PR #20, opened 2014-11-25
Salt localsettings

dannyroberts:
This is to the point where a fab preindex is working on it (as long as you're running `fab development awesome_deploy --set code_branch=EMAIL_USE_TLS`, until the `EMAIL_USE_TLS` branch gets merged in.)

It's failing on compressor stuff now. I think it's fair to review this and merge now, and just accept it's not fully complete. To note:

- `fab development deploy` (from the [`fab-ansible` branch of `commcare-hq`](https://github.com/dimagi/commcare-hq/compare/fab-ansible)) fails on `python manage.py compress`, which is due to lessc not being installed on the proxy. To different versions need to be installed: the latest by git cloning the lessc repo to /opt/lessc and 1.3.0 through npm (see [commcare-hq README](https://github.com/dimagi/commcarehq-ansible/blob/master/README.md) for details).
- If you do comment out the compressor related stuff, apache static files now 404. Not sure what's up with that; I thought it was a permissions thing and started to debug here (https://github.com/dimagi/commcarehq-ansible/commit/f061019b). (It's also definitely possible that I'm wrong and I haven't commented out all of the proper localsettings.)


dannyroberts:
Seeing as there are other branches working off of this, I'm going ahead and merging. @cewing thanks for your comments, and I'm looking into that cyclical dependency you mentioned. I remember bumping into it and "working around" it, but I'll scrutinize my "solution" a little more and come up with something that works.

cewing:
It's one of those things that I think only shows up when you are first running the setup.  Once the users exist, you won't see it again.  In fact, the second half of the dependency I have not yet verified to break things, because I had the setup running by the time I noticed it.

dannyroberts:
I so far haven't been able to reproduce any error, but I agree this is a hacky/fragile setup

PR #21, opened 2015-01-02
Subtree merge bennojoy.*

dannyroberts:


dannyroberts:
@TylerSheffels I think this should be our approach to including things from galaxy. It lets us edit their code while theoretically still allowing us to merge in their new code if that's something we ever want to do. See https://help.github.com/articles/about-git-subtree-merges/

PR #23, opened 2015-01-06
Rabbitmq

TylerSheffels:
@dannyroberts

### DIFF
--- a/ansible/vars/dev.yml
+++ b/ansible/vars/dev.yml
@@ -9,7 +9,7 @@ ECRYPTFS_PASSWORD: 'xxBvYiMZ3YVIVkFR'
 KSPLICE_ACTIVE: no
 KSPLICE_ACTIVATION_KEY:
 
-AMQP_USER: hq_worker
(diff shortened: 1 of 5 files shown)


ansible/vars/dev.yml
```diff
@@ -9,7 +9,7 @@ ECRYPTFS_PASSWORD: 'xxBvYiMZ3YVIVkFR'
 KSPLICE_ACTIVE: no
 KSPLICE_ACTIVATION_KEY:
 
-AMQP_USER: hq_worker
+AMQP_USER: "{{ deploy_env }}_worker"
```
dannyroberts: Is this right? On prod is this `www_worker` and on staging it's `staging_worker` and on india it's `india_worker`?

TylerSheffels: On staging its `staging_worker`, on prod its `hq_worker`.  I had only checked staging.  I believe that a full `ansible-playbook deploy_stack.yml ...` call would appropriately switch the worker name though.  It would add the user in rabbitmq, then update the localsettings to point at that new user.

PR #27, opened 2015-01-26
set max_local_storage_nodes to 1

czue:
@dannyroberts - this will prevent the machine from accidentally booting multiple versions of elasticsearch


czue:
i noticed some problems with this so closing for now (it's been undone on the india server)

PR #30, opened 2015-01-27
tag users role

dannyroberts:


dannyroberts:
@TylerSheffels

TylerSheffels:
@dannyroberts is there anywhere you are tracking what tags are available?

dannyroberts:
I'm not, but that's a great idea. I'll add to README.

PR #31, opened 2015-01-27
Activate New Relic

dannyroberts:
@TylerSheffels this includes your suggestion of listing tags in README

### DIFF
--- a/ansible/roles/common/meta/main.yml
+++ b/ansible/roles/common/meta/main.yml
@@ -5,4 +5,5 @@ dependencies:
   - role: supervisor
   - {role: users, tags: users}
   - role: edit
-  - role: ksplice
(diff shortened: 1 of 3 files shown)


ansible/roles/common/meta/main.yml
```diff
@@ -5,4 +5,5 @@ dependencies:
   - role: supervisor
   - {role: users, tags: users}
   - role: edit
-  - role: ksplice
+  - {role: ksplice, tags: ksplice}
+  - {role: newrelic, tags: newrelic, when: NEW_RELIC_KEY is defined}
```
TylerSheffels: I really like this pattern.

dannyroberts: Yeah I like it too. It's taking me a while to figure out good idioms for ansible.

PR #32, opened 2015-01-30
Clean inventory

dannyroberts:
Get rid of redundant groups


ansible/inventories/rs-danny-test
```diff
@@ -27,18 +27,6 @@
 [elasticsearch]
```
TylerSheffels: Does it make sense to remove this file, as I deleted these RS machinces?

dannyroberts: So much sense

dannyroberts:
@TylerSheffels g2g

PR #37, opened 2015-02-02
add newrelic.ini file

snopoke:
@dannyroberts @TylerSheffels 

Not 100% sure this is the best way to do this but thought I'd give it a go. Setup should be as follows:

Each machine that is running a python process which reports to NewRelic has a `newrelic.ini` file in the `www_home` folder. The contents of the files are basically the same except for the name which should vary according to host function e.g. celery, django, pillowtop.

### DIFF
--- a/ansible/inventories/development
+++ b/ansible/inventories/development
@@ -1,5 +1,5 @@
 [webworkers]
-192.168.33.15
+192.168.33.15 newrelic_app_name=django
 
 [postgresql]
 192.168.33.16
@@ -17,7 +17,7 @@
 192.168.33.16
 
 [celery]
-192.168.33.16
+192.168.33.16 newrelic_app_name=celery
 
 [elasticsearch]
 192.168.33.16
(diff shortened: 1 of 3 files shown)


ansible/inventories/development
```diff
@@ -1,5 +1,5 @@
 [webworkers]
-192.168.33.15
```
TylerSheffels: @snopoke I think you want to add a file in https://github.com/dimagi/commcarehq-ansible/tree/master/ansible/group_vars where we define this variable.  That way it will get picked up for all environments, alternatively this might be the right place to define it if we don't want to install newrelic on staging/india/development.  @dannyroberts will know better than me.

ansible/roles/newrelic/tasks/main.yml
```diff
@@ -17,3 +17,12 @@
 
 - name: Make sure newrelic-sysmond has started
   service: name=newrelic-sysmond state=started
+
+- name: copy newrelic ini file
```
TylerSheffels: Do we need to restart the newrelic daemon after this?

dannyroberts:
@snopoke, nice! I'm curious what prompted this, and whether we're going to need to run this once merged, or whether it's forward-looking

snopoke:
This is completely unrelated to the newrelic daemon that runs on the
servers. This file provides config to the newrelic agent that 'runs' the
python app (celery or django) and collects app level data e.g. transaction
times etc.

PR #38, opened 2015-02-03
Supervisor upstart

dannyroberts:


dannyroberts:
@TylerSheffels

dannyroberts:
Just tested this on the india cloud, as well as hqpillowtop0, and it works

dannyroberts:
@snopoke

ansible/roles/supervisor/tasks/main.yml
```diff
@@ -1,24 +1,56 @@
...
   sudo: yes
-  sudo_user: "{{ cchq_user }}"
-  notify:
-    - stop supervisor
-    - start supervisor
+  template:
+    src: supervisord.upstart.conf.j2
+    dest: /etc/init/supervisord.conf
+
+- name: restart supervisor to pick up configuration changes
```
snopoke: Will this execute every time this play gets run?

dannyroberts: It will, just because it's annoying and error prone to try to figure out whether something needs to be stopped and started, just started, or neither. When I was running this a lot, it kept landing in an almost-there state where the only thing I needed to do was restart it. I think once this is part of the real workflow, we can revisit and optimize this?

dannyroberts: (this is from the commit called "just always restart supervisord" :smile:)

snopoke: Fair enough

PR #40, opened 2015-02-03
break out dev_group name as a variable

dannyroberts:
needs to be set to something different on prod. (Matters very little)


dannyroberts:
@esoergel

PR #42, opened 2015-02-03
fix change that made supervisor not install

dannyroberts:


dannyroberts:
@snopoke @esoergel

ansible/roles/supervisor/tasks/main.yml
```diff
@@ -4,6 +4,7 @@
   shell: 'which supervisord'
   register: which_supervisord
   changed_when: false
+  failed_when: which_supervisord.rc != 0 and which_supervisord.rc != 1
```
snopoke: I'm not sure I understand why this should fail if `rs != 1`. Doesn't an rc of 1 mean that it failed to find it?

snopoke:
Merged anyway

ansible/roles/supervisor/tasks/main.yml
dannyroberts: The point is that I'm trying to calculate whether supervisor is installed, so I don't want it to _fail_ if it's not installed, I want to be able to use the answer in the next tasks. (If a task fails, it doesn't move on to the next one.)

PR #47, opened 2015-02-06
deprecate salt localsettings workflow

dannyroberts:
localsettings.salt.py.j2 is just renamed to localsettings.py.j2 (and the old localsettings.py.j2 which wasn't being used is deleted)


dannyroberts:
@snopoke

PR #49, opened 2015-02-06
fix error detecting whether repo exists

dannyroberts:
Previously this was just always updating the repo, which we don't want it to do yet (that's what deploy is for)


dannyroberts:
@snopoke

PR #50, opened 2015-02-06
remove functionality in webworkers we are not using

dannyroberts:
- some of it was just logically wrong (like running syncdb on each webworker)
- some of it we are just not ready for (updating webworkers one at a time,
  doesnt make sense because we actually deploy with fab now


dannyroberts:
@snopoke (didn't have a chance to test this but I've been wanting to do this cleanup for a while)

snopoke:
looks good

PR #51, opened 2015-02-10
add cron job to delete old touchforms sessions

snopoke:
http://manage.dimagi.com/default.asp?155377#870144

@dannyroberts @TylerSheffels


czue:
nice

PR #58, opened 2015-03-20
Purge old restore files

benrudolph:
Note, this is dependent on this PR: https://github.com/dimagi/commcare-hq/pull/5956

@TylerSheffels

### DIFF
--- a/ansible/roles/commcarehq/tasks/main.yml
+++ b/ansible/roles/commcarehq/tasks/main.yml
@@ -76,3 +76,11 @@
     - { path: "{{ preindex_home }}", deploy_env_name: "{{ deploy_env }}_preindex" }
   tags:
     - localsettings
+
+- name: Create payload purging cron job
+  sudo: yes
+  cron:
+    name: "Purge old restore payload files"
+    special_time: daily
+    job: "find {{ restore_payload_dir }} -name \"*.xml\" -maxdepth 1 -mtime +4 -delete"
+    user: root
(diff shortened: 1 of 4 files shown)


czue:
i'll let @TylerSheffels or @dannyroberts review, but i think it's fine to merge this before https://github.com/dimagi/commcare-hq/pull/5956

dannyroberts:
Sweet, @benrudolph! What machine do you want this deployed to? Django workers? I think it'd make sense to either make a new role for this or add it to a role that only gets deployed to the inventory group you're targeting.

benrudolph:
hmm not too sure. i'm assuming it'd have to be deployed to all machines which run the application so i'm guessing that'd be django workers? is that the `webworker` role? makes sense to only deploy to that. let me know if we have that role (not too familiar with the setup), else i'll make a new one

dannyroberts:
Yeah, so https://github.com/dimagi/commcarehq-ansible/blob/64f0be510930ef74f53224dd9c72de6a3a043a37/ansible/deploy_webworker.yml is the entrypoint for things that run only on webworkers (the line `hosts: webworkers` is say, run these roles on the `webworkers` inventory group).

dannyroberts:
You can either add to the `webworkers` role, or add your own role ('commcaresync' or whatever) and make sure to reference that from this file.

benrudolph:
thanks for the pointers, will look today

benrudolph:
Decided to just add the task to the `webworker` role since it seemed like overkill to make an entirely new role for one minor cron job

dannyroberts:
Ok, sounds good!

benrudolph:
@dannyroberts i think this is now good to go

dannyroberts:
Awesome! Thanks for being persistent in working out the kinks!

snopoke:
Nice

PR #59, opened 2015-03-25
Restore payload tasks

benrudolph:
This smells bad to me, but I couldn't figure out a better way to do this. I'm trying to accomplish two things: one, setup the `restore_payloads` directory and a cron job to periodically clean the dir; two, fill the `RESTORE_PAYLOAD_DIR` variable in localsettings template. Originally I had the `restore_payload_dir` variable in the `webworker` role and that worked fine, but since the localsettings template is under the `commcarehq` role, i can't access that variable. I didn't want to declare the variable twice and couldn't figure out a way to share variables between roles. The reason this works for touchforms is that the task of creating the directory is in the same role as creating the localsettings file. This solution just declares a global `restore_payload_dir` variable, but that seems like overkill. let me know if you have a better solution @dannyroberts @TylerSheffels

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -214,6 +214,8 @@ SOIL_DEFAULT_CACHE = "redis"
 
 REPORT_CACHE = 'redis'
 
+RESTORE_PAYLOAD_DIR = {{ restore_payload_dir }}
+
 memcached = {
     'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
     'LOCATION': '{{ localsettings.MEMCACHED_HOST }}:{{ localsettings.MEMCACHED_PORT }}',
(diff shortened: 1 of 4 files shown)


TylerSheffels:
:+1: from me

dannyroberts:
@benrudolph I know how you feel, but this seems like as good a solution as any! Let's go with it

dannyroberts:
Ansible kind of does that to you—either you can use a variable in one place, or else it's basically a global. Us programmers are used to being able to pass variables around in clean ways through function calls :)

ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -214,6 +214,8 @@ SOIL_DEFAULT_CACHE = "redis"
 
 REPORT_CACHE = 'redis'
 
+RESTORE_PAYLOAD_DIR = {{ restore_payload_dir }}
```
snopoke: @benrudolph @dannyroberts This should be in quotes

dannyroberts: Oops I'm sorry I didn't catch that in review!

PR #62, opened 2015-03-27
add segment analytics private variable

sravfeyn:
For https://github.com/dimagi/commcare-hq/pull/6067

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -158,7 +158,8 @@ DEFAULT_PROTOCOL = "https"
 
 ANALYTICS_IDS = {
     'GOOGLE_ANALYTICS_ID': '{{ localsettings.GOOGLE_ANALYTICS_ID }}',
-    'PINGDOM_ID': '{{ localsettings.PINGDOM_ID }}'
+    'PINGDOM_ID': '{{ localsettings.PINGDOM_ID }}',
+    'SEGMENT_ANALYTICS_KEY': '{{ localsettings.SEGMENT_ANALYTICS_KEY }}',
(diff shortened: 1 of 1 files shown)


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -158,7 +158,8 @@ DEFAULT_PROTOCOL = "https"
 
 ANALYTICS_IDS = {
     'GOOGLE_ANALYTICS_ID': '{{ localsettings.GOOGLE_ANALYTICS_ID }}',
-    'PINGDOM_ID': '{{ localsettings.PINGDOM_ID }}'
+    'PINGDOM_ID': '{{ localsettings.PINGDOM_ID }}',
+    'SEGMENT_ANALYTICS_KEY': '{{ localsettings.SEGMENT_ANALYTICS_KEY }}',
```
czue: fyi @sravfeyn @kaapstorm if you add an environment variable like this you need to make sure it exists in all environments (this was missing from india) or ansible will choke on it. you can do a more protective style like https://github.com/dimagi/commcarehq-ansible/pull/70/ if you don't want to add it everywhere

TylerSheffels:
@czue @dannyroberts I think this is actual showing a larger issue, that when we add a new localsettings we need a better process for adding that settings to each environment in the configuration.  A more conservative approach like #70 will choke on any code that relies on that settings in hq, right?

czue:
it depends - if you put a sensible default in `settings.py` then leaving it out entirely is fine. for something that's required and can't have a reasonable default then it will choke

sravfeyn:
Sorry and thanks cory. Added this note to https://confluence.dimagi.com/display/internal/Ansible

PR #63, opened 2015-03-31
add variables for new invoice template

nickpell:
For http://manage.dimagi.com/default.asp?157984

@benrudolph

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -489,3 +489,6 @@ SMS_PREFIXES_BY_COUNTRY = '{{ localsettings.SMS_PREFIXES_BY_COUNTRY }}'
 # See corehq/apps/builds/README.md for more information.
 BASE_ADDRESS = '{{ SITE_HOST }}'
 EMAIL_USE_TLS = {{ localsettings.EMAIL_USE_TLS }}
+
+BANK_ROUTING_NUMBER_ACH = '{{ localsettings.BANK_ROUTING_NUMBER_ACH }}'
+BANK_ROUTING_NUMBER_WIRE = '{{ localsettings.BANK_ROUTING_NUMBER_WIRE }}'


benrudolph:
looks good. just checking, the actual variables have been added to the server?

benrudolph:
(merging cause the comment doesn't make a different)

nickpell:
@benrudolph Not yet, going to try that now that this is merged (first time working on ansible)

ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -489,3 +489,6 @@ SMS_PREFIXES_BY_COUNTRY = '{{ localsettings.SMS_PREFIXES_BY_COUNTRY }}'
 # See corehq/apps/builds/README.md for more information.
 BASE_ADDRESS = '{{ SITE_HOST }}'
 EMAIL_USE_TLS = {{ localsettings.EMAIL_USE_TLS }}
+
+BANK_ROUTING_NUMBER_ACH = '{{ localsettings.BANK_ROUTING_NUMBER_ACH }}'
+BANK_ROUTING_NUMBER_WIRE = '{{ localsettings.BANK_ROUTING_NUMBER_WIRE }}'
```
czue: fyi @nickpell @benrudolph  if you add an environment variable like this you need to make sure it exists in all environments (this was missing from india) or ansible will choke on it. you can do a more protective style like #70 if you don't want to add it everywhere

nickpell: ahh, thanks

czue: if we need to add something specific we should do that - i just deployed leaving them blank

nickpell: that's fine, most of billing (invoicing in particular) is unused on india.

benrudolph: i was wondering what all those if statements were supposed to be doing

PR #64, opened 2015-04-01
Setup shared directory

benrudolph:
@dannyroberts @TylerSheffels @biyeun 

Working on setting up a shared directory for our web workers (see this for context as to why: https://github.com/dimagi/commcare-hq/pull/5956). I've been following this as a guide: https://www.digitalocean.com/community/tutorials/how-to-set-up-an-nfs-mount-on-ubuntu-12-04. This is a WIP, but I think it's somewhat ready to be tested on staging. Shouldn't really have any negative side effects. You'll notice I setup my role a bit differently than the others, this made it easier to reuse tasks and  use better encapsulation. I got the idea from this repo: https://github.com/gc3-uzh-ch/ansible-playbooks/tree/981ac476a38b24e391ff969549dab260d7dcd915


ansible/roles/nfs_shared_dir.yml
```diff
@@ -0,0 +1,17 @@
+---
+
+- name: SharedDirHost
+  hosts: shared_dir_host  # Should be one
+  sudo: yes
+  vars_files:
```
benrudolph: FYI we can use this pattern to share variables between two roles and not put them in the global variable file

ansible/roles/shared_dir.yml
```diff
@@ -0,0 +1,17 @@
+---
+
+- name: SharedDirHost
+  hosts: shared_dir_host  # Should be one
+  sudo: yes
+  vars_files:
```
benrudolph: FYI we can use this pattern to share variables between two roles and not put them in the global variable file

dannyroberts: Not sure I see how this shares variables between roles. Seems like this vars file is only in this role. Are you just saying this is a way to do it in the future if needed?

benrudolph: yea exactly, like i could have specified a different directory

benrudolph:
works! :fireworks: 

@TylerSheffels @dannyroberts I figured out a better solution. I _believe_ the reason it wasn't working before under the same group id was that when a client would create a file, it'd set the group to the anonymous group type which did not have permissions to write to the mount. You can set a setting that lets you choose the anonymous gid, which i conveniently set to the gid of nfs. Now any user part of the nfs group can read and write to the shared mount and users not part of the group, cannot

benrudolph:
this is so cool!

dannyroberts:
Haha, yeah this is sweet :)

PR #66, opened 2015-04-02
Nginx

TylerSheffels:
Add support for Nginx as a webserver.  Currently running on proxy3 as a test environment.

@snopoke @dannyroberts @benrudolph


ansible/roles/nginx/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+  notify:
+   - reload nginx
+
+- name: Create the links to enable site configurations
+  sudo: yes
+  file: path=/etc/nginx/sites-enabled/{{ item['server']['file_name'] }} state=link src=/etc/nginx/sites-available/{{ item['server']['file_name'] }}
+  with_items: nginx_sites
+  when: nginx_sites|lower != 'none'
+
+- include_vars: '{{deploy_env}}.yml'
```
TylerSheffels: I'm not incredibly proud of this method for including variable configurations for each environment.  I chose this because:
* It let's us use one nginx template
* It let's us control site configurations by environment (e.g. ones that depend on inventory setup)
* It also lets us deploy sites that are independent of inventory (e.g. slow.)

ansible/roles/nginx/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+
+- name: Create the configurations for sites based on inventory
```
benrudolph: How is this different from "Create the configurations for sites independent of inventory"

very possible i'm missing something

ansible/roles/nginx/tasks/main.yml
```diff
@@ -0,0 +1,77 @@
...
+  sudo: yes
+  template: src=site.j2 dest=/etc/nginx/sites-available/{{ item['server']['file_name'] }}
+  with_items: nginx_sites
+  when: nginx_sites|lower != 'none'
+  notify:
+   - reload nginx
+
+- name: Create the links to enable site configurations
```
benrudolph: Looks the same as the one above

ansible/roles/nginx/vars/production.yml
```diff
@@ -0,0 +1,96 @@
...
+   location1:
+    name: /
+    return: "301 https://$server_name$request_uri"
+   location2:
+    name: "~* /(?:home|users|support|plans|images|pdfs|style|js|tour|div2|docs|fr|poc|es)/"
+    alias: "{{ public_site_path }}/$1"
+    try_files: "$uri $uri/index.html $uri/ =404"
+- server:
+   file_name: commtrack
+   listen: "8443 ssl"
```
benrudolph: was this meant to be 8443?

benrudolph: jk it was

benrudolph:
nice push on this. don't have too much feedback for the actual settings used for nginx, but they all seem reasonable to me

ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -0,0 +1,29 @@
...
+}
+
+
+http {
+
+        include /etc/nginx/mime.types;
+        default_type application/octet-stream;
+
+        access_log {{ log_home }}/{{ nginx_access_log_name }};
+        error_log {{ log_home }}/{{ nginx_error_log_name }};
```
benrudolph: may want to add `warn` just in case? at least at the beginning. i don't actually know if we use warn in our code though

ansible/roles/nginx/tasks/main.yml
TylerSheffels: The line I commented on above includes the variables file for that env, e.g. `production.yml` below.  It's a bit confusing and I'm open to a better suggestion.

ansible/roles/nginx/tasks/main.yml
TylerSheffels: Same comment as above

ansible/roles/nginx/templates/nginx.conf.j2
TylerSheffels: Good thought, will update.

ansible/roles/nginx/vars/production.yml
TylerSheffels: Yeah, looks like you saw this, but Commtrack hits a separate proxy before it goes to hqproxy.

TylerSheffels:
@benrudolph Once we feel decently good about this pr my plan is to switch staging to point at it and then run testing on staging.

benrudolph:
comments make sense. i'm cool with that testing plan, i'll let @dannyroberts or @snopoke take a look before merging

dmyung:
Does nginx still require the chunkin module to be compiled and installed separately? Or is the chunked encoding posts not a concern anymore? The reason why we've been on apache this long has bee because nginx didn't support chunked encoding at all for the phone submissions. Apache's mod_proxy would buffer and add a content length to let django accept the submissions (apache receives as http/1.1, buffers, and forwards to django as http/1.0). 

So unless django can handle 1.1 requests better with zero content-length, or nginx out of the box has the chunkin handler to buffer and add content length, I'd make sure to test this extensively with the phone submissions.

dannyroberts:
@dmyung, @czue can chime in, but I'm pretty sure nginx now supports this better than apache does (due to progress in the last 4 years)

snopoke:
@benrudolph I won't be able to look at this till Tuesday so don't wait for
me

czue:
@dmyung yeah nginx just supports this nicely out of the box now

ansible/roles/nginx/tasks/main.yml
dannyroberts: Is it possible for these variables to be stored in the same place as the other private variables?

ansible/roles/nginx/vars/main.yml
```diff
@@ -0,0 +1,21 @@
+---
+
+ubuntu_pkg: nginx
+
+nginx_sites:
```
dannyroberts: I don't understand why this is the default

dannyroberts:
In general I'm trying to understand the organization here, and it's kind of challenging. I don't want to derail your timeline for deploying this in production, but it would be nice for the design to be easier to understand and modify.

The ideal design in my mind would be to have some preset types of configs (cchq, commtrack, hq_status, tableau, etc.) available, and then to have each environment declare what it'd like to have deployed in some mixture of its inventory and private variables file. This might include having to call the nginx role with specific sites to deploy (`cchq=yes` or `commtrack=yes`) each time it's called or something.

It's up to you whether and when you want to do this depending on what the timeline is like and whether this feels easy or daunting. Feel free to consult with me on implemention and @fyniac on timeline

TylerSheffels:
@dannyroberts Updated to use individual files for each site config

TylerSheffels:
Not ready to go yet, I'll not here when it is.

dannyroberts:
fyi just rebased due to https://github.com/dimagi/commcarehq-ansible/pull/71 (but didn't squash any commits :))

TylerSheffels:
@dannyroberts This should be ready for a second look now.

dannyroberts:
@TylerSheffels, I sort of skimmed this since we talked through a lot of it together. Can you add to the README a section a basic outline of how to configure and deploy nginx sites?

TylerSheffels:
@dannyroberts Added readme

ansible/roles/nginx/README.md
```diff
@@ -0,0 +1,21 @@
...
+`proxy_type: (apache2|nginx)`
+
+## Adding new nginx sites
+
+* Add a new file in vars/<site_name>
+  * Each file here corresponds to an nginx site that will listen on a port/subdomain combination, e.g. slow.commcarehq.com:80
+* Add a new line in ../../deploy_proxy.yml
+  * `- { role: nginx, when: proxy_type == 'nginx' and active_sites.<site_name> == True, action: site, site_name: <site_name> }`
+* Add to the active sites list in the all the config yaml files
+  * `cchq_static: True`
```
dannyroberts: `<site_name>: True`?

dannyroberts:
Thanks, this looks sweet!

PR #68, opened 2015-04-06
Refactor restore payload

benrudolph:
Don't think we need this `java` dependency anymore

@dannyroberts

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -13,6 +13,5 @@ restore_payload_dir: "{{ shared_mount_dir }}/restore_payloads"
 
 cchq_user: cchq
 dev_group: dev
-shared_dir_groupname: nfs
 
 django_port: 9010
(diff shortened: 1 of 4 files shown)


benrudolph:
It allows the cchq user to write to the folder since nogroup anonymous id is changed to 1500 (that was set in the other PR). So nogroup is effectively the `nfs` group. I have no idea why it doesn't work to just set the group as `nfs`. I have no idea why that isn't essentially the same thing, but it fails on the `chown` when I try it

dannyroberts:
Is it possible for the files themselves to be owned by the nfs group? Or do those also get written as nobody:nogroup?

benrudolph:
they get written as nobody:nogroup when written by the `cchq` user on the client machiens

benrudolph:
i believe, if you are the host machine you can own the files as `nfs`

dannyroberts:
Interesting. Well, thanks for the explanation and for digging into all this!

PR #69, opened 2015-04-08
Refactor nfs

benrudolph:
@dannyroberts 

This does a couple things. The dir permissions are pretty obvious. The other thing is `all_squash`.  This `all_squash` makes things work. Not in the most ideal way though. `all_squash` - "Convert incoming requests, from ALL users, to the anonymous uid and gid.". Unfortunately nfs does not have an option that says "Convert requests, from SOME users, to the anonymous uid and gid." So if we don't want to have to write as root, we have to allow every user on the machine to write to the shared dir. I tried really hard to make this work in such a way that only `cchq` could write the files, but short of syncing all the uids on all the servers, I think it's impossible. Here's a quote from one of the trouble shooting [guides](http://nfs.sourceforge.net/nfs-howto/ar01s07.html) I read:

"If you cannot have a single uid for all instances of a username, suboptimal steps must be taken. In some instances you could make the directory and files world-readable, thereby enabling all users to read it. It could also be made world-writeable, but that's always a bad idea. It could be mounted all_squash with a specific anonuid and/or a specific anongid to cure the problem, but once again, at least from the NFS viewpoint, that's equivalent to making it world readable or writeable."

Why it was working before:
Not quite sure, but I think I changed the permissions using root at one point and made the shared directory owned by the users on the staging webworkers.

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -8,6 +8,7 @@ service_home: "{{ cchq_home }}/services"
 virtualenv_home: "{{ www_home }}/python_env"
 virtualenv_preindex_home: "{{ www_home }}/python_env_preindex"
 
+shared_dir_gid: 1500  # This GID cannot already be allocated
 shared_mount_dir: "/mnt/shared_data"
 restore_payload_dir: "{{ shared_mount_dir }}/restore_payloads"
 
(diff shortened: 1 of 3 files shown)


dannyroberts:
@benrudolph looks good, haven't totally wrapped my head around why we still need a `shared_dir_gid` when everything's being accessed through anonymous user/groups. Maybe once this is done it'd be worth writing up a kind of retro/summary of what you learned about NFS and roadblocks you hit along the way, and how you ended up where you ended up. I think that'll not only be informative to curious onlookers, but also will likely provide really good documentation to the next person who needs to dive in and solve some problem (possibly even your future self)

PR #72, opened 2015-04-10
add missing environment var to dev env

czue:
@dannyroberts is this the right place to do this?

context: https://github.com/dimagi/commcarehq-ansible/pull/62/files#r28164439


dannyroberts:
Yeah that looks right

dannyroberts:
That file doesn't affect production in any way, but is a good example config file

PR #74, opened 2015-04-15
tag ecryptfs role

dannyroberts:
@benrudolph something like this


benrudolph:
ah just saw this

benrudolph:
does this then add a tag to run all the tasks in main.yml? if so, this is the better solution

dannyroberts:
@benrudolph yeah this is like shorthand for adding `tags: ecryptfs` to every task in the role

PR #75, opened 2015-04-15
Add tags for mounting ecrypt drive

benrudolph:
@dannyroberts @orangejenny 

We just used this for remounting the drive. It'll probably be useful in the future

code buddy: @millerdev


benrudolph:
also, fyi, @czue

benrudolph:
https://github.com/dimagi/commcarehq-ansible/pull/74

PR #77, opened 2015-04-16
Pg logging

benrudolph:
@snopoke @dannyroberts 

cc: @millerdev 

This enables logging for postgres. Most of these values are the default settings. Important ones that aren't:
```
log_destination = 'csvlog' # this logs the statements in a csv format (default is stderr)
log_min_duration_statement = 300 # logs queries over 300ms
```

### DIFF
--- a/ansible/deploy_db.yml
+++ b/ansible/deploy_db.yml
@@ -14,7 +14,7 @@
   hosts: postgresql
   sudo: yes
   roles:
-    - postgresql
+    - {role: postgresql, tags: 'postgresql'}
 
 - name: Couchdb
   hosts: couchdb
--- a/ansible/roles/postgresql/templates/postgresql.conf.j2
+++ b/ansible/roles/postgresql/templates/postgresql.conf.j2
@@ -277,21 +277,21 @@ shared_buffers = 24MB           # min 128kB
 
 # - Where to Log -
 
-#log_destination = 'stderr'     # Valid values are combinations of
+log_destination = 'csvlog'     # Valid values are combinations of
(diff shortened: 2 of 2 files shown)


millerdev:
Turning on logging can seriously affect the performance of Postgres, depending on disk configuration. For production systems I believe it's generally recommended to put logs on a separate physical volume from data. Do we have any way of testing if/how this will affect performance of our database?

czue:
@millerdev of course.

image

benrudolph:
lol

czue:
(half kidding, is a good suggestion - not sure if @benrudolph or @snopoke want to chime in)

benrudolph:
doing some research now. there's a decent article here: http://www.depesz.com/2013/09/05/what-is-the-overhead-of-logging/. i think it's OK to try on staging, but it's hard to measure performance of logging vs. not logging when you're not logging haha

snopoke:
Logging to a separate disk is definitely a good idea. Query running will
increase CPU load but don't think that's an issue on hqdb0. Also having
specified the timing threshold we shouldn't be generating that many logs.

Can always deploy this to staging and run some benchmarks.

benrudolph:
Been looking into writing logs to a different disk. Any suggestions as to where to put the logs? One article I read suggested using rsyslog to write logs to a different server that way the performance hit is on the network vs the I/O. This makes sense to me because hqdb0 is always pushing its limits in terms of I/O so we might as well try and reduce that. @dannyroberts @snopoke

snopoke:
I don't have any experience with this but rsyslog looks good. Have you
looked at syslog-ng for comparison?

PR #78, opened 2015-04-16
Tags for updating es conf

benrudolph:
@dannyroberts 
cc: @millerdev


millerdev:
Not sure how that works, but :smile: :+1:

dannyroberts:
@millerdev tagging individual tasks lets you then run the `ansible-playbook` command with `--tags=es_conf` at the end to only run those 4 tasks, skipping everything else

PR #82, opened 2015-04-16
Postgres conf hostnames

benrudolph:
@dannyroberts this allows ansible to lookup the IP address of a hostname. Requires the version bump in ansible.

Other changes:
- Change `host` to `hostssl` (this is what we use in our prod environment)
- Allow access to postgres via devreadonly user (also used in prod environment)

The domain lookup was tested using my Vagrant machine

### DIFF
--- a/ansible/requirements.txt
+++ b/ansible/requirements.txt
@@ -1,2 +1,4 @@
 ansible==1.9.0.1
+netaddr
+dnspython
 passlib
--- a/ansible/requirements.txt
+++ b/ansible/requirements.txt
@@ -1,2 +1,2 @@
-ansible==1.8.2
+ansible==1.9.0.1
 passlib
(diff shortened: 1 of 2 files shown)


benrudolph:
cc: @millerdev

czue:
neat. does this mean india isn't setup with ssl? is that a problem?

benrudolph:
that's correct. as to whether it's a problem, it's probably not the best practice, but i don't think it's too big of a deal. since the pg connection is from inside the network doesn't that mean you'd have to be vpn'd in to possibly snoop on the connection?

czue:
cool, yeah I agree with that logic, though not sure i've fully thought through the consequences.

dannyroberts:
Excerpt from http://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html:

> `hostssl`
>
> This record matches connection attempts made using TCP/IP,
> but only when the connection is made with SSL encryption.

Presumably you've checked that all connections to postgres are indeed made over SSL? If so this looks good!

benrudolph:
I haven't explicitly, and I'm not really sure how. However our prod db uses `hostssl` it so I'm assuming that's the standard we should submit to

benrudolph:
(also ran this on staging and no issues with the ssl connection)

dannyroberts:
ok sounds good

PR #83, opened 2015-04-17
Update postgres conf regardless of pg_install

benrudolph:
@dannyroberts


dannyroberts:
Having trouble finding where `pg_install` ever isn't `True`—what made you find this?

benrudolph:
It's in the `default.yml` file and it is always true. when i run the command i change it to `False` though i'm sure there much be a way to specify the variable in the command line

dannyroberts:
Ah, I see. And yeah there is, I believe it's `-e pg_install=False`

PR #84, opened 2015-04-17
Restart supervisor when PG updates con

benrudolph:
@czue


czue:
seems convenient, though it might be nicer to only do this on celery and pillowtop roles. is that possible?

benrudolph:
ah, i think this might not even work on prod. this will run hqdb0 which is fine for staging but not prod. also yea, it should be possible to do that, just need to figure out the command to only restart celery and pillowtop

PR #86, opened 2015-04-22
Add handling to nginx to route 502 + 503 errors to a static page

TylerSheffels:


TylerSheffels:
@snopoke This is an old pr, but nginx is merged now so this should be good to go

PR #87, opened 2015-04-22
Redirect to .org

benrudolph:
@TylerSheffels would this work?
cc: @emord @millerdev


TylerSheffels:
@benrudolph I think you want to do something like https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/nginx/vars/cchq_redirect.yml and https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/deploy_proxy.yml#L16

The downside is that this requires adding a couple variables to the config settings

benrudolph:
thanks for the info. closing for now until i have opportunity to fix

benrudolph:
k should be good, just need to update the variables on salt

TylerSheffels:
@benrudolph You will also need to add in a line in deploy_proxy.yml: 
https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/deploy_proxy.yml#L16

benrudolph:
yep thanks, i strangely woke up last night and that was at the forefront my mind

PR #88, opened 2015-04-29
add gettext in common installs

dannyroberts:


czue:
thanks!

czue:
can you ping on https://github.com/dimagi/commcare-hq/pull/6481 when this is g2g?

PR #91, opened 2015-05-04
More npm tasks

benrudolph:
Had to do this on the fly today during deploy, but I think we should have it standardized in our deploy that'd be optimal. `n` allows you to update `node` really easily

### DIFF
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -34,7 +34,10 @@
   with_items:
     - less@1.3.1
     - n@1.3.0
-    - node@0.10.29
+
+- name: Update Node
+  command: 'n 0.10.29'
+  sudo: yes
 
 - name: Install lessc from source
   sudo: yes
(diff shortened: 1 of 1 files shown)


benrudolph:
@dannyroberts @TylerSheffels 
cc: @millerdev

czue:
:+1: from me but will let an ansible person merge

ansible/roles/common_installs/tasks/main.yml
```diff
@@ -33,6 +33,11 @@
   sudo: yes
   with_items:
     - less@1.3.1
+    - n@1.3.0
```
dannyroberts: Does this really install an executable called `n` on your path!? That's such a... presumptuous name for a utility for managing node's version...

benrudolph: haha yup: http://davidwalsh.name/upgrade-nodejs
i guess it's so presumptuous that no one does it and thus isn't used

millerdev: The comment stream on that page you linked to is less than inspiring. Isn't there a way to do this with the system package manager (and doesn't ansible know all about such things)?

PR #92, opened 2015-05-06
Allowing RC4 cipher in order for older phones to connect

TylerSheffels:
Hard to see actual change, only thing was removing `!` before RC4


czue:
nice! does this fix things?!

TylerSheffels:
Yes, this allows the Nokia phones to connect

czue:
wild

PR #93, opened 2015-05-06
Update pg conf

benrudolph:
@snopoke @dannyroberts 
cc: @esoergel 

Used these as reference:
http://pastebin.com/pXk09qte - `pg_hba.conf` diff
http://pastebin.com/kDRYdZqJ - `postgres.conf` diff

needs: https://github.com/dimagi/commcare-hq/pull/6590

### DIFF
--- a/ansible/roles/postgresql/templates/postgresql.conf.j2
+++ b/ansible/roles/postgresql/templates/postgresql.conf.j2
@@ -65,7 +65,7 @@ listen_addresses = '*'      # what IP address(es) to listen on;
                     # defaults to 'localhost', '*' = all
                     # (change requires restart)
 port = 5432             # (change requires restart)
-max_connections = 100           # (change requires restart)
+max_connections = 300           # (change requires restart)
 # Note:  Increasing max_connections costs ~400 bytes of shared memory per
 # connection slot, plus lock space (see max_locks_per_transaction).
 #superuser_reserved_connections = 3 # (change requires restart)
@@ -110,7 +110,7 @@ ssl = true              # (change requires restart)
 
 # - Memory -
 
-shared_buffers = 24MB           # min 128kB
+shared_buffers = 1024MB           # min 128kB
                     # (change requires restart)
 #temp_buffers = 8MB         # min 800kB
 #max_prepared_transactions = 0      # zero disables the feature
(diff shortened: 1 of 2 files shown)


ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -65,7 +65,7 @@ listen_addresses = '*'      # what IP address(es) to listen on;
                     # defaults to 'localhost', '*' = all
                     # (change requires restart)
 port = 5432             # (change requires restart)
-max_connections = 100           # (change requires restart)
+max_connections = 300           # (change requires restart)
```
benrudolph: @snopoke do you think this number is actually too high? everytime postgres goes tits up we have to kill a bunch of connections. trying to do a bit of research into it, but it seems like a complicated question that depends on a variety of factors. http://serverfault.com/a/482302/246939
https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server

snopoke: I don't think this number is to high though don't have anything to back
that up with. I was going to look at this next week and see if I could find
anything in the Apache logs leading up to the would indicate cause.

Ultimately though I think we should install pgbouncer which handles
connection pooling etc.

benrudolph: ah ok cool. i downloaded [pgbadger](http://dalibo.github.io/pgbadger/) and ran it on the last 7 days of our postgres logs. i can send you the html report if you're interested. it actually has some pretty cool stuff in it. i can't point to anything specific that would cause it, but you might find something useful. pgbouncer sounds like the right way to go

snopoke: That would be great thanks

benrudolph: wasn't super sure on how to do a secure transfer, so i just `scp`d it onto hqdb0 under /home/brudolph. it's 80MB so it'll probably take your browser a bit to load. also feel free to delete once you've copied it. i have a local copy

benrudolph: (called `out.html`)

PR #94, opened 2015-05-07
leave shared_buffers at default value

snopoke:
@benrudolph 

Was getting this error after the conf changes:

```
IpcMemoryCreate: shmget(key=5432001, size=415776768, 03600) failed: Invalid argument 

This error usually means that PostgreSQL's request for a shared memory 
segment exceeded your kernel's SHMMAX parameter. You can either 
reduce the request size or reconfigure the kernel with larger SHMMAX. 
To reduce the request size (currently 415776768 bytes), reduce 
PostgreSQL's shared_buffers parameter (currently 50000) and/or 
its max_connections parameter (currently 12).
```

@sravfeyn buddy ping

### DIFF
--- a/ansible/roles/postgresql/templates/postgresql.conf.j2
+++ b/ansible/roles/postgresql/templates/postgresql.conf.j2
@@ -110,7 +110,7 @@ ssl = true              # (change requires restart)
 
 # - Memory -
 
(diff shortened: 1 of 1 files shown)


benrudolph:
ah nice

PR #96, opened 2015-05-08
Transfer folder

snopoke:
@benrudolph

This adds another shared folder and also adds the nfs drive to the proxy machines. This is in prep for using django-transfer.

buddy ping @sravfeyn

### DIFF
--- a/ansible/roles/shared_dir.yml
+++ b/ansible/roles/shared_dir.yml
@@ -10,7 +10,7 @@
     - include: shared_dir/tasks/setup_host.yml
 
 - name: SharedDirClient
-  hosts: webworkers
+  hosts: webworkers:proxy
   sudo: yes
   vars_files:
     - shared_dir/vars/main.yml
--- a/ansible/roles/shared_dir/tasks/setup_host.yml
+++ b/ansible/roles/shared_dir/tasks/setup_host.yml
@@ -18,7 +18,9 @@
     line: "{{ shared_data_dir }} {{ item }}(rw,sync,all_squash,no_subtree_check,anongid={{ shared_dir_gid }})"
     regexp: "^{{ shared_data_dir }} {{ item }}"
     state: present
-  with_items: groups.webworkers
+  with_flattened:
+    - groups.webworkers
+    - groups.proxy
   tags:
     - nfs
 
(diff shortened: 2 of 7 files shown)


ansible/roles/webworker/tasks/main.yml
```diff
@@ -1,23 +1,10 @@
 # Tasks to be run on machines running django
 
-- name: create restore payloads directory
```
snopoke: I moved these to the 'host' machine since it didn't seem to make sense to run them on all the webworkers when they are all looking at the same shared folders.

ansible/roles/shared_dir/tasks/setup_host.yml
```diff
@@ -27,3 +29,32 @@
...
+    - "{{ transer_payload_dir }}"
+  tags:
+    - shared_files
+
+- name: Create purging cron jobs
+  sudo: yes
+  cron:
+    name: "{{ item.name }}"
+    special_time: daily
+    job: "find {{ item.dir }} -type f -maxdepth 1 -mtime +4 -delete"
```
snopoke: Removed the `-name "*.xml"` matching and replaced it with `-type f` (only files)

ansible/roles/shared_dir.yml
```diff
@@ -10,7 +10,7 @@
     - include: shared_dir/tasks/setup_host.yml
 
 - name: SharedDirClient
-  hosts: webworkers
+  hosts: webworkers:proxy
```
benrudolph: should this be:
```
- webworkers
- proxy
```
i thought that syntax was for specifying parent/child stuff

ansible/roles/shared_dir/tasks/setup_host.yml
benrudolph: nice

ansible/roles/webworker/tasks/main.yml
benrudolph: makes sense

benrudolph:
this is cool. good to see the shared dir being used. what exactly are we storing in the transfer payloads?

ansible/roles/shared_dir.yml
snopoke: no this syntax is correct: http://docs.ansible.com/intro_patterns.html

snopoke:
@dannyroberts @benrudolph We can use this to allow apache / nginx to handle file downloads instead of Django: https://github.com/dimagi/commcare-hq/pull/6615

ansible/roles/shared_dir.yml
benrudolph: cool, that's pretty neat

PR #97, opened 2015-05-08
fix spelling and use shared var name

snopoke:
@sravfeyn


snopoke:
@benrudolph

benrudolph:
haha shoot, nice fix

PR #98, opened 2015-05-08
add xsendfile apache mod

snopoke:
@benrudolph @TylerSheffels


snopoke:
@TylerSheffels @benrudolph  This should be good to go.

PR #100, opened 2015-05-08
logrotate.conf

benrudolph:
Rotate dem logs!
:evergreen_tree: :evergreen_tree:


benrudolph:
@TylerSheffels @dannyroberts 
elf: @esoergel

ansible/roles/commcarehq/templates/logrotate.conf.j2
```diff
@@ -0,0 +1,29 @@
...
+    create 0660 root utmp
+    rotate 1
+}
+
+{{ log_home }}/*.log {
+    weekly
+    missingok
+    compress
+    delaycompress # compress on second rotation
+    olddir {{ rotated_log_home }}
```
dannyroberts: In other places, we just leave these in the same directory (though maybe you're saying that's not the best plan). One nice thing about that is you're in there grep'ing, realize that what you're looking for is more than a month ago, and then you can just sandwich your previous command with `z`...`.1` and run it again (zgrep for grep'ing gzipped files). That's such a weak argument in absolute terms that if you have any sort of reason to do this it probably trumps that.

benrudolph: ah was just about to mention the that they were gzipped. `zgrep` sounds cool. i did do it on purpose because i've seen other places that are littered with log rotates that it's hard to find what real files are there

dannyroberts: OK sounds good. Before applying this to anywhere that currently has log
rotation set up though, you'll need to (1) disable whatever logrotation is
currently set up and (2) move all the current rotated logs into the new
directory.

You might also want to look into what configuration we use on hqproxy0.

Cheers,
Danny

benrudolph: i think it actually might be best to use the log rotate python class rather than this. a lot of the logs are being rotated that way currently and it would be a mess to try and change that

PR #105, opened 2015-05-12
add deploy_buildservers.yml for stripped-down env

dannyroberts:
for running unit tests on a remote server

- split out couchdb tasks
- split out postgresql tasks

This is from my PID yesterday, thought I'd at least get it reviewed now


czue:
guessing this is for @TylerSheffels or @benrudolph

benrudolph:
pretty sweet, makes sense to me. did you want to play around with it more before merging to master or is it g2g?

dannyroberts:
I tested that it works for setting up a testing environment, but I didn't check that it works for the normal thing of deploying a new server; all I did was split out the tasks into separate files, but it's always possible I messed up some of that logic. Not sure what the best way to test this is; maybe I should just run that in my own time later.

PR #106, opened 2015-05-12
Update logging

benrudolph:
@dannyroberts @czue 
elf: @esoergel 
Still working on doing a simple version for localsettings.example.py, but thought that it'd be good to get logging working properly asap

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -31,7 +31,118 @@ CTABLE_BACKENDS = {'SQL': 'ctable.backends.SqlBackend'}
 CTABLE_PREFIX = '{{ localsettings.CTABLE_PREFIX }}'
 {% endif %}
 
-DJANGO_LOG_FILE = os.path.join('..','log',"{{ localsettings.DEPLOY_MACHINE_NAME }}-django.log")
+DJANGO_LOG_FILE = os.path.join('..','log',"{{ localsettings.DEPLOY_MACHINE_NAME }}-commcarehq.django.log")
+ACCOUNTING_LOG_FILE = os.path.join('..','log',"{{ localsettings.DEPLOY_MACHINE_NAME }}-commcarehq.accounting.log")
+
+LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': True,
+    'formatters': {
+        'verbose': {
+            'format': '%(levelname)s %(asctime)s %(module)s %(process)d %(thread)d %(message)s'
+        },
+        'simple': {
+            'format': '%(asctime)s %(levelname)s %(message)s'
+        },
+        'pillowtop': {
+            'format': '%(asctime)s %(levelname)s %(module)s %(message)s'
+        },
+    },
+    'handlers': {
+        'pillowtop': {
+            'level': 'INFO',
+            'class': 'logging.StreamHandler',
+            'formatter': 'pillowtop'
+        },
+        'console': {
+            'level': 'INFO',
+            'class': 'logging.StreamHandler',
+            'formatter': 'simple'
+        },
+        'file': {
+            'level': 'INFO',
+            'class': 'logging.handlers.RotatingFileHandler',
+            'formatter': 'verbose',
+            'filename': DJANGO_LOG_FILE,
+            'maxBytes': 10 * 1024 * 1024,  # 10 MB
+            'backupCount': 20  # Backup 200 MB of logs
+        },
+        'accountinglog': {
+            'level': 'INFO',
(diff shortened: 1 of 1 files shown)


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -31,7 +31,118 @@ CTABLE_BACKENDS = {'SQL': 'ctable.backends.SqlBackend'}
 CTABLE_PREFIX = '{{ localsettings.CTABLE_PREFIX }}'
 {% endif %}
 
-DJANGO_LOG_FILE = os.path.join('..','log',"{{ localsettings.DEPLOY_MACHINE_NAME }}-django.log")
+DJANGO_LOG_FILE = os.path.join('..','log',"{{ localsettings.DEPLOY_MACHINE_NAME }}-commcarehq.django.log")
```
benrudolph: changed the name so it resembles what people are used to seeing from the settings.py

ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -31,7 +31,118 @@ CTABLE_BACKENDS = {'SQL': 'ctable.backends.SqlBackend'}
...
+        'verbose': {
+            'format': '%(levelname)s %(asctime)s %(module)s %(process)d %(thread)d %(message)s'
+        },
+        'simple': {
+            'format': '%(asctime)s %(levelname)s %(message)s'
+        },
+        'pillowtop': {
+            'format': '%(asctime)s %(levelname)s %(module)s %(message)s'
+        },
+    },
```
benrudolph: removed the `require_debug_false` thing

ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -31,7 +31,118 @@ CTABLE_BACKENDS = {'SQL': 'ctable.backends.SqlBackend'}
...
+            'formatter': 'pillowtop'
+        },
+        'console': {
+            'level': 'INFO',
+            'class': 'logging.StreamHandler',
+            'formatter': 'simple'
+        },
+        'file': {
+            'level': 'INFO',
+            'class': 'logging.handlers.RotatingFileHandler',
```
benrudolph: Changed `file` to use a [RotatingFileHandler](https://docs.python.org/2/library/logging.handlers.html#rotatingfilehandler). I set `maxBytes` to 10MB thinking that that would be small enough to grep through quickly. don't have too good of a sense on this though so you have better suggestions let me know. Also, this will remove logs after 200MB. We can disable this feature if we want to keep them forever. A lot of our systems like celery have been deleting them after 10 backups

ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -31,7 +31,118 @@ CTABLE_BACKENDS = {'SQL': 'ctable.backends.SqlBackend'}
...
+            'formatter': 'verbose',
+            'filename': DJANGO_LOG_FILE,
+            'maxBytes': 10 * 1024 * 1024,  # 10 MB
+            'backupCount': 20  # Backup 200 MB of logs
+        },
+        'accountinglog': {
```
benrudolph: same as above

ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -31,7 +31,118 @@ CTABLE_BACKENDS = {'SQL': 'ctable.backends.SqlBackend'}
...
+            'level': 'INFO',
+            'propagate': False,
+        },
+        'accounting': {
+            'handlers': ['accountinglog', 'sentry', 'console', 'couchlog', 'mail_admins'],
+            'level': 'INFO',
+            'propagate': False,
+        },
+    }
+}
```
benrudolph: everything else should be the same as what we already have

ansible/roles/commcarehq/templates/localsettings.py.j2
czue: seems fine for now - imagine we'll figure out deterministically if we want to change this in the future.

esoergel: @benrudolph do you know about how far back 200MB goes?  A week, a month, a year?

benrudolph: no, but our current log is at ~1.5 GB. apparently we can't get the file creation time on our servers, but i think it's safe to say it's been there for a while.  (this was the command i tried to run: `stat -c '%w' /home/cchq/www/production/code_root/commcarehq.django.log`). 

fyi, there are two ways python lets you do this. one is by size and one is by time. i personally like by time, but a bunch of our other logs use the by size metric.

PR #112, opened 2015-05-15
fix india http redirect issue

dannyroberts:
Ask me not why this works on www, for I do not know. This fixes it on india.

This keeps apache from rewriting the Location header (that django sends back on a redirect) to use http


dannyroberts:
@TylerSheffels

czue:
weird. are they running the same version of apache?

TylerSheffels:
@czue Nope, both are running version 2.2.22

PR #114, opened 2015-05-19
enable autovacuum

snopoke:
@benrudolph 

I though we had autovacuum setup but its definitely not running now. Perhaps a change when we updated the config files.


benrudolph:
That is weird, I thought it's turned on by default

snopoke:
Ha, turns out it is. I was looking at the wrong column.

PR #117, opened 2015-05-20
pgbouncer

snopoke:
@benrudolph @TylerSheffels

### DIFF
--- a/ansible/roles/postgresql/tasks/main.yml
+++ b/ansible/roles/postgresql/tasks/main.yml
@@ -1,6 +1,23 @@
 ---
 # roles/postgresql/tasks/main.yml
 
+- name: Add PosgreSQL apt repo
+  apt_repository: repo='deb http://apt.postgresql.org/pub/repos/apt/ {{ ansible_lsb.codename}}-pgdg main' state=present
+  register: add_postgresql_repo
+  tags:
+    - pgbouncer
+
+- name: Add PosgreSQL apt key
+  apt_key: url=https://www.postgresql.org/media/keys/ACCC4CF8.asc state=present
+  tags:
+    - pgbouncer
+
+- name: Update package list
+  apt: update_cache=yes
+  when: add_postgresql_repo|changed
+  tags:
+    - pgbouncer
+
 - name: Install PostgreSQL & dependencies
   apt: name={{ item }} state=present
   with_items:
@@ -70,3 +87,40 @@
     state: present
   when:
     pg_install == True
+
+- name: Install pgbouncer
+  apt: name=pgbouncer state=present
+  when:
+    pg_install == True
+  tags:
+    - pgbouncer
+
+- name: pgbouncer configuration
+  template: src=pgbouncer.ini.j2 dest=/etc/pgbouncer/pgbouncer.ini
+  register: pgbouncer_configuration_updated
+  tags:
+    - pgbouncer
+
+- name: pgbouncer users
+  template: src=pgbouncer.users.j2 dest=/etc/pgbouncer/userlist.txt
+  register: pgbouncer_configuration_updated
+  tags:
+    - pgbouncer
+
+- name: pgbouncer defaults
+  template: src=pgbouncer.defaults.j2 dest=/etc/default/pgbouncer
+  register: pgbouncer_configuration_updated
+  tags:
+    - pgbouncer
+
+- name: Apply pgbouncer configuration changes
+  service: name=pgbouncer state=reloaded
+  when: pgbouncer_configuration_updated.changed
+  tags:
+    - pgbouncer
+
+- name: start pgbouncer
+  service: name=pgbouncer state=started
+  tags:
+    - pgbouncer
+
--- a/ansible/roles/postgresql/tasks/main.yml
+++ b/ansible/roles/postgresql/tasks/main.yml
@@ -5,17 +5,20 @@
   apt_repository: repo='deb http://apt.postgresql.org/pub/repos/apt/ {{ ansible_lsb.codename}}-pgdg main' state=present
   register: add_postgresql_repo
   tags:
+    - postgresql
     - pgbouncer
 
 - name: Add PosgreSQL apt key
   apt_key: url=https://www.postgresql.org/media/keys/ACCC4CF8.asc state=present
   tags:
+    - postgresql
     - pgbouncer
 
 - name: Update package list
   apt: update_cache=yes
   when: add_postgresql_repo|changed
   tags:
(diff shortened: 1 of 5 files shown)


ansible/roles/postgresql/tasks/main.yml
```diff
@@ -1,6 +1,26 @@
 ---
 # roles/postgresql/tasks/main.yml
 
+- name: Add PosgreSQL apt repo
```
snopoke: Not strictly necessary but allows us to get the latest version

ansible/roles/postgresql/templates/pgbouncer.defaults.j2
```diff
@@ -0,0 +1,11 @@
+
+# Set to 1 after you've configured pgbouncer in
+# /etc/pgbouncer
+
+START=1
```
snopoke: By default when you install pgbouncer this is set to 0 and so it won't start up

ansible/roles/postgresql/tasks/main.yml
benrudolph: cool does this mean it'll auto upgrade the postgres version if we run this? not sure if it makes sense to lock to certain version or not

ansible/roles/postgresql/templates/pgbouncer.defaults.j2
benrudolph: lol killer default

ansible/roles/postgresql/templates/pgbouncer.ini.j2
```diff
@@ -0,0 +1,207 @@
...
+; If server was used more recently that this many seconds ago,
+; skip the check query.  Value 0 may or may not run in immediately.
+;server_check_delay = 30
+
+;;;
+;;; Connection limits
+;;;
+
+; total number of clients that can connect
+max_client_conn = 200
```
benrudolph: do we need to keep this in sync with `max_connections` in postgres conf?

ansible/roles/postgresql/templates/pgbouncer.ini.j2
```diff
@@ -0,0 +1,207 @@
...
+
+; If off, then server connections are reused in LIFO manner
+;server_round_robin = 0
+
+;;;
+;;; Timeouts
+;;;
+
+;; Close server connection if its been connected longer.
+;server_lifetime = 1200
```
benrudolph: if this is in seconds, then a good portion of our queries won't be able to finish

benrudolph: wait might be confused on what this parameter does

benrudolph: well if this closes a connection during a query, we may need to bump it to 2 hours to be safe?

ansible/roles/postgresql/templates/pgbouncer.ini.j2
```diff
@@ -0,0 +1,207 @@
...
+;server_connect_timeout = 15
+
+;; If server login failed (server_connect_timeout or auth failure)
+;; then wait this many second.
+;server_login_retry = 15
+
+;; Dangerous.  Server connection is closed if query does not return
+;; in this time.  Should be used to survive network problems,
+;; _not_ as statement_timeout. (default: 0)
+;query_timeout = 0
```
benrudolph: just saw this, ignore above comments

ansible/roles/postgresql/templates/pgbouncer.users.j2
```diff
@@ -0,0 +1 @@
+"{{ localsettings.PG_DATABASE_USER }}" "{{ localsettings.PG_DATABASE_PASSWORD }}"
```
benrudolph: does this mean it will only clean the hqweb2postgres connnection? may want to add the devreadonly and/or sentry though those i'm sure are way less important

benrudolph:
:tada: hope this fixes those postgres issues!

ansible/roles/postgresql/templates/pgbouncer.ini.j2
```diff
@@ -0,0 +1,207 @@
+;; database name = connect string
+;;
+;; connect string params:
+;;   dbname= host= port= user= password=
+;;   client_encoding= datestyle= timezone=
+;;   pool_size= connect_query=
+[databases]
+{% for db in postgresql_dbs | unique %}
+{{ db }} = host=localhost port=5432
```
TylerSheffels: Does this require a change to the postgres conf?

TylerSheffels: By that I mean, does this port overlap with postgres and do we need to change what port postgres listens on?

ansible/roles/postgresql/tasks/main.yml
snopoke: no, it will only upgrade if the ansible state is `latest` but we always use `present` which won't upgrade once its installed

ansible/roles/postgresql/templates/pgbouncer.defaults.j2
snopoke: yea, took a while to figure that one out

ansible/roles/postgresql/templates/pgbouncer.ini.j2
snopoke: no, this is telling pgbouncer how to connect to the database. pgbouncer runs on port 6432 so we'll need to update our localsettings before we actually start using it

ansible/roles/postgresql/templates/pgbouncer.ini.j2
snopoke: Good question. I think this is totally independent of the postgres limit but agree that for now it should probably be at least as high.

ansible/roles/postgresql/templates/pgbouncer.users.j2
snopoke: Yes, we should run all our connections through here but we can add those later

ansible/roles/postgresql/templates/pgbouncer.ini.j2
snopoke: That was a good question though. I don't know if it will kill the connection if there is a query still running but I would guess not.

snopoke:
@benrudolph @TylerSheffels updated

benrudolph:
nothing blocking, merging

benrudolph:
(though be careful with the whole tag thing)

PR #124, opened 2015-05-21
adjust pgbouncer settings

snopoke:
@benrudolph

### DIFF
--- a/ansible/vars/dev.yml
+++ b/ansible/vars/dev.yml
@@ -47,6 +47,12 @@ postgres_users:
     username: "{{ localsettings.PG_DATABASE_USER }}"
     password: "{{ localsettings.PG_DATABASE_PASSWORD }}"
 
+postgres_config:
+  max_connections: 20
+  pgb_default_pool: 15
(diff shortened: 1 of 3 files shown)


snopoke:
I think the total connections (default_pool_size + reserve_pool_size) should always be below the max_connections setting for postgresql.

ansible/vars/dev.yml
```diff
@@ -47,6 +47,12 @@ postgres_users:
     username: "{{ localsettings.PG_DATABASE_USER }}"
     password: "{{ localsettings.PG_DATABASE_PASSWORD }}"
 
+postgres_config:
```
TylerSheffels: Isn't this a huge decrease in max_connections? Did you use a different number on prod?

benrudolph: this is just for dev.yml. presumably the real ones will be higher

benrudolph:
dont' think anything is blocking here, so i'll go ahead and merge since @TylerSheffels is independent of this PR

PR #134, opened 2015-05-25
updates to get touchforms working on staging

snopoke:
see https://github.com/dimagi/touchforms/commit/f5957b989545b8d1e3b72fc62999222bdad24a37

@benrudolph


benrudolph:
wow, nice. i was hurting friday afternoon when i was seeing those errors and changing the port didn't Just Work

PR #137, opened 2015-05-27
add newrelic plugin agent

snopoke:
@TylerSheffels @benrudolph

### DIFF
--- a/ansible/roles/nginx/templates/default.j2
+++ b/ansible/roles/nginx/templates/default.j2
@@ -1,1 +1,7 @@
 # {{ ansible_managed }}
+
+location /nginx_stub_status {
+    stub_status on;
+    allow 127.0.0.1;
+    deny all;
+}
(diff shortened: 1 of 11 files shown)


ansible/roles/nginx/templates/default.j2
```diff
@@ -1 +1,7 @@
 # {{ ansible_managed }}
+
+location /nginx_stub_status {
```
TylerSheffels: :+1:

snopoke: Note I haven't actually enabled the nginx monitoring (not sure of the best pattern to use to get apache on one proxy machine and nginx on the other). Any ideas would be helpful.

TylerSheffels:
@snopoke Don't know enough about the plugin agent, but this looks awesome to me.  Maybe test conf on staging?

benrudolph:
:clap: :clap:

czue:
just got a demo of this. :clap:

PR #138, opened 2015-05-28
Don't default to DEFAULT_CACHE to memcached

TylerSheffels:
This occurs when localsettings.DEFAULT_CACHE is unset, as it was on the india server.


TylerSheffels:
@benrudolph 
@orangejenny

PR #139, opened 2015-05-29
Setup for pgbench host and runner server

benrudolph:
@czue @snopoke, given the server has postgres setup, this should make it pretty easy to setup `pgbench`. i'm in the process of running it now. the actual command is `pgbench9.1-*test`. While this takes a bit away from configurability, it will help with consistency and ensuring we can repeatably run the same tests on multiple machines. haven't tried setting up `9.4` yet but should be pretty straight forward. would be cool to be able to kick this off via fabric. if you do end up playing around with it make sure to override the email by using `-e "receiver=someemail@abc.com"` so that the results will be sent to you

also the 1 commit definitely doesn't do justice to the actual number of commits, but didn't want to pull a @TylerSheffels :wink: 
screencapture-github-com-dimagi-commcarehq-ansible-pull-66-1428442004708


czue:
:bus: :bus: :bus: :bus: 
@TylerSheffels

snopoke:
Nice. Is this gtg?

Are we happy with the default pgbench script?
```
BEGIN;
UPDATE pgbench_accounts SET abalance = abalance + :delta WHERE aid = :aid;
SELECT abalance FROM pgbench_accounts WHERE aid = :aid;
UPDATE pgbench_tellers SET tbalance = tbalance + :delta WHERE tid = :tid;
UPDATE pgbench_branches SET bbalance = bbalance + :delta WHERE bid = :bid;
INSERT INTO pgbench_history (tid, bid, aid, delta, mtime) VALUES (:tid, :bid, :aid, :delta, CURRENT_TIMESTAMP);
END;
```

snopoke:
A little bit of reading shows that this benchmark has been marked as obsolete by the TCP: http://www.tpc.org/tpcb/default.asp

However we can still use it as a starting point but we should definitely think about trying a more realistic test (perhaps using [BenchmarkSQL](http://sourceforge.net/projects/benchmarksql/) or [HammerDB](http://hammerora.sourceforge.net/index.html))

snopoke:
Other options:
* [oltpbenchmark](http://oltpbenchmark.com/wiki/index.php)
* (YCSB](https://github.com/brianfrankcooper/YCSB)

benrudolph:
wait, seriously? postgresql's default pgbench metric is obsolete and they continue to have it in their contrib package? damn i'll take a look into some of those other benchmarking tools

benrudolph:
Also for now think I'll leave this open til I iron out some kinks

ansible/roles/load_test/templates/pgbench-test.j2
```diff
@@ -0,0 +1,39 @@
...
+
+        # Initialize database
+        scale=$((scales[$i - 1]))
+        pgbench{{ postgresql_version }} $conn_options -i -s $scale {{ db_name }}
+
+        # Run actual test
+        {% if item == "read" %}
+        out+=$(pgbench{{ postgresql_version }} $conn_options -c {{ connections }} -j {{ threads }} -T {{ time_in_sec }} -S {{ db_name }})
+        {% elif item == "writeread" %}
+        out+=$(pgbench{{ postgresql_version }} $conn_options -c {{ connections }} -j {{ threads }} -T {{ time_in_sec }} {{ db_name }})
```
benrudolph: @snopoke when this is deployed, the wiki [tests](https://wiki.postgresql.org/wiki/Pgbenchtesting) are created under the `writeread` script. (should already be deployed but the receiver email is wrong and you may want to change the T for the first few trails). The reason I went with this approach is so we see a graph of all the different scales (1, 10, 100, 1000, 3000, 5000, 10000) rather than using the formula provided in pgwiki. i did this cause i wasn't sure the wiki formula would be correct given that it might depend on more parameters than the wiki suggests. don't know if that is a real concern though

snopoke: Great, thanks. Based on the graph you sent out previously I think the wiki formula will put us somewhere in the right zone. I did the calc yesterday and came up with a scale of ~4500 for 'mostly cache'

snopoke:
Added setup code for https://github.com/dimagi/pgbench-tools which should make running the tests and comparing results quite a bit easier

benrudolph:
Great, pgbench-tools looks like the way to move forward. saving the results in a database is SO much better than emailing haha. will delete my old scripts on Monday

ansible/roles/load_test/README.md
```diff
@@ -0,0 +1,45 @@
...
+  
+  
+Get the results
+---------------
+
+  $ tar -czvf results.tar.gz results/*
+  
+And then locally:
+  
+  $ scp {load_test_runner}:/home/cchq/pgbench-tools/results.tar.gz
```
benrudolph: :+1:

PR #142, opened 2015-06-10
Updates to Nginx Config

TylerSheffels:
This:
* Resolve the change in behavior from 1.6.2 -> 1.8 with nginx alias
* Switch to use nodejs instead of npm in the package manager (which resolves an issue with apt-get versions)
* Move SSL cert handling into Ansible
* Handle another case of `service restart nginx` not returning a nonzero return code
* Redirect commtrack traffic to commcarehq

@dannyroberts @czue


ansible/deploy_proxy.yml
```diff
@@ -11,8 +11,8 @@
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.cchq_redirect == True, action: site, site_name: cchq_redirect }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.cchq_static == True, action: site, site_name: cchq_static }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.cchq_static_redirect == True, action: site, site_name: cchq_static_redirect }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack == True, action: site, site_name: commtrack }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_static == True, action: site, site_name: commtrack_static }
+    - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_ssl == True, action: site, site_name: commtrack_ssl }
+    - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_http == True, action: site, site_name: commtrack_http }
```
czue: should we make this more consistent with the cchq naming?

ansible/roles/nginx/tasks/install.yml
```diff
@@ -38,3 +38,51 @@
 - name: Create the link for site enabled specific configurations
   sudo: yes
   file: path=/etc/nginx/sites-enabled/default state=link src=/etc/nginx/sites-available/default
+
+- name: Create directories for certificates
+  file:
+    path: "{{ item }}"
+    owner: "root"
+    group: "root"
```
czue: are these supposed to be run by root?

ansible/roles/nginx/vars/commtrack_ssl.yml
```diff
@@ -0,0 +1,15 @@
...
+- server:
+   file_name: zz_commtrack #There was a bug where a phone would default to connecting to commtrack on login and use this cert, then failing.  Renaming the file makes it load later and work.
+   listen: "443 ssl"
+   server_name: "{{ COMMTRACK_SITE_HOST }}"
+   client_max_body_size: 100m
+   use_balancer: True
+   proxy_set_header: "Host $http_host"
+   location1:
+    name: /
+    return: "301 https://{{ SITE_HOST }}$request_uri"
```
czue: could add a redirect for commtrack.org/home to the new supply anchor

czue:
:+1: after root question is answered

ansible/roles/nginx/tasks/install.yml
TylerSheffels: The certs are current owned by root, and apache is running as www-data (groups: www-data cchq). I think fine to leave these as root.

TylerSheffels:
@czue This should be ready to go now.

PR #143, opened 2015-06-10
Load test tsung

benrudolph:
@snopoke this is a very early draft of Tsung setup targeting postgres. I have to do a bunch more reading on exactly what all the levers and buttons in the tsung configuration file do, but from what i've read so far it seems really powerful. One interesting thing we can do to simulate real load, is to take the logs from our postgres database, run them through pgbadger to generate a `.tsung` file which creates sessions for us to use. just throwing this up so you know where i'm at with it. (note this is a PR into `load-test` branch)

elf: @dannyroberts


snopoke:
Nice, thanks for this. One issue with using the logs from prod is we'd have to use the same DB schema. Useful to know for future though or if we want to profile our prod DB.

snopoke:
I wouldn't worry too much about using ansible to set up the session files since we'll probably end up tweaking those quite a but and won't want to have to keep updating them in ansible repo. Actually it might make sense to create a separate repo for them.

benrudolph:
Yea, the session generation from logs would be a late game thing.

i'm torn on how to keep the session files. i'd really like a better way than to keep updating ansible, but i'd also like to keep it under version control. i'd be game for putting them in a different repo, but wouldn't that be pretty much the same workflow as committing them to ansible? or are you imagining something where we log on to the runner machine and just git pull the repo

snopoke:
Yea, just pulling directly from the repo on the test runner. We can use the
'mobile-endpoint' repo I setup.

On 11 June 2015 at 16:02, Ben <notifications@github.com> wrote:

> Yea, the session generation from logs would be a late game thing.
>
> i'm torn on how to keep the session files. i'd really like a better way
> than to keep updating ansible, but i'd also like to keep it under version
> control. i'd be game for putting them in a different repo, but wouldn't
> that be pretty much the same workflow as committing them to ansible? or are
> you imagining something where we log on to the runner machine and just git
> pull the repo
>
> —
> <https://github.com/dimagi/commcarehq-ansible/pull/143#issuecomment-111146211>
> .
>



-- 
Simon Kelly

benrudolph:
sounds good

PR #144, opened 2015-06-15
Add default_server for nginx processing and correct http spelling

TylerSheffels:
@dannyroberts 

I still want to figure out how this will work if two servers are specified as default  (e.g. staging and prod in this configuration).

### DIFF
--- a/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -2,7 +2,7 @@
 nginx_sites:
 - server:
    file_name: "{{ deploy_env }}_commcare"
-   listen: "443 ssl default_server"
+   listen: "443 ssl {{ nginx_set_default_server }}"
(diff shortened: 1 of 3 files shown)


TylerSheffels:
@czue This has the change we talked about this morning and a couple spelling fixes.

ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -2,7 +2,7 @@
 nginx_sites:
 - server:
    file_name: "{{ deploy_env }}_commcare"
-   listen: "443 ssl"
+   listen: "443 ssl {{ nginx_set_default_server }}"
```
czue: nitpicky but this variable name seems odd to me. (not sure where the `set` part comes from)

czue:
nothing blocking. the other bugs you found were all in untracked files?

TylerSheffels:
@czue Yeah, there were a couple changes in the untracked configuration, but this was the blocking bug.

PR #147, opened 2015-06-18
Nginx Bug Fixes

TylerSheffels:
This addresses:

* http://manage.dimagi.com/default.asp?171347 by raising the timeout and only ever trying one of the servers in the load balancing group
* http://manage.dimagi.com/default.asp?171406#960292 by adding a wiki and help endpoint
* The fact that it was hard to find the transfer server variable by making it dependent on the proxy_type (which it implicitly is)

@dannyroberts @czue @snopoke

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -1,4 +1,4 @@
-#
+    #
 # {{ ansible_managed }}
 #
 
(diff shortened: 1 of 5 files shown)


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -334,7 +334,7 @@ SHARED_DRIVE_ROOT = '{{ shared_mount_dir }}'
 RESTORE_PAYLOAD_DIR_NAME = '{{ restore_payload_dir_name }}'
 TRANSFER_FILE_DIR_NAME = '{{ transfer_payload_dir_name }}'
 SHARED_TEMP_DIR_NAME = '{{ shared_temp_dir_name }}'
-TRANSFER_SERVER = '{{ localsettings.TRANSFER_SERVER }}'
+TRANSFER_SERVER = '{{ (proxy_type == 'nginx') | ternary('nginx', 'apache') }}'
```
czue: what's this change doing?

TylerSheffels: Previously there were two settings that needed to be changed 'proxy_type' and 'localsettings.TRANSFER_SERVER'. they both need to change when the proxy changes, otherwide async downloads wont work, but TRANSFER_SERVER expects 'apache' and proxy_type expects 'apache2'

PR #148, opened 2015-06-19
Fix 302 redirects

TylerSheffels:
Somehow help.commcarehq.org was resulting in constant 302 redirects with the previous configuration.  I think it has to do with the response from confluence but haven't fully looked into it.

### DIFF
--- a/ansible/roles/nginx/vars/wiki.yml
+++ b/ansible/roles/nginx/vars/wiki.yml
@@ -4,10 +4,9 @@ nginx_sites:
    file_name: wiki
    listen: "443 ssl"
    server_name: "wiki.commcarehq.org help.commcarehq.org"
-   proxy_set_header: "Host $http_host"
    location1:
     name: /
-    proxy_pass: http://confluence.internal.dimagi.com
+    proxy_pass: http://confluence.dimagi.com
    location2:
     name: "= /"
     return: "302 https://$server_name/display/commcarepublic/Home"


czue:
is this the permanent solution? was apache pointing at the internal dns or the external one?

TylerSheffels:
This is the fix I did last night. Apache was pointing at the internal DNS. Mostly I wanted to revisit this with fresh eyes, and know that this is a stable configuration.

czue:
(no strong feelings from me on this. if it works and you and @dannyroberts think it's a fine change then I'm signed off)

ansible/roles/nginx/vars/wiki.yml
```diff
@@ -4,10 +4,9 @@ nginx_sites:
...
    listen: "443 ssl"
    server_name: "wiki.commcarehq.org help.commcarehq.org"
-   proxy_set_header: "Host $http_host"
    location1:
     name: /
-    proxy_pass: http://confluence.internal.dimagi.com
+    proxy_pass: http://confluence.dimagi.com
    location2:
     name: "= /"
     return: "302 https://$server_name/display/commcarepublic/Home"
```
dannyroberts: Can you explain the difference between location1 and location2 to me? (And "/" vs "= /"?)

dannyroberts:
I'm going to merge this, but still want to understand

PR #151, opened 2015-06-23
Variable for uses postgres

benrudolph:
@TylerSheffels @wspride 
easier to set default var
elf: @czue


czue:
offlined deploy strategy with @benrudolph and he'll follow up

benrudolph:
:+1:

PR #152, opened 2015-06-25
Add timing log format and timing to cchq

TylerSheffels:
@dannyroberts 

@snopoke Buddy!

### DIFF
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -17,6 +17,8 @@ http {
         access_log {{ log_home }}/{{ nginx_access_log_name }};
         error_log {{ log_home }}/{{ nginx_error_log_name }} warn;
 
+        log_format timing '$request $status $request_time';
+
 {% for k,v in nginx_http_params.iteritems() %}
         {{ k }}  {{ v }};
 {% endfor %}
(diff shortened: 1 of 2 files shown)


czue:
neat - does this tell us timing of every request?

dannyroberts:
Yup, and tyler and I mocked up a log parser that users commcare-hq/django's
url resolution to replace urls with the corresponding django view path,
which tyler's taken off with to parse the new timing log:
https://github.com/dimagi/commcare-hq/compare/nginx-logparsing. The
intended product of passing the timing log through this is to answer not
only "which single requests have the longest run time", but also which
views do our workers spend most of their total time processing (i.e. sum of
time of all requests to that view; think long running views that are
frequently called), and for a given view, what's the distribution of run
times (not just the single longest run time).

In particular, I think sorting by total time spent by prod web workers
processing all requests to a view is a good place to start for finding
low-hanging fruit for speedups—good mix of "does it take a long time to
run" and "is this view actually used".

Thoughts?

czue:
:+1: sounds like a good plan to me

snopoke:
Not to say this isn't useful but you can already get that from NewRelic:
https://rpm.newrelic.com/accounts/366142/applications/5058236/transactions

czue:
wow, that does look exactly like what @dannyroberts just described.

dannyroberts:
huh! Well there we go, we should just use that then!

TylerSheffels:
Damn, that is exactly what we were looking for.

PR #156, opened 2015-07-06
Set proxy headers for Tableau

TylerSheffels:
Without these headers Tableau Server tries to guess the server name.


TylerSheffels:
@dannyroberts @czue

dannyroberts:
Any reason this is tableau specific? Can it be part of the default somehow?

TylerSheffels:
@dannyroberts My understanding of X-* Headers is they are application specific (although very prolific and pretty well defined).  I added them to the Tableau config because they were preventing a specific bug, but they shouldn't hurt to add to more of the configurations

PR #157, opened 2015-07-08
Adding timing log back

TylerSheffels:
@dannyroberts @czue 

Even though new relic has this same data, I would like to have it in the nginx logs in order to look at things like requests per hour, or timing at a particular time of day.  @snopoke let me know if there is a better way/new relic does this already.


czue:
no objections from me, will let @snopoke or @dannyroberts merge

PR #159, opened 2015-07-13
Add read timeout for tableau

TylerSheffels:
@dannyroberts @czue 

This value is how long nginx waits for a response from the downstream server. I am on the fence for increasing it across the board, because I think its good to see what services are taking more than a minute to return.  If want to do that we would need a way to see 504 in the nginx logs, does that sound worth doing?


czue:
happy to just change it on a server-by-server basis for now. i think generally logging more requests (and especially the 504's) is a good plan - especially if that's easy. wouldn't spend a ton of time on it if we don't know how we'll use them.

going to merge this since it looks obviously harmless and let you and @dannyroberts decide whether to take any additional steps

PR #162, opened 2015-07-21
setup for prototpe service

snopoke:
@NoahCarnahan

### DIFF
--- a/ansible/deploy_load_test.yml
+++ b/ansible/deploy_load_test.yml
@@ -20,3 +20,15 @@
     - include: roles/load_test/tasks/install_host.yml
   roles:
     - {role: load_test, tags: 'load_test'}
+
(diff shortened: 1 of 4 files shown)


snopoke:
depends on https://github.com/dimagi/mobile-endpoint/pull/21/

ansible/deploy_load_test.yml
```diff
@@ -20,3 +20,15 @@
     - include: roles/load_test/tasks/install_host.yml
   roles:
     - {role: load_test, tags: 'load_test'}
+
+- name: Prototype service
+  hosts:
+    - load_test_host
```
snopoke: @NoahCarnahan wasn't sure which machine to deploy this on but picked this one since it required less setup.

PR #168, opened 2015-07-27
customize redis role to our needs

snopoke:
This removes the bennojoy.redis submodule and replaces it with a tailored version which only handles ubuntu.

The only addition is to add a task to add the `ppa:chris-lea/redis-server` apt repo so that we get the latest version of redis.

@NoahCarnahan 
cc @TylerSheffels


TylerSheffels:
@snopoke Looks great, unmergeable now though.

NoahCarnahan:
Is using an "unofficial" ppa a security concern?

snopoke:
Potentially I guess though this one in particular is used and referenced
quite a lot.

On 29 July 2015 at 17:09, Noah Carnahan <notifications@github.com> wrote:

> Is using an "unofficial" ppa a security concern?
>
> —
> <https://github.com/dimagi/commcarehq-ansible/pull/168#issuecomment-125983301>
> .
>



-- 
Simon Kelly

TylerSheffels:
Sounds wrong to say this, but, we already use a couple unofficial PPA's. Don't know if that makes it any better or us more comfortable. It's probably doable to find the md5 on the redis site and test the download as well.

PR #169, opened 2015-07-27
reduce timeout

czue:
related to http://manage.dimagi.com/default.asp?177250

@dannyroberts @benrudolph @TylerSheffels


benrudolph:
:+1: 
seems like the root cause could be elsewhere though, just not entirely sure where. this smells a lot like the issues we were seeing with the postgres connections, cept i'm pretty sure redis already uses a pooler

czue:
agree this is definitely a band aid

PR #170, opened 2015-07-27
Revert "reduce timeout"

czue:
Reverts dimagi/commcarehq-ansible#169

after further debugging we now no longer suspect this is necessary


czue:
@benrudolph

PR #173, opened 2015-07-28
Tweaks

snopoke:
Read by commit

@TylerSheffels FYI some of the redis config is not compatible with v3+ (see second commit)


ansible/roles/postgresql/defaults/main.yml
```diff
@@ -1,6 +1,7 @@
 pg_install: True
 postgresql_version: '9.1'
 postgresql_port: 5432
+postgresql_ssl_enabled: True
```
TylerSheffels: I think we want to default this to False now as we don't us postgres SSL, correct? I thought we already made this change when adding pgbouncer, as that doesn't use SSL to connect.

TylerSheffels:
@snopoke 1 Question, otherwise :+1:

ansible/roles/redis/templates/redis.conf.j2
```diff
@@ -345,80 +345,6 @@ slowlog-log-slower-than 10000
...
-# Max number of VM I/O threads running at the same time.
-# This threads are used to read/write data from/to swap file, since they
-# also encode and decode objects from disk to memory or the reverse, a bigger
-# number of threads can help with big objects even if they can't help with
-# I/O itself as the physical device may not be able to couple with many
-# reads/writes operations at the same time.
-#
-# The special value of 0 turn off threaded I/O and enables the blocking
-# Virtual Memory implementation.
-vm-max-threads 4
```
benrudolph: if the new redis module will deploy an upgrade to redis, does it make sense to also delete `maxclients` from this conf as well? non-blocking, but was looking at the conf today to see what it'd take to deploy on staging

snopoke: maxclients is still in the conf (though it can't have a '0' value)

ansible/roles/postgresql/defaults/main.yml
snopoke: prod still uses some ssl connections e.g. sentry

PR #174, opened 2015-07-31
use notify logger for notify logger in template

czue:
related to http://manage.dimagi.com/default.asp?177091

@snopoke cc @esoergel


esoergel:
:+1: Didn't realize we overrode the log settings

PR #175, opened 2015-07-31
Update redis.conf

benrudolph:
@snopoke most everything seems g2g. been using this to check settings and read about them: http://download.redis.io/redis-stable/redis.conf. in version 2.6 redis has the ability to automatically set maxclients. if we specify 10000 and the OS can't handle that many clients that redis will automatically change the maxclients setting to what it can handle. thus i think it's best to leave it at 10000 and let redis figure it out. this seems to be the number that is used in most examples too. here's more info: http://redis.io/topics/clients#maximum-number-of-clients

fyi @orangejenny, @dannyroberts

### DIFF
--- a/ansible/roles/redis/defaults/main.yml
+++ b/ansible/roles/redis/defaults/main.yml
@@ -16,7 +16,7 @@ redis_db_dir: /var/lib/redis
 redis_role: master
 redis_requirepass: false
 redis_pass: None
-redis_max_clients: 128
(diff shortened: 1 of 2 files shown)


benrudolph:
if you're ok with this, i'll go ahead and deploy this to staging

dannyroberts:
Looks good to me! Thanks for the research!

snopoke:
:+1:

PR #176, opened 2015-08-03
Pin version to 3.0

benrudolph:
@snopoke @dannyroberts this is needed in order for ansible to actually upgrade the redis version. this is in effect on staging

@orangejenny

### DIFF
--- a/ansible/roles/redis/vars/main.yml
+++ b/ansible/roles/redis/vars/main.yml
@@ -8,6 +8,4 @@ env:
 
 redis_ubuntu_pkg:
   - python-selinux
-  - redis-server
-
-  
+  - redis-server=3:3.0.*


ansible/roles/redis/vars/main.yml
```diff
@@ -8,6 +8,4 @@ env:
 
 redis_ubuntu_pkg:
   - python-selinux
-  - redis-server
-
-  
+  - redis-server=3:3.0.*
```
benrudolph: this dumb `3:` prefix caused me a big headache

dannyroberts: huh, pretty non-standard

dannyroberts:
Awesome!

PR #178, opened 2015-08-06
least_conn load balance

benrudolph:
@TylerSheffels was looking into the `OSError: [Errno 12] Cannot allocate memory` error we occasionally get. This ends up coming from us not being able to allocate enough space to run the `form_translate.jar`. I was looking at newrelic and our webworkers aren't that evenly balanced. hqdjango5 was maxed out in memory while hqdjango3 was only using 50%. I'm wondering if you think using a different load balancing schema will work better? was reading about it here: http://nginx.org/en/docs/http/load_balancing.html

buddy: @snopoke

### DIFF
--- a/ansible/roles/nginx/templates/site.j2
+++ b/ansible/roles/nginx/templates/site.j2
@@ -4,8 +4,9 @@
 
 {% if item.server.get('use_balancer', False) == True %}
 upstream {{ item.server.file_name }} {
+  least_conn;
 {% for host in groups['webworkers'] %}
-  server {{ host }}:{{ django_port }} weight=2;
+  server {{ host }}:{{ django_port }};
 {% endfor %}
 }
 {% endif %}


benrudolph:
wow, buddy: @czue

czue:
:+1: from me and seems like a good experiment. I suspect it has to do with a few requests taking huge amounts of memory rather than the number of connections being way off, but this seems to be a strict improvement. I'll let @TylerSheffels and @dannyroberts review/merge deploy

snopoke:
:+1: will be interesting to see the difference

If you setup the [balancer manger](http://httpd.apache.org/docs/2.2/mod/mod_proxy_balancer.html#balancer_manager) I think you can change this on the fly which would make it much easier to test out.

benrudolph:
sounds good. balancer manager is only for apache right?

snopoke:
Right, forgot we're on nginx now.

On 6 August 2015 at 11:51, Ben <notifications@github.com> wrote:

> sounds good. balancer manager is only for apache right?
>
> —
> <https://github.com/dimagi/commcarehq-ansible/pull/178#issuecomment-128311689>
> .
>



-- 
Simon Kelly

dannyroberts:
@TylerSheffels I'm merging this but holler if you see a problem with it

TylerSheffels:
@benrudolph This looks great.  Interested to see if we notice a difference. Reading through that page it also looks like ip-hash would benefit from some of the caching we do around user/domain, but that's completely a gut feeling and would be hard to measure.

benrudolph:
i did take a look at the ip-hash balancer, but didn't think about any
caching benefits. interesting, i have a feeling most of our caching is done
via redis vs. memory, but can't really say for sure. also could potentially
be bad to introduce dependencies on a user hitting the same server on every
request. maybe if you get time today you can show me a deploy

PR #181, opened 2015-08-13
The nginx service log should point to a non environment dependent loc…

TylerSheffels:
…ation

@dannyroberts @benrudolph


dannyroberts:
But the particular production/staging access logs still end up in e.g. /home/cchq/www/production/log/ right?

TylerSheffels:
@dannyroberts Yes, this is only the log for nginx the service.  Without this change if you deployed staging then production it would flip this log in between the respective folders.

dannyroberts:
:+1:

PR #183, opened 2015-08-17
Better find for redis module

benrudolph:
@dannyroberts django gets screwy when we import `django-redis` before defining the `CACHES` variable. this fixes that

cc: @biyeun


dannyroberts:
Fun. Great fix!

PR #188, opened 2015-08-19
Commtrack purely redirects to commcare, and the supply anchor now, so…

TylerSheffels:
… this is unused

@benrudolph


benrudolph:
sweet, so we don't have to worry about redeploying?

PR #190, opened 2015-08-24
swith to nginx

snopoke:
@TylerSheffels


ansible/roles/nginx/templates/default.j2
```diff
@@ -1,5 +1,8 @@
 # {{ ansible_managed }}
 server {
+  listen 127.0.0.1:80;
+  listen [::1]:80;
```
snopoke: This is necessary to get around the SSL redirects

PR #196, opened 2015-09-01
HSTS config

sravfeyn:
This PR adds HSTS (HSTS - a response header that tells browser that HQ is HTTPS only). This is not only a good security practice but also cuts rountrip cost of HTTP -> HTTPS redirect, as modern browsers will then automatically request HTTPS even when user enters an HTTP URL (see note on [this PR](https://github.com/dimagi/commcare-hq/pull/8076) about redirect time). But I have few questions before getting this merged in.
- Do we serve any HTTP traffic at all?
- Is 2 months expiry too high or too low?
- Should we include all subdomains as well?

### DIFF
--- a/ansible/roles/nginx/templates/site.j2
+++ b/ansible/roles/nginx/templates/site.j2
@@ -55,6 +55,9 @@ server {
 {% if nginx_ssl_ciphers|default('') %}
   ssl_ciphers {{ nginx_ssl_ciphers }};
 {% endif %}
+{% if nginx_hsts_max_age %}
+  add_header Strict-Transport-Security "max-age={{ nginx_hsts_max_age }}";
+{% endif %}
 {% endif %}
 
 {% if server_tokens|default('off') %}
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -46,7 +46,7 @@ elasticsearch_cluster_name: 'prodhqes-1.x'
 
 nofile_limit: 65536
 
-nginx_hsts_max_age: 86400 # 1 day
+nginx_hsts_max_age: 300 # 5 minutes
 
 # commcarehq.org of certs
 nginx_combined_cert_value: "{{ ssl_secrets.certs.commcarehq_org }}"
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -46,6 +46,8 @@ elasticsearch_cluster_name: 'prodhqes-1.x'
 
 nofile_limit: 65536
 
+nginx_hsts_max_age: 86400 # 1 day
+
 # commcarehq.org of certs
 nginx_combined_cert_value: "{{ ssl_secrets.certs.commcarehq_org }}"
 nginx_key_value: '{{ ssl_secrets.private_keys.commcarehq_org }}'


snopoke:
@dannyroberts I think you may be best to review this

dannyroberts:
@sravfeyn Have you tested this manually at all on staging? I wouldn't want to deploy this without testing that the underlying nginx conf change works the way we expect it to

TylerSheffels:
@sravfeyn Currently the only thing we serve over HTTP is slow.commcarehq.org.  Any reason not to update slow.commcarehq.org to use HTTPS? I don't have a strong feeling about the 2 month expiry, but it seems pretty high. Is there a standard? Even a day expiry seems like it would make a huge difference here.  The configuration you have here will get applied to all subdomains, which I think is what we want because of the first point.

sravfeyn:
Re. Expiry standard: Usually people set it from a day to as long as an year. For that duration when can not serve any HTTP traffic under enabled domain/subdomain. What governs this number is usually the question - 'Are we going to serve any content on HTTP in near future? How long in future are we sure that we won't serve HTTP content'. Facebook sets it at 2 months, I have seen some sites putting it at an year. 
Testing - I am planning to test it on staging subdomain. But I thought I have to get these changes merged in first, isn't it so?

dannyroberts:
You can log into the salt/ansible machine, and then check out your hsts
branch and deploy it to staging without having it merged first. Usually by
the time it's merged that means we're ready to deploy it to prod.

sravfeyn:
Do we maintain commcarehq-ansible/config directory in git? How do I manually change it? I want to add `uses_hsts: true` config in `staging.yml` and then add the above header only when an env has `uses_hsts` set to true

czue:
you might want to read this: https://confluence.dimagi.com/display/internal/Ansible

sravfeyn:
Ha, thanks Cory. I now remember, I added a commit once to that repo

sravfeyn:
Doesn't feel great to put settings like this in a repo without PRing. Instead, if we split that config into secrets and configs, we can maintain configs in a hosted repo and PR a change before writing it

dannyroberts:
That's a good point. I'm not totally sure what an ideal version of this
would look like. One idea is to split those things into two files like you
said, and then use Ansible Vault
<http://docs.ansible.com/ansible/playbooks_vault.html> on the secrets file,
and then push this to a private github repository. (Of course it'd be
critical that we start over with a new repo and don't save any of the
current git history, since that all contains plaintext passwords.)

I'm not planning on driving any of these changes in the short term, but
it'd definitely be an improvement, and please feel free to spearhead.

dannyroberts:

sravfeyn:
This was a good to have, we can still try it. I have updated the PR to test it on prod for a day. I can deploy this tomorrow, if no concerns on prod.

emord:
Whoa I was thinking about doing this recently. Can you rebase this instead of merging master since it's been open for so long?

I think we should set this to something low like 5 min while testing. That way we can easily revert it without causing too much chaos. After that I don't see any issues with increasing it.

Another note is we should make sure that this doesn't affect j2me endpoint since that's on http now

sravfeyn:
Haha, nice!

I put a day, so that we would could compare impact of this change on performance (due to lesser redirects). But we can do 5 minutes as well.

dannyroberts:
Anything left to do here? Looks like

- change value to 5 minutes
- rebase commits in this PR

these are easy enough that I'm tempted to do it myself, but since I'm not deep into this and could be missing something, I will let someone else do it.

emord:
I haven't looked at this closely, but I think that's all that's needed. Eventually we'd want something larger than 5 min, but I think that's a decent place to start. We can bump it up as long as no issues are reported

dannyroberts:
OK I'll make that commit then and let you approve it.

dannyroberts:
I just confirmed j2me site conifg is unchanged in two ways: by doing a dry run of proxy changes from this branch, and by inspecting the code (this header is in the `{% if 'ssl' in item.server.listen %}` section of `site.j2`, and the [`includeSubDomains` directive](https://https.cio.gov/hsts/) is not set).

sravfeyn:
Thanks Danny!

PR #200, opened 2015-09-02
install python-netaddr on the controller

czue:
@dannyroberts i was getting this error setting up postgres.

`fatal: [192.168.33.16] => {'msg': 'AnsibleFilterError: The ipaddr filter requires python-netaddr be installed on the ansible controller', 'failed': True}`

running this command on the controller fixes it, but I wasn't 100% sure this was the right way to make sure that happens.

### DIFF
--- a/provisioning/control.sh
+++ b/provisioning/control.sh
@@ -1,5 +1,6 @@
 sudo apt-get update
 sudo apt-get install -y python-software-properties
+sudo apt-get install -y python-netaddr
 sudo apt-add-repository ppa:ansible/ansible
 sudo apt-get update
 sudo apt-get install -y ansible


TylerSheffels:
@dannyroberts Should this go in deploy_control.yml maybe? I don't have strong feelings either way.

dannyroberts:
@czue curious what you're up to. Is this just trying to run ansible locally on VMs? I think the `provisioning/` dir is really only for using with vagrant. Currently setting up an actual new control machine for a new cluster isn't very well document, or well tested since it's something I've done once as opposed to repeating over and over to iron out the kinks.

deploy_control.yml (documented a tiny bit [here](https://confluence.dimagi.com/display/internal/Ansible#Ansible-Addingdevuserstothecontrol(salt)machine)) would be a good place to put something you want to be run on the control machine, but it's unclear that the thing you're running into is currently a problem in a production setting.

czue:
none of this is for prod - just trying to get a dev environment stood up

PR #201, opened 2015-09-02
don't install multiple versions of postgres

czue:
@benrudolph not sure if this is the right thing to do or not.

however as it currently works it installs both 9.4 and 9.1 and then fails trying to start them both up because there's a port conflict.


benrudolph:
Good catch

benrudolph:
The idea is to only install one but be able to install a different version on the same machine if necessary

PR #203, opened 2015-09-04
Switch to use new tableau machine

TylerSheffels:


TylerSheffels:
@dannyroberts @sravfeyn This switches us to using the new larger machine.

ansible/roles/nginx/vars/tableau.yml
```diff
@@ -13,7 +13,7 @@ nginx_sites:
    - "X-Forwarded-Host $host:443"
    location1:
     name: /
-    proxy_pass: http://tableau1.internal.commcarehq.org
+    proxy_pass: http://tableau3.internal.commcarehq.org
```
dannyroberts: I feel like there's a world where this lives in the config file, and that that world is a slightly better world. Maybe next time you have to move the server or reconfigure another one on a different setup, you can think about how to enter that world.

TylerSheffels: Agreed this is not the right place long term for this to live.

PR #209, opened 2015-09-10
Gzip all of the things

NoahCarnahan:
Right now we only gzip `text/html`. I expect performance gains if we gzip css and js as well. I came across this optimization potential while preparing for the performance session for the upcoming summit. Hold off on merge because I haven't actually tested this yet.

### DIFF
--- a/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/ansible/roles/nginx/templates/nginx.conf.j2
@@ -25,6 +25,21 @@ http {
 
         gzip on;
         gzip_disable "msie6";
+        gzip_types
+            text/html
+            text/css
+            text/javascript
+            text/xml
+            text/plain
+            text/x-component
+            application/javascript
+            application/json
+            application/xml
+            application/rss+xml
+            font/truetype
+            font/opentype
+            application/vnd.ms-fontobject
+            image/svg+xml;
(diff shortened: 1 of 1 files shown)


czue:
neat idea.

benrudolph:
yea big win

TylerSheffels:
Big +1, wonder if there is a good way to take a baseline before.

benrudolph:
would anything in the browser tab be an indicator? https://rpm.newrelic.com/accounts/366142/browser/5058236

NoahCarnahan:
Yeah, I think the best way to compare is to compare the "network" segment of the graph ben linked before and after.

NoahCarnahan:
I've deployed the change to hqproxy5, which is the proxy for staging, and static files on staging are now being gziped. Should we wait a little while to see if any issues arise before merging this? I can't think of what would go wrong, but I want to use some caution before merging a site-wide change.

czue:
not sure what else I'd recommend, seems like an easy thing to test and revert if issues.

NoahCarnahan:
(I've requested some basic QA just in case)

NoahCarnahan:
QA passed. Good to merge.

czue:
sweet

PR #220, opened 2015-09-30
add ucrreadonly in pg_hba

sravfeyn:


sravfeyn:
@TylerSheffels

czue:
@sravfeyn I think you need to also add code that creates this user if you want to roll this out globally

czue:
(will let @TylerSheffels @dannyroberts or @benrudolph confirm)

sravfeyn:
I was trying to find way to create the user globally in all environments, but couldn't find it easily. So I manually created it for india as it was urgent.

Any pointers to have ansible create this user? I found the task [here](https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/postgresql/tasks/main.yml#L91) but couldn't figure where to put the new username var.

benrudolph:
hey @sravfeyn that's a good example of how to create a user globally. since that info in sensitive we don't store it in the ansible repo itself. we have a `dev.yml` (https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/vars/dev.yml#L67) that shows you what that file looks like. on `salt` and `india6` you can modify the `india.yml` in the config dir in order to set those localsettings variable. happy to offline with you if you need me to point you to it

benrudolph:
just curious what the status of this is? is there a potential bug if the ucr user isn't in the prod database?

sravfeyn:
hey thanks for pinging ben. I haven't prioritized this as it wasn't as urgent. But I am planning to do it once bunch of ICDS things settle down.

benrudolph:
Sounds good, thanks

dannyroberts:

sravfeyn:
Also got done in a different way, so closing this one.

PR #221, opened 2015-10-01
Djangobower

benrudolph:
the ansible component for djangobower: https://github.com/dimagi/commcare-hq/pull/8364

### DIFF
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -7,10 +7,14 @@
   apt_repository: repo='ppa:git-core/ppa' state=present
   register: add_git_repo
 
+# http://stackoverflow.com/a/16303380/835696
+- name: Add git nodejs repo
+  apt_repository: repo='ppa:chris-lea/node.js' state=present
+  register: add_nodejs_repo
 
 - name: Update package list
   apt: update_cache=yes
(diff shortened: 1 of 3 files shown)


ansible/roles/common_installs/tasks/main.yml
```diff
@@ -32,7 +36,7 @@
     - htop
     - tmpreaper
     - mailutils
-    - npm
+    - nodejs
```
benrudolph: this caused me quite a bit of pain. we have different installs of node on different machines which made the path to `bower` different. i think this should be the right way; i know we've gone back and forth on nodejs and npm many times. with this ppa nodejs should install npm as well. see: http://stackoverflow.com/a/16303380/835696

@dannyroberts @TylerSheffels

TylerSheffels: I agree, this always seems to be a problem. is it worth specifying the version here e.g. `- nodejs=0.10.37` or whatever the exact version is?

benrudolph: we specify it when we use `n` to upgrade:
```
 - name: Update Node
   command: 'n 0.10.29'
```

benrudolph:
@proteusvacuum @sravfeyn this is g2g

PR #222, opened 2015-10-08
Add postgres log rotation

benrudolph:
@snopoke @TylerSheffels Adding postgres log rotation. went to debug some postgres stuff the other day and `ls` took minutes to run because there are so many files in there

### DIFF
--- a/ansible/deploy_db.yml
+++ b/ansible/deploy_db.yml
@@ -65,7 +65,7 @@
     - role: ansible-logrotate
       logrotate_scripts:
         - name: "postgresql_{{ deploy_env }}"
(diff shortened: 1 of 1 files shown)


benrudolph:
deployed this to staging, but not prod yet. will wait for review

snopoke:
Much needed, will do a closet look tomorrow

TylerSheffels:
Looks great to me, and with daily rotation I don't think there is a need for a size limit.

TylerSheffels:
@benrudolph Assuming that you will deploy this to prod and india

benrudolph:
yep will do. also i think postgres has its own sizing limit

PR #224, opened 2015-10-08
Optional scripts for machines

benrudolph:
@dannyroberts here was my idea for automating the log pulls for amelia. also think it'd be nice to use this playbook to deploy things optionally, like the touchforms restarter script that we have on only the prod machine
cc: @TylerSheffels


benrudolph:
(note i haven't actually run this code yet, will play around with it once i deploy the log format)

dannyroberts:
neat! Looks good

benrudolph:
cool g2g

PR #227, opened 2015-10-09
Revert "Revert "Custom ucr databases""

benrudolph:
@TylerSheffels


benrudolph:
@TylerSheffels do we need this still?

TylerSheffels:
@benrudolph Maybe at some point, but for now I will close

PR #232, opened 2015-10-16
install datadog cloudant integration

snopoke:
@benrudolph 

cc @millerdev


snopoke:
Already ran on prod

czue:
cool stuff! will let one of the ansible guys review

TylerSheffels:
This is really cool. Means that we could write any sort of monitoring integration that we felt would be helpful using this approach

benrudolph:
:+1:

PR #233, opened 2015-10-19
Remove sentry (and raven, the sentry client) from ansible settings

TylerSheffels:
Goes with: https://github.com/dimagi/commcare-hq/pull/8707/files
@dannyroberts @benrudolph 

Once this is merged, I will need to run (on each environment):
- [ ] Stop sentry service
- [ ] Remove sentry from config .yml files
- [ ] deploy_proxy
- [ ] postgres configuration (pg_hba.conf)
- [ ] --tags=localsettings


benrudolph:
should localsettings and restart services come before postgres conf? or does it not really matter

TylerSheffels:
@benrudolph I don't think it will really matter, the step I didn't include here is actually stopping the sentry service at which point the rest of this shouldn't have any effect.  Added that step in.

benrudolph:
cool,  just checking, i'll go ahead and merge then

PR #234, opened 2015-10-20
set log level for elasticsearch to ERROR

snopoke:
http://manage.dimagi.com/default.asp?184654


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -172,6 +172,11 @@ LOGGING = {
             'level': 'INFO',
             'propagate': False,
         },
+        'elasticsearch': {
+            'handlers': ['file', 'sentry'],
```
benrudolph: little late, but `sentry` is on its last legs

PR #236, opened 2015-10-21
Add S3 access and secret keys to localsettings

TylerSheffels:
@gcapalbo @dannyroberts 

Goes with: https://github.com/dimagi/commcare-hq/pull/8750


benrudolph:
ooooh yeaaa. also i can make a request to get staging machines with at least 8GB of RAM?

PR #238, opened 2015-10-21
Tweaks to get a clean install working again

czue:
Also adds config files that let you setup a local monolith instead of a 3-box cluster.

@benrudolph


benrudolph:
nice

benrudolph:
this is definitely more reasonably then continuously deploying to staging

TylerSheffels:
:+1:

PR #242, opened 2015-10-22
add zookeeper and kafka install scripts

czue:
keen to get feedback on this. borrowed heavily from https://github.com/rackerlabs/ansible-kafka

@benrudolph @TylerSheffels @dannyroberts @snopoke


ansible/roles/kafka/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+apache_mirror: apache.mirror.anlx.net
+kafka_version: 0.8.2.2
+kafka_user: kafka
+kafka_group: kafka
+kafka_data_dir: /opt/data/kafka/data
```
czue: putting on the encrypted drive for now, though i don't think is strictly necessary since only metadata is planned to pass through

ansible/roles/kafka/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
...
+kafka_conf_dir: /etc/kafka/config
+kafka_log_dir: /var/log/kafka
```
czue: not sure if this is right or it should go in `{{ log_home }}` (looked like I would have had to wrangle with permissions to make that work so just did this)

ansible/roles/kafka/templates/server.properties.j2
```diff
@@ -0,0 +1,119 @@
...
+broker.id={{ kafka_broker_id }}
+
+############################# Socket Server Settings #############################
+
+# The port the socket server listens on
+port=9092
+
+# Hostname the broker will bind to. If not set, the server will bind to all interfaces
+#host.name=localhost
+host.name={{ ansible_eth1['ipv4']['address'] }}
```
czue: this worked locally, but i'm guessing shouldn't be like this. what's the appropriate way to get the public IP of the current machine?

ansible/inventories/monolith
```diff
@@ -25,6 +25,12 @@
 [rabbitmq]
 192.168.33.21
 
+[zookeeper]
+192.168.33.21
+
+[kafka]
+192.168.33.21 kafka_broker_id=0
```
czue: modeled after how elasticsearch cluster names worked (each broker_id must be unique to the cluster)

czue:
have only played with this locally. will play around on staging now

ansible/roles/kafka/defaults/main.yml
TylerSheffels: My general model has been that logs that will output "commcareish errors" would go in `log_home`, things like pillowtop errors.  I leave things in `/var/log` that are more system errors, e.g. `/var/log/nginx/error.log` has errors when nginx cant start, or the process ends unexpectedly.

ansible/roles/kafka/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
+---
+
+- name: create kafka group
+  group: name={{ kafka_group }} system=yes state=present
+
+- name: add user "kafka"
+  user: name={{ kafka_group }} group={{ kafka_group }} shell=/sbin/nologin system=yes state=present
+
+- name: Download Kafka
+  get_url: url="http://{{ apache_mirror }}/kafka/{{ kafka_version }}/kafka_2.10-{{ kafka_version }}.tgz" dest=/opt/kafka_2.10-{{ kafka_version }}.tgz
```
benrudolph: what is the 2.10 if not the version? just curious

czue: yeah is a bit weird. is the scala version (which we aren't going to use/care about). the [downloads page](http://kafka.apache.org/downloads.html) recommended 2.10 if you don't care.

benrudolph: :+1:

ansible/roles/kafka/templates/server.properties.j2
snopoke: You could do this: https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/postgresql/templates/pg_hba.conf.j2#L110-110

snopoke:
Looks good

ansible/roles/kafka/templates/server.properties.j2
benrudolph: i think the line you're point to is actually the filter that will return true or false if the host is an ip address or not. i usually use this if i want to get the IP for the inventory_host (also in pg_hba):
```
{{ lookup('dig', inventory_host, wantlist=True)[0] }}
```
i hesitated to respond because i think there might be a better way?

TylerSheffels: I think you kind of stumbled on this, but you want the "gathered facts" about the system: 
http://docs.ansible.com/ansible/playbooks_variables.html#information-discovered-from-systems-facts

TylerSheffels: Specifically 'ansible_default_ipv4'

czue:
this is deployed to prod, india and staging. i think it's good to merge

benrudolph:
:confetti_ball:

PR #243, opened 2015-10-22
reduce slowlog max len

snopoke:
We aren't really using this data and even if we do it doesn't seem like we need that much

@benrudolph


benrudolph:
agreed

PR #245, opened 2015-10-28
Datadog nginx parser

benrudolph:
@snopoke have yet to give it a spin, but this is the general idea. 
related: https://github.com/dimagi/datadog-parsers/pull/1
cc: @esoergel


czue:
neat. can we do this with the couch logs too?

snopoke:
Looks good. :+1: for couch logs too.

PR #247, opened 2015-10-31
A few fixups for the datadog parser

benrudolph:
@snopoke i deployed this and here were a few changes i had to make. it's unclear to me whether or not it's working. i can't find the metric yet in datadoghq. i'd imagine it'd be under `nginx.timings` since that what i called it [here](https://github.com/dimagi/datadog-parsers/pull/1/files#diff-04f4e797004c16d4737c308c9202e55cR28). are there datadog log files that you know i could check out see what's happening?

### DIFF
--- a/ansible/roles/datadog/templates/datadog.conf.j2
+++ b/ansible/roles/datadog/templates/datadog.conf.j2
@@ -15,6 +15,6 @@ gce_updated_hostname: yes
 # DEPRECATED: use conf.d/disk.yaml instead to configure it
 use_mount: no
 
-{% if inventory_host in groups['proxy'] %}
-dogstreams: {{ log_home }}/{{ deploy_env }}-timing.log:{{ datadog_parsers_dest }}/nginx/timings:parse_nginx_timings
+{% if inventory_hostname in groups['proxy'] %}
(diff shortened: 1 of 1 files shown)


benrudolph:
(nvm found them in `/var/log/datadog`)

benrudolph:
woah! this is incredible. it does exactly what i hoped it'd do! you can now view times based on a url, status_code, etc.

also issue was resolved by properly putting 0755 on log dir. somehow the permissions got messed up and `all` wasn't able to read and execute

czue:
> woah! this is incredible. it does exactly what i hoped it'd do! you can now view times based on a url, status_code, etc.

@benrudolph where can i check this out??

czue:
([found it](https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1446225424050&to_ts=1446398224050&tile_size=m&exp_metric=nginx.timings&exp_scope=url%3A%2Fhome%2F&exp_agg=avg&exp_row_type=metric))

now we need to add things like `domain` and `report_type` as queryable fields :)

benrudolph:
yeah seriously that'd be pretty sweet. but this isn't a bad start. we could make a dashboard for product people that showed different important urls like `/home/` restore, receiver, locations etc.

PR #248, opened 2015-10-31
add web worker couch parser

benrudolph:
@snopoke @czue goes with https://github.com/dimagi/datadog-parsers/pull/2

### DIFF
--- a/ansible/roles/datadog/templates/datadog.conf.j2
+++ b/ansible/roles/datadog/templates/datadog.conf.j2
@@ -19,6 +19,6 @@ use_mount: no
 dogstreams: {{ log_home }}/{{ deploy_env }}-timing.log:{{ datadog_parsers_dest }}/nginx/timings:parse_nginx_timings
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


ansible/roles/datadog/templates/datadog.conf.j2
```diff
@@ -18,3 +18,7 @@ use_mount: no
 {% if inventory_host in groups['proxy'] %}
 dogstreams: {{ log_home }}/{{ deploy_env }}-timing.log:{{ datadog_parsers_dest }}/nginx/timings:parse_nginx_timings
 {% endif %}
+
+{% if inventory_hostname in groups['webworkers'] %}
```
czue: should we not also monitor pilllowtop and celery?

czue: if we do add them it'd be nice if we could also add the machine or service type to the log format

benrudolph: good call, forgot about that

PR #249, opened 2015-11-02
Settings rename / cleaup

snopoke:
See https://github.com/dimagi/commcare-hq/pull/8893


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -662,7 +649,7 @@ SMS_PREFIXES_BY_COUNTRY = '{{ localsettings.SMS_PREFIXES_BY_COUNTRY }}'
 {% endif %}
 
 {% if localsettings.get('MIA_THE_DEPLOY_BOT') %}
-MIA_THE_DEPLOY_BOT = '{{ localsettings.MIA_THE_DEPLOY_BOT }}'
+MIA_THE_DEPLOY_BOT_API = '{{ localsettings.MIA_THE_DEPLOY_BOT }}'
```
benrudolph: i think this should be `localsettings.MIA_THE_DEPLOY_BOT_API`

benrudolph: actually that depends if you're planning on changing the name in the `config` dir

benrudolph:
may be worthwhile to also change the var names in the `dev.yml` (and the config dirs) so that there remains a consistency between the two

sravfeyn:
This looks good. This needs to be coordinated with the deploy, right?

snopoke:
I've added back in the old settings as well so that I can deploy this now and merge the HQ PR. Once HQ has been deployed I'll remove the old settings.

czue:
you still want a don't merge label on this?

snopoke:
removed label, thanks

PR #256, opened 2015-11-05
Revert 64e49b673cc560264db2f900506ebc480c3717c7

benrudolph:
Reverts dimagi/commcarehq-ansible#222

@proteusvacuum we're using a different strategy for log rotation now, need to get rid of this


benrudolph:
jk it's already deleted

PR #260, opened 2015-11-11
Don't create log dir

benrudolph:
@czue @TylerSheffels this looks to be the culprit for the changing log dir permissions. we don't need to do this cause we already create the `log_home` in common with different permissions: https://github.com/dimagi/commcarehq-ansible/blob/nginx-log-dir-fix/ansible/roles/commcarehq/tasks/main.yml#L15

cc: @proteusvacuum


czue:
:+1: from me but will let @TylerSheffels merge

PR #261, opened 2015-11-13
Better dogstream format

benrudolph:
@dannyroberts this is a better dogstream format. wasn't as trivial as i had first thought, luckily the conf allows for trailing commas

@proteusvacuum


benrudolph:
already ran on prod

dannyroberts:
:pray: very nice of them on the trailing commas. Looks good to me, thanks!

PR #263, opened 2015-11-17
install ntp on all machines

czue:
the root cause of http://manage.dimagi.com/default.asp?186591 is that the clocks are out of sync on the machines by more than 10 minutes. yuck.

also verified that older machines (e.g. hqdb0) have ntp installed properly, but newer ones (e.g. hqdjango12) do not. 

I didn't run deploy_common because i was worried about side effects so just manually updated india. I *think* that prod might be fine because rackspace might handle it a level above our vms (?)

### DIFF
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -36,6 +36,7 @@
     - htop
     - tmpreaper
     - mailutils
+    - ntp
     - nodejs
     - at
     - screen


czue:
@TylerSheffels @dannyroberts

snopoke:
:+1:

benrudolph:
good find :mag:. i'm sure you already do this, but just in case, if i ever need to run just a few commands and not the whole `deploy_common`, i manually add a tag to the task i want to run. if you're worried about installing something weird, you can always comment out the other packages (note: this command uses `present` so it won't upgrade/install anything unless that package is not there even if there's a newer version). this way you don't have to log onto every machine to install it.

czue:
i didn't know that and good tip. thanks! :tophat:

PR #268, opened 2015-11-19
Fix syntax in template

benrudolph:
@emord @czue err slightly awkward commits posing as dan, but fixed this when deploying datadog


benrudolph:
actually just gonna close so people don't get confused when they do a git blame

czue:
lol

PR #270, opened 2015-11-19
use primary_ssl_env to determine the default nginx site

czue:
http://manage.dimagi.com/default.asp?187354

@TylerSheffels how does this look? cc @proteusvacuum


czue:
(updated the prod and staging vars on the ansible machine already and verified this works)

TylerSheffels:
Much better option than mine.

PR #271, opened 2015-11-19
Add X-Frame-Options

TylerSheffels:
This prevents someone from embedding HQ in another page using an Iframe and hijacking clicks to perform other actions. @czue 

cc: @dannyroberts


czue:
seems legit. hope no one was depending on this workflow!

benrudolph:
haha if they were we should definitely stop them. click jacking ain't cool and loading our site in an iframe is pretty sketch too

PR #272, opened 2015-11-19
update to new local settings

benrudolph:
@wpride

### DIFF
--- a/ansible/roles/touchforms/templates/localsettings.py.j2
+++ b/ansible/roles/touchforms/templates/localsettings.py.j2
@@ -3,7 +3,8 @@
 #
 import os
 
-URL_ROOT = "https://{{ SITE_HOST }}/a/{{ '{{' }}DOMAIN{{ '}}' }}"
+URL_HOST = "https://{{ SITE_HOST }}"
+URL_ROOT = "%s/a/{{ '{{' }}DOMAIN{{ '}}' }}" % URL_HOST
 PERSISTENCE_DIRECTORY = "{{ touchforms_data_dir }}"
 FORMPLAYER_LOG_FILE = os.path.join('{{ log_home }}',"formplayer.clean-{{ inventory_hostname_short }}.log")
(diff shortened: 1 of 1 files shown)


ansible/roles/touchforms/templates/localsettings.py.j2
```diff
@@ -3,7 +3,8 @@
 #
 import os
 
-URL_ROOT = "https://{{ SITE_HOST }}/a/{{ '{{' }}DOMAIN{{ '}}' }}"
+URL_HOST = "https://{{ SITE_HOST }}"
+URL_ROOT = "%s/a/{{ '{{' }}DOMAIN{{ '}}' }}" % URL_HOST
```
czue: i know it just moved but what is the deal with this syntax?

benrudolph: lol i honestly have no clue why it is that. probably should figure that out at some point. i always thought it was some sort of magic

snopoke: This is trying to produce the following `/a/{{ DOMAIN }}` which then gets rendered by touchforms in a horribly complicated way: https://github.com/dimagi/touchforms/blob/f49f796ddf21ca68f4e0bce5df97d2229a37d56f/touchforms/backend/util.py#L124-L124

The weird syntax is necessary to escape the braces however it would be much clearer if we use the `raw` template tag:
```
{% raw %}
URL_ROOT = "%s/a/{{ DOMAIN }}" % URL_HOST
{% endraw %}
```

Alternately we could just change the syntax to something that doesn't need to be escaped.

benrudolph: ah wow thanks for digging into that. makes sense. for now i'll update with `raw` just so i don't mess up other people's localsettings but agree probably a different syntax is best

czue: :dart: thanks for the info

PR #279, opened 2015-11-27
Can't use wildcards in package versions

kaapstorm:
Another little one. @benrudolph cc @gcapalbo


benrudolph:
Strange missed that. Thanks

PR #288, opened 2015-12-17
ddog logger

benrudolph:
@snopoke ansible version


benrudolph:
goes with: https://github.com/dimagi/commcare-hq/pull/9623

PR #289, opened 2015-12-17
add user,noauto to /opt/data mount

snopoke:
@kaapstorm

### DIFF
--- a/ansible/roles/ecryptfs/tasks/main.yml
+++ b/ansible/roles/ecryptfs/tasks/main.yml
@@ -34,8 +34,13 @@
 
 - name: Register mount configuration
   sudo: yes
-  shell: "cat /etc/mtab | grep '/opt/data' | sed -e 's/ rw/ user,noauto,rw/'"
-  register: mount_configuration
+  shell: "cat /etc/mtab | grep '/opt/data'"
+  register: mount_conf
+
+- name: add user, noauto to mount conf
+  shell: "echo {{ mount_conf.stdout }} | sed -e 's/rw/user,noauto,rw/'"
+  when: "'noauto' not in mount_conf.stdout"
+  register: updated_mount_conf
 
 - name: Create entry in fstab
   sudo: yes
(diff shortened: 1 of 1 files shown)


snopoke:
@dannyroberts @TylerSheffels this was an issue on SoftLayer. The 'user,noauto' were missing from /etc/fstab which prevented the machines from coming up after boot (halted while trying to mount /opt/data)

ansible/roles/ecryptfs/tasks/main.yml
```diff
@@ -34,7 +34,7 @@
 
 - name: Register mount configuration
   sudo: yes
-  shell: "tail /etc/mtab -n1"
+  shell: "cat /etc/mtab | grep '/opt/data' | sed -e 's/ rw/ user,noauto,rw/'"
```
snopoke: Not that comfortable with this. Any better ideas?

benrudolph: could you use [lineinfile](http://docs.ansible.com/ansible/lineinfile_module.html)? it looks like you're trying to change which users have access to `/opt/data` in that file, which seems like you could use that module

benrudolph: what's supposed to go into `mount_configuration`?

snopoke: suppose could have another task:
```
- name: add user, noauto
  when: "'noauto' not in mount_configuration.stdout"
  shell: "echo {{ mount_configuration.stdout }} | sed -e 's/ rw/ user,noauto,rw/'"
  register: mount_configuration
```

snopoke: The issue is that the line we're adding is coming from the output of `tail /etc/mtab` but we want to alter the line before writing it to the file.

benrudolph: oh i see. that is tricky. seems like `sed` is the easiest solution. it also seems odd though that it isn't there already when it adds `user,noauto` when you mount: https://github.com/dimagi/commcarehq-ansible/blob/ecryptfs/ansible/roles/ecryptfs/tasks/main.yml#L22

snopoke: yea, it is a bit odd but its definitely not there

dannyroberts:
Looks fine to me; @benrudolph might have more do say since he made the original setup

kaapstorm:
@snopoke subsequently discovered that in Ubuntu 12.04 /etc/mtab includes the "user,noauto" options but Ubuntu 14.04 drops them for some reason.

benrudolph:
it's @TylerSheffels's baby!

PR #293, opened 2015-12-20
Default to PostgreSQL 9.4

kaapstorm:
@snopoke buddies @sravfeyn @calellowitz

### DIFF
--- a/ansible/roles/postgresql/defaults/main.yml
+++ b/ansible/roles/postgresql/defaults/main.yml
@@ -1,5 +1,5 @@
 pg_install: True
-postgresql_version: '9.1'
+postgresql_version: '9.4'
 postgresql_port: 5432
 postgresql_ssl_enabled: True
 postgresql_dir_original_path: /var/lib/postgresql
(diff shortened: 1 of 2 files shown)


ansible/roles/postgresql/defaults/main.yml
```diff
@@ -1,5 +1,5 @@
 pg_install: True
-postgresql_version: '9.1'
+postgresql_version: '9.4'
```
sravfeyn: Does this place also need a version change? https://github.com/dimagi/commcarehq-ansible/blob/pg94/ansible/roles/common_installs/defaults/main.yml#L1

kaapstorm: Yes! Thank you! I was sure there were two places, but I only found this one. I'll update.

kaapstorm:
:white_check_mark:

snopoke:
How come it's defined in 2 places? Can't we change that?

kaapstorm:
That's a good idea, @snopoke. The one place is for the client and the other for the server, but iirc the client is common to all hosts, and the server should just use the client's version. It's on my to-do list.

PR #294, opened 2015-12-22
Secure Elasticsearch

kaapstorm:
**This is not tested.**

The new `elasticsearch_host` variable should be overridden in host-specific vars, and should be set to the internal IP address of the host. 

These changes are based on [Elasticsearch's blog post about the vulnerability](https://www.elastic.co/blog/scripting-security).

@czue, @snopoke, buddies @calellowitz, @sravfeyn

### DIFF
--- a/ansible/roles/elasticsearch/defaults/main.yml
+++ b/ansible/roles/elasticsearch/defaults/main.yml
@@ -1,3 +1,4 @@
 ---
 elasticsearch_version: 0.90.5
 elasticsearch_memory: 4096m
+elasticsearch_host: 127.0.0.1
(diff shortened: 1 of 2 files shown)


czue:
thanks! i'm going to do some testing with this so put a don't merge on it. I think we are going to need to tweak it a bit since we need ES to be available on the VPN interface and not just localhost but the direction seems good.

czue:
@kaapstorm I reverted https://github.com/dimagi/commcarehq-ansible/commit/7afba087cb329e5ffa840840cdd28e94c77ea2a7 in favor of https://github.com/dimagi/commcarehq-ansible/pull/295/ but the other change is g2g

PR #295, opened 2015-12-22
use ufw to restrict certain interfaces

czue:
@kaapstorm @snopoke this is deployed to softlayers and seems to be working well. rather than manually managing all our services to run on the right interfaces we can just disable all inbound traffic with a software firewall.

### DIFF
--- a/ansible/deploy_common.yml
+++ b/ansible/deploy_common.yml
@@ -8,7 +8,7 @@
   hosts: '!proxy'
   sudo: yes
   roles:
-    - {role: ufw, tags: ufw, when: ufw_enabled and ufw_private_interface is defined}
+    - {role: ufw, tags: ufw, when: ufw_private_interface is defined}
 
 
 - name: Datadog agent
--- a/ansible/deploy_common.yml
+++ b/ansible/deploy_common.yml
@@ -4,6 +4,13 @@
   roles:
     - common
 
+- name: ufw (firewall)
+  hosts: '!proxy'
+  sudo: yes
+  roles:
+    - {role: ufw, tags: ufw, when: ufw_enabled and ufw_private_interface is defined}
+
+
 - name: Datadog agent
   hosts: all
   sudo: yes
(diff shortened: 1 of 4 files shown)


ansible/roles/ufw/readme.md
```diff
@@ -0,0 +1,7 @@
+# UFW Module
+
+This module configures a simple firewall rule using [ufw](https://wiki.ubuntu.com/UncomplicatedFirewall).
+
+To enable it just specify a `ufw_private_interface` (e.g. `eth0`) in your environment variables.
+If you do this, inbound traffic on _all interfaces except that one_ will be blocked.
```
TylerSheffels: This seems semi dangerous, e.g. if the interface was named differently for some reason. Can we add a rule that is something like: `ufw: rule=allow direction=in from_ip={{ ansible_control_host_ip }}`. I don't know the order of rule resolution, or what the appropriate value for `ansible_control_host_ip` would be, but this would let us recover the server.

czue: that's a good idea. will look into figuring something like that out.

a simpler thing we could do is always allow ssh traffic, though that comes with some (though probably many fewer) risks.

czue: @TylerSheffels updated

czue:
I [tried to figure out a better way to get the control machine IP](http://stackoverflow.com/questions/34560622/best-way-to-get-the-ip-address-of-the-ansible-control-machine) but came up empty.

ansible/deploy_common.yml
```diff
@@ -4,6 +4,13 @@
   roles:
     - common
 
+- name: ufw (firewall)
+  hosts: '!proxy'
+  sudo: yes
+  roles:
+    - {role: ufw, tags: ufw, when: ufw_private_interface is defined and control_machine_ip is defined }
```
kaapstorm: :+1:

kaapstorm:
lgtm

PR #297, opened 2015-12-23
disallow root ssh access

czue:


TylerSheffels:
This prevents using a password to login as root, correct? I feel like we need this on the Indiacloud but will defer to @dannyroberts

czue:
oh good point. we could make it configurable or we could just wait until until the softlayers migration and then merge it.

it seems bad that we allow that on those machines and like this should always be the default setup

TylerSheffels:
Agree that this should be the protocol, would be happy to merge on the softlayer migration is complete. We should also be able to set up ssh keys between the india control machine and the other india machines if we want to merge this sooner.

czue:
that's fine, hopefully will be done in a week or less

czue:
@TylerSheffels I think this can be merged now

PR #298, opened 2015-12-29
Switch to use SAMEORIGIN for exframe options

TylerSheffels:
Fixes: http://manage.dimagi.com/default.asp?189184#1061462
@benrudolph, @nickpell, @orangejenny


TylerSheffels:
Deploying this to Prod + India now

PR #299, opened 2015-12-30
Remove tableau handling from nginx (project spinning down)

TylerSheffels:
Without new cases of Ebola in Guinea they are beginning to spin down the project and we won't be hosting Tableau any more
CC: @benrudolph


czue:
yay for epidemics ending

PR #302, opened 2016-01-07
Use raw and endraw for tf local settings

benrudolph:
@snopoke @TylerSheffels https://github.com/dimagi/commcarehq-ansible/pull/272/files#r49047688

### DIFF
--- a/ansible/roles/touchforms/templates/localsettings.py.j2
+++ b/ansible/roles/touchforms/templates/localsettings.py.j2
@@ -4,7 +4,9 @@
 import os
 
 URL_HOST = "https://{{ SITE_HOST }}"
-URL_ROOT = "%s/a/{{ '{{' }}DOMAIN{{ '}}' }}" % URL_HOST
+{% raw %}
+URL_ROOT = "%s/a/{{ DOMAIN }}" % URL_HOST
+{% endraw %}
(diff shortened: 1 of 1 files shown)


TylerSheffels:
@benrudolph Have you tested this? This thread seems to indicate it might not work like you think:  https://github.com/ansible/ansible/issues/4638

benrudolph:
yikes good catch. no not yet

benrudolph:
seems to work as expected despite that ticket:
screen shot 2016-01-07 at 11 28 29 am

benrudolph:
though that was just with `--check`

TylerSheffels:
looks good enough for me, thanks for checking

czue:
:+1: well that's quite a bit easier to understand

tamsky:
FYI nifty macro can do the same thing:

``` jinja
{# Macro which emits `expr` surrounded by literal double braces that Jinja ignores. #}
{%- macro _ (expr) -%}
  {{ '{{' }}{{ expr }}{{ '}}' }}
{%- endmacro -%}

...
URL_ROOT = "%s/a/{{ _('DOMAIN') }}" % URL_HOST
...
```

benrudolph:
thanks for the tip @tamsky! we're getting rid of this in the end, but will keep it in mind

PR #304, opened 2016-01-07
Remove host header to fix redirection (not clear cause)

TylerSheffels:
@sheelio


TylerSheffels:
Self merging to deploy this for Sheel

PR #305, opened 2016-01-11
Add shared dir to pillowtop server

millerdev:
Fix `BlobDeletionPillow` failure:

`corehq.blobs.exceptions.Error: cannot initialize blob db: shared drive path is not a directory: '/mnt/shared_data'`

@dannyroberts @czue


czue:
:+1: legit

millerdev:
I applied this on prod and started `commcare-hq-production-pillowtop-BlobDeletionPillow`

PR #306, opened 2016-01-14
Add shared_data_dir task for single-server setup

millerdev:
This is for servers like zambia that don't currently have a shared data dir. Use of the blob db requires a shared data dir.

@dannyroberts


ansible/roles/shared_data_dir/tasks/setup_host.yml
```diff
@@ -0,0 +1,42 @@
...
+- name: Create shared data dir
+  sudo: yes
+  file:
+    path: "{{ shared_data_dir }}"
+    owner: "nobody"
+    group: "{{ shared_dir_gid }}"
+    mode: "u=rwx,g=rwx,o=r"
+    state: directory
+  tags:
+    - shared_files
```
millerdev: Everything in this file was moved from `ansible/roles/shared_dir/tasks/setup_host.yml`. This line is the only thing that changed: `nfs` -> `shared_files`

dannyroberts:
I'll let @benrudolph take a look

benrudolph:
offlined with @millerdev, seems good to me

PR #307, opened 2016-01-15
Add sudo to SSL cert commands when running ansible as not root user

TylerSheffels:
This was a confusing one for me. On softlayer we don't run ansible as root, we smartly run it as the ansible user. The ansible user is a member of the sudoers group, so it can run any command it needs to, but for commands that require elevated permissions we need to specify `sudo: yes` on the actual task. We probably didn't notice this before because other environments run as root.

cc: @orangejenny @benrudolph @dannyroberts

### DIFF
--- a/ansible/roles/nginx/tasks/install.yml
+++ b/ansible/roles/nginx/tasks/install.yml
@@ -49,6 +49,7 @@
 
 
 - name: Copy cchq SSL cert
+  sudo: yes
   copy:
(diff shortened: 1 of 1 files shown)


czue:
nice. how did this work the first time?

TylerSheffels:
@czue Not really sure, might have initially run -u root? Couldn't really figure out how it got into the current state.

PR #309, opened 2016-01-18
Update MOTECH server config To properly pass host to DHIS2

sheelio:
@TylerSheffels can you review and merge?


czue:
looks ok to me but will wait for @TylerSheffels

TylerSheffels:
@sheelio Did you test and see that this was the config that made it work?

PR #311, opened 2016-01-19
Fix shared dirs in proxy config files

millerdev:
@benrudolph @dannyroberts @emord


benrudolph:
:+1: , does `--check` on prod return empty now?

millerdev:
I did not try that, but I will...

millerdev:
I tested on staging and zambia, which each have different values for `shared_drive_enabled`

millerdev:
Close to empty. I think fine as far as this PR is concerned. Messaging you offline with changes that I observed.

PR #317, opened 2016-01-22
Add postgres backups

TylerSheffels:
This is a first pass at adding manual backups of postgres. There are a number of improvements that I would like to make, but will come in a future PR I think:
- [ ] Monitoring of this job
- [ ] Backup other directories we care about (looking at you blobdb)

@czue @benrudolph @orangejenny


ansible/roles/postgresql/templates/create_postgres_dump.sh.j2
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+BACKUP_TYPE=$1
+DAYS_TO_RETAIN_BACKUPS=$2
+
+TODAY=$(date +"%Y_%m_%d")
+pg_dumpall --clean | gzip  > "/opt/data/backups/postgres_$BACKUP_TYPE_$TODAY.db.gz"
```
czue: nit but should we template this to use `postgresql_backup_dir`?

ansible/roles/postgresql/defaults/main.yml
```diff
@@ -4,6 +4,7 @@ postgresql_port: 5432
 postgresql_ssl_enabled: True
 postgresql_dir_original_path: /var/lib/postgresql
 postgresql_dir_path: /opt/data/postgresql
+postgresql_backup_dir: /opt/data/backups
```
czue: we might want to consider putting this in `/opt/data/backups/postgres` just so the backups are more organized when we are backing up other stuff

czue:
this looks great!

ansible/roles/postgresql/defaults/main.yml
snopoke: :+1:

benrudolph:
sweet! do we care about storing backups on a different machine than the actual database? if the hqdb0 goes (like the hard disk gets struck by :zap:), then the backups go down with it no?

czue:
in addition to this we're installing software that backs up the directory

TylerSheffels:
@benrudolph As Cory said, this is only half the backup process, the whole /opt/data/backups directory is pulled onto a separate disk after this completes (but that is configured through softlayer ui)

benrudolph:
ah awesome. makes sense, thanks for the info

PR #319, opened 2016-01-22
set up ansible with a password

emord:
http://manage.dimagi.com/default.asp?204574#1078258

Set up the ansible user with a password.

Used http://docs.ansible.com/ansible/faq.html#how-do-i-generate-crypted-passwords-for-the-user-module

After deploying  this using --ask-sudo-pass with ansible user should work on prod

cc @sravfeyn

### DIFF
--- a/ansible/vars/dev.yml
+++ b/ansible/vars/dev.yml
@@ -24,6 +24,7 @@ active_sites:
 
 deploy_env: dev
 fake_ssl_cert: yes
+ansible_user_password: 'YIN9OONyKqkxY' # blahblah
 
 dev_users:
   present:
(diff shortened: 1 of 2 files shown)


ansible/vars/dev.yml
```diff
@@ -24,6 +24,7 @@ active_sites:
 
 deploy_env: dev
 fake_ssl_cert: yes
+ansible_user_password: 'YIN9OONyKqkxY' # blahblah
```
czue: heh, for some weird reason I'd be more comfortable if this said "secret" or something just so it was obviously a test password...

PR #320, opened 2016-01-25
Test that localsettings are python parsable

benrudolph:
can't tell you how many times i've deployed bad strings to our localsettings only to realize an hour later when the webworker tries to restart and fails and causes the entire site to go down :sweat_smile: 

cc: @czue @esoergel


czue:
nice idea! looks like it was a battle!

benrudolph:
:facepunch:

PR #323, opened 2016-01-26
Be more specific with postgres backup dir, parametrize shell script

TylerSheffels:
@czue @orangejenny


czue:
good stuff, thanks for follow up

PR #324, opened 2016-01-27
Encrypt blob DB

millerdev:
Objective: move blob db data to encrypted drive.

This move should be transparent to the running application once completed because there will be a symlink in the location where the blobdb data was previously located.

Deployment plan (only in setups that do not use NFS):

Run the following (in screen) directly on servers to be migrated. When the
initial migration from couch to blob DB was performed there were no objects in
the blob DB for the setups on which this will be performed, so it is not a
conern that this will entail a short period in which blob DB access will error.
The second and third rsync commands can probably be omitted, but are included
here for completeness in case of unexpected data volume.

```sh
SHARED_DATA_DIR=/opt/shared_data...
rsync -av $SHARED_DATA_DIR/blobdb /opt/data/
# if that took a while, run again to bring up to update (should be quicker)
rsync -av $SHARED_DATA_DIR/blobdb /opt/data/

mv $SHARED_DATA_DIR/blobdb $SHARED_DATA_DIR/blobdb-old
# blobdb is now offline (error on access)
rsync -av --delete $SHARED_DATA_DIR/blobdb-old/ /opt/data/blobdb/
ln -s /opt/data/blobdb $SHARED_DATA_DIR/blobdb
# blobdb is now online
```

Finally, on ansible control machine

```sh
# should see no changes since everything is already in place
ENV=... && aps --tags=ecryptfs-blobdb --diff --check
```

@dannyroberts @benrudolph cc @emord

### DIFF
--- a/ansible/roles/ecryptfs/tasks/main.yml
+++ b/ansible/roles/ecryptfs/tasks/main.yml
@@ -22,6 +22,19 @@
   shell: "mount -t ecryptfs -o key=passphrase:passphrase_passwd={{ ECRYPTFS_PASSWORD }},user,noauto,ecryptfs_cipher=aes,ecryptfs_key_bytes=32,ecryptfs_unlink_sigs,ecryptfs_enable_filename_crypto=y,ecryptfs_fnek_sig={{ password_hash.stdout }},verbosity=0 /opt/data/ /opt/data/"
   when: "'/opt/data' not in mtab_contents.stdout"
 
+- name: Create /opt/data/blobdb directory
+  sudo: yes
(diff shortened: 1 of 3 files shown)


benrudolph:
plan sounds good. is there any concern that there will be some unmigrated data left in the shared dir? seems like it's somewhat dependent on how fast you type the commands. it's probably not worth the effort to ensure that there is absolutely no chance of unmigrated data since the machines you're doing it on aren't super active

millerdev:
> is there any concern that there will be some unmigrated data left in the shared dir?

I think the final rsync during the "offline" period will take care of any unmigrated data, so there shouldn't be any left when it's over. I'll probably delete `$SHARED_DATA_DIR/blobdb-old` when I'm finished.

PR #325, opened 2016-01-27
pin npm to latest version

dannyroberts:
Looks like we weren't specifying a version of npm before. I checked locally that `npm install -g npm@3.6.0` works.

@benrudolph

### DIFF
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -56,10 +56,11 @@
   command: 'npm config set registry http://registry.npmjs.org/'
   sudo: yes
 
-- name: Install npm lessc
+- name: Install npm items
   npm: name={{ item }} state=present global=yes
   sudo: yes
   with_items:
+    - npm@3.6.0
     - less@1.3.1
     - n@1.3.0
     - bower@1.5.3


benrudolph:
does that work? that's essentially saying `npm install npm -g`. we've traditionally used this to upgrade node: https://github.com/dimagi/commcarehq-ansible/blob/pin-npm/ansible/roles/common_installs/tasks/main.yml#L69

dannyroberts:
@benrudolph I saw that but I'm trying to upgrade `npm` not `node`. The assumption of course is that upgrading `npm` and upgrading `node` isn't the same thing. Is that wrong?

benrudolph:
ah interesting. i think you're right as is. sorry about the confusion

TylerSheffels:
@dannyroberts I assume your change is correct, but I remember running into a multitude of issues with npm containing a specific version of node, or the opposite, and one of those versions being very out of date. Don't remember the exact issues that came up, but just a note if you see something similar crop up.

dannyroberts:
OK, good to know. So far I've run this on staging and prod with no apparent problems. It might be that once you install npm indirectly through apt-get node, you can then self-upgrade it `npm update npm@x.x.x` just fine.

PR #329, opened 2016-02-01
use service instead of command

benrudolph:
@millerdev fixes the issue about starting riak and it failing hard. riak is now running on staging. you should also check out `ansible/roles/riakcs/README.md`

cc: @emord

### DIFF
--- a/ansible/roles/riakcs/tasks/admin_user.yml
+++ b/ansible/roles/riakcs/tasks/admin_user.yml
@@ -12,7 +12,7 @@
 
 # Note this can't be notify because we need it to restart immediately
 - name: Restart RiakCS
-  command: riak-cs restart
+  service: name=riak-cs state=started enabled=yes
 
 - name: Create admin user
   command: "curl -H 'Content-Type: application/json'   -XPOST http://{{ inventory_hostname }}:{{ riakcs_port }}/riak-cs/user   --data '{\"email\":\"{{ riakcs_admin_email }}\", \"name\":\"{{ riakcs_admin_name }}\"}'"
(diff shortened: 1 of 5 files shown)


ansible/roles/riakcs/tasks/admin_user.yml
```diff
@@ -12,7 +12,7 @@
 
 # Note this can't be notify because we need it to restart immediately
 - name: Restart RiakCS
-  command: riak-cs restart
+  service: name=riak-cs state=started enabled=yes
```
millerdev: Why are there two tasks in the file called `Restart RiakCS`? (see also https://github.com/dimagi/commcarehq-ansible/blob/fix-riak-start/ansible/roles/riakcs/tasks/admin_user.yml#L67)

millerdev: Also this is making sure the service is started. It does not restart the service if it's already running ([docs](http://docs.ansible.com/ansible/service_module.html#options)). Use `state=restarted` if it should always restart the service.

benrudolph: you're right, this should just be named Start RiakCS. we just need to make sure it's running. the command below needs to be a restart for the changes to take effect. will update that

benrudolph: the reason there's a bunch of these is that you can run this playbook in portions using tags/defaults and i make no assumption up the status of the service (beyond the fact that it's installed)

benrudolph: (actually it is supposed to be restart)

benrudolph: https://github.com/dimagi/commcarehq-ansible/commit/3b98546be6633d23fe65ea5eed05bcb1e8192d5e

PR #330, opened 2016-02-05
Blobdb backups

TylerSheffels:
This adds a similar script to the postgres backup script I wrote, but this one tar's the blobdb drive. I would like to do a bit of refactoring (this is a bit of copy/paste) to the postgres backup script after this is ironed out, but wanted to wait until I am clear what will end up being parametrized.

@kaapstorm @snopoke

### DIFF
--- a/ansible/deploy_db.yml
+++ b/ansible/deploy_db.yml
@@ -10,6 +10,7 @@
   sudo: yes
   roles:
     - {role: ecryptfs, tags: 'ecryptfs'}
+    - {role: backups, tags: 'backups'}
 
 - name: Postgresql
   hosts: postgresql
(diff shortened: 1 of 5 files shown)


snopoke:
isn't this just creating a copy on the same drive?

TylerSheffels:
Ha, pretty much. It is dependent on a second process which backs up a specific drive to external media. The only advantage of this vs just backing up the drive as is would be the compression from gzip and putting everything into a tarball.

kaapstorm:
(Most probably not relevant here because this is a DB, but sometimes useful for backups -- depending on `$DAYS_TO_RETAIN_BACKUPS` on a single device, and the amount of churn on the files to be backed up, you can sometimes get pretty good use of space without compression but with deduplication by using rsync for backups with `rsync --link-dest` to hardlink to the previous backup's unchanged files. (Assumes a filesystem that supports hardlinks.) Restores are very quick because there is no compression, and size can be comparable, or better if you have low churn and high retention.)

czue:
:+1:

PR #331, opened 2016-02-05
allow groovy scripts on ES

snopoke:
@esoergel @TylerSheffels

### DIFF
--- a/ansible/es_rolling_restart.yml
+++ b/ansible/es_rolling_restart.yml
@@ -6,11 +6,13 @@
   vars_prompt:
     - name: "pillows_stopped"
       prompt: "Have you stopped all the elastic pillows? [YES/NO]"
+      private: no
     - name: "confirmation"
       prompt: "Are you sure you want to do a rolling restart of the ES cluster? REALLY SURE? [YES/NO]"
+      private: no
   tasks:
     - assert:
-      that:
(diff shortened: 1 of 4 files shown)


snopoke:
Already run on prod

benrudolph:
:v: groovy. looks good, i know in the past we've just restarted ES. what's the advantage here of shutting down the pillows and doing these extra steps?

benrudolph:
(just saw your other PR with the explanation)

snopoke:
Yea, if you just restart all the nodes then any ES functions on HQ are basically down (error).

TylerSheffels:
:+1: Based on the number of retries + timeout on each one, this potentially runs a really long time. I don't think this is necessarily a bad thing, e.g. just set it running then do something else but it might make for a really long iteration time (although, I assume ES doesn't take all that long to restart).

snopoke:
Yea, this took about 2 mins per server on prod but can definitely take longer with more nodes / shards etc. or if you don't stop the pillows.

esoergel:
Looks good, thanks Simon!

czue:
classic pact being a bummer.

czue:
this is totally baller btw.

snopoke:
:grinning:

PR #334, opened 2016-02-17
adding analytics log file

calellowitz:
logging was set up for analytics a while ago by @sravfeyn but the log files were never created or added to localsettings
@millerdev


sravfeyn:
Thanks Cal! This will make things easy to debug. (I didn't setup the original analytics logging)

PR #336, opened 2016-02-23
J2me

emord:
Deployed this on staging at j2mestaging.commcarehq.org

Assuming we want a real QA on this, but I was able to restore a user and submit a form.

cc: @calellowitz


ansible/deploy_proxy.yml
```diff
@@ -25,6 +25,7 @@
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.wiki == True, action: site, site_name: wiki_http }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.motech == True, action: site, site_name: motech}
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.motech == True, action: site, site_name: motech_http }
+    - { role: nginx, when: proxy_type == 'nginx' and active_sites.cchq_ssl_j2me == True, action: site, site_name: cchq_ssl_j2me }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_ssl == True, action: site, site_name: commtrack_ssl }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.riakcs == True, action: site, site_name: riakcs }
```
emord: Wasn't sure if riackcs should be down here after commtrack_ssl

czue:
I think submit and sync are basically all that we need to test (on j2me)

ansible/deploy_proxy.yml
benrudolph: yeah for now OK to leave. it sets up the config for proxying to riakcs. not used on prod/india, just staging

PR #337, opened 2016-02-26
STATIC_CDN variable

benrudolph:
@dannyroberts 
cc: @NoahCarnahan


benrudolph:
err don't merge with @emord commits yet

benrudolph:
boom. this works. :boom:

ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -20,6 +20,7 @@ nginx_sites:
    location2:
     name: "/static"
     alias: "{{ nginx_static_home }}"
+    add_header: "Access-Control-Allow-Origin *;"
```
TylerSheffels: I'm not sure if this opens us up to a CSRF attack. I think it doesn't, because this should only affect static files, but does that sound correct?

http://stackoverflow.com/questions/19322973/security-implications-of-adding-all-domains-to-cors-access-control-allow-origin/20186782#20186782

benrudolph: that's correct. we just need the static resource to be able to be loaded from a different origin. namely our CDN URL. don't think this really opens us up to any security thing because as you mentioned this is just static files. here are some cURLs:
screen shot 2016-02-28 at 11 06 05 am

benrudolph: second packet doesn't have Acces-Control-Allow-Origin set

emord: I don't think this should have the `;` at the end.  causes it to write `add_header Access-Control-Allow-Origin *;;` which makes nginx fail

benrudolph: thanks, was hand editting the nginx conf and hadn't given it a go yet. will definitely ensure that it runs on staging before deploying to prod

PR #338, opened 2016-02-29
add cache-control header

benrudolph:
@dannyroberts did some reading on the cache control header. it's still a little unclear to me why our static files were being cached at all since we never added that header. i'm assuming it has to do with the fact that it was coming from the same origin. this allows any system to cache our static files, which i think is fine since anyone can access our static files from our site anyways.

this will cache for a month. we should start seeing this effect take place on staging over the next 24 hours when the amazon cloudfront cdn starts expiring static files. note the Cache-Control header
screen shot 2016-02-29 at 3 14 30 pm


dannyroberts:
Sweet!

PR #339, opened 2016-03-02
adding port to psql address for ucr

calellowitz:
implementing the idea of @proteusvacuum from http://manage.dimagi.com/default.asp?217882
UCR postgres requests were trying to connect on the wrong port. it was only a problem on staging but limited testing on prod indicates it shouldnt break anything
this is deployed to staging and i tested it in a private release on prod
cc: @emord

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -26,7 +26,7 @@ SQL_REPORTING_OBJECT_OWNER = '{{ localsettings.REPORT_OBJECT_OWNER }}'
 {% endif %}
 
 {% if localsettings.get('UCR_DATABASE_NAME') %}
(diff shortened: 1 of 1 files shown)


proteusvacuum:
Nice. Thanks for picking that up. :+1:

czue:
just to check - before we didn't specify a port and it was just using the default?

calellowitz:
Most of my context here comes from what was happening on staging, but yes we did not specify a port before so it was trying to connect to postgres on port 5432 (the default) but pgbouncer runs on 6432 and postgres was not accecpting connections on 5432.
It is not entirely clear to me on prod why the default was working, but when i set up a private release and added the port number to the localsettings, I was able to execute ucr functions successfully so this fix, seems harmless on prod as well.

czue:
all seems legit

PR #341, opened 2016-03-02
Custom icds cert

TylerSheffels:
Because the ICDS tableau server is hosted on the India cloud we need a new certificate for an endpoint that isn't india.commcarehq.org.  There is a bit of copy pasting in the copying certs to proxy step that could be refactored in the future.

@nickpell

### DIFF
--- a/ansible/deploy_proxy.yml
+++ b/ansible/deploy_proxy.yml
@@ -13,8 +13,6 @@
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.cchq_http == True, action: site, site_name: cchq_http }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.cchq_http_redirect == True, action: site, site_name: cchq_http_redirect }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.hq_status == True, action: site, site_name: hq_status }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.icds_tableau == True, action: site, site_name: icds_tableau }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.icds_tableau == True, action: site, site_name: icds_tableau_http }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.dhis == True, action: site, site_name: dhis }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.dhis == True, action: site, site_name: dhis_http }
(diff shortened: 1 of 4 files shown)


czue:
Do you need to change this based on http://manage.dimagi.com/default.asp?219729?

TylerSheffels:
I don't believe so, I think the appropriate change is adding this line to the softlayer configuration:
`primary_ssl_env: "softlayer"`

ansible/deploy_proxy.yml
```diff
@@ -25,10 +23,12 @@
...
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.motech == True, action: site, site_name: motech}
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.motech == True, action: site, site_name: motech_http }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_ssl == True, action: site, site_name: commtrack_ssl }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.riakcs == True, action: site, site_name: riakcs }
     # Don't put any .commcarehq.org ssl configurations below commtrack, the ssl cert variable that gets loaded persists between roles
+    - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_ssl == True, action: site, site_name: commtrack_ssl }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_http == True, action: site, site_name: commtrack_http }
+    - { role: nginx, when: proxy_type == 'nginx' and active_sites.icds_tableau == True, action: site, site_name: icds_tableau }
+    - { role: nginx, when: proxy_type == 'nginx' and active_sites.icds_tableau == True, action: site, site_name: icds_tableau_http }
```
czue: mostly just pattern matching, but confirming this shouldn't read `active_sites.icds_tableau_http`?

czue:
one question, but otherwise looks good. confirming that you will test/fix the `primary_ssl_env` thing before merge/deploy

ansible/deploy_proxy.yml
TylerSheffels: Yeah, at some point I realized that anytime you want to deploy the SSL endpoint you also want to deploy the redirection endpoint, so specifying icds_tableau will deploy both endpoints.

TylerSheffels:
Confirmed that the primary_ssl_env corrects the login issue.

czue:
:+1:

PR #345, opened 2016-03-08
monolith setup

benrudolph:
@czue some instructions

cc: @gcapalbo


czue:
thanks for the :books: !

PR #348, opened 2016-03-09
Fix ecrpytfs mount check and write to fstab

millerdev:
`/etc/fstab` was getting two lines to mount `/dev/xvdb1` and none to mount ecryptfs.

@benrudolph @dannyroberts @TylerSheffels cc @snopoke


dannyroberts:
nice

PR #349, opened 2016-03-10
Fix broken nginx config

millerdev:
Basically reverts dimagi/commcarehq-ansible#270

Not sure what changed, but recently when deploying the proxy it wants to do this:
```nginx
TASK: [nginx | Create the site configurations] ********************************
--- before: /etc/nginx/sites-available/production_commcare
+++ after: /home/salt/ansible/commcarehq-ansible/ansible/roles/nginx/templates/site.j2
@@ -16,21 +16,21 @@

   ssl_certificate /etc/pki/tls/certs/production_nginx_combined.crt;
   ssl_certificate_key /etc/pki/tls/private/production_nginx_commcarehq.org.key;
   ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
   ssl_ciphers ...removed for brevity...;

       server_name www.commcarehq.org;
                     add_header X-Frame-Options SAMEORIGIN;
       client_max_body_size 100m;
       access_log /home/cchq/www/production/log/production-timing.log timing;
-      listen 443 ssl;
+      listen 443 ssl default_server;

     proxy_set_header Host $http_host;
     proxy_set_header X-Forwarded-For $remote_addr;
```
which causes nginx to not start. I had this happen on staging, and subsequently edited the config file to get nginx to start. @benrudolph had it happen on prod; he also reverted manually.

I did some checking, and the `default_server` is not anywhere in the prod nginx configs (on the proxy: `grep -rni default_server /etc/nginx/` returns nothing).

@czue @TylerSheffels

### DIFF
--- a/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -2,7 +2,7 @@
 nginx_sites:
 - server:
    file_name: "{{ deploy_env }}_commcare"
-   listen: "443 ssl{{ ' default_server' if deploy_env == primary_ssl_env else '' }}"
+   listen: "443 ssl"
    server_name: "{{ SITE_HOST }}"
    client_max_body_size: 100m
    balancer: webworkers
--- a/ansible/roles/nginx/vars/main.yml
+++ b/ansible/roles/nginx/vars/main.yml
@@ -29,7 +29,6 @@ nginx_http_params:
   types_hash_max_size: 2048
 
 nginx_separate_logs_per_site: True
-primary_ssl_env: ""
 commcarehq_errors_repository: "https://github.com/dimagi/commcare-hq-errorpages.git"
 errors_home: "{{ www_home }}/error_root"
 error_pages:


czue:
weird. what's not valid / what error do you get? it seems like it should be a supported option: http://nginx.org/en/docs/http/ngx_http_core_module.html#listen

czue:
on the one hand, if prod is already setup this way it seems like this is safe. on the other, that bug was a huge nightmare to track down, and we've been bit by bad nginx config/routing causing various random device/ssl/auth issues, so I'd like to understand what changed and when.

millerdev:
@czue 
> what's not valid / what error do you get?

```
2016/03/07 20:59:46 [emerg] 6780#0: a duplicate default server for 0.0.0.0:443 in /etc/nginx/sites-enabled/staging_commcare_j2me:25
```

czue:
so I think that's the problem. `staging_commcare_j2me` should not have it enabled.

millerdev:
@benrudolph do you remember what error you saw on prod?

benrudolph:
so on prod that change hadn't been made when i deployed. as soon as i deployed the proxy went down and i immediately reverted those changes. it still didn't start. i ended up having to remove a stale icds site in the sites enabled. i didn't take the time to figure out if it would've worked with the `default_server` in and the icds site out. working on an update so the nginx deploy will remove stale sites

benrudolph:
to be clear, the error from the icds site was it had a fault private key path. i did not check to see if `default_server` caused any error, but it does some odd that it hadn't been deployed

millerdev:
Sounds inconclusive that it breaks prod, so closing for now. It definitely breaks staging, but that probably needs a different fix. I'm not sure how `default_server` is ending up in multiple files since it appears that `deploy_proxy.yml` only references `cchq_ssl` once.

czue:
so are we confident that we can deploy prod and not mess anything up without this?

millerdev:
I think we're not sure. @benrudolph's responses above didn't lead me to believe that prod was definitely broken by the thing that this removes.

emord:
Sorry about this. it was related to the j2me work. 

Right now our ansible deploys don't clean up extra nginx configs, and you can only have 1 `default_server` in an nginx config. For now, prod needs to have `default_server` on the prod site for j2me phones to work as they don't support [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication).

I don't think it should have affected the private key path on icds on prod.

PR #350, opened 2016-03-10
Load Riak CS keys for localsettings in check mode

millerdev:
@benrudolph

### DIFF
--- a/ansible/deploy_commcarehq.yml
+++ b/ansible/deploy_commcarehq.yml
@@ -5,7 +5,7 @@
   # - celery
   # - proxy ?? maybe
   # - pillowtop
-  hosts: all
+  hosts: all:!riakcs:!stanchion
   sudo: yes
   roles:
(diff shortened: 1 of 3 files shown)


ansible/deploy_commcarehq.yml
```diff
@@ -5,7 +5,7 @@
   # - celery
   # - proxy ?? maybe
   # - pillowtop
-  hosts: all
+  hosts: all:!riakcs:!stanchion
```
millerdev: Don't put localsettings on riakcs servers. We should add more host groups here (or only list host groups where it makes sense to have localsettings). `hqes2` and `hqes3` failed when I ran this against prod, and I assume that's fine and normal.

benrudolph: if riakcs is deployed to a machine where we do need localsettings will this work? such as staging?

benrudolph: actually most definitely yes

PR #351, opened 2016-03-10
Add datadog agent config for riak and riakcs

millerdev:
This has been deployed to the prod cluster.

@snopoke cc @dannyroberts @benrudolph

### DIFF
--- a/ansible/group_vars/riakcs.yml
+++ b/ansible/group_vars/riakcs.yml
@@ -1,0 +1,7 @@
+#datavol_device=... # defined in commcare-hq/fab/inventories/* if applicable
+# http://docs.basho.com/riak/latest/ops/tuning/linux/#Storage-and-File-System-Tuning
+datavol_opts: barrier=0,data=writeback
+
+datadog_integrations:
+  - riak
+  - riakcs
--- a/ansible/roles/datadog/templates/riakcs.yaml.j2
+++ b/ansible/roles/datadog/templates/riakcs.yaml.j2
@@ -2,5 +2,8 @@ init_config:
 
 # key values are set by ../../roles/riakcs/sync_keys/main.yml
 instances:
-  - access_id: access-key
+  - host: {{ inventory_hostname if inventory_hostname|ipaddr else lookup('dig', inventory_hostname, wantlist=True)[0] }}
+    port: {{ riakcs_port }}
+    is_secure: False
+    access_id: access-key
     access_secret: access-secret
--- a/ansible/roles/datadog/templates/riakcs.yaml.j2
+++ b/ansible/roles/datadog/templates/riakcs.yaml.j2
@@ -1,0 +1,6 @@
+init_config:
+
+# key values are set by ../../roles/riakcs/sync_keys/main.yml
+instances:
+  - access_id: access-key
+    access_secret: access-secret
--- a/ansible/roles/riakcs/README.md
+++ b/ansible/roles/riakcs/README.md
@@ -8,7 +8,9 @@ For a system overview look [here](https://docs.google.com/document/d/1F5KjtyXmvG
 
 2. Run `deploy_proxy` to update Nginx with the new configs to allow webworkers to communicate with RiakCS cluster
 
-3. Run `deploy_riakcs` to deploy the Riak CS cluster
+3. Optional: run `deploy_stack --tags=datadog --limit=riakcs` to setup the datadog monitoring agent.
+
+4. Run `deploy_riakcs` to deploy the Riak CS cluster
 
 ## Role layout
 
--- a/ansible/roles/riakcs/sync_keys/tasks/main.yml
+++ b/ansible/roles/riakcs/sync_keys/tasks/main.yml
@@ -24,8 +24,8 @@
 - name: Update access_id in riakcs datadog agent config
   lineinfile:
     dest: "/etc/dd-agent/conf.d/riakcs.yaml"
-    regexp: "([ -]+access_id: ).*$"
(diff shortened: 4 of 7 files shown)


ansible/group_vars/riakcs.yml
```diff
@@ -0,0 +1,7 @@
+#datavol_device=... # defined in commcare-hq/fab/inventories/* if applicable
+# http://docs.basho.com/riak/latest/ops/tuning/linux/#Storage-and-File-System-Tuning
+datavol_opts: barrier=0,data=writeback
```
millerdev: @dannyroberts this is moving from [hq prod inventory](https://github.com/dimagi/commcare-hq/blob/3508bc5446f97ecb34ccfe2a9c6a4873581fc9e0/fab/inventory/production#L61-L62) since it's a configuration detail. The only thing left in the inventory is whether or not the hosts have a `datavol_device`. I'll be submitting a PR to remove those lines from the prod inventory soon.

ansible/roles/riakcs/sync_keys/tasks/main.yml
```diff
@@ -15,3 +15,21 @@
...
+  notify: restart datadog
+
+- name: Update access_secret in riakcs datadog agent config
+  lineinfile:
+    dest: "/etc/dd-agent/conf.d/riakcs.yaml"
+    regexp: "([ -]+ access_secret: ).*$"
+    line: "\\1{{ hostvars[riakcs_control].get('riak_secret', 'not-set') }}"
+    backrefs: yes
+  when: "name == 'riakcs' and 'riak_key' in hostvars[riakcs_control] and 'riak_secret' in hostvars[riakcs_control]"
+  notify: restart datadog
```
millerdev: These new tasks mean that datadog must be installed before riakcs can be installed. Does that seem like a reasonable dependency? Should I make these tasks conditional on whether datadog is present or not? Actually, I think the dependency should be on the presence of the `datadog-agent` service and possibly `/etc/dd-agent/conf.d`. Thoughts welcome.

dannyroberts: I think if we ever want to install a node without datadog, we can fix it then? If it's easy to add the conditional I don't think it'd hurt, but this doesn't offend me too much, unless we're already *not* installing datadog in some environments. (Not sure, for example, if staging, india (softlayer), etc. have datadog installed.)

millerdev: It was easy. Also, we're not using datadog on staging, so it would have been a problem there.

ansible/roles/datadog/templates/riakcs.yaml.j2
```diff
@@ -0,0 +1,9 @@
+init_config:
+
+# key values are set by ../../roles/riakcs/sync_keys/main.yml
+instances:
+  - host: {{ inventory_hostname if inventory_hostname|ipaddr else lookup('dig', inventory_hostname, wantlist=True)[0] }}
+    port: {{ riakcs_port }}
+    is_secure: False
+    access_id: access-key
+    access_secret: access-secret
```
millerdev: This config appears to be correct, but the datadog riakcs integration is still reporting an error. See https://github.com/DataDog/dd-agent/issues/2340 for more details.

ansible/group_vars/riakcs.yml
snopoke: nit: to me this should belong in `roles/riakcs/defaults/main.yml`

ansible/roles/riakcs/README.md
```diff
@@ -8,7 +8,9 @@ For a system overview look [here](https://docs.google.com/document/d/1F5KjtyXmvG
 
 2. Run `deploy_proxy` to update Nginx with the new configs to allow webworkers to communicate with RiakCS cluster
 
-3. Run `deploy_riakcs` to deploy the Riak CS cluster
+3. Optional: run `deploy_stack --tags=datadog --limit=riakcs` to setup the datadog monitoring agent.
+
+4. Run `deploy_riakcs` to deploy the Riak CS cluster
```
snopoke: Would it not make more sense to run the datadog tasks after RiakCS is setup and use the `get_admin_keys` tasks to populate the keys in the context? That would make the datadog command idempotent which I don't think it is now.

ansible/group_vars/riakcs.yml
millerdev: no such directory or file exists. `roles/riakcs` contains multiple sub-roles. Also, it's referenced in `roles/ecryptfs/tasks/main.yml`, so I'm not sure if putting it in the `riakcs` role would work.

ansible/roles/riakcs/README.md
millerdev: I agree, that would be nice, but it would be a larger structural change: the datadog tasks would need to be moved after `deploy_riakcs`, which is included by `deploy_db`. Currently the datadog tasks are part of `deploy_common`, which happens before anything else.

Actually, it might make a lot of sense to move the datadog tasks to the end of `deploy_stack` since it would be better to configure the services that datadog will be monitoring before setting up the monitoring service itself so we don't end up with service outage alerts during initial configuration of new servers.

millerdev: @snopoke is this a blocker for you? I'd prefer to merge this PR and work on moving datadog to somewhere later in the stack setup as a separate task. However, that doesn't feel like a priority right now.

snopoke: Nope, not a  blocker. Feel free to merge.

PR #353, opened 2016-03-10
Delete sites that are set to false

benrudolph:
@TylerSheffels @millerdev i refactored the nginx conf a bit to delete any sites that are set to false. wondering if you think i should just delete everything in the `sites-enabled` dir even if it's not explicitly set to false in the conf (that would be slightly more difficult given the setup). this at least will provide us with a little less of a headache. still need to test on staging/prod. have just been testing this locally

cc: @gcapalbo


ansible/deploy_proxy.yml
```diff
@@ -9,26 +9,20 @@
...
+    - { role: nginx, when: proxy_type == 'nginx', action: site, site_name: motech}
+    - { role: nginx, when: proxy_type == 'nginx', action: site, site_name: riakcs }
     # Don't put any .commcarehq.org ssl configurations below commtrack, the ssl cert variable that gets loaded persists between roles
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_ssl == True, action: site, site_name: commtrack_ssl }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.commtrack_http == True, action: site, site_name: commtrack_http }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.icds_tableau == True, action: site, site_name: icds_tableau }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.icds_tableau == True, action: site, site_name: icds_tableau_http }
+    - { role: nginx, when: proxy_type == 'nginx', action: site, site_name: commtrack_ssl }
+    - { role: nginx, when: proxy_type == 'nginx', action: site, site_name: commtrack_http }
+    - { role: nginx, when: proxy_type == 'nginx', action: site, site_name: icds_tableau }
```
millerdev: I like this change, if for no other reason than the lines are shorter and easier to read.

millerdev:
I don't feel like I have a lot of context here, so maybe not the best judge.

PR #364, opened 2016-03-22
easy_install setup tools

benrudolph:
@emord upgrades our setup tools. staging was already on 18, this bumps us to 20. no breaking changes that looked relevant to us. will leave on staging for a bit before deploying to the rest of the machines

cc: @gcapalbo

### DIFF
--- a/ansible/roles/commcarehq/tasks/main.yml
+++ b/ansible/roles/commcarehq/tasks/main.yml
@@ -60,6 +60,12 @@
     - slow
   when: not gitdir.stat.exists and not testing|default(False)
 
+- name: Upgrade setuptools
+  easy_install:
+    # HACK: -U allows us to force update. In ansible 2.0 we can just use the state property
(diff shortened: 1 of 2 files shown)


ansible/roles/commcarehq/tasks/main.yml
```diff
@@ -60,6 +60,12 @@
     - slow
   when: not gitdir.stat.exists and not testing|default(False)
 
+- name: Upgrade setuptools
+  easy_install:
+    # HACK: -U allows us to force update. In ansible 2.0 we can just use the state property
+    name: "-U setuptools"
```
benrudolph: super surprised this actually worked

emord:
didn't mean to merge this, but merge and comment buttons are both green.

anyways it looks good to me, but can revert if you think you need to test it more

benrudolph:
yep should be fine

PR #365, opened 2016-03-22
Rackspace cloud agent

emord:
@czue @benrudolph @calellowitz 

http://manage.dimagi.com/default.asp?221540#1113032

I have a manual backup running at the moment on staging (will auto email devops once done), but looks like this works. Most files are just from installing https://github.com/rack-roles/cloud-backup-agent


benrudolph:
solid :computer: :recycle: :computer:

czue:
nice. let's run this on prod stat.

PR #366, opened 2016-03-23
Make deploy_stack work

esoergel:
Ran into a number of problems while trying to do an ansible deploy to new machines.  These are the changes I made to get it to pass
@benrudolph @millerdev @dannyroberts

### DIFF
--- a/ansible/roles/elasticsearch/tasks/main.yml
+++ b/ansible/roles/elasticsearch/tasks/main.yml
@@ -85,7 +85,7 @@
   sudo: yes
   service: name=elasticsearch state=started
   register: result
-  failed_when: result.rc != 0 and 'already running' not in result.stderr
+  failed_when: result.state != "started"
 
 - name: Restart Elasticsearch
   sudo: yes
--- a/ansible/roles/touchforms/tasks/main.yml
+++ b/ansible/roles/touchforms/tasks/main.yml
@@ -47,6 +47,13 @@
     - sqlite3
     - libsqlite3-dev
 
+- name: Check formplayer existence
+  stat:
+    path: "{{ code_home }}/submodules/formplayer"
+  register: formplayer_path
+
+#- include_vars: "supervisor_tasks.yml"
+#- include: ../../supervisor/tasks/supervised_tasks.yml
 - name: Copy Spring application.properties
   sudo: yes
   template:
@@ -57,3 +64,4 @@
     mode: 0644
   tags:
     - tf-localsettings
+  when: formplayer_path.stat.exists


ansible/group_vars/all.yml
```diff
@@ -31,8 +31,8 @@ stanchion_data_dir: "{{ encrypted_root }}/stanchion"
 # commcare-hq connects to an S3-compatible service, which is Riak CS
 s3_blob_db_enabled: "{{ 'true' if groups.get('riakcs') else '' }}"
 s3_blob_db_url: "http://{{ groups['proxy'][0] }}:{{ riakcs_port }}"
-s3_blob_db_access_key: "{{ hostvars[riakcs_control]['riak_key'] if s3_blob_db_enabled else '' }}"
-s3_blob_db_secret_key: "{{ hostvars[riakcs_control]['riak_secret'] if s3_blob_db_enabled else '' }}"
+s3_blob_db_access_key: "{{ hostvars[riakcs_control].get('riak_key', 'not-set') if s3_blob_db_enabled else '' }}"
+s3_blob_db_secret_key: "{{ hostvars[riakcs_control].get('riak_secret', 'not-set') if s3_blob_db_enabled else '' }}"
```
esoergel: @millerdev does this look like the right approach to you?  I saw that `'riak_secret'` was being used as a fallback everywhere else this variable is accessed, so I copied it here too.

benrudolph:
nice :+1: will wait for feedback from daniel

ansible/group_vars/all.yml
millerdev: I don't think this is right. The `riak_key` and `riak_secret` variables should be set any time that these variables are accessed and `s3_blob_db_enabled` evaluates to true. If they're not set then it seems like something is wrong with the environment. Can you give me reproduction steps for the problem that caused you to write this?

ansible/roles/riakcs/join_cluster.yml
```diff
@@ -15,7 +15,7 @@
 
 #- debug: var=riak_cluster_status
 
-- set_fact: riak_control_node_name="{{ hostvars[riakcs_control]['riak_nodename'] }}"
+- set_fact: riak_control_node_name="{{ riak_nodename }}"
```
millerdev: I'm pretty sure this is wrong. On all hosts `riak_control_node_name` should be the `riak_nodename` of the control node.

dannyroberts: Oh! @esoergel that makes a lot of sense after doing our research. Somehow this ended up working? Maybe we deployed just that one time off of master instead of this branch? @millerdev I think at some point, running this wasn't working and that's why Ethan and I looked into it and ended up making this change.

esoergel: Yeah, that definitely sounds right.  I'll revert this change.  Still curious what made this line fail that one time then.

ansible/group_vars/all.yml
esoergel: There's a lot of state floating around, so I imagine reproduction will be difficult.  I was following [these steps](https://confluence.dimagi.com/display/commcarehq/Adding+and+removing+webworkers) on new staging machines.  The command that failed was step 8 (deploy the full stack):
```bash
$ aps --limit=$NEW_MACHINES --diff --check
```

millerdev: > The command that failed was step 8 (deploy the full stack)...

Did it only fail with `--check` (that isn't too surprising, but I should try to fix it), or did it fail without `--check` as well?

esoergel: It failed without `--check`.

millerdev: I think I need more details to troubleshoot this further. The way it's designed the Riak CS machines should be configured and keys should be set by the time these `s3_blob_db_...` variables are used to write localsettings.py. I would need to know where in the process the failure occurred and what had happened before that to know why it failed.

esoergel: Yeah, I think I'll just revert this change and see if it comes up again when I start migrating to virginia

dannyroberts:
(not mergeable)

esoergel:
:white_check_mark:

PR #367, opened 2016-03-29
postgresql_ssl_enabled defaults to False

nickpell:
based on discussion with @benrudolph

### DIFF
--- a/ansible/roles/postgresql/defaults/main.yml
+++ b/ansible/roles/postgresql/defaults/main.yml
@@ -1,7 +1,7 @@
 pg_install: True
 postgresql_version: '9.4'
 postgresql_port: 5432
-postgresql_ssl_enabled: True
+postgresql_ssl_enabled: False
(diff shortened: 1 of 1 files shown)


benrudolph:
:+1: maybe @dannyroberts to confirm?

dannyroberts:
I'm not super familiar with what changing `postgresql_ssl_enabled` does on a production environment. Can you describe what'll happen when we deploy this to prod?

benrudolph:
postgres will stop using ssl for connections to it over the network if this is set to false. but i believe that's already the case on prod. https://dimagi.slack.com/archives/web-team/p1459324647000130

dannyroberts:
Cool, thanks. Let's make sure to run this with `--check` on each instance we have to make sure we're either not changing things or changing things correctly. (I imagine there's a change that needs to happen on both postgres clients like web workers, etc. and in postgres's config itself.)

Good to merge from my side once you've removed the hold off on merge tag.

PR #368, opened 2016-04-01
Install ranger and kafkat

esoergel:
ranger everywhere and kafkat on kafka machines
@benrudolph

### DIFF
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -56,6 +56,7 @@
     - mosh
     - libjpeg-dev
     - gradle
+    - ranger  # cmdline file browser
 
 - name: Set NPM registry
   command: 'npm config set registry http://registry.npmjs.org/'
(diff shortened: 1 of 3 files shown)


benrudolph:
:beetle: :cat:

benrudolph:
:+1: on :white_check_mark:

czue:
what do these do?

benrudolph:
@esoergel can probably give you more of an explanation, but from my discussion with him, `kafkat` is a command line utility for working with kafka since kafka's mgmt via the command line is very difficult

esoergel:
Yup, kafkat is a cmdline kafka tool.  I found it very difficult to see what's going on with a kafka cluster (see my [migration notes](https://confluence.dimagi.com/display/internal/Migrating+a+Kafka+cluster) for examples), and kafkat looks much more manageable.

Ranger is a command line file browser.  I could totally be talked out of it if we're concerned about bloat, but I find it very useful.  Basically it lets you move through files and directories much more easily than `cd` and `ls`.

czue:
:+1: doesn't bother me, was just wondering in case they were useful (sounds like they are)

PR #373, opened 2016-04-06
Add libfreetype6-dev to common package installs

millerdev:
This is so django-simple-captcha can write text in images

@benrudolph cc @kaapstorm


benrudolph:
lgtm, no tests

PR #377, opened 2016-04-11
Use redis host property

wpride:
goes with https://github.com/dimagi/formplayer/pull/26


ansible/roles/touchforms/templates/application.properties.j2
```diff
@@ -1,2 +1,3 @@
 commcarehq.host=https://{{ SITE_HOST }}
 server.port=8181
+redis.host={{ groups.redis.0 }}
```
snopoke: This should be `redis.hostname` to match `@Value("${redis.hostname}")`

snopoke: Or visa versa

PR #380, opened 2016-04-14
Icds db

sravfeyn:


ansible/roles/postgresql/tasks/main.yml
```diff
@@ -84,11 +84,12 @@
 - name: Create PostgreSQL users
   sudo_user: postgres
   postgresql_user:
-    name: "{{ localsettings.PG_DATABASE_USER }}"
-    password: "{{ localsettings.PG_DATABASE_PASSWORD }}"
-    role_attr_flags: CREATEDB
+    name: "{{ item.username }}"
+    password: "{{ item.password }}"
+    role_attr_flags: "{{ item.role_attr_flags }}"
```
snopoke: maybe specify a default value: "{{ item.role_attr_flags|default('LOGIN') }}"

ansible/deploy_stack.retry
```diff
@@ -0,0 +1 @@
+192.168.33.16
```
snopoke: shouldn't commit this file

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -103,6 +104,15 @@
   when:
```
snopoke: Should also add owner to the `postgresql_dbs` config list

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -98,11 +99,31 @@
...
+- name: Grant PostgreSQL privileges
+  sudo_user: postgres
+  postgresql_privs: type=database database="{{ item.1.db_name }}" roles="{{ item.0.username}}" privs="{{ item.1.priv}}"
+  with_subelements:
+    - "{{ postgres_users.values() }}"
+    - privileges
+    - flags:
+      skip_missing: True
+
+- name: Set readonly permissions
```
sravfeyn: This is granting more permissions than readonly

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -98,11 +99,31 @@
...
+  sudo_user: postgres
+  postgresql_privs: db="{{ item.1.db_name }}" roles="{{ item.0.username}}" privs=SELECT objs=ALL_IN_SCHEMA
```
snopoke: `db` should be `database`

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -98,11 +99,31 @@
...
     pg_install == True
 
+- name: Grant PostgreSQL privileges
+  sudo_user: postgres
+  postgresql_privs: type=database database="{{ item.1.db_name }}" roles="{{ item.0.username}}" privs="{{ item.1.priv}}"
+  with_subelements:
+    - "{{ postgres_users.values() }}"
+    - privileges
+    - flags:
+      skip_missing: True
```
snopoke: Task needs a when clause `item.readonly == False`

ansible/roles/postgresql/tasks/main.yml
snopoke: I suspect we need to do something like this: https://gist.github.com/jirutka/afa3ce62b1430abf7572#file-pg_grant_read_to_db-sh

one issue though is that setting the default privileges needs to be done by the `commcarehq` user so I guess using the `shell` or `command` module is necessary

here is an eg where we do something similar: https://github.com/dimagi/commcarehq-ansible/blob/757d383105114e0c900c193a578664174b69bd5b/ansible/roles/postgresql/tasks/main.yml#L173-L173
 
GitHub
dimagi/commcarehq-ansible
Contribute to commcarehq-ansible development by creating an account on GitHub.

ansible/roles/postgresql/tasks/main.yml
sravfeyn: Will change it, but it's an alias anyway.

ansible/roles/postgresql/tasks/main.yml
sravfeyn: We want to grant `CONNECT` privilege to readonly users as well, so this should be here, no?

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -92,17 +92,52 @@
   when:
     pg_install == True
 
+- name: Create PostgreSQL ICDS users
```
snopoke: why did you split this up again? seemed fine to create all the users with this task

snopoke: also if the ICDS users are in the `posetgres_users` list then they'll get added to pgbouncer userlist

sravfeyn:
This is pretty much ready to go. I will remove the hold-off tag once I get to rename some existing db objects manually tomorrow.

sravfeyn:
@snopoke I renamed existing db objects on india manually. This can now be merged.

sravfeyn:
Actually this task requires `ansible 2.0.1.0`. We need to upgrade ansible before merging this in

snopoke:
TODO:
- [ ] notify team of ansible upgrade verify compatability
- [ ] once merged update all envs config (swiss, prod, zambia, india)

benrudolph:
fwiw, i had attempted to upgrade to ansible 2.0 and most of the stuff worked fine and the error messages were much more clear. i didn't go through with it because i couldn't tell if i was conflating errors from the ansible upgrade or from a broken ansible build. when i upgraded to 1.9 i ran through the stack on vagrant to make sure everything still worked

snopoke:
Tried running the full stack deploy but got stuck with riakcs:
```
PLAY [Install Riak and Riak CS] ************************************************

TASK [setup] *******************************************************************
ok: [192.168.33.16]
ok: [192.168.33.17]

TASK [riakcs/install : Check if riak is installed] *****************************
ok: [192.168.33.16]
ok: [192.168.33.17]

TASK [riakcs/install : set_fact] ***********************************************
ok: [192.168.33.16]
ok: [192.168.33.17]

TASK [riakcs/install : Fetch riak apt repo install script] *********************
fatal: [192.168.33.16]: FAILED! => {"failed": true, "msg": "'defaults' is undefined"}
fatal: [192.168.33.17]: FAILED! => {"failed": true, "msg": "'defaults' is undefined"}
```
@millerdev how is the `defaults` dict meant to work (https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/riakcs/install/tasks/main.yml#L16)?

snopoke:
Made one or 2 unrelated commits

sravfeyn:
I spent around 2 hours to prepare a doc of what code changes need to be made for ansible upgrade. I found good resources, but I will need more time to be able to list out the required changes, as the ansible context I have is not enough right now (things like dynamic includes, finding them in our playbooks etc). Is it worth me putting more time into this?

snopoke:
Just share what you have do far.  Don't think it's worth more effort at
this point.

millerdev:
@snopoke wrote:
> how is the defaults dict meant to work (https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/riakcs/install/tasks/main.yml#L16)?

It's a reference to these vars https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/riakcs/install/defaults/main.yml Not sure how it needs to be changed for Ansible 2.0.

sravfeyn:
I found a porting guide for ansible 2.0 https://github.com/ansible/ansible/blob/devel/docsite/rst/porting_guide_2.0.rst This lists all changes that needs to be done, but some of them needs good context of our ansible task and I think ~2hours is not enough to do this. I made this small changes to fix some of deprecation warnings https://github.com/dimagi/commcarehq-ansible/pull/390/files

benrudolph:
yikes, that definitely looks like more work then anticipated

snopoke:
Is it possible to get this working pin 1.9 so that we can separate it from
the upgrade?
On 27 Apr 2016 19:10, "Ben" <notifications@github.com> wrote:

yikes, that definitely looks like more work then anticipated

—
You are receiving this because you were mentioned.

Reply to this email directly or view it on GitHub
<https://github.com/dimagi/commcarehq-ansible/pull/380#issuecomment-215151203>

sravfeyn:
I can try (by probably restructuring the variables), it won't be as neat. But this is not urgent at all (as in nothing is blocking ICDS reports). I am happy to wait for upgrade which gives bunch of useful new features. Your call.

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -107,6 +108,30 @@
...
+- name: Set readonly PostgreSQL roles
+  sudo_user: postgres
+  shell: "PGPASSWORD={{ postgres_users.get(item.1.readonly_master).get('password') }}
+    {{postgres_install_dir}}/bin/psql -U {{ item.1.readonly_master }} -h localhost {{item.1.db_name}} -c
+    'GRANT USAGE ON SCHEMA PUBLIC to {{ item.0.username }}; 
+    GRANT SELECT ON ALL TABLES IN SCHEMA PUBLIC to {{ item.0.username }}; 
+    ALTER DEFAULT PRIVILEGES IN SCHEMA PUBLIC GRANT SELECT ON TABLES TO {{ item.0.username }}'"
+  when:
+    item.1.readonly == True
+  with_subelements:
```
snopoke: @benrudolph I looked at this again and the issue is that `with_subelements` isn't well supported in ansible < 2.0.

benrudolph: aha thanks for looking into that

sravfeyn:
I was trying to do a merge just before ending my day. I shouldn't have. I will try to fix this tomorrow.

dannyroberts:

sravfeyn:
This got done in a different (better) way. So it's not needed anymroe.

PR #384, opened 2016-04-20
tailon for captain

benrudolph:
Allows us to tail log files for captain @millerdev


benrudolph:
might actually switch to this: https://github.com/kilianc/rtail

PR #386, opened 2016-04-25
Move libffi to common installs

benrudolph:
@millerdev this is required now that github3py will be used in our prod environments

### DIFF
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -58,6 +58,7 @@
     - libfreetype6-dev
     - gradle
     - ranger  # cmdline file browser
+    - libffi-dev  # Required for installing github3py
 
 - name: Set NPM registry
   command: 'npm config set registry http://registry.npmjs.org/'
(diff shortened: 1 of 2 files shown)


benrudolph:
(has been run on all envs)

ansible/roles/common_installs/tasks/main.yml
```diff
@@ -58,6 +58,7 @@
     - libfreetype6-dev
     - gradle
     - ranger  # cmdline file browser
+    - libffi-dev  # Required for installing github3py
```
millerdev: This task is installing lots of stuff that is not needed on many machines. For example, `libffi-dev` (and many others too) is not needed by Riak CS machines. Not a blocker for this PR, but would be nice if we moved a lot of this stuff to the `commcarehq` role.

benrudolph: true agreed. this package is pretty small ~300kb so it's not a huge deal, but is nice for managing complexity

millerdev: Yeah, size is not remotely the issue. Putting stuff on servers that is not needed is a bad practice for other reasons: it makes maintenance more difficult, can result in implicit/undocumented dependencies over time, and in some cases can introduce security vulnerabilities.

PR #387, opened 2016-04-26
Update main.yml

midav:
Actual add 'user,noauto,rw' to /etc/mtab


benrudolph:
thanks @midav! looks good

PR #389, opened 2016-04-27
pip to upgrade supervisor

benrudolph:
@millerdev @snopoke going to try and upgrade supervisord and see if it helps with deploy errors revolving around weirdness in supervisor. our version is 3 years old (3.0b2). seems like they've made a fair number of bug fixes since: http://supervisord.org/changes.html. will go through breaking changes and do staging upgrade first

### DIFF
--- a/ansible/roles/supervisor/tasks/main.yml
+++ b/ansible/roles/supervisor/tasks/main.yml
@@ -11,18 +11,9 @@
   changed_when: false
   failed_when: which_supervisord.rc != 0 and which_supervisord.rc != 1
 
-- name: Add git apt repo
-  apt_repository: repo='ppa:schooltool-owners/trunk' state=present
-  register: add_supervisor_ppa
-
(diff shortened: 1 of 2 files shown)


benrudolph:
the change log also mentions performance improvements to starting and stopping lots of processes (aka pillowtop!)

millerdev:
should we test this out for a while (few days) on staging first before merging and deploying to production?

benrudolph:
Yep definitely

benrudolph:
Brain dumping so I remember where I was:
1. Install new version of supervisor with `pip`. Gets stored in `/usr/local/bin`
2. Remove apt version of supervisor. stored in `/usr/bin`
3. Redeploy upstart conf with proper supervisor path
4. Shut down all supervisor processes. Shutdown supervisord
5. Start supervisord back up with new process
6. Run `--tags=users` to update sudoers files

PR #391, opened 2016-04-28
add nginx 500 page

sravfeyn:
Requires https://github.com/dimagi/commcare-hq-errorpages/pull/7

This shows ngnix 500 page if Django is down @snopoke


snopoke:
I don't think this is a good idea. Django will still return a 500 response even if it handles the error.

PR #392, opened 2016-04-28
add lines to hosts

snopoke:
@kaapstorm  @sravfeyn

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -1,0 +1,5 @@
+- name: "Build hosts file"
+  lineinfile: dest=/etc/hosts regexp='.*{{ item }}$' line="{{item}}" state=present
+  with_items: etc_host_lines
+  tags:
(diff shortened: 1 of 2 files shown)


ansible/roles/common/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+- name: "Build hosts file"
+  lineinfile: dest=/etc/hosts regexp='.*{{ item }}$' line="{{item}}" state=present
```
kaapstorm: Is the purpose of that `regexp` to uncomment `{{ item }}` if it exists and is commented out?

kaapstorm: Unclear from [the docs] whether `line` will be inserted if `regexp` isn't found. 

  [1]: http://docs.ansible.com/ansible/lineinfile_module.html

kaapstorm:
Considering we've used this, we know it works, despite my question. Merging.

PR #393, opened 2016-04-29
j2me http

nickpell:
This enables using J2ME by hitting http endpoints, and will be necessary once we drop support for SHA1.

For india, we will jump immediately to this solution while production will continue to use https until we have to switch.  Prod will be setup to redirect http to https (in a separate PR) once CommCare on J2ME supports redirects.

@emord @ctsims


nickpell:
also hoping to get this in today when http://manage.dimagi.com/default.asp?225900 is fixed
@biyeun

nickpell:
good to merge

@emord cc: @biyeun

PR #396, opened 2016-05-01
Captain Softlayer vpn

benrudolph:
@snopoke this allows captain to access softlayer vpn

cc: @millerdev

### DIFF
--- a/ansible/roles/captain/vars/main.yml
+++ b/ansible/roles/captain/vars/main.yml
@@ -1,5 +1,9 @@
 deploy_env: captain
 # supervisor_task_conf: "{{ service_home }}/supervisor"
+#
+# Defined in config directory:
+# softlayer_vpn_username: ...
+# softlayer_vpn_password: ...
 
 captain_dir: "{{ cchq_home }}/captain"
 captain_log_dir: "{{ captain_dir }}/log"
--- a/ansible/roles/captain/tasks/main.yml
+++ b/ansible/roles/captain/tasks/main.yml
@@ -82,3 +82,6 @@
 
 - name: Igor Integration
   include: igor.yml
+
+- name: Configure Softlayer VPN
+  include: softlayer.yml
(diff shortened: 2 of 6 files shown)


snopoke:
:+1: but not mergable

ansible/roles/captain/vars/main.yml
```diff
@@ -1,5 +1,9 @@
 deploy_env: captain
 # supervisor_task_conf: "{{ service_home }}/supervisor"
+#
+# Defined in config directory:
+# softlayer_vpn_username: ...
+# softlayer_vpn_password: ...
```
millerdev: @benrudolph: I'm assuming you copied these credentials from keepass into the config repo. I think this was one of the reasons we did not implement deploy from remote host in the past. I had tried to get it setup, but we decided against it at the time because it required putting login credentials on a file system on a remote server somewhere, which we were uncomfortable with at the time. Did you discuss this with other members of the team, or are you going with it's easier to ask forgiveness than permission?

benrudolph: I think I went with the forgiveness route, but I don't think what I did was unreasonable. One I made a new user that doesn't have as many permissions as the current user in the KeyPass (though still enough that if someone had access they could obviously just delete everything) and two, if we don't believe our servers in our rackspace vpn are secure (given our no root ssh and only public key access), then we shouldn't have our cloudant password or any of our passwords stored on machines. You can make the same argument for storing our cloudant password on a remote server in a localsettings file. I guess both are still non-ideal, but if one is insecure then I'd argue the other is also insecure

czue: We should follow up offline on this. Practical security aside we should also understand the legal constraints (e.g. HIPAA) and make sure it's ok from that perspective.

millerdev: Can I be included in this (offline) discussion when it happens? Also, not sure if this has already been deployed. If it has should the credentials be removed until we have made a decision?

benrudolph: Sure just got backed so happy to talk about it this week or next. This has been deployed I can remove the creds until then. but just so i understand correctly your concern is having the credentials on hqcaptain not on salt machine right?

benrudolph: (also removed all passwords)

PR #397, opened 2016-05-01
Captain igor

benrudolph:
@snopoke @millerdev this moves igor off of the jenkins machine and on to hqcaptain where it can better be maintained


benrudolph:
hmm igor should probably be its own role and have its own entry in the inventory file

PR #399, opened 2016-05-02
Add cas site

emord:
Still need to buy a cert
https://trello.com/c/r03fYqIH/105-override-the-base-url-at-an-app-level

This adds cas.commcarehq.org for the india server. It's going to be used until the gov servers are set up. It'll only be used by phones, so not copying over the /static, /ws etc paths
cc: @NoahCarnahan


NoahCarnahan:
Looks fine to me, but don't know very much about nginx, so might be worth pinging someone else if you're looking for a more thorough review (or I can do some research :P).

emord:
Successfully deployed this to india. (this is just copied over from the cchq_ssl.yml file but without the static files)

new code buddy @gcapalbo

gcapalbo:
@emord looks ok to me.  is there anyone else with more context that you'd like to review this, or are you saying this is already live?

PR #400, opened 2016-05-03
Comment on why the hostvars[riak_control] dict may be empty

esoergel:
@millerdev @danielroberts
Danny just had this same issue coming up.  I think it's triggered by limiting the machines that you deploy - I think the variables aren't set on the control node if it's not processed in the ansible command

### DIFF
--- a/ansible/roles/riakcs/join_cluster.yml
+++ b/ansible/roles/riakcs/join_cluster.yml
@@ -15,6 +15,9 @@
 
 #- debug: var=riak_cluster_status
 
+# If this line fails, it might be because you're not also deploying to the
+# control node This means it doesn't get processed and this stuff is never set.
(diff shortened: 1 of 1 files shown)


ansible/roles/riakcs/join_cluster.yml
```diff
@@ -15,6 +15,9 @@
 
 #- debug: var=riak_cluster_status
 
+# If this line fails, it might be because you're not also deploying to the
+# control node This means it doesn't get processed and this stuff is never set.
+# Try including the first entry in [riakcs] in your `--limit=` declaration.
```
millerdev: ah, that makes a lot of sense

esoergel:
Whoops, totally pinged the wrong @dannyroberts ...

dannyroberts:
👍 thanks

PR #401, opened 2016-05-05
allow downloading form XML over http

nickpell:
@emord @ctsims cc: @biyeun


nickpell:
@emord this is good to go

PR #402, opened 2016-05-05
proxy redirect urls hit by j2me from http to https

nickpell:
For the interim period before sha1 is switched to sha256.  Blocking on j2me support for 302 redirects.


nickpell:
looks like we're not going this route

PR #403, opened 2016-05-10
set couch-stale setting to update_after

sravfeyn:
@proteusvacuum 
FYI @snopoke

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -460,7 +460,8 @@ INDICATOR_NAMESPACES = {
         ('mvp_indicators', "MVP"),
     ),
 }
-COUCH_STALE_QUERY = 'ok'
+# should be 'ok' if couchdb version < 1.1.0, else 'update_after'
+COUCH_STALE_QUERY = 'update_after'
 
 API_THROTTLE_REQUESTS = 5
 API_THROTTLE_TIMEFRAME = 10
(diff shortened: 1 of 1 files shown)


snopoke:
:+1: just need to make sure this is overridden in the env vars files where it needs to be (prod and softlayers)

ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -460,7 +460,8 @@ INDICATOR_NAMESPACES = {
         ('mvp_indicators', "MVP"),
     ),
 }
-COUCH_STALE_QUERY = 'ok'
+# should be 'ok' if couchdb version < 1.1.0, else 'update_after'
```
czue: nit, but it should also be 'ok' on any cloudant environment

snopoke: Maybe also say that it should be 'ok' for cloudant.

ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -460,7 +460,8 @@ INDICATOR_NAMESPACES = {
...
+COUCH_STALE_QUERY = 'update_after'
```
czue: like @snopoke said - we don't want to change this for prod/india so should be overrideable and the default should probably remain 'ok'

czue: like @snopoke said - we don't want to change this for prod/india so should be overrideable and the default should probably remain 'ok'

sravfeyn:
Makes sense, closing this

sravfeyn:
Updated

sravfeyn:
Bump. This is urgent, exports are dysfunctional on swiss without this

PR #404, opened 2016-05-10
don't expose ES to public

sravfeyn:
@proteusvacuum 
FYI @snopoke 
http://manage.dimagi.com/default.asp?225659


ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -9,6 +9,9 @@ node.name: "{{ elasticsearch_node_name }}"
 
 transport.tcp.port: {{ elasticsearch_tcp_port }}
 
+# don't expose ES to public
+network.bind_host: localhost
```
czue: I don't think this will work on any non-monolith environment since it needs to be open on the private interface. guessing this also needs to be overrideable

sravfeyn:
@czue update

ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -9,6 +9,11 @@ node.name: "{{ elasticsearch_node_name }}"
 
 transport.tcp.port: {{ elasticsearch_tcp_port }}
 
+{% if elasticsearch_bind_to_localhost %}
```
benrudolph: i think a better way to do this check is to check the length of `group['all']` and if it's greater than 1 than it's not a monolith. that way the person ansible deploying doesn't have to know about this variable to do a correct deploy

sravfeyn: Good idea, updated `elasticsearch_bind_to_localhost`.

benrudolph: nice thanks!

PR #407, opened 2016-05-13
couchdb stale = 'ok' and j2me

proteusvacuum:
Needed quotes. Also, added a check that the j2me setting existed before setting it.


proteusvacuum:
@benrudolph

benrudolph:
what a battle!

proteusvacuum:
All your ansible knowledge is belong to us.

PR #408, opened 2016-05-16
task to upgrade openssl

sravfeyn:
https://docs.google.com/document/d/1Urbjz78Q6NnVEmLDFrcpv7nQtqx4Xvi07ZMQ5FREs90/edit
@snopoke


ansible/deploy_touchforms.retry
```diff
@@ -0,0 +1 @@
+192.168.33.16
```
snopoke: These shouldn't be committed - could add to `.gitignore`

sravfeyn:
Updated

PR #410, opened 2016-05-17
Monolith ufw

benrudolph:
@proteusvacuum @nickpell this only allows certain traffic for monoliths, so no one can see our entire database...

http://manage.dimagi.com/default.asp?225659

cc: @millerdev

### DIFF
--- a/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
+++ b/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
@@ -9,11 +9,6 @@ node.name: "{{ elasticsearch_node_name }}"
 
 transport.tcp.port: {{ elasticsearch_tcp_port }}
 
-{% if elasticsearch_bind_to_localhost %}
-# don't expose ES to public
-network.bind_host: localhost
-{% endif %}
-
 http.port: {{ elasticsearch_http_port }}
 
 http.cors.enabled: true
--- a/ansible/roles/ufw/tasks/proxy_ufw.yml
+++ b/ansible/roles/ufw/tasks/proxy_ufw.yml
@@ -22,7 +22,7 @@
   ufw:
     rule: allow
     direction: in
-    port: 60003
+    port: "60000:61000"
     proto: udp
   when: is_monolith
 
--- a/ansible/roles/ufw/tasks/proxy_ufw.yml
+++ b/ansible/roles/ufw/tasks/proxy_ufw.yml
@@ -1,0 +1,35 @@
+# Rules for monoliths
+- name: allow web traffic
+  ufw:
+    rule: allow
+    direction: in
+    port: "{{ item }}"
+    proto: tcp
+  with_items:
+    - 443  # https
+    - 80   # http
+  when: is_monolith
+
+- name: allow ssh traffic
+  ufw:
+    rule: allow
+    direction: in
+    port: 22
+    proto: any
+  when: is_monolith
+
+- name: allow mosh traffic
+  ufw:
+    rule: allow
+    direction: in
+    port: 60003
+    proto: udp
+  when: is_monolith
+
+- name: default to denying all incoming traffic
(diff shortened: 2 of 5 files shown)


ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -9,11 +9,6 @@ node.name: "{{ elasticsearch_node_name }}"
 
 transport.tcp.port: {{ elasticsearch_tcp_port }}
 
-{% if elasticsearch_bind_to_localhost %}
-# don't expose ES to public
-network.bind_host: localhost
```
benrudolph: this was actually just wrong. `host` is set to an IP address not localhost and thus it made it impossible to even access elastic search. the reason it didn't surface sooner is that no one restarted ES when the change was made

benrudolph: by default `bind_host` == `host`

ansible/roles/ufw/tasks/proxy_ufw.yml
```diff
@@ -0,0 +1,35 @@
...
+    direction: in
+    port: 22
+    proto: any
+  when: is_monolith
+
+- name: allow mosh traffic
+  ufw:
+    rule: allow
+    direction: in
+    port: 60003
```
proteusvacuum: I think mosh takes a range of ports: `60000:61000`

ansible/roles/ufw/tasks/proxy_ufw.yml
```diff
@@ -0,0 +1,35 @@
...
+    rule: allow
+    direction: in
+    port: "{{ item }}"
+    proto: tcp
+  with_items:
+    - 443  # https
+    - 80   # http
+  when: is_monolith
+
+- name: allow ssh traffic
```
proteusvacuum: do these allow ssh and mosh when not connected to the vpn?

ansible/roles/ufw/tasks/proxy_ufw.yml
benrudolph: ah good to know, thanks

ansible/roles/ufw/tasks/proxy_ufw.yml
benrudolph: yes, no sure what to do about that though since neither swiss or zambia are behind a vpn

benrudolph: we need to disable password ssh...

proteusvacuum: Ah. Do we need to add a second factor on monoliths then? Like google-auth or something?
(probably for discussion later, but just flagging now)

benrudolph: good question, i don't think we've thought about it too hard. agree that it's a parallel discussion, this is still better than every port being open 😄

proteusvacuum:
:+1: on :heavy_check_mark:

ansible/roles/ufw/tasks/proxy_ufw.yml
millerdev: are there not multiple interfaces, one for the VPN and one public? can we not have ufw be more permissive on the VPN interface than on the public one?

ansible/roles/ufw/tasks/proxy_ufw.yml
```diff
@@ -0,0 +1,35 @@
...
+    direction: in
+    port: 22
+    proto: any
+  when: is_monolith
+
+- name: allow mosh traffic
+  ufw:
+    rule: allow
+    direction: in
+    port: "60000:61000"
```
millerdev: This is a big port range to have open to the public. See comment above about public vs VPN interfaces. That seems ideal. (However, I also agree that this is better than what we had before.)

PR #411, opened 2016-05-20
500MB limit for client_max_body_size

proteusvacuum:
http://manage.dimagi.com/default.asp?226167#1148928

This was failing because the zip is almost 400MB, so nginx barfs. Manually edited on softlayer, but should probably be run on all other environments. 
@benrudolph @nickpell interrupt buddies

cc @dannyroberts code buddy


benrudolph:
yikes. that's a lot of mulitmedia!

benrudolph:
📷 📹 🎤

PR #413, opened 2016-05-26
Setup proxy for a single MOTECH endpoint

sheelio:
Moving to a model where motech servers are at motech.commcarehq.org/INSTANCENAME

This configuration sets this up for the servers required for the UN Supply project.  (It also removes the configuration for the unused endos-motech.commcarehq.org machine).  

@czue @calellowitz @emord 
Can one of you review, merge and deploy?


emord:
@sheelio Will ping you separately once deployed

PR #416, opened 2016-06-01
Fix nginx config for MOTECH

sheelio:
@emord 

Figured out the problem with my Nginx config.  This should fix it.


sheelio:
@emord 

Added one more change that I'm pretty sure fixes the issue (at least based on my local testing)

PR #417, opened 2016-06-02
seperate couchdb-install as it requires sudo privilege

sravfeyn:
@czue

### DIFF
--- a/ansible/roles/couchdb/tasks/main.yml
+++ b/ansible/roles/couchdb/tasks/main.yml
@@ -24,7 +24,7 @@
     dest: ~/{{ couchdb_distro_filename }}
   when: not couchdb_path.stat.exists
 
-- name: Compile + install CouchDB
+- name: Compile CouchDB
   shell: "{{ item }}"
   args:
     chdir: ~/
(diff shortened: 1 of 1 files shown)


ansible/roles/couchdb/tasks/main.yml
```diff
@@ -33,9 +33,18 @@
...
-    - sudo make install -C ~/apache-couchdb-{{ couchdb_version }}
-    - rm -rf ~/apache-couchdb-{{ couchdb_version }}*
   when: not couchdb_path.stat.exists
+  register: couchdb_compile
+
+- name: Install CouchDB
+  shell: "{{ item }}"
+  args:
+    chdir: ~/
+  sudo: yes
```
sravfeyn: Compiling is done without `sudo` so as to avoid permission problems see [commit message](https://github.com/dimagi/commcarehq-ansible/commit/0c2951a3f7ddfee0c854a907dcd18ef39b018a3f). But installation requires sudo as shown in the command here. If sudo is used via raw commands, ansible hangs at the password prompts. sudo commands should be done via `sudo: yes`. So this PR seperates out the part that needs sudo into a separate sudo task

snopoke:
:+1:

PR #421, opened 2016-06-02
Add message about restarting node

emord:
@czue I think this is what happened with es2 earlier. Thoughts on adding a message to warn the restarter they might need to go and do things manually?

cc @proteusvacuum

### DIFF
--- a/ansible/roles/elasticsearch/tasks/rolling_restart.yml
+++ b/ansible/roles/elasticsearch/tasks/rolling_restart.yml
@@ -18,6 +18,8 @@
   sudo: true
   service: name=elasticsearch state=started
 
+- debug: msg="Sometimes we try to start the node too soon. If hung start node manually"
+
 - name: wait for node to restart
   shell: curl -I -s -m 2 http://localhost:9200 | head -n 1
   register: result


czue:
I think adding a message about manual intervention is a good idea

ansible/roles/elasticsearch/tasks/rolling_restart.yml
```diff
@@ -18,6 +18,8 @@
   sudo: true
   service: name=elasticsearch state=started
 
+- debug: msg="Sometimes we try to start the node too soon. If hung start node manually"
```
czue: what does this do? (not familiar with it)

emord: the output is 

```
TASK [debug] *******************************************************************
ok: [192.168.33.16] => {
    "msg": "Sometimes we try to start the node too soon. If hung start node manually"
}
```

seemed like the easiest way to add a message, but i didn't look super hard either

czue: neat

PR #423, opened 2016-06-03
add /etc/sudoers.d/ansible file

dannyroberts:
to auto-manage which users can `sudo -iu ansible bash` without a password


dannyroberts:
Goes well with https://github.com/dimagi/ansible-dotfiles/pull/5

PR #424, opened 2016-06-03
Disable ssh password

emord:
buddy @proteusvacuum


czue:
nice. :closed_lock_with_key:

PR #429, opened 2016-06-09
There's a hard failure if this isn't set

esoergel:
This line depends on this variable:
https://github.com/dimagi/commcarehq-ansible/blob/b3002b43dd42e986380125acf2f85956250e19a3/ansible/roles/datadog/tasks/integrations.yml#L7
@snopoke @millerdev


millerdev:
👍  will let @snopoke merge since he may have more context

PR #432, opened 2016-06-13
Update config file name

gcapalbo:
@benrudolph: @wpride was proposing to rename this file from <env>.properties to application.properties. Any reason it was named <env>.properties originally? Seems like it always is copied in a specific environment's code root anyway.


gcapalbo:
added hold off on merge label so that @benrudolph can sign off

benrudolph:
👍 signed off

PR #435, opened 2016-06-15
remove apache role

snopoke:
@nickpell


benrudolph:
i was also considering doing this, but i don't think zambia is off of apache yet. we may want to keep it around until we do that switch? don't have that strong of opinions on it though

snopoke:
wow you're right, crazy. Fine to leave it for now.

czue:
wow, good catch. sorry about the bad advice

benrudolph:
seems like this is OK to merge now

czue:
nice

PR #440, opened 2016-06-17
Include sudo for running individual roles

esoergel:
@benrudolph @dannyroberts @snopoke
Are there any instances where we don't want this?


snopoke:
Not that I know of

dannyroberts:
Yeah ntiko

PR #442, opened 2016-06-21
Add ability to skip checking riak control machine

esoergel:
@millerdev @benrudolph


esoergel:
No can do on the `{% if %}` syntax - I think it first gets parsed as yml, then the strings get parsed as j2

benrudolph:
that's a bummer, well 👍 on ✔️

millerdev:
👍 on ✅

PR #443, opened 2016-06-23
Ce/pg backup

calellowitz:
@czue @benrudolph cc: @dannyroberts 
This fixes a bunch of small errors that were preventing our current postgres backup script from running. I tested running the script as the postgres user on staging. I have not yet tried to restore from the dump.

### DIFF
--- a/ansible/roles/backups/tasks/main.yml
+++ b/ansible/roles/backups/tasks/main.yml
@@ -5,7 +5,7 @@
     dest: "/etc/cron.d/create_blobdb_backup.sh"
     group: root
     owner: root
-    mode: 700
+    mode: 0700
     backup: yes
   when: backup_blobdb
 
--- a/ansible/roles/postgresql/templates/create_postgres_dump.sh.j2
+++ b/ansible/roles/postgresql/templates/create_postgres_dump.sh.j2
@@ -3,7 +3,7 @@ BACKUP_TYPE=$1
 DAYS_TO_RETAIN_BACKUPS=$2
 
 TODAY=$(date +"%Y_%m_%d")
-pg_dumpall --clean | gzip  > "{{ postgresql_backup_dir }}/postgres_$BACKUP_TYPE_$TODAY.db.gz"
+pg_dumpall --clean | gzip  > "{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}.db.gz"
 
 # Remove old backups of this backup type
-find {{ postgresql_backup_dir }} -mtime "+$DAYS_TO_RETAIN_BACKUPS" -name "postgres_$BACKUP_TYPE_*" -exec rm {} \;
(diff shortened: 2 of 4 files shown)


ansible/roles/backups/tasks/main.yml
```diff
@@ -5,7 +5,7 @@
     dest: "/etc/cron.d/create_blobdb_backup.sh"
     group: root
     owner: root
-    mode: 700
+    mode: 0700
```
calellowitz: ansible file modes need to have a leading zero, so i went back and updated existing files that lacked them.

benrudolph: good find

ansible/roles/postgresql/templates/create_postgres_dump.sh.j2
```diff
@@ -3,7 +3,7 @@ BACKUP_TYPE=$1
 DAYS_TO_RETAIN_BACKUPS=$2
 
 TODAY=$(date +"%Y_%m_%d")
-pg_dumpall --clean | gzip  > "{{ postgresql_backup_dir }}/postgres_$BACKUP_TYPE_$TODAY.db.gz"
+pg_dumpall --clean | gzip  > "{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}.db.gz"
```
benrudolph: just curious, was `$BACKUP_TYPE` not working? i had thought that would work with double quotes

benrudolph:
👍 on ✔️

ansible/roles/postgresql/templates/create_postgres_dump.sh.j2
calellowitz: It don't think it could figure out where `$BACKUP_TYPE` ended so it was leaving it out. probably looking for a variable called `$BACKUP`. I did a bit of playing around and variables with underscores seem to only work correctly with the braces or at the end of a string

benrudolph: ah makes sense, thanks

gcapalbo:
just a few questions i had while reviewing this:
- does the backup get saved on an encrypted drive?
- what user do we run the command as? seems like https://www.postgresql.org/docs/9.4/static/app-pg-dumpall.html mentions needing to run it as a database superuser
- have we tested this out on staging?

calellowitz:
@gcapalbo 
To answer your questions
The backup is saved on the encrypted drive https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/postgresql/defaults/main.yml#L6
The command is run by the user `postgres` so we should be fine there.
I tested out deploying the playbook and running the command on staging, as `postgres`, and it successfully completed but I have not yet tried to restore from that dump file. I also ran it manually so we will have to wait until tomorrow to see if it will run successfully from `cron`.

gcapalbo:
:+1: thanks @calellowitz

PR #444, opened 2016-06-23
kafka monitoring

benrudolph:
@snopoke initial setup with kafka monitoring. running into a couple issues that i'm still working out:

1. i can't seem to figure out or connect to the JMX port on prod or staging (can do so locally). without it, datadog can't pick up kafka stats
2. i'm confused as to how to figure out what the consumer groups are for each topic. i had thought that they would be the group-id for the consumer, but datadog doesn't seem to be picking any changes up. maybe @czue knows something

### DIFF
--- a/ansible/group_vars/kafka.yml
+++ b/ansible/group_vars/kafka.yml
@@ -1,0 +1,3 @@
+datadog_integrations:
+  - kafka
+  - kafka_consumer
--- a/ansible/group_vars/kafka.yml
+++ b/ansible/group_vars/kafka.yml
@@ -1,3 +1,4 @@
 datadog_integrations:
   - kafka
   - kafka_consumer
+  - zk
--- a/ansible/roles/datadog/templates/kafka.yaml.j2
+++ b/ansible/roles/datadog/templates/kafka.yaml.j2
@@ -1,0 +1,383 @@
+#
+# {{ ansible_managed }}
+#
+instances:
+  - host: localhost
+    port: 9999 # This is the JMX port on which Kafka exposes its metrics (usually 9999)
+    tags:
+      kafka: broker_{{ kafka_broker_id }}
+    name: kafka_instance
+
+init_config:
+  is_jmx: true
+
+  # Metrics collected by this check. You should not have to modify this.
+  conf:
+    # v0.8.2.x Producers
+    - include:
+        domain: 'kafka.producer'
+        bean_regex: 'kafka\.producer:type=ProducerRequestMetrics,name=ProducerRequestRateAndTimeMs,clientId=.*'
+        attribute:
+          Count:
+            metric_type: rate
+            alias: kafka.producer.request_rate
+    - include:
+        domain: 'kafka.producer'
+        bean_regex: 'kafka\.producer:type=ProducerRequestMetrics,name=ProducerRequestRateAndTimeMs,clientId=.*'
+        attribute:
+          Mean:
+            metric_type: gauge
+            alias: kafka.producer.request_latency_avg
+    - include:
+        domain: 'kafka.producer'
+        bean_regex: 'kafka\.producer:type=ProducerTopicMetrics,name=BytesPerSec,clientId=.*'
+        attribute:
+          Count:
+            metric_type: rate
+            alias: kafka.producer.bytes_out
+    - include:
+        domain: 'kafka.producer'
+        bean_regex: 'kafka\.producer:type=ProducerTopicMetrics,name=MessagesPerSec,clientId=.*'
+        attribute:
+          Count:
+            metric_type: rate
+            alias: kafka.producer.message_rate
+
+
+    # v0.9.0.x Producers
+    - include:
+        domain: 'kafka.producer'
+        bean_regex: 'kafka\.producer:type=producer-metrics,client-id=.*'
+        attribute:
+          response-rate:
+            metric_type: gauge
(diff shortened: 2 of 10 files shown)


ansible/roles/datadog/templates/kafka_consumer.yaml.j2
```diff
@@ -0,0 +1,35 @@
...
+  zk_timeout: 5
+#  Customize the Kafka connection timeout here
+  kafka_timeout: 5
+
+instances:
+  - kafka_connect_str: localhost:9092  # Should always be deployed to broker so localhost is OK
+    zk_connect_str: "{{ groups['zookeeper'][0] }}:{{ zookeeper_client_port }}"
+    zk_prefix: /0.8
+    consumer_groups:
+      'form-processsor':
```
benrudolph: does this seem right to you?

snopoke: As far as I know the groups are what gets passed to the `KafkaChangeFeed` class as `group_id`. e.g.
```
KafkaChangeFeed(topics=[GROUP], group_id='groups-to-es')
```

snopoke: Sorry, misunderstood you're question. Yes I think this looks correct.

snopoke:
Re JMX, do we have the JMX_PORT environment variable set?
```
Before you begin, you must verify that Kafka is configured to report metrics via JMX, and that you can communicate with ZooKeeper, usually on port 2181. For Kafka, that means confirming that the JMX_PORT environment variable is set before starting your broker (or consumer or producer), and then confirming that you can connecting to that port with JConsole.
```

For 2 it seems like you might need to setup [Burrow](https://www.datadoghq.com/blog/collecting-kafka-performance-metrics/#burrow) to get consumer metrics. If that is the case then probably worth doing it as a separate task.

benrudolph:
we don't have that port set by default. i didn't need to set locally, but i did try setting it to 9999 on staging and seeing if i could connect to it from my local machine using jconsole. couldn't seem to connect to it and was hard verify if it was actually working. will give it another go

will look into burrow thanks

ansible/roles/datadog/templates/kafka_consumer.yaml.j2
czue: the list looks good, though maintaining this here in duplicate with the code doesn't feel like a great long-term solution.

we don't really use groups at all yet. i think the way we would start using them is if we needed to have multiple pillows of the same type because we can't handle the throughput on a single one. also as you noted our partition setup is pretty hard coded for a single partition per topic. 

i don't think any of this blocks anything on this PR, just noting that there are some architectural issues with scaling out the environment if and when we need to do that

benrudolph: agreed that this list would be hard to maintain, though it doesn't have to be up to date. only if we want stats on a certain group. i'm imagining those groups won't change too often. 

good to note on the groups wasn't certain if we were using them correctly

benrudolph:
Made some progress on this: https://p.datadoghq.com/sb/5c4af2ac8-2b131e4065. Got the JMX Port thing to work so it's not sending broker metrics to datadog and java metrics. I haven't taken the time to learn all about the JVM metrics young-gen vs. old-gen garbage collection, so I didn't include any of that stuff yet on the dashboard. 

Didn't get the group stuff working with burrow, but I think that's OK for now. I think it makes sense to merge this and work on that later as you mentioned.

One thing that still confusing me is how datadog is telling me to setup the `kafka.yaml` file:
> kafka.yaml
> You can use the example configuration provided whether you are monitoring your brokers, producers, consumers, or all three. Just change the host and port appropriately. Though you could monitor the entirety of your deployment from one host, it is recommended that you install the Agent on each of your producers, consumers and brokers, and configure each separately.

Do you think that means I need to put `kafka.yaml` on pillowtop and webworker machines? Aren't those where the produces and consumers live?

ansible/group_vars/kafka.yml
```diff
@@ -0,0 +1,4 @@
+datadog_integrations:
+  - kafka
+  - kafka_consumer
+  - zk
```
benrudolph: this should actually be in ` ansible/group_vars/zookeeper.yml` but i think since zookeeper and kafka are on the same machine the datadog_integrations variable gets overridden so it doesn't properly deploy the zk.yaml file

benrudolph: for now we always deploy zk on kafka so i think we can cross that bridge when we get there

ansible/roles/datadog/templates/kafka.yaml.j2
```diff
@@ -0,0 +1,383 @@
...
+    tags:
+      kafka: broker_{{ kafka_broker_id }}
+    name: kafka_instance
+
+init_config:
+  is_jmx: true
+
+  # Metrics collected by this check. You should not have to modify this.
+  conf:
+    # v0.8.2.x Producers
```
snopoke: Seems like we can remove the sections for older versions of kafka

snopoke:
don't think we need to have a datadog agent on each producer / consumer to monitor. Also don't really see how that would work since we're using the python client.

benrudolph:
makes sense, thanks, wasn't sure if datadog was magically monitoring the traffic we were sending on a port or something

PR #445, opened 2016-06-24
This kafka path was wrong

esoergel:
@proteusvacuum
You can see the correct path being used (hardcoded) here:
https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/kafka/files/kafka.sh#L2-L2


proteusvacuum:
:+1: Should this be a variable instead? `{{kafka_path}}` that is used in both places?

esoergel:
It could be, but I don't expect it to change.

PR #446, opened 2016-06-27
use template variables

snopoke:
I'm sure I this get fixed recently but can't find it.
@nickpell


benrudolph:
haha that is in my open PR. good to do this separately though!

PR #447, opened 2016-06-27
fix syntax for supervisor install check

dannyroberts:
I happened to come upon this, but didn't end up running it again with this fix. This seems to be the preferred way to check for empty stdout, looking at the docs, and the current way was failing (though I can't say whether it's for this reason).


dannyroberts:
It was apparently not for this reason. Closing.

PR #448, opened 2016-06-27
run non-destructive command in check mode too

dannyroberts:


dannyroberts:
@esoergel just learned you can use this to make a shell command run in check mode (which you only want to do if it's non-destructive). Very helpful for making check mode not needlessly broken.

dannyroberts:
@snopoke thanks!

snopoke:
👍 very good to know

PR #449, opened 2016-06-27
run "find supervisorctl path" in check mode

dannyroberts:
Now `deploy_common.yml` can run start to finish in check mode without failing! (That exclamation point is pretty depressing...)


dannyroberts:
@snopoke one more!

PR #452, opened 2016-07-04
Remove deprecated rackspace role

emord:
@esoergel I think we discovered that this is deprecated and it's all online now right?

cc @snopoke


esoergel:
Yup, this is all done through the rackspace UI.

ansible/inventories/test
```diff
@@ -40,5 +40,3 @@ localhost ansible_connection=local
 [riakcs]
 
 [stanchion]
-
-[rackspace_cloud_backup]
```
esoergel: This should also be removed from the corresponding inventory files

PR #454, opened 2016-07-11
Ce/pg standby

calellowitz:
@benrudolph @czue @millerdev @snopoke 
Here is my first pass at a hot standby setup for postgres. Setting it all up still requires a few manual actions (setting up ssh between the db and standby machines, taking a filesystem backup of the main db and rsyncing it to the standby machine, moving the `recovery.conf` file to the data directory on the standby machine, and restarting postgres on the standby machine), but I wanted to get the deployment work reviewed as soon as possible, so we can get the standby up.


ansible/roles/pg_standby/templates/create_postgres_dump.sh.j2
```diff
@@ -3,7 +3,7 @@ BACKUP_TYPE=$1
 DAYS_TO_RETAIN_BACKUPS=$2
 
 TODAY=$(date +"%Y_%m_%d")
-pg_dumpall --clean | gzip  > "{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}.db.gz"
+pg_basebackup -D "{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}" -Ft -Xs -z
```
calellowitz: I switched to filesystem level backups in our backup scripts because they are often faster to execute, but definitely faster to restore from. If both the main and standby servers go down restoring to the most recent backup is as simple as extracting the tar file and then turning on postgres.

ansible/roles/pg_standby/tasks/main.yml
```diff
@@ -0,0 +1,61 @@
...
+- name: Copy postgres backup script
+  sudo: yes
+  template:
+    src: "create_postgres_dump.sh.j2"
+    dest: "/etc/cron.d/create_postgres_dump.sh"
+    group: postgres
+    owner: postgres
+    mode: 0700
+    backup: yes
+  when: backup_postgres
```
calellowitz: I moved the backup scripts to the standby machine to take that load off our main machine

ansible/roles/pg_standby/templates/recovery.conf.j2
```diff
@@ -0,0 +1,153 @@
...
+#
+restore_command = 'cp {{ archive_dir }}/%f %p'		# e.g. 'cp /mnt/server/archivedir/%f %p'
+#
+#
+# archive_cleanup_command
+#
+# specifies an optional shell command to execute at every restartpoint.
+# This can be useful for cleaning up the archive of a standby server.
+#
+archive_cleanup_command = 'pg_archivecleanup {{ archive_dir }} %r'
```
calellowitz: this deletes WAL archives as they are replicated into the standby to free up disk space.

ansible/roles/postgresql/templates/pg_hba.conf.j2
```diff
@@ -98,7 +98,7 @@ host    all             all             127.0.0.1/32            md5
 host    all             all             ::1/128                 md5
 # Allow replication connections from localhost, by a user with the
 # replication privilege.
-#local   replication     postgres                                peer
+local   replication     postgres                                peer
```
calellowitz: I needed to allow local replication for `pg_basebackup` to be run, which is required as the base for a standby server

ansible/roles/pg_standby/templates/recovery.conf.j2
```diff
@@ -0,0 +1,153 @@
...
+# restore_command and/or primary_conninfo.
+#
+standby_mode = on
+#
+# primary_conninfo
+#
+# If set, the PostgreSQL server will try to connect to the primary using this
+# connection string and receive XLOG records continuously.
+#
+#primary_conninfo = ''		# e.g. 'host=localhost port=5432'
```
snopoke: Do we not want this?

ansible/roles/pg_standby/templates/recovery.conf.j2
```diff
@@ -0,0 +1,153 @@
...
+#
+#primary_slot_name = ''
+#
+# By default, a standby server keeps restoring XLOG records from the
+# primary indefinitely. If you want to stop the standby mode, finish recovery
+# and open the system in read/write mode, specify a path to a trigger file.
+# The server will poll the trigger file path periodically and start as a
+# primary server when it's found.
+#
+#trigger_file = ''
```
snopoke: Don't we need this to do a failover?

calellowitz: Its an optional failover method. We can also use `pg_ctlcluster promote`, which seemed easier, but there isn't a downside to defining a trigger file if you prefer that method.

ansible/roles/pg_standby/templates/recovery.conf.j2
calellowitz: This is only needed for streaming replication. Its currently setup for simple log shipping, so the standby is just reading the log files from the archive directory and doesn't need to connect to the main db server. Managing that connection seemed like an extra bit of complexity that was unnecessary to get a simple standby up and running.

calellowitz:
just tested a failover on staging to a standby db, and it all works as expected.

ansible/roles/pg_standby/tasks/main.yml
benrudolph: nit: this may be easier to configure with ansible's cron module: http://docs.ansible.com/ansible/cron_module.html

benrudolph: ah i see that that module doesn't let you do templating and that you already found the cron module below

ansible/roles/pg_standby/templates/recovery.conf.j2
benrudolph: that's pretty neat

ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -201,7 +200,7 @@ vacuum_cost_delay = 20ms        # 0-100 milliseconds
 
 # These settings are ignored on a standby server
 
-#max_wal_senders = 0        # max number of walsender processes
+max_wal_senders = 3        # max number of walsender processes
```
benrudolph: how come these two properties aren't encompassed in the `if` block?

calellowitz: `max_wal_senders` is necessary for `pg_basebackup`, and therefore our backup scripts, so its potentially useful even in deployments without a standby, but more generally, neither setting should affect the normal running of a postgres server. The `hot_standby` is explicitly ignored if the server isn't in recovery mode and increasing the `max_wal_senders` didn't seem to have a downside on a running machine

calellowitz:
are there any blocking questions here? i would love to get this set up in prod as soon as possible

snopoke:
None from me though I would like to understand a bit more about exactly how
this works and what the failover procedure is. I think a wiki page would be
a good place to start.

calellowitz:
@snopoke I wrote up a short confluence page describing both the process of standing up a standby server and what a failover would look like. its at https://confluence.dimagi.com/display/commcarehq/Postgres+Standby if you want to take a look

snopoke:
@calellowitz thanks - can you add details like how long the lag between the primary and standby is?

Also, how does with work with pgbouncer?

calellowitz:
@snopoke the lag between the primary and the standby is pretty dependent on how heavy a load the primary db is under. As soon as the logs are completed they are sent to the standby, but if there are few writes to the primary it could be a while before each WAL segment is sent. I dont have a good sense of how long that will generally be on prod, but I would guess no more than an hour, and likely at worst on the order of a few minutes. We can also set a max timeout, where WAL segments would be sent regardless of whether they were full. My understanding is that that would be the bottleneck and the standby should have no trouble keeping up with WAL segments at whatever rate they are sent.

As far as pgbouncer, since we do a full ansible postgres deploy to the standby machine, it has its own instance of pgbouncer, which should always be running. When we switch the db host it still points to the port pgbouncer is on, so new requests hit pgbouncer on the standby machine instead of the primary machine.

ansible/roles/postgresql/templates/postgresql.conf.j2
snopoke: Got an error when testing this locally (without a hot_standby server):

FATAL:  WAL streaming (max_wal_senders > 0) requires wal_level "archive", "hot_standby", or "logical"

PR #459, opened 2016-07-13
Revert "Let the ansible user sudo without a password"

esoergel:
Reverts dimagi/commcarehq-ansible#458
This actually isn't necessary, and it's overwridden by the cchq sudoers file anyways.  I think you just need to do something manual during initial set-up.


esoergel:
@dannyroberts @czue

PR #460, opened 2016-07-13
devreadonly user

benrudolph:
@snopoke i couldn't find anywhere where we actually set this user up. i went through some iterations on this to try and get it right. still trying to figure out why i'm only able to read from some tables

cc: @wpride @orangejenny

### DIFF
--- a/ansible/roles/postgresql/tasks/main.yml
+++ b/ansible/roles/postgresql/tasks/main.yml
@@ -95,25 +95,15 @@
 - name: Create PostgreSQL devreadonly user
   sudo_user: postgres
   postgresql_user:
-    name: postgres_users.devreadonly.username
-    password: postgres_users.devreadonly.password
+    name: "{{ postgres_users.devreadonly.username }}"
+    password: "{{ postgres_users.devreadonly.password }}"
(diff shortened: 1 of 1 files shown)


snopoke:
You should take a look at this: https://github.com/dimagi/commcarehq-ansible/pull/380/files#diff-ca3206fc1d7501d5550eed5478ab055fR120

This also described it quite well: http://stackoverflow.com/a/762649

benrudolph:
ah great thanks!

benrudolph:
man can't believe i missed that

snopoke:
I think that PR hasn't been merged because it contains some Ansible 2.0
stuff. @sravfeyn might recall exactly what.

sravfeyn:
Yeah, that PR is supposed to manage various pg user and db creation, and permissions via a user-db-permission mapping config. It needs Ansible2.0. I didn't prioritize it as it was not needed for ICDS, yet. 

That PR is more general purpose than this, but I am fine with having this merged for now.

benrudolph:
closing for now as this is no longer blocking formplayer and seems the work in the other PR is the better route

PR #461, opened 2016-07-18
Add a read-only elasticsearch endpoint

NoahCarnahan:
This adds an endpoint for accessing elasticsearch that only accepts GET requests, and is therefore read only. It can be accessed on port 9201.

### DIFF
--- a/ansible/roles/nginx/templates/site.j2
+++ b/ansible/roles/nginx/templates/site.j2
@@ -53,7 +53,10 @@ server {
     {% if v.get('is_internal', False) == True %}
         internal;
     {% endif %}
-    {% for x,y in v.iteritems() if x not in ['name', 'balancer', 'is_internal', 'proxy_set_headers'] %}
+    {% if v.get('if', False) %}
+        if {{ v['if'] }}
+    {% endif %}
+    {% for x,y in v.iteritems() if x not in ['name', 'balancer', 'is_internal', 'proxy_set_headers', 'if'] %}
         {{ x }} {{ y }};
     {% endfor %}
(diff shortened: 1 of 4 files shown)


ansible/roles/nginx/templates/site.j2
```diff
@@ -53,7 +53,10 @@ server {
     {% if v.get('is_internal', False) == True %}
         internal;
     {% endif %}
-    {% for x,y in v.iteritems() if x not in ['name', 'balancer', 'is_internal', 'proxy_set_headers'] %}
+    {% if v.get('if', False) %}
+        if {{ v['if'] }}
```
NoahCarnahan: The special case here doesn't add a `;` to the end of the statement.

NoahCarnahan:
Forgot to ping @emord

emord:
Looks good. Does this add any benefits other than having a safer endpoint we can use for es head?

Also could we change the write endpoint to only listen to requests from inside the vpn?

NoahCarnahan:
No other benefits, just thought it would be nice to have a safer endpoint to use when inspecting  prod data. Might be possible to make the write endpoint safer too, but going to punt that for now.

snopoke:
Does this not give the world access to our ES data? Or do you still have to be on the VPN to access that port.

emord:
Sounds reasonable. I've often wondered/worried how easy it would be for me to screw something up using es head. I don't think I've done it yet, but was alwasy in the back of my mind. Thanks for looking into this.

NoahCarnahan:
Good point Simon. Looks like this we should put this nginx site on the ES machines instead. Does that sound correct to you?

PR #463, opened 2016-07-20
pg fixes

snopoke:
@calellowitz

### DIFF
--- a/ansible/inventories/development
+++ b/ansible/inventories/development
@@ -43,3 +43,5 @@
 
 [shared_dir_host]
 192.168.33.16
+
+[pg_standby]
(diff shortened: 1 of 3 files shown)


snopoke:
:white_check_mark:

calellowitz:
looks good. systems without a standby will no longer be able to run the new backup scripts, but i those are only installed on standby machines now anyway. if thats fine by you happy to merge.

snopoke:
These are just inventories for local testing so no backups necessary.

calellowitz:
:+1:

emord:
> systems without a standby will no longer be able to run the new backup scripts

@calellowitz Seems bad for monoliths like swiss? Or is there some other backup process they're following

calellowitz:
@emord
currently our monoliths dont have any backup process so this change wont affect them per se. It just means they will need a different snapshot strategy than we use on prod.

PR #464, opened 2016-07-21
ansible support for multiple db machines

snopoke:
@benrudolph looking for feedback on this approach

cc @calellowitz 

#### Inventory
```
[postgresql]
192.168.33.16 hot_standby_server=192.168.33.18
192.168.33.17 hot_standby_server=192.168.33.19

[pg_standby]
192.168.33.18
192.168.33.19
```

#### vars.yml
All databases on `192.168.33.16` except `commcare_p3` and `commcare_p4`.
DB users created on ALL DB machines.
```
postgresql_dbs:
  - django_alias: default
    name: "{{localsettings.PG_DATABASE_NAME}}"
  - django_alias: proxy
    name: commcarehq_proxy
  - django_alias: p1
    name: commcarehq_p1
    shards: [0, 99]
  - django_alias: p2
    name: commcarehq_p2
    shards: [100, 199]
  - django_alias: p3
    name: commcarehq_p3
    shards: [200, 299]
  - django_alias: p4
    name: commcarehq_p4
    host: 192.168.33.17
    shards: [300, 399]
  - django_alias: p5
    name: commcarehq_p5
    host: 192.168.33.17
    shards: [400, 511]
```


ansible/roles/postgresql/tasks/main.yml
```diff
@@ -45,27 +42,24 @@
...
-  when: pg_install == True and orig_path.stat.exists
+  when: orig_path.stat.exists
 
 - name: generate en_US.UTF-8 locale
   locale_gen: name='en_US.UTF-8' state=present
-  when:
-    pg_install == True
+
+- name: assert hot standby config correct
+  assert: { that: "hot_standby_server is not defined or hot_standby_server in groups.pg_standby"}
```
snopoke: This doesn't check that multiple servers don't point to the same hot standby.

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -89,8 +83,10 @@
     role_attr_flags: CREATEDB
     state: present
     port: "{{ postgresql_port }}"
-  when:
-    pg_install == True
+
+- name: check postgres database hosts
+  assert: { that: "item.host is not defined or item.host in groups.postgresql" }
+  with_items: "{{ postgresql_dbs }}"
```
snopoke: validate that host specified in vars file is actually a DB host

ansible/roles/postgresql/defaults/main.yml
```diff
@@ -1,4 +1,3 @@
-pg_install: True
```
benrudolph: 🙌 . one thing to ensure is that this role is idempotent when pg_install is True (i think it is)

benrudolph:
looks good and way less invasive than i thought it would be. the only concerns i have are implicit assumptions we have that the postgresql group being one host. some grepping in the ansible repo convinced me that we actually don't have too many.

`ansible/roles/touchforms/templates/application.properties.j2` this file assumes pg bouncer is installed on the same machine as the first host in the postgres group (which is probably true, but not necessarily)

also i think changing the inventory has some implications on our deploy, probably not too many, but good to check.

👍

ansible/roles/postgresql/defaults/main.yml
snopoke: checked that this is true

snopoke:
This is GTG. Also https://github.com/dimagi/commcare-hq-deploy/pull/83

benrudolph:
👍

PR #465, opened 2016-07-21
Danny cal

calellowitz:
@dannyroberts @esoergel @millerdev 
This gets adds a few the roles from `common` that are germane to the control machine. I still plan on breaking up `common_installs` to get the truly common parts and adding that here as well, but i wanted to make sure we could deploy the necessary ssh changes first, and that one may require a bit more review.


millerdev:
👍 on ✅

PR #466, opened 2016-07-22
add separate credentials for cloudant monitoring

dannyroberts:
@snopoke @gcapalbo buddy

### DIFF
--- a/ansible/vars/dev.yml
+++ b/ansible/vars/dev.yml
@@ -162,6 +162,8 @@ localsettings:
   COUCH_PASSWORD: commcarehq
   COUCH_SERVER_ROOT: "{{ groups.couchdb.0 }}:5984"
   COUCH_USERNAME: commcarehq
+  COUCH_MONITORING_USERNAME: commcarehq
+  COUCH_MONITORING_PASSWORD: commcarehq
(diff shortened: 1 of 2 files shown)


ansible/vars/dev.yml
```diff
@@ -162,6 +162,8 @@ localsettings:
   COUCH_PASSWORD: commcarehq
   COUCH_SERVER_ROOT: "{{ groups.couchdb.0 }}:5984"
   COUCH_USERNAME: commcarehq
+  COUCH_MONITORING_USERNAME: commcarehq
+  COUCH_MONITORING_PASSWORD: commcarehq
```
snopoke: What's the reason for adding it to the `localsettings` dict?

dannyroberts: Good question. I didn't think about it so much, I just put them next to the other couchdb settings. But now that I *do* think about it, the next step here is to integrate it with the systems page, so it will eventually go into localsettings. So I guess I'd advocate keeping it here even though it's not going into localsettings.py in this PR

dannyroberts:
Bump

dannyroberts:
Thanks!

PR #467, opened 2016-07-22
Allow ansible deploy from dev account on control machine

millerdev:
This should make it possible to remove the ansible key pair from all machines and switch to using dev keys for authentication.

When logging into the ansible control machine, it will be important to use `ssh -A` or add `ForwardAgent yes` to your ssh config. Not sure if `mosh` has support for agent forwarding yet.

Inline comments to follow. Feedback welcome.

@dannyroberts cc @proteusvacuum


ansible/roles/ssh/tasks/main.yml
```diff
@@ -11,3 +11,10 @@
               line="PasswordAuthentication no"
               state=present
   notify: Restart ssh
+
+- name: Set SSH logging level to VERBOSE
+  lineinfile: dest=/etc/ssh/sshd_config
+              regexp="^LogLevel"
+              line="LogLevel VERBOSE"
+              state=present
+  notify: Restart ssh
```
millerdev: This allows us to see which key was used to login to a given account. Log output is something like:
```
Jul 22 13:40:15 precise64 sshd[27718]: Connection from 192.168.33.14 port 37519
Jul 22 13:40:15 precise64 sshd[27718]: Found matching RSA key: 49:33:77:25:3e:05:ba:51:59:2a:c9:aa:f1:bd:24:29
Jul 22 13:40:15 precise64 sshd[27718]: Postponed publickey for dmiller from 192.168.33.14 port 37519 ssh2 [preauth]
Jul 22 13:40:15 precise64 sshd[27718]: Found matching RSA key: 49:33:77:25:3e:05:ba:51:59:2a:c9:aa:f1:bd:24:29
Jul 22 13:40:15 precise64 sshd[27718]: Accepted publickey for dmiller from 192.168.33.14 port 37519 ssh2
```
See also https://help.ubuntu.com/community/SSH/OpenSSH/Configuring#Log_More_Information

control/init.sh
```diff
@@ -0,0 +1,61 @@
+#! /bin/bash
```
millerdev: This script initializes a developer account for ansible deploys. There is no longer a single shared ansible repo. Instead, each developer has a clone from which to initiate deploys. Should help with people stepping on other people's work. The control repo is symlinked in from the `/home/ansible` (more on that below).

control/init.sh
```diff
@@ -0,0 +1,61 @@
...
+    mkvirtualenv ansible
+    pip install -r ~/commcarehq-ansible/ansible/requirements.txt
+else
+    workon ansible
+fi
+
+if ! [ -L ~/commcarehq-ansible/config -o -d ~/commcarehq-ansible/config ]; then
+    # TODO setup permissions (add user to group?) to allow write to config repo
+    #su ansible -c "usermod -a -G ansible $USER"
+    ln -s /home/ansible/commcarehq-ansible/config ~/commcarehq-ansible/config
```
millerdev: This is still shared, so still has most of the same problems we had before: people stepping on other people's work or accidentally leaving uncommitted changes. I still need to figure out how commits to this repo will workr. One advantage of committing from separate dev accounts is that each dev will have a separate git config, so commits should show correct author details.

control/init.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+if [ ! -d ~/commcare-hq-deploy ]; then
+    echo "Cloning commcare-hq-deploy..."
+    git clone git@github.com:dimagi/commcare-hq-deploy ~/commcare-hq-deploy
+    if [ ! -d ~/commcare-hq ]; then
+        # keep old commands working: create symlink to simulate cchq repo
+        mkdir ~/commcare-hq
+        ln -s ~/commcare-hq-deploy/fab ~/commcare-hq/fab
```
millerdev: I think this was the only thing that was used by the deploy, so no need to have a clone of the cchq repo for each dev since inventories have moved to a separate deploy repo. Is that correct?

provisioning/control.sh
```diff
@@ -13,6 +9,12 @@ ssh-keyscan 192.168.33.15 >> /home/vagrant/.ssh/known_hosts
 ssh-keyscan 192.168.33.16 >> /home/vagrant/.ssh/known_hosts
 ssh-keyscan 192.168.33.17 >> /home/vagrant/.ssh/known_hosts
 
+ln -s /vagrant /home/vagrant/commcarehq-ansible
+mkdir /home/vagrant/commcare-hq-deploy
+sudo pip install virtualenv virtualenvwrapper
+sudo -H -u vagrant /vagrant/control/init.sh
```
millerdev: Vagrant control now uses the same script for ansible setup. The aliases don't make much sense though since vagrant doesn't have a proper `config` dir.

provisioning/control.sh
```diff
@@ -13,6 +9,12 @@ ssh-keyscan 192.168.33.15 >> /home/vagrant/.ssh/known_hosts
...
+echo '[ -t 1 ] && source ~/init-ansible' >> /home/vagrant/.profile
```
millerdev: Do not `init-ansible` unless connected to a TTY

snopoke:
Very exciting

control/init.sh
millerdev: Poll: should we...
- Fiddle with permissions and try to make it so everyone can modify and commit to a single central git repo at `/home/ansible/commcarehq-ansible/config`. This is very similar to what we are doing now. Keeping the permissions correct so everyone can write could be tricky. Vote with 🎉 
- Put a central bare repo at `/home/ansible/commcarehq-ansible-config.git` or some similar path and make a separate clone of that repo in each dev account. Typical git collaboration model (pull, edit, commit, push), and does not require special permissions. Vote with ❤️ 

@dimagi/team-commcare-hq please vote by adding a reaction to this message. Add another comment if you have a better idea.

benrudolph: Will this all be mitigated if we switch over to ansible Vault so we can commit the config repo on github?

millerdev: Yes. That would also allow deploys directly from your own machine rather than having to login to the control node to do it. It would be very nice, but out of scope for this PR.

millerdev: Everyone's voting for the same option so I'm going to implement the ❤️. This will entail a small adjustment to our current process: changes in the config repo in the ansible account will also need to be pushed to the central repo while the ansible account is still in use. Feel free to point out if I've overlooked some issue concerning security of our secrets. I think the security is unchanged, but could have overlooked something.

control/init.sh
benrudolph: yes! that's correct, been on my backlog forever to implement

benrudolph: these will probably have to be updated: https://github.com/dimagi/ansible-dotfiles/blob/master/.bash_aliases#L1

benrudolph: nvm i see you have it below

control/init.sh
millerdev: Implemented ❤️ in https://github.com/dimagi/commcarehq-ansible/pull/467/commits/e143f119bf7d4e754e743b96fb478e241feb36fb

control/init.sh
dannyroberts: This file seems to copy a lot of stuff from ansible-dotfiles. Is the idea that this will replace that?

millerdev:
Setup steps:
- In the "ansible" account on the control machine: checkout `dm/secure-deploy` and run `ENV=... && aps --tags=users,ssh`. This needs to be done for each ENV; so far it's only been done on staging. This does two things:
  - Add dev public keys to /home/ansible/.ssh/authorized_keys.
  - Set sshd log level to verbose.
- [Setup your dev account](https://github.com/dimagi/commcarehq-ansible/blob/80aa7a0f987f8e4f6e0b17a19a50a77fdc9a6c27/README.md#setting-up-a-dev-account-on-ansible-control-machine) on the control machine.
- Test something like `ENV=staging && aps --diff --tags=ssh --check`

Sometime later when everything is working smoothly from dev accounts, decommission the old ansible keys:
- Remove "ansible" and "cchq" from `dev_users.present` list (do not add  to `dev_users.absent`)  in `config/$ENV/$ENV.yml`. "ansible" and "cchq" should not have their own private keys since they are shared accounts.
- Remove the "ansible" private and public keys from the control machine. Also remove "cchq" keys (do they exist?)

control/init.sh
millerdev: Yes. It replaces that entirely. The reason I did this is so we have one repo to clone (this one) and then one command to run `. commcarehq-ansible/control/init.sh` to bootstrap a dev account.

ansible/roles/users/tasks/main.yml
```diff
@@ -9,6 +9,12 @@
   user: name={{ item.name }} state=present shell=/bin/bash groups={{ dev_group }}
   with_items: '{{ dev_users.present }}'
 
+- name: Add public keys
+  sudo: yes
+  authorized_key: user={{ item.name }} state=present key="{{ item.key }}" exclusive=yes
+  with_items: '{{ dev_users.present }}'
+  when: item.name != 'ansible'
```
millerdev: Exclude "ansible" because `exclusive=yes` option would remove dev keys from `/home/ansible/.ssh/authorized_keys`.

control/init.sh
snopoke: FYI I am in the process of implementing the Vault solution

millerdev: Should work fine with this (and a few extra small changes), correct?

snopoke: yea, just need to change the aliases

millerdev:
✅ ready to merge.

millerdev:
I will roll out `aps --tags=users,ssh` for each env once this is merged. Setting up dev account can be done by each dev individually. Deploys can still be done from the ansible account until we decommission the ansible key pair, although we should all try to use dev accounts to make sure everything works as expected.

benrudolph:
🔐 👪

proteusvacuum:
@millerdev were you ever able to update the ansible docs? This [page](https://confluence.dimagi.com/display/internal/Ansible) is very out of date now.

millerdev:
I did not update that page on confluence. Last time I worked on it I updated the [ansible README](https://github.com/dimagi/commcarehq-ansible/blob/master/README.md), specifically these sections:

- https://github.com/dimagi/commcarehq-ansible#use-your-ssh-key-to-authenticate
- https://github.com/dimagi/commcarehq-ansible#setting-up-a-dev-account-on-ansible-control-machine
- https://github.com/dimagi/commcarehq-ansible#simulate-dev-user-setup-on-vagrant-control-machine

millerdev:
^ @proteusvacuum

PR #469, opened 2016-07-25
putting shared drive on data disk

calellowitz:
@snopoke @czue @millerdev @benrudolph 
Our prod db machine comes with a 160GB data disk that no one seems to know existed and has remained unmounted since the migration to virginia. This PR allows us to mount that drive make it the nfs drive, to deal with repeated issues we have running out of disk space on db0 and the shared drive. Once we switch to the new drive, i can manually rsync the contents of the old one over.

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -9,7 +9,7 @@ encrypted_root: "/opt/data"
 
 shared_drive_enabled: true  # overridden in ../config/$ENV/$ENV.yml
 shared_dir_gid: 1500  # This GID cannot already be allocated
-shared_dir_name: "shared_data{{ '_' ~ deploy_env if deploy_env != 'production' else '' }}"
+shared_dir_name: "shared{{ '_' ~ deploy_env if deploy_env != 'production' else '' }}"
 shared_data_dir: "/opt/{{ shared_dir_name }}"
 shared_mount_dir: "/mnt/{{ shared_dir_name }}"
 is_monolith: '{{ groups["all"]|length == 1 }}'
--- a/ansible/roles/shared_dir/tasks/setup_host.yml
+++ b/ansible/roles/shared_dir/tasks/setup_host.yml
@@ -12,6 +12,20 @@
     - nfs
     - shared_files
 
+- name: Format system data disk
+  sudo: yes
+  filesystem: fstype=ext4 dev={{ datadisk_device }}1
+  when: datadisk_device is defined
+  tags:
+    - shared_files
+
+- name: Mount system data disk
+  sudo: yes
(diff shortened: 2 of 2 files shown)


ansible/group_vars/all.yml
```diff
@@ -9,7 +9,7 @@ encrypted_root: "/opt/data"
 
 shared_drive_enabled: true  # overridden in ../config/$ENV/$ENV.yml
 shared_dir_gid: 1500  # This GID cannot already be allocated
-shared_dir_name: "shared_data{{ '_' ~ deploy_env if deploy_env != 'production' else '' }}"
+shared_dir_name: "shared{{ '_' ~ deploy_env if deploy_env != 'production' else '' }}"
```
calellowitz: i changed the name of the shared drive to create a new mountpoint for the device. if the old directory name is preferred, we can switch it back once the new drive is set up.

ansible/roles/shared_dir/tasks/setup_host.yml
```diff
@@ -12,6 +12,20 @@
     - nfs
     - shared_files
 
+- name: Format system data disk
+  sudo: yes
+  filesystem: fstype=ext4 dev={{ datadisk_device }}1
```
benrudolph: why does the `1` get added here? i've seen it before but wasn't sure why it was hardcoded here instead of the device name

ansible/roles/shared_dir/tasks/setup_host.yml
```diff
@@ -12,6 +12,20 @@
...
+  when: datadisk_device is defined
+  tags:
+    - shared_files
```
benrudolph: i think this should be `- nfs`. shared_files is used for monolith machines

calellowitz: i used the `shared_files` tag because these tasks seemed to fit more with the other tasks using that tag. both get run when you deployed the shared directory on non monolith machines and this isnt specific to the nfs. if monolith machines had data disks we could do this too.

ansible/roles/shared_dir/tasks/setup_host.yml
calellowitz: the first partition on the device is the device name followed by 1. so here the data disk is `/dev/xvde` and its first partition, which is the entire drive in this case is `/dev/xvde1`. the `1` is only added here. when we mount it, its the name without the 1

benrudolph: nice thanks for the explanation

czue:
nice

millerdev:
👍 from me. Little scared to merge though.

ansible/group_vars/all.yml
benrudolph: seems fine to me

PR #470, opened 2016-07-25
Beginnings of smtp email for formplayer

benrudolph:
@wpride settings for formplayer. this is still pending on being able to set authentication params in the formplayer


benrudolph:
k this has been deployed and g2merge

PR #471, opened 2016-07-26
proper dir for inventory file

benrudolph:
@millerdev my `aps` command was having trouble finding the correct dir. i think this needs to be updated, but could be missing something


benrudolph:
nvm the script failed i think

PR #472, opened 2016-07-26
Clarify ssh ForwardAgent instructions

esoergel:
@millerdev


millerdev:
👍 I was just re-reading that and thinking it should be re-worded. thanks.

esoergel:
Actually, it could also use something along the lines of "ansible uses your local ssh key to authenticate via a mechanism called `ForwardAgent`", I'll offline

PR #474, opened 2016-07-27
Deploy local settings to both dirs

benrudolph:
@wpride @esoergel on the way to making the formplayer build a separate repo


benrudolph:
(this can be merged now with no risk of bad side effects, tested it out on staging)

PR #475, opened 2016-07-28
allow db auth on hot standby

snopoke:
@calellowitz


snopoke:
Also https://github.com/dimagi/commcare-hq-deploy/pull/86

PR #476, opened 2016-07-28
disable per host keys

snopoke:
This allows us to run ansible tasks without having to
ssh into the Vagrant control machine.
https://docs.ansible.com/ansible/guide_vagrant.html

@calellowitz @millerdev @benrudolph


millerdev:
Cool! FWIW, it was already possible to run ansible tasks from the host machine (using your own SSH keys for auth) if `PUBKEY=~/.ssh/id_rsa.pub reset-vms` was used to provision the VMs.

snopoke:
cool, didn't know that. `reset-vms` doesn't work for me at all (syntax
different of xargs between mac and linux I think). Also this relies on the
provisioning scripts working which also don't for me.

millerdev:
can't merge

snopoke:
updated

PR #478, opened 2016-07-28
do not encourage storing vault password in a plaintext file

dannyroberts:
use `--ask-vault-pass` instead of `--vault-password-file`

changed in the `ap`/`aps` aliases and in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,24 +225,18 @@ ansible-playbook -u ansible --ask-sudo-pass -i inventories/development \
 **IMPORTANT**: Install the git hooks to help ensure you never commit secrets into the repo: `./git-hooks/install.sh`
 
 All the secret variables and private data required for the different environments is included
-in this repository as encrypted files (`*_vault.*`).
+in this repository as encrypted files (`${ENV}_vault.yml`).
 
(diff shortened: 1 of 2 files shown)


benrudolph:
looks good! unfortunately with security comes inconvenience 😞

snopoke:
Downvoted just cause it sucks to have to enter 2 passwords - but agree with the logic

dannyroberts:
:( I wonder if there's a way to move SUDO pass into the (vaulted) passwords file instead of `--ask-sudo-pass` then we'd be back to one password, and the sudo password would be encrypted

snopoke:
No real need. I was just complaining.

dannyroberts:
It's actually pretty easy, I just tested it out. Any objection to me making that change? All I have to do is add

```
ansible_sudo_pass: <I'm not going to put it here>
```

to `vars/${ENV}/${ENV}_vault.yml` (which is encrypted of course) and remove `--ask-sudo-pass` from `ap`.

snopoke:
Yea I guess that's fine as long as we have password auth off for ssh logins
which we do. Nice.

dannyroberts:
https://github.com/dimagi/commcarehq-ansible/pull/485

PR #484, opened 2016-07-29
Change deploy-control to output the commands used to deploy control

esoergel:
@snopoke @benrudolph @millerdev 
We could probably tune the user permissions a bit better so we can deploy control from our normal users, but this approach isn't too bad for an infrequent workflow.


millerdev:
👍 

In addition to "this approach isn't too bad for an infrequent workflow" it also requires an extra factor of authentication before changing user accounts, which is probably a good thing.

PR #486, opened 2016-08-02
do not put app props in submodule directory

benrudolph:
@wpride @czue more removing of the formplayer submodule


benrudolph:
ready

PR #487, opened 2016-08-02
enabling datadog for swiss

calellowitz:
enabling datadog for swiss in light of recent performance issues. hoping to get a sense of what is going on next time performance drops.
for context http://manage.dimagi.com/default.asp?231012#1193845
@czue @nickpell cc: @snopoke


benrudolph:
@calellowitz from my understanding from our conversation yesterday, there's no extra cost to turning on datadog for india and staging. does that sound correct to you?

calellowitz:
from my reading of http://docs.datadoghq.com/guides/billing/ it seems like thats the case. we may have to change a few more of our triggers to be environment specific because i think some of them are still global but yeah. we also may need to update new relic as well, if we still use their apdex calculations

calellowitz:
@benrudolph looks like i was wrong about pricing. the pro plan described in the FAQ is only for pricing tiers, it does not come with that number of hosts. We get charged $15 (or $18 depending on how we pay) per host per month

esoergel:
@fyniac mentioned that we DO pay per machine.  I imagine the costs are worth it though.

esoergel:
(oh, looks like @calellowitz beat me to it - I didn't see that last comment 'till after submitting)

snopoke:
Pretty sure we get billed per server so I'd be surprised if there's no
extra cost

czue:
@snopoke :eyes: like a :bird:

PR #488, opened 2016-08-03
Ce/proxy ufw

calellowitz:
@czue @benrudolph @snopoke

### DIFF
--- a/ansible/roles/ufw/tasks/main.yml
+++ b/ansible/roles/ufw/tasks/main.yml
@@ -1,9 +1,6 @@
 - name: allow incoming traffic on private interface
   ufw: rule=allow direction=in interface={{ ufw_private_interface }}
 
-- name: allow ssh traffic from control machine
-  ufw: rule=allow from={{ control_machine_ip }}
-
 - name: default to denying all incoming traffic
   ufw: policy=deny direction=incoming
 
--- a/ansible/roles/ufw/tasks/proxy_ufw.yml
+++ b/ansible/roles/ufw/tasks/proxy_ufw.yml
@@ -1,4 +1,4 @@
-# Rules for monoliths
+# Rules for proxies
 - name: allow web traffic
   ufw:
     rule: allow
@@ -8,7 +8,6 @@
   with_items:
     - 443  # https
     - 80   # http
-  when: is_monolith
 
 - name: allow ssh traffic
   ufw:
@@ -18,6 +17,15 @@
     proto: any
   when: is_monolith
 
+- name: allow only private ssh traffic
+  ufw:
+    rule: allow
+    direction: in
+    port: 22
+    proto: any
+    interface: {{ ufw_private_interface }}
+  when: not is_monolith
+  
 - name: allow mosh traffic
   ufw:
     rule: allow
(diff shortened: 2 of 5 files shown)


ansible/roles/ufw/tasks/main.yml
```diff
@@ -1,9 +1,6 @@
 - name: allow incoming traffic on private interface
   ufw: rule=allow direction=in interface={{ ufw_private_interface }}
 
-- name: allow ssh traffic from control machine
-  ufw: rule=allow from={{ control_machine_ip }}
-
```
calellowitz: this seemed redundant with the line above. if its in a private cloud, the control machine should be able to reach it on the private interface. If not, as in monoliths, port 22 should be open anyway for ssh

calellowitz:
The retro from the india downtime caused me to do some digging into our firewall policies on various servers. This is an attempt to make consistent policies across all our machines. The main changes are to the proxy, all proxies get ufw now, but i tried to do some cleanup in the main role as well. 
From a security standpoint the biggest difference is that this closes a loophole by which one does not need VPN access to ssh into our india machines. It is now possible to ssh into the india proxy while not on the vpn, since it has a real public ip. If agent forwarding is on, one can then ssh into any of our other india machines, all without needing to get on the VPN. This fixes that by only allowing ssh traffic into the proxy on the private interface

ansible/roles/ufw/tasks/main.yml
czue: yeah I think this was just paranoia

czue:
good find, @calellowitz. this looks good to me. any additional testing you think we need to do or good to ship? just being a little cautious since locking ourselves out of servers is not fun.

calellowitz:
I deployed to staging yesterday and can still ssh into the proxy. That is the only machine that should change because the rest of the india machines already have ufw running from the other role. The only real concern I have is if we are using other ports on prod that we dont use on monoliths for machines in the cluster to connect to the proxy, and i close those by mistake, but staging appears to be up and running. I can check the ufw logs there as well to make sure it isnt dropping packets that look important.

ansible/roles/ufw/tasks/proxy_ufw.yml
```diff
@@ -18,6 +17,15 @@
...
   when: is_monolith
 
+- name: allow only private ssh traffic
+  ufw:
+    rule: allow
+    direction: in
+    port: 22
+    proto: any
+    interface: '{{ ufw_private_interface }}'
+  when: not is_monolith|bool
```
calellowitz: for future reference and to avoid others having the same nightmare i had `not is_monolith|bool` is necessary here because both `is_monolith` and `not is_monolith` evaluate to `False`, likely because of implicit type coercions regarding strings, but explicit coercion to a bool fixes it.

benrudolph:
nice looks good to me

benrudolph:
let me know when you're happy with the prod port testing and i can merge

calellowitz:
ufw logging appears to have stopped at midnight UTC but it looks like it was blocking internal connections with the db machine based on the logs we do have. is there any objection to me opening non monolith proxies up to all communications on the private interface. the other machines are already like that and this way it wont block attempts to communicate with rest of the cluster.

calellowitz:
@czue @benrudolph updated to allow all traffic on the private interface which seems less likely to cause unforeseen problems and align proxy settings with those of the other machines in the cluster

benrudolph:
yeah agree with that. i think that's an assumption we make a lot

calellowitz:
in that case, ready to merge from my end

PR #489, opened 2016-08-04
creating cleanup playbook and cleaning up shared dir

calellowitz:
@benrudolph cc: @snopoke 
This cleans up after the shared directory move that happened a week ago. It unmounts the old shared directory from machines that used to have it, removes it from the list of mountable drives on db0 and deletes the directory.
In theory, the cleanup playbook can be added upon in the future to cleanup after other tasks that change state in ways that have side effects, such as changing drive locations, removing old volumes, etc.


benrudolph:
any reason you decided not to make a cleanup role for this? isn't one advantage of having a separate playbook/tasks for cleanup is that we can remove that code from cluttering our main roles?

benrudolph:
👍 on the logic though

calellowitz:
The idea was that the separate playbook `deploy_cleanup` kept it out of our main playbooks, and it was in its own playbook in `shared_dir`, not `main.yml`, but it is still code relating to the shared directory so it made sense to be in the same place. Im not attached to that idea at all if you think it makes more sense to have it totally separate, i can make an entirely new cleanup role.

benrudolph:
ok sounds good, there are pros and cons to both. in the current setup we can reuse variables that have been declared in that role which is nice. i think it's possible in the future we may find it nice to have all cleanup in one area, but don't feel too strongly about it now. happy to merge this and see how it grows as we need it more

PR #492, opened 2016-08-08
fixing cas cert ref

calellowitz:
@snopoke 
I think this was just an oversight when certs were pulled out of ansible vault, but it was breaking `deploy_proxy`


snopoke:
Thanks

PR #493, opened 2016-08-08
Add softlayer HQ_PRIVATE_KEY

emord:
http://manage.dimagi.com/default.asp?234436#1198429

@gcapalbo should we wait for @ctsims on this one?

cc @czue


gcapalbo:
(probably fine to wait for confirmation in the ticket)

PR #494, opened 2016-08-08
Fix git hook

emord:
@benrudolph this should fix the git hook

cc @czue

### DIFF
--- a/git-hooks/pre-commit.sh
+++ b/git-hooks/pre-commit.sh
@@ -16,7 +16,7 @@ cr='
 '
 for f in $(git diff --cached --name-only | grep -E $FILES_PATTERN)
 do
-  MATCH=`grep --invert-match --no-messages $REQUIRED $f | head -n1`
+  MATCH=`grep -L $REQUIRED $f | head -n 1`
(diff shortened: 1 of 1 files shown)


git-hooks/pre-commit.sh
```diff
@@ -16,7 +16,7 @@ cr='
 '
 for f in $(git diff --cached --name-only | grep -E $FILES_PATTERN)
 do
-  MATCH=`grep --invert-match --no-messages $REQUIRED $f | head -n1`
+  MATCH=`grep -L $REQUIRED $f | head -n 1`
```
emord:
```
       -L, --files-without-match
              Suppress normal output; instead print the name of each input file from which no output would normally have been printed.  The scanning will stop on the first match.
```

emord: So I think that this looks for the required text `ANSIBLE_VAULT` in the file `$f` and if it isn't found in a file then it prints the file's name.

Previously it was returning file lines, so it was returning every line except for the first one of the vault file

benrudolph:
👍 thanks

PR #497, opened 2016-08-11
De-authorize ansible public key

millerdev:
@dannyroberts @snopoke cc @esoergel @NoahCarnahan

### DIFF
--- a/ansible/vars/production/production_public.yml
+++ b/ansible/vars/production/production_public.yml
@@ -35,7 +35,6 @@ primary_ssl_env: "production"
 dev_group: 'dimagidev'
 dev_users:
   present:
-    - ansible
     - biyeun
     - brudolph
     - cellowitz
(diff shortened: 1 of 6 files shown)


ansible/vars/production/production_public.yml
```diff
@@ -35,7 +35,6 @@ primary_ssl_env: "production"
 dev_group: 'dimagidev'
 dev_users:
   present:
-    - ansible
```
millerdev: Would be nice to move `dev_group` and `dev_users` into a single vars file so they don't have to be duplicated on each environment. I didn't look closely, but guessing they're all the same.

snopoke: They are all the same so we could do it. It would mean you'd have to add an extra command line arg with that vars file.

dannyroberts:
👍

ansible/vars/production/production_public.yml
millerdev: What do you think about putting it in `group_vars/all.yml`? Looks like [`dev_group` is already in there](https://github.com/dimagi/commcarehq-ansible/blob/30a9104cffbfc3d7a8de9858485742b57f5298ec/ansible/group_vars/all.yml#L22).

snopoke: Makes sense

millerdev: https://github.com/dimagi/commcarehq-ansible/pull/500

PR #498, opened 2016-08-12
bump kafka log retention to 4 weeks

czue:
@benrudolph @snopoke I think this is safer than the current 1 week


czue:
i double checked the current disk usage on prod and it doesn't look like it will be a major issue

snopoke:
👍

benrudolph:
nice agree this is safer

czue:
(just rolled this out to prod and india)

PR #499, opened 2016-08-12
Ce/pg dump

calellowitz:
The old cron job wasnt being run, which i think had to with `/etc/crontab` not being reloaded but im not sure. This cron method works, and is the same as we use for the purging of the cache files on the shared drive. I also cut down how long we keep the weekly backups to free up some space.
@czue @snopoke

### DIFF
--- a/ansible/roles/pg_standby/templates/create_postgres_dump.sh.j2
+++ b/ansible/roles/pg_standby/templates/create_postgres_dump.sh.j2
@@ -3,7 +3,10 @@ BACKUP_TYPE=$1
 DAYS_TO_RETAIN_BACKUPS=$2
 
 TODAY=$(date +"%Y_%m_%d")
-pg_basebackup -D "{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}" -Ft -Xs -z
+DIRECTORY="{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}"
+pg_basebackup -D $DIRECTORY -Xs
+tar -cjf "${DIRECTORY}.tar.bz" $DIRECTORY
+rm -rf $DIRECTORY
 
 # Remove old backups of this backup type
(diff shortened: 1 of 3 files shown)


ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -191,13 +191,14 @@ vacuum_cost_delay = 20ms        # 0-100 milliseconds
...
-wal_level = hot_standby
 archive_mode = on
 archive_command = 'rsync -a %p postgres@{{ hot_standby_server }}:{{ archive_dir }}/%f'   # archive directory lives on first standby machine specified in the inventory file
 
-max_wal_senders = 3
 {% endif %}
 
+wal_level = hot_standby
+max_wal_senders = 3
+
```
calellowitz: This allows us to use the same backup scripts on clusters that have a standby and ones that don't, and according to postgres docs, raising the `wal_level` should not have an impact on performance.

ansible/roles/pg_standby/templates/create_postgres_dump.sh.j2
```diff
@@ -3,7 +3,10 @@ BACKUP_TYPE=$1
 DAYS_TO_RETAIN_BACKUPS=$2
 
 TODAY=$(date +"%Y_%m_%d")
-pg_basebackup -D "{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}" -Ft -Xs -z
+DIRECTORY="{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}"
+pg_basebackup -D $DIRECTORY -Xs
+tar -cjf "${DIRECTORY}.tar.bz" -C "${DIRECTORY}" .
+rm -rf $DIRECTORY
```
calellowitz: The backup process now does a filesystem level dump that is then compressed, and the original dump is removed. This does mean we always need as much free space as the size of the db, but the db would have to more than double its current size before this starts to become a problem. I switched from gzip to bzip2 because it gives a higher compression ratio and while it is slower, the speed seemed less important.

PR #504, opened 2016-08-15
changing playbook to support the cas proxy removal from proxy group

calellowitz:
This pr should make the cas proxy no longer dependent on the proxy role. Then we can take it out of the proxy group, so we dont install the cchq site, and the bug where it has 2 default servers, prevent nginx from starting will be fixed.
@benrudolph cc: @snopoke

### DIFF
--- a/ansible/deploy_common.yml
+++ b/ansible/deploy_common.yml
@@ -5,13 +5,13 @@
     - common
 
 - name: ufw (firewall)
-  hosts: '!proxy'
+  hosts: '!proxy:!cas_proxy'
   sudo: yes
   roles:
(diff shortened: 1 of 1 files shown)


ansible/deploy_common.yml
```diff
@@ -5,13 +5,13 @@
     - common
 
 - name: ufw (firewall)
-  hosts: '!proxy'
+  hosts: '!proxy:!cas_proxy'
```
benrudolph: does this work if cas_proxy is not defined in the inventory?

calellowitz:
@benrudolph 
yeah, tested on staging and there is no issue if you specify a group that isnt in the inventory. It just does not match any hosts, or in this case it matches all of them.

PR #509, opened 2016-08-23
disabling challenge response authentication

calellowitz:
@millerdev @emord cc: @sravfeyn 
still not sure how that line got into the config but this should make sure it is removed. did a little readability cleanup as well


millerdev:
I had tried setting `ChallengeResponseAuthentication no`, but I think I forgot to restart sshd so it didn't appear to fix the issue. Thanks for following up on this.

PR #513, opened 2016-08-31
Route formplayer requests based on sessionid

benrudolph:
@wpride @dannyroberts ok so here's one option we talked about. Not positive it'll work, but this'll require formplayer to make all its requests like `/formplayer/my-domain/my-username` then it'll get proxied to the the `formplayer` upstream and be balanced with `hash $request_uri consistent` which will then hash based on domain and username. _hopefully_ this'll mean that data is divided equally among all servers that we have. and we can then avoid using the nfs and store directly on `/opt/data`, the encrypted drive

### DIFF
--- a/ansible/roles/nginx/templates/site.j2
+++ b/ansible/roles/nginx/templates/site.j2
@@ -5,7 +5,7 @@
 {% for balancer in item.server.get('balancers', []) %}
 upstream {{ balancer.name }} {
   {% if balancer.get('method') %}
-  {{ balancer.method }}
+  {{ balancer.method }};
   {% else %}
   least_conn;
   {% endif %}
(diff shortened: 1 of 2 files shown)


benrudolph:
https://github.com/dimagi/commcare-hq/pull/13080

benrudolph:
waaayyy simpler to just hash based on the sessionid in cookie. thanks @dannyroberts!

benrudolph:
✔️ works on staging

dannyroberts:
nice work! This is maybe the simplest "turn on scale" PR I've ever seen.

benrudolph:
haha it was a windy road!

dannyroberts:
OK, merging it! ⚡

PR #514, opened 2016-08-31
Add CDC motech server

sheelio:
New server is in Virginia, and I already configured the DNS for motech3.internal.commcarehq.org


benrudolph:
are you planning on getting interrupt to deploy proxy?

PR #516, opened 2016-09-01
Add local app to softlayer

NoahCarnahan:
This should not be merged and deployed until https://github.com/dimagi/commcare-hq/pull/13079 has been deployed.

Once this is deployed, the migrations in the local app "icds_reports" will be run

@orangejenny @kaapstorm


NoahCarnahan:
never mind, this app is already installed on all environments

PR #518, opened 2016-09-02
Revert "Revert "Add CDC motech server""

benrudolph:
Reverts dimagi/commcarehq-ansible#517
@sheelio no upstream was set for this, sorry i should have looked more closely. this should be deployed to staging first


benrudolph:
```
nginx: [emerg] host not found in upstream "motech3.internal.commcarehq.org" in /etc/nginx/sites-enabled/motech:44
```

benrudolph:
actually i'm entirely sure why this failed. i don't think there should be an upstream

benrudolph:
perhaps that machine just wasn't found. will have to investigate

PR #519, opened 2016-09-02
Add CDC MOTECH Server

sheelio:
This previously failed because the DNS wasn't configured correctly for motech3.internal.commcarehq.org.   This has now been fixed.

### DIFF
--- a/ansible/roles/nginx/vars/motech.yml
+++ b/ansible/roles/nginx/vars/motech.yml
@@ -20,3 +20,6 @@ nginx_sites:
      - name: /supply-dhis
        proxy_pass: http://motech-main.internal.commcarehq.org:8888/supply-dhis
        proxy_read_timeout: 180s
+     - name: /cdc-moz
+       proxy_pass: http://motech3.internal.commcarehq.org:8080/cdc-moz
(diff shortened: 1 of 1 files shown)


benrudolph:
tested on staging, deploying it to prod

ansible/roles/nginx/vars/motech.yml
```diff
@@ -20,3 +20,6 @@ nginx_sites:
      - name: /supply-dhis
        proxy_pass: http://motech-main.internal.commcarehq.org:8888/supply-dhis
        proxy_read_timeout: 180s
+     - name: /cdc-moz
+       proxy_pass: http://motech3.internal.commcarehq.org:8080/cdc-moz
```
biyeun: @sheelio I think this config is incorrect. If this is on virginia it's `internal-va`

dannyroberts: Uh, yeah confusing @biyeun but I think this is "correct" in the sense that this points to the right machine, it's just that the DNS @sheelio made didn't conform to the convention we've been using. I don't think it's a huge deal but we can easily change that.

PR #521, opened 2016-09-05
Formplayer fly on its own

benrudolph:
@wpride @dannyroberts this decouples formplayer from touchforms so we can scale it separately. adding label so i can do some testing with it


dannyroberts:
merge conflict

benrudolph:
ran the `deploy_formplayer.yml` playbook and there were no errors. i think this is good to go

czue:
:airplane:

PR #522, opened 2016-09-06
Add nginx to cchq sudo abilities

benrudolph:
@snopoke @dannyroberts this'll allow cchq to run `nginx` commands


dannyroberts:
I think I'm ok with that

PR #524, opened 2016-09-07
es snapshots

calellowitz:
@snopoke cc: @czue @benrudolph 
Here is the work for s3 backed es snapshots. It does a daily snapshot that it keeps for 3 days. Anything longer seemed excessive since it takes less than that to build most of these indices from scratch. Its currently on staging and the snapshots have been working but it has not been on there long enough to test deletion of old snapshots. Wanted to open for code review and assuming the deletion works tomorrow we can go forward.


ansible/roles/common_installs/tasks/main.yml
```diff
@@ -95,6 +95,13 @@
     - fabric
     - virtualenv
     - sh
+    - httplib2
+
+- name: Install elasticsearch python client
```
benrudolph: nit we may want to move this to the elastic search role since we only need it for those machines

benrudolph:
nice! looks good to me. 🔍 🏭

snopoke:
looks good

calellowitz:
i made a management command to simplify restoring from backup. pr is https://github.com/dimagi/commcare-hq/pull/13154

PR #525, opened 2016-09-07
Proxy to next server on 403 from Riak CS

millerdev:
Hopefully this will fix Riak node down causing error on client.

I've repeatedly observed errors like this in Riak CS access log when it's corresponding Riak service is down:
```
/var/log/riak-cs/access.log.2016_08_25_22:172.24.32.20 - - [25/Aug/2016:22:46:51 +0000] "PUT /buckets/blobdb/objects/commcarehq%2Fd<snip-production-doc-name>.xml.adf0e9982920427fb8353942c3828b58 HTTP/1.0" 403 309 "" "Boto3/1.2.3 Python/2.7.6 Linux/3.13.0-86-generic Botocore/1.3.30 Resource"
```
The cluster is usually still operational when this occurs, it (usually) just has one node down. This change should make the proxy try the request on the next node, which is hopefully not down and can service the request without causing the client to error out.

OOM killer has been the cause of Riak going down, and is the root cause/motivation for this workaround.

I'll deploy it after sanity check from code reviewer.

@dannyroberts cc @NoahCarnahan

### DIFF
--- a/ansible/roles/nginx/vars/riakcs.yml
+++ b/ansible/roles/nginx/vars/riakcs.yml
@@ -22,3 +22,6 @@ nginx_sites:
        proxy_read_timeout: 90
        proxy_buffer_size: 64k
        proxy_buffers: 8 64k
+       # http_403 from Riak CS is an indication that Riak is down
+       proxy_next_upstream: error timeout http_403
+       proxy_next_upstream_tries: 3


millerdev:
✅

NoahCarnahan:
Looks fine to me. Seems sort of weird that we are receiving a 403 when the node is down, so might be worth investigating why that is if you don't already know, but non-blocking from my perspective.

millerdev:
I think it's the standard the failure mode of Riak CS when it's Riak node is down. Can't find a good source for that though. Note: `riak-cs` is a separate service from `riak`, so it's not surprising that we're getting a valid HTTP response when `riak` is down but `riak-cs` is not, although 403 does seem a little odd for the "backend db is down" scenario.

dannyroberts:
Will be interesting to see whether this results in domino behavior as load redistributed causes a second node to also OOM and so on

PR #527, opened 2016-09-08
Fix noisy ansible managed comment

millerdev:
Comment was different for each user doing deploys causing annoying comment-only file changes. I would have liked to include the relative file path inside the ansible repo in the comment, but I couldn't find a way to format the `{file}` variable :(

@calellowitz @NoahCarnahan


benrudolph:
👍

PR #528, opened 2016-09-08
Change Motech 3 internal URL to internal-va

sheelio:
@biyeun 

Let me know if this looks right (already created the DNS entry for this)


dannyroberts:
@sheelio I think i came late to the game, but check my note here (https://github.com/dimagi/commcarehq-ansible/pull/519#discussion_r78074407) and see if you still think we need this change. My current guess is that the suggestion was a false alarm.

sheelio:
@dannyroberts 

Not a hard change to make, and I already setup the DNS to point motech3.internal-va.commcarehq.org to the correct machine.

If the change is better long term, I'd move to merge this.

dannyroberts:
sweet. Thanks

PR #530, opened 2016-09-09
add riak_monitoring role for restarting riak

dannyroberts:


dannyroberts:
Going to run this first

dannyroberts:
OK, ran this on all the machines and it seems to work!

ansible/roles/riak_monitoring/templates/riak-restarter.sh.j2
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+# set the path so that service can be found
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games
+result=`service riak status`
```
millerdev: the conventional way to do that is to explicitly call `/usr/sbin/service` instead of using `PATH`

ansible/roles/riak_monitoring/templates/riak-restarter.sh.j2
```diff
@@ -0,0 +1,14 @@
...
+date=`date`
+expected='pong
+riak is running'
+if [ "$result" = "$expected" ]; then
+    echo "riak is up as of $date" > status.txt
+else
+    service riak start > service-output.txt 2> service-stderr.txt
+    echo "server was restarted at $date" >> restart.txt
```
millerdev: It's typically best to rotate log files rather than letting them grow forever. Do you think we should be concerned about that here?

PR #533, opened 2016-09-13
swiss https -> http

benrudolph:
forgot they don't have a cert @wpride @gcapalbo


benrudolph:
this kind of turned into a "fix" swiss session. it makes postgres connect to pgbouncer. it also properly changes the db hostname so that the postgres tasks aren't skipped over

PR #534, opened 2016-09-14
pg s3

calellowitz:
@snopoke @czue @benrudolph 
This sends our postgres backups to s3 so we are not tied to rackspace or in case something happens to our standby machine where they are stored locally. I have configured to the s3 bucket to delete them after 30 days but we can shorten that time period if necessary. This will replace our rackspace backups that used to be on prod, and have the same goal as the evault backups we do in India.

### DIFF
--- a/ansible/roles/pg_standby/tasks/main.yml
+++ b/ansible/roles/pg_standby/tasks/main.yml
@@ -45,7 +45,7 @@
     owner: postgres
     mode: 0700
     backup: yes
-  when: backup_postgres
+  when: postgres_s3
 
 - name: Create Daily Cron job
   sudo: yes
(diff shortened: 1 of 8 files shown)


benrudolph:
nice looks good to me. glad we're backing up somewhere outside of rackspace

czue:
> glad we're backing up somewhere outside of rackspace

agreed thanks for this. we can all sleep slightly safer at night

ansible/roles/pg_standby/tasks/main.yml
```diff
@@ -36,6 +36,23 @@
...
+    name: tinys3
+    version: 0.1.11
+
+- name: Copy s3 upload script
+  sudo: yes
+  template:
+    src: "backup_snapshots.py.j2"
+    dest: "/etc/cron.d/backup_snapshots.py"
+    group: postgres
+    owner: postgres
```
snopoke: Does this need to be owned by postgres?

PR #535, opened 2016-09-19
Back to localhost for swiss

benrudolph:
@gcapalbo this returns PG to working with localhost. also fixes cloudcare to properly request localhost since we don't have a cert on swiss, the call to the host would always fail

http://manage.dimagi.com/default.asp?237630


gcapalbo:
:+1: on :white_check_mark:

PR #536, opened 2016-09-21
Ce/pg s3

calellowitz:
@gcapalbo @benrudolph 
Final fixes for postgres s3 uploads. These have run successfully on prod. S3 has a size limit for individual file uploads that we were hitting on prod but not in my testing on staging, but allows multipart uploads. Boto3 handles that automatically while tinys3 was just throwing errors so i switched libraries but otherwise maintained the same functionality.


benrudolph:
👍

PR #537, opened 2016-10-04
Ce/india pg s3

calellowitz:
@czue @benrudolph @snopoke @millerdev cc: @gcapalbo 
This pr allows us to move our offsite postgres backups in India from evault to s3. Evault charges us significantly more per GB and we pay for a fixed amount of space whether or not we use all of it, plus any overages. Additionally its significantly more of a blackbox and therefore far harder to manage. After this PR we will have one offsite backup and restore system that is the same for both prod and softlayer


ansible/vars/softlayer/softlayer_public.yml
```diff
@@ -62,7 +62,9 @@ keystore_file: '../config/DimagiKeyStore'
 backup_blobdb: False
 backup_postgres: True
 backup_es: True
-postgres_s3: False
+postgres_s3: True
+
+aws_region: 'ap-south-1'
```
calellowitz: this is the amazon mumbai data center

calellowitz:
@millerdev 
Backups are encrypted on disk at amazon (by amazon) and transmitted over an encrypted connection but we dont send encrypted data.
As far as HIPAA compliance, amazon seems to indicate that compliance exists on a service level, making no distinction between the regions and s3 is on the list of compliant services but I can double check with Kevin to make sure our baa covers all regions.

calellowitz:
spoke with kevin, our BAA is not region specific at all so we should be good.

czue:
thanks!

PR #540, opened 2016-10-10
track function calls in postgresql

snopoke:
@sravfeyn 

https://www.postgresql.org/docs/9.4/static/runtime-config-statistics.html#GUC-TRACK-FUNCTIONS


benrudolph:
nice

PR #541, opened 2016-10-10
adding archive timeout

calellowitz:
@czue @snopoke @benrudolph 
It looks like the standby is lagging behind the master server (at least in the afternoon in boston) probably due to the low volume of writes. This forces the master to send an update to the standby ever 60 seconds even if the WAL is not filled which should cut down lag to a minute at most.


benrudolph:
sweet

snopoke:
@calellowitz I assume you'll roll this out

PR #545, opened 2016-10-12
move elastic datadog integration to ES machines

snopoke:
We're already running the agent on the ES machines. This just puts the ES integration on those machines instead of on the PG machine. This should make the builtin dashboards work better: https://app.datadoghq.com/screen/integration/elasticsearch


snopoke:
@ben / @calellowitz 
cc @kaapstorm

benrudolph:
nice this is sweet. i was looking at the actual dashboard too, do you know why it has queries from hqdb0? it also says our disk is 86% full which seems kind of high

snopoke:
It lists hqdb0 because we were running the integration on hqdb0. I think
that will stay there until there is no data from hqdb0 in the time window.

Disk space does seem a high though it's ~22GB per node so probably not
serious for now.

benrudolph:
ah gotcha. should've realized it would take time to propagate the changes to the dashboard

PR #546, opened 2016-10-13
specifying endpoint in es snapshot setup

calellowitz:
@snopoke @czue @benrudolph 
It turns out that the aws plugin for elasticsearch has a hardcoded list of supported regions which does not include the mumbai datacenter, so to get it to work in india i had to specify an endpoint instead of a region. (This is deployed and verified that the snapshots are succeeding)


benrudolph:
nice

PR #547, opened 2016-10-13
Add Pathfinder LAD server

sheelio:
This adds the entry for Pathfinder LAD motech server

@NoahCarnahan @emord @snopoke


emord:
curious, what does pi stand for? (would have expected pf)

sheelio:
Ah - they prefer the initials PI (for Pathfinder International).

PR #551, opened 2016-10-14
adding restore command for pg s3 backups

calellowitz:
@snopoke @czue @benrudolph 
This provides scripts to allow 1 command for restoring postgres from a backup stored in s3.

### DIFF
--- a/ansible/roles/pg_standby/templates/restore_from_backup.sh.j2
+++ b/ansible/roles/pg_standby/templates/restore_from_backup.sh.j2
@@ -1,0 +1,28 @@
+#!/bin/bash
+BACKUP=$1
+
+read -p 'Are you sure you want to restore postgres to a past state [y/N]: ' -r confirm
+if [ $confirm != 'y' ]; then
+    exit 1
+fi
+read -p 'This will delete the current db. Are you SURE you are sure [y/N]: ' -r confirm2
+if [ $confirm2 != 'y' ]; then
+    exit 1
+fi
+
+echo 'retrieving backup from amazon'
+/root/retrieve_backup.py "${BACKUP}" &&
+
+echo 'extracting backup' &&
+tar -xjf "${BACKUP}" -C "{{ postgresql_backup_dir }}" &&
+
+echo 'stopping postgresql' &&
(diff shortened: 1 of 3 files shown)


ansible/roles/pg_standby/templates/restore_from_backup.sh.j2
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+BACKUP=$1
+
+read -p 'Are you sure you want to restore postgres to a past state [y/N]: ' -r confirm
+if [ $confirm != 'y' ]; then
+    exit 1
+fi
+read -p 'This will delete the current db. Are you SURE you are sure [y/N]: ' -r confirm2
```
benrudolph: 😨

benrudolph: i hope to never have to run this command

ansible/roles/pg_standby/templates/restore_from_backup.sh.j2
```diff
@@ -0,0 +1,28 @@
...
+/root/retrieve_backup.py "${BACKUP}" &&
+
+echo 'extracting backup' &&
+tar -xjf "${BACKUP}" -C "{{ postgresql_backup_dir }}" &&
+
+echo 'stopping postgresql' &&
+/etc/init.d/postgresql stop &&
+
+echo 'restoring from backup' &&
+rsync -avz --delete "{{ postgresql_backup_dir }}/${BACKUP}/" "{{ postgresql_home }}" &&
```
snopoke: I'd another confirmation here before we actually go ahead (to be paranoid)

ansible/roles/pg_standby/templates/restore_from_backup.sh.j2
calellowitz: Yeah, it sounds scary on purpose but realistically the most probable scenario for running this command is on a new machine if the main db (and standby for prod) is hosed. I dont really see this being run on a machine with existing data, because those machines have backups already on the machine.

PR #552, opened 2016-10-17
restart datadog agent after pulling git changes

snopoke:
@emord 

buddy @sravfeyn

### DIFF
--- a/ansible/roles/datadog/tasks/main.yml
+++ b/ansible/roles/datadog/tasks/main.yml
@@ -38,5 +38,6 @@
     accept_hostkey: yes
     update: yes
   sudo_user: "{{ cchq_user }}"
+  notify: restart datadog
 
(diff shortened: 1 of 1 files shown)


ansible/roles/datadog/tasks/main.yml
```diff
@@ -38,5 +38,6 @@
     accept_hostkey: yes
     update: yes
   sudo_user: "{{ cchq_user }}"
+  notify: restart datadog
```
emord: Does this always restart datadog, or only restarts it when there are new commits pulled?

snopoke: It should only do it if there are new changes that get pulled (though the docs don't mention anything about it examining the source confirms). Also not a huge issue to restart the datadog agent.

PR #555, opened 2016-10-18
cchq owns java

benrudolph:
@dannyroberts @wpride this changes the owner of the java installations to cchq instead of root. this is necessary for monitoring the formplayer application. the formplayer service is run by cchq, but in order to access JMX it needs to read a couple files within the java directory. if you don't know what jmx is (i didn't), it's the recommended way to implement monitoring in java applications. i'm still new to the technology and wrapping my head around it, but it's built into java, which is why the files are located in the java directory. from the docs:

> The JMX specification defines an architecture, the design patterns, the APIs, and the services for application and network management and monitoring in the Java programming language.

> Using JMX technology, a given resource is instrumented by one or more Java objects known as Managed Beans, or MBeans. These MBeans are registered in a core managed object server, known as an MBean server, that acts as a management agent and can run on most devices enabled for the Java programming language.

it's mostly managed by the java framework we're using for formplayer and there's almost no additional code we need to write at this time. see this other PR as well: https://github.com/dimagi/commcare-hq-deploy/pull/134

### DIFF
--- a/ansible/requirements.txt
+++ b/ansible/requirements.txt
@@ -2,4 +2,4 @@ ansible==1.9.6
 netaddr
 dnspython
 passlib
-cryptography
+cryptography==1.4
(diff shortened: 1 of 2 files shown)


dannyroberts:
Restarted the build. It ended on an error message that was something like "I'm sorry but this version only supports 100 groups". (I ought to have copied it before restarting.)

benrudolph:
build failure most likely due to the fact that we don't version our crypto library: https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/requirements.txt#L5

benrudolph:
✅ froze crypto package at the one thta was installed on the production ansible machine

dannyroberts:
Excellent work

PR #561, opened 2016-11-01
use correct host for es curl commands

snopoke:
ES is configured to bind to it's IP so curls to `localhost` don't work
@benrudolph  @emord  @esoergel 
buddy @biyeun


benrudolph:
👍 this probably won't work for our monolith setup, but i think that's fine since we don't need a rolling restart for monolith

snopoke:
I think it should since this uses the same value as get's used in the ES
config file.

benrudolph:
ok cool sounds good

PR #565, opened 2016-11-08
Clean up MOTECH servers

sheelio:
This removes the old test MOTECH server and adds another MOTECH server for the possible health project

@biyeun @mkangia to review and deploy.


sheelio:
I've already removed the corresponding server from Rackspace and the DNS

czue:
@sheelio wonder if it's worth getting you setup on ansible so you can deploy on your own?

emord:
> I've already removed the corresponding server from Rackspace and the DNS

@sheelio next time can you wait on doing that until after an ansible deploy? It blocks the normal deploys because nginx checks if those servers are up

sheelio:
@emord Sorry lesson learned.  

@czue Hoping that setting up these servers doesn't last for more than one or two more projects (and we move to MOTECH v2 that's multi-tenant) :)

snopoke:
Is that issue documented somewhere? Could be added to the devops team's
todos.

emord:
:+1: added it

benrudolph:
is the proposal to remove the `nginx -t` check during deploy? i actually think it's a fairly nice thing to have, it's just weird that `nginx` fails hard when those servers aren't available. maybe there's a way to configure that

benrudolph:
jk, realize the proposal was to clean up nginx sites. +1000

PR #567, opened 2016-11-09
check nginx -t

emord:
@nickpell @czue 

adapted from https://github.com/jdauphant/ansible-role-nginx/blob/cc5cb82c5742d5a442b98c8ca2515459e75a07d1/handlers/main.yml#L17-L22

buddy @esoergel


czue:
nice, thanks

PR #569, opened 2016-11-15
add script template for transferring export migration logs to s3

biyeun:
@calellowitz @benrudolph 

This isn't complete yet as I'm not sure how to get this properly deployed with ansible. This script worked when running manually from the postgres user on hqdb0.


benrudolph:
actually don't really think this is necessary. this is very much a one off. we could add it to some sort of cleanup role and allow for arbitrary files to uploaded, but don't think it's worth it

biyeun:
@benrudolph thanks! will go ahead and close

PR #570, opened 2016-11-15
Set auto_create_index to false

esoergel:
I deployed this on staging already.  Tested it out by attempting to submit a doc via curl:
```
$ curl -XPUT 'http://es-server.example.com:9200/fakeindex/type1/1?pretty' -d'
{
    "test" : "Ethan is testing"
}'
```
This succeeded before the change, and fails after with an IndexMissingException.  I did confirm that we are still able to create indices the way we normally do (in python, `es.indices.create(index="fakeindex")`).

The rolling restart didn't work for me on staging (I got a "failed to connect" error) - I had to manually call the start command.  For this reason I'll do the restart for prod in the office while pairing just to make sure we minimize ES downtime.

@calellowitz @dannyroberts @snopoke


snopoke:
Re the rolling restart not being able to connect - the restart script attempts to connect from the ES machine using the host name in the inventory file (this was changed recently).

czue:
nice, great to finally have this in

PR #571, opened 2016-11-16
Revert "Remove amazon keys from staging"

esoergel:
Reverts dimagi/commcarehq-ansible#564
@calellowitz @emord 

```
TASK: [{{ role }} | Copy Elasticsearch Config] ******************************** 
fatal: [hqes0-staging.internal-va.commcarehq.org] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'dict object' has no attribute 'AMAZON_S3_ACCESS_KEY'", 'failed': True}
```
Reverting because this causes the elasticsearch role to fail on staging.  I'm guessing we need a smarter check or something, but I'm already like 3 levels removed from the change I'm actually trying to make so I'm gonna punt that.


esoergel:
actually, I'm not sure how to handle this, I'm just comment this out for now.  @calellowitz can you follow up to make sure this variable is either defined for all envs or not needed?

PR #578, opened 2016-11-25
host tableau at /reporting/ hq url suffix for icds env

sravfeyn:
@calellowitz @czue 

The dirty hack to tableau hosting at `icds-cas.gov.in/reporting/`.


czue:
looks good on a glance but i'll let @calellowitz review more closely

calellowitz:
Looks good to me. I have some stylistic questions but given the temporary nature of this hack I'm not sure they matter

PR #581, opened 2016-12-08
activate preview A/B test by setting ONBOARDING_DOMAIN_TEST_DATE

biyeun:
@benrudolph


ansible/vars/production/production_public.yml
```diff
@@ -204,6 +204,7 @@ localsettings:
   MAPS_LAYERS: "{{ localsettings_private.MAPS_LAYERS }}"
   MIA_THE_DEPLOY_BOT_API: "{{ localsettings_private.MIA_THE_DEPLOY_BOT_API }}"
   MOBILE_INTEGRATION_TEST_TOKEN: "{{ localsettings_private.MOBILE_INTEGRATION_TEST_TOKEN }}"
+  ONBOARDING_DOMAIN_TEST_DATE: (2016, 12, 8)
```
benrudolph: just checking that this will be a python string, not a tuple when it's put in the localsettings

biyeun: EXCELLENT CATCH THANK YOU. Updated.

biyeun:
@benrudolph any idea if we're good to deploy this?

benrudolph:
yep we're good :running_woman:

PR #582, opened 2016-12-12
add password protected location for files

snopoke:
@czue not sure if this should get merged into the main repo or not.


czue:
yeah not sure either. this just exposes that endpoint with an empty listing for servers that haven't had anything uploaded?

snopoke:
For servers that don't have any data in the location and no `.htpasswd` file nginx will return a 502 Bad Gateway

PR #585, opened 2016-12-27
Misc fixes

benrudolph:
anyone. no matter what, there are always ansible error on a deploy_stack

### DIFF
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -93,7 +93,6 @@
   pip: name={{ item }}
   sudo: yes
   with_items:
-    - fabric
     - virtualenv
     - sh
     - httplib2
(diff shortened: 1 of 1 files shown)


ansible/roles/common_installs/tasks/main.yml
```diff
@@ -92,7 +93,6 @@
   pip: name={{ item }}
   sudo: yes
   with_items:
-    - fabric
```
benrudolph: @dannyroberts do you know why this is here? i don't think this is used anywhere so figured i'd remove

dannyroberts: Yeah I'm really not sure. I traced it back all the way to the first commit of the repo https://github.com/dimagi/commcarehq-ansible/commit/699d5d945b706db23a55327b8c20d668b259ef3c#diff-4d74503b386f59bf023b93f3dd8d5a3cR18, that's been moved around a dozen times to wind up where it is now, and was originally added by someone we very randomly contracted. My guess is that if you move it, nothing will break. It certainly wasn't added there for a specific purpose by a dimagi employee recently (or ever).

benrudolph: ah! that diff shows the deploy script `fabfile.py`. i didn't know we used to keep that in here. guessing that's where that requirement came from. based on that i think we can actually get rid of `sh` too, not too big of a deal though

PR #589, opened 2017-01-10
Je/enikshay proxy

emord:
https://github.com/dimagi/commcare-hq-deploy/pull/170

This sets up the proxy for uatbc. I wanted to make this a bit more general since it's basically the same code for CAS and UATBC, but I decided not to when I realized our config uses globals for the certs

@proteusvacuum @czue


czue:
nice. yeah i don't think we need full site functionality.

PR #592, opened 2017-01-17
Configure leveldb backend for riak

millerdev:
This will eventually be rolled out to all Riak clusters. #590 and #591 makes it safe to have the new configuration in ansible even though it is different from the effective configuration on nodes before they are migrated to leveldb backend.

@snopoke


ansible/roles/riakcs/configure/tasks/main.yml
```diff
@@ -68,16 +61,6 @@
...
-  template: >
-    src={{ extra_conf }}.j2
-    dest={{ _extra_conf.stat.path }}
-    owner=root
-    group=riak
-    mode=0640
-  notify: restart {{ name }}
-  when: extra_conf != "" and (_force_riak_config or not _extra_conf.stat.exists)
-
 - name: Drop {{ name }} {{ drop_conf }}
```
millerdev: This drop extra conf task snuck into an earlier PR by accident. It should have been added in this PR.

kaapstorm:
:+1: Thanks @millerdev

PR #595, opened 2017-01-20
Ce/swiss cert

calellowitz:
change to new swiss certs (already deployed just adding to codebase)
@sravfeyn @emord @snopoke

### DIFF
--- a/ansible/vars/swiss/swiss_public.yml
+++ b/ansible/vars/swiss/swiss_public.yml
@@ -37,10 +37,10 @@ elasticsearch_memory: '4096m'
 elasticsearch_cluster_name: 'deves'
 #elasticsearch_node_name: '??'
 
-apache2_cert_file: '../config/{{ deploy_env }}/ssl/server.crt'
-apache2_key_file: '../config/{{ deploy_env }}/ssl/server.key'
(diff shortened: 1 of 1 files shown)


ansible/vars/swiss/swiss_public.yml
```diff
@@ -37,10 +37,10 @@ elasticsearch_memory: '4096m'
 elasticsearch_cluster_name: 'deves'
 #elasticsearch_node_name: '??'
 
-apache2_cert_file: '../config/{{ deploy_env }}/ssl/server.crt'
-apache2_key_file: '../config/{{ deploy_env }}/ssl/server.key'
+apache2_key_file: '../config/{{ deploy_env }}/ssl/swiss.commcarehq.org.key'
+apache2_cert_file: '../config/{{ deploy_env }}/ssl/swiss.commcarehq.org.crt'
```
emord: I don't think this is used anymore? It should just be the nginx file I think

emord: Looks like that variable is in all the var files, but isn't used anymore

benrudolph: confirming apache is no longer used

PR #598, opened 2017-01-26
newrelic java

benrudolph:
@snopoke this install the newrelic java agent on our machines. just need to update our supervisor conf to point to it

cc: @emord


ansible/roles/common_installs/tasks/main.yml
```diff
@@ -60,6 +60,7 @@
     - ranger  # cmdline file browser
     - libffi-dev  # Required for installing github3py
     - tmux
+    - unzip
```
snopoke: :+1:

benrudolph:
this is g2g

PR #601, opened 2017-02-02
Add Perm Gen bean

benrudolph:
@snopoke @wpride according to the datadog [docs](https://help.datadoghq.com/hc/en-us/articles/207874586-Collecting-Composite-type-JMX-attributes) this is how you add a composite data bean... i'm not confident this will work, but going to try and play around with it since it'd be really nice to have

cc: @emord


benrudolph:
nice it worked!

PR #606, opened 2017-02-15
Initial setup for fp load testing

benrudolph:


benrudolph:
don't need this as we are going to load test from our local machines for now

PR #607, opened 2017-02-16
add iotop to default installs

snopoke:
@dimagi/devops


snopoke:
Very useful for figuring out what's misbehaving

PR #608, opened 2017-02-17
add sentry api key to localsettings

snopoke:
cross: https://github.com/dimagi/commcare-hq/pull/15022


snopoke:
ah, yea, so painful

PR #609, opened 2017-02-17
Use riakcs swap for postgres too

kaapstorm:
Would this work, @benrudolph @snopoke ? (I have not tested this!)

### DIFF
--- a/ansible/deploy_db.yml
+++ b/ansible/deploy_db.yml
@@ -72,6 +72,14 @@
 
 - include: deploy_riakcs.yml
 
+- name: Configure swap
+  hosts:
+    - riakcs
+    - postgresql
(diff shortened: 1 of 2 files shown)


kaapstorm:
Yes.

benrudolph:
my preference for this would be to just selectively add swap to the machines we need. don't think it's great to rely on swap. that said not super convinced it matters that much

snopoke:
+1 to only have this on machines we want. I haven't looked closely but this might only get applied to hosts with the `swap_size` var set (which gets set in inventory).

emord:
closing this out because https://github.com/dimagi/commcarehq-ansible/pull/706 likely supersedes it

PR #612, opened 2017-02-23
nginx config when fake_ssl_cert

kaapstorm:
When fake_ssl_cert is true, nginx_ssl_(cert|key) won't be copied, resulting in an invalid nginx config. Instead of copying fake SSL certs, just point nginx at the ones it already has.

@snopoke cc @NoahCarnahan


kaapstorm:
@millerdev, you might have an opinion in this approach too.

snopoke:
seems ok to me

PR #613, opened 2017-02-23
Document DimagiKeyStore

kaapstorm:
DimagiKeyStore should be mentioned in the docs before the user is told to run a full deployment.

"Obtain a copy" is pretty vague. Open to suggestions.

@millerdev cc @NoahCarnahan


millerdev:
👍 I don't have any suggestions for better wording right now

PR #617, opened 2017-02-24
Upgrade Ansible to 2.2.1

kaapstorm:
:tropical_fish: The first commit just surrounds variables in quotation marks. The interesting stuff happens after that. 

This started out as a need to use a `block` clause. The `block` that began this is in the last commit. pip requirements were not being installed because cloning HQ would cause the `when` condition always to be false.

@benrudolph @snopoke cc @millerdev @NoahCarnahan 

(Hold off until I've tested this One More Time.)


ansible/requirements.txt
```diff
@@ -1,4 +1,4 @@
-ansible==1.9.6
+ansible==2.2.1
```
benrudolph: 😂

ansible/roles/commcarehq/tasks/main.yml
```diff
@@ -28,37 +28,37 @@
...
+    - name: install pip requirements
+      sudo_user: "{{ cchq_user }}"
+      with_items: commcarehq_requirements
+      pip:
+        requirements: "{{ code_home }}/requirements/{{ item }}"
+        virtualenv: "{{ virtualenv_source }}"
+      tags:
+        - slow
+      when: not testing|default(False)
+  when: not gitdir.stat.exists
```
benrudolph: nice clean up

benrudolph:
test failure real
```
fatal: [localhost]: FAILED! => {"failed": true, "msg": "Failed to set permissions on the temporary files Ansible needs to create when becoming an unprivileged user (rc: 1, err: chown: changing ownership of `/tmp/ansible-tmp-1487950995.05-21792329834171/': Operation not permitted\nchown: changing ownership of `/tmp/ansible-tmp-1487950995.05-21792329834171/git.py': Operation not permitted\n). For information on working around this, see https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user"}
```

millerdev:
Looking good. I'm very happy to see this upgrade \o/

ansible/ansible.cfg
```diff
@@ -1,2 +1,5 @@
 [defaults]
 ansible_managed = Ansible managed, do not edit directly
+
+# Added to allow HQ to be cloned as a non-privileged user in Ansible >2.0.
+allow_world_readable_tmpfiles = True
```
kaapstorm: This is what Ansible was silently doing up to now, and what you have to explicitly allow after Ansible 2.0. It seems OK to me, but it might be better to clone HQ and install pip requirements as root, and then change ownership to the cchq user afterwards. It seems the only affected tasks are ansible/roles/commcarehq/tasks/main.yml.

ansible/ansible.cfg
```diff
@@ -1,5 +1,6 @@
 [defaults]
 ansible_managed = Ansible managed, do not edit directly
+role_path = /opt/ansible_roles
 
 # Added to allow HQ to be cloned as a non-privileged user in Ansible >2.0.
 allow_world_readable_tmpfiles = True
```
kaapstorm: This is what Ansible was silently doing up to now, and what you have to explicitly allow after Ansible 2.0. It seems OK to me, but it might be better to clone HQ and install pip requirements as root, and then change ownership to the cchq user afterwards. It seems the only affected tasks are ansible/roles/commcarehq/tasks/main.yml.

kaapstorm:
@benrudolph, re testing, I've run through deploy_stack.yml on Vagrant.

benrudolph:
awesome! excited to have it in

ansible/ansible.cfg
millerdev: +1 for clone as root and then `chown cchq`. It seems better to not have this enabled as we're developing new code because we may inadvertently end up depending on it in places where we should not.

kaapstorm:
@javierwilson That last commit should be what we need to clone HQ as part of `deploy_stack`. There are other files that also sudo as non-privileged user:

  - [ ] ansible/roles/captain/tasks/igor.yml
  - [ ] ansible/roles/captain/tasks/main.yml
  - [ ] ansible/roles/commcarehq/tasks/main.yml
  - [ ] ansible/roles/datadog/tasks/main.yml
  - [ ] ansible/roles/load_test/tasks/install_host.yml
  - [ ] ansible/roles/load_test/tasks/install_prototype.yml
  - [ ] ansible/roles/load_test/tasks/install_runner.yml
  - [ ] ansible/roles/newrelic-java/tasks/main.yml
  - [ ] ansible/roles/newrelic/tasks/main.yml
  - [ ] ansible/roles/nginx/tasks/errorpages.yml
  - [ ] ansible/roles/pg_standby/tasks/main.yml
  - [ ] ansible/roles/postgresql/tasks/main.yml
  - [ ] ansible/roles/touchforms/tasks/main.yml

I haven't checked them yet.

kaapstorm:
@javierwilson, @benrudolph, @snopoke I went through those tasks that sudo as an unprivileged user, and they don't seem to be causing problems. This PR seems to be ready to merge. Is there anything you can think of that I might be missing?

benrudolph:
👍 our testing infrastructure is non existent at this point, so if you're able to run through a full stack deploy, i think that's good (and it works). there are lots of benefits to getting this merged, so i'm happy to merge and then incrementally fix minor things that aren't working

snopoke:
Finally!! Thanks @kaapstorm

millerdev:
🎉 🎉 🎉

PR #618, opened 2017-02-24
icds updates

calellowitz:
@snopoke @sravfeyn 
made some final updates to get the icds proxy deployed. have deployed based on this branch and all seems to be working fine.


ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -44,3 +44,43 @@ nginx_sites:
       balancer: "{{ deploy_env }}_formplayer/"
       proxy_next_upstream_tries: 1
       proxy_read_timeout: 900s
+    - name: /files
```
calellowitz: @snopoke i added all of these to the main nginx template because there doesnt seem to be a way to conditionally add locations. this does mean that they will be deployed to all the other environments but i do not think there is anything wrong with having them in site configs. I assume they were manually added the first time but if there was other ansible work with them done better i would be happy to change this.

ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -10,7 +10,7 @@ nginx_sites:
       port: "{{ formplayer_port }}"
       method: "hash $cookie_sessionid consistent"
    file_name: "{{ deploy_env }}_commcare"
-   listen: "443 ssl{{ ' default_server' if deploy_env == primary_ssl_env }}"
+   listen: "443 ssl{{ ' default_server' if deploy_env == primary_ssl_env else '' }}"
```
sravfeyn: Dirty, sorry. Thanks for the fix!

sravfeyn:
I will let Simon review the files part

ansible/roles/nginx/vars/cchq_ssl.yml
snopoke: yes they were manually added. Is there no way to have them site specific? Could we not put a new section into the icds vars file? Or at least have a `when` condition that prevents them from being added to the other envs?

calellowitz: it looks like based on our current templates, we dont have a way to conditionally add locations. I can try to change our template to allow that, something like `env_locations` that pulls from the vars file, and could be added to the locations list here. does that seem worth it?

snopoke: I think it's worth it to not contaminate the other envs with all these locations.

calellowitz:
@snopoke I took a stab at separating out those extra locations. To test, I did a proxy deploy with `--check` and the only thing that changed was ordering of the configuration within each location, which should be fine. Let me know if this is what you were looking for

calellowitz:
@emord removed second loop

snopoke:
Nice, thanks @calellowitz

PR #620, opened 2017-02-27
fix localsettings changes

snopoke:


sravfeyn:
Sorry about that. Thanks for fixing. I will be more cautious

PR #621, opened 2017-03-02
test out data dog metrics for processes

benrudolph:
Going to experiment with this a bit to see if it can help with our celery memory usage monitoring

http://docs.datadoghq.com/integrations/process/

I figured it might be useful to monitor all our processes, so I turned it on for all hosts. This initial version only captures celery processes


benrudolph:
deployed this to NIC, think it's fine to merge

PR #625, opened 2017-03-03
rate limit nginx

emord:
Tested this on staging with ethan also hitting the server.

This will rate limit based on the server name so "cas.commcarehq.org" (which I think is what all phones will be hitting) It doesn't seem necessary to block the main site as that's only used via the web correct? if that's necessary, It'll also take some refactoring as we use the same base yml file for prod, softalyer and icds

@snopoke 

@gcapalbo @benrudolph

### DIFF
--- a/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/ansible/roles/nginx/vars/cas_ssl.yml
@@ -13,6 +13,7 @@ nginx_sites:
       port: "{{ formplayer_port }}"
    limit_req_zones:
     - name: "server_name"
+      zone: "$server_name"
       size: "10m"
       rate: "50r/s"
    file_name: "{{ deploy_env }}_cas_commcare"
--- a/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/ansible/roles/nginx/vars/cas_ssl.yml
@@ -16,6 +16,10 @@ nginx_sites:
       zone: "$server_name"
       size: "10m"
       rate: "50r/s"
+    - name: "remote_ip_addr"
+      zone: "$binary_remote_addr"
+      size: "10m"
+      rate: "5r/s"
    file_name: "{{ deploy_env }}_cas_commcare"
    listen: "443 ssl default_server"
    server_name: "{{ CAS_SITE_HOST }}"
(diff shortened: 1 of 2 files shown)


ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -11,6 +11,11 @@ nginx_sites:
     - name: "{{ deploy_env }}_cas_formplayer"
       hosts: formplayer
       port: "{{ formplayer_port }}"
+   limit_req_zones:
+    - name: "server_name"
+      zone: "$server_name"
+      size: "10m"
+      rate: "50r/s"
```
emord: This rate and burst are arbitrary and should likely be changed, but wasn't sure what we were expecting per second

benrudolph:
nice this is cool. looks good to me once we figure out what we want as our rate

benrudolph:
current rates on datadog peak around 90r/s. not sure how useful that is, but at leasts grounds some sort of estimate

emord:
Forgot to mention that we could also limit this to certain endpoints by copying the nginx location configuration, but with a different limit_req and/or limit_req_zone. Could be useful if we don't want to limit phones updating the app or if we want restores and submissions to be limited differently

benrudolph:
was looking at the nginx dashboard, but i forgot to filter out softlayer requests. once i limited to hqproxy0 i see 15r/s at the peak

emord:
Realized we'd probably want to have one limit for the server and another per user/ip address so i added one of those as well

snopoke:
Nice, was also thinking the same.

emord:
@snopoke do you have any recommended limits for this? may be good to get this out before the rest of the phones are deployed

snopoke:
Thanks for pinging - been meaning to come back to this. This is what I think we want:
* global - 200r/s (burst 0, nodelay = true)
* per IP - 5r/s (burst 0, nodelay = True) - this won't work for web requests but since this is only for request from the mobile I think it's OK.

emord:
@snopoke thanks i updated the limits

emord:
> this won't work for web requests but since this is only for request from the mobile I think it's OK.

do we want to limit for web? I was thinking that that would be unnecessary since there should be much less activity on web that mobile

snopoke:
Yes I agree, no need to limit web requests. I was just noting that the per
IP rate limit would be to low for web requests.

Turns out it's quite useful to have mobile submitting to a separate site.

PR #628, opened 2017-03-07
icds postgresql performance tuning, fogbugz 248996

javierwilson:
adding performance tuning options to ICDS, and postgresql.conf template support for these options. @snopoke @kaapstorm

### DIFF
--- a/ansible/vars/icds/icds_public.yml
+++ b/ansible/vars/icds/icds_public.yml
@@ -61,6 +61,9 @@ postgres_users: "{{ secrets.POSTGRES_USERS }}"
 
 postgres_config:
   max_connections: 300
+  work_mem: '8MB'
+  shared_buffers: '512MB'
+  effective_cache_size: '8GB'
   pgb_max_connections: 800
   pgb_default_pool: 290
   pgb_reserve_pool: 5
--- a/ansible/roles/postgresql/templates/postgresql.conf.j2
+++ b/ansible/roles/postgresql/templates/postgresql.conf.j2
@@ -115,6 +115,8 @@ ssl = {% if postgresql_ssl_enabled %}true{% else %}false{% endif %}             
 # - Memory -
 
 #shared_buffers = 1024MB           # min 128kB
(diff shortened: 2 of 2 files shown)


ansible/vars/icds/icds_public.yml
```diff
@@ -61,6 +61,9 @@ postgres_users: "{{ secrets.POSTGRES_USERS }}"
 
 postgres_config:
   max_connections: 300
+  work_mem: '8MB'
+  shared_buffers: '512MB'
```
benrudolph: >  a reasonable starting value for shared_buffers is 25% of the memory in your system

how come we aren't using that heuristic?

ansible/vars/icds/icds_public.yml
```diff
@@ -61,6 +61,9 @@ postgres_users: "{{ secrets.POSTGRES_USERS }}"
...
+  effective_cache_size: '8GB'
```
benrudolph: > Setting effective_cache_size to 1/2 of total memory would be a normal conservative setting, and 3/4 of memory is a more aggressive but still reasonable amount.

also for this one, what is the ram on our db machine?

snopoke:
@benrudolph good point re memory numbers, one factor to consider is that we don't have dedicate PG machines, they are also running other processes e.g. Riak

ansible/vars/icds/icds_public.yml
javierwilson: Same as shared_buffers, I wanted to increase and test.  Until we reached 25% and 1/2, just being cautious.

benrudolph: ok sounds good 👍

ansible/vars/icds/icds_public.yml
javierwilson: We increased from 128MB to 512MB, I want to increase and test a few times before we go as high as recommended, we can try 2GB next and 8GB... but not if the system is not actually taking advantage of it.

PR #631, opened 2017-03-13
Fix riak cs config

millerdev:
This configuration cannot be applied to an existing cluster. All data must be migrated to a new cluster via the API.
http://lists.basho.com/pipermail/riak-users_lists.basho.com/2017-March/019142.html

@javierwilson @snopoke


javierwilson:
I think `extra_conf: advanced.config` should be `extra_conf: advanced.conf` since the template is called `advanced.conf.j2`
Also, `extra_conf: ""`, should be explicit whenever `riakcs/configure` is invoked.

millerdev:
@javierwilson Good catch. The template should be named `advanced.config.j2`. Fixed in 413216f

ansible/vars/staging/staging_public.yml
```diff
@@ -19,7 +19,7 @@ active_sites:
   motech2: False
   riakcs: True
 
-riak_ring_size: 64
+riak_ring_size: 128
```
millerdev: this should be done for softlayer (and eventually production), but not staging

ansible/roles/swap/tasks/main.yml
```diff
@@ -48,7 +48,7 @@
 - name: Set swappiness
   sysctl:
     name: vm.swappiness
-    value: 0
+    value: 1
```
emord: @millerdev should this change get pulled out to master and deployed to all environments?

millerdev: @emord Yes. I wanted to check that it was working as expected on softlayer, which I can now confirm.

ansible/roles/riakcs/configure/templates/riakcs.conf.j2
```diff
@@ -421,7 +421,7 @@ platform_log_dir = /var/log/riak-cs
 ## 
 ## Acceptable values:
 ##   - text
-nodename = riakcs-{{ inventory_hostname_short }}@{{ inventory_hostname }}
+nodename = riakcs@{{ inventory_hostname }}
```
millerdev: this change may need to be made conditional to accommodate envs that are using the old naming style (some of those envs are also using bitcask, so that needs to be addressed before this PR can be merged as well).

ansible/roles/riakcs/configure/defaults/main.yml
```diff
@@ -8,8 +8,9 @@
 #stanchion_port: ...
 
 riak_data_root_leveldb: "{{ riak_data_dir }}/leveldb"
+riak_data_root_blocks: "{{ riak_data_dir }}/blocks"
 riak_ring_dir: "{{ riak_data_dir }}/ring"
-riak_nodename: "riak-{{ inventory_hostname_short }}@{{ inventory_hostname }}"
-riak_ring_size: 64
+riak_nodename: "riak@{{ inventory_hostname }}"
```
millerdev: See https://github.com/dimagi/commcarehq-ansible/pull/631/files#r153820979

millerdev:
I think this was marked as merged when #1107 was merged. I came here to close it after doing that only to realize it had already been merged...by me. I do not remember merging this PR, so unless I'm going crazy...

PR #633, opened 2017-03-16
Jw/increase max stack depth

javierwilson:
related fogbugz # 249655
@snopoke @benrudolph


snopoke:
Do we not also have to change the kernel settings?

PR #637, opened 2017-03-22
prioritize ucr pillows for icds

emord:
https://github.com/dimagi/commcare-hq/pull/15479

This creates specific pillows for curs that are important for ICDS. @sheelio can you take a look at this list to make sure I got the important ones correct?

@snopoke May also be good to split the "non-priority" pillow into two or three, but I'm not familiar with specs on the ICDS machines, so I don't want to overwhelm them. I think we'll likely want (number of cores - 1 or 2) number of static pillows running.

I think I recall you mentioning that our web workers are underutilized. Maybe we could start one or two pillows on those as well?

### DIFF
--- a/ansible/vars/icds/icds_public.yml
+++ b/ansible/vars/icds/icds_public.yml
@@ -291,14 +291,14 @@ localsettings:
             - 'location'
           include_ucrs:
             - 'static-awc_location'
-      - name: 'kafka-ucr-static-ccs_record_cases,'
+      - name: 'kafka-ucr-static-ccs_record_cases'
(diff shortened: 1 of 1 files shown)


emord:
Not entirely happy with this new change as it means there needs to be an ansible deploy/pillow fast forwards when we want to change this, but I really like that this will split out what ucrs actually fast/slow and it also ignores topics it doesn't need.

I checked the pillow machine and it has 16 threads (8 cores w/ hyperthreading i think) and we're nowhere near capacity even though we were really behind. I think this will be a good stopgap until we can get a better async solution (celery/gevent based)

PR #638, opened 2017-03-23
Updated: new and old riakcs

millerdev:
This adds localsettings changes to @javierwilson's original branch [jw/new_and_old](https://github.com/dimagi/commcarehq-ansible/compare/jw/new_and_old_riakcs). The first commit is his commit rebased on master. My objective was to design it such that it:

- supports all existing envs (with old riak cluster)
- supports envs with two riak clusters (old and new)
- supports envs where the migration is complete (new cluster only)

Managing the proxy ports is particularly tricky because we need to make sure there is no downtime between when the proxy is updated with new cluster configuration(s) and when localsettings are updated on django workers, etc. I think no subsequent ansible code changes will be needed except for environment configurations.

Additionally, this can be merged and we can continue to run ansible against the old riak cluster until #631 and [commcare-hq-deploy:jw/new_riak_cluster](https://github.com/dimagi/commcare-hq-deploy/compare/jw/new_riak_cluster) are merged.

Prerequisites for using this to prepare staging for migration:
- [x] new staging riak cs cluster must be operational
- [x] add `[riakcs_new]` group (with hosts) to [commcare-hq-deploy](https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/inventory/staging)
- [x] keys for the new cluster must be added to `ansible/vars/staging/staging_vault.yml`

The [migration process](https://github.com/dimagi/commcarehq-ansible/commit/eca837265e81a4c535359f360119489e2a4ab49f#diff-607e6e5925d0031dafa79eb80d198640R69) is more complicated than hoped. Is there a better way or are there simplifications that I'm overlooking?


ansible/group_vars/all.yml
```diff
@@ -52,6 +52,8 @@ dev_users:
 
 django_port: 9010
 riakcs_port: 8080
+temp_riakcs_proxy_port: 8081
+riakcs_proxy_port: "{{ temp_riakcs_proxy_port if 'BLOB_DB_MIGRATING_FROM_S3_TO_S3' in localsettings else riakcs_port }}"
```
snopoke: I think it would be clearer if this logic was in `deploy_proxy.yml` rather than some here and some there.

ansible/deploy_proxy.yml
```diff
@@ -18,7 +18,25 @@
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.motech == True, action: site, site_name: motech}
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.motech == True, action: site, site_name: motech_http }
     - { role: nginx, when: proxy_type == 'nginx' and active_sites.cchq_http_j2me == True, action: site, site_name: cchq_http_j2me }
-    - { role: nginx, when: proxy_type == 'nginx' and active_sites.riakcs == True, action: site, site_name: riakcs }
+    - role: nginx
+      action: site
+      site_name: riakcs
+      hosts_group: "{{ 'riakcs_new' if 'BLOB_DB_MIGRATING_FROM_S3_TO_S3' in localsettings else 'riakcs' }}"
+      when: proxy_type == 'nginx' and active_sites.riakcs == True
```
snopoke: moving the `riakcs_proxy_port` from `vars/all.yml` to here would make this easier to reason with

ansible/group_vars/all.yml
millerdev: Agreed, but then I'd need to duplicated it in localsettings.py.j2 as well since `s3_blob_db_url` uses this var too.

millerdev:
> Is it necessary for the new cluster to end up on the same port as the old one?

Good question. I think the before/after migration configuration is simpler if the new cluster ends up on the same port as the old one. If we wanted to have the new cluster on a new port after the migration then we'd need to add another setting in `{{deploy_env}}_public.yml` that either is the proxy port number itself, or indicates that this particular migration is complete. We could effectively replace the `TEMP_RIAKCS_PROXY` variable with this new variable, but it would need to live in the configuration indefinitely (or at least until all clusters are migrated) instead of being temporary and reusable for future migrations. As is, the configuration before and after the migration looks the same with the exception of the hosts in the inventory file.

millerdev:
@snopoke responded to your question above. Happy to go to voice next week if you want to discuss more details.

millerdev:
@snopoke I realized there is a better/simpler way: 2bd7a31

I've tested proxy and localsettings migration states with vagrant, and it works as expected.

millerdev:
@javierwilson ran this against staging just now. It's working!

snopoke:
Nice!

PR #639, opened 2017-03-23
Etc hosts

benrudolph:
@snopoke @javierwilson Adds fully qualified hostname to /etc/hosts file. in the future we may want to do this for all hosts

cc: @millerdev


javierwilson:
i think this may also solve the problem of providers adding the public IP against the local hostname in /etc/hosts

PR #641, opened 2017-03-24
forward proxy deployment

javierwilson:
it improves on existing role: adds ftp proxy, uses https "insteadOf" for git, etc. already tested on my vagrant environment but you're welcome to give it a try as well.

### DIFF
--- a/ansible/deploy_http_proxy.yml
+++ b/ansible/deploy_http_proxy.yml
@@ -1,5 +1,5 @@
 - name: http proxy for external calls
-  hosts: all
+  hosts: patch_clients
   sudo: yes
   roles:
     - http_proxy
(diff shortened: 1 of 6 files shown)


ansible/deploy_http_proxy.yml
```diff
@@ -1,5 +1,5 @@
 - name: http proxy for external calls
-  hosts: all
+  hosts: patch_clients
```
javierwilson: No other PR, actually we just need to exclude the patch server itself. I will look it up, something liek "all but patch"

ansible/roles/http_proxy/tasks/main.yml
```diff
@@ -2,49 +2,80 @@
...
     - 'no_proxy'
     - 'NO_PROXY'
 
+- name: Bower proxy
+  template:
+    src: bowerrc.j2
+    dest: "{{ item.home }}/.bowerrc"
+  become: yes
+  become_user: "{{item.user}}"
+  ignore_errors: True
```
javierwilson: Sometimes user does not exist, I thinkg cchq is not created in all servers am I wrong, to be honest this also helped quite a bit while testing on my own environment.

PR #646, opened 2017-03-27
no remote ip based rate limit

sravfeyn:
@emord


ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -16,10 +16,6 @@ nginx_sites:
       zone: "$server_name"
       size: "10m"
       rate: "500r/s"
-    - name: "remote_ip_addr"
```
emord: You don't actually need to get rid of this. It's enough to not use it under limit_reqs

PR #647, opened 2017-03-27
refresh_interval to improve performance

benrudolph:
@snopoke @emord https://www.elastic.co/guide/en/elasticsearch/reference/1.7/index-modules.html

cc: @millerdev

### DIFF
--- a/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
+++ b/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
@@ -52,6 +52,8 @@ index.search.slowlog.threshold.query.info: 5s
 index.search.slowlog.threshold.query.debug: 2s
 index.search.slowlog.threshold.query.trace: 500ms
 
+index.refresh_interval: 5s
+
 # somewhat of a security risk but required by pact custom reports
 script.engine.groovy.inline.aggs: true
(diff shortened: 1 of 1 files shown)


ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -52,6 +52,8 @@ index.search.slowlog.threshold.query.info: 5s
 index.search.slowlog.threshold.query.debug: 2s
 index.search.slowlog.threshold.query.trace: 500ms
 
+index.refresh_interval: 5s
```
emord: I don't know that we want this here. This means that it will be for all indexes. We may not want that on users/groups/apps.

Also I think this will get overwritten on each deploy. https://github.com/dimagi/commcare-hq/blob/8667a4159ad39d3307f92a3812d74006d610fd1b/corehq/ex-submodules/pillowtop/es_utils.py#L20

benrudolph: thanks, didn't know we did that in the django code. yes i did think about doing it for all indices. i don't really see a downside in increasing it for every index since we don't use it for anything real time. but i suppose in general it'd be best to just do it for forms and cases. i could imagine a giant user upload doing something similar though (or giant ucr insert)

PR #654, opened 2017-03-31
combine icds pillows

emord:
goes with https://github.com/dimagi/commcare-hq-deploy/pull/227

@sravfeyn @snopoke


emord:
updated this and the related deploy module to break up other_cases (it's falling behind) and to keep child_health_cases and ccs_record_cases separate from the monthly partitions for now. Did this because they are able (so far) to keep up with the changes because they don't rely on related docs as much and don't need to get the forms so it's not as big of a win for caching

PR #655, opened 2017-04-01
purge restore_payloads +mmin 8 hours

javierwilson:
i wonder if max_age should allow a different setting per environment, please comment.

### DIFF
--- a/ansible/roles/shared_dir/tasks/setup_host.yml
+++ b/ansible/roles/shared_dir/tasks/setup_host.yml
@@ -46,13 +46,13 @@
   cron:
     name: "{{ item.name }}"
     special_time: daily
-    job: "find {{ item.dir }} -type f -maxdepth 1 -mtime +{{ item.max_age }} -delete"
+    job: "find {{ item.dir }} -maxdepth 1 -type f -mmin +{{ item.max_age }} -delete"
     user: root
     cron_file: purge_cache_files
   with_items:
-    - {name: 'Purge old restore files', dir: "{{ restore_payload_dir_host }}", max_age: 2}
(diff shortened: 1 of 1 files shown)


ansible/roles/shared_dir/tasks/setup_host.yml
```diff
@@ -46,13 +46,13 @@
...
+    job: "find {{ item.dir }} -maxdepth 1 -type f -mmin +{{ item.max_age }} -delete"
     user: root
     cron_file: purge_cache_files
   with_items:
-    - {name: 'Purge old restore files', dir: "{{ restore_payload_dir_host }}", max_age: 2}
-    - {name: 'Purge old transfer files', dir: "{{ transfer_payload_dir_host }}", max_age: 2}
-    - {name: 'Purge old temp files', dir: "{{ shared_temp_dir_host }}", max_age: 1}
+    - {name: 'Purge old restore files', dir: "{{ restore_payload_dir_host }}", max_age: 480 } # 8 hours in minutes
+    - {name: 'Purge old transfer files', dir: "{{ transfer_payload_dir_host }}", max_age: 2880 } # 2 days in minutes
+    - {name: 'Purge old temp files', dir: "{{ shared_temp_dir_host }}", max_age: 1440 } # 1 day in minutes
```
snopoke: I actually think we can reduce these even more, I'd say 3 hours max for all of them. I don't think there's any reason to keep then longer than even 1 hour but can make it 3 hours to be conservative.

Also you'll need to change the cron time so that it runs more frequently (hourly) on line 48.

javierwilson: just to be sure, this is true for all environments, right?

PR #656, opened 2017-04-02
enable password hashing on staging

mkangia:
PR for password hashing is on staging now
https://github.com/dimagi/commcare-hq/pull/15490

Corresponding change to localsettings for staging to enable it


snopoke:
:+1: though I assume we don't want this set permanently right?

mkangia:
@snopoke this is going to be up just for auditors. Will bring this down once tested.

PR #658, opened 2017-04-04
File limits increase

benrudolph:
this'll bump the file limits. the default on softlayer was 4096 when formplayer began running out of file descriptors and it wasn't anywhere near running out of memory (8GB machine). doesn't quite work, i think in the current form it requires a restart to take affect. looking into a work around

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -35,9 +35,7 @@
   tags:
     - nofile_limits
   with_items:
-    - 'rabbitmq'
-    - 'elasticsearch'
-    - '{{ cchq_user }}'
+    - '*'
     - 'root'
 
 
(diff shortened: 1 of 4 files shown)


benrudolph:
```
brudolph@hqdb0:~$ ulimit -Hn
65536
brudolph@hqdb0:~$ ulimit -n
65536
```
i believe by requiring the `pam_limits.so` it gives the new nofile on restart of service

ansible/roles/common/tasks/main.yml
```diff
@@ -23,3 +23,40 @@
   apt: name="libssl1.0.0" state=present
   tags:
     - upgrade_openssl
+
+
+- name: Hard file limits
+  sudo: yes
+  pam_limits:
```
sravfeyn: This module is in ansible `>2.0`, so this is failing on our prod ansible (`1.9.6`). Are we planning to upgrade our ansible?

kaapstorm: This PR should have done that already: https://github.com/dimagi/commcarehq-ansible/pull/617 (specifically commit 2e88b685)

PR #659, opened 2017-04-07
mail relay configuration

javierwilson:
creates deploy_mailrelay that forces outbound email to go through the server defined at the "mailrelay" group (if defined at all)
also sets django mail server to localhost for ICDS (since we're taking care of thing at the MTA level)

### DIFF
--- a/ansible/deploy_mailrelay.yml
+++ b/ansible/deploy_mailrelay.yml
@@ -13,7 +13,7 @@
         dest: /etc/postfix/main.cf
         regexp: '^relayhost ='
         line: "relayhost = {{ mailrelay }}"
-      when: groups['mailrelay'] is defined
+      when: groups['mailrelay'] is defined and mailrelay is defined
 
 - name: "mail relay clients deploy"
   hosts: all:!mailrelay
--- a/ansible/deploy_mailrelay.yml
+++ b/ansible/deploy_mailrelay.yml
@@ -2,7 +2,8 @@
   hosts: all
   sudo: yes
   tasks:
-    - package: name=postfix state=present
+    - debconf: name=postfix question="postfix/main_mailer_type" value="'Internet Site'" vtype="string"
+    - apt: name=postfix state=present force=yes
 
 - name: "mail relay deploy"
   hosts: mailrelay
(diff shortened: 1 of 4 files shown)


ansible/deploy_mailrelay.yml
```diff
@@ -0,0 +1,27 @@
...
+
+- name: "mail relay deploy"
+  hosts: mailrelay
+  sudo: yes
+  tasks:
+    - name: "sets our relayhost"
+      lineinfile:
+        dest: /etc/postfix/main.cf
+        regexp: '^relayhost ='
+        line: "relayhost = {{ mailrelay }}"
```
sravfeyn: To handle undefined case

sravfeyn: Just to avoid the confusion between actual `mailrelay` host and the postfix (or 'proxy') `mailrelay` host variables, can we rename them on the lines of `mailrelay_postfix_host` and `mailrelay_host` or some other name?

javierwilson: maybe "mailrelay" is ok for the inventory, since all of those are hosts, but maybe use smtp_server instead of the "mailrelay" variable. would that do the trick?

sravfeyn: Sure

calellowitz:
@javierwilson do we want to add this to the full stack deploy?

ansible/deploy_mailrelay.yml
javierwilson: i'm actually moving this variable to the inventory and calling it "parent_mailrelay" which makes more sense since this variable is dependant and affects only the mailrelay host  i've modified the dev inventory to test this in my dev environment.

javierwilson:
@sravfeyn  i'm actually moving this variable to the inventory and calling it "parent_mailrelay" which makes more sense since this variable is dependant and affects only the mailrelay host i've modified the dev inventory to test this in my dev environment.
@calellowitz not just yet... however, since it only affects deployment if it has the [mailrelay] group defined, it shouldn't be a problem. we'll test this soon in icds and come back to you on this one.

snopoke:
What happens when we get throttled by NIC?

javierwilson:
Postfix has its own way to deal with this, it actually depends on *how* NIC does it, if they send temp fails, they postfix will retry later, which should be the right way.
Also, postfix is pretty flexible, we could configure how/when to retry, how often, etc, but the defaults I think are the polite way.

PR #663, opened 2017-04-10
Revert "no remote ip based rate limit"

emord:
Reverts dimagi/commcarehq-ansible#646

I think this should be reverted once the trainings are complete, so maybe a couple weeks?

@sravfeyn @calellowitz @benrudolph

### DIFF
--- a/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/ansible/roles/nginx/vars/cas_ssl.yml
@@ -16,6 +16,10 @@ nginx_sites:
       zone: "$server_name"
       size: "10m"
       rate: "500r/s"
+    - name: "remote_ip_addr"
+      zone: "$binary_remote_addr"
+      size: "10m"
+      rate: "5r/s"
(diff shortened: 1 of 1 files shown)


sravfeyn:
I have asked ICDS folks when we can do this, I will ping back

ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -34,6 +38,8 @@ nginx_sites:
        limit_reqs:
          - zone_name: "server_name"
            nodelay: True
+         - zone_name: "remote_ip_addr"
```
emord: This will require a burst attribute since https://github.com/dimagi/commcarehq-ansible/pull/690 requires this now

emord:
@dimagi/scale-team I think we want to drop this, but pinging in case we want to revisit it.

This would the ability to rate limit based on IP so no one user could hit us too many times. I don't think it's possible at the moment or will be possible as trainings are still going on

snopoke:
Agree that we can't do this for now

PR #664, opened 2017-04-10
Remove run_once

benrudolph:
@dannyroberts either this is new functionality in 2.0 or we were getting lucky on prod. when `run_once` is specified, it only runs once even if it's skipped. so this never gets run unless celery is the first machine loaded

fyi: @emord @calellowitz


dannyroberts:
that's a really weird thing for `run_once` to mean... wonder why one would ever want that!

PR #665, opened 2017-04-10
Couch log rotate

benrudolph:
@javierwilson @emord @calellowitz @sravfeyn this rotates the couch log. looks like this:

```
+# Ansible managed, do not edit directly
+
+"/usr/local/var/log/couchdb/*.log" {
+  daily
+  size 100M
+  rotate 10
+  missingok
+  compress
+  delaycompress
+  copytruncate
+  nocreate
+  notifempty
+  }
```
not yet deployed. this will run through the log file pretty quickly. personally i don't find the logs that useful, but let me know if you'd like me to increase the size or rotate so we can keep them around longer


emord:
About how long do you think these logs will last? I wonder if it'd be useful to have an ansible task to copy logs so that we could copy quickly if a service fails

benrudolph:
well in one day it wrote ~3GB of logs... so about 8 hours? i also think we could bump the log level to something higher: https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/couchdb/templates/local.ini.j2#L44

emord:
whoa we're logging debug? does seem like we should be logging info at most on prod

benrudolph:
i'm not sure what the default is, but it is usually convention to have the commented out value be the default. this doc makes me think it's `info` https://wiki.apache.org/couchdb/Configurationfile_couch.ini

benrudolph:
(deployed this to nic)

PR #670, opened 2017-04-13
datadog use pipelining (fails otherwise on ansible 2)

javierwilson:
datadog role will fail because of sudo_user in ansible 2, this fixes it.  this change however i have not actually tested on my dev environment. but since others almost identical have worked, i hope for the best!


benrudolph:
nice

PR #671, opened 2017-04-13
adding auto compact to couch for the two bad databases

calellowitz:
@benrudolph @emord 
Here is auto compaction for couch. when the data size is less 30% of the db or 40% of the view, it will trigger a compact but only between 11pm and 4am. I only did it for the dbs that we have seen get large, because it seems like wasted resources otherwise, but happy to make it the default for all databases.


calellowitz:
put a dont merge tag on while i look into what time zone the compaction daemon is using.

calellowitz:
relatively certain couch is in GMT so adjusted times accordingly (and pushed forward start since 10:30 seems plenty late)

calellowitz:
@benrudolph I was concerned about saving resources on a stretched machine and felt the others were too small to matter, but happy to change it if you think the other way is better.

PR #672, opened 2017-04-14
fix flattened and pipelining

sravfeyn:
@javierwilson


sravfeyn:
Fixed some more issues while deploying webworkers. I have tested all these so good to go

PR #675, opened 2017-04-14
remaking couch user whenever config changes

calellowitz:
@kaapstorm @emord @benrudolph 
couch stores admin users by appending their name and password hash to the config file. every time we update the couch config the users gets deleted. this ensures that we remake the user each time we change the config.

### DIFF
--- a/ansible/roles/couchdb/tasks/main.yml
+++ b/ansible/roles/couchdb/tasks/main.yml
@@ -56,6 +56,7 @@
 
 - name: Apply CouchDB config
   template: src=local.ini.j2 dest=/usr/local/etc/couchdb/local.ini
+  register: update_config
 
 - name: CouchDB ownership permissions
   file: path="{{ item }}" owner=couchdb group=couchdb recurse=yes state=directory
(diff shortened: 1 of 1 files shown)


calellowitz:
@javierwilson

ansible/roles/couchdb/tasks/main.yml
```diff
@@ -80,16 +81,16 @@
...
+    method: PUT
   when: couch_response.status == 404
+
+- name: Set CouchDB username and password
+  uri:
+    url: "{{ couchdb_base_url }}/_config/admins/{{ localsettings.COUCH_USERNAME }}"
+    method: PUT
+    body: '"{{ localsettings.COUCH_PASSWORD }}"'
+    body_format: raw
+  when: update_config|changed
```
emord: I think this used to run when couch_response.status == 404. does that still need to happen?

calellowitz: yeah but it should still do so. `couch_response.status` is 404 when the dbs havent been created. i cant really foresee a situation where the dbs needed to be created from scratch but the config is unchanged. if you think thats possible i can add in an `or`

benrudolph:
Time to relax!

PR #679, opened 2017-04-17
sets timezone to UTC

javierwilson:
fitst step to allow setting timezone to UTC in all servers


sravfeyn:
Looks good, but do you want to wait till we get confirmation from others?

benrudolph:
did this get deployed? still seeing nginx record logs in india timezone. would be great to start getting metrics

PR #680, opened 2017-04-17
allows to set hostname

javierwilson:
this allows to set the hostname when "hostname" is defined , and domain name if specified

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -60,3 +60,22 @@
     line: "session required pam_limits.so"
   tags:
     - nofile_limits
+
+
+- name: "Sets hostname"
+  become: yes
+  hostname:
+    name: "{{ hostname }}"
+  when: hostname is defined
+  tags:
+    - hostname
+
+- name: "Sets /etc/hosts"
+  become: yes
+  lineinfile:
+    dest: /etc/hosts
+    regexp: '^{{ ansible_default_ipv4.address }} '
+    line: '{{ ansible_default_ipv4.address }} {{ hostname }}.{{ internal_domain_name }} {{ hostname }}'
+  when: hostname is defined and internal_domain_name is defined
+  tags:
+    - hostname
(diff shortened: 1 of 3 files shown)


ansible/roles/common/tasks/main.yml
```diff
@@ -60,3 +60,22 @@
...
+  when: hostname is defined
+  tags:
+    - hostname
+
+- name: "Sets /etc/hosts"
+  become: yes
+  lineinfile:
+    dest: /etc/hosts
+    regexp: '^{{ ansible_default_ipv4.address }} '
+    line: '{{ ansible_default_ipv4.address }} {{ hostname }}.{{ internal_domain_name }} {{ hostname }}'
```
sravfeyn: Nice!

ansible/roles/common/tasks/main.yml
```diff
@@ -60,3 +60,22 @@
     line: "session required pam_limits.so"
   tags:
     - nofile_limits
+
+
+- name: "Sets hostname"
+  become: yes
+  hostname:
+    name: "{{ hostname }}"
```
sravfeyn: Where is this defined? Some of already set hostnames on these servers might not match with what you have set in the DNS. May be double check this before deploying?

ansible/vars/icds/icds_public.yml
```diff
@@ -2,6 +2,7 @@ SITE_HOST: 'www.icds-cas.gov.in'
 CAS_SITE_HOST: 'cas.commcarehq.org'
 NO_WWW_SITE_HOST: 'icds-cas.gov.in'
 deploy_env: icds
+internal_domain_name: 'internal-icds.commcarehq.org'
```
benrudolph: may also want to set this for the rest of the envs

ansible/roles/common/tasks/main.yml
javierwilson: I do this at the hostname inventory level like  127.0.0.1 hostname="localhost"

ansible/roles/common/tasks/main.yml
sravfeyn: There's `[etc_hosts](https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/common/tasks/main.yml#L5)` tags that does similar editing already. Should we merge these two?

PR #681, opened 2017-04-18
fixes for new riak machine

calellowitz:
@javierwilson @millerdev cc: @dannyroberts (i think you did the initial setup on this)
When trying to spin up a new riak machine on icds today i discovered that despite our configuration, existing nodes had been created with a ring size of 64, so in order to allow the new node to join its ring size, and all other new nodes, had to be 64 as well. I know we had good reasons to use 128 at first so I am curious if we think this is a long term problem or manageable.

### DIFF
--- a/ansible/vars/icds/icds_public.yml
+++ b/ansible/vars/icds/icds_public.yml
@@ -36,7 +36,7 @@ active_sites:
   riakcs: True
   cas_ssl: True
 
-riak_ring_size: 128
+riak_ring_size: 64
 
 riakcs_s3_access_key: "{{ secrets.RIAKCS_S3_ACCESS_KEY }}"
 riakcs_s3_secret_key: "{{ secrets.RIAKCS_S3_SECRET_KEY }}"
@@ -175,7 +175,7 @@ etc_host_lines:
   - '10.247.24.13		www.icds-cas.gov.in'
 etc_host_lines_removed: ['etc_host_lines']
 
-http_proxy_address: '10.247.63.132'
+http_proxy_address: '10.247.24.16'
 http_proxy_port: '3128'
 
 site_locations:


javierwilson:
This has to change to 128 and is part of the planned riak migration. We're just taking too long :(

ansible/vars/icds/icds_public.yml
```diff
@@ -175,7 +175,7 @@ etc_host_lines:
   - '10.247.24.13		www.icds-cas.gov.in'
 etc_host_lines_removed: ['etc_host_lines']
 
-http_proxy_address: '10.247.63.132'
+http_proxy_address: '10.247.24.16'
```
benrudolph: how come the proxy address changed?

emord: According to cal This changed because the new machines have not been setup with the patch server so this tunnels it through es0 to the proxy. I needed this change for celery1, but i'm not sure that this should get commited, if it's not hard to setup the patch server to work with new machines

ansible/vars/icds/icds_public.yml
```diff
@@ -36,7 +36,7 @@ active_sites:
   riakcs: True
   cas_ssl: True
 
-riak_ring_size: 128
+riak_ring_size: 64
```
millerdev: woah! I thought ICDS had ring size of 128. I don't think this should change, and I'm not understanding why you had to change it now.

calellowitz: it has to change now because all of the existing nodes have a ring size of 64 and new ones need to have a ring of the same size. My hunch as to why it changed is that when we were initially setting up, the riak machines were deployed with a ring size of 64, and when we went to change the config, nothing took affect because the machines had already been commissioned with the smaller ring size. even if there is no data, in order to change the ring size you have to manually delete the old ring, and nobody did that. so all of the old machines show `ring_creation_size = 128` which is the value in the config but `ring_num_partitions = 64` which is the actual value in the cluster.
from the riak docs:
```
If ring_num_partitions and ring_creation_size do not agree, that means that the ring_creation_size value was changed too late and that the proper steps were not taken to start over with a new ring.

Note: Riak will not allow two nodes with different ring sizes to be joined into a cluster.
```

millerdev: Thanks for the detailed explanation, @calellowitz. That's both interesting and disappointing. We should change this when we setup the new higher-capacity riak cluster for ICDS.

calellowitz:
merging this, because both of those problems surfaced again when deploying machines today

ansible/vars/icds/icds_public.yml
snopoke: what's the status of this? Should this still be tunneling?

emord: @snopoke From what I understand all new machines will need to be tunneled through 24.16 until they are whitelisted on the patch server. There has been no resolution in our email. search for subject "additional VMs to patch server" @mkangia may know more, but they haven't responded for over a week on that thread

javierwilson: that's right, i think in the meantime it's best that everybody uses es0 as proxy instead of dealing with a two-proxy configuration.

emord: @javierwilson what happens if es0 gets its' proxy set as itself though?

javierwilson: no problem, it just uses itself as a proxy, it's part of the local network allow to use it. sames goes for the patch server, it can use itself as a proxy, it gains the benefits of cached files.

PR #686, opened 2017-04-19
ansibale -> ansible

emord:
@javierwilson @calellowitz

### DIFF
--- a/ansible/roles/commcarehq/tasks/main.yml
+++ b/ansible/roles/commcarehq/tasks/main.yml
@@ -55,6 +55,7 @@
       pip:
         requirements: "{{ code_home }}/requirements/{{ item }}"
         virtualenv: "{{ virtualenv_source }}"
+        chdir: "{{ code_home }}"
       tags:
         - slow
       when: not testing|default(False)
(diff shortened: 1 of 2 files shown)


ansible/roles/http_proxy/tasks/main.yml
```diff
@@ -49,9 +49,8 @@
     ansible_ssh_pipelining: true
   ignore_errors: True
   with_items:
-    - {'user': 'ansible', 'home': '/home/ansibale' }
+    - {'user': 'ansible', 'home': '/home/ansible' }
     - {'user': 'root', 'home': '/root' }
-    - {'user': '{{ cchq_user }}' , 'home': '{{ code_home }}'}
```
emord: Removing this line because it's not needed and it was erroring. this tried to become cchq and then copy a file from /tmp/ansible which it doesn't have access to.

This doesn't stop bower from working because it uses the machines general http/https proxy

ansible/roles/commcarehq/tasks/main.yml
```diff
@@ -55,6 +55,7 @@
       pip:
         requirements: "{{ code_home }}/requirements/{{ item }}"
         virtualenv: "{{ virtualenv_source }}"
+        chdir: "{{ code_home }}"
```
emord: Need to have this because requirements now have a requirement from a submodule

https://github.com/dimagi/commcare-hq/blob/bb238246bf98a5fe6f3702bee038404b03d2b073/requirements/requirements.txt#L101

PR #687, opened 2017-04-19
replacing root with cchq user in commcarehq role

calellowitz:
@emord @gcapalbo @kaapstorm cc: @benrudolph 
I switched these tasks back to the cchq user because otherwise the `current` symlink was owned by root which made cloning the virtualenv across releases fail.

### DIFF
--- a/ansible/roles/commcarehq/tasks/main.yml
+++ b/ansible/roles/commcarehq/tasks/main.yml
@@ -63,19 +63,6 @@
       tags:
         - slow
       when: not testing|default(False)
-
-    - name: chown clone
-      become: true
-      become_user: "{{ cchq_user }}"
-      file:
-        path: "{{ item }}"
-        state: directory
-        owner: "{{ cchq_user }}"
-        group: "{{ cchq_user }}"
-        recurse: yes
-      with_items:
-        - "{{ code_source }}"
-        - "{{ virtualenv_source }}"
   when: not gitdir.stat.exists
 
 - name: Upgrade setuptools
--- a/ansible/roles/commcarehq/tasks/main.yml
+++ b/ansible/roles/commcarehq/tasks/main.yml
@@ -9,7 +9,7 @@
 #    state: absent
 
 - name: create required directories
-  sudo: yes
+  become: true
   file:
     path: "{{ item }}"
     owner: "{{ cchq_user }}"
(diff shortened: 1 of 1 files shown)


calellowitz:
@kaapstorm This reverts a change you had made before, just want to make sure there was not a reason for that that i am missing in this PR

benrudolph:
makes sense

benrudolph:
looks like test failure though

kaapstorm:
Yeah, I'm guessing that's not going to work, because Ansible isn't going to be able to do that as the cchq user without violating 2.0's security restrictions (specifically, saving a password in a world-readable temp file iirc). You should probably try setting the cchq user as the owner of the symlink.

kaapstorm:
... you'll need to `chown -h` to affect the symlink itself, and not the directory it's referring to. I can't see how you would do that with Ansible's **file** module, so you may need to use the `chown` command itself.

gcapalbo:
in case it helps - checking my notes again i think the issue might not have been the owner of the symlink which prevented cloning the virtual envs, that might have just been a red herring. it looks like any user can still follow and delete symlinks owned by root.

the other problem i saw (and maybe the only problem) appears to have been that there was a file deep inside the virtualenv called setuptools.pth which was owned by root, and this was the file that generated the permissions error. switching that file to be owned by cchq fixed the problem.

emord:
@gcapalbo interesting. I would have though that should have been taken care of by https://github.com/dimagi/commcarehq-ansible/pull/687/files#diff-7fa3c99e5bd9021910b1b2e3b0a0eb1bL63 since it recurses into both the virutalenv and hte code directory

benrudolph:
> the other problem i saw (and maybe the only problem) appears to have been that there was a file deep inside the virtualenv called setuptools.pth which was owned by root, and this was the file that generated the permissions error. switching that file to be owned by cchq fixed the problem.

if anyone accidentally installed something by hand using root the file gets owned by root. seen it happen on staging, but that's about it

emord:
@benrudolph @gcapalbo We did do some stuff by hand while debugging this last night. however that should have been in the previous release directories (there were around 5 release directories when we finished that we manually deleted). The last ansible deploy ran through completely in a new release directory and we didn't do anything by hand in that directory.

calellowitz:
closing in favor of https://github.com/dimagi/commcarehq-ansible/pull/792

PR #690, opened 2017-04-20
Require burst

benrudolph:
@emord I think this is what was causing the rate limit issues. 

> By default, the maximum burst size is equal to zero.

it still seems weird though... as soon as i added a burst size equal to the rate limit size, we stopped getting rate limiting errors

### DIFF
--- a/ansible/roles/nginx/templates/site.j2
+++ b/ansible/roles/nginx/templates/site.j2
@@ -76,7 +76,7 @@ server {
       {% endfor %}
     {% endif %}
     {% for limit in location.get('limit_reqs', []) %}
-        limit_req zone={{ limit.zone_name }}{% if limit.get('burst') %} burst={{limit.burst}}{% endif %}{% if limit.get('nodelay') %} nodelay{% endif %};
+        limit_req zone={{ limit.zone_name }} burst={{limit.burst}}{% if limit.get('nodelay') %} nodelay{% endif %};
     {% endfor %}
   }
 {% endfor %}
--- a/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/ansible/roles/nginx/vars/cas_ssl.yml
@@ -33,7 +33,7 @@ nginx_sites:
        proxy_read_timeout: 900s
        limit_reqs:
          - zone_name: "server_name"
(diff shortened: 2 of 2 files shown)


emord:
This looks good to me although I've updated the trello with what i found https: //trello.com/c/CxXQJtQ6/52-investigate-into-why-rate-limiting-didn-t-work

It seems like the past two days we've been rate limiting very heavily without user issues which seems strange to me that this would be the cause for issues from today since we blocked a lot less

emord:
> we stopped getting rate limiting errors

Oh. did you think that those "error" logs were nginx "errors". anytime we 503 due to the rate limit it logs it as an error. we can change the level to limit_req_log_level.

If we're going to be bursting we might want to log the delays as well which will require logs of "warn" by default. (limit_req_log_level - 1)

benrudolph:
> anytime we 503 due to the rate limit it logs it as an error. we can change the level to limit_req_log_level.

yea that's what i was assuming. logging the delays sounds like a good idea. thanks for diving in

ansible/roles/nginx/templates/site.j2
```diff
@@ -76,7 +76,7 @@ server {
       {% endfor %}
     {% endif %}
     {% for limit in location.get('limit_reqs', []) %}
-        limit_req zone={{ limit.zone_name }}{% if limit.get('burst') %} burst={{limit.burst}}{% endif %}{% if limit.get('nodelay') %} nodelay{% endif %};
+        limit_req zone={{ limit.zone_name }} burst={{limit.burst}}{% if limit.get('nodelay') %} nodelay{% endif %};
```
emord: Also curious why you made burst required? This should work without dooing htat. and we may not want bursting when adding ip based limiting

emord:
I changed this to 750 burst as that's what's currently deployed

ansible/roles/nginx/templates/site.j2
benrudolph: i made it required because at the time i thought that in order to perform well it needed the attribute. now i'm not as sure

benrudolph:
👍

PR #691, opened 2017-04-20
fix up fp settings for icds

benrudolph:
@sravfeyn @javierwilson fixes some settings for formplayer. still needs to be tested


benrudolph:
@javierwilson there's a mail host in these settings that is set to `relay.nic.in` however that is not the same setting as `localsettings.EMAIL_SMTP_HOST`. wondering if that discrepancy was intentional

javierwilson:
not intentional, i think django should always use localhost. then all mail routing should be taken care by the machine.

emord:
Added open for review until after your testing is complete

benrudolph:
cool ran this and it's good 2 go

PR #692, opened 2017-04-20
use proper hostname

benrudolph:
this should allow us to get proper datadog timings from out couch logs

### DIFF
--- a/ansible/roles/datadog/vars/main.yml
+++ b/ansible/roles/datadog/vars/main.yml
@@ -18,7 +18,7 @@ datadog_parsers:
     python_file: "{{ datadog_parsers_dest }}/couch/parsers.py"
     python_function: parse_couch_logs
     hosts: "{{ groups['webworkers'] + groups['pillowtop'] + groups['celery'] }}"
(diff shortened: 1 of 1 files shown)


ansible/roles/datadog/vars/main.yml
```diff
@@ -14,11 +14,11 @@ datadog_parsers:
     python_file: "{{ datadog_parsers_dest }}/nginx/timings.py"
     python_function: parse_nginx_counter
     hosts: "{{ groups['proxy'] }}"
-  - input: "{{ log_home }}/{{ inventory_hostname_short }}-commcarehq.couch.log"
+  - input: "{{ log_home }}/{{ localsettings.DEPLOY_MACHINE_NAME }}-commcarehq.couch.log"
```
snopoke: why not use full hostname from ansible?

benrudolph: because we change which one use based on the env :/ https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/commcarehq/templates/localsettings.py.j2#L61

PR #701, opened 2017-04-22
swappiness

emord:
@dimagi/scale-team possibly part of the cause of our disk latency issues. On NIC the swappiness was set to 60 so postgres/riak were using swap even when there are gigs of memory available....

I don't think there's anything wrong with always setting swap to 0 no matter what so, i've removed the when check. 

I'm also not clear why there's swap in the first place. There is no swap_size defined in the inventory which is why this was never being run

@calellowitz run this on the new machines (esp pg3) before you fully migrate the db

buddy @millerdev

### DIFF
--- a/ansible/roles/swap/tasks/main.yml
+++ b/ansible/roles/swap/tasks/main.yml
@@ -12,25 +12,25 @@
   tags: swap
 
 - name: Create swap space
-  sudo: yes
+  become: true
   command: fallocate -l {{ swap_size }} /swapfile
   when: swap_size is defined and not swapfile.stat.exists
   tags: swap
 
 - name: Set swap file mode
-  sudo: yes
+  become: true
   file: path=/swapfile state=file mode=600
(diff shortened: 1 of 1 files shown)


emord:
Merging b/c this is already deployed to NIC and these branches can quickly get out of hand

snopoke:
👍

dannyroberts:
Would this change if deployed to production also set swapiness to 0 in production? Is that what we want for a machine that does actually use swap? (Which we sometimes rely on to keep costs down even though adding another machine would be better)

emord:
swappiness to 0 just means that it won't use swap until it runs out of memory, which I think is what we want. for riak nodes it was already set to 0.

On icds it was set to 60 (default for ubuntu) which means when we hit 40% mem used it begins swapping

dannyroberts:
Oh dear. Yeah I see how that's a problem. Sounds like the answer is yes this will make swappiness 0 in production and that's what we want. Think I get it now, thanks for the explanation!

PR #703, opened 2017-04-23
fixes

emord:
Some issues ran into when deploying ES nodes


ansible/vars/icds/icds_public.yml
```diff
@@ -48,7 +48,7 @@ ansible_user_password_sha_512: "{{ secrets.ANSIBLE_USER_PASSWORD_SHA_512 }}"
 fake_ssl_cert: no
 
 elasticsearch_endpoint: '{{ groups.elasticsearch.0 }}'
-elasticsearch_cluster_name: 'icds-hqes-1.x'
+elasticsearch_cluster_name: 'prodhqes-1.x'
```
emord: I think this was set to prodhqes when es0 and es1 were set up so that's the cluster name now.

We might want to look into changing that

PR #708, opened 2017-04-24
icds riak

emord:
@dimagi/scale-team this is what's currently being deployed to the new nodes. Already deployed to pg2 and pg3, will be es2 and es3 later today. It's mostly based on #631 

If you use this branch please make sure it includes later changes of #631 and master

@millerdev


ansible/roles/riakcs/configure/templates/advanced.config.j2
```diff
@@ -0,0 +1,16 @@
...
+    {riak_kv, [
+        {add_paths, ["{{ riakcs_ebin }}"]},
+        {storage_backend, riak_cs_kv_multi_backend},
+        {multi_backend_prefix_list, [{<<"0b:">>, be_blocks}]},
+        {multi_backend_default, be_default},
+        {multi_backend, [
+            {be_default, riak_kv_eleveldb_backend, [
+                {data_root, "{{ riak_data_root_leveldb }}"}
+            ]},
+            {be_blocks, riak_kv_eleveldb_backend, [
```
emord: On the old nodes this uses bitcask but daniel said it's ok that these will be using this

ansible/roles/riakcs/configure/templates/riak.conf.j2
```diff
@@ -501,3 +503,8 @@ buckets.default.allow_mult = true
 javascript.map_pool_size = 0
 javascript.reduce_pool_size = 0
 javascript.hook_pool_size = 0
+
+# leveldb.compression = on
```
emord: These aren't available in 2.1.1

ansible/roles/riakcs/configure/tasks/main.yml
```diff
@@ -56,6 +57,14 @@
   notify: "restart {{ name }}"
   when: drop_conf != ""
 
+- name: "Create {{ name }} log dir"
+  file: >
+    path="{{ log_path | dirname }}"
+    state=directory
+    owner={{ name }}
+    group=riak
+  when: log_path != ''
```
millerdev: rather than adding another task here you could add an entry to `data_dirs` to the configure role for each of [riak](https://github.com/dimagi/commcarehq-ansible/blob/dm/riak-config/ansible/deploy_riakcs.yml#L82-L86), [riak-cs](https://github.com/dimagi/commcarehq-ansible/blob/dm/riak-config/ansible/deploy_riakcs.yml#L110), and [stanchion](https://github.com/dimagi/commcarehq-ansible/blob/dm/riak-config/ansible/deploy_riakcs.yml#L104)

emord:
@millerdev this was originally based on yoru pr #631 what's the status of these changes? are they safe to merge to master?

millerdev:
No, this does not appear to be safe to merge to master. It looks like it re-introduces bitcask backend, which is not used on production or softlayer (and possibly not on enikshay either, although I'm not sure there). If we want to support different backends for different envs then we should make that configurable on a per-env basis.

My plan is to merge #631 once the prod migration is complete, but that appears to be stalled in the SQL attachments migration (no status updates have been printed for over a week) and I haven't looked into why yet.

dannyroberts:

millerdev:
I'd wait to hear from @emord or @snopoke, but I think it should probably be closed.

emord:
Yeah. I think we can close this. Thanks for resolving these branches @millerdev

PR #709, opened 2017-04-25
no 500 email handler for NIC

sravfeyn:


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -653,4 +653,6 @@ SENTRY_API_KEY = '{{ localsettings.SENTRY_API_KEY }}'
 
 {% if localsettings.get('GREENHOUSE_API_KEY') %}
 GREENHOUSE_API_KEY = '{{ localsettings.GREENHOUSE_API_KEY }}'
-{% endif %}
\ No newline at end of file
+{% endif %}
+
+LOCAL_LOGGING_LOGGERS = {{ localsettings.get('LOCAL_LOGGING_LOGGERS', {}) }}
```
sravfeyn: This will override https://github.com/dimagi/commcare-hq/pull/15880/files#diff-bdf3ecebd8379ca98cc89e545fc90899R1124

sravfeyn: Not sure if there is better way to disable 500 emails. Also, confirming that the `notify_exception` will still be logged in Sentry after this change

ansible/vars/icds/icds_public.yml
```diff
@@ -486,4 +486,7 @@ localsettings:
   ENABLE_DRACONIAN_SECURITY_FEATURES: yes
   TABLEAU_URL_ROOT: "https://{{ TABLEAU_HOST }}/"
   ICDS_DOMAIN: "{{ localsettings_private.ICDS_DOMAIN }}"
-
+  LOCAL_LOGGING_LOGGERS:
+    notify:
+      handlers:
+        - 'sentry'
```
snopoke: You need the other fields here too (`level` and `propagate`)

snopoke: You also need to override this one which is the django error logger:
```
'django': {
            'handlers': ['mail_admins', 'sentry'],
            'level': 'ERROR',
            'propagate': True,
        },
```

sravfeyn: Sorry, updated

PR #710, opened 2017-04-25
update localsettings for staging to enable used passwords check

mkangia:


mkangia:
its just staging. Merging it.

PR #711, opened 2017-04-25
adding back max connection variable

calellowitz:
@snopoke @javierwilson this was dropped in the reorganization of pg settings. noticed yesterday when migrating. considered also bumping up to 3500 since some machines are already there but figured that could be a different pr


emord:
Can bump to 3500 in a separate pr this is a better default

benrudolph:
sorry maybe dumb question, but how does it help to bump to 3500 when PG's max_connections is 400?

snopoke:
@benrudolph that's pgouncer max client connections which is connections coming into pgbouncer (not from pgbouncer to postgresql)

benrudolph:
ok so it can maintain it's own backlog of connections and swap a connection into the pool used to connect to PG when it's ready

PR #712, opened 2017-04-25
reduce size of nginx logs

dannyroberts:
hqproxy0 disk usage is > 85%, which we get alerted about. Log dir is 18G, total disk is 32G, so it's a significant portion.

Do you all think this is a reasonable response to that alert?

### DIFF
--- a/ansible/deploy_proxy.yml
+++ b/ansible/deploy_proxy.yml
@@ -66,7 +66,7 @@
           path: "{{ log_home }}/*.log"
           options:
             - monthly
-            - size 1G
+            - size 750M
             - rotate 14
             - missingok
             - compress


dannyroberts:
@orangejenny @snopoke

snopoke:
Similar problem on icds

dannyroberts:
ok cool

ansible/deploy_proxy.yml
```diff
@@ -66,7 +66,7 @@
           path: "{{ log_home }}/*.log"
           options:
             - monthly
-            - size 1G
+            - size 750M
```
benrudolph: note this won't make a difference if more than 750 MB is written in a month. i think we should drop the check down to `daily` and lower the size further and just keep more compressed logs

PR #721, opened 2017-04-28
Enable {nodelay, true}

gcapalbo:
@dimagi/scale-team


ansible/roles/couchdb/templates/local.ini.j2
```diff
@@ -19,7 +19,7 @@ bind_address = {{ inventory_hostname|ipaddr }}
 ; Options for the MochiWeb HTTP server.
 ;server_options = [{backlog, 128}, {acceptor_pool_size, 16}]
 ; For more socket options, consult Erlang's module 'inet' man page.
-;socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
+socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
```
gcapalbo: these values for recbuf and sndbuf are enabled by default.ini so this is just keeping them here

PR #722, opened 2017-04-29
Fix couch restarter

gcapalbo:
@dimagi/scale-team 
(the fix has already been applied)

### DIFF
--- a/ansible/roles/couchdb/templates/couch-restarter.sh.j2
+++ b/ansible/roles/couchdb/templates/couch-restarter.sh.j2
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # set the path so that service can be found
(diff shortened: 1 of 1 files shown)


ansible/roles/couchdb/templates/couch-restarter.sh.j2
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
gcapalbo: sh doesn't support the regular expression syntax for the wildcard matching, so the if condition was always evaluating to false and so the email was being sent every minute.  the emails are going to keep trickling in due to the backlog so i changed the subject in the message so we can tell which ones are real.

javierwilson: good catch!

benrudolph:
very nice, sorry about that 😞

PR #726, opened 2017-05-01
Endpoint rate limiting

emord:
Ran `awk -F\" '{print $2}' icds_cas_commcare-nginx_access.log* | awk '{print $2}' | cut -d/ -f 1,2,3,4,5 | sort | uniq -c | sort -nr` on the past two weeks of logs. Everything with > 1000 hits

```
21136256 /a/icds-cas/receiver/secure
4053911 /a/icds-cas/apps/download
1758144 /a/icds-cas/phone/restore
 740852 /hq/multimedia/file/CommCareAudio
 275377 /a/icds-cas/phone/keys
 243680 /hq/multimedia/file/CommCareImage
  39699 /a/icds-cas/receiver/9c3c742c280ef42b5a941496bb4de626
  24762 /accounts/login/
   6473 /hq/multimedia/file/CommCareVideo
   1345 /
   1329 /a/icds-cas/apps/dozZyXC.
```

Seems really strange that /hq/multimedia is being hit. I thought that those would only be hit on new installs and that all the phones had the app preinstalled. Either way it seems inefficient to have django handle those since they're just static files.

I'm also assuming /phone/keys is part of authentication.

Authentication and app installs are not rate limited
Restores are 100 r/s with 100 burst
Submissions are 100 r/s with 0 burst
Everything else is 100 r/s with 0 burst

@dimagi/scale-team

### DIFF
--- a/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/ansible/roles/nginx/vars/cas_ssl.yml
@@ -12,10 +12,6 @@ nginx_sites:
       hosts: formplayer
       port: "{{ formplayer_port }}"
    limit_req_zones:
-    - name: "authentication"
-      zone: "$server_name"
-      size: "10m"
-      rate: "500r/s"
     - name: "restore"
       zone: "$server_name"
       size: "10m"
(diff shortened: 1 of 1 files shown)


snopoke:
That's pretty interesting. The app installs are mostly because they've been re-installing as a way to try and make the app work and because a small % don't have it installed (for some unknown reason).

I'm not sure that we want to rate limit app installs at all since any failure there will make the entire install fail. Also not sure about auth - I think probably OK not to rate limit it at all.

Re app installs - can we not put the CCZ contents on the proxy and have nginx serve it directly?

emord:
@snopoke Updated to not rate limit app installs or auth. We may be able to do a one off for the icds server to put those files on the proxy. It's not clear how easy that will be as they are all stored in riak right now

PR #727, opened 2017-05-02
datadog integrations tweaks

snopoke:


ansible/roles/datadog/templates/process.yaml.j2
```diff
@@ -4,37 +4,56 @@
 init_config:
 
 instances:
+{% if datadog_integration_celery %}
```
benrudolph: interesting, i had assumed it would pretty much be a noop if the process didn't exist on that server. did you find otherwise?

snopoke:
it shows up as errors in the logs and datadog UI. Need to fix case when
there aren't any processes to be monitored on the host.

PR #729, opened 2017-05-02
couch config

snopoke:
@gcapalbo


ansible/roles/couchdb/templates/couchdb.config.j2
```diff
@@ -5,4 +5,4 @@ COUCHDB_STDOUT_FILE=/dev/null
 COUCHDB_STDERR_FILE=/usr/local/var/log/couchdb/couch.err
 COUCHDB_RESPAWN_TIMEOUT=5
 COUCHDB_OPTIONS=
-ENV_ERL_FLAGS="+A 48"
+ENV_ERL_FLAGS="+A 16"
```
gcapalbo: for some reason when i try to examine how many threads there are, htop shows a number of threads that is always 20 more than what we put here. so when i saw it was 24 before i think we were legitimately only running 4. i bumped this to 16 for now so it's 1 per core

ansible/roles/couchdb/templates/couchdb.bin.j2
```diff
@@ -21,7 +21,7 @@ ERL_OS_MON_OPTIONS="-os_mon \
     disk_space_check_interval 1 \
     disk_almost_full_threshold 1"
 ERL_START_OPTIONS="$ERL_OS_MON_OPTIONS -sasl errlog_type error +K true +A 4"
-HEART_BEAT_TIMEOUT=11
+HEART_BEAT_TIMEOUT=120
```
gcapalbo: maybe there's a better way to do this by updating this script and pulling from a different location but for now i wanted to get this out. there's no way i could find to configure this param out of the box

ansible/roles/couchdb/templates/couchdb.init.j2
```diff
@@ -88,6 +88,9 @@ start_couchdb () {
         chown $COUCHDB_USER "$RUN_DIR"
     fi
     command="$COUCHDB -b"
+    if test -n "$ENV_ERL_FLAGS"; then
+        command="env ERL_FLAGS=\"$ENV_ERL_FLAGS\" $command"
+    fi
```
gcapalbo: the actual command gets run with su, which cleans the environment variables before running the command. so it had to be added explicitly to the env variables for the command. i renamed it to `ENV_ERL_FLAGS` to avoid confusion

gcapalbo:
(just a little cleanup left like adding the path for the bin file)

emord:
Deployed this to couch0 and pillow1 (no significant changes from what was already there)

emord:
Should we make some of these params configurable? I imagine that swiss won't handle a lot more threads very well (not that we'll redeploy that anytime soon) and we may want different configs for l10k and enikshay. Maybe it could wait until couch 2 though

gcapalbo:
@emord yeah i had thought about parameterizing some of these values. i think before this gets merged we should definitely do that. in the meantime i think we're still trying to figure out what's helping and what's not and which ones we'd need to parameterize or not

PR #732, opened 2017-05-02
disable transparent huge pages THP

javierwilson:
this is for DBs in ICDS

### DIFF
--- a/ansible/deploy_db.yml
+++ b/ansible/deploy_db.yml
@@ -126,5 +126,16 @@
     - name: Sets THP defrag to never
       raw: echo never > /sys/kernel/mm/transparent_hugepage/defrag
       when: thp_defrag.stat.exists
-
-  
+    - name: Adds transparent_hugepage=never to default grub config
+      lineinfile:
+        dest: /etc/default/grub
+        regexp: '^GRUB_CMDLINE_LINUX_DEFAULT=""$'
+        line: 'GRUB_CMDLINE_LINUX_DEFAULT="transparent_hugepage=never"'
+        backrefs: yes
+        state: present
+    - name: Adds transparent_hugepage=never to current grub config
+      replace:
+        dest: /boot/grub/grub.cfg
+        regexp: '(.*)linux(.*)/boot/vmlinuz(.*) ro$'
+        replace: '\1linux\2/boot/vmlinuz\3 ro transparent_hugepage=never'
+        backup: yes
--- a/ansible/deploy_db.yml
+++ b/ansible/deploy_db.yml
@@ -99,3 +99,24 @@
     - celery
   roles:
     - {role: java, tags: java}
+
+# Disables transparent huge pages (THP)
+- name: Disable THP
+  become: true
+  hosts:
+    - postgresql
+    - couchdb
+    - redis
+    - elasticsearch
+    - rabbitmq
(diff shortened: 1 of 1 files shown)


ansible/deploy_db.yml
```diff
@@ -99,3 +99,32 @@
...
+  tasks:
+    - stat:
+        path: /sys/kernel/mm/transparent_hugepage/enabled
+      register: thp_enabled
+    - stat:
+        path: /sys/kernel/mm/transparent_hugepage/defrag
+      register: thp_defrag
+    - name: Sets THP enabled to never
+      raw: echo never > /sys/kernel/mm/transparent_hugepage/enabled
+      when: thp_enabled.stat.exists
```
calellowitz: if those files dont exist does that mean it is not enabled?

ansible/deploy_db.yml
```diff
@@ -99,3 +99,32 @@
...
+  become: true
+  hosts:
+    - postgresql
+    - couchdb
+    - redis
+    - elasticsearch
+    - rabbitmq
+    - zookeeper
+    - riakcs
+    - formplayer
```
emord: If this is just for dbs, why are rabbitmq, zookeeper and formplayer here?

ansible/deploy_db.yml
```diff
@@ -99,3 +99,43 @@
...
+    - stat:
+        path: /sys/kernel/mm/transparent_hugepage/defrag
+      register: thp_defrag
+    - name: Sets THP enabled to never
+      raw: echo never > /sys/kernel/mm/transparent_hugepage/enabled
+      when: thp_enabled.stat.exists
+    - name: Sets THP defrag to never
+      raw: echo never > /sys/kernel/mm/transparent_hugepage/defrag
+      when: thp_defrag.stat.exists
+    - name: Adds transparent_hugepage=never to default grub config
```
emord: If this causes an error in the grub config that causes us to not be able to boot, what would be the best way to troubleshoot it?

ansible/deploy_db.yml
javierwilson: actually, i just copied the list from the "Common Database Machine Setup"

ansible/deploy_db.yml
javierwilson: very true, i've tried grub-script-check but doesn't seem very helpful.  I am leaning towards leaving grub.cfg alone and just change /etc/default/grub and let the job of configuring grub to update-grub.

javierwilson: just for our curren peace of mind, i tried:
grub-mkconfig > /tmp/grub.tmp
diff /tmp/grub.tmp /boot/grub/grub.cfg
with no difference, but as I said I think it's better to leave grub.cfg alone and let grub re-configure it with the parameter in /etc/default/grub which syntax is a lot easier to check for humans.

javierwilson:
I have removed the modification to the current grub config, and left a comment about grub-mkconfig, which is always safer.

ansible/deploy_db.yml
benrudolph: if this is in general good for database access, then it may also be good for formplayer due to its frequent access to sqlite. though it's not clear to me what the tradeoffs are for other things like running a normal web server

PR #734, opened 2017-05-04
formplayer needs couch-users db

sravfeyn:


snopoke:
This should work for ICDS but won't for any other ENV's since `couchdbs.users` won't exist. We probably have to add another setting telling ansible which CouchDB should be used by Formplayer:
```
couchdbs[formplayer_couchdb].host
```

sravfeyn:
@snopoke I have updated rest of env settings to have `couchdb.users` and point by default to the default db. Is this better.

dannyroberts:

sravfeyn:
Probably not needed, thanks for pinging Danny!

PR #737, opened 2017-05-08
compact synclogs more

emord:
@dimagi/scale-team 

Was looking at https://app.datadoghq.com/metric/explorer?live=true&page=0&is_auto=false&from_ts=1491675022153&to_ts=1494267022153&tile_size=m&exp_metric=couchdb.by_db.disk_size&exp_scope=&exp_group=db&exp_agg=avg&exp_row_type=metric#db:commcarehq_synclogs and saw that synclogs gets to almost 300 GB and hasn't been compacted ~1.5 weeks.

I think this should make compaction happen around every 4 days or so which seems better. Could probably do something more aggressive if we wanted as well

### DIFF
--- a/ansible/roles/couchdb/templates/local.ini.j2
+++ b/ansible/roles/couchdb/templates/local.ini.j2
@@ -22,6 +22,7 @@ bind_address = {{ inventory_hostname|ipaddr }}
 socket_options = [{recbuf, 262144}, {sndbuf, 262144}, {nodelay, true}]
 
 [compactions]
+commcarehq__synclogs = [{db_fragmentation, "50%"}, {view_fragmentation, "50%"}, {from, "17:00"}, {to, "22:30"}]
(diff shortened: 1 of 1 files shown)


benrudolph:
wonder if we could just remove all the attachments. we no longer save them to the sync log. it's unfortunate they hang around taking up space

emord:
@benrudolph I didn't look too much further into that. There's a card in the backlog to purge old synclogs so I'm leaving it for whoever picks up that card https://trello.com/c/ylowaGGK/90-purge-old-synclogs

benrudolph:
👍 mostly musing

emord:
I haven't done this yet since I haven't wanted to take down the site. @javierwilson could you deploy this whenever you do the resource reclaiming on couch0 so we only go down once?

javierwilson:
i'll do it, tomorrow, I hope.

PR #738, opened 2017-05-09
Allow executable

benrudolph:
@millerdev @javierwilson was debugging why formplayer wasn't working on swiss. it turns it out it couldn't download the application because when it'd attempt to, the nginx worker (user `www-data`) didn't have permissions to open the file. you cannot read a file unless you have access to the dir, so i figured that was the intent of the permissions already


benrudolph:
this isn't an issue on prod because the `nfs` has different `user:group` owners

PR #743, opened 2017-05-11
exclude integrations from swiss

snopoke:
I don't think we're using this data at all so removing it.


snopoke:
I think for addis we'd want to start with full monitoring and only remove
it if we find we don't need it.

PR #745, opened 2017-05-15
add setting couch cloudant domain

snopoke:
https://github.com/dimagi/datadog-checks/pull/7


snopoke:
after further digging I think what we're experiencing is a change in cloudant monitoring api

PR #748, opened 2017-05-15
couch2

emord:
@dimagi/scale-team here's the start of the couch2 ansible work

The first commit is just installing some roles from galaxy my changes from that are in the commits after that. (https://github.com/dimagi/commcarehq-ansible/compare/e6d12ea...je/couch2)

This uses a new inventory (couchdb2) and will use different ports so that we can run couch 2 and 1.6 at the same time and we can also run multiple nodes on the same machine (likely what we'll do until the new infrastructure)

Things to do:

- [ ] add arguments to run multiple nodes on the same machine
- [x] change nginx config to support endpoints and load balance
- [ ] remove changes to vagrant file
- [ ] standardize couchdb2 usage (vs couch2)


snopoke:
Hey. Is it necessary to copy the entire role into our own repo? Can we not just install it as a dependency (https://docs.ansible.com/ansible/galaxy.html#installing-roles).

I know that this is a longer route but if there are changes that we need to make to the imported role (e.g. making the install location configurable) it would be better to PR those to his repo.

For additional things we need like creating users we can put those in our own role that depends on his one. 

Keen to hear your thoughts on this.

emord:
@snopoke Ah i see. I used the ansible-galaxy install to do this. I'll look into changing it to a dependency agreed that's a better route

emord:
I'm currently requiring some cloned copies of the roles from my own account. I will PR them to his once I am done iterating

emord:
(closing during cleanup)

PR #751, opened 2017-05-18
couch2

emord:
Cleaned this up so that it's easy to follow up commit by commit. This PR will not change what the workers point to. that can come in a later PR

I think that I've made any options configurable that we need from https://github.com/dimagi/commcarehq-ansible/pull/729/files, but I haven't implemented those configuration changes (like changing # threads to 64 and not logging to couch.out) Because I think it would be good to keep the defaults until it's proven that couch2 fails on those. Would be great if @gcapalbo or @snopoke can verify that the repositories contain the appropriate ways to configure their changes.

I'll PR the following changes into andrew's repository once we're happy with the configuration changes made (didn't want too much churn on an external repo)

https://github.com/andrewrothstein/ansible-couchdb/compare/master...emord:master
https://github.com/andrewrothstein/ansible-couchdb-cluster/compare/master...emord:master

buddy @calellowitz


snopoke:
LGTM

emord:
On initial deploy it's necessary to wget https://www.apache.org/dist/couchdb/source/2.0.0/ to tmp. This is because we are deploying to trusty and the default python (2.7.6) is not capable of using SNI.

I also changed the nginx proxy's couchdb port to 25984 because we need to run couch 1 couch2 and couch2 proxy on one machine for l10k for now

emord:
Updated with some more config changes that I found after load testing

gcapalbo:
@emord to answer your question about the configuration - yes it looks like everything we had done before is still configurable now except for the heartbeat timeout (which i think is fine).

i also noticed that setting the log level to warning is possible but is not being overridden yet (which i think is also fine).

emord:
@gcapalbo thanks for double checking. I'm doing that intentionally and have included in the trello https://trello.com/c/A83zKCac/87-couch-2-0

> change couch logging to notice or warn (to be done after assuring everything's ok on NIC)

gcapalbo:
@emord sounds good. i'm not sure there was a huge difference when we set it to warning, so i'm even fine leaving as is unless you think it's too much output

emord:
Travis isn't running tests for this PR for some reason. These commits have passed tests in #771 though and this is deployed to ICDS right now so going to merge

PR #752, opened 2017-05-19
Fix host spoofing

emord:
https://github.com/yandex/gixy/blob/master/docs/en/plugins/hostspoofing.md

ran gixy on my local vagrant setup and saw this error that's detailed in the link above. Should probably get tested on staging first

Also saw

```                                                                                                                                                             
Problem: [add_header_redefinition] Nested "add_header" drops parent headers.                                                                                                                                       
Description: "add_header" replaces ALL parent headers. See documentation: http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header                                                                    
Additional info: https://github.com/yandex/gixy/blob/master/docs/en/plugins/addheaderredefinition.md                                                                                                               
Reason: Parent headers "x-frame-options" was dropped in current level                                                                                                                                              
Pseudo config:                                                                                                                                                                                                     
include /etc/nginx/sites-enabled/dev_commcare;                                                                                                                                                                     
                                                                                                                                                                                                                   
        server {                                                                                                                                                                                                   
                server_name 192.168.33.21;                                                                                                                                                                         
                add_header X-Frame-Options SAMEORIGIN;                                                                                                                                                             
                                                                                                                                                                                                                   
                location /static {                                                                                                                                                                                 
                        add_header Access-Control-Allow-Origin *;                                                                                                                                                  
                }                                                                                                                                                                                                  
        }                                                                                                                                                                                                          
```

but that seems like less of a priority. it just means that `add_header X-Frame-Options SAMEORIGIN;` has no effect

buddy @calellowitz


emord:
deployed to staging and looks like it's working
will merge master and deploy prod/softlayer tomorrow

emord:
deployed and looking good

PR #756, opened 2017-05-22
remove custom ICDS logging

snopoke:
Now the same across all envs: https://github.com/dimagi/commcare-hq/pull/16258


sravfeyn:
👍

PR #754, opened 2017-05-22
l10k inventory

javierwilson:
initial testing environment

PR #759, opened 2017-05-23
l10k couch2

emord:
PRing this into @javierwilson branch #754 

This is what I deployed to l10k today. There's a lot of commits here, but it's really just master + #751 + https://github.com/dimagi/commcarehq-ansible/commit/cb1593d64efa52b5d9060aef731bcd383e32bbcf.

Currently l10k is using couch2 with nginx working as a proxy/load balancer (but only one node). I made this change manually to the localsettings.py (just changed the couch url's port to 25984).

I also set up replication between both couch1 and couch2, so any change to couch1 should be reflected in couch2 and vice versa.

I think this should be ready for any load testing that needs to be done (assuming the rest of this deploy is finished)

@dimagi/scale-team fyi


snopoke:
Looking good. Didn't review closely though. What's the need for the bi-directional replication (or even couch1 at all)? Is it just to test it out for the ICDS migration?

Also what changed in the vault files?

emord:
@snopoke no need to review closely. This is going to javier's branch. I think most of those changes are just from commits from master that are not in his branch + my couch2 changes from my other PR. 

I set up bidirectional replication to test out icds. I also thought since we are load testing we may want a comparison between couch1 and 2 and that would be the easiest way. I can take down couch1 if we don't think that's necessary

snopoke:
That all sounds fine. Good to get couch1 vs couch2 test for comparison.

ansible/roles/couchdb2/vars/main.yml
```diff
@@ -1,16 +1,16 @@
 ---
 couchdb_port : 15984
 couchdb_node_port : 15986
-couchdb_dir: /usr/local/couchdb2/couchdb
+couchdb_dir: /usr/local/couch2/couchdb
```
emord: I ended up changing this after deploying to l10k. I'll clean this up this week.

PR #761, opened 2017-05-24
cannot set mode if no swapfile

javierwilson:


javierwilson:
travis error seems unrelated, a network error

ansible/roles/swap/tasks/main.yml
```diff
@@ -20,7 +20,7 @@
 - name: Set swap file mode
   become: true
   file: path=/swapfile state=file mode=600
-  when: swap_size is defined and not swapfile.stat.exists
+  when: swap_size is defined and swapfile.stat.exists
```
millerdev: I think this `Set swap file mode` task should be moved down below the `Make swap` task. Also, I don't think this change is correct. Instead we should change this to simply `when: swap_size is defined`

javierwilson: the file itself is created at L14 `Create swap space`, I do not see why move it down, I have had trouble with this task before, but you're right, swapfile should exist. Also `Make swap` has `when: swap_size is defined and not swapfile.stat.exists` which makes no sense to me... I may be reading the whole file wrong.  Or maybe it ended up working because I had to run it a second time.

PR #763, opened 2017-05-24
not all servers use /etc/default/grub (softlayer)

javierwilson:


kaapstorm:
Huh! Which servers don't use GRUB? ... Or they do use GRUB, but it doesn't read `/etc/default/grub`? (Not important, just curious.)

javierwilson:
i think they use grub, just not /etc/default/grub, they may use /boot/default/grub, didn't look too deep into it.  but they do use grub, as far as i could tell, lilo days are over.

PR #764, opened 2017-05-24
set flower url for icds

snopoke:
This is used by the datadog integration as well. We could separate it into 2 different settings but waiting on feedback from @dannyroberts as to why this was commented out in the first place: https://github.com/dimagi/commcarehq-ansible/commit/0481f094022826160469637bee036e12fcf80dc2


dannyroberts:
long story short, I'm really not sure why I commented it out in the first place—probably to make it past some ansible step at the time I was helping set up icds cluster—but I'm guessing this should be fine, and if it's not you'll find it out quickly and probably painlessly :)

PR #769, opened 2017-05-26
final step to remove old riak cluster in softlayer

javierwilson:


javierwilson:
@calellowitz  please review

PR #776, opened 2017-06-02
couch2 improvements

emord:
@dimagi/scale-team deployed these changes to icds

buddy @orangejenny

### DIFF
--- a/ansible/roles/nginx/vars/couchdb2.yml
+++ b/ansible/roles/nginx/vars/couchdb2.yml
@@ -18,6 +18,8 @@ nginx_sites:
        balancer: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
        proxy_redirect: ' off'
        client_max_body_size: ' 10m'
+       client_body_buffer_size: ' 512k'
+       proxy_max_temp_file_size: ' 0'
      - name: "~ ^/(.*)/_changes"
        balancer: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
        proxy_redirect: ' off'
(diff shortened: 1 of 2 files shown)


ansible/roles/nginx/vars/couchdb2.yml
```diff
@@ -18,6 +18,8 @@ nginx_sites:
        balancer: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
        proxy_redirect: ' off'
        client_max_body_size: ' 10m'
+       client_body_buffer_size: ' 512k'
```
emord: Needed for some puts/posts to couch

@millerdev I noticed some similar logs in the riak errors on nic. I assume the same is happening on prod/softlayer. Do you think it would be good to bump this up for riak as well? https://serverfault.com/questions/511789/nginx-client-request-body-is-buffered-to-a-temporary-file

ansible/roles/nginx/vars/couchdb2.yml
```diff
@@ -18,6 +18,8 @@ nginx_sites:
...
+       proxy_max_temp_file_size: ' 0'
```
emord: Needed for large responses coming out of couch

orangejenny: what's up with the leading spaces in values in this file? i'm assuming it's needed, but why?

emord: I don't think that they're required. I started this file by copying another file and it had a space there. i've been meaning to go back and make sure they aren't required

PR #777, opened 2017-06-02
update flower url

emord:
@javierwilson this changed with your recent pr. I fixed this on the webworkers manually

buddy @orangejenny


gcapalbo:
@javierwilson did this get deployed to all softlayer machines? i think only the webworkers were updated manually but the rest of them would need it too.

javierwilson:
done

PR #781, opened 2017-06-05
support http2 where possible

emord:
saw this in a blog post the other day and was surprised how easy it could be.

some of our servers are using 1.8.1 which doesn't support this (>1.9.5)

@orangejenny


emord:
Actually to do this we would need to upgrade our openssl to 1.0.2 library which would require upgrding ubuntu... not going to go down that road @ctsims fyi since you asked about this

PR #782, opened 2017-06-06
Localsettings tags

benrudolph:
@emord as discussed. if it looks ok, i'll notify the team so people are aware

cc: @mkangia

### DIFF
--- a/ansible/roles/commcarehq/tasks/main.yml
+++ b/ansible/roles/commcarehq/tasks/main.yml
@@ -93,6 +93,7 @@
     - { path: "{{ code_home }}", deploy_env_name: "{{ env_name|default(deploy_env) }}" }
   tags:
     - localsettings
+    - hq-localsettings
 
 - name: chown localsettings
   become: true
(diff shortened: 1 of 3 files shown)


emord:
so localsettings is everything and now there's hq- settings to only deploy those? that looks good to me

benrudolph:
that's correct

@dimagi/team-commcare-hq fyi if you deploy `localsettings` it will now also update formplayer and touchforms. This is so they get proper machine updates when machines change. If you want to only deploy hq localsettings, use the `hq-localsettings` tag

millerdev:
@benrudolph what does this mean for a person who doesn't know much about formplayer and touchforms? Should we just continue to deploy `localsettings` and not worry about the fact that it's doing more than just `localsettings`? Or is `hq-localsettings` the new common case, and we should only deploy `localsettings` in special cases?

emord:
@millerdev I think usually formplayer settings should not change, or it should be obvious why they change. If it's not there's a backup for deploying a subset of these changes (like hq-localsettings)

this came up because I deployed localsettings to NIC but not formplayers. As a result formplayer was authenticating against couch 1 instead of couch. so I think this change is strictly better so things like that don't happen again

PR #783, opened 2017-06-07
give postgres all the memory it deserves (icds)

javierwilson:


emord:
@benrudolph we only have 32 GB on pg machines fwiw

benrudolph:
oh sorry meant 32GB, shared buffers is recommend to be 1/4 of the RAM

PR #785, opened 2017-06-08
icds test staging

emord:
@sheelio This will also need to change on icds and softlayer if you want to test there later

buddy @orangejenny


emord:
This has already been deployed to staging so going ahead with merge

PR #787, opened 2017-06-09
we don't need no java for web workers!

javierwilson:
also, automatic download of JDK7 is broken, which pushed to review if installing java was absolutely needed.  Time for openjdk?


benrudolph:
👍

PR #792, opened 2017-06-09
chown link to ~/www/env/current to cchq

emord:
I had to do this manually on the last couple of deploys since the webworkers were redeployed @calellowitz and I tried to fix this before during the scale up but the PR had other issues. I think this should work correctly

@orangejenny


orangejenny:
test failure related?

javierwilson:
i don't think you can recurse on a symlink, not sure. if you need to change owner of the symlink itself you may need to move this to line 45 where the symlink is created, unless this is already merged in 2.2.1 https://github.com/ansible/ansible/issues/21537

PR #793, opened 2017-06-09
creating users in read-only standbys fails

javierwilson:


calellowitz:
is this going to be a problem if we need to fail over to the standby

javierwilson:
but these users already exist in the standby, as long as they are created in the master

PR #794, opened 2017-06-14
deploys etckeeper

javierwilson:


ansible/deploy_etckeeper.yml
```diff
@@ -0,0 +1,8 @@
+---
+
+- name: Put /etc under version control using etckeeper
+  hosts: all
+  become: True
+  roles:
+    - role: debops-contrib.etckeeper
+      tags: 'etckeeper'
```
benrudolph: should this be referenced in `deploy_stack` so that it always gets deployed? or do you only want it only deployed when specified?

ansible/requirements.yml
```diff
@@ -2,3 +2,6 @@
 - src: git+https://github.com/emord/ansible-couchdb-cluster.git
   version: df205e7
   name: emord.ansible-couchdb-cluster
+- src: git+https://github.com/debops-contrib/ansible-etckeeper.git
```
snopoke: I think `src` can just be `debops-contrib.etckeeper` with `version: trusty`

ansible/deploy_etckeeper.yml
javierwilson: i thought about it, we want CVS as part of deploy_stack, however this specific role, I do want to test a bit more first, perhaps in the env I know the most or l10k.

ansible/requirements.yml
javierwilson: cool, i'm just getting familiar with galaxy...

PR #795, opened 2017-06-14
update couchdb role for service script

emord:
adds https://github.com/emord/ansible-couchdb-cluster/commit/826df6010e191f17c35be598e9e4ba79d538c47f

@dimagi/scale-team 

buddy @orangejenny

### DIFF
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -1,4 +1,4 @@
 ---
 - src: git+https://github.com/emord/ansible-couchdb-cluster.git
-  version: df205e7
+  version: 826df60
   name: emord.ansible-couchdb-cluster


benrudolph:
nice. do you know if it will retry after a period? or does this pretty much mean we'll have to bring up couch on restart?

emord:
@benrudolph that's something we'll need to solve elshwere in our stack. like after ecryptfs is run.

Right now we also have to start ES psql and riak manually

emord:
```
root@pillow1:~# service couchdb2 start
couchdb2 could not start because data directory does not exist
```
I've never been so excited for a database to fail to start

gcapalbo:
haha, nice!

PR #796, opened 2017-06-15
mail relay deploy should be part of deploy_stack

javierwilson:
besides installing postfix, this tasks are only played if mailrelay group is defined. not including this in deploy_stack before is what left the new webworkes with no email communication enabled.


sravfeyn:
Nice!

PR #798, opened 2017-06-15
set celery timezone for ICDS

snopoke:
I've tested this locally by finding the next fire time for a task which lined up with the scheduled time in IST.


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -279,6 +279,9 @@ CELERY_SEND_TASK_ERROR_EMAILS = {{ localsettings.get('CELERY_SEND_TASK_ERROR_EMA
 CELERY_RESULT_BACKEND = '{{ localsettings.CELERY_RESULT_BACKEND }}'
 CELERY_PERIODIC_QUEUE = '{{ localsettings.CELERY_PERIODIC_QUEUE }}'
 CELERY_FLOWER_URL = '{{ localsettings.CELERY_FLOWER_URL }}'
+{% if CELERY_TIMEZONE in localsettings and localsettings.CELERY_TIMEZONE %}
```
gcapalbo: looks like build failed, i think this needs to be `if 'CELERY_TIMEZONE' in`

PR #804, opened 2017-06-19
Combine static cases and child_health case pillows

emord:
@calellowitz I'll deploy this today


emord:
I'll also look at this tomorrow and may add anothr 8 partitions and processes to cases if it falls behind. Currently there's a lot of load on pillow0 and hoping this helps out by having fewer processes

PR #805, opened 2017-06-21
after reboot tasks

javierwilson:
first effort to try to automate tasks to be run after a machine is rebooted

### DIFF
--- a/ansible/roles/couchdb/tasks/main.yml
+++ b/ansible/roles/couchdb/tasks/main.yml
@@ -75,6 +75,7 @@
 
 - name: Start CouchDB service
   service: name=couchdb state=started
+  tags: after-reboot
 
 - name: Check CouchDB database
   uri:
--- a/ansible/roles/elasticsearch/tasks/main.yml
+++ b/ansible/roles/elasticsearch/tasks/main.yml
@@ -101,12 +101,14 @@
   service: name=elasticsearch state=started
   register: result
   failed_when: result.state != "started"
+  tags: after-reboot
 
 - name: Restart Elasticsearch
   sudo: yes
   service: name=elasticsearch state=restarted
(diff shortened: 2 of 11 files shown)


ansible/roles/elasticsearch/tasks/main.yml
```diff
@@ -101,12 +101,14 @@
...
   register: result
   failed_when: result.state != "started"
+  tags: after-reboot
 
 - name: Restart Elasticsearch
   sudo: yes
   service: name=elasticsearch state=restarted
   when: copy_elasticsearch_conf_result.changed or copy_elasticsearch_upstart_result.changed or copy_elasticsearch_logging_result.changed
   tags:
+    - after-reboot
```
millerdev: why is this one needed? seems odd to restart immediately after starting

ansible/roles/couchdb/tasks/main.yml
```diff
@@ -75,6 +75,7 @@
 
 - name: Start CouchDB service
   service: name=couchdb state=started
+  tags: after-reboot
```
emord: This will work for couch 1 (swiss) but not couch 2 (icds). confusing we currently have two couch roles for the moment

ansible/roles/riakcs/configure/tasks/main.yml
```diff
@@ -58,3 +58,4 @@
 - name: "Start {{ name }}"
   service: name="{{ service }}" state=started enabled=yes
   when: not riakcs_deb_check.get("skipped")
+  tags: after-reboot
```
emord: Does this start riak as well?

millerdev: yes, it should start all of riak, stanchion and riak-cs (assuming they're stopped)

ansible/roles/elasticsearch/tasks/main.yml
javierwilson: good catch

PR #808, opened 2017-06-21
uses tmpreaper instead of role celery_hq

javierwilson:


javierwilson:
this requires #807 , once that is merged, i will test again...

ansible/deploy_commcarehq.yml
```diff
@@ -16,4 +16,6 @@
     - celery
   sudo: yes
   roles:
-    - {role: celery_hq, tags: celery_hq}
+    - role: tmpreaper
+      tags: commcarehq, celery_hq, tmpreaper, cron
+      tmpreaper_time: '2d'
```
gcapalbo: do we need to specify the directory or does it only operate on /tmp?

gcapalbo:
:+1: looks good to me. so just looks like we're replacing the cron job with tmpreaper to clear tmp files?

also just curious if this has been tested

ansible/deploy_commcarehq.yml
javierwilson: yes, this works in /tmp and /var/tmp, we can add more dirs if neede

PR #810, opened 2017-06-21
ping hosts and checks status

javierwilson:
this is a first effort, more complex tests should be added.  this is like a os-level check_services

### DIFF
--- a/ansible/ping.yml
+++ b/ansible/ping.yml
@@ -51,10 +51,10 @@
       warn: False
 
 - hosts:
-  - zookeper
+  - zookeeper
   tasks:
-  - command: service zookeper status
-    name: Check zookeper status
+  - command: service zookeeper status
+    name: Check zookeeper status
     become: True
     args:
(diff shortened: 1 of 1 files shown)


ansible/ping.yml
```diff
@@ -0,0 +1,147 @@
...
+  - redis
+  tasks:
+  - command: service redis-server status
+    name: Check redis status
+    become: True
+    args:
+      warn: False
+
+- hosts:
+  - zookeper
```
benrudolph: `zookeeper`

ansible/ping.yml
```diff
@@ -0,0 +1,147 @@
...
+- hosts:
+  - shared_dir_host
+  tasks:
+  - command: service nfs-kernel-server status
+    name: Check NFS status
+    args:
+      warn: False
+
+- hosts:
+  - riakcs
```
benrudolph: could create a ping for `riak` as well

PR #812, opened 2017-06-21
Set boto3 signature_version

millerdev:
No idea why this is necessary, but recently a new machine was allocated on softlayer and it refused to connect to Riak CS.

Error on get object: `ClientError: An error occurred (InvalidRequest) when calling the GetObject operation: The authorization mechanism you have provided (AWS4-HMAC-SHA256) is not supported.`

On comparing this new machine (pillow3) to an existing machine (db0) I noticed they had slightly different client configs:

```
pillow3:
In [31]: bu.meta.client.meta.config._user_provided_options
Out[31]:
{'connect_timeout': 60,
 'max_pool_connections': 10,
 'read_timeout': 60,
 'region_name': u'us-east-1',
 's3': None,
 'signature_version': 's3v4',
 'user_agent': 'Boto3/1.4.4 Python/2.7.6 Linux/4.4.0-71-generic Botocore/1.5.71 Resource'}

db0:
In [12]: bu.meta.client.meta.config._user_provided_options
Out[12]:
{'connect_timeout': 60,
 'max_pool_connections': 10,
 'read_timeout': 60,
 'region_name': None,
 's3': None,
 'signature_version': 's3',
 'user_agent': 'Boto3/1.4.4 Python/2.7.6 Linux/3.13.0-65-generic Botocore/1.5.35 Resource'}
```

Setting `signature_version` in the config fixed the issue.

@nickpell @javierwilson

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -163,12 +163,14 @@ S3_BLOB_DB_SETTINGS = {
     'url': '{{ s3_blob_db_url }}',
     'access_key': '{{ s3_blob_db_access_key }}',
     'secret_key': '{{ s3_blob_db_secret_key }}',
+    'config': {'signature_version': 's3'},
 }
 {% if 'BLOB_DB_MIGRATING_FROM_S3_TO_S3' in localsettings %}
 OLD_S3_BLOB_DB_SETTINGS = {
     'url': '{{ old_s3_blob_db_url }}',
     'access_key': '{{ old_s3_blob_db_access_key }}',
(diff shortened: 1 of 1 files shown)


millerdev:
I don't think it's urgent to deploy this to existing environments if they're working.

emord:
Ah. I see it now. I think this is necessary because boto3 1.4.4 has a requiremnt of botocore>=1.5.0,<1.6.0. botocore 1.5.71 must have changed the default signature version in a version since .35.

thanks for digging into it

millerdev:
bad outcome of a patch version bump. according to semantic versioning that should only contain backwards-compatible bug fixes

PR #814, opened 2017-06-22
update cache before nfs and install openjdk

emord:
couple of issues we ran into when deploying the new celery server

@sravfeyn this second commit should install openjdk 7 instead of trying oracle. will likely be needed to install elasticsearch tomorrow

@calellowitz


ansible/roles/shared_dir/tasks/install.yml
```diff
@@ -1,5 +1,10 @@
 ---
 
+- name: Update the apt cache
+  when: shared_drive_enabled
+  apt:
+    update_cache: yes
+
```
javierwilson: I think you should be able to add "update_cache: yes" to the apt task below.

ansible/roles/java/tasks/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Install Java
   sudo: yes
-  script: ../files/install_java.sh creates=/usr/lib/jvm/success.txt
+  apt: name=openjdk-7-jdk state=present
```
benrudolph: 😂

ansible/roles/shared_dir/tasks/install.yml
emord: Since it uses with_items would that try to call `apt-get update` twice? that was the original reason I moved it here

PR #815, opened 2017-06-22
Keep fewer proxy logs around

emord:
Softlayer's proxy was keeping 14 rotations for riak. seems excessive. not sure why the other proxy's have less rotations, but I think it's related to the logrotate not applying correctly

@calellowitz @millerdev

### DIFF
--- a/ansible/deploy_proxy.yml
+++ b/ansible/deploy_proxy.yml
@@ -76,7 +76,7 @@
           options:
             - monthly
             - size 750M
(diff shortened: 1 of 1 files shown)


millerdev:
As discussed on slack, I think we might want to keep proxy logs around for a while as an audit trail. I'm not sure if or where such a log retention policy is defined though, and I'm not sure who to ask about it.

PR #823, opened 2017-06-29
Updated couch.yaml for new datadog integration.

emord:
This is going to need the username for the new couch2 datadog integration.

buddy @esoergel

### DIFF
--- a/ansible/roles/datadog/templates/couch.yaml.j2
+++ b/ansible/roles/datadog/templates/couch.yaml.j2
@@ -4,5 +4,7 @@ instances:
 {% for couch_config in couch_dbs.values() %}
 {% if couch_config.host == inventory_hostname %}
     - server: "http{% if couch_config.is_https %}s{% endif %}://{{ couch_config.host }}:{{ couch_config.port }}"
+      user: {{ couch_config.username }}
+      password: {{ couch_config.password }}
 {% endif %}
 {% endfor %}


ansible/roles/datadog/templates/couch.yaml.j2
```diff
@@ -4,5 +4,7 @@ instances:
 {% for couch_config in couch_dbs.values() %}
 {% if couch_config.host == inventory_hostname %}
     - server: "http{% if couch_config.is_https %}s{% endif %}://{{ couch_config.host }}:{{ couch_config.port }}"
+      user: {{ couch_config.username }}
+      password: {{ couch_config.password }}
```
esoergel: Does this mean that datadog will have write access to our prod couch db?  Is it possible to do read-only access or something, or is write access necessary for the integration?

emord: This will only be for ICDS it won't be deployed on prod.

The reason we need the user/pass is because in couch2 to find all the nodes you need to request couch/_membership which is only available to server admins. it may be possible to find a way to specify a locked down role.

Luckily I've touched a majority of the datadog integration personally so I know it only hits _membership and _stats. which are read only dbs

emord: I think ddog also has write access to ES redis and maybe riak as well. So its something we should definitely look into

PR #826, opened 2017-07-06
Updated couch proxy to be on couch0.

emord:
couch0 has always been running nginx for couch2 because I set it up in case anything went wrong during the initial deploy. @dimagi/scale-team

### DIFF
--- a/ansible/vars/icds/icds_public.yml
+++ b/ansible/vars/icds/icds_public.yml
@@ -242,7 +242,7 @@ couchdb2:
 
 couch_dbs:
   default:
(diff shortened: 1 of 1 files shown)


ansible/vars/icds/icds_public.yml
```diff
@@ -242,7 +242,7 @@ couchdb2:
 
 couch_dbs:
   default:
-    host: 10.247.24.13
+    host: 10.247.24.18
```
emord: in the future this should probably be changed to use someting from the inventory instead of an IP

PR #828, opened 2017-07-09
use md5 hash of password if available

mkangia:
Minor change for password for enikshay private API password.


mkangia:
I tested this with a --check run and looks fine. Merging it.

PR #829, opened 2017-07-09
moving kafka logs to /opt/data/kafka

javierwilson:
applied for softlayer, i applied same solution we sued for icds, however, i think kafka "logs" should probably not be in the same diretory as kafka logs which should be moved back to /var/log/kafka


javierwilson:
I am just moving them from /var/log to /opt/data which is not encrypted, do you think this should be encrypted. I will write another PR to split logs from kafka logs.  And yes, still /opt/data/ecrypt :(

PR #831, opened 2017-07-10
add support for different 50x.html pages

dannyroberts:
and use icds branding on icds

Puts https://github.com/dimagi/commcare-hq-errorpages/pull/12 into action


dannyroberts:
Already deployed on ICDS

PR #834, opened 2017-07-11
Added off peak timings for ICDS.

emord:
for https://github.com/dimagi/commcare-hq/pull/16967

@dimagi/scale-team


gcapalbo:
i might have missed but how are we setting the autoscaler setting, or is that coming in a different pr

emord:
@gcapalbo yeah. this is one of htose changes that spanned all 3 repos https://github.com/dimagi/commcare-hq-deploy/pull/321 that sets the laoder. the loader sets the autoscaler https://github.com/dimagi/commcare-hq/pull/16967/files#diff-c6ed677c387177036e5f5d81d23d7a6aR267

gcapalbo:
nice, thanks

PR #835, opened 2017-07-12
Je/utc

emord:
@dimagi/scale-team The celery worker schedules tasks based on IST, but I used `datetime.utcnow()` to do the comparison. 

Could possibly change the loader to use the timezone, but would have to be careful to get the celery setting instead of the server setting as the server is still in UTC

buddy @esoergel


snopoke:
Would be useful to update the comment in `settings.py`

emord:
@snopoke will do. Do you have any strong feelings between using:

1) UTC always
2) Server time
3) Celery time

Seems fine to use #1 to me since this is a localsetting anyways, but it sucks that there's 3 possibly different times (although now I think 1 and 2 are always the same)

snopoke:
I think we should never use `2` and can use either `1` or `3` depending on the situation. 

In this case I don't think it matters as long it's clear which is the case. If we were to use celery time here it would make the code a bit more complex so I'm happy to go with UTC.

PR #836, opened 2017-07-12
Point dashboard to real data.

emord:
@sheelio This should probably change in sol dev's code eventually but it's fine for now

buddy @esoergel


esoergel:
This seems potentially dangerous - is it possible to wipe the "test" database?  I'm okay doing it if we're confident that this isn't any more dangerous, but it feels sketchy to have something called "test" pointing to the real deal.

emord:
@esoergel the databases this points to gets wiped every night by sheel's aggregation script. 

I think this alias was originally created so that we could have old v1 tables alongside new v2 tables on softlayer for testing. this is the alias that's in the router https://github.com/dimagi/commcare-hq/blob/master/corehq/sql_db/routers.py#L65-L67

PR #844, opened 2017-07-13
Restart ucr indicator queue every 20 min

emord:
This is not a great solution but I've tried a few things with no success.

Added extra logging to the autoscaler and tried to force it to stay at max concurrency: https://github.com/dimagi/commcare-hq/compare/je/log-more-for-autoscaler

from the logs it appears that even when we tell celery to scale down, it ignores the request and keeps 16 workers up.  those workers eventually become idle and do nothing while the load goes down to 0.

restarting the queue works for now @gcapalbo

### DIFF
--- a/ansible/roles/commcarehq/tasks/main.yml
+++ b/ansible/roles/commcarehq/tasks/main.yml
@@ -110,3 +110,26 @@
   tags:
     - localsettings
     - hq-localsettings
+
+- name: Copy celery restart script
+  sudo: yes
+  template:
+    src: "restart_celery.sh.j2"
+    dest: "/root/restart_celery.sh"
+    group: root
+    owner: root
+    mode: 0700
+    backup: yes
+  tags:
+    - hq-cronjob
+
+- name: Create Cron job
+  sudo: yes
+  cron:
+    name: "Restart celery ucr indicator"
+    job: /root/restart_celery.sh
+    user: root
+    cron_file: restart_celery
+    minute: "*/20"
+  tags:
+    - hq-cronjob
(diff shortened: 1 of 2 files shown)


emord:
I also attempted to lower max tasks per child. when that equaled 2, celery1 went to 0 load after 20 min (which is where 20 min came from)

gcapalbo:
shoot, that's too bad. i think we might have to be careful with restarting the worker so often though, as it might create a lot of zombie processes, or processes that are left finishing ongoing tasks. does it have to be every 20 minutes?

emord:
updated to one hour

emord:
@gcapalbo yeah. I want to dig in a little more digging into celery and possibly upgrade our libraries (although I don't think that will help).

using a cron to start and stop the queue is not a bad idea for the web workers. i was hoping to do the laod based so we could relibably predict the load and not interfere with other tasks, but looks like that may not work

ansible/roles/commcarehq/tasks/main.yml
```diff
@@ -110,3 +110,26 @@
   tags:
     - localsettings
     - hq-localsettings
+
+- name: Copy celery restart script
+  sudo: yes
+  template:
+    src: "restart_celery.sh.j2"
+    dest: "/root/restart_celery.sh"
```
javierwilson: This is not very importat, but just to keep things tidy, this should either go to /usr/local/sbin, if for some reason, we prefer to keep them at root's home, they should go in /root/bin/

emord: Sounds good @javierwilson I was copying htis from the couch restarter scrpit. will be cleaning it up today

emord:
Removed this from the celery workers.

PR #846, opened 2017-07-14
cron task to start/stop celery queues

emord:
taking @gcapalbo idea from my other PR. this will start at 1230 UTC (1800 IST) and stop at 230UTC (8 IST)

could be dryer but I don't think this should last very long (or get deployed to other environments) because a deploy between those hours would start all of the queues. safe for now because we don't generally deploy during those times.


emord:
I also moved the queues off of the load based scaler https://github.com/dimagi/commcare-hq-deploy/pull/329/files

emord:
Removed from ICDS

PR #848, opened 2017-07-17
fix couchdb2 requirement

javierwilson:
:( had the wrong repo
@mkangia thisis important to fix and do a  ansible-galaxy install --force

### DIFF
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -2,6 +2,9 @@
 - src: git+https://github.com/andrewrothstein/ansible-couchdb-cluster.git
   version: b7253a9f980a0a9a86f68881047775dff8aa554d
   name: emord.ansible-couchdb-cluster
+- src: git+https://github.com/andrewrothstein/ansible-couchdb.git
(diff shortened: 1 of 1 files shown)


javierwilson:
@emord actually the problem is (i think) that the couchdb-cluster repo requires the couchdb repo, but not the latest one, so it keeps pulling the one which doesn't create the parent dir :( We can get around it, re-installing the appropiate couchdb repo manually. or fixing this in your repo.

emord:
@javierwilson I think we could also require the both repositories in our requirements yml?

javierwilson:
yes, as long as we require the cloudb-cluster repo first, it will work, still a force install is required. testing and commiting...

javierwilson:
@emord this fixes the requirements issue, there's still an issue regarding the data directory not existing, but that will go in a different PR

emord:
@javierwilson ah. i think I know what happened with the data directory. originally I had the data directory created by the ansible task, but the maintainer asked me to put it in the init script. I removed that when I discovered that that would create the directory even if the encrypted drive wasn't mounted. I probably didn't add that back to the ansible task.

PR #851, opened 2017-07-19
Added pg stats to ICDS and softlayer UCR.

emord:
@nickpell


emord:
(this is on icds. haven't put it on softlayer yet)

emord:
Removed changes to softlayer since it will get migrated to enikshay cluster later anyways

PR #855, opened 2017-07-21
couchdb2 restarter

snopoke:
No relaxation happening here :x: :relaxed: 

buddy  @orangejenny


emord:
I don't think this would get properly deployed. it should be in the couchdb2 role

snopoke:
https://github.com/dimagi/commcarehq-ansible/pull/858

PR #857, opened 2017-07-21
nginx proxy caching

emord:
@dimagi/scale-team

### DIFF
--- a/ansible/roles/nginx/templates/site.j2
+++ b/ansible/roles/nginx/templates/site.j2
@@ -53,7 +53,7 @@ server {
 {% endif %}
 
 {% for k,v in item.server.iteritems() %}
-  {% if k.find('location') == -1 and k not in ['file_name', 'balancers', 'proxy_set_headers', 'upstream_port', 'limit_req_zones'] %}
+  {% if k.find('location') == -1 and k not in ['file_name', 'balancers', 'proxy_set_headers', 'upstream_port', 'limit_req_zones', 'proxy_cache_path'] %}
   {{ k }} {{ v }};
   {% endif %}
 {% endfor %}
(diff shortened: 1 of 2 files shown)


emord:
Followed https://www.nginx.com/blog/nginx-caching-guide/ for this

emord:
Turns out that multimedia already allowed browser caching, so I added a line to cache that in nginx. I still think it would be good to allow other app files to be cached there as well though

emord:
@snopoke by default it's `$scheme$proxy_host$request_uri` and $request_uri contains the query parameters

The 60m here means that it's flushed from the cache if it isn't accessed for 60m. so it'd likely get flushed overnight, but be around all day.

if we struggle tomorrow or monday, I think patching https://github.com/dimagi/commcare-hq/pull/17136 and restarting the webworkers will relieve some more stress if that's a safe change. This ansible PR is already deployed to ICDS so that patch would allow caching xml and media_profile immediately I believe

ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -24,6 +24,10 @@ nginx_sites:
       zone: "$server_name"
       size: "10m"
       rate: "100r/s"
+   proxy_cache_path:
```
emord: This adds a dictionary in a server. will fix tomorrow

ansible/roles/nginx/templates/site.j2
```diff
@@ -22,7 +22,7 @@ limit_req_zone {{ zone.get('zone', '$binary_remote_addr') }} zone={{ zone.name }
 
 {% for cache in item.server.get('proxy_cache_path', []) %}
 proxy_cache_path {{ cache.path }} levels=1:2 keys_zone={{ cache.name }}:10m
-                 max_size={{ cache.max_size}} inactive=60m use_temp_path=off;
+                 max_size={{ cache.max_size}} inactive=24h use_temp_path=off;
```
emord: This should be a variable

PR #858, opened 2017-07-24
couch2 restarter in correct location

snopoke:


snopoke:
I've updated couch0 & pillow1 on ICDS with these changes

snopoke:
No, I've moved the ones we had to where they should be according to this PR

PR #861, opened 2017-07-27
tmpreaper for formplayer

benrudolph:
@javierwilson @emord i tried using `tmpreaper` for this. i tested on staging doing `tmpreaper  7d /opt/data/formplayer` and checked the conf to make sure it updated correctly. i'm a little worried that using tmpreaper is too blunt. if formplayer gets deployed to a machine that needs a different config for tmpreaper than i'm not sure how we do that. this may make sense to use our old cron way of killing old files. what do you think?

cc: @proteusvacuum

### DIFF
--- a/ansible/group_vars/formplayer.yml
+++ b/ansible/group_vars/formplayer.yml
@@ -1,1 +1,4 @@
 datadog_integration_jmx: true
+
+tmpreaper_dirs: "{{ encrypted_root }}/formplayer" "/tmp/."
+tmpreaper_time: 7d
--- a/ansible/group_vars/formplayer.yml
+++ b/ansible/group_vars/formplayer.yml
@@ -1,4 +1,1 @@
 datadog_integration_jmx: true
-
-tmpreaper_dirs: "{{ encrypted_root }}/formplayer /tmp/."
-tmpreaper_time: 7d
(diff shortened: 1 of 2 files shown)


emord:
Maybe we'll have to add it to the inventory file? Is there a way to use different tmpreaper settings per folder?

benrudolph:
it wasn't obvious to me if it does exist. according to this: https://github.com/ANXS/tmpreaper. you just set some variables and those variables get written to `/etc/tmpreaper.conf`

benrudolph:
I think I'm going to change this back to our old way of deleting files for the above reasons. @javierwilson that OK with you?

javierwilson:
I think the global tempreaper.conf config will not help here, however, you can still use tmpreaper instead of "find", it has some advantages over find. 
So maybe just add the cron task but use the command: "tmpreaper 1d {{ encrypted_root }}/formplayer"
To delete files older than 1 day.

benrudolph:
ok switched to tmpreaper. the cron looks like this in check mode:
```
+#Ansible: Purge sqlite db files
+@daily root tmpreaper /opt/data/ecrypt/formplayer 2d

changed: [10.162.36.231] => (item={u'time_spec': u'2d', u'name': u'Purge sqlite db files', u'dir': u'/opt/data/ecrypt/formplayer'})
```

javierwilson:
small change...

benrudolph:
small but big. thanks, i've been running it manually and for some reason didn't catch that

PR #864, opened 2017-08-02
Seppuku

benrudolph:
image


emord:
Also need to remove yourself here https://github.com/dimagi/commcarehq-ansible/blob/0a46e3441963b0968c50e67b1959245d3b52cc39/ansible/group_vars/all.yml#L27

benrudolph:
couldn't even kill myself completely!

PR #866, opened 2017-08-03
upgrade to redis 4

javierwilson:
https://raw.githubusercontent.com/antirez/redis/4.0/00-RELEASENOTES


emord:
Nice. Are you thinking about using a library from ansible galaxy now that we're trying to cluster redis?

javierwilson:
we should, i've tested a bit, and wanted to see if you guys raised any red flag regarding the upgrade, it seems to be backwards compatible and has worked with no problems so far.

snopoke:
Great, I don't see any issues.

PR #877, opened 2017-08-15
Cap ES at 30 GB

emord:
the reason to not go over 32 GB is because that's when ES starts using 64 bit pointers. Due to overhead that actually happens at 30.5 GB, but apparently 30 GB is more common usage. Could have caused issues when our servers on icds had 64 GB (48 GB now)

@dimagi/scale-team


dannyroberts:
just saw this going through old PRs. Still useful?

emord:
@dannyroberts Yeah. forgot about this. I think it will still be useful to have this if we ever have 64 GB servers

PR #881, opened 2017-08-16
pg1 query stats

emord:
@dimagi/scale-team Just deployed this to icds. Will check in on it tomorrow


javierwilson:
@emord so, before this commit we would loop through the dbs and try to install the extension regardless of their presence (or not) on the current host?

emord:
@javierwilson that's correct. I didn't notice it before because I was always using --limit to limit it to the database I wanted to chagne

PR #891, opened 2017-08-24
using monit to restart redis

javierwilson:
we can easily add more services (to be monitored), just wanted to start with redis

### DIFF
--- a/ansible/roles/redis/tasks/main.yml
+++ b/ansible/roles/redis/tasks/main.yml
@@ -30,3 +30,21 @@
   service: name="{{ redis_service }}" state=started enabled=yes
   tags: after-reboot
 
+- name: redis monit config
+  sudo: yes
+  template:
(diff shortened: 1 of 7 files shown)


ansible/roles/redis/tasks/main.yml
```diff
@@ -30,3 +30,21 @@
   service: name="{{ redis_service }}" state=started enabled=yes
   tags: after-reboot
 
+- name: redis monit config
+  sudo: yes
+  template:
+    src: "monit.redis.j2"
+    dest: "/etc/monit/conf.d/redis"
+    group: root
+    owner: root
```
emord: should this be owned by a monit user or the redis user?

javierwilson: monit runs as root, there is no monit user :(

PR #893, opened 2017-08-24
Settings for pg0

emord:
@dimagi/scale-team We don't have the ability in ansible to change these per db machine yet, so just documenting this here. this is live on pg0


snopoke:
You can override on a per DB bases by putting them in the inventory file

emord:
Thanks. I'll look into that depending on the result of javier's test

PR #894, opened 2017-08-24
Bump max daily rule updates for enikshay

gcapalbo:
@orangejenny


gcapalbo:
@orangejenny yup, thanks - 500,000 is correct. now that we're on the new cluster going to make this 10x higher

PR #901, opened 2017-08-29
monit couchdb2 / restarter

javierwilson:
also fixing removal of redis restarter and absense of couchdb_dir

### DIFF
--- a/ansible/roles/couchdb2/tasks/main.yml
+++ b/ansible/roles/couchdb2/tasks/main.yml
@@ -93,10 +93,7 @@
     group: root
     owner: root
     mode: 0640
-  tags: monit
-
-- name: restart monit
-  service: name="monit" state=reloaded
+  notify: reload monit
   tags: monit
 
 - monit:
--- a/ansible/roles/couchdb2/tasks/main.yml
+++ b/ansible/roles/couchdb2/tasks/main.yml
@@ -21,13 +21,8 @@
 
 - name: Copy couch restart script
   sudo: yes
-  template:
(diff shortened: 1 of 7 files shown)


ansible/roles/couchdb2/tasks/main.yml
```diff
@@ -83,3 +80,26 @@
...
+    src: "monit.couchdb2.j2"
+    dest: "/etc/monit/conf.d/couchdb2"
+    group: root
+    owner: root
+    mode: 0640
+  tags: monit
+
+- name: restart monit
+  service: name="monit" state=reloaded
+  tags: monit
```
snopoke: I think this should be a handler and not a task

ansible/roles/couchdb2/templates/monit.couchdb2.j2
```diff
@@ -0,0 +1,6 @@
+ check process couchdb2 with pidfile /usr/local/couchdb2/couchdb/var/run/couchdb2.pid
+   group database
+   group couchdb2
+   start program = "/etc/init.d/couchdb2 start"
+   stop program = "/etc/init.d/couchdb2 stop"
+   if failed host {{ inventory_hostname|ipaddr }} port 15984 then restart
```
snopoke: should use the `couchdb2_port` variable instead of hard coding

ansible/roles/couchdb2/tasks/main.yml
javierwilson: using handlers now...

javierwilson:
@snopoke i've made some changes, please review

PR #904, opened 2017-08-30
handle read-only databases separately

snopoke:
buddy @kaapstorm 

along with https://github.com/dimagi/commcare-hq/pull/17643

### DIFF
--- a/ansible/vars/icds/icds_public.yml
+++ b/ansible/vars/icds/icds_public.yml
@@ -135,10 +135,12 @@ postgresql_dbs:
   - name: commcarehq_ucr
     host: "{{ groups.postgresql.0 }}"
     django_alias: ucr
+    django_migrate: False
   - django_alias: icds-ucr
     name: commcarehq_icds_aggregatedata
     host: "{{ groups.postgresql.0 }}"
     query_stats: True
+    django_migrate: False
   - name: commcarehq_icds_testdata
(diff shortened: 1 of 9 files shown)


snopoke:
Deployed this to all envs so HQ change is GTG

ansible/vars/icds/icds_public.yml
```diff
@@ -135,10 +135,12 @@ postgresql_dbs:
   - name: commcarehq_ucr
     host: "{{ groups.postgresql.0 }}"
     django_alias: ucr
+    django_migrate: False
   - django_alias: icds-ucr
     name: commcarehq_icds_aggregatedata
     host: "{{ groups.postgresql.0 }}"
     query_stats: True
+    django_migrate: False
```
emord: @snopoke It's possible that we do want to migrate this https://github.com/dimagi/commcare-hq/tree/master/custom/icds_reports/migrations

I think we have moved most of the table generation into sql scripts that run druing the aggregation, but historically there have been migrations on icds-ucr. @sheelio can you confirm

sheelio:
Yes, we have SQL Scripts that have migrated those tables (mainly adding or
removing columns).

ansible/vars/icds/icds_public.yml
snopoke: abf9e86

PR #906, opened 2017-08-30
Add icds ucr database for softlayer

emord:
@snopoke This is needed for QA

@biyeun @proteusvacuum


emord:
This is already deployed and deploying without this change would break QA. so self merging

PR #907, opened 2017-08-30
Add localsettings playbook

emord:
1uzpro

Still needs to use --tags=localsettings at the command line. according to https://stackoverflow.com/a/25675696 there's no way to get around specifying it at the command line unless we go through and restructure it to not use tags

@biyeun @proteusvacuum

### DIFF
--- a/ansible/deploy_localsettings.yml
+++ b/ansible/deploy_localsettings.yml
@@ -1,0 +1,7 @@
+---
+# deploy-stack.yml
+
+- include: deploy_riakcs.yml
+- include: deploy_commcarehq.yml
+- include: deploy_touchforms.yml
+- include: deploy_formplayer.yml


proteusvacuum:
Are there docs anywhere on what the changes are here?

ansible/deploy_localsettings.yml
```diff
@@ -0,0 +1,7 @@
+---
+# deploy-stack.yml
+
+- include: deploy_riakcs.yml
+- include: deploy_commcarehq.yml
+- include: deploy_touchforms.yml
+- include: deploy_formplayer.yml
```
millerdev: @emord I think this means all these things get done every time someone wants to deploy localsettings, which seems it might defeat the purpose of this PR which is to speed up the localsettings deploy. Seems like you'd definitely want to always run `deploy_localsettings.yml` with `--tags=localsettings`. Too bad you can't do an `include` with a tag filter so you only include tasks with a certain tag...

Just commentary, not really sure I know what to do about it other than document the part about always using `--tags=localsettings`. Maybe add a task here in this file to confirm that it's really what you wanted if `--tags=localsettings` is not being used?

millerdev: I don't think so. We might be able to remove it if we stopped using the thing where ansible reads the key off of the riak control node, but that's going to be a bigger effort than I want to think about right now. Is it possible to do a conditional include? If so, it might make sense add a `when: riak_key is undefined or riak_secret is undefined` to that include (although neither of those vars are set for staging, so it wouldn't help there).

PR #911, opened 2017-08-31
postgres complains if recovery.conf is in config dir

javierwilson:
@calellowitz this will require a change in the wiki instructions, apparently this change (not allowing recovery in config dir) came from ubuntu packaging see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=853868

I blame Myon, from #postgresql:

<xocolatl> Myon: I wonder if pg_ctlcluster should error out if it finds a recovery.conf in /etc
<Myon> possibly, yes

@nickpell this solves the issue you encountered

### DIFF
--- a/ansible/roles/pg_standby/tasks/main.yml
+++ b/ansible/roles/pg_standby/tasks/main.yml
@@ -11,7 +11,7 @@
     state: directory
 
 - name: PostgreSQL recovery configuration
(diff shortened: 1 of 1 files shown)


snopoke:
Looks fine but don't have much context on this

nickpell:
thanks Javier!

PR #914, opened 2017-09-03
allow ssh password authentication

javierwilson:
let  me know what you think about this 

CtrlS is locked out of VMs since we disable password logins. This change is to enable access for them for monitoring and support.


snopoke:
Can they not give us an ssh key to install on the servers?

javierwilson:
we asked for it, we got a negative answer.

PR #917, opened 2017-09-04
pgbackrest as a postgresql backup solution

javierwilson:
this doesn't take full advantage of pgbackrest, specially the "backup from standby" option :( we could benefit from it is we use pgbackrest for the standby configuration itself, this would also automate standby installations more easily.  tested only on dev enviroment so far.


ansible/roles/pgbackrest/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+    owner=postgres
+    group=postgres
+    mode=640
+
+- name: Creating postgres stanza
+  become: True
+  become_user: postgres
+  vars:
+    ansible_ssh_pipelining: True
+  shell: pgbackrest --stanza=backup --log-level-console=info stanza-create
```
snopoke: Is this OK to run multiple times?

javierwilson: it is, however it is marked as "changed" everytime,  i am going to check if {{postgresql_backup_dir}}/archive/backup exists before running it again.

javierwilson: done. however i still have the problem of postgresql restarting twice because of the changes done to the archive_command line by the two different playbooks (postgresql and pgbackrest)

snopoke:
If we're using handles it should only restart once (unless it's doing a
reload and a restart)

javierwilson:
true. it restarts only once now, still, it shouldn't restart at all. problem is both plays change the config file, everytime.

snopoke:
Do you have a plan for testing this further?

javierwilson:
this is being tested in softlayer

javierwilson:
@snopoke @gcapalbo any reason not to merge this?

ansible/roles/pgbackrest/tasks/main.yml
snopoke: yea, that's unfortunate

gcapalbo:
@javierwilson this is already deployed to india and enikshay right?

i'm curious if you still think this is a better backup solution. i've noticed some limitations around using pgbackrest, mainly:

- you can only delete the oldest backup, so if you accidentally did a full backup and it's the latest one, there's no way to delete it without deleting all backups.

- you can only delete incremental backups by deleting the full backup they're associated with, so it becomes basically impossible to delete even very old incremental backups. for this reason i had to switch to using rolling differential backups otherwise we ran out of space fast.

- for us, the incremental and differential backups seem to take up just about as much space as the full backups do

- it does allow us to do a point in time recovery for as long we keep the wal archive (which is something we couldn't do before), but keeping wal for a long time takes up too much space so i had to reduce wal retention to the last backup, which means we'd only be able to do the PITR for any time since the last backup

javierwilson:
It doesn't sound as wonderful as it once did :(

ansible/roles/postgresql/templates/pg_hba.conf.j2
```diff
@@ -99,6 +99,9 @@ host    all             all             ::1/128                 md5
 # Allow replication connections from localhost, by a user with the
 # replication privilege.
 local   replication     postgres                                peer
+{% if hot_standby_server|default(False) %}
+host    replication     postgres        {{ hot_standby_server }}/32 trust
```
snopoke: If this get's changed from `trust` to `md5` where would we put the password for pg_backrest?

snopoke: Just trying to figure out how we merge this with https://github.com/dimagi/commcarehq-ansible/pull/962/files#diff-d959b4d4814d1bc5d1d095125136a90aR105

javierwilson: From the docs "It is possible to configure a password in the primary_conninfo setting but using a .pgpass file is more flexible and secure."
But our current pgbackrest setup doesn't need it, I used this to be able to remotely pg_basebackup, instead of doing it locally and then rsync (saves an step).

javierwilson: it's ok to remove this or change to md5

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -60,6 +60,14 @@
 - name: assert hot standby config correct
   assert: { that: "hot_standby_server is not defined or hot_standby_server in groups.pg_standby"}
 
+- name: Create postgresql_log_directory
+  file:
+    dest: "{{ postgresql_log_directory }}"
+    state: directory
+    owner: postgres
+    group: postgres
+  when: postgresql_log_directory|default(false)
```
snopoke: why is this necessary? this seems unrelated to pg_backrest

javierwilson: following advise on https://github.com/pgbackrest/pgbackrest/issues/413
just to save disk space

PR #918, opened 2017-09-05
log cache status to timing log

snopoke:


dannyroberts:
@snopoke maybe I missed it—what's the motivation behind logging the cache status? From the docs, $upstream_cache_status seems to refer to whether it went on to make a request to "upstream" (django web workers in our case?) or just returned from the cache. Is that right?

dannyroberts:
Ah, probably related to this email excerpt:

> We did do some changes last time which makes nginx cache the app updates so they don't hit Django as much. This should help dramatically.

snopoke:
@dannyroberts correct, we now use nginx caching for app downloads so this would allow us to monitor how well it's working.

snopoke:
Here's the PR where we introduced that: https://github.com/dimagi/commcare-hq/pull/17171

PR #921, opened 2017-09-06
Set celery timezone for enikshay

emord:
@javierwilson @mkangia @proteusvacuum 

I'm pulling this from the icds config https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/vars/icds/icds_public.yml

I didn't notice anything going wrong, just thought this might be a better setting. I think without this the enikshay task [0] is running at midnight UTC (530 IST). this would cause it to run at midnight IST

buddy @sravfeyn 

[0] https://github.com/dimagi/commcare-hq/blob/af4c010ee38dfeda5e6149a880be374c9dd16843/custom/enikshay/tasks.py#L62


proteusvacuum:
Repeaters fire every 5 minutes with heartbeat, so this shouldn't affect that. This is a good change, I didn't realize this was running at 5:30AM.

PR #922, opened 2017-09-06
PNA server installation

javierwilson:


ansible/vars/pna/pna_public.yml
```diff
@@ -0,0 +1,213 @@
+SITE_HOST: 'pna.commcarehq.org'
```
kaapstorm: `commcare.pna.sn` now, I guess.

javierwilson: yes. changed.

javierwilson:
@kaapstorm this is important. we need to be able to change ssh port otherwise UFW will lock us out.

ansible/roles/ecryptfs/tasks/main.yml
```diff
@@ -74,29 +74,41 @@
...
 
 - name: "Create {{ encrypted_root }}/blobdb directory for setups that do not use NFS"
-  sudo: yes
+  become: true
   when: not shared_drive_enabled
   file: path="{{ encrypted_root }}/blobdb" owner=nobody group="{{ shared_dir_gid }}" mode=0775 state=directory
   tags:
     - ecryptfs-blobdb
 
+- name: Create shared data dir
```
kaapstorm: Is there a way to reuse the `Create shared data dir` task that's in ansible/roles/shared_dir/tasks/setup_host.yml? ... Or if it's not possible, maybe add a comment to point to each other, so that if we change this task, we know to change it in both places.

javierwilson: good call, let me see what I can do.

emord:
Any reason not to merge this? It doesn't look like it would affect anything in prod

PR #923, opened 2017-09-06
cache prelogin pages

biyeun:
@dannyroberts @javierwilson 

I'm going off of the `cas_ssl` example and these docs: 
https://www.nginx.com/resources/admin-guide/content-caching/

### DIFF
--- a/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -9,6 +9,11 @@ nginx_sites:
       hosts: formplayer
       port: "{{ formplayer_port }}"
       method: "hash $cookie_sessionid consistent"
+   proxy_cache_path:
+    - name: "prelogin_cache"
+      path: "{{ www_home }}/prelogin_cache"
+      max_size: "3g"
+      inactive_time: "24h"
    file_name: "{{ deploy_env }}_commcare"
    listen: "443 ssl{{ ' default_server' if deploy_env == primary_ssl_env else '' }}"
    root: "/var/www/html"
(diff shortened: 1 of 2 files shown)


ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -20,6 +25,24 @@ nginx_sites:
    add_header: "X-Frame-Options SAMEORIGIN"
    access_log: "{{ log_home }}/{{ deploy_env }}-timing.log timing"
    locations:
+    - name: "^~ /(home|impact|pricing|services|software|solutions|askdemo)$"
```
dannyroberts: I'm confused what `^~ ` is supposed to mean here. According to http://nginx.org/en/docs/http/ngx_http_core_module.html#location

> If the longest matching prefix location has the “^~” modifier then regular expressions are not checked.

it seems to mean "don't check regular expressions", but this seems like definitely a regular expression. (Maybe I interpreted that sentence wrong—there's not much there.

The only other thing I can think of is that you transposed the `^`, `~`, and ` ` characters and meant

```
"~ ^/(home|impact|pricing|services|software|solutions|askdemo)/$"
```
in which case the tilde means "this is a regular expression, and the `^` is just part of the regular expression.

ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -20,6 +25,24 @@ nginx_sites:
...
+      balancer: "{{ deploy_env }}_commcare"
+      proxy_redirect: "http://{{ SITE_HOST }} https://{{ SITE_HOST }}"
+      proxy_next_upstream_tries: 1
+      proxy_read_timeout: 900s
+      proxy_cache: prelogin_cache
+      proxy_cache_use_stale: "error timeout updating http_500 http_502 http_503 http_504"
+      proxy_cache_methods: GET
+      add_header: X-Cache-Status $upstream_cache_status
+    - name: "/lang/"
```
dannyroberts: Scanning these settings with my eyeballs it looks like they're they same as those above (except for `name`). If I'm not missing an important difference, can this just be added to the regex above?

biyeun:
@dannyroberts updated

emord:
@biyeun nginx won't actually cache these pages unless the headers are set properly. We have a decorator and middleware to do this https://github.com/dimagi/commcare-hq/blob/master/corehq/middleware.py#L140-L170

also there will need to be some way to bust the cache on updates to the prelogin module. It's possible that the static files have a hash on the url to get around this, but I don't think the endpoints that hit django do.

biyeun:
@emord added the ability to force purge the NGINX caches using the `PURGE` request to any of the endpoints, per these docs: https://www.nginx.com/products/nginx/caching/#purging

also added docs for how to force purge and updated the headers on the `prelogin` views to enable caching. https://github.com/dimagi/commcarehq-prelogin/pull/138

emord:
This looks like a fairly manual process that we will inevitably forget to do many times. Would it be possible to add those commands to the deploy? Preferably it would only happen if the prelogin submodule was updated, but I think it would be acceptable to do on every deploy

dannyroberts:
+1 to adding this to deploy. Maybe the easiest thing to do is after `PURGE`ing, write the prelogin submodule commit id to a file somewhere; and then the next time before `PURGE`ing, read the file to see if it's different from the current submodule commit id; if it's different then `PURGE`, otherwise skip.

biyeun:
@dannyroberts @emord I'm kind of preoccupied with dimagi.com stuff right now. is there a change we can merge this today so that the static site gets cached for the time being? I'll add the deploy script on round two.

dannyroberts:
I'm okay with that.

dannyroberts:
Do you have everything you need to test this on staging? (Maybe you already did that?)

PR #928, opened 2017-09-10
start riak after reboot

javierwilson:
after-reboot won't start riak because the whole "configuration and start riak" play is skipped, so I put this task separably. I am not sure it's the most elegant solution. opinions welcomed.

### DIFF
--- a/ansible/deploy_riakcs.yml
+++ b/ansible/deploy_riakcs.yml
@@ -83,7 +83,14 @@
         - "{{ riak_data_dir }}"
         - "{{ riak_ring_dir }}"
         - "{{ riak_data_root_leveldb }}"
-      when: _force_riak_config | bool
+      when: _force_riak_config|default(false) | bool
+
+- name: Start Riak
+  become: true
+  hosts: riakcs
+  tasks:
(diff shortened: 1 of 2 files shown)


ansible/deploy_riakcs.yml
```diff
@@ -83,7 +83,14 @@
         - "{{ riak_data_dir }}"
         - "{{ riak_ring_dir }}"
         - "{{ riak_data_root_leveldb }}"
-      when: _force_riak_config | bool
+      when: _force_riak_config|default(false) | bool
```
millerdev: why is the `| default(false)` needed? is `| bool` still needed?

ansible/deploy_riakcs.yml
```diff
@@ -83,7 +83,14 @@
...
+
+- name: Start Riak
+  become: true
+  hosts: riakcs
+  tasks:
+    - service: name="riak" state=started enabled=yes
+      tags: after-reboot
```
millerdev: I think this should be fine since it should be fast/no-op in cases where riak is already started (although not 100% sure about the "fast" part).

PR #930, opened 2017-09-12
remove process integration

snopoke:


dannyroberts:
This PR was a long time ago, but wanted to note that i just had a problem with dd-agent starting on enikshay proxy0 that I traced back to the presence of a `process.yaml` file (with an empty instances section). Might have been worth adding a task assuring process.yaml is absent, not sure if it's worth doing anymore since I don't totally get how/why this problem was happening.

snopoke:
Sorry @dannyroberts my bad. We do already have that task (https://github.com/dimagi/commcarehq-ansible/pull/930/files#diff-ac5996d925ce320638570db035716995R19) but I should have done the removal in 2 stages.

dannyroberts:
Oh, right, cool so this seems like an easy fix. What do you think about https://github.com/dimagi/commcarehq-ansible/pull/1084?

PR #935, opened 2017-09-13
enikshay fixes

snopoke:


ansible/vars/enikshay/enikshay_public.yml
```diff
@@ -33,7 +34,7 @@ riakcs_s3_secret_key: "{{ secrets.RIAKCS_S3_SECRET_KEY }}"
 
 proxy_type: 'nginx'
 
-primary_ssl_env: "enikshay"
+primary_ssl_env: "enikshay.in"
```
emord: I think this should be enikshay https://github.com/dimagi/commcarehq-ansible/blob/8c7c648e45eb7f70cbbfeed2a9e0adc236c57654/ansible/roles/nginx/vars/cchq_ssl.yml#L13

snopoke: Not this is correct. On ICDS we have `deploy_env = 'icds'` and `primary_ssl_env=cas`.

Having this as `enikshay` makes both the `enikshay.in` and the `enikshay.commcarehq.org` configs default which nginx complains about.

emord: Ah. yeah that makes sense. thanks

PR #937, opened 2017-09-13
remove icds proxy sites

snopoke:


snopoke:
updated to remove `enikshay_ssl`

Note that this won't actually change anything on the proxy

PR #940, opened 2017-09-14
ansible logging

snopoke:
FYI @dimagi/devops

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -1,3 +1,4 @@
 [defaults]
 ansible_managed = Ansible managed, do not edit directly
 roles_path = /opt/ansible_roles
+log_path=/var/log/ansible.log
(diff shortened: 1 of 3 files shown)


ansible/ansible.cfg
```diff
@@ -1,3 +1,4 @@
 [defaults]
 ansible_managed = Ansible managed, do not edit directly
 roles_path = /opt/ansible_roles
+log_path=/var/log/ansible.log
```
millerdev: do we need to rotate this log file?

javierwilson:
nice. i assume takes place in the managed machine, in that case ansible needs permission to write to the file, which we cannot guarantee unless we use ansible to create it (but...) so I wonder if it's better to use ~/ansible.log. unless ansible is clever and uses sudo to create the file.

snopoke:
This file is created by ansible is writable by the `dimagidev` group which all devs are part of. I've deployed this to prod, enikshay and icds

Re logratation - that's probably a good idea

PR #944, opened 2017-09-14
using ssl secrets from vault

javierwilson:
testing this for enikshay first, then the world...


ansible/vars/enikshay/enikshay_public.yml
```diff
@@ -55,13 +55,17 @@ elasticsearch_cluster_name: 'enikshayes-1.x'
 
 apache2_cert_file: '../config/{{ deploy_env }}/ssl/enikshay.in.crt'
 apache2_key_file: '../config/{{ deploy_env }}/ssl/enikshay.in.key'
+apache2_key_value: '{{ ssl_secrets.private_keys.enikshay_in }}'
```
snopoke: Just a note that since we don't use apache anymore we can rename these

PR #947, opened 2017-09-15
logrotate every hour

emord:
I think this should help solve the space issues we've seen recently on proxy workers. 

By default logrotate is only run every day. This moves it to every hour. This means that if the log is at 749M when it gets run, then it could balloon over the next night and not be caught by th erotation

I couldn't make this change in the normal ansible-logrotate because that's a submodule to a repo we don't control. that should get moved to galaxy at some point.

@javierwilson @snopoke @sravfeyn


ansible/deploy_proxy.yml
```diff
@@ -90,6 +90,8 @@
             - missingok
             - compress
             - delaycompress
-            - copytruncate
             - nocreate
             - notifempty
+            - sharedscripts
```
emord: This means the post rotate script is only run once

ansible/deploy_proxy.yml
```diff
@@ -90,6 +90,8 @@
...
+          scripts:
+            postrotate: "[ -s /run/nginx.pid ] && kill -USR1 `cat /run/nginx.pid`"
```
emord: Tells nginx to change its file handles for logging

ansible/deploy_proxy.yml
```diff
@@ -90,6 +90,8 @@
             - missingok
             - compress
             - delaycompress
-            - copytruncate
```
emord: having this meant that htere was a portion of time that the we lost logs...

emord:
deployed to nic and ctrls

ansible/roles/logrotate/tests/test.yml
```diff
@@ -0,0 +1,5 @@
+---
+- hosts: localhost
+  remote_user: root
+  roles:
+    - logrotate
```
snopoke: Nice, will this actually run?

emord: Good question. I'm not sure. I just  used `ansible-galaxy init logrotate` to initialize the role

emord:
thanks. added it to deploy_db

PR #948, opened 2017-09-15
fix couchdb2 datadog stuff for non-couchdb2 envs

dannyroberts:


ansible/roles/datadog/tasks/add_integrations.yml
```diff
@@ -27,7 +27,7 @@
     sha256sum: "{{ datadog_couchdb2_sha256sum }}"
     force: yes
   notify: restart datadog
-  when: inventory_hostname == groups.couchdb2[0]
+  when: "'couchdb2' in groups and inventory_hostname == groups.couchdb2[0]"
```
kaapstorm: Do you have a rule of thumb / aesthetic / style guide for when to wrap conditionals in quotation marks? I just ask because I hadn't seen it before.

dannyroberts: This errored if I didn't have this, and ansible said to put the whole thing in quotes to fix it. I think the problem is that if it starts with a `'` (as in `'couchdb2`), then it assumes that that quote is for the whole line and gets confused if the quoted text terminates early. So my rule of thumb is basically, put double quotes around the whole thing if ansible complains.

kaapstorm: :laughing: that sounds fair.

PR #950, opened 2017-09-18
Add playbook to deploy chaos to host(s)

millerdev:
Currently only supports one type of chaos: BurnIO, which I'm hoping to use for IO load testing on web workers.

@calellowitz cc @javierwilson @emord

### DIFF
--- a/ansible/roles/chaos/tasks/main.yml
+++ b/ansible/roles/chaos/tasks/main.yml
@@ -10,21 +10,18 @@
     - start
 
 - name: Install chaos scripts
-  sudo: yes
   template:
-    src: "{{ item }}.j2"
-    dest: "{{ chaos_dir }}/start/{{ item }}"
+    src: "{{ strategy }}.j2"
+    dest: "{{ chaos_dir }}/start/{{ strategy }}"
     group: root
     owner: root
     mode: 0755
     backup: yes
-  with_items: "{{ enable.split(',') }}"
   tags:
     - start
 
 - name: Start chaos
-  shell: "{{ chaos_dir }}/start/{{ item }}"
-  with_items: "{{ enable.split(',') }}"
+  shell: "{{ chaos_dir }}/start/{{ strategy }}"
   tags:
     - start
 
--- a/ansible/roles/chaos/tasks/main.yml
+++ b/ansible/roles/chaos/tasks/main.yml
@@ -1,0 +1,50 @@
+# Chaos
+
+- name: Setup chaos directories
+  file: path="{{ item }}" owner=root group=root mode=0755 state=directory
+  with_items:
+    - "{{ chaos_dir }}"
+    - "{{ chaos_dir }}/start"
+    - "{{ chaos_dir }}/stop"
+  tags:
+    - start
+
+- name: Install chaos scripts
+  sudo: yes
+  template:
+    src: "{{ item }}.j2"
+    dest: "{{ chaos_dir }}/start/{{ item }}"
+    group: root
+    owner: root
+    mode: 0755
+    backup: yes
+  with_items: "{{ enable.split(',') }}"
+  tags:
+    - start
+
+- name: Start chaos
+  shell: "{{ chaos_dir }}/start/{{ item }}"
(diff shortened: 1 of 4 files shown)


ansible/roles/chaos/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+- name: "Find stop scripts"
+  shell: "find {{ chaos_dir }}/stop -executable -type f"
+  register: stop_scripts
+  changed_when: false
+  failed_when: false
+  tags:
+    - debug_stop
+    - stop
+
+- name: Stop chaos
```
calellowitz: I am not sure how we plan on using this, but are concerned at all about being locked out of the machine. At times in the past when cpu got too high we have been unable to ssh into machines, and since this relies on being able to reconnect to the machine and issues a series of commands to stop it, that seems like potentially a problem here.

ansible/roles/chaos/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+- name: Install chaos scripts
+  sudo: yes
+  template:
+    src: "{{ item }}.j2"
+    dest: "{{ chaos_dir }}/start/{{ item }}"
+    group: root
+    owner: root
+    mode: 0755
+    backup: yes
+  with_items: "{{ enable.split(',') }}"
```
calellowitz: this seems like a natural fit for tags, something like `ap deploy_chaos.yml --tags=burnio`. then each chaos script could be its own task. I see that you use tags later on to do stop/start but maybe those belong in separate playbooks. so it would end up `ap start_chaos.yml --tags=burnio` and `ap stop_chaos.yml --tags=burnio`, that also might give slightly more control over what to stop, instead of always stopping everything. I am not sure this idea is strictly better, but passing in an enable variable feels clunky.

ansible/roles/chaos/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+  shell: "{{ item }}"
+  with_items: "{{ stop_scripts.stdout_lines }}"
+  tags:
+    - debug_stop
+    - stop
+
+- name: Remove chaos scripts
+  file: path="{{ chaos_dir }}" state=absent
+  tags:
+    - stop
```
calellowitz: it looks like passing no tags starts and then stops whatever you are enabling, is that on purpose?

ansible/roles/chaos/tasks/main.yml
emord: For this particular one I don't think tihs is an issue. It should only need to run one dd command .

ansible/roles/chaos/templates/burnio.j2
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Script for BurnIO Chaos
+#
+# based on https://github.com/Netflix/SimianArmy/blob/master/src/main/resources/scripts/burnio
```
emord: This link needs an .sh at the end. Also would be nice to have a commit hash instead of master since master changes where it points to

ansible/roles/chaos/templates/burnio.j2
```diff
@@ -0,0 +1,28 @@
...
+
+cat << EOF > {{ chaos_dir }}/burnioloop
+#!/bin/bash
+# do burnio chaos
+while true;
+do
+    dd if=/dev/urandom of=$BURNFILE bs=1M count=128 iflag=fullblock
+    sleep {{ burnio_wait }}
+done
+EOF
```
emord: The start script looks like it is blocking. Does this mean you'd need to open up a second shell in order to stop it?

snopoke:
@millerdev curious what you plan on doing with this

ansible/roles/chaos/templates/burnio.j2
emord: Thanks. I missed the nohup command to put it in the background. Original question was if this would block the ansible playboook (so that the playbook never exited). Looks like that isn't a problem

PR #953, opened 2017-09-21
Don't rate limit when users are submitting authorization

emord:
@dimagi/scale-team pretty sure this will let us not rate limit after a 401.

According to https://en.wikipedia.org/wiki/Digest_access_authentication the authorization header is only on the second request. 

basing the nginx config off of this https://serverfault.com/questions/784852/nginx-conf-for-limit-req-based-on-http-header

@snopoke @sravfeyn

### DIFF
--- a/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/ansible/roles/nginx/vars/cas_ssl.yml
@@ -19,15 +19,19 @@ nginx_sites:
     - name: "restore"
       zone: "$server_name"
       size: "10m"
-      rate: "100r/s"
+      rate: "25r/s"
     - name: "submissions"
       zone: "$digest_submission"
       size: "10m"
-      rate: "200r/s"
+      rate: "20r/s"
     - name: "server_name"
       zone: "$server_name"
       size: "10m"
-      rate: "100r/s"
+      rate: "25r/s"
+    - name: "app_downloads"
+      zone: "$server_name"
+      size: "10m"
+      rate: "50r/s"
    proxy_cache_path:
     - name: "app_cache"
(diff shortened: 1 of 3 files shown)


ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -12,13 +12,16 @@ nginx_sites:
       hosts: formplayer
       port: "{{ formplayer_port }}"
       method: "hash $cookie_sessionid consistent"
+   mappings:
+    - name: "$digest_submission"
+      variable: "$http_authorization"
```
emord: If this header is included (after 401) then the key for the rate limiting will be "authorization_remote_ip".

If it isn't (before 401) then the key is "".

ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -12,13 +12,16 @@ nginx_sites:
...
    limit_req_zones:
     - name: "restore"
       zone: "$server_name"
       size: "10m"
       rate: "100r/s"
     - name: "submissions"
-      zone: "$server_name"
+      zone: "$digest_submission"
       size: "10m"
       rate: "200r/s"
```
emord: crazy we used to have 200r/s

emord:
Really kicking myself for this one. It looks like I read the map definition incorrectly.

for the mapping:

```
map $http_authorization $limit {
    ""     $binary_remote_addr;
    default     ""; 
}
```

If we don't include the header it is rate limited by $binary_remote_addr, the default is a blank string which is not rate limited (where already authorized requests fall in)

I've changed remote_addr to be configurable and started lmiting based on server_name again

PR #958, opened 2017-09-28
Whitelist web apps on NIC

emord:
Changes from this branch:

```
@@ -18,10 +18,16 @@
   server 10.247.24.20:8181;
 }
 

-map $http_authorization $digest_submission {
-  ""     $server_name;
-  default     "";
-}
+geo $is_remote {
+  default 1;
+      10.247.24.0/24 0;
+  }
+
+map "$is_remote:$http_authorization" $digest_submission {
+      "~^1:" $server_name;
+      "~^1:.+" "";
+      default "";
+  }
```

What this means to me:
1) assign $is_remote a value of 1 if remote or a value of 0 if a local IP
2) If a request is from a remote IP and has a blank HTTP authorization header, use $server_name to rate limit
3) if a request is from a remote IP and has an HTTP authorization header, do not rate limit
4) If it is a local IP do not rate limit

sources:
mapping multiple variables https://serverfault.com/questions/760380/how-to-specify-several-variables-in-the-map-directive-of-nginx
whitelisting by ip example: https://www.nginx.com/blog/rate-limiting-nginx/

@snopoke buddies @gcapalbo @calellowitz

### DIFF
--- a/ansible/roles/nginx/templates/site.j2
+++ b/ansible/roles/nginx/templates/site.j2
@@ -18,8 +18,9 @@ upstream {{ balancer.name }} {
 
 {% for var_map in item.server.get('mappings', []) %}
 map {{ var_map.variable }} {{ var_map.name }} {
-  ""     {{ var_map.value }};
-  default     "";
+  {% for mapping in var_map.mappings %}
+    {{ mapping.key }} {{ mapping.value }};
+  {% endfor %}
 }
 {% endfor %}
 
--- a/ansible/roles/nginx/templates/site.j2
+++ b/ansible/roles/nginx/templates/site.j2
@@ -16,6 +16,13 @@ upstream {{ balancer.name }} {
 }
 {% endfor %}
 
+geo $is_remote {
+  default 1;
+  {% for local_ips in item.server.get('local_ips', []) %}
+    {{ local_ips }} 0;
+  {% endfor %}
+}
+
 {% for var_map in item.server.get('mappings', []) %}
 map {{ var_map.variable }} {{ var_map.name }} {
   {% for mapping in var_map.mappings %}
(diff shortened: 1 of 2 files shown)


ansible/roles/nginx/templates/site.j2
```diff
@@ -16,10 +16,18 @@ upstream {{ balancer.name }} {
 }
 {% endfor %}
 
+geo $is_remote {
```
snopoke: could put this whole block in an if statement that checks if `local_ips` is set

ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -12,10 +12,18 @@ nginx_sites:
...
+    - '10.247.24.0/24'
    mappings:
     - name: "$digest_submission"
-      variable: "$http_authorization"
-      value: "$server_name"
+      variable: '"$is_remote:$http_authorization"'
+      mappings:
+        - key: '"~^1:"'
+          value: "$server_name"
+        - key: '"~^1:.+"'
```
snopoke: Is this mapping needed? Since it has the same results as the default I think you can leave it out.

ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -103,6 +109,11 @@ nginx_sites:
          - zone_name: "submissions"
            burst: 100
            nodelay: True
+     - name: "/hq/admin/session_details"
+       balancer: "{{ deploy_env }}_cas_commcare"
+       proxy_redirect: "http://{{ CAS_SITE_HOST }} https://{{ CAS_SITE_HOST }}"
+       proxy_next_upstream_tries: 1
+       proxy_read_timeout: 900s
```
snopoke: strange that this should need a separate zone since the traffic is from an local IP.

emord: Because we have multiple rate limiting configurations the IP white list will need to be added to all of them. This was hitting our generic '/' rate limiting (by server_name) so it needed to be pulled out https://github.com/dimagi/commcarehq-ansible/blob/3664dd35f550937df8e230295022321c06c8054f/ansible/roles/nginx/vars/cas_ssl.yml#L117-L124

emord:
This didn't appear to work. Seems that it was rate limiting far more today than it should have

PR #959, opened 2017-09-28
Add BurnIO scale

millerdev:
Hopefully this will allow me to create more IO disruption on hosts with very low IO activity. Worked very well when tested locally on vagrant.

@snopoke


millerdev:
just realized they're all writing to the same file... need to fix that

millerdev:
fixed

PR #960, opened 2017-09-28
Couchdb2 clean install

dannyroberts:
we also noticed that lvm role was formatting drives as FAT12, because the last few args were off by one. Cal checked and it seems to be affecting at least ICDS as well.


ansible/roles/couchdb2-preinstall/tasks/main.yml
```diff
@@ -0,0 +1,17 @@
...
+    - ndg-httpsclient
+
+- name: Create couchdb2 dirs
+  sudo: yes
+  file:
+    path: "{{ item }}"
+    state: directory
+    mode: 0755
+    owner: "{{ couchdb_user }}"
+    group: "{{ couchdb_group }}"
```
dannyroberts: Now that i think of it, this might not work because this user & group probably won't be created by this point on a clean install. @emord not sure if we want to move some of this into your ansible-galaxy package

emord: I don't think it matters if the parent_srcs_dir is owned by the couchdb user.

It makes sense to move teh data directory creation to that. I think that stopped working after I made this change https://github.com/andrewrothstein/ansible-couchdb-cluster/commit/826df6010e191f17c35be598e9e4ba79d538c47f

dannyroberts: thanks for the tip. I'll PR a change to explicitly create that dir to https://github.com/andrewrothstein/ansible-couchdb-cluster then, since it was relying on lazy creation before.

ansible/roles/lvm/tasks/main.yml
```diff
@@ -15,7 +15,7 @@
 
 - name: Create partition on block storage volume
   sudo: yes
-  shell: echo -e "o\nn\np\n1\n\n\nt\n1\n8e\nw" | fdisk {{ item.item }}
```
javierwilson: @dannyroberts this worked fine last time I deployed (ctrls) did you use 14.04 to test? fdisk changed this behavior (smarter autoselect partition number) in a commit, this is why our LVM is a bit risky or dangerous, for example it would fail if server language is set to French.

dannyroberts: Really? I think @calellowitz and I checked enikshay (ctrls) and saw some FAT12-formatted drives there as well, presumably for the same reason. When you said it worked fine, do you mean it didn't halt mid-way and you did end up with a consolidated drive, or do you mean you're sure they were formatted as Linux (or whatever the target format is)?

javierwilson: i see... yeah, that's what i meant by "it worked", you're right, partitions were created with FAT12 type, but just so we're clear they were all correctly formatted as ext4. as far as I understand, otherwise they would have failed.

dannyroberts: Ah, interesting. I wasn't making a distinction between those two things. Sounds like this change is still technically an improvement, though, even if it's inconsequential in the scheme of things, right?

ansible/requirements.txt
```diff
@@ -1,4 +1,4 @@
-ansible==2.2.1
+ansible==2.4
```
dannyroberts: Anyone see an issue with this? I'm going to run a deploy to the couch machine with this to confirm it doesn't cause any problems.

ansible/roles/lvm/tasks/main.yml
javierwilson: yeah, absolutely

ansible/requirements.txt
javierwilson: perhaps some deprecated warnings will become errors now, but i think it's a good to go for it even if it requires extra work to update some playbooks

dannyroberts: Hmm, so looks like we can't do this because there's an active bug in the monit module in 2.4 that prevents monit 5.18 and below from working; we're on monit 5.6 and that's the latest available in apt.

dannyroberts:
@javierwilson Added a number of changes here, all of this is already deployed.

- I made a fix to `andrewrothstein.ansible-couchdb-cluster`, got it merged, and then updated to the new commit
- Other changes had been made to that that now made it require ansible 2.4
- So I updated to ansible 2.4
- Ansible 2.4 had a bug in it that broke monit commands for us, so I fixed that; that's now merged into the `devel` branch and will be in 2.4.1. when that's released
- In the meantime, I forked to dimagi/ansible, checked out the 2.4.0 release (`v2.4.0.0-1`) and cherry-picked my commit onto it, and pinned our ansible version (in this repo) to _that_.

ansible/roles/rabbitmq/tasks/main.yml
```diff
@@ -1,29 +1,29 @@
 ---
 # Add and configure rabbitmq
 - name: Add rabbitmq official apt repository
-  sudo: true
+  become: true
```
emord: why become:true here and become:yes elsewhere (no action needed. existed before)

ansible/vars/swiss/swiss_public.yml
```diff
@@ -21,6 +21,10 @@ active_sites:
   riakcs: False
   couchdb2: False
 
+couchdb2:
```
emord: Swiss doesn't have this yet right? also would need the cluster settings I think

dannyroberts: good catch. I'm not sure whether it makes a difference, but yeah I shouldn't have included this commit in this PR. Reverting here: https://github.com/dimagi/commcarehq-ansible/pull/968

ansible/roles/rabbitmq/tasks/main.yml
dannyroberts: `yes` and `true` evaluate to the same thing in yaml (== json `true`), so it doesn't matter which one you use. As to why we don't just pick one... I think we've just never bothered to agree on a repo-wide preference for one or the other.

ansible/roles/couchdb2-preinstall/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+- name: Install ndg-httpsclient so we can download couch
```
javierwilson: nice work-around...

PR #962, opened 2017-09-29
[WIP] use streaming replication for pg standbys

snopoke:


ansible/roles/postgresql/tasks/main.yml
```diff
@@ -74,49 +87,46 @@
...
-    role_attr_flags: CREATEDB
-    state: present
-    port: "{{ postgresql_port }}"
-  when: not is_pg_standby|default(False)
-  tags:
-    - postgres_users
+- name: restart postgresql if necessary
+  become: true
+  service: name=postgresql state=restarted
+  when: pg_config_changed.changed and not pg_started.changed
```
snopoke: This feels very hacky but a restart is required after setting `max_replication_slots` before any replication slots can be created.

ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -36,19 +36,19 @@ vacuum_cost_delay = 20ms
 
 # WRITE AHEAD LOG
 
-{% if hot_standby_server|default(None) %}
 archive_mode = on
-archive_command = 'rsync -a %p postgres@{{ hot_standby_server }}:{{ archive_dir }}/%f'   # archive directory lives on first standby machine specified in the inventory file
+archive_command = 'test ! -f {{ postgresql_archive_dir }}/%f && cp %p {{ postgresql_archive_dir }}/%f'
```
snopoke: This changes the archiving to keep them locally instead of shipping them to the standby. PG will take care of removing them as needed according to any replication slots that are in use.

ansible/vars/dev/dev_public.yml
```diff
@@ -89,10 +89,23 @@ backup_es: False
...
+    password: devreadonly
+    role_attr_flags: 'NOSUPERUSER,NOCREATEROLE,NOCREATEDB'
+    privs:
+      -
+        privs: SELECT
+        objs: ALL_IN_SCHEMA
+  replication:
+      username: 'hqrepl'
+      password: 'hqrepl'
+      role_attr_flags: 'LOGIN,REPLICATION'
```
snopoke: Similar changes were applied to the vault files

ansible/vars/dev/dev_public.yml
```diff
@@ -89,10 +89,23 @@ backup_es: False
...
+
+pg_replication_slots:
+ 192.168.33.16:
+   - standby1
```
snopoke: Since this is per host it could be set in the inventory or in a group vars file for each host.

ansible/roles/postgresql/tasks/main.yml
millerdev: Assuming you're saying that this is hacky because it restarts postgres on _any_ configuration change, not just a change in `max_replication_slots`. If that's correct, would it be possible to read `max_replication_slots` from the config file at the beginning of the ansible run and only restart if it changed?

snopoke: Switched to use `flush_handlers` (4499137)

snopoke:
I still want to test this locally before merging.

PR #963, opened 2017-10-01
Switch to using differential backups

gcapalbo:
From what I'm seeing on enikshay, the incremental backups are taking up almost the same amount of room as the the full backups (looking at `repository backup size` in `pgbackrest info`).

I also don't see a way to expire incremental backups until the full backups they depend on expire, so this is causing us to run out of space on the standby machine unless we aggressively expire full backups.

What I'm proposing is that we switch from using incremental backups to differential backups so that we can use `retention-diff` to expire them. With this approach, based on our default settings in our ansible repo, we would then keep 4 weeks of weekly backups (`retention-full`) and 3 days of daily backups (`retention-diff`), mimicking the old pg basebackup method.

@javierwilson @snopoke

### DIFF
--- a/ansible/roles/pgbackrest/tasks/main.yml
+++ b/ansible/roles/pgbackrest/tasks/main.yml
@@ -71,8 +71,8 @@
 - name: Create Daily Cron job
   sudo: yes
   cron:
-    name: "Backup postgres daily (incremental)"
-    job: "/usr/bin/pgbackrest --type=incr --stanza=backup backup"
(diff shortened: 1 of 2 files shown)


gcapalbo:
If there's no objections I'll deploy this to enikshay tomorrow before the daily backup runs

gcapalbo:
deployed this change and https://github.com/dimagi/commcarehq-ansible/pull/961 to softlayer as well since we're running into space issues there too

PR #964, opened 2017-10-03
Vagrant setup of multiple ES clusters

sravfeyn:


sravfeyn:
I was able to get this working, I couldn't test port forwarding of ES yet

@snopoke

PR #965, opened 2017-10-03
Je/dd apm

emord:
buddy @millerdev 

docs are here: http://pypi.datadoghq.com/trace/docs/#module-ddtrace.contrib.django


emord:
@millerdev We're doing the 2 week trial for some dashboard load testing on ICDS. It's not clear if we'll keep it once that's done

PR #966, opened 2017-10-05
couch setup tweaks

snopoke:


nickpell:
looks good but there are merge conflicts

ansible/roles/couchdb2/tasks/main.yml
```diff
@@ -14,11 +14,26 @@
...
+    user: "{{ couchdb2.username }}"
+    password: "{{ couchdb2.password }}"
+    force_basic_auth: true
+    status_code: 201, 412 # 412 means already set up (conflict)
+    body: '{}'
+  when: inventory_hostname == groups.couchdb2.0
+  with_items:
+   - _users
+   - _replicator
+   - _global_changes
```
dannyroberts: Really, this is the solution?! Couch makes us create its built-in dbs for it? Some things about couch just still blow my mind...

snopoke: It does create them for you under some circumstances but when I ran the playbook they weren't created.

It seems like couch2.0.0 has only half implemented cluster setup functionality.

PR #973, opened 2017-10-11
Move Sentry properties into sentry.properties, add proxy settings

wpride:
Try to get Sentry working for ICDS

cc @dannyroberts @snopoke 

**DO NOT MERGE** I have never submitted an Ansible PR before and have no idea how to test. 

image


ansible/roles/formplayer/templates/sentry.properties.j2
```diff
@@ -0,0 +1,5 @@
+sentry.dsn={{ formplayer_sentry_dsn }}
+{% if http_proxy_address is defined %}
+http.proxy.host={{ http_proxy_address }}
+http.proxy.port={{ http_proxy_port }}
+{% endif %}
```
dannyroberts: What is `sentry.properties.j2`? Is it used somewhere?

Or, from the other side, what do you want to end up on the server / in the formplayer directory?

snopoke:
In addition to this you'll need a task that copies the file to the right location on the server such as this one for the 'application.properties' file: https://github.com/dimagi/commcarehq-ansible/blob/43bd9e91ee7200ee5062afe2d59d7715a221a335/ansible/roles/formplayer/tasks/main.yml#L43

You can probably just re-purpose that task to do both of them as follows:
```

- name: Copy Spring application.properties
  become: yes
  template:
    src: "{{ item.src }}"
    dest: "{{ formplayer_build_dir }}/{{ item.dest}} "
    group: cchq
    owner: cchq
    mode: 0644
  tags:
    - fp-localsettings
    - localsettings
  with_items:
    - {'src': 'application.properties.j2', 'dest': 'application.properties'}
    - {'src': 'sentry.properties.j2', 'dest': 'sentry.properties'}
```

wpride:
@snopoke Thanks! Updated, and also returned `sentry.dsn` to `application.properties`

ansible/roles/formplayer/templates/sentry.properties.j2
```diff
@@ -0,0 +1,4 @@
+{% if http_proxy_address is defined %}
+http.proxy.host={{ http_proxy_address }}
+http.proxy.port={{ http_proxy_port }}
```
dannyroberts: I assume these two variables exist?

wpride: Danny, when you ASSUME you make an ASS of U and ME

wpride: I found them here https://github.com/dimagi/commcarehq-ansible/blob/3567e2b0946f8c23fc9570e287868d52e2dbe261/ansible/vars/icds/icds_public.yml#L193-L193

wpride:
@dannyroberts OK, despite the documentation I could find no evidence/other mentions of `sentry.properties` being a valid place for these properties. So, I've moved them into `application.properties`. Locally this seems to be working (invalid proxy fails, valid proxy works, no proxy works). LMK when you want to give this another go.

PR #983, opened 2017-10-16
kill processes when removing users

snopoke:


ansible/roles/users/tasks/remove_users.yml
```diff
@@ -0,0 +1,41 @@
+---
+
+- name: Get running processes
+  shell: "ps aux | grep -v grep | grep -Ew \"{{ dev_users.absent|join('|') }}\" | awk '{print $2}'"
```
kaapstorm: You might want `grep -Ew \"^{{ dev_users.absent|join('$|^') }}$\"` in case we ever have a dev whose username is a substring of another dev's username.

kaapstorm:
Just the one comment, otherwise LGTM.

ansible/roles/users/tasks/remove_users.yml
snopoke: I don't think I can do that since we're not matching whole lines

kaapstorm: nvm. `grep -w` already only matches whole words

ansible/roles/users/tasks/remove_users.yml
```diff
@@ -0,0 +1,41 @@
...
+  register: running_processes
+
+- name: Kill running processes
+  become: yes
+  shell: "kill {{ item }}"
+  with_items: "{{ running_processes.stdout_lines|default([]) }}"
+
+- wait_for:
```
millerdev: should make this timeout after 30 or 60s instead of 300

snopoke: Made thee timeout 15s (which is for each process)

PR #985, opened 2017-10-20
Revert "Je/dd apm"

emord:
Reverts dimagi/commcarehq-ansible#965

Turning this off since we're done with load testing


emord:
Reverted this and also updated the datadog api key with what was actually on ICDS

emord:
merging so we don't get charged for APM

PR #986, opened 2017-10-23
stop wal archiving

calellowitz:
@snopoke @javierwilson cc: @kaapstorm 
The replication slots should ensure the correct number of WAL logs (i know thats redundant but it sounds better in my head) are kept for the standby, and nothing currently cleans up our WAL archive, which means these continue to pile up and have almost filled the disks on 3 of our machines. There is also nothing configured to read from this archive so it doesn't seem to serve much purpose. The postgres docs seem to imply that replication slots supersede WAL archiving for standbys.

Longer term if we want to keep this, it should be possible to properly cleanup but in the short term i think it makes more sense to get rid of it until we have a plan on what to do with it.

### DIFF
--- a/ansible/roles/postgresql/templates/postgresql.conf.j2
+++ b/ansible/roles/postgresql/templates/postgresql.conf.j2
@@ -36,10 +36,6 @@ vacuum_cost_delay = 20ms
 
 # WRITE AHEAD LOG
 
-archive_mode = on
-archive_command = 'test ! -f {{ postgresql_archive_dir }}/%f && cp %p {{ postgresql_archive_dir }}/%f'
-archive_timeout = 60
-
 wal_level = hot_standby
 max_wal_senders = 3
 checkpoint_segments = {{ postgresql_checkpoint_segments|default(16) }}


emord:
reasoning seems good to me, but would need @javierwilson @snopoke feedback tomorrow to know what to do long term

snopoke:
Did some more reading on this and it seems that the archiving is only necessary when setting up the standby to make sure that there are enough logs between the beginning of online-backup and the startup of streaming replication. If `pg_basebackup --xlog-method=stream` is used to create the standby then this probably isn't necessary (assuming the standby is started up immediately).

The `archive_command` and `wal_keep_segments` are both involved in this decision. See comments below:

```
# To prevent the primary server from removing the WAL segments required for
# the standby server before shipping them, set the minimum number of segments
# retained in the pg_xlog directory. At least wal_keep_segments should be
# larger than the number of segments generated between the beginning of
# online-backup and the startup of streaming replication. If you enable WAL
# archiving to an archive directory accessible from the standby, this may
# not be necessary.
wal_keep_segments = 32

# Enable WAL archiving on the primary to an archive directory accessible from
# the standby. If wal_keep_segments is a high enough number to retain the WAL
# segments required for the standby server, this is not necessary.
archive_mode    = on
archive_command = 'cp %p /path_to/archive/%f'
```

ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -36,10 +36,6 @@ vacuum_cost_delay = 20ms
 
 # WRITE AHEAD LOG
 
-archive_mode = on
-archive_command = 'test ! -f {{ postgresql_archive_dir }}/%f && cp %p {{ postgresql_archive_dir }}/%f'
-archive_timeout = 60
-
```
gcapalbo: @javierwilson isn't this is going to interfere with pgbackrest? https://github.com/dimagi/commcarehq-ansible/pull/917/files#diff-1de0fd6b211ae1ef3b9020f83c0635d0R32

PR #987, opened 2017-10-25
digest auth rate limiting for restores

snopoke:
Already applied to nginx conf


ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -18,13 +18,13 @@ nginx_sites:
       value: "$server_name"
    limit_req_zones:
     - name: "restore"
-      zone: "$server_name"
+      zone: "$digest_submission"
```
emord: Since these have the same zone, doesn't that mean that they will effect each others limits?

snopoke: This just changes the zone key. The result will look like this:
```
limit_req_zone $digest_submission zone=restore:10m rate=2r/s;
limit_req_zone $digest_submission zone=submissions:10m rate=10r/s;
limit_req_zone $server_name zone=server_name:10m rate=25r/s;
limit_req_zone $server_name zone=app_downloads:10m rate=50r/s;
```

The zone names are still unique.

PR #990, opened 2017-10-27
couch_dbs now run on 15984

taylordowns2000:
Hey @snopoke , it looks like couchdb2 should now be on 15984. am i reading this right?


emord:
Thanks for catching that. This is correct in our default setup, but you should be able to override it using these variables https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/couchdb2/vars/main.yml#L2-L3

snopoke:
Also when running in production we put nginx in front of couchdb2 so that we can load balance across multiple nodes. You can set that up by adding this to the `couchdb2: True` list:
```
active_sites:
  couchdb2: True
```

PR #991, opened 2017-10-27
makes sure pgbouncer is restarted after reboot

javierwilson:


ansible/roles/postgresql/tasks/main.yml
```diff
@@ -180,6 +180,16 @@
   tags:
     - pgbouncer
 
+- shell: /bin/true
+  register: no_tags
```
kaapstorm: This is a nice approach! Are there other things that need to happen once, that we can use this with too?

snopoke: can you elaborate on whet this is for?

javierwilson: this will only be run when using no tags, this allows to run tasks only when a specific tag is used.

snopoke: interesting, I think it's worth a comment (maybe even link to the SO answer: https://serverfault.com/a/748864)

emord: I think this is causing pgbouncer to restart unconditionally (but not during `--check`). I ran deploy_db with `--tags=pgbouncer` and in check mode it said this restart was skipped, but when I ran without `--check` pgbouncer was restarted. I haven't figured it out why that is though :/

snopoke: https://github.com/dimagi/commcare-cloud/pull/3151

PR #992, opened 2017-10-30
deploy netvault

javierwilson:
not perfect but will help

### DIFF
--- a/ansible/deploy_netvault.yml
+++ b/ansible/deploy_netvault.yml
@@ -9,7 +9,7 @@
     - name: "Unarchive Netvault"
       unarchive:
         src: /tmp/netvault.tar.gz
-        dest: /tmp/netvault/
+        dest: /tmp/
         remote_src: yes
     - name: "Create symlink for lib64"
       file:
--- a/ansible/deploy_netvault.yml
+++ b/ansible/deploy_netvault.yml
@@ -2,24 +2,58 @@
   hosts: pg_standby
   become: true
   tasks:
+
     - name: "Download Netvault"
       get_url:
         url: "{{ netvault_url }}"
         dest: /tmp/netvault.tar.gz
+
     - name: "Unarchive Netvault"
       unarchive:
         src: /tmp/netvault.tar.gz
         dest: /tmp/
         remote_src: yes
+
     - name: "Create symlink for lib64"
       file:
         src: /usr/lib
         dest: /usr/lib64
         state: link
+
     # FIXME: deal with prompt for options
     - name: "Runs install script"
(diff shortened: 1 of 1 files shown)


ansible/deploy_netvault.yml
```diff
@@ -0,0 +1,24 @@
+- name: "Install Netvault Client"
+  hosts: pg_standby
+  become: true
+  tasks:
+    - name: "Download Netvault"
+      get_url:
+        url: "{{ netvault_url }}"
```
snopoke: where is `netvault_url`  defined?

javierwilson: should be defined in vault, i think but i am not sure this is a long-lasting variable, so maybe extra_vars?

snopoke: Is it a 'private' url? If not can just put it in here directly since this is only going to be run on ICDS.

javierwilson:
I've added configuration of user netvault. I am using extra-vars for now, will add the vault once it is confirmed that it works (meaning meets the requirements of netvault server) @sravfeyn

ansible/deploy_netvault.yml
```diff
@@ -2,24 +2,58 @@
...
+  tasks:
+
+    - name: "Create PostgreSQL netvault user"
+      become_user: postgres
+      vars:
+        ansible_ssh_pipelining: true
+      postgresql_user:
+        name: "{{ netvault_username }}"
+        password: "{{ netvault_password }}"
+        role_attr_flags: SUPERUSER
```
calellowitz: is a read only user not possible?

ansible/deploy_netvault.yml
```diff
@@ -2,24 +2,58 @@
...
+
     # FIXME: deal with prompt for options
     - name: "Runs install script"
       #command: /tmp/netvault/install
       command: /bin/true
       when: false
+
+    - name: "Grants PostgreSQL access to netvault user"
+      blockinfile:
+        path: "{{ postgresql_config_home }}/pg_hba.conf"
```
calellowitz: any reason not to add this to our pg_hba.conf template

ansible/deploy_netvault.yml
javierwilson: according to the documentation netvault needs a superuser, this is a superuser on a standby, i hope it's enough... not sure.

ansible/deploy_netvault.yml
javierwilson: i thought about it. but since i am running this using extra_vars didn't want to mess with that template.  i'll improve this once we get confirmation that it is an accepted solution.

PR #993, opened 2017-10-30
making sure services are running and not just started

javierwilson:
runs various network, os and app tests.  tests connectivity among hosts.


ansible/ping.yml
```diff
@@ -115,6 +115,19 @@
       warn: False
 
 - hosts:
+  - elasticsearch
+  tasks:
+  - uri: url="http://{{ inventory_hostname }}:9200/_cluster/health" return_content=yes
```
emord: We may want to have a wait on this. If ES has just started it could take a while for it to become green. Something like check every 5 seconds and timeout after x seconds would probably be ideal

javierwilson: the way i see "ping" is like something to run when we firefight, to detect if something is off. it should check that all services are running and running well.

emord: Ah. that makes sense. Didn't realize this was in ansible/ping.yml

PR #994, opened 2017-10-30
Postgres logging

emord:
https://trello.com/c/7OWkUCFl/173-improve-postgres-log-rotation-to-avoid-disk-filling-up

@calellowitz @snopoke @javierwilson A starting point for this. I haven't fully tested it being deployed, but think the basic idea is here. Workflow will be

1) postgres starts logging to files with the format postgresql-timestamp.log
2) postgres starts writing to a new log file after 100 MB
3) Every 15 minutes a script runs:
  a) it removes the the log files except for the 20 newest ones
  b) out of those 20 it gzips the 17 newest ones

Can easily change the 100 MB / 20 log files, but wasn't sure how much we want to keep around

buddy @mkangia

### DIFF
--- a/ansible/roles/postgresql/tasks/main.yml
+++ b/ansible/roles/postgresql/tasks/main.yml
@@ -20,7 +20,6 @@
 
 - name: generate en_US.UTF-8 locale
   locale_gen: name='en_US.UTF-8' state=present
-
 - name: Install PostgreSQL & dependencies
   apt: name="{{ item }}" state=present
   with_items:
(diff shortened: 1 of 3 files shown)


snopoke:
Just curious, did you consider using syslog to do the logging?

emord:
@snopoke Didn't really consider it because in their docs they say:

> On many systems, however, syslog is not very reliable, particularly with large log messages; it might truncate or drop messages just when you need them the most. Also, on Linux, syslog will flush each message to disk, yielding poor performance. (You can use a "-" at the start of the file name in the syslog configuration file to disable syncing.)

https://www.postgresql.org/docs/9.2/static/logfile-maintenance.html

calellowitz:
Dont know the state of this but just wanted to flag that enikshay db0 ran out of space again yesterday because of the same logging issues, so would love to push this forward if we can.

snopoke:
I think this is GTG. Just needs to be rolled out.

PR #1000, opened 2017-10-31
Change couchbd2 compaction

emord:
We've had enough issues with synclogs not compacting that I'm not convinced this works, so I'm thinking about changing all of these to 10% which means that once the file size is 10% more than the data size it will compact. this should not really affect other dbs that much since they have a lot less data than synclogs. thoughts?

@dannyroberts @snopoke


snopoke:
Seems worth trying, strange that the other one wouldn't work

PR #1001, opened 2017-11-02
Optimize pg mem settings on enikshay server

millerdev:
Not yet deployed. More background here: https://docs.google.com/document/d/13Jlei5joToyME4DbnlHxRhQEpnrhUWvlQCtEDklz1po/edit#

This will require postgres downtime (restart).

@snopoke @dimagi/devops

### DIFF
--- a/ansible/vars/enikshay/enikshay_public.yml
+++ b/ansible/vars/enikshay/enikshay_public.yml
@@ -84,6 +84,9 @@ nofile_limit: 65536
 
 postgres_users: "{{ secrets.POSTGRES_USERS }}"
 
+postgresql_shared_buffers: '10GB'
+postgresql_max_stack_depth: '6MB'
+postgresql_effective_cache_size: '32GB'
 postgresql_max_connections: 600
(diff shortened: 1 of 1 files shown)


calellowitz:
@millerdev 
I imagine the riak process on this machine is also trying to cache as much as possible, do you have any concerns that having postgres plan queries assuming this much cache, will cause it and riak to fight over the cache more (leading to many of the same cache churn issues)

millerdev:
@calellowitz If riak and postgres are fighting over the OS cache (which I'm fairly confident they already are), I don't believe there's anything we can do about that other than move them to separate machines or add more RAM to the machine until the OS can cache parts of the disk that they access frequently. I think this server doesn't even have enough RAM to accommodate all commonly accessed data used by postgres shard dbs, so some amount of memory contention seems likely.

calellowitz:
@millerdev Makes sense to me. Thanks for that explanation

millerdev:
My plan is to roll this out on Monday after 2pm. Will contact the support team about making a downtime announcement.

PR #1002, opened 2017-11-02
add package sysstat to common installs

jhoffman-dimagi:
adding sysstat to our base os installs so we can gather better metrics on IO


jhoffman-dimagi:
@emord    had to do this over a few times to get it right, but should be quicker checking in changes down the road, thanks for adding this to the deploy, if it looks good please accept.

emord:
This failed on two servers (.16 and .26) due to some apt key not being up to date i think. not related to this PR though

PR #1003, opened 2017-11-02
Upgrade ansible 2.4.1

emord:
@dannyroberts your pr was released.

side note that we should probably pin the others to a version

buddy @snopoke


dannyroberts:
🎉

PR #1004, opened 2017-11-03
PNA site to submit configuration

javierwilson:


javierwilson:
you mean the fact that we get keys from vault or files? that's just to make our transition to only-vault easier.  If you mean having to configure almost an identical YML for each server, yeah, i'm with you, we should come up with something better DRY-ish.

snopoke:
yea, the duplicate yml, not the keys in vault files.

PR #1005, opened 2017-11-03
reject form submissions and restore requests for IPM senegal

snopoke:


ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -20,6 +20,10 @@ nginx_sites:
    add_header: "X-Frame-Options SAMEORIGIN"
    access_log: "{{ log_home }}/{{ deploy_env }}-timing.log timing"
    locations:
+    - name: "/a/ipm-senegal/phone/restore"
```
emord: This looks right. Any thoughts about also doing an HQ PR to check the domain on submission and restores?

snopoke: Good idea, can't be too careful.

emord:
Deployed this to prod. verified that both of these 503 and bring up the maintenance page

PR #1008, opened 2017-11-06
disable temp file logging

calellowitz:
@snopoke @millerdev @emord

### DIFF
--- a/ansible/roles/postgresql/templates/postgresql.conf.j2
+++ b/ansible/roles/postgresql/templates/postgresql.conf.j2
@@ -76,7 +76,7 @@ log_connections = on
 log_disconnections = on
 log_line_prefix = '%t '
 log_lock_waits = on
-log_temp_files = 0
+log_temp_files = -1
 
 
 # RUNTIME STATISTICS


calellowitz:
Described this in an email that I believe you were all linked to but just to have it here as well, when db0 ran out of space sunday a lot of the log files includes hundreds of temp files created for each query, so it was logging the same massive query 276 times in some instances. This should dramatically cut back on what we log so that even when this query gets slow we dont fill up our disk.

millerdev:
This seems reasonable, but I think we might be shooting the messenger. Queries should not be writing massive amounts of temp files like that, so it's likely a sign of other bad configuration.

millerdev:
I see now the default is -1, and I don't have any idea how many temp files are created during normal operation in a well-tuned system, so maybe better to revert to the default for now.

calellowitz:
I think you are right about shooting the messenger and in an ideal world (or even just when we get proper log rotation) we can bring this back. It just feels like in the short term this is a good way to retain info about what queries are slow without logging them hundreds of times each, and given the number of times the enikshay disks have filled up already, this does not really feel optional. We should be able to tell from `EXPLAIN` or `EXPLAIN ANALYZE` anyway if a slow query is using temp files.
Given your last comment im going to go ahead and merge to get it into your pg settings deploy, but I think its a good idea to revisit this once things settle down.

PR #1009, opened 2017-11-07
use language code as cluster setting

javierwilson:


snopoke:
Have we tested this locally? I don't think we've ever done this before.

javierwilson:
Besides changing the label from English to French, it doens't really change any text, is this the wrong way to change a default language on installation? Or should we change HQ to take this into account? @snopoke

snopoke:
This just sets the fallback value. The actual lang is determined by the logged in user if possible.

PR #1011, opened 2017-11-07
Update postgres log file pattern

emord:
@millerdev you'll want this before you deploy to enikshay

@snopoke


ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -61,7 +61,7 @@ effective_cache_size = {{ postgresql_effective_cache_size }}
 log_destination = 'csvlog'
 logging_collector = on
 log_directory = 'pg_log'
-log_filename = 'postgresql-%x-%r.log'
```
emord: before: could not open log file "pg_log/postgresql-11/07/17-07:37:32 PM.log": No such file or directory

ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -61,7 +61,7 @@ effective_cache_size = {{ postgresql_effective_cache_size }}
...
+log_filename = 'postgresql-%Y-%m-%d_%H%M%S.log'
```
emord: This is the default. specified https://www.postgresql.org/docs/9.1/static/runtime-config-logging.html#RUNTIME-CONFIG-LOGGING-WHERE

emord:
deployed to NIC

PR #1012, opened 2017-11-08
Fix postgres rotation script

emord:
@snopoke @javierwilson this is spamming emails. there's not really an issue with that, it just means these aren't getting removed/gzipped today. apparently I had an alias for gz locally. it should be gzip.


javierwilson:
i think we'd have to use -name "*.csv", sorryr didn't see this one, created my own branch

snopoke:
Closing in favor of #1014

PR #1021, opened 2017-11-10
cache facts

snopoke:
This speeds up successive playbook runs and also runs that span multiple playbooks

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -2,3 +2,8 @@
 ansible_managed = Ansible managed, do not edit directly
 roles_path = /opt/ansible_roles
 log_path=/var/log/ansible.log
+
+gathering = smart
+fact_caching = jsonfile
+fact_caching_connection = /tmp/ansiblecache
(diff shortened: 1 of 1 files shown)


dannyroberts:
All my ansible commands started failing like this

```
(ansible) droberts@control:~/commcarehq-ansible/ansible$ ENV=staging && ap --diff --check deploy_localsettings.yml --tags=localsettings
Vault password:
ERROR! error in 'jsonfile' cache, configured path (/tmp/ansiblecache) does not have necessary permissions (rwx), disabling plugin
```

and I noticed `/tmp/ansiblecache` was owned by `jemord:jemord` with permissions 775. Running `chown -R jemord:dimagidev /tmp/ansiblecache` as root fixed this for me, but I've only done that on production control machine.

Just wanted to give a heads-up

snopoke:
Thanks. I actually meant to change this to cache in the user's home. I'll
PR that tomorrow.

snopoke:
https://github.com/dimagi/commcarehq-ansible/pull/1036

PR #1025, opened 2017-11-10
rework pgbackrest

snopoke:
I took a shot at re-working the `pgbackrest` role to incorporate it into the `postgresql` role along with the `pg_standby` role (which was mostly backup related).

The goal here is to be able to merge this now even though we don't have pgbackrest rolled out everywhere.

This should also make it less prone to errors if someone doesn't include the right tags.

I still need to test this.


snopoke:
Tested this locally and looking good. Will do check tests on other envs next week before merging.

ansible/deploy_db.yml
```diff
@@ -16,7 +16,7 @@
     - {role: backups, tags: 'backups'}
 
 - name: Postgresql
-  hosts: postgresql
+  hosts: postgresql,pg_standby
```
emord: +1 really like this. although it still may have issues when using `--limit=pg0` or something similar

snopoke:
ran on enikshay, going to merge now

ansible/roles/postgresql/tasks/backup.yml
```diff
@@ -0,0 +1,20 @@
...
+  become: yes
+  file:
+    path: '{{ postgresql_backup_dir }}'
+    owner: postgres
+    group: postgres
+    mode: 0755
+    state: directory
+
+- include_tasks: backup_plain.yml
+  when: backup_postgres == 'plain' and (is_pg_standby or not hot_standby_server|default(False))
```
dannyroberts: @snopoke this line seems to be failing for my quick aws env setup (which I'm doing without a `pg_standby`). Can you explain this line to me / when this is supposed to run?

My understanding is `hot_standby_server` is only ever set as a variable on a `pg_standby` host, so if `is_pg_standby` is false, then `hot_standby_server` is undefined, `hot_standby_server|default(False))` is `False`, and `not hot_standby_server|default(False))` is `True`—so `(is_pg_standby or not hot_standby_server|default(False))` is always `True`. Is something in this logic inverted, or am I maybe missing something?

javierwilson: This will be true whenever the current host is a standby (is_pg_standby) or no standby is defined.  I will not be true when running on a master with no standby.

hot_standby_server is only set on a master (not a slave as you said), hot_standby_master is set on the slave.

javierwilson: I will work on creating a pg_backup group. to get rid of the "standby has to be backup server" constraint.  Does that sound like a good solution to you?

snopoke: As @javierwilson pointed out the `hot_standby_server` var is only set on the master, not the standby node. So the logic here is:
* if backup is 'on' and
  * we're on a standby node
  * or there is no standby node configured

dannyroberts: OK thanks, that helps, I should be able to debug from here. Sounds like part of what I was missing is that even if you don't have a `pg_standby`, we still do this backup, just on the same machine? (The other part was that I was confusing the variables `hot_standby_server` and `hot_standby_master`.)

PR #1027, opened 2017-11-10
Only send daily deploy for certain environments

sheelio:
@dannyroberts 
Feedback about the daily deploy emails, so changing them to only go eNikshay / ICDS / production. 

Does this work?

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -146,3 +146,4 @@ report_builder_add_on_email: sales@dimagi.com
 eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
+daily_deploy_email: null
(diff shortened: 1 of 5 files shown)


dannyroberts:
I don't think this approach is quite right.

These emails used to go to commcarehq-ops@, so rather than telling it not to send (which I don't think this PR _quite_ does, but that's the obvious intention), I'd say just change the email on those environments back to the old one. Does that work?

sheelio:
These emails actually still go to both (ops and daily announce).  This just turns it off for daily-announce but keeps the ops behaviour the same.

dannyroberts:
Ah, I see. In that case, I'd suggest making a change like https://github.com/dimagi/commcare-hq/pull/18497 and then just setting the value to `null` in the yml config files for the envs. As you have it now, the `settings.py` default (an `@example.com`) address will bleed through, and that's not really what we want.

ansible/group_vars/all.yml
```diff
@@ -146,4 +146,3 @@ report_builder_add_on_email: sales@dimagi.com
 eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
-daily_deploy_email: tech-announce-daily@dimagi.com
```
dannyroberts: I believe you can override this directly in the *`_public.yml` files, so you can just set it to `null` here and add `daily_deploy_email: tech-announce-daily@dimagi.com` in those three other files without changing `localsettings.py.j2`.

sheelio:
@dannyroberts 
Thanks - updated

PR #1029, opened 2017-11-13
makes sure ulimit can be set high

javierwilson:
for nginx, pgbouncer and redis

### DIFF
--- a/ansible/roles/postgresql/templates/pgbouncer.init.j2
+++ b/ansible/roles/postgresql/templates/pgbouncer.init.j2
@@ -1,0 +1,99 @@
+#!/bin/sh
+
+### BEGIN INIT INFO
+# Provides:          pgbouncer
+# Required-Start:    $local_fs $remote_fs $network $syslog $named
+# Required-Stop:     $local_fs $remote_fs $network $syslog $named
+# Should-Start:      postgresql
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Short-Description: start pgbouncer
+# Description: pgbouncer is a connection pool server and replication
+#              proxy for PostgreSQL.
+### END INIT INFO
+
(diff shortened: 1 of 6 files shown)


ansible/vars/dev/dev_public.yml
```diff
@@ -127,6 +127,9 @@ backup_blobdb: True
 postgres_s3: False
 nadir_hour: 18
 
+nofile_limit: 65536
+nginx_worker_rlimit_nofile: "{{ nofile_limit }}"
```
snopoke: where are these used?

ansible/roles/postgresql/templates/pgbouncer.init.j2
```diff
@@ -0,0 +1,99 @@
...
+
+  status)
+    status_of_proc -p $PIDFILE $DAEMON $NAME && exit 0 || exit $?
+    ;;
+
+  *)
+    echo "Usage: /etc/init.d/$NAME {start|stop|reload|force-reload|restart|status}"
+    exit 1
+    ;;
+esac
```
snopoke: What's different in this from the standard one?

ansible/vars/dev/dev_public.yml
javierwilson: nofile_limit is the general max files open and nginx_worker_rlimit_nofile only for nginx max files open,  maybe this one is redudant?

snopoke: OK, I thought you had introduced these in this PR but I see they are already in use.

ansible/roles/postgresql/templates/pgbouncer.init.j2
javierwilson: it applies ulimit if $ULIMIT exists (usually this means it has been set in /etc/defaults) I am copying this from how other services do it (redis, nginx)

PR #1030, opened 2017-11-14
 possibility of adding static routes to hosts

javierwilson:
FYI currently net_static_route only works on some network devices (cisco routers ans similar) not in debian, and it's still experimental :(

### DIFF
--- a/ansible/deploy_common.yml
+++ b/ansible/deploy_common.yml
@@ -32,3 +32,14 @@
   hosts: all
   roles:
     - {role: monit, tags: monit}
+
+- name: Configure static routes
+  hosts: all
+  tasks:
+  - name: Adds static routes
+    shell: 'if [ `ip route show {{ item.prefix }} | wc -l` -eq "0" ];  then route add -net {{ item.prefix }} gw {{ item.gateway }}; else echo "already_exists"; fi'
+    with_items: "{{ static_routes }}"
+    become: True
+    register: route_show
+    changed_when: '"already_exists" not in route_show.stdout'
+    tags: static_routes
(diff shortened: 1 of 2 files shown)


ansible/deploy_common.yml
```diff
@@ -32,3 +32,16 @@
   hosts: all
   roles:
     - {role: monit, tags: monit}
+
+- name: Configure static routes
+  hosts: all
+  tasks:
+  - name: Adds static routes
+    shell: 'if [ `ip route show {{ item.prefix }} | wc -l` -eq "0" ];  then route add -net {{ item.prefix }} gw {{ item.gateway }}; else echo "already_exists"; fi'
```
snopoke: rather than doing this in one command can we do it in 2 where the first one checks if it's necessary and the 2nd one does the change?

javierwilson: something like https://stackoverflow.com/a/35128533 ?

javierwilson:
@snopoke i added the option to delete routes too.

ansible/deploy_common.yml
```diff
@@ -32,3 +32,26 @@
...
   roles:
     - {role: monit, tags: monit}
+
+- name: Configure static routes
+  hosts: all
+  tasks:
+
+  - name: Checks existing static routes
+    shell: 'ip route show {{ item.prefix }} | wc -l'
+    with_items: "{{ static_routes }}"
```
emord: This doesn't work if static_routes is undefined.

emord: Should it have a default value? what default value should it be?

javierwilson: perhaps when: static_routes|default(false) ? i'll test it.

snopoke:
Probably just an empty list

PR #1031, opened 2017-11-14
deploy users changes

emord:
@snopoke Found these when deploying users to hqpillow0-staging. It's not clear why iterating over dev_users.present requires a .name now though

### DIFF
--- a/ansible/roles/users/tasks/main.yml
+++ b/ansible/roles/users/tasks/main.yml
@@ -6,12 +6,12 @@
 
 - name: Add users for current devs
   become: yes
-  user: name="{{ item }}" state=present shell=/bin/bash groups={{ dev_group }}
+  user: name="{{ item.name }}" state=present shell=/bin/bash groups={{ dev_group }}
   with_items: '{{ dev_users.present }}'
 
 - name: Add public keys for current devs
   become: yes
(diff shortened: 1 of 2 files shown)


emord:
actually even more confusingly, htis only happened when deploying as root with --tags=users to bootstrap. when running as ansible  master  worked

emord:
Closing this as I don't know why it was failing when I was boostraping

emord:
@nickpell ran into this too. it appears that

`$ ansible-playbook -u root -i ../../commcare-hq/fab/inventory/$ENV deploy_stack.yml -e "@../config/$ENV/$ENV.yml" --limit=$NEW_MACHINE --tags=users --ask-pass`

does not work for some reason

emord:
@snopoke @javierwilson @sravfeyn Just want to give a heads up that we might run into this for the new NIC servers. It appears that dev_users is being interpreted differently between the command above and running `aps`

snopoke:
`-e "@../config/$ENV/$ENV.yml"` looks suspicious. Shouldn't it be `-e "@vars/$ENV/{$ENV}_public.yml" -e "@vars/$ENV/{$ENV}_vault.yml"`

emord:
@snopoke thanks for seeing that. i followed our out of date docs too closely. this should be solved by https://github.com/dimagi/commcarehq-ansible/pull/1052

PR #1032, opened 2017-11-14
Commcare cloud

dannyroberts:
This adds a `commcare-cloud` CLI tool, which currently has one command, `ansible-playbook`. To usage is essentially

```
commcare-cloud <env> ansible-playbook <...>
```

e.g.

```
commcare-cloud staging ansible-playbook deploy_localsettings.yml --tags=localsettings
```

The default behavior is to run with `--check` (`--diff` is always on), tell you whether it succeeded, and then prompt you whether to abort (default) or apply it. You can also run it with `--skip-check` to skip the check portion (in which case it will still ask you to confirm.

Other nice features:
- you only have to enter the vault password once per run and it'll use it for both check and apply.
- if you are not on master, it will fail, and tell you what cli arg to add to make it not fail.
- there's an additional shortcut for what we've been calling "localsettings" but I'm calling config as per https://12factor.net/config: `commcare-cloud staging update-config` is equivalent to `commcare-cloud staging ansible-playbook deploy_localsettings.yml --tags=localsettings`
- ⭐️ it installs itself, so once this is merged the install path is
  - log in
  - `update-code`
  - run `. init-ansible` if you can remember that, or else just log out and log in
  ...and you're ready to go.

### DIFF
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -1,6 +1,7 @@
 # coding=utf-8
 from __future__ import print_function
 import getpass
+import os
 from six.moves import input
 from argparse import ArgumentParser
 import subprocess
@@ -21,15 +22,15 @@ class AnsiblePlaybook(object):
         def anisible_playbook(environment, vault_password, *cmd_args):
             cmd = (
                 'ansible-playbook',
-                '-u', '~/.commcare-cloud/ansible',
-                '-i', '~/.commcare-cloud/inventory/{env}'.format(env=environment),
-                '-e', '"@~/.commcare-cloud/vars/{env}/{env}_vault.yml"'.format(env=environment),
-                '-e', '"@~/.commcare-cloud/vars/{env}/{env}_public.yml"'.format(env=environment),
-                '--ask-vault-pass',
+                '-u', os.path.expanduser('~/.commcare-cloud/ansible'),
+                '-i', os.path.expanduser('~/.commcare-cloud/inventory/{env}'.format(env=environment)),
+                '-e', '"@{}"'.format(os.path.expanduser('~/.commcare-cloud/vars/{env}/{env}_vault.yml'.format(env=environment))),
+                '-e', '"@{}'.format(os.path.expanduser('~/.commcare-cloud/vars/{env}/{env}_public.yml"'.format(env=environment))),
+                '--vault-password-file=/bin/cat',
                 '--diff',
             ) + cmd_args
             print(' '.join(cmd))
-            p = subprocess.Popen(cmd, stdin=subprocess.PIPE)
+            p = subprocess.Popen(' '.join(cmd), stdin=subprocess.PIPE, shell=True)
             p.communicate(input='{}\n'.format(vault_password))
             return p.returncode
 
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -79,6 +79,19 @@ class AnsiblePlaybook(object):
         exit(exit_code)
 
 
+class UpdateConfig(object):
+    @staticmethod
+    def make_parser(parser):
+        parser.add_argument('--skip-check', action='store_true', default=False)
+        parser.add_argument('--branch', default='master')
+
+    @staticmethod
+    def run(args, unknown_args):
+        args.playbook = 'deploy_localsettings.yml'
+        unknown_args += ('--tags=localsettings',)
+        AnsiblePlaybook.run(args, unknown_args)
+
+
 def git_branch():
     return subprocess.check_output("git branch | grep '^*' | cut -d' ' -f2", shell=True,
                                    cwd=os.path.expanduser('~/.commcare-cloud/ansible')).strip()
@@ -100,10 +113,13 @@ def main():
     subparsers = parser.add_subparsers(dest='command')
 
     AnsiblePlaybook.make_parser(subparsers.add_parser('ansible-playbook'))
+    UpdateConfig.make_parser(subparsers.add_parser('update-config'))
 
     args, unknown_args = parser.parse_known_args()
     if args.command == 'ansible-playbook':
         AnsiblePlaybook.run(args, unknown_args)
+    if args.command == 'update-config':
+        UpdateConfig.run(args, unknown_args)
 
 if __name__ == '__main__':
     main()
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -23,7 +23,7 @@ class AnsiblePlaybook(object):
     def run(args, unknown_args):
         check_branch(args)
 
-        def anisible_playbook(environment, playbook, vault_password, *cmd_args):
+        def ansible_playbook(environment, playbook, vault_password, *cmd_args):
             cmd = (
                 'ANSIBLE_CONFIG={}'.format(os.path.expanduser('~/.commcare-cloud/ansible/ansible.cfg')),
                 'ansible-playbook',
@@ -41,10 +41,10 @@ class AnsiblePlaybook(object):
             return p.returncode
 
         def run_check():
-            return anisible_playbook(args.environment, args.playbook, get_ansible_vault_password(), '--check', *unknown_args)
+            return ansible_playbook(args.environment, args.playbook, get_ansible_vault_password(), '--check', *unknown_args)
 
         def run_apply():
-            return anisible_playbook(args.environment, args.playbook, get_ansible_vault_password(), *unknown_args)
+            return ansible_playbook(args.environment, args.playbook, get_ansible_vault_password(), *unknown_args)
 
         def get_ansible_vault_password():
             if not get_ansible_vault_password.value:
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -22,7 +22,7 @@ class AnsiblePlaybook(object):
         def anisible_playbook(environment, vault_password, *cmd_args):
             cmd = (
                 'ansible-playbook',
-                '-u', os.path.expanduser('~/.commcare-cloud/ansible'),
+                '-u', 'ansible',
                 '-i', os.path.expanduser('~/.commcare-cloud/inventory/{env}'.format(env=environment)),
                 '-e', '"@{}"'.format(os.path.expanduser('~/.commcare-cloud/vars/{env}/{env}_vault.yml'.format(env=environment))),
                 '-e', '"@{}'.format(os.path.expanduser('~/.commcare-cloud/vars/{env}/{env}_public.yml"'.format(env=environment))),
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -1,0 +1,88 @@
+# coding=utf-8
+from __future__ import print_function
+import getpass
+from six.moves import input
+from argparse import ArgumentParser
+import subprocess
+from clint.textui import puts, colored
+
+
+def ask(message):
+    return 'y' == input('{} [y/N]'.format(message))
+
+
+class AnsiblePlaybook(object):
+    @staticmethod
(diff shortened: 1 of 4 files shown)


commcare-cloud/commcare_cloud.py
```diff
@@ -0,0 +1,125 @@
...
+    def make_parser(parser):
+        parser.add_argument('--skip-check', action='store_true', default=False)
+        parser.add_argument('--branch', default='master')
+        parser.add_argument('playbook')
+
+    @staticmethod
+    def run(args, unknown_args):
+        check_branch(args)
+
+        def anisible_playbook(environment, playbook, vault_password, *cmd_args):
```
emord: ansibile -> ansible

commcare-cloud/commcare_cloud.py
```diff
@@ -0,0 +1,125 @@
...
+                '-i', os.path.expanduser('~/.commcare-cloud/inventory/{env}'.format(env=environment)),
+                '-e', '"@{}"'.format(os.path.expanduser('~/.commcare-cloud/vars/{env}/{env}_vault.yml'.format(env=environment))),
+                '-e', '"@{}'.format(os.path.expanduser('~/.commcare-cloud/vars/{env}/{env}_public.yml"'.format(env=environment))),
+                '--vault-password-file=/bin/cat',
+                '--diff',
+            ) + cmd_args
+            print(' '.join(cmd))
+            p = subprocess.Popen(' '.join(cmd), stdin=subprocess.PIPE, shell=True)
+            p.communicate(input='{}\n'.format(vault_password))
+            return p.returncode
```
snopoke: Any reason not to run the playbook using the python API?

dannyroberts: Yeah, nothing solid, I was just interested in mapping this directly onto the commands we're used to running, and if we use the API I'm less familiar with the behavior, and depending on the level we hook into we wouldn't get the command-line checks on args etc. I felt like I had a lot more certainty about how it would behave if I accessed it through the shell.

commcare-cloud/commcare_cloud.py
```diff
@@ -0,0 +1,125 @@
+# coding=utf-8
```
kaapstorm: `#!/usr/bin/env python`?

dannyroberts: I made this executable through the `entry_points.console_scripts` setting in setup.py here: https://github.com/dimagi/commcarehq-ansible/pull/1032/files#diff-1028db3bff7bbf8d651eee1350b2c8f9R10. So I don't actually need to make this file itself executable.

kaapstorm: Ah! That's cool. I didn't know that was a thing.

commcare-cloud/commcare_cloud.py
```diff
@@ -0,0 +1,125 @@
...
+                os.path.expanduser('~/.commcare-cloud/ansible/{playbook}'.format(playbook=playbook)),
+                '-u', 'ansible',
+                '-i', os.path.expanduser('~/.commcare-cloud/inventory/{env}'.format(env=environment)),
+                '-e', '"@{}"'.format(os.path.expanduser('~/.commcare-cloud/vars/{env}/{env}_vault.yml'.format(env=environment))),
+                '-e', '"@{}'.format(os.path.expanduser('~/.commcare-cloud/vars/{env}/{env}_public.yml"'.format(env=environment))),
+                '--vault-password-file=/bin/cat',
+                '--diff',
+            ) + cmd_args
+            print(' '.join(cmd))
+            p = subprocess.Popen(' '.join(cmd), stdin=subprocess.PIPE, shell=True)
```
millerdev: It would be safer to use `' '.join(shlex_quote(arg) for arg in cmd)` (and `from six.moves import shlex_quote`) instead of `' '.join(cmd)`. As is will break if any of the arguments contain spaces.

commcare-cloud/commcare_cloud.py
```diff
@@ -0,0 +1,125 @@
...
+            return p.returncode
+
+        def run_check():
+            return ansible_playbook(args.environment, args.playbook, get_ansible_vault_password(), '--check', *unknown_args)
+
+        def run_apply():
+            return ansible_playbook(args.environment, args.playbook, get_ansible_vault_password(), *unknown_args)
+
+        def get_ansible_vault_password():
+            if not get_ansible_vault_password.value:
```
millerdev: should say `if get_ansible_vault_password.value is None:`—the password could be an empty string.

commcare-cloud/commcare_cloud.py
```diff
@@ -0,0 +1,125 @@
...
+        exit(-1)
+
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument('environment')
+    subparsers = parser.add_subparsers(dest='command')
+
+    AnsiblePlaybook.make_parser(subparsers.add_parser('ansible-playbook'))
+    UpdateConfig.make_parser(subparsers.add_parser('update-config'))
```
millerdev: What do you think of renaming "update-config" to "deploy-localsettings"? It's not obvious what "config" it's referring to without reading the code in this file.

control/init.sh
```diff
@@ -35,8 +34,15 @@ echo "Downloading dependencies from galaxy"
 export ANSIBLE_ROLES_PATH=~/.ansible/roles
 ansible-galaxy install -r ~/commcarehq-ansible/ansible/requirements.yml
 
+pip install -r ~/commcarehq-ansible/ansible/requirements.txt
```
millerdev: this was purposely left out in the past because it makes the init script take _much_ longer to run, and many people have it in their profile, which can make starting a new shell take an inconvenient amount of time.

control/init.sh
```diff
@@ -35,8 +34,15 @@ echo "Downloading dependencies from galaxy"
...
+pip install -e ~/commcarehq-ansible/commcare-cloud
+
 # convenience: . init-ansible
 [ ! -f ~/init-ansible ] && ln -s ~/commcarehq-ansible/control/init.sh ~/init-ansible
+[ ! -d ~/.commcare-cloud ] && mkdir ~/.commcare-cloud
+[ ! -d ~/.commcare-cloud/ansible ] && ln -s ~/commcarehq-ansible/ansible ~/.commcare-cloud/
+[ ! -d ~/.commcare-cloud/vars ] && ln -s ~/commcarehq-ansible/ansible/vars ~/.commcare-cloud/
+[ ! -d ~/.commcare-cloud/inventory ] && ln -s ~/commcare-hq-deploy/fab/inventory ~/.commcare-cloud/
```
millerdev: Why is this indirection (symlinks in `~/.commcare-cloud/`) necessary? Would it work to remove them and replace `~/.commcare-cloud/` with `~/commcarehq-ansible/ansible/`?

commcare-cloud/commcare_cloud.py
millerdev: Oh, I see you mentioned this in the PR description. I'm not a fan of renaming for consistency with a best practice we're not following: [Store config in the environment](https://12factor.net/config) (aka _environment variables_). Also from that page "Note that this definition of “config” does not include internal application config, such as `config/routes.rb` in Rails"—I think that applies to django localsettings.

dannyroberts: So, I definitely think there's an argument to be made that whether it's industry practice or not, we call these things "localsettings" here at Dimagi, so we shouldn't confuse people by now calling them "config". That said, in my reading of that page, our `localsettings.py` files are absolutely "config"; we just aren't following their recommendation that config be stored in the environment. The negative example they gave of `config/routes.rb` would for us by `urls.py` files—i.e. "configuration" of the framework (rails/django), but an immutable part of your codebase. I'm currently still a fan of my choice not to call this deploy-localsettings (I think it keeps us focussed on the implementation rather than the intention; already one of the "localsettings" files update is formplayer's `applications.properties`). I guess it's inconvenient that I already rolled this out and announced it, but let me know if you feel strongly about pushing back on this name.

control/init.sh
dannyroberts: Yeah, I considered this. It definitely puts a bunch more useless output on the screen (which of course we could clean up), but the total start time from hitting enter on your `ssh` command to receiving a usable prompt is something like 5 seconds in steady state. (I just timed it.) Of course, you're right that it will be longer than that if you have an install you have to do. That's rarely the case, but when it _is_ the case, my guess is that auto-updating will save you more time and energy debugging your system after a minor upgrade than you'll lose waiting for it.

My primary goal here is to make the context from which we run cloud maintenance commands as self-managing as possible so that we don't all have to be experts in the nitty-gritty details of our setup.

What do you think?

control/init.sh
dannyroberts: So, I'm not sure it's "necessary", but I decided to do it because I think it gives you something nice. In terms of why it's nice to have `~/.commcare-cloud/` at all, it makes configuring `commcare-cloud` to run on a different environment a little easier, so for example if I run it on my own machine (as I was doing) I can symlink the correct locations there and it works because `~/commcarehq-ansible/ansible/vars` for example isn't hard-coded. But maybe that wasn't your question; maybe you were asking more specifically about 3 symlinks versus one.

`ansible` and `vars` are symlinks to `~/commcarehq-ansible/ansible` and a subdirectory with the same name (respectively), but `inventory` is symlinked to a different location entirely. Could I have forgone `vars` entirely and referenced it in my code as `~/.commcare-cloud/ansible/vars`? Well, yes I could have in this case. However, `vars` and `inventory` stand out to me in that they are the only parts of the code that deal with _our_ cloud environment, as opposed to _a_ commcare cloud environment. You could imagine a world for example where those two directories live together in a private repository as opposed to living apart in two different public repos that are made up of otherwise generic code.

Not sure if that gets at what you were asking, happy to talk more!

emord: This seemed weird to me las tnight, but  I didn't have a different solution. What about using environment variables?

commcare-cloud/commcare_cloud.py
dannyroberts: Thanks good call: https://github.com/dimagi/commcarehq-ansible/pull/1038

commcare-cloud/commcare_cloud.py
dannyroberts: Thanks, also in https://github.com/dimagi/commcarehq-ansible/pull/1038

commcare-cloud/commcare_cloud.py
millerdev: I see your point about `routes.rb` being more like `urls.py`. I still think it's unnecessary to rename it since it requires more context to understand what it's actually doing, but it might just be opinion/personal preference.

control/init.sh
millerdev: My concern is that sometimes the control machine is also used for other things (db0 on enikshay), and I don't necessarily want to wait for something I'm not even doing when I log into that machine. I guess I will just remove the init script from my profile and run it manually when I want to use ansible.

Another idea would be to remove the init script entirely and do those things as part of `commcare-cloud`.

control/init.sh
millerdev: I'd prefer to have these be options that can be passed to `commcare-cloud`. Something like `--ansible-dir`, `--ansible-vars-dir`, and `--inventory-dir` with values set here as defaults. It could use environment variables following the pattern of [some of our nose plugins](https://github.com/dimagi/commcare-hq/blob/64fabc2799cf6a2f7c49e2db8929c77eb4f86d3b/corehq/tests/noseplugins/dividedwerun.py#L35).

Having this setup/configuration in the init script means there is more scattered code to read to understand what the tool is actually doing. It feels OK for this tool to be opinionated about some of these things (at least by default) since it is specifically made to work with the commcare ansible setup—it even has "commcare" in its name.

control/init.sh
dannyroberts: Yeah, I see your point. To make it at least a little faster, I opened https://github.com/dimagi/commcarehq-ansible/pull/1050, but agree this is ripe for rethinking soon

PR #1036, opened 2017-11-15
cache per user

snopoke:


ansible/ansible.cfg
```diff
@@ -5,5 +5,5 @@ log_path=/var/log/ansible.log
 
 gathering = smart
 fact_caching = jsonfile
-fact_caching_connection = /tmp/ansiblecache
+fact_caching_connection = ~/.ansiblecache
```
kaapstorm: I'm sure using `~` here is fine, but it's useful to know when it won't work. Most shells (any shell we might run ansible inside) will expand `~` to a home directory if `~` is the start of a string. So `~foo` is good, but `FOO_HOME=~foo` is not. See the [Bash reference manual](http://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html) for more info.

`$HOME` will always work.

PR #1040, opened 2017-11-15
add commcarehq-ansible/config if it doesn't exist

dannyroberts:
so that the init script doesn't error on new machines (even if it is
harmless, it's annoying)

A better long term fix is to move the few remaining things from config/
into the ansible vault files

### DIFF
--- a/control/update_code.sh
+++ b/control/update_code.sh
@@ -32,6 +32,7 @@ function update_repo() {
     git submodule update --init --recursive
 }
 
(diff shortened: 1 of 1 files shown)


dannyroberts:
@javierwilson do you agree about moving things from config/ to vault files? I think they're just there because we added them before vault supported files.

javierwilson:
yes, absolutely.
I already moved certs to the vault for pna.

PR #1043, opened 2017-11-16
adds simple bash completion for commcare-cloud

javierwilson:


control/init.sh
```diff
@@ -49,6 +49,9 @@ alias aps='ap deploy_stack.yml'
 alias update-code='~/commcarehq-ansible/control/update_code.sh && . ~/init-ansible'
 alias update_code='~/commcarehq-ansible/control/update_code.sh && . ~/init-ansible'
 
+[ ! -f ~/.bash_completion ] source  ~/commcarehq-ansible/control/.bash_completion
+cp ~/commcarehq-ansible/control/.bash_completion ~/
```
dannyroberts: Shouldn't these two lines be sort of switched? I.e. you want to create the directory of it doesn't exist and wenn run `~/.bash_completion` regardless? Or does `~/.bash_completion` already get automatically run?

javierwilson: exactly, bash automatically runs ~/.bash_complettion, so this is just to update it (and run if it was not there already)

PR #1048, opened 2017-11-16
Make enikshay pg work_mem more reasonable (and update default)

millerdev:
The current very low `work_mem` setting on enikshay is a likely cause of lots of disk access and inefficient query plans.

Numerous sources recommend at least 16MB, including http://pgtune.leopard.in.ua/ (when combined with other parameters from the enikshay server). Some sources say if you can't set `work_mem` to 16MB or more you need more RAM.

Note: I'm offline all next week, so unless you want to oversee this rollout, let's wait to merge until I'm back. Although feel free to roll it out before I'm back if you want to.

@snopoke @javierwilson

### DIFF
--- a/ansible/roles/postgresql/defaults/main.yml
+++ b/ansible/roles/postgresql/defaults/main.yml
@@ -19,7 +19,7 @@ pgbouncer_ini: "{{ pgbouncer_config_home }}/pgbouncer.ini"
 pgbouncer_users: "{{ pgbouncer_config_home }}/userlist.txt"
 
 postgresql_max_connections: 20
-postgresql_work_mem: '1MB'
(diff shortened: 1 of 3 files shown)


emord:
This might want to get moved to commcare-hq-deploy when we have multiple machines (terrible that we have two repos for configuration :/ ) https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/inventory/icds#L32

millerdev:
> This might want to get moved to commcare-hq-deploy...

Oh my, I didn't even realize we were doing that. Thanks @emord

millerdev:
Also turning temp file logging back on so we can tell if this makes a difference there. FYI @calellowitz

PR #1052, opened 2017-11-18
Easy bootstrap

dannyroberts:
hackathon

This is just a start and I got stuck with some permissions errors towards the end of running

```
commcare-cloud hq-3ua7yab bootstrap-users --branch=easy-bootstrap --skip-check
```

when I come back to this I'll start from scratch with a new env and continue debugging.

This PR also includes some good general improvements to `commcare-cloud`


snopoke:
wow, pretty sweet

commcare-cloud/commcare_cloud.py
```diff
@@ -44,30 +51,50 @@ def make_parser(parser):
...
+                cmd_args += ('-u', 'ansible')
+
+            if ask_vault_pass:
+                cmd_parts += ('--vault-password-file=/bin/cat',)
+
+            common_ssh_args = []
+            if pem:
+                common_ssh_args.extend(['-i', pem])
+            if not strict_host_key_checking:
+                common_ssh_args.append('-o StrictHostKeyChecking=no')
```
emord: This is what we get when a provider cycles IPs right? Do we want this around long term? We should probably start up a repository for the hashes we can trust

dannyroberts: I added this as a workaround for easy bootstraping on AWS for temporary environments. I agree that even better would be proactively fetch and whitelist the fingerprints, but I wasn't finding an easy way to do that on my first round of quick googling, so I went with this. Agree that it'd be nice to have a better long-term way to manage this for our prod VMs.

emord: Yeah. that makes sense. I was thinking about htis when having to deal with the new icds cluster. Maybe there's a way we could deploy ~/.known_hosts with a static file? I don't think host keys are private (you get prompted with the host key before authentication), so there shouldn't be any privacy issues with that

PR #1057, opened 2017-11-22
filter dbs that run

czue:


czue:
confirmed this worked

PR #1058, opened 2017-11-22
icds new

snopoke:
@dimagi/scale-team


ansible/vars/icds/icds_public.yml
```diff
@@ -3,19 +3,19 @@ CAS_SITE_HOST: 'cas.commcarehq.org'
 NO_WWW_SITE_HOST: 'icds-cas.gov.in'
 deploy_env: icds
 internal_domain_name: 'internal-icds.commcarehq.org'
-tableau_server: 10.247.24.11:80
+# is tableau necessary? tableau_server: 10.247.24.11:80
```
emord: I think that tableau isn't being used in the new cluster and the users will be using the new dasbhoard built by sol ddev

ansible/vars/icds/icds_public.yml
```diff
@@ -300,7 +296,7 @@ localsettings:
   DROPBOX_APP_NAME: 'CommCareHQFiles'
   DROPBOX_KEY: "{{ localsettings_private.DROPBOX_KEY }}"
   DROPBOX_SECRET: "{{ localsettings_private.DROPBOX_SECRET }}"
-  ELASTICSEARCH_HOST: "{{ groups.elasticsearch.0 }}"
+  ELASTICSEARCH_HOST: "{{ groups.elasticsearch0.0 }}"
```
emord: why did this change?

snopoke:
I changed these to refer to specific hosts instead of "the first item in
the group's list of hosts". It means we aren't depending on the order and
also means you don't have to look at the inventory file to know which host
it is.

But now that I look at it again it would be `groups.es0.0`

I'll fix that tomorrow.

ansible/vars/icds/icds_public.yml
snopoke: going to keep it around for now just in case we do end up needing it

ansible/vars/icds/icds_public.yml
```diff
@@ -2,20 +2,21 @@ SITE_HOST: 'www.icds-cas.gov.in'
 CAS_SITE_HOST: 'cas.commcarehq.org'
 NO_WWW_SITE_HOST: 'icds-cas.gov.in'
 deploy_env: icds
+env_name: icds-new
 internal_domain_name: 'internal-icds.commcarehq.org'
-tableau_server: 10.247.24.11:80
+# is tableau necessary? tableau_server: 10.247.24.11:80
```
sravfeyn: No

ansible/roles/datadog/templates/datadog.conf.j2
```diff
@@ -8,7 +8,7 @@ dd_url: https://app.datadoghq.com
 api_key: {{ DATADOG_API_KEY }}
 tags: {% for group, group_hosts in groups.items() -%}
     {%- if inventory_hostname in group_hosts %}group:{{ group }}, {% endif -%}
-{%- endfor %} environment:{{ deploy_env }}
+{%- endfor %} environment:{{ env_name|default(deploy_env) }}
```
sravfeyn: Aren't we creating separate datadog environment for 100k?

snopoke: eventually we want the environment to be the same so that the datatog metrics etc keep working. The new name is just temporary until we migrate

sravfeyn: Won't some of metrics get messed up?

snopoke: I don't think so

PR #1059, opened 2017-11-22
Apache no longer serves kafka 0.8.2.2 here

taylordowns2000:
0.10 is the oldest available on apache.mirror.anlx.net. Switching to the archive site works for me. Alternatives to my proposed change include:
- changing the Kafka version
- setting the mirror above to `archive.apache.org` and adding the `/dist` to https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/roles/kafka/tasks/main.yml#L16

I went with this single change because it was easier to PR from github—no offence taken if you implement one of the other methods above!


snopoke:
Thanks @taylordowns2000 upgrading to a newer version is on our backlog

PR #1060, opened 2017-11-22
move icds certs into ansible vars file

snopoke:
I've checked this on ICDS:
```
ENV=icds && ap deploy_proxy.yml --diff --tags=update-cert --limit=10.247.24.13 --check
```
No diff output.


snopoke:
Yea. I just kept it around in case we do need it.

PR #1070, opened 2017-11-27
Remove unused MOTECH servers

sheelio:
Once this is deployed, I can actually pull down the servers from Rackspace.

@calellowitz @mkangia  (since you're on interrupt).


mkangia:
@sheelio Dont have much context here. Are these just domain specific forwarders not in use anymore?

sheelio:
Yup, old servers that are not needed.

nickpell:
Ran into an issue deploying production today because these changes hadn't been ansible deployed.  Would be good to make sure that ansible changes happen on the necessary environments after merging.

PR #1071, opened 2017-11-27
Fix find expression in log rotation script

millerdev:
Was missing .csv.gz and .log files, which means old log files were not being deleted at all because all but the most recent 3 end with .gz

This is too complicated...

@emord cc @javierwilson @snopoke @calellowitz

### DIFF
--- a/ansible/roles/postgresql/templates/rotate_logs.sh.j2
+++ b/ansible/roles/postgresql/templates/rotate_logs.sh.j2
@@ -1,6 +1,9 @@
 #!/bin/bash
 # keep latest {{ postgresql_num_logs_to_keep }} log files
-find {{ postgresql_log_directory }} -type f -name '*.csv' -printf '%T@\t%p\n' \
+find {{ postgresql_log_directory }} \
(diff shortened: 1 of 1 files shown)


ansible/roles/postgresql/templates/rotate_logs.sh.j2
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 # keep latest {{ postgresql_num_logs_to_keep }} log files
-find {{ postgresql_log_directory }} -type f -name '*.csv' -printf '%T@\t%p\n' \
+find {{ postgresql_log_directory }} \
+    -type f \
+    \( -name '*.csv' -o -name '*.csv.gz' -o -name '*.log' \) \
```
emord: nit: This is getting pretty nasty. maybe we should break these up, or use a for loop for them

millerdev: I'm not sure how we'd do this with a for loop. Feel free to improve it.

PR #1072, opened 2017-11-27
Fix shared directory variables.

emord:
@calellowitz @snopoke 

Looks like this is because we specify the vars file for share directory. before ansible 2.1.1.0 it also loaded group vars from the current working directory which was a "bug" (disagreement here: https://github.com/ansible/ansible/issues/16878)

### DIFF
--- a/ansible/roles/shared_dir.yml
+++ b/ansible/roles/shared_dir.yml
@@ -4,6 +4,7 @@
   hosts: shared_dir_host  # Should be one
   become: yes
   vars_files:
+    - ../group_vars/all.yml
     - shared_dir/vars/main.yml
   tasks:
     - include: shared_dir/tasks/install.yml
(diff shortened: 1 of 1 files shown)


ansible/roles/shared_dir.yml
```diff
@@ -4,6 +4,7 @@
   hosts: shared_dir_host  # Should be one
   become: yes
   vars_files:
+    - ../group_vars/all.yml
     - shared_dir/vars/main.yml
```
emord: Possible we could just move these into group_vars? I mostly just did this as mvp to fix the error. not sure about best practices.

snopoke: or we could just get rid of the shared dir completely, I did a quick look and it's not used in many places and almost everywhere it is used it's optional

snopoke: Merging this for now and adding it to the tech dept trello

PR #1073, opened 2017-11-28
set pass

sravfeyn:


emord:
Was this supposed to be fixed here?  https://github.com/dimagi/commcarehq-ansible/blob/master/ansible/vars/icds-new/icds-new_public.yml#L249 

This fix is in the icds, not icds-new

PR #1080, opened 2017-11-29
removing icds-new env name

calellowitz:
This is only used in localsettings, specifically the `SERVER_ENVIRONMENT` and `EMAIL_SUBJECT_PREFIX` variables that was causing trouble for dimagi/commcare-hq#18705
Now the environment on the new cluster will also be icds, which fits with the rest of the work we have done and makes an easier replacement for the old cluster.

### DIFF
--- a/ansible/vars/icds/icds_public.yml
+++ b/ansible/vars/icds/icds_public.yml
@@ -2,6 +2,7 @@ SITE_HOST: 'www.icds-cas.gov.in'
 CAS_SITE_HOST: 'cas.commcarehq.org'
 NO_WWW_SITE_HOST: 'icds-cas.gov.in'
 deploy_env: icds
+env_name: icds-new
 internal_domain_name: 'internal-icds.commcarehq.org'
 tableau_server: 10.247.24.11:80
 daily_deploy_email: tech-announce-daily@dimagi.com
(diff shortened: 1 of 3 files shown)


snopoke:
The reason I put that there in the first place was for datadog so that we don't have all the alerts etc mixed up with current datadog stuff. If we remove this then we should add a 'datadog-env' var or something to allow us to keep them separate until after the migration.

calellowitz:
I switched localsettings to use the deploy env (so datadog can keep the custom name), it seems like the localsettings environment should match the deploy env used throughout the rest of the playbooks anyway, and already does in all the others since `env_name` is not defined anywhere else.
Also reran grep and confirmed that env_name is used only in the commcarehq playbook and datadog one.
Let me know if this seems like a reasonable middle ground.

ansible/vars/icds/icds_public.yml
```diff
@@ -2,6 +2,7 @@ SITE_HOST: 'www.icds-cas.gov.in'
 CAS_SITE_HOST: 'cas.commcarehq.org'
 NO_WWW_SITE_HOST: 'icds-cas.gov.in'
 deploy_env: icds
+env_name: icds-new
```
snopoke: This is in the wrong vars file. Should be in `icds-new_public.yml`

snopoke:
@calellowitz bump

calellowitz:
@snopoke 
fixed

PR #1081, opened 2017-11-29
pass through extra args

snopoke:
buddy @biyeun


dannyroberts:
Oh wait never mind, this is a different thing. I fixed my confusion yesterday by putting quotes around my multi-word command. This is about passing in other --flags. Anyway, still approve.

PR #1083, opened 2017-11-29
playbooks for setting up PG standby nodes and for failover

snopoke:
buddy @biyeun 

Taken from https://opensource.com/article/17/6/ansible-postgresql-operations

### DIFF
--- a/ansible/roles/postgresql/tasks/main.yml
+++ b/ansible/roles/postgresql/tasks/main.yml
@@ -89,15 +89,6 @@
   tags: postgresql_conf
   notify: Reload postgres
 
-- name: PostgreSQL recovery configuration
-  template:
-    src: "recovery.conf.j2"
-    dest: "{{ postgresql_home }}/recovery.conf"
-    owner: "postgres"
-    group: "postgres"
-  notify: Restart postgres
-  when: is_pg_standby
-
 - name: start postgresql
   become: true
   service: name=postgresql state=started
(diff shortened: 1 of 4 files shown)


ansible/roles/postgresql/tasks/main.yml
```diff
@@ -89,15 +89,6 @@
...
 
-- name: PostgreSQL recovery configuration
-  template:
-    src: "recovery.conf.j2"
-    dest: "{{ postgresql_home }}/recovery.conf"
-    owner: "postgres"
-    group: "postgres"
-  notify: Restart postgres
-  when: is_pg_standby
-
```
snopoke: removed this from there since it prevents PG setup. Once the file is created it doesn't need to be managed by ansible I don't think

snopoke:
bump

ansible/setup_pg_standby.yml
```diff
@@ -0,0 +1,73 @@
...
+- hosts:
+    - "{{ standby }}"
+  become: yes
+  vars_files:
+    - roles/postgresql/defaults/main.yml
+  tasks:
+    - assert:
+        that:
+          - "hot_standby_master is defined"
+        msg: "Standby node ('{{ standby }}') provided does not have 'host_standby_master var set. Are you sure it's the right node?"
```
calellowitz: there is a typo here, it says `host_standby_master`

PR #1088, opened 2017-11-29
15 forks

dannyroberts:
Closes https://github.com/dimagi/commcarehq-ansible/issues/1086

### DIFF
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -84,6 +84,9 @@ class AnsiblePlaybook(object):
             if '-u' not in unknown_args:
                 cmd_parts += ('-u', 'ansible')
 
+            if '-f' not in unknown_args and '--forks' not in unknown_args:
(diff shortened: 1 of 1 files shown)


commcare-cloud/commcare_cloud.py
```diff
@@ -47,6 +48,26 @@ def get_common_ssh_args(public_vars):
...
+    check whether a conceptual arg is present in a list of command line tokens
+
+    :param unknown_args: list of command line tokens
+    :param short_form: dash followed by single letter, e.g. '-f'
+    :param long_form: double dash followed by work, e.g. '--forks'
+    :return: boolean
+    """
+
+    assert re.match(r'^-[a-zA-Z0-9]$', short_form)
+    assert re.match(r'^--\w+$', long_form)
```
millerdev: Not sure how `unknown_args` has already been parsed by this point, but will this catch arguments of the form `--forks=15`?

Also, for the previous line, single-character args can be stacked, so you might have `-DCu`

PR #1090, opened 2017-11-29
control monit state

javierwilson:
this will work if run with deploy_common.  it may not work if run with deploy_stack since later the cuch role will reload monit, and reload starts a service if stopped, or so i've read.


javierwilson:
In that case we could just -m service -a "name=monit state=stopped"
unless there are more things we need to worry about stopping during the migration.

PR #1091, opened 2017-11-30
keeping wal segs to allow plain backup from master

javierwilson:


ansible/vars/enikshay/enikshay_public.yml
```diff
@@ -77,9 +77,10 @@ keystore_file: '../config/DimagiKeyStore'
 encrypted_root: '/opt/data'
 
 backup_blobdb: False
-backup_postgres: pgbackrest
+backup_postgres: plain
 postgresql_backup_days: 2
 postgresql_backup_weeks: 2
+postgresql_backup_master: True
```
calellowitz: why are we backing up directly from the master on enikshay

javierwilson: i first decided to do this in order to avoid missing wal issues we got in icds. can the slave keep them as well? i can test to make sure we don't *have* to use master.

ansible/roles/postgresql/templates/backup/plain/create_postgres_dump.sh.j2
```diff
@@ -1,12 +1,20 @@
 #!/bin/bash
 BACKUP_TYPE=$1
 DAYS_TO_RETAIN_BACKUPS=$2
+ENV="{{ deploy_env }}"
 
 TODAY=$(date +"%Y_%m_%d")
-DIRECTORY="{{ postgresql_backup_dir }}/postgres_${BACKUP_TYPE}_${TODAY}"
+DIRECTORY="{{ postgresql_backup_dir }}/postgres_${ENV}_${BACKUP_TYPE}_${TODAY}"
+
+{% if postgresql_backup_master|default(false) and hot_standby_master|default(false) %}
```
snopoke: Shouldn't this be `if not postgresql_backup_master`

ansible/vars/enikshay/enikshay_public.yml
snopoke: Correct me if I'm wrong but since we're using `-Xs` as an option for pg_basebackup PostgreSQL will keep as many segments as are required for the backup so we shouldn't need to have a high value for `wal_keep_segments`

javierwilson: -Xs doens't allow compression, also before 9.6 doesn't guarantee keeping all necessary wals.  WIthout compression we couldn't even afford one backup.  So we use -Xf

ansible/roles/postgresql/templates/backup/plain/create_postgres_dump.sh.j2
javierwilson: probable a bad var name, I meant backup_from_master

snopoke: The name is fine and that's what I assumed it meant but if we're backing up from the master why do we need to supply the host option (`-h`)?

javierwilson: we do, line 11, pg_basebackup -h $MASTER -Ft -D "${DIRECTORY}" -Z9 -Xf

snopoke: offlined, I didn't realize the script was running on the standby node

ansible/vars/enikshay/enikshay_public.yml
calellowitz: I think the standby can keep them as well. I guess its worth testing but we do it now from the standby so i don't see why they couldnt. And if it could, then we would be able to back up locally and not worry about load on the master.

javierwilson: yeah, only we don't have much space in the standby, but neither do we in the master so... ok, done.

ansible/roles/postgresql/tasks/backup_plain.yml
```diff
@@ -1,4 +1,14 @@
 ---
+- name: Copy replication credentials
+  become: yes
+  template:
+    src: "backup/plain/pgpass.j2"
+    dest: "{{ postgresql_dir_original_path }}/.pgpass"
```
dannyroberts: @javierwilson 
I tried again and it turns out the issue I mentioned in https://github.com/dimagi/commcarehq-ansible/pull/1025/files#r154541574 actually happens here, not there. When it renders the template `backup/plain/pgpass.j2` (below in this PR) it fails with the message "AnsibleUndefinedVariable: 'hot_standby_master' is undefined".

I'm testing off (a very-slightly modified version of) the [dev](https://github.com/dimagi/commcarehq-ansible/tree/master/ansible/vars/dev) variables, with nothing in the `[pg_standby]` group and no `hot_standby_master` defined anywhere, very much like [staging](https://github.com/dimagi/commcare-hq-deploy/blob/master/fab/inventory/staging#L51). Was this change maybe made assuming that this would only ever be run on a pg_standby machine, with `hot_standby_master` always defined?

dannyroberts: Was able to fix this by just adding `when: is_pg_standby` here. Sound good if I PR that?

javierwilson: yes, i think so... that file should only go on backup server (standby) i'll apply the same fix for https://github.com/dimagi/commcarehq-ansible/pull/1104

PR #1094, opened 2017-11-30
Run shell command sudo

dannyroberts:
Make `commcare-cloud <env> run-shell-command ...` work with just `--become` or `--become-user` flags to do something as another user (i.e. sudo). If (and only if) a password is needed, based on the become user, it will ask you for the vault password. Example usage:

```
commcare-cloud <env> run-shell-command <group> '<command>' --become
```

If you were already using `commcare-cloud <env> run-shell-command ... --become --ask-become-pass` the main differences here are

1. You can drop the `--ask-become-pass` and
2. You will have to use the _vault_ password instead of the ansible user password
3. In an edge case (maybe `dev` environment currenty?) where the ansible user password is not included in the vault/private file, you can still use `--ask-become-pass` to use the ansible user password.

In addition, there is some syntax onboarding if you make the mistake of using `sudo` in your command directly:

screen shot 2017-12-04 at 10 58 07 am

Closes https://github.com/dimagi/commcarehq-ansible/issues/1099


dannyroberts:
Oh, dear. I had missed that as an option entirely. It does sound simpler. It's _kind of_ nice that with this PR it asks you for the same vault password you would use with ansible-playbook, but I probably wouldn't have written this if I'd thought about `-bK`. OK, well I guess I'll close this then!

dannyroberts:
What if we just automatically include `--ask-become-pass` when you use `--become` or `--become-user root` (but not `--become-user cchq`)? I know it seems small, but I'm trying to pack as much herd knowledge as possible into this tool so no one has to think about these things.

In which case, might as well just keep this PR and stick with requiring the same password as `ansible-playbook`, right? Sorry I'm flip flopping on this. What do you think @snopoke

snopoke:
Yea, seems fine.

dannyroberts:
@millerdev also made another suggestion here: https://github.com/dimagi/commcarehq-ansible/issues/1099

dannyroberts:
And I implemented that here: https://github.com/dimagi/commcarehq-ansible/pull/1101

dannyroberts:
Picking this PR back up and taking the main good idea from https://github.com/dimagi/commcarehq-ansible/pull/1101 (warning when you start a command with 'sudo ') and just added it on top of this.

commcare-cloud/commcare_cloud.py
```diff
@@ -247,24 +252,65 @@ def make_parser(parser):
         parser.add_argument('-u', '--user', default='ansible', help=(
             "The user to run the commands as."
         ))
+        parser.add_argument('--become', action='store_true', help=(
```
snopoke: could you add the short form as well: `'-b'`

millerdev: oops, I should have waited to merge until @dannyroberts had a chance to respond to this (probably by adding 4 characters to this PR).

dannyroberts: Ah, yes, of course. Thanks, will do!

dannyroberts: https://github.com/dimagi/commcarehq-ansible/pull/1112

PR #1097, opened 2017-12-01
commcare-cloud: detect remote branch

dannyroberts:
was previously giving `You are not on branch master. To deploy anyway, use --branch=(HEAD`

### DIFF
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -247,23 +247,14 @@ class RunShellCommand(object):
 
 def git_branch():
     cwd = os.path.expanduser('~/.commcare-cloud/ansible')
-
-    def get_detached_ref():
-        """
-        :return None if on a local branch, commit or remote branch name otherwise
-        """
-        return subprocess.check_output(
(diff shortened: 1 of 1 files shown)


commcare-cloud/commcare_cloud.py
```diff
@@ -246,8 +246,24 @@ def run(args, unknown_args):
...
-    return subprocess.check_output("git branch | grep '^*' | cut -d' ' -f2", shell=True,
-                                   cwd=os.path.expanduser('~/.commcare-cloud/ansible')).strip()
+    cwd = os.path.expanduser('~/.commcare-cloud/ansible')
+
+    def get_detached_ref():
+        """
+        :return None if on a local branch, commit or remote branch name otherwise
+        """
+        return subprocess.check_output(
+            "git branch | grep '\* (HEAD detached at .*)' | cut -d' ' -f5 | cut -d')' -f1",
```
millerdev: Interesting using the shell/grep/cut to do regex matching and string manipulation. Python has nice tools in the standard library that do this kind of thing, with (imo) in a more readable syntax. Also, would only need a single subprocess call (`subprocess.check_output("git branch")`) if matching/string manipulation was done in Python.

dannyroberts: OK fine :)

PR #1098, opened 2017-12-01
add restart-elasticsearch command

dannyroberts:
this pushes the prompt/confirmation/user interaction
from es_rolling_restart.yml into commcare-cloud

Closes https://github.com/dimagi/commcarehq-ansible/issues/1093

### DIFF
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -14,6 +14,13 @@ def ask(message):
     return 'y' == input('{} [y/N]'.format(message))
 
 
+def ask_YES(message):
+    r = input('{} [YES/NO]'.format(message))
+    while r not in ('YES', 'NO', 'N', 'n', 'no', ''):
+        r = input('YES or NO? '.format(message))
+    return 'YES' == r
+
+
 def arg_skip_check(parser):
     parser.add_argument('--skip-check', action='store_true', default=False, help=(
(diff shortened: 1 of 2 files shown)


emord:
We should probably change this to restart all the nodes at once. Rolling restarts are only helpful when you have replica shards, so it really only helps on users in some environments (enikshay and monoliths don't hvae replicas)

commcare-cloud/commcare_cloud.py
```diff
@@ -165,37 +172,52 @@ def get_ansible_vault_password():
...
+    help = (
+        "Do a rolling restart of elasticsearch."
+    )
+
+    @staticmethod
+    def run(args, unknown_args):
+        args.playbook = 'es_rolling_restart.yml'
+        if not ask_YES('Have you stopped all the elastic pillows?'):
+            exit(0)
+        puts(colored.yellow("For ES clusters without data redundancy, "
```
calellowitz: I am pretty sure this is all clusters now. I believe even on clusters with some replica shards its only on a couple indices, so this might want to be worded more strongly.

dannyroberts:
@calellowitz @emord thanks for the review. I'm not opposed to changing how we restart elasticsearch, but I will let one of you do that if you choose. (I don't use this much, so I'm really just limiting this change to making it available through `commcare-cloud`.)

PR #1100, opened 2017-12-01
Auto bash completion

dannyroberts:
Bash completion now pulls options list from `commcare-cloud -h`, caching it in an env variable so it only has to calculate those once per login/session. This keeps us from having to independently update this file.

While I was at it I also made the playbook file completion a little less noisy.


dannyroberts:
You mean nicer than parsing the output of `commcare-cloud -h`? Haha, yeah.

PR #1101, opened 2017-12-01
interpret shell command starting with sudo

dannyroberts:
```
commcare-cloud <env> run-shell-command <group> 'sudo <cmd>'
```
is now short for
```
commcare-cloud <env> run-shell-command <group> '<cmd>' --become --ask-become-pass
```

and

```
commcare-cloud <env> run-shell-command <group> 'sudo -u <user> <cmd>'
```
is short for
```
commcare-cloud <env> run-shell-command <group> '<cmd>' --become-user <user> [--ask-become-pass]
```

(`--ask-become-pass` is skipped for user `cchq` because you don't need it).

You have to say 'y' to a confirmation, and will also tell you that you can run the raw command instead by adding `--raw`. (I can't actually think of a time you'd use this, but leaving it as an escape hatch.)

Closes https://github.com/dimagi/commcarehq-ansible/issues/1099

### DIFF
--- a/commcare-cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud.py
@@ -234,13 +234,34 @@ class RunShellCommand(object):
     def run(args, unknown_args):
         public_vars = get_public_vars(args.environment)
         shell_command = args.shell_command
-        become_root = False
+        become = False
+        become_user = None
+        ask_become_pass = False
         if shell_command.strip().startswith('sudo '):
-            puts(colored.blue("Using `--become --ask-become-pass` rather than `sudo`. "))
+            parts = shell_command.strip().split()
+            if len(parts) > 2 and parts[1] == '-u':
+                become_user = parts[2]
+                cmd_parts = parts[3:]
+            else:
+                become_user = None
+                cmd_parts = parts[1:]
+            ask_become_pass = become_user not in ('cchq',)
+            if become_user:
+                if ask_become_pass:
+                    puts(colored.blue(
+                        "Using `--become-user {}` --ask-become-pass rather than `sudo`. "
+                        .format(become_user)))
+                else:
+                    puts(colored.blue(
+                        "Using `--become-user {}` rather than `sudo`. ".format(become_user)))
+            else:
+                puts(colored.blue("Using `--become --ask-become-pass` rather than `sudo`. "))
+
             puts(colored.blue("If you want to run the raw command as written, use --raw."))
-            if ask("Run command as root?"):
-                shell_command = shell_command.strip()[len('sudo '):]
-                become_root = True
+
+            if ask("Run command as {}?".format('root' if not become_user else become_user)):
+                shell_command = ' '.join(cmd_parts)
+                become = True
             else:
                 exit(0)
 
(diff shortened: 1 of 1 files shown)


commcare-cloud/commcare_cloud.py
```diff
@@ -224,20 +224,64 @@ def make_parser(parser):
...
     @staticmethod
     def run(args, unknown_args):
         public_vars = get_public_vars(args.environment)
+        shell_command = args.shell_command
+        become = False
+        become_user = None
+        ask_become_pass = False
+        if shell_command.strip().startswith('sudo '):
+            parts = shell_command.strip().split()
+            if len(parts) > 2 and parts[1] == '-u':
```
millerdev: This is fragile. I sometimes pile options together (like `sudo -iu user`), which would have confusing/wrong results here. Maybe it would be better if it didn't try to be so smart?

dannyroberts:
@snopoke thanks for the thoughts, I think your instinct here that this is maybe too magical/unintuitive to a long-time ansible user is good.

As for the problem I'm trying to solve, I think we touched on this in https://github.com/dimagi/commcarehq-ansible/pull/1094#issuecomment-348496184, but the problem I'm trying to solve is distributing little pieces of knowledge like "stick `-bK` (or `--become --ask-become-pass`) to the end of your `run-shell-command` to run it as root (or just `--become-user cchq` to run as `cchq`)" around the tech team. While I think it's good for some people on the team to be very familiar with ansible, I want to make being very familiar with ansible not a requirement. I think we should look at instances of experienced team members like myself and Daniel (but the same would be true of new team members) not knowing something that seemed obvious ("you can't run `ansible` ad-hoc commands that begin with `sudo` because it'll just hang; you should use `ansible` cli flags instead") as an indication that our tool could be more helpful in distributing such knowledge, or providing shortcuts that don't require us to know it in the first place.

As for this particular solution, I agree with @millerdev's comment that trying to parse the `sudo` command and do the "right" thing is too smart and will only lead to more confusion, so I think I'm going to close this PR.

I'm going to go back to my suggestion here https://github.com/dimagi/commcarehq-ansible/pull/1094#issuecomment-348496184, and work from there. The main thing I'll add is to also print a helpful error message if your command starts with 'sudo ' (like, "To run a command as another user use `--become` (for root) or `--become-user <user>`. Using 'sudo' directly in the command is nonstandard practice and in the normal event that a password is required it will silently hang indefinitely. Do you know what you're doing and want to run this anyway? [y/N]").

snopoke:
Thanks @dannyroberts. I do agree with your thinking here in terms of making the tool self explanatory or self guiding so that it is easy to use by anyone but I think the way it should do that is by guiding users to the standard practice / options rather than trying to interpret the user's intent.

One thing I've been wanting is to be able to show the normal ansible command options. Since we pass those through it would be nice to be able to know what they are. It's still possible to just run `ansible-playbook --help` to get them so I guess we could just add that as a note to the help docs.

dannyroberts:
@snopoke good point, I was just thinking the same thing while tweaking the help text. Moving discussion to https://github.com/dimagi/commcarehq-ansible/issues/1113

PR #1104, opened 2017-12-04
pg_backup group and role

javierwilson:
work in progress. this need further testing and a change in all inventories.


ansible/deploy_db.yml
```diff
@@ -15,12 +15,18 @@
...
 
-- name: Postgresql
+- name: PostgreSQL
   hosts: postgresql,pg_standby
   become: true
   roles:
     - {role: postgresql, tags: 'postgresql'}
 
+- name: PostgreSQL Backup
+  hosts: all
```
snopoke: I think this can be `postgresql:pg_backup`

ansible/roles/pg_backup/defaults/main.yml
```diff
@@ -0,0 +1,18 @@
...
+backup_postgres: plain
+postgres_s3: false
+hot_standby_server: false
+pg_backup_from: false
+pg_backup_to: false
+
+# default to false (must be set to true by group vars)
+is_pg_backup: false
+is_pg_master: false
+is_pg_standby: false
```
snopoke: These shouldn't be necessary should they? You can just check `'pg_standby' in group_names`

ansible/roles/pg_backup/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+  apt_repository: repo='deb http://apt.postgresql.org/pub/repos/apt/ {{ ansible_lsb.codename}}-pgdg main' state=present
+  register: add_postgresql_repo
+  when: is_pg_backup
+
+- name: Add PosgreSQL apt key
+  apt_key: url=https://www.postgresql.org/media/keys/ACCC4CF8.asc state=present
+  when: is_pg_backup
+
+- name: Update package list
+  apt: update_cache=yes
```
snopoke: Add `cache_valid_time=7200` to avoid unnecessary updates

ansible/roles/pg_backup/templates/plain/aws_credentials.j2
```diff
@@ -0,0 +1,6 @@
+
+# {{ ansible_managed }}
+
+[default]
+aws_access_key_id={{ localsettings.AMAZON_S3_ACCESS_KEY }}
+aws_secret_access_key={{ localsettings.AMAZON_S3_SECRET_KEY }}
```
snopoke: Is this (and other similar files) just moving from the `postgresql` role? If so I think you forgot to remove the other files.

ansible/roles/pg_backup/defaults/main.yml
```diff
@@ -0,0 +1,18 @@
...
+postgresql_port: 5432
+postgresql_backup_dir: "{{ encrypted_root }}/backups/postgresql"
+postgresql_dir_original_path: /var/lib/postgresql
+
+nadir_hour: 0
+postgresql_backup_days: 3
+postgresql_backup_weeks: 4
+backup_postgres: plain
+postgres_s3: false
+hot_standby_server: false
```
snopoke: What's the significance of this in this role?

javierwilson: pgbackrest uses this variable a couple of times, this is to replace using |default(false) repeatedly

PR #1107, opened 2017-12-04
Add support for different riak config per env

millerdev:
Makes #631 obsolete (this PR is based on that branch). Use `ansible-playbook deploy_riak_cs_new.yml ...` to deploy changes to the second/new cluster. I ran this on all riak envs except icds-new. Notes:

- staging - benign changes applied, nodes restarted
- production - benign changes applied
- production `deploy_riakcs_new` - benign changes applied, one node restarted to verify changes did not prevent start
- softlayer - changes not applied
  - riak.conf:
    - riak version comment (new with this PR)
    - `erlang.distribution.port_range` (not important for normal operation)
    - hostname -> IP address (don't know why this is happening)
- enikshay - changes not applied
  - data dir permissions
  - riak.conf: riak version comment (new with this PR)
- icds - changes not applied:
  - riak.conf
    - various comments (riak version, etc.)
    - `erlang.distribution.port_range` (not important for normal operation)
    - `ring_size`: config currently says 128 even though the actual ring size is 64 (verified on riak node with `riak-admin status | grep ring`; `ring_num_partitions` and `ring_creation_size` differ, which means the config has been changed and riak restarted since the ring was initialized [[source](http://lists.basho.com/pipermail/riak-users_lists.basho.com/2011-October/006148.html)]; config should be changed back to 64 and riak restarted to resolve this difference, although I don't think it does any harm as is)
    - `leveldb.maximum_memory.percent` commented out (it's overridden anyway, so was ineffective)
  - advanced.config: whitespace only on 3 nodes
- icds-new - not run

@emord @javierwilson @snopoke


ansible/deploy_riakcs_new.yml
```diff
@@ -0,0 +1,197 @@
...
+  roles:
+    - {role: swap}
+
+- name: Load admin keys from Riak CS conf
+  become: true
+  hosts: riakcs_new_control_host
+  vars_files:
+    - roles/riakcs/install/defaults/main.yml
+  tasks:
+    - include: roles/riakcs/get_admin_keys.yml tags=localsettings
```
snopoke: Not neccessary now but could make a custom fact file to get these: http://serverascode.com/2015/01/27/ansible-custom-facts.html

ansible/deploy_riakcs_new.yml
```diff
@@ -0,0 +1,197 @@
...
+    - {role: riakcs/sync_keys, name: riakcs}
+
+- name: Sync admin user keys to Stanchion
+  become: true
+  hosts:
+    - stanchion_new
+  roles:
+    - {role: riakcs/sync_keys, name: stanchion}
+
+- name: Remove Riak Restarter
```
snopoke: Is this still necessary?

ansible/vars/enikshay/enikshay_public.yml
```diff
@@ -28,6 +28,7 @@ couchdb2:
   username: "{{ localsettings_private.COUCH_USERNAME }}"
   password: "{{ localsettings_private.COUCH_PASSWORD }}"
 
+riak_backend: "leveldb-bad-config"
```
snopoke: Well that sucks

ansible/roles/riakcs/configure/templates/riak.conf.j2
```diff
@@ -453,7 +460,12 @@ riak_control.auth.mode = off
 ## 
 ## Acceptable values:
 ##   - an integer
+{% if _riak_backend == "leveldb-bad-config" %}
 leveldb.maximum_memory.percent = 70
+{% else %}
+## Overridden by total_leveldb_mem_percent in advanced.config
+#leveldb.maximum_memory.percent = 70
```
snopoke: does this also apply for if the backend is `bitcask`? I see this setting in the leveldb advanced config with a value of 70 but in the bitcask advanced config the setting has a value of 30. I assume that's correct but just wanted to check.

snopoke:
@millerdev I ran this on `icds-new` with `force_riak_config` and there were only a few benign changes to `riak.conf`

ansible/roles/riakcs/configure/defaults/main.yml
```diff
@@ -1,15 +1,39 @@
...
+riak_data_root_blocks: "{{ riak_data_dir }}/blocks"
+riak_data_dir: "{{ encrypted_root }}/riak"
 riak_ring_dir: "{{ riak_data_dir }}/ring"
-riak_nodename: "riak-{{ inventory_hostname_short }}@{{ inventory_hostname }}"
-riak_ring_size: 64
+
+# This parameter cannot be changed once the cluster has been configured and
+# has data in it. Use a lower ring size (64) in envs where RAM is very limited
+# since higher ring size requires more RAM, even for leveldb-only backend.
+riak_ring_size: 128
```
emord: We generally default to running our sevices on low hardware unless we really have to even if it impacts performance (not saying I agree to this, just mentioning it's a recurring theme)

What's the reason to default to 128 instead of 64? originally I thought there was going to be a point where our cluster could no longer grow, but I think that this config value just impacts the performance. Is that correct?

ansible/vars/enikshay/enikshay_public.yml
millerdev: yeah, disappointed that this cluster got setup that way 😞

ansible/roles/riakcs/configure/templates/riak.conf.j2
millerdev: Yes, my understanding is that if `total_leveldb_mem_percent` is present in advanced.config then this setting in riak.conf is ignored.

ansible/roles/riakcs/configure/defaults/main.yml
millerdev: It's a somewhat grey area. The [Riak documentation](http://docs.basho.com/riak/kv/2.2.3/setup/planning/cluster-capacity/#ring-size-number-of-partitions) recommends ring size >64 for cluster sizes >6 nodes. However, I've gleaned from talking to folks at Basho and discussions on the mailing list that the main concern of too low a ring size may be more about efficient resource usage (using maximal amounts of RAM) on very beefy hosts than it is about absolute ability to run. This of course assumes the cluster will always have less than 64 nodes (probably less than 32 or even 16, practically speaking) since 64 would be an absolute maximum number of nodes for a ring size of 64.

I would prefer to default to ring size of 128 since that allows clusters to grow much bigger before something drastic, like complete migration to a new cluster with a different ring size, needs to be done. As we know, that's a painful process.

As for running our clusters with very lean resources, I think this is really inefficient and a bad trade-off because of the number of man hours required to troubleshoot silly resource issues that should not even be an issue, but that's a discussion for a different place.

millerdev:
> I think this should also be able to close out #708 right?

Yes, although I haven't compared the two very closely so wasn't quite sure if there was anything going on there that I had not accounted for here. However, since @snopoke ran this on icds-new and saw only benign changes then I'm guessing we can close #708.

ansible/deploy_riakcs_new.yml
millerdev: Probably can be removed since I don't think it's been applied everywhere. Leaving for a separate PR.

ansible/roles/riakcs/configure/defaults/main.yml
emord: Thanks for the explanation of ring size. Definitely makes sense to default to 128.

> As for running our clusters with very lean resources, I think this is really inefficient and a bad trade-off because of the number of man hours required to troubleshoot silly resource issues that should not even be an issue, but that's a discussion for a different place.

agreed on all points

millerdev: Incidentally, I've noticed that a ring size of 128 with cluster size of 6 nodes seems to be problematic.  Apparently the default ring balancing algorithm does not like that combination. During recent cluster resizing on prod I had a really hard time getting a plan that didn't include "WARNING: Not all replicas will be on distinct nodes," and there is one node (hqriak20) that has consistently higher CPU usage than the others. After many clear/plan iterations I thought I got (and applied) a plan that didn't have that warning, but later troubleshooting has made me suspect that something is still wrong. Unfortunately the diagnostic tools built into riak are broken on the version of riak we're using, so I'm not certain about any of this.

PR #1110, opened 2017-12-05
 set secure proxy ssl header setting

snopoke:
This allows Django to know whether a request is secure or not
https://docs.djangoproject.com/en/1.11/ref/settings/#secure-proxy-ssl-header

Without this `request.build_absolute_uri()` will use `http` as the request scheme instead of `https`. This is currently breaking dashboard feed URLs : https://manage.dimagi.com/default.asp?266533

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -252,6 +252,7 @@ SEND_BROKEN_LINK_EMAILS = True
 
 OVERRIDE_LOCATION = "https://www.commcarehq.org"
 DEFAULT_PROTOCOL = "https"
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTOCOL', 'https')
 
 ANALYTICS_IDS = {
     'GOOGLE_ANALYTICS_API_ID': '{{ localsettings.GOOGLE_ANALYTICS_API_ID|default('', true) }}',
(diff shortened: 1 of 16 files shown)


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -252,6 +252,7 @@ SEND_BROKEN_LINK_EMAILS = True
 
 OVERRIDE_LOCATION = "https://www.commcarehq.org"
 DEFAULT_PROTOCOL = "https"
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTOCOL', 'https')
```
javierwilson: I think it is HTTP_X_FORWARDED_PROTO

ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -45,6 +45,7 @@ nginx_sites:
    proxy_set_headers:
    - "Host $host"
    - "X-Forwarded-For $remote_addr"
+   - "X-Forwarded-Protocol  $scheme"
```
javierwilson: RFC says it is  X-Forwarded-Proto

ansible/roles/commcarehq/templates/localsettings.py.j2
snopoke: It doesn't matter what it's called as long as it matches up with what we put in nginx

PR #1122, opened 2017-12-06
cleanup and revamp of deploy_proxy

snopoke:
Quite a bit of cleanup and also change to using `include_role` which support more advanced options and also is a dynamic import so decreases the console output dramatically.

I've tested this on ICDS

Best reviewed per commit :tropical_fish: 

Other things to do (separate PR):

- [ ] remove duplication of couchdb2 role (`active_sites.couchbd2` and `couchdb2_proxy` group)

PR #1123, opened 2017-12-06
only install couchdb2 proxy on couchdb2_proxy group

snopoke:
PR into #1122 

Goes with https://github.com/dimagi/commcare-hq-deploy/pull/463


emord:
Nice. For a bit of background I think I did this because I originally had the couchdb localsettings pointing to nginx on icds to see if we could get away with only one NGINX instance while having couchdb2_proxy as a backup in case things went wrong.

This seems like a good way forward and we should probably break our riak proxy config into it's own server as well

PR #1128, opened 2017-12-07
switch prod synclogdb

snopoke:
New synclog DB is up to date so we're ready to make the switch:

image

image


emord:
Wow. that's such a huge difference even though therere are 100k more docs in the new db (I think this is because the new db haven't been pruned?)

looks like the tests have a real failure. (surprised those were good for something)

snopoke:
@emord I'm not sure how long it took. I didn't watch it closely. I'll deploy this next week.

snopoke:
Old DB removed

PR #1132, opened 2017-12-09
Add analytics debug flag and log level

orangejenny:
For https://github.com/dimagi/commcare-hq/pull/18841 and https://github.com/dimagi/commcare-hq/pull/18859 but can be merged/deployed independently.

@calellowitz


snopoke:
Just one check, if those PR's get merged before this is deployed that won't break anything will it?

orangejenny:
Yeah, it's fine if the other PRs get merged first. The settings are only accessed in templates, as part of the ANALYTICS_CONFIG dict, so they're blank if not defined, which is fine.

orangejenny:
Verifying localsettings have been updated on staging and prod.

PR #1134, opened 2017-12-11
pg_dump method for backup

javierwilson:
mostly for very small clusters. needs https://github.com/dimagi/commcarehq-ansible/pull/1104


ansible/roles/pg_backup/tasks/backup_plain_clean.yml
```diff
@@ -14,9 +14,9 @@
   become: true
   cron:
     name: "Backup postgres weekly"
-    job: "/usr/local/sbin/create_postgres_dump.sh weekly 21"
```
calellowitz: looks like this changes the number of backup weeks (the default added is 4, which would make this 28)

javierwilson: true, default should be 3

PR #1137, opened 2017-12-12
monit should be manually started (after-reboot)

javierwilson:
autostarting monit is crazy, since must services need to be started after "after-reboot" is ran (ecrypt mounted)

### DIFF
--- a/ansible/roles/monit/tasks/main.yml
+++ b/ansible/roles/monit/tasks/main.yml
@@ -14,5 +14,7 @@
 
 - name: start monit
   become: yes
-  service: name="monit" state=started enabled=yes
-  tags: monit
+  service: name="monit" state=started enabled=no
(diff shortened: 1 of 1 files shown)


ansible/roles/monit/tasks/main.yml
```diff
@@ -14,5 +14,7 @@
 
 - name: start monit
   become: yes
-  service: name="monit" state=started enabled=yes
-  tags: monit
+  service: name="monit" state=started enabled=no
```
snopoke: why the change from `enabled=yes` to `enabled=no`?

javierwilson: enabled means autostart, if we start monit as soon as ubuntu boots (with no other service started) monit will keep trying and failing to start services

PR #1144, opened 2017-12-14
update set standby with progress

snopoke:
This makes the pg basebackup run as an async task and then runs a loop of other tasks to compare the size of the standby node data with that of the master as a way of estimating progress

### DIFF
--- a/ansible/roles/postgresql/tasks/standby_progress.yml
+++ b/ansible/roles/postgresql/tasks/standby_progress.yml
@@ -1,0 +1,21 @@
+- name: Check sync status
+  async_status:
+    jid: "{{ async_results.ansible_job_id }}"
+  register: async_poll_results
+  when: not async_poll_results.get('finished', True)
+
+- name: get standby size
+  shell: du -s {{ postgresql_home }}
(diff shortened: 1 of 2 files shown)


ansible/roles/postgresql/tasks/standby_progress.yml
```diff
@@ -0,0 +1,21 @@
...
+  when: not async_poll_results.get('finished', True)
+
+- name: get standby size
+  shell: du -s {{ postgresql_home }}
+  register: standby_size_output
+  when: not async_poll_results.get('finished', True)
+
+- debug:
+    msg:
+      - "Progress = {{ (100 * standby_size_output.stdout.split()[0]|int / master_size_output.stdout.split()[0]|int)|int }}%"
```
millerdev: Will integer division result in wrong/misleading output? `99 / 100 == 0`

snopoke: Not in this case because of the expression order: `100 * A / B` (`100 * 99 / 100 == 99`)

PR #1147, opened 2017-12-15
clean up touchforms

javierwilson:
this cleans up touchforms garbage when touchforms is disabled or mvoed.
@kaapstorm should we touchforms_enabled: false for PNA?


snopoke:
Looks good. Re PNA the only reason to have touchforms is if they are doing SMS forms or SMS stock reporting.

kaapstorm:
I think they use SMS stock reporting. I'll check with Kelly.

PR #1149, opened 2017-12-15
set aws credentials for other roles/users

javierwilson:
@kaapstorm PNA backup to AWS still needs this to work


ansible/roles/couchdb/tasks/main.yml
```diff
@@ -105,13 +105,9 @@
 
 - name: Copy couch restart script
```
kaapstorm: "Delete couch restart script", no?

javierwilson: good point.

PR #1160, opened 2017-12-17
change icds-new ES cluster name to icds-2.0

dannyroberts:
prodhqes-1.x is a holdover from the old cluster. The migration docs seemed to pretty clearly suggest we want the ES cluster on the new environment to be named icds-2.0.


snopoke:
Thanks

PR #1163, opened 2017-12-18
run module

snopoke:
Can review by commit :fish:


commcare-cloud/commcare_cloud/commcare_cloud.py
```diff
@@ -323,63 +343,121 @@ def make_parser(parser):
...
-        cmd_parts += get_common_ssh_args(public_vars)
-        cmd = ' '.join(shlex_quote(arg) for arg in cmd_parts)
-        print(cmd)
-        p = subprocess.Popen(cmd, stdin=subprocess.PIPE, shell=True, env=ansible_context.env_vars)
-        if ask_vault_pass:
-            p.communicate(input='{}\n'.format(ansible_context.get_ansible_vault_password()))
-        else:
-            p.communicate()
-        return p.returncode
+        RunAnsibleModule._run(args, 'shell', args.shell_command, unknown_args)
```
dannyroberts: Very nice.

commcare-cloud/commcare_cloud/commcare_cloud.py
```diff
@@ -323,63 +343,121 @@ def make_parser(parser):
...
+    command = 'run-shell-command'
+    help = 'Run an arbitrary command via the shell module.'
+
+    @staticmethod
+    def make_parser(parser):
+        RunAnsibleModule._make_parser(
+            parser,
+            module_args_argument_name='shell_command',
+            module_args_argument_help="The shell command you want to run"
+        )
```
dannyroberts: So the way i solved this same problem in `AnsiblePlaybook` and the command classes that rely on it is to give `AnsiblePlaybook` one hard-coded interface, and then have the command classes that rely on it edit their own `args` and `unknown_args` variables to fit its interface. (For the shared parser args I use some good-enough intermediate base-class inheritance kind of thing, but I don't have a preference between that and something like what you did here with `_make_parser`.)

I like my edit-args-and-pass-on style because the central class needs to anticipate the needs of the commands that rely on it less, and there's less indirection code that needs to be in place to make it work. I'm happy to do this myself as a refactor later, though, if you're happy with that.

commcare-cloud/commcare_cloud/commcare_cloud.py
```diff
@@ -306,6 +325,7 @@ def make_parser(parser):
         parser.add_argument('--become-user', help=(
             "run operations as this user (default=root)"
         ))
+        arg_skip_check(parser)
```
dannyroberts: Does this mean `run-shell-command` now defaults to running in check mode first? If so, I'd suggest modifying it to always pass in `--skip-check`, since the shell module's check mode is useless as far as I know

snopoke: good point

snopoke:
@dannyroberts thanks for the improvements

PR #1165, opened 2017-12-18
pg 9.6

snopoke:
https://docs.google.com/document/d/1FJSKtPerljiIKyxM7a8yOfAo4f3Dx4SDc-1DZ_wY2FM/edit

### DIFF
--- a/ansible/roles/postgresql/templates/postgresql.conf.j2
+++ b/ansible/roles/postgresql/templates/postgresql.conf.j2
@@ -42,7 +42,9 @@ archive_timeout = 60
 
 wal_level = hot_standby
 max_wal_senders = 8
+{% if postgresql_version == '9.4' %}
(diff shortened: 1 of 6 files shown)


ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -31,6 +31,7 @@ work_mem = {{ postgresql_work_mem }}
 max_stack_depth = {{ postgresql_max_stack_depth }}
 maintenance_work_mem = {{ postgresql_maintenance_work_mem }}
 vacuum_cost_delay = 20ms
+dynamic_shared_memory_type = posix
```
snopoke: get's added during the upgrade

snopoke:
@javierwilson I've tested this locally but would be good for someone else to run though it too. Thanks

PR #1170, opened 2017-12-18
Dr/run module

dannyroberts:
Responding to my own feedback in https://github.com/dimagi/commcarehq-ansible/pull/1163. This PR merges into that PR.

This mainly a refactor but
- https://github.com/dimagi/commcarehq-ansible/pull/1170/commits/2389a6af4ab14e237b0852ed116559639bf50547 make run-shell-command always run with --skip-check

is a functional change.

### DIFF
--- a/commcare-cloud/commcare_cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud/commcare_cloud.py
@@ -336,8 +336,6 @@ class RunAnsibleModule(object):
 
     @staticmethod
     def run(args, unknown_args):
-        module = args.module
-        module_args = args.module_args
         ansible_context = AnsibleContext(args)
         public_vars = get_public_vars(args.environment)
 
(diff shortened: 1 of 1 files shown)


commcare-cloud/commcare_cloud/commcare_cloud.py
```diff
@@ -436,17 +417,17 @@ def run_apply():
...
     @staticmethod
     def make_parser(parser):
-        RunAnsibleModule._make_parser(
-            parser,
-            module_args_argument_name='shell_command',
-            module_args_argument_help="The shell command you want to run"
-        )
+        parser.add_argument('inventory_group', help=(
+            "The inventory group to run the command on. Use 'all' for all hosts."
+        ))
```
dannyroberts: Only minor sad part of this refactor is that this statement is copy-pasted between the two. That said, the two commands do have a sort of fundamentally different set of positional arguments, so I think it's clearer to repeat the one argument they have in common and make it explicit right here what this command's positional arguments are than to over parametrize this and make you untangle indirection to figure out what this command's args are.

Having said this, I realized the obvious, that the best of both worlds would be to pull this out into an `arg_inventory_group` helper function. I'll add that to this PR.

PR #1171, opened 2017-12-18
env separation

snopoke:
This separates what the env name is that's used in Datadog and Django from what the env that's used for any folders or files.

So we can have `/home/cchq/www/icds/` and have the env on Datadog be `icds-new`.

This seems like a reasonable separation to me.


snopoke:
@sravfeyn @calellowitz bump

PR #1177, opened 2017-12-19
actually fix tags on Copy sentry.properties so it gets run on update-…

dannyroberts:
…config this time


dannyroberts:
gonna go ahead and merge this without review

PR #1179, opened 2017-12-19
Merge commcare hq deploy

dannyroberts:
Sorry for the long diff here. To review this PR
- Consider your thoughts about merging commcare-hq-deploy into commcarehq-ansible/fab
- Then skip down all the way to the bottom and review the commits starting with https://github.com/dimagi/commcarehq-ansible/pull/1179/commits/09cb79714574556f51dc2c0782c7ebcd4dd48293.
- See whether you think https://github.com/dimagi/commcare-hq/pull/18959 is an okay bridge for easing the change.

When this is merged, people will be able to log into the control machine and (assuming they've already run the 2 steps to set it up) can run

```bash
update-code
```
and then (and every time they log in thereafter) they can run all our fab commands like we always have (without changing directories or anything).

Locally the change is going to be a little less automated but it's basically

```bash
workon ansible
cd commcarehq-ansible
git pull
pip install -r fab/requirements.txt
./control/check_install.sh
cd fab
```

and then you can run the fab command. Thereafter to run fab commands you must `workon ansible` and choose one of three options:
- enter the fab directory:
  ```bash
  cd commcarehq-ansible/fab
  fab production deploy
  ```
- use the `-f` option on fab
  ```bash
  fab -f ~/.commcare-cloud/repo/fab/fabfile.py production deploy
  ```
- put an alias in your bash profile
  ```bash
  echo "alias fab=fab -f ~/.commcare-cloud/fab/fabfile.py" >> ~/.bash_profile
  # forever after, from any directory:
  fab production deploy
  ```
I prefer the third one, which is what's done on control machines, and I suggest that in https://github.com/dimagi/commcare-hq/pull/18959/files


snopoke:
👍 from me. I think this makes a lot of sense. I've been think about making it a submodule for a while but I think this is better

snopoke:
This also breaks ground on getting the services deploy out of HQ deploy. I could also see us breaking the `environments.yml` file up and putting it with the other env vars files.

dannyroberts:
Yes! Exactly. Glad to hear we're on the same page. Those were also two of the top things on my wish-list related to this change.

dannyroberts:
It just occurred to me that we could possibly do this in two steps, where we merge this one first, and then merge the commcare-hq counterpart of this a week or two later. That way we can make sure fab is working in commcarehq-ansible for a number of people before pulling out the rug.

The only thing we'd have to do is to remember to make all changes in the interim to the current commcare-hq-deploy submodule and then I could make sure they get more or less continuously merged into this repo. Not sure it's worth the effort, but just wanted to throw that out there and highlight that merging this would initially be a very quiet change (but would mean committing to a louder change within a couple weeks).

commcare-cloud/commcare_cloud/commcare_cloud.py
```diff
@@ -12,6 +12,12 @@
 
 from .parse_help import filtered_help_message, add_to_help_text
 
+REPO_BASE = os.path.expanduser('~/.commcare-cloud/repo')
+ANSIBLE_DIR = os.path.join(REPO_BASE, 'ansible')
+VARS_DIR = os.path.join(REPO_BASE, 'ansible', 'vars')
+FAB_DIR = os.path.join(REPO_BASE, 'fab')
+INVENTORY_DIR = os.path.join(REPO_BASE, 'fab', 'fab', 'inventory')
```
snopoke: Some repetition from bootstrap. Could maybe pull out into a constants file?

snopoke:
your rollout plan seems good. I doubt much will change in the next few days

snopoke:
just tested this on icds-new and it worked fine:
```
(ansible) skelly@kafka0:~/commcarehq-ansible/fab$ fab icds-new restart_services
```

snopoke:
@dannyroberts FYI there are some commits in commcare-hq-deploy that need to be merged in here.

dannyroberts:
Looked at https://github.com/dimagi/commcare-hq-deploy/commits/master (latest commit on Dec. 21) and it's all been merged in; must have done that last week after you wrote this but before I read it.

millerdev:
I like where this is going. Really excited not to have to merge ansible+inventory changes in two separate repos anymore. However, one thing concerns me:

> put an alias in your bash profile
>
> ```sh
> echo "alias fab=fab -f ~/.commcare-cloud/fab/fabfile.py" >> ~/.bash_profile
> # forever after, from any directory:
> fab production deploy
> ```

I think this means you cannot use fabric (via the `fab` command) for anything other than HQ deploy-related tasks from that specific repo unless you unset that alias. The convention is that `fab` works relative to a `fabfile.py`, and breaking that convention globally is unintuitive and possibly dangerous. If you want a globally accessible command to initiate deploys from anywhere, why not wrap fab deployments into `commcare-cloud`? Although I'm more a fan of making commands that work relative to a local context; most bash commands/programs are like that: `ls`, `cd`, `git`, `fab`, `npm` (without `-g`), etc. This is a powerful feature (more below).

True story: for a long time I've had two separate copies of the HQ repo on my machine, which up until now included two separate copies of the `commcare-hq-deploy` submodule. One is (typically) used for day-to-day development and the other is used for deployment related tasks. This allows me to maintain the state of my dev environment (current branch, uncommitted files, etc.) while also being able to deploy HQ without messing with that state (20+ minute deploys have a way of encouraging this type of innovation). However, I could still run fab from either repo to do deployment-related tasks as necessary. An alias like this would make that workflow unintuitively **use the wrong fabfile!!** when running `fab` in the repo that is not referenced by the alias.

dannyroberts:
Yeah, @millerdev this is a really good point. I didn't love this workaround either, but it was the first way I thought of that made for very convenient usage. Having revisited it, I was able to pretty easily do this a different and, I think you'll agree, better way by dropping a dummy fabfile.py in the root directory, basically as we did before: https://github.com/dimagi/commcarehq-ansible/pull/1212. Once that's merged, I'll also update the commcare-hq documentating fab output.

dannyroberts:
Here it is: https://github.com/dimagi/commcare-hq/pull/19006

PR #1182, opened 2017-12-20
enable tableu on icds-new

sravfeyn:


ansible/roles/postgresql/templates/pg_hba.conf.j2
```diff
@@ -21,6 +21,10 @@ host    replication     {{ postgres_users.replication.username }}   {{ cluster_i
 host    all             {{ postgres_users.backup.username }}   {{ cluster_ip_range|default('0.0.0.0/0') }}    md5
 {% endif %}
 
+{% if postgres_users.get('tableau_server') %}
+host    all     {{ postgres_users.tableau_readonly.username }}   {{ tableau_server.split(':')[0] }}    md5
```
dannyroberts: I think you need a /32 in there, like
```
host    all     {{ postgres_users.tableau_readonly.username }}   {{ tableau_server.split(':')[0] }}/32    md5
```

PR #1186, opened 2017-12-21
old es0 as squid

sravfeyn:
@javierwilson

### DIFF
--- a/ansible/vars/icds-new/icds-new_public.yml
+++ b/ansible/vars/icds-new/icds-new_public.yml
@@ -193,7 +193,7 @@ nameservers:
   - 164.100.17.3
   - 164.100.3.1
 
-http_proxy_address: '10.247.63.132'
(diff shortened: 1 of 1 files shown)


ansible/vars/icds-new/icds-new_public.yml
```diff
@@ -193,7 +193,7 @@ nameservers:
   - 164.100.17.3
   - 164.100.3.1
 
-http_proxy_address: '10.247.63.132'
+http_proxy_address: '10.247.24.16'
```
dannyroberts: this could also eventually just be 
```
http_proxy_address: {{ groups.squid.0 }}
```
if we start setting that group correctly, or even just
```
http_proxy_address: {{ groups.patch0 }}
```
right?

javierwilson: correct

PR #1188, opened 2017-12-21
auto upgrade pip

dannyroberts:
I've had reports of people getting tripped up on having an old version of pip on their control env, and I've had this happen to me as well.

I think this is all we need to permanently fix it, unless anyone wants to advocate for the more conservative approach of pinning pip to a specific version, which I at this point think is probably overkill.


snopoke:
👍

PR #1191, opened 2017-12-22
more pillows for processing cases

snopoke:
@dimagi/scale-team

### DIFF
--- a/fab/fab/environments.yml
+++ b/fab/fab/environments.yml
@@ -484,6 +484,14 @@ icds-new:
         num_processes: 1
       UserPillow:
         num_processes: 1
+      kafka-ucr-static-cases:
+        start_process: 0
+        num_processes: 6
+        total_processes: 36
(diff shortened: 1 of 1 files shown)


snopoke:
I see kafka only has 24 case-sql partitions so we can't have more than 24 processes reading from it

snopoke:
I've added 12 more partitions to `case-sql` topic brining total to 36

snopoke:
@czue I'm basing them off stats in the (Change Feed Dashboard on DD](https://app.datadoghq.com/dash/256236/change-feeds?live=true&page=0&is_auto=false&from_ts=1513930533574&to_ts=1513934133574&tile_size=m&tpl_var_env=icds-new&fullscreen=false)

And the system load for the VMs

emord:
fwiw I don't think we saw any help when we moved from 8 to 12. My guess is that we're limted by ES. more shards would mean we could index more quickly and we may be limited by only having 5

PR #1198, opened 2017-12-23
double again

czue:
doesn't look like we're keeping up with the incoming rate of indicators yet so increasing this some more


snopoke:
By incoming rate do you mean from submissions or do you mean from the queue?

czue:
Not sure the source of the incoming indicators (form submissions doesn't seem right given timing) but I meant with the overall number of async indicators to process

PR #1208, opened 2018-01-02
allow adding multiple msp users; add new msp user for enikshay

mkangia:
Building on https://github.com/dimagi/commcarehq-ansible/pull/938/files
1. to allow multiple users to be added
2. to add a new user for enikshay


ansible/vars/enikshay/unx40423.pub
```diff
@@ -0,0 +1 @@
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDvNQLnUdNc+Y6gk+IcYjeasQ7hyCFn+OFBeNfe5hTZvAnOEPDzstTrs+SOcNLa/Qoc7XxfxIQBWKFkQUrJ17vKmm6suE6rXO6MmfAtAsxSQMPPTc6BjakraVncP0OcnG6qVV6P+r0HQJTicBMSU08heTsZfxBXIQbt39bjcjJiaA4E3wF57cgitcz0AEfmi9L5ah+0L2dST3FCHWqU9uXbdHE560PHW6PhXfisEx7F7XQahsniwOpT7dkmxaJKL3Pm7QTZdatUrJ5TceJxYisybxUKzmGDBuAONtIgcJ+dSRxiZZqlqu/rq8EwPpOYhndFafKm28u9EuhG16Y21Q/X unx40423@ubuntu
```
javierwilson: we don't need this, i think

javierwilson:
looks good to me! also, thanks this make things more flexible for other clients too.

ansible/vars/enikshay/unx40423.pub
mkangia: @javierwilson this is the different pub key that should be added for ssh for new user `unx40423` like we had for msp?

javierwilson: got it. yeah we keep those in files.

PR #1209, opened 2018-01-02
add new VMs to enikshay

mkangia:


mkangia:
Going to merge this now to update master since we are going to avoid making any more changes this week.
@calellowitz @millerdev

PR #1214, opened 2018-01-03
Ce/cas body buffer

calellowitz:
@czue @mkangia  cc: @snopoke @javierwilson @emord 
hoping to resolve some of the load on nginx by allowing to buffer larger requests in memory without resorting to temp files. 5m is what the rest of the routes use.

### DIFF
--- a/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/ansible/roles/nginx/vars/cas_ssl.yml
@@ -99,7 +99,7 @@ nginx_sites:
        proxy_redirect: "http://{{ CAS_SITE_HOST }} https://{{ CAS_SITE_HOST }}"
        proxy_next_upstream_tries: 1
        proxy_read_timeout: 900s
-       client_body_buffer_size: 1m 
+       client_body_buffer_size: 5m 
        limit_reqs:
          - zone_name: "submissions"
            burst: 100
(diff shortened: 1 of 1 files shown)


emord:
Guessing it has to buffer for submitted media files (images?)

It looks like we're getting pretty close to max memory during the day on web0. Any concern this could push it over the edge and run out of memory on some requests? From my understanding the max memory before would have been `# connections * 512k` and this would make it `# connections * 5m` 

may want to experiment with a dedicated proxy machine so that webworkers aren't interfering with the memory usage

emord:
@calellowitz any update on this?

> hoping to resolve some of the load on nginx

I think the best way to do this is going to be moving riak's proxy from web0 to a riak node (or some other machine), similar to how we have nginx on a couch node to proxy the couch requests. Thoughts?

calellowitz:
No real update on this. We ended up reverting this change as the issues continued, and we were scared it might be exacerbating them for the memory reasons you pointed out. Reverting it didn't change anything either, and I think the consensus is now that the issues we saw were almost certainly NIC firewall or networking issues, not our own proxy. For that reason, I would be in favor of redeploying this as we do buffer a lot of form submissions to files.

To your point about a dedicated proxy machine or moving the riak proxy, both seem like good ideas, although it is now not clear that proxy load was ever really the issue. Particularly has load has remained the same and we have not seen any more of those browser timeouts.

calellowitz:
@dimagi/scale-team (@emord / @snopoke in particular). Any objection to merging this. Despite the irrelevance to the issues we were seeing, it still seems like a good change. Otherwise, I am going to go ahead and close this PR

snopoke:
@calellowitz OK by me in principle. Were there any memory issues when this was deployed?

PR #1216, opened 2018-01-03
move inventory, ansible vars, and environments.yml sections to environments/<env>/

dannyroberts:
We've talked about this before, but I thought I'd give it a shot—now that these are all in the same repo, it wasn't that hard.

This takes all `inventory/{env}`, `{env}_public.yml`, and `{env}_vault.yml` files, as well as the {env} section of `environments.yml`, and moves them to `environments/{env}/`. This makes all setup that's specific to an environment live in a single directory. Each environment directory has a parallel structure, so the sum total looks like this

- `environments`
  - `{env}`
    - `app-processes.yml` (from the env's section of `environments.yml`)
    - `public.yml`
    - `vault.yml`
    - `inventory.ini`
- `environmental-defaults` (for setting defaults that the individual environments can override)
  - `app-processes.yml` (currently the only thing in here)

with of course one directory `{env}` per environment.

This PR also includes all the code changes necessary to pull from the new file locations. (Other than moving these files and their contents, these settings/vars files themselves weren't changed.)

**Why?** My own rationale is as follows: some of this static config, strewn across different directories, is mutually redundant because fab fetchs from file X and ansible fetches from file Y (though for the most part we've been pretty good about this), and all of these files are of interest to both fab and ansible (even if some variables/files are only currently used by one of them), so having some live in `fab/` and some live in `ansible/` is arbitrary and puts up barriers of convenience and mindset. Additionally, these things are more organization-specific config than they are code—if Igamid, Inc came along and wanted to deploy using our tools, everything now in `environments` would be totally different for them, whereas they could use everything else out of the box. So, while I'm not going to argue for it living in another repo or something at this point, it does seem worth delineating that clearly in the code structure itself.

Including my buddies and @snopoke and @millerdev for providing good feedback in the past on this kind of stuff. Let me know what you think! Also let me know if you think I should solicit a review from anyone else.


scripts/inventory/getinventory.py
```diff
@@ -48,7 +48,7 @@ def import_read_inventory_file():
 def get_instance_group(instance, group):
     read_inventory_file = import_read_inventory_file()
     servers = read_inventory_file(
-        os.path.join(ROOT, 'fab', 'fab', 'inventory', instance))
+        os.path.join(ROOT, 'environments', instance, 'inventory.ini'))
```
millerdev: Ah, I was wondering if this would be addressed since I ran into another issue with it today. I noticed that you must be in (have cwd set to) the root of the repo to run `./scripts/inventory/ssh`. I was in the `./ansible` subdir and tried
```
(ansible)::ansible$ ../scripts/inventory/ssh softlayer ansible@riakcs:0
python: can't open file 'scripts/inventory/getinventory.py': [Errno 2] No such file or directory
```
Would be really cool if we could `commcare-cloud ssh softlayer ansible@riakcs:0`  from anywhere (assuming the virtualenv is activated).

dannyroberts:
@millerdev yeah I think we're on the same wavelength because the filename thing bugged me too, but when I went through each of the alternatives, I didn't like them. The only contender in my mind is to prefix each file with `{env}_` so that we have e.g. `production/production_inventory.ini`, `icds-new/icds-new_public.yml`, etc. but I don't love the redundancy. My editor (PyCharm) detects the parallel structure (or maybe just the fact that it's a common file name or something?) and names the tabs e.g. `enikshay/public.yml` rather than just `public.yml` as it might for another file.

scripts/inventory/getinventory.py
dannyroberts: Yeah that's occurred to me too, could be pretty neat. The one thing I'm not sure about is that `commcare-cloud` is a python script and python scripts aren't that great at handing over control to whatever process they're forking (or maybe I just do this wrong)—so I'm not sure how well dumping the user in an interactive shell works. Do you know what I mean here, and do you know how to get around that?

dannyroberts: Another lower-touch way to fix this is to use `$(dirname $0)` to get the script's directory. PR'd that over here: https://github.com/dimagi/commcarehq-ansible/pull/1218

dannyroberts:
Just did some final tests (a fab deploy and a few ansible-playbook commands), and this looks good to go. I'm going to merge it before it gets stale, but we can revisit organization if there are things we want to improve.

dannyroberts:
Hey @dimagi/team-commcare-hq quick note on these changes. All environment specific config—inventory, ansible vars files, the former `environments.yml` contents (which are now split up by env)—now live under the top level `environments/` directory as e.g. `environments/production/inventory.ini`, etc. Shared environmental defaults that can be overridden by files in `environments` are now stored in `environmental-defaults/`—currently that's just the `base:` section of the former `environments.yml`.

Last night after merging this, I went ahead and merged master into all open PRs in this repo that were affected by the file move; all merges were straightforward with no conflicts because of git's (pretty impressive) renaming detection.

Lastly, apologies for any inconvenience that comes with doing this with much of the team offline. If you feel like you should have been consulted before this was merged or have any feedback of that nature, please let me know—I'm always trying to strike a balance between pushing things through quickly and painlessly and making sure what ends up going through doesn't feel to anyone like a midnight coup!

Similarly, now that it's in, if you have suggestions for better organization, this doesn't have to be the end-all-be-all—I'm hoping that this can just be a good jumping off point.

PR #1219, opened 2018-01-03
add nginx.error_log via datadog_parsers var

dannyroberts:
Followup to https://github.com/dimagi/datadog-parsers/pull/21


dannyroberts:
@ctsims this is deployed to enikshay

PR #1222, opened 2018-01-04
commcare-cloud ssh/mosh

millerdev:
Incorporate `ssh` and `mosh` scripts into `commcare-cloud`

@dannyroberts


commcare-cloud/commcare_cloud/commcare_cloud.py
```diff
@@ -516,6 +510,36 @@ def run_fab(args, unknown_args):
...
 
 
+class Ssh(object):
+    command = 'ssh'
+    help = "Connect to a remote host with ssh"
+
+    @classmethod
+    def make_parser(cls, parser):
+        cls.parser = parser
+        parser.add_argument("group",
```
dannyroberts: It might be worth mentioning that you can also name a server by "name" (since technically "named" servers are just putting the IP or DNS name under a group that contains just it), so `commcare-cloud softlayer ssh django0` and `commcare-cloud softlayer ssh webworkers:0` will incidentally both do the same thing. Even though your help string here is technically correct (`django0` is also a server "group"), I think it's worth calling out individually.

Nitpick: Also, since in the end this arg is supposed to signify a single server + username (which defaults via ssh/mosh to yours) and not a group of servers, I might call this arg `remote` or `server` or `remote_spec` or `server_spec` or something like that.

Not blocking, fine to do after if you like (or tell me to do it 😄)

commcare-cloud/commcare_cloud/commcare_cloud.py
```diff
@@ -516,6 +510,36 @@ def run_fab(args, unknown_args):
...
+                 "user and may be terminated with ':<n>' to choose one of "
+                 "multiple servers if there is more than one in the group. "
+                 "For example: webworkers:0 will pick the first webworker.")
+
+    @classmethod
+    def run(cls, args, ssh_args):
+        def exit(message):
+            cls.parser.error("\n" + message)
+        address = get_server_address(args.environment, args.group, exit)
+        cmd_parts = [cls.command, address] + ssh_args
```
dannyroberts: Ha! Clever. Maybe too clever? Since `command` here means "the name of the subcommand of `commcare-cloud`", maybe hijacking it to also mean "the name of the executable to pass these args to" is not worth the DRYness. Also not blocking, but I'd personally go with

```
command = 'ssh'
ssh_executable = 'ssh'
...
command = 'mosh'
ssh_executable = 'mosh'
```

(and then use `cls.ssh_executable` here and a few lines below) over repurposing `command` to mean both. Super-nit.

commcare-cloud/commcare_cloud/paths.py
```diff
@@ -0,0 +1,19 @@
...
+def get_public_vars_filepath(environment):
+    return os.path.join(ENVIRONMENTS_DIR, environment, 'public.yml')
+
+
+def get_vault_vars_filepath(environment):
+    return os.path.join(ENVIRONMENTS_DIR, environment, 'vault.yml')
+
+
+def get_inventory_filepath(environment):
+    return os.path.join(ENVIRONMENTS_DIR, environment, 'inventory.ini')
```
dannyroberts: Very nice, I approve of this reorg.

commcare-cloud/commcare_cloud/getinventory.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python
```
dannyroberts: I guess it's still worth keeping this file executable (as this line does)? I occasionally run `./scripts/inventory/getinventory.py` directly to print out the IP/DNS name of a machine, rather than opening the inventory file up in my editor and searching, so now I guess I'd run `commcare-cloud/commcare_cloud/getinventory.py`. Might be worth recasting this as `commcare-cloud <env> lookup <server>` but happy to leave that for another day.

commcare-cloud/commcare_cloud/commcare_cloud.py
millerdev: Thanks for the notes. I'll do this in a separate PR.

commcare-cloud/commcare_cloud/commcare_cloud.py
millerdev: Going to call a YAGNI here. We can add that if we ever override this class for a command that has a different executable name than whatever `command` value we pick.

commcare-cloud/commcare_cloud/getinventory.py
millerdev: Ha, I completely overlooked this line. It doesn't work that way anymore (I removed that functionality). Unfortunately my changes are not easy to see because I changed too much and git didn't realize the file was moved. If you want I will rebase this and do the move and refactoring in separate commits to make it more obvious.

I'll be happy to add the `lookup` command in a separate PR.

dannyroberts:
Ah okay, no worries. Sounds good to me!

PR #1229, opened 2018-01-05
Experiment: Allow for default public variables

dannyroberts:
This is a work in progress, but I thought I'd open up the approach for comment. The goal is to add a file `environmental-defaults/public.yml` that the other `public.yml` conceptually inherit from. (It inherits recursively so that `localsettings.ALLOWED_HOSTS` gets the parent value, rather than `localsettings` being blown over entirely.)

The approach here is to leave the code in an intermediate situation where each environment has a `public.yml` and an `expected-public.yml`. Running `./commcare-cloud/tests/test_vars.sh` checks that the defaults + `public.yml` = `expected-public.yml`, printing out

```
enikshay
✓ OK
icds
✓ OK
icds-new
✓ OK
l10k
✓ OK
pna
✓ OK
production
✓ OK
softlayer
✓ OK
staging
✓ OK
swiss
✓ OK
```

(or a diff if the end-values differ). That test is run by travis. It's also _super_ useful when converting repeated variables to default variables—you can prove to yourself nothing changed.

I wouldn't want to merge this before doing more work to pull out identical variables (I think there are tons) and leaving this in a better state, but I'm mostly interested to hear whether people would find that end-state valuable, or if you think another approach is better. Tell me what you think. This will be easiest to review commit by commit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,4 @@ script:
   - ".travis/tests.sh"
   - "diff <(ansible -h) commcare-cloud/commcare_cloud/help_cache/ansible.txt"
   - "diff <(ansible-playbook -h) commcare-cloud/commcare_cloud/help_cache/ansible-playbook.txt"
+  - "./commcare-cloud/tests/test_vars.sh"
--- a/commcare-cloud/commcare_cloud/commcare_cloud.py
+++ b/commcare-cloud/commcare_cloud/commcare_cloud.py
@@ -13,6 +13,7 @@ import yaml
 
 from .getinventory import get_server_address
 from .parse_help import filtered_help_message, add_to_help_text
+from .var_files import get_consolidated_vars, get_public_vars, get_expected_public_vars
 from .paths import (
     ANSIBLE_DIR,
     ENVIRONMENTS_DIR,
@@ -73,12 +74,6 @@ def arg_stdout_callback(parser):
         help=("The callback plugin to use for generating output. "
               "See ansible-doc -t callback -l and ansible-doc -t callback [ansible|minimal]"
     ))
-
-
-def get_public_vars(environment):
-    filename = get_public_vars_filepath(environment)
-    with open(filename) as f:
-        return yaml.load(f)
 
 
 def get_common_ssh_args(public_vars):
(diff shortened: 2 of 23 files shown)


millerdev:
Am I understanding correctly that if we were to adopt this we would need to maintain two vars files: `public.yml` and `expected-public.yml`? The first being a slimmed down, hopefully containing nothing more than differences from the default `public.yml`, and the second being what is expected when the default and env-specific vars are merged. Is it true that you could just drop the defaults and maintain only `expected-public.yml` for each env and you'd have the same result and less code?

Forgive my skepticism, but something has a faint odor here. I can't quite put my finger on it. I think most of the time the changes we'll make in `public.yml` will be so similar to the corresponding changes in `expected-public.yml` that it will just feel like redundant work that didn't tell us anything meaningful. Kinda reminds me of the feeling I get when I write a bad test and I feel like the test didn't really test anything, it was just a mirror of the real code. It's brittle and needs to be updated whenever anything in the code changes. Maybe the experience of actually doing it will turn out to be different?

Also the merge algorithm does not appear to handle the case where you *want* to replace an entire dict with a different dict rather than merging them. In other words, it looks like there is no way to say *delete this key, which is in the default dict, from the final result*. Maybe that doesn't matter in practice?

Sorry for not getting on board with this immediately. I'll try to be open-minded if you feel like this is worthwhile, despite my complaining.

dannyroberts:
@millerdev Yeah sorry I didn't explain this that well. Maintaining expected-public.yml was supposed to be a temporary thing that bridges the transition from "all vars repeated in each env to "all shard stuff lives in the default file" while leaving as little room for human error as possible. Step two would be to remove this scaffolding and leave just the default public.yml and each environment's public.yml.

As for your question about the merge algorithm, you're absolutely correct. You can always set a key back to `null`, but you would have to do so explicitly and you can't remove a key. I think that's not a horrible convention to have (that absent and `null` should mean the same thing), and something we could probably live with. My concern is more whether we think this default / environment-specific override pattern is something that would make these files easier to maintain or harder to maintain.

I could see the criticism that the need for this is to some extent caused more by poorly a poorly organized variable structure to begin with. For example, the `active_sites` variable makes us specify _which nginx `sites-enabled` files we want to create_, which is a super low-level thing to put in a settings file. Rather than expose

```
NO_WWW_SITE_HOST: 'commcarehq.org'
...
  cchq_www_redirect_secure: True
  cchq_www_redirect_insecure: True
```

in the settings, you could have those nginx files be governed by conditions, or even make `NO_WWW_SITE_HOSTS` a list, and include one redirect config per host. And there are probably a bunch of other examples.

On the other hand, there's tons of boilerplate like this, where the simplest way to DRY that would be to put it in a default public.yml file:
```
localsettings:
...
  AMAZON_S3_ACCESS_KEY: "{{ localsettings_private.AMAZON_S3_ACCESS_KEY }}"
  AMAZON_S3_SECRET_KEY: "{{ localsettings_private.AMAZON_S3_SECRET_KEY }}"
  SECRET_KEY: "{{ localsettings_private.SECRET_KEY }}"
  SENTRY_PUBLIC_KEY: "{{ localsettings_private.SENTRY_PUBLIC_KEY }}"
  SENTRY_PRIVATE_KEY: "{{ localsettings_private.SENTRY_PRIVATE_KEY }}"
  SENTRY_PROJECT_ID: "{{ localsettings_private.SENTRY_PROJECT_ID }}"
  SENTRY_QUERY_URL: "{{ localsettings_private.SENTRY_QUERY_URL }}"
  SENTRY_API_KEY: "{{ localsettings_private.SENTRY_API_KEY }}"
  SIMPLE_API_KEYS: "{{ localsettings_private.SIMPLE_API_KEYS }}"
  SMS_GATEWAY_PARAMS: "{{ localsettings_private.SMS_GATEWAY_PARAMS }}"
```
so I think there's a case to be made for both cleaning up the "contract" (what variables need to be set in public.yml in the first place) _and_ allowing for this type of default/override inheritance.

I may have just picked a bad first example (`active_sites`) to try this on because it was at the top of the yaml file.

Thoughts?

millerdev:
> Maintaining expected-public.yml was supposed to be a temporary thing that bridges the transition

Ah, that's very helpful and makes a lot more sense. I'm fine with this if it's just temporary and that period doesn't last for too long.

> My concern is more whether we think this default / environment-specific override pattern is something that would make these files easier to maintain or harder to maintain.

That's a good question. I can see pros and cons in both directions. I'm not sure which is better.

Also agree that improving the organization of the variable structure, resulting in less boilerplate in environment settings, is a good idea. I'd love to see more effort focused on that. An example that has worked out well so far is the [`riak_backend`](https://github.com/dimagi/commcarehq-ansible/blob/b06e900c394bff1d54c842e229072b5fc1dfb742/ansible/roles/riakcs/configure/defaults/main.yml#L1-L20) setting, which controls a whole set of [other settings related to riak](https://github.com/dimagi/commcarehq-ansible/blob/b06e900c394bff1d54c842e229072b5fc1dfb742/ansible/roles/riakcs/configure/vars/main.yml#L7-L25).

dannyroberts:
Neat, yeah. Well it seems like that's a promising direction to start with then (cleaning up what has to go in public.yml) rather than this. Then we can see if something like this is still needed/worth it. I'll go ahead and close this for now, thanks for the thoughts!

PR #1231, opened 2018-01-05
Ce/aliases

calellowitz:
@dannyroberts @javierwilson @esoergel @czue cc: @mkangia 
This reimplements the old (and now broken) aliases for ansible to use danny's new command. I updated the printout at the top but they work similarly to the old ones. It is now `ap <env>` and `aps <env>`.

I really liked the ease of the old aliases and got tired of how verbose the new commands are, but agree the new ui is a big improvement. When the variables moved and the old aliases were broken it seemed like the right time to fix them. Let me know if you fee there is a better way.

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -32,13 +32,20 @@ wait
 # convenience: . init-ansible
 [ ! -f ~/init-ansible ] && ln -s ~/commcarehq-ansible/control/init.sh ~/init-ansible
 ~/commcarehq-ansible/control/check_install.sh
-alias ap='ansible-playbook -u ansible -i ~/commcarehq-ansible/fab/fab/inventory/$ENV -e "@vars/$ENV/${ENV}_vault.yml" -e "@vars/$ENV/${ENV}_public.yml" --ask-vault-pass'
-alias aps='ap deploy_stack.yml'
 alias update-code='~/commcarehq-ansible/control/update_code.sh && . ~/init-ansible'
 alias update_code='~/commcarehq-ansible/control/update_code.sh && . ~/init-ansible'
 
 [ ! -f ~/.bash_completion ] && source  ~/commcarehq-ansible/control/.bash_completion
 cp ~/commcarehq-ansible/control/.bash_completion ~/
+
+function ap() {
+    commcare-cloud $1 ansible-playbook
+}
+
+
+function aps() {
+    ap $1 deploy_stack.yml
+}
 
 function ae() {
     ansible $1 -m shell -a "$2" -u ansible -i ~/commcarehq-ansible/fab/fab/inventory/$ENV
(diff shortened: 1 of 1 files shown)


control/init.sh
```diff
@@ -32,14 +32,25 @@ wait
...
 alias update-code='~/commcarehq-ansible/control/update_code.sh && . ~/init-ansible'
 alias update_code='~/commcarehq-ansible/control/update_code.sh && . ~/init-ansible'
 
 [ ! -f ~/.bash_completion ] && source  ~/commcarehq-ansible/control/.bash_completion
 cp ~/commcarehq-ansible/control/.bash_completion ~/
 
+function ap() {
+    ENV=$1
+    shift
+    commcare-cloud $ENV ansible-playbook $@
```
millerdev: Should have `$@` in quotes for proper quoting of args containing spaces: `"$@"`

control/init.sh
```diff
@@ -67,8 +78,8 @@ function ansible-control-banner() {
     printf "${BLUE}update-code${NC} - update the ansible repositories (safely)\n"
     printf "${BLUE}workon ansible${NC} - activate the ansible virtual env\n"
     printf "${BLUE}commcare-cloud${NC} - CLI wrapper for ansible. See ${YELLOW}commcare-cloud -h${NC} for more details.\n"
-    printf "ap - Use ${YELLOW}commcare-cloud <env> ansible-playbook${NC} instead.\n"
-    printf "aps - Use ${YELLOW}commcare-cloud <env> ansible-playbook deploy_stack.yml${NC} instead.\n"
+    printf "ap - Alias for ${YELLOW}commcare-cloud <env> ansible-playbook${NC} e.g. ap production deploy_proxy.yml \n"
+    printf "aps - Alias for ${YELLOW}commcare-cloud <env> ansible-playbook deploy_stack.yml${NC} e.g. aps production \n"
```
millerdev: You've changed the expected arguments of these aliases, so they are not technically backward compatible with how they worked before.
```
# before
ENV=staging && aps --diff

# after
aps staging --diff
```
Seems like that could be confusing for anyone who's not paying close attention to all these changes recently.

millerdev: Personally I'd like to advocate for having the documented and recommended way to execute these commands be to use the `cchq` alias for `commcare-cloud` (@dannyroberts is working on documenting that now). `cchq` is pretty slick and actually requires less typing than before when paired with the new bash completion @dannyroberts implemented.

If you still want very terse aliases like `ap` and `aps`, what do you think about adding them to your bash profile?

millerdev: [`cchq` alias is now documented](https://github.com/dimagi/commcarehq-ansible/tree/master/commcare-cloud#optional-hook-up-the-bells-and-whistles)

dannyroberts: (I want to also give a shoutout to @javierwilson for doing the initial setup and most of the work for the commcare-cloud bash completion, which has proven totally awesome.)

dannyroberts: @calellowitz Worth weighing back in here? Good to decide whether we're going to go with this, close and go with @millerdev's suggestion, or something different.

calellowitz: I would still like this included, as I do think it remains easier to use than the the new command (even with the slightly shorter alias) and I think encourages people to do the "right thing" (by making `deploy_stack.yml` the easiest possible command as it used to be). I am not committed to this version if you or @millerdev have an alternative suggestion, but I do think `deploy_stack` should be treated separately and made as easy as possible, and preferably different from other playbooks to encourage people to use it.

All that said, if there is strong resistance to that, I am happy to follow daniel's suggestion and add this to my own profile. I just think it might be useful/good for others.

calellowitz: @dannyroberts forgot to mention

dannyroberts:
@calellowitz how do you feel about `cchq staging aps`: https://github.com/dimagi/commcare-cloud/pull/1306

Given that previously you had to write

```
$ ENV=staging && aps --diff --check
<wait, verify>
$ ENV=staging && aps --diff
```

I'd say this is _more_ terse than what we had before. It's all in how you look at it ;) I guess we could get it all the way down to `aps staging` by adding another alias to the bash profile or something, but I don't feel strongly about it.

calellowitz:
sounds great. thanks @dannyroberts

PR #1233, opened 2018-01-05
remove ap and aps control machine aliases

dannyroberts:
if these are not out of date already, they soon will be. I think it is overall dangerous to keep these around much longer


dannyroberts:
ha! just saw https://github.com/dimagi/commcarehq-ansible/pull/1231/files—never mind!

PR #1234, opened 2018-01-05
document commcare cloud install

dannyroberts:
This adds a README to `commcare-cloud` with install instructions, and also documents how to hook up extra features like bash completion and running `commcare-cloud` anywhere outside the virtualenv.

### DIFF
--- a/commcare-cloud/README.md
+++ b/commcare-cloud/README.md
@@ -27,13 +27,14 @@ workon ansible
 ```
 
 ```
-pip install -r requirements.txt
+pip install pip --upgrade
 ```
 
-To be able to run ansible commands, you must additionally run
-
 ```
-ansible-galaxy install -r ansible/requirements.yml
+pip install -e commcare-cloud
+pip install -r ansible/requirements.txt  # to be able to run ansible commands
+ansible-galaxy install -r ansible/requirements.yml  # to be able to run ansible commands
+pip install -r fab/requirements.txt  # to be able to run fab commands
 ```
 
 Finally, finish your install with
--- a/commcare-cloud/README.md
+++ b/commcare-cloud/README.md
@@ -1,0 +1,75 @@
+# Set up commcarehq-ansible
+
+Clone the commcarehq-ansible repo (suggested location alongside commcare-hq):
+
+```
+git clone https://github.com/dimagi/commcarehq-ansible.git
+cd commcarehq-ansible
+```
+
+Now make a virtualenv for ansible:
+
+```
+mkvirtualenv ansible
+```
+
+**Note**: If you already have commcarehq-ansible cloned, then just enter that directory
+and update it with
+
+```
+git pull
+```
+
+# Install commcare-cloud
+
+```
+workon ansible
+```
+
+```
+pip install -r requirements.txt
+```
+
+To be able to run ansible commands, you must additionally run
+
+```
+ansible-galaxy install -r ansible/requirements.yml
+```
+
+Finally, finish your install with
+
+```
+./control/check_install.sh
+```
+
+# Optional: Hook up the bells and whistles
+
+If you want to have access to the `commcare-cloud` command outside your virtualenv,
+you can link the executable (within your vitualenv) to somewhere on your PATH.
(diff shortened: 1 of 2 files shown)


commcare-cloud/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+mkvirtualenv ansible
+```
+
+If you want `workon ansible` to always bring you to this directory, then you can also run
+
+```
+pwd > ~/.virtualenvs/ansible/.project
+```
```
esoergel: virtualenvwrapper has a command for this, actually: https://github.com/dimagi/commcarehq-ansible/pull/1240

commcare-cloud/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+pip install pip --upgrade
+```
+
+```
+pip install -e commcare-cloud
+pip install -r ansible/requirements.txt  # to be able to run ansible commands
+ansible-galaxy install -r ansible/requirements.yml  # to be able to run ansible commands
```
esoergel: does this make sense locally too?

commcare-cloud/README.md
```diff
@@ -0,0 +1,84 @@
...
+pip install -r fab/requirements.txt  # to be able to run fab commands
+```
```
esoergel: How would you feel about a top level `requirements.txt` file like:
```
-e commcare-cloud
-r ansible/requirements.txt
-r fab/requirements.txt
```
That way everything is updated in one go

commcare-cloud/README.md
```diff
@@ -0,0 +1,84 @@
...
+# Optional: Hook up the bells and whistles
+
+If you want to have access to the `commcare-cloud` command outside your virtualenv,
+you can link the executable (within your vitualenv) to somewhere on your PATH.
+While it's technically redundant, I like to accomplish this by adding the link command
+to my bash profile:
+
+```
+ln -sf ~/.virtualenvs/ansible/bin/commcare-cloud /usr/local/bin
+```
```
esoergel: Personally I like the idea of adding that `bin` dir to your path rather than symlinking stuff:
```export PATH="~/virtualenvs/ansible/bin:$PATH"```
no strong feelings though

File: commcare-cloud/README.md [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+
+(Instead of `/usr/local/bin`, you will need to use a directory that is owned by your user
+and on your PATH.)
+
+If you want to have access to bash completion for the commcare-cloud command, you will
+need to add
+
+source ~/.commcare-cloud/repo/control/.bash_completion
```
esoergel: Forgot the triple backticks here

esoergel:
Awesome, thanks for writing this all up, Danny!  It mirrors my experiences pretty well too.

commcare-cloud/README.md
dannyroberts: Yeah it can; I sometimes run simple commands from my computer. It's not strictly necessary for everything, but it's nice to get everyone doing all the installs so there's some kind of consistency.

commcare-cloud/README.md
dannyroberts: Thanks for bringing this up. I was initially going to tell you that I actually had it this way in my first commit on this PR, but that it complained because both `ansible/requirements.txt` and `fab/requirements.txt` contained ansible (the same version even). But then I thought I'd give it one more shot, and was able to ascertain that that only happens with an old version of pip. So as long as you `pip install pip --upgrade` (which I later added to this README and which you have to do anyway for some of the installs), you can actually put these in a single file as I tried and as you just suggested!

I'm going to go ahead and do that. Thanks!

commcare-cloud/README.md
dannyroberts: Yeah this occurred to me early on, but I ruled it out pretty immediately because of the other things in that directory, which include e.g. `python`. I think making this always on the path would lead to confusing behavior and overriding other programs like this that people aren't expecting.

commcare-cloud/README.md
dannyroberts: Thanks will fix

PR #1242, opened 2018-01-05
Link to executables dir

dannyroberts:
OK I found what I believe to be a much nicer way to do this. A lot easier to tell the user what to do in the README.

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -39,6 +39,7 @@ alias update_code='~/commcarehq-ansible/control/update_code.sh && . ~/init-ansib
 
 [ ! -f ~/.bash_completion ] && source  ~/commcarehq-ansible/control/.bash_completion
 cp ~/commcarehq-ansible/control/.bash_completion ~/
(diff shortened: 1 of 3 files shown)


control/init.sh
```diff
@@ -39,6 +39,7 @@ alias update_code='~/commcarehq-ansible/control/update_code.sh && . ~/init-ansib
 
 [ ! -f ~/.bash_completion ] && source  ~/commcarehq-ansible/control/.bash_completion
 cp ~/commcarehq-ansible/control/.bash_completion ~/
```
millerdev: Unrelated to this PR: this will overwrite `~/.bash_completion` if it exists. I'm not sure that's a good idea since it's not a particularly project-specific name and people tend to not always have backups of all their files...

dannyroberts: Yeah, seems better to just replace these two lines with `source  ~/commcarehq-ansible/control/.bash_completion`, right?

millerdev: yes

PR #1249, opened 2018-01-09
Turn on analytics debug flag on staging

orangejenny:
Don't know why I set this to `False` in https://github.com/dimagi/commcarehq-ansible/pull/1132 ...can't test analytics on staging if debug is off.

Verified that the staging vault contains staging-specific keys for google and kissmetrics, and no key for hubspot, so we don't send staging data to real services.

@calellowitz


orangejenny:
deployed to staging

PR #1252, opened 2018-01-10
password hashing

sravfeyn:
This is required on softlayer as AKS is auditing ICDS dashboard on it.


calellowitz:
@sravfeyn looks like this got lost somewhere, is this still necessary

sravfeyn:
Nope, closing!

PR #1258, opened 2018-01-10
Make fab auto-create tasks for each env

dannyroberts:
Move any differences between envs into `app-settings.yml` so that creating the `@task` for an env is rote; then just auto-generate those tasks. Removes a lot of boilerplate code and puts these differences together into a config file in the environment.

### DIFF
--- a/fab/fab/fabfile.py
+++ b/fab/fab/fabfile.py
@@ -28,6 +28,7 @@ Server layout:
 from __future__ import absolute_import
 from __future__ import print_function
 import datetime
+import functools
 import os
 import posixpath
 from getpass import getpass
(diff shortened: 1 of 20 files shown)


fab/fab/fabfile.py
```diff
@@ -167,107 +160,59 @@ def get_env_dict(path):
...
+    if window:
+        d = datetime.datetime.now(pytz.timezone(window['timezone']))
+        if window['hour_start'] <= d.hour < window['hour_end']:
+            return
+    else:
         return
 
     message = (
-        "Woah there bud! You're deploying '%s' during the day. "
+        "Whoa there bud! You're deploying '%s' during the day. "
```
czue: [the more you know](https://www.grammarly.com/blog/whoa-woah/)

dannyroberts: It apparently even has a [known etymology](https://www.etymonline.com/word/whoa). You learn interesting things as a programmer.

emord: whoa

PR #1260, opened 2018-01-11
fix fab/fab/etc_hosts.py

dannyroberts:
I came across this and could see it right away it hadn't been updated for the recent fab reorganization. This both fixes it and simplifies it to use shared helper functions.

### DIFF
--- a/fab/fab/etc_hosts.py
+++ b/fab/fab/etc_hosts.py
@@ -4,30 +4,17 @@ from __future__ import print_function
 
 import logging
 import os
+import sys
 import re
 from argparse import ArgumentParser
-from os.path import dirname
 
-from ansible.inventory import Inventory
-from ansible.parsing.dataloader import DataLoader
-from ansible.vars import VariableManager
+from commcare_cloud.paths import get_available_envs
+
+sys.path.append(os.path.dirname(os.path.dirname(__file__)))  # this lets us import from utils file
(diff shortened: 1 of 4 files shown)


dannyroberts:
Yeah, agree, I also didn't know it existed, just stumbled upon it. @snopoke might be the best person to say what it's for / when to use it. I think the point of it is to give you something you can copy to your own laptop's /etc/hosts to be able to ssh into machines by their aliases. (The same thing can be achieved albeit more verbosely per command with `commcare-cloud <env> ssh <machine>`, though the /etc/hosts aliasing may open up other workflows I'm not thinking of.)

snopoke:
@dannyroberts @czue I find it useful to have these in my hosts file since it much easier than using the old ssh script which requires you to be in the right folder etc. Now that it's moved to the commcare-cloud tool I'll try that approach. One other thing that I do though is for ICDS there are 2 routes to the servers depending on which VPN you are connected to. I have my ssh config setup to route all IPs in the ICDS range via the shared VPN but if I'm using the other VPN then I ssh using the mapped host name.

I'd be curious how other people manage this workflow.

czue:
> I'd be curious how other people manage this workflow.

I usually use my bash history / local brain memory and - when I need - look up the IP manually in the inventory file like a neanderthal :monkey:

PR #1261, opened 2018-01-11
Move shared vars to group_vars/all.yml

dannyroberts:
This has annoyed me for a long time now. First pass at moving boilerplate logic into `group_vars/all.yml` instead of having it repeated in every `public.yml`.


ansible/group_vars/all.yml
```diff
@@ -145,3 +140,21 @@ eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
 daily_deploy_email: null
+
+
+AMQP_PASSWORD: "{{ secrets.AMQP_PASSWORD }}"
```
emord: There are a lot of passwords in here. I don't think we should be sharing passwords between envs.

What about having these as `None` and having checks where they are used to ensure they've been set in the environments

dannyroberts: I think I don't understand your comment. Maybe we have different mental models of what's happening here.

My perspective is that the `secrets` variable, a dict, gets set in the environment's `vault.yml` file, which is of course different for each environment. These key/value pairs need to be copied into variables of the same name in global scope, and previously that was being done separately in each `public.yml`. Now, it's done in `group_vars/all.yml` regardless of environment. I think conceptually there's no difference here; it's just a matter of not repeating the same logic once per env.

emord: Ohh. I see. so `secrets.AMQP_PASSWORD` is just sharing the definition of a secret and `secrets` is loaded from the vault file for the current env you're deploying to. Similar to usage of localsettings. This makes sense. Originally I thought these secrets were going to be shared since you were putting them in the shared group vars

snopoke:
I think when I originally set up the vault files the loading order of variables prevented this approach but perhaps that's changed since then.

If it works I'm all for it!

PR #1269, opened 2018-01-15
make netvault install work

sravfeyn:
@snopoke 
This makes netvault install work via Ansible. I have tested this on local vagrant setup.


ansible/deploy_netvault.yml
```diff
@@ -1,11 +1,11 @@
 - name: "Install Netvault Client"
-  hosts: pg_standby:couchdb2
+  hosts: pg_backup
```
javierwilson: pg_backup doens't really have a running instance of postgresql so pg_dump won't work there, and netvault uses pg_dump, so I am not sure this will work, unless we try offline backups with netvault.

ansible/deploy_netvault.yml
```diff
@@ -20,8 +20,14 @@
         dest: /usr/lib64
         state: link
 
-    # FIXME: deal with prompt for options
     - name: "Runs install script"
-      #command: /tmp/netvault/install
-      command: /bin/true
-      when: false
+      expect:
```
javierwilson: cool, thanks for this

ansible/deploy_netvault.yml
sravfeyn: Yeah, it was a wrong update (used for local test). I have updated it now

javierwilson:
Creaing its own group makes sense, it looks good to me.

sravfeyn:
I have already deployed this on NIC (it went OK)

PR #1271, opened 2018-01-16
fix ssh command

dannyroberts:
broken since #1260 (author: me).


esoergel:
merging before build because travis is down

PR #1273, opened 2018-01-17
replace commcarehq-ansible with commcare-cloud throughout

dannyroberts:


snopoke:
@dannyroberts brought this up yesterday. His thinking is that since this repo now contains more than just ansible code and is turning into a 'HQ infrastructure tools' repo that renaming it might be a good idea especially if we're heading in the direction of external parties making more use of it.

dannyroberts:
Yeah, that's right. To add a bit to that, we're intentionally setting ourselves up to make our tooling more externally accessible, something we can report out to commcare developers forum about, blog about, keep visible to our funders and partners, etc.—and be able to share these tooling improvements we're making with the community. That's what motivated giving this repo a name that's more in line with what it's becoming (unified tooling for managing a CommCare cloud environment) rather than what it started out as. I'm hoping the naming switch will be relatively painless internally, though I do accept that it does cause some bit of shuffling, especially if people want to rename their own local copies (which isn't at all required to keep things working).

czue:
:+1: makes sense thanks

dannyroberts:
Added a few commits to fix control machine setup after testing on all control machines. Should be good to go now

PR #1276, opened 2018-01-17
Dmr/swiss updates

dannyroberts:
A few changes from touching swiss today.

### DIFF
--- a/environments/swiss/public.yml
+++ b/environments/swiss/public.yml
@@ -52,6 +52,7 @@ pgbouncer_pool_mode: transaction
 pgbouncer_pool_timeout: 1
 pgbouncer_reserve_pool: 5
 postgresql_max_connections: 300
+postgresql_wal_keep_segments: 0
 postgresql_work_mem: '1MB'
 
 formplayer_db_name: formplayer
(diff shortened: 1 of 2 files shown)


environments/swiss/public.yml
```diff
@@ -46,12 +46,14 @@ blobdb_s3: False
...
 pgbouncer_default_pool: 290
-pgbouncer_reserve_pool: 5
-pgbouncer_pool_timeout: 1
+pgbouncer_max_connections: 400
 pgbouncer_pool_mode: transaction
+pgbouncer_pool_timeout: 1
+pgbouncer_reserve_pool: 5
+postgresql_max_connections: 300
+postgresql_wal_keep_segments: 0
+postgresql_work_mem: '1MB'
```
czue: don't know what the last two do, but assume this is just documenting what's already there?

dannyroberts: Yeah that's right. The point here was to update the config here to what swiss currently has, so that deploying to swiss doesn't change it in any significant way.

javierwilson: i agree with this, we should make it ansible-deployable first, then start changing stuff using ansible.

PR #1277, opened 2018-01-17
make commcare-cloud <env> ssh respect port

dannyroberts:


commcare-cloud/commcare_cloud/commcare_cloud.py
```diff
@@ -556,6 +556,9 @@ class Ssh(Lookup):
     @classmethod
     def run(cls, args, ssh_args):
         address = cls.lookup_server_address(args)
+        if ':' in address:
+            address, port = address.split(':')
+            ssh_args += ['-p', port]
```
millerdev: Might want to insert these at the beginning of `ssh_args`. Something like
```py
ssh_args = ['-p', port] + ssh_args
```
As is I'm not sure it will work for a command like `cchq production ssh redis:2222 -t screen -DR` since the `-p 2222` args would be passed to screen instead of ssh.

dannyroberts: Keen eye, will do. Thanks @millerdev!

dannyroberts:
(retested and it still works)

PR #1279, opened 2018-01-17
Make icds an alias to icds-new

dannyroberts:
Thoughts?

### DIFF
--- a/environments/icds
+++ b/environments/icds
@@ -1,0 +1,1 @@
+icds-new
--- a/environments/icds/fab-settings.yml
+++ b/environments/icds/fab-settings.yml
@@ -1,4 +1,0 @@
-acceptable_maintenance_window:
-  hour_start: 0
-  hour_end: 7
-  timezone: Asia/Kolkata


environments/icds
```diff
@@ -0,0 +1 @@
+icds-new
```
dannyroberts: This is a symlink if that wasn't clear.

snopoke:
@czue The folder structure is already just `icds` and not `icds-new`.

czue:
oh great! ignore me.

dannyroberts:
Yeah that's right. We've tried to make it so the folder name _only_ affects how you reference it in tools like ansible/fab/commcare-cloud, and does _not_ affect any of the setup. Folder names, supervisorctl task names, and such are determined by the (somewhat redundant) variables `environment` in `app-processes.yml` and `deploy_env` in `public.yml`.

dannyroberts:
Oh and `env_name` in `public.yml`, which is meant to be unique, for use in things like datadog, email subjects, e.g. "icds-new". That's versus `deploy_env` and `environment` which are meant to be the "conceptual" instance, and can be shared between two environments that represent the same deployment (esp. during a migration), e.g. "icds".

Writing that down, it strikes me that that's super not clear from the variable names.

PR #1281, opened 2018-01-18
increase proxy buffer size to 64k

dannyroberts:
@ctsims noticed a lot of warnings related to this in enikshay logs. The default limits seem super low, so I'm upping this a bit to see if it changes the volume of these errors (which we can see in datadog).


ansible/roles/nginx/vars/riakcs.yml
```diff
@@ -23,6 +23,7 @@ nginx_sites:
        proxy_read_timeout: 90
        proxy_buffer_size: 64k
        proxy_buffers: 8 64k
+       proxy_buffer_size: 64k
```
emord: This is double defined

dannyroberts: Oh dear, I was distracted and put this in the wrong file. Thanks

PR #1286, opened 2018-01-19
give nginx_hsts_max_age a default value

dannyroberts:
currently envs without this defined fail


dannyroberts:
Yeah, adding a default is my preference. I usually search for variables, and having a nice `some_variable: null` or other default value in one of the default value files I think is nicer than checking for existence. Another nice thing about it is you don't get into a situation where the only way to get a certain default behavior is to _remove_ a variable, because then there's no way to set it explicitly and if someone tries to (by setting it to null) then it the behavior may not be what they expect. I guess there might be arguments on the other side, but these are my reasons.

PR #1288, opened 2018-01-20
modification to make Swiss Ansible deployable 

NitigyaS:
- In `ansible/roles/postgresql/templates/pg_hba.conf.j2`
devreadonly is added since it was there in the pg_hba.conf of Swiss Postgres and it's vault file

### DIFF
--- a/ansible/roles/postgresql/templates/pg_hba.conf.j2
+++ b/ansible/roles/postgresql/templates/pg_hba.conf.j2
@@ -17,9 +17,7 @@ local   replication     postgres                                peer
 {% if postgres_users.get('replication') %}
 host    replication     {{ postgres_users.replication.username }}   {{ cluster_ip_range|default('0.0.0.0/0') }}    md5
 {% endif %}
-{% if postgres_users.get('devreadonly') %}
-host    replication     {{ postgres_users.devreadonly.username }}   {{ cluster_ip_range|default('0.0.0.0/0') }}    md5
-{% endif %}
+
 {% if postgres_users.get('backup') %}
 host    all             {{ postgres_users.backup.username }}   {{ cluster_ip_range|default('0.0.0.0/0') }}    md5
 {% endif %}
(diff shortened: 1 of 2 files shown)


ansible/roles/postgresql/templates/pg_hba.conf.j2
```diff
@@ -17,6 +17,9 @@ local   replication     postgres                                peer
 {% if postgres_users.get('replication') %}
 host    replication     {{ postgres_users.replication.username }}   {{ cluster_ip_range|default('0.0.0.0/0') }}    md5
 {% endif %}
+{% if postgres_users.get('devreadonly') %}
+host    replication     {{ postgres_users.devreadonly.username }}   {{ cluster_ip_range|default('0.0.0.0/0') }}    md5
```
javierwilson: i wonder why we need this, are we using replication in swiss? does devreadonly need replication permissions?

environments/swiss/public.yml
```diff
@@ -59,11 +59,15 @@ formplayer_db_name: formplayer
 
 postgresql_dbs:
   - name: commcarehq_ucr
+    query_stats: True
+    host: "{{ groups.postgresql.0 }}"
```
javierwilson: my guess is the role complained if hosts wansn't defined here? we  did the same in PNA so i guess this is the way to go.

ansible/roles/postgresql/templates/pg_hba.conf.j2
snopoke: I think it's used by Datadog

environments/swiss/public.yml
snopoke: Currently hosts defaults to `localsettings.PG_DATABASE_HOST` though I'm if we get rid of that default (and probably the `localsettings.PG_DATABASE_HOST` altogether).

ansible/roles/postgresql/templates/pg_hba.conf.j2
snopoke: I think in the other envs (can check the vault files) the `devreadonly` user is defined in the vault file and so it gets added below on line 37: https://github.com/dimagi/commcare-cloud/pull/1288/files#diff-d959b4d4814d1bc5d1d095125136a90aR37

I think it would be better to be consistent and add it that way. The reason we have special lines for `replication` and `netvault` is that they require 'special' setting.

javierwilson: yes, i wondered why devreadonly also needed "replication" permissions or be setup , i think it doesn't. and i agree, let's do it as it's done in other envs (users in the vault)

PR #1292, opened 2018-01-23
use ip address

javierwilson:
ipaddr filter returns False if not an IP address.
i've changed it to ansible_default_ipv4, which works in monoliths, not sure if it works in prod.
perhaps we could just use {{ inventory_hostname }} ?


javierwilson:
could this be a problem https://github.com/dimagi/commcare-cloud/blob/master/ansible/roles/couchdb2/vars/main.yml#L9 ? if used in environments with hostnames instead of ip numbers it may translate to False.

PR #1304, opened 2018-01-23
Allow for commcare-cloud command aliases

dannyroberts:
This PR adds an alias `ap` for `ansible-playbook`

```
commcare-cloud <env> ap <playbook> ...
```

as well as doing some restructuring that makes adding other aliases trivial.

This required adding the latest pip argparse, which comes with python3 but not python2, as a dependency, and then doing a few lines of black magic to manually import that instead of the standard library argparse. (Python 2 Standard library `argparse` doesn't include support for aliases.) If this feels too gross to people, I can always scrap the alias feature and just PR the first commit (which which gives all command classes a shared base class), which is a nice change in itself.

### DIFF
--- a/commcare-cloud/commcare_cloud/argparse14.py
+++ b/commcare-cloud/commcare_cloud/argparse14.py
@@ -1,3 +1,13 @@
+"""
+This file allows us to import argparse 1.4, a future version
+
+not available in the python 2.7's standard library (but is in python3).
+This file allows us to use the following relative import statement
+
+  from .argparse14 import ArgumentParser
+
+rather than inlining this manual import magic.
+"""
 from __future__ import print_function
 from __future__ import absolute_import
 import pkgutil
--- a/commcare-cloud/commcare_cloud/argparse14.py
+++ b/commcare-cloud/commcare_cloud/argparse14.py
@@ -11,14 +11,10 @@ rather than inlining this manual import magic.
 from __future__ import print_function
 from __future__ import absolute_import
 import pkgutil
-import sys
-import os
+from commcare_cloud.environment import get_virtualenv_site_packages_path
 
-for filepath in sys.path:
-    if os.path.isdir(filepath) and 'argparse.py' in os.listdir(filepath):
-        argparse = pkgutil.get_importer(filepath).find_module('argparse').load_module('argparse')
-else:
-    import argparse
+site_packages = get_virtualenv_site_packages_path()
+argparse = pkgutil.get_importer(site_packages).find_module('argparse').load_module('argparse')
 
 ArgumentParser = argparse.ArgumentParser
 
--- a/commcare-cloud/commcare_cloud/argparse14.py
+++ b/commcare-cloud/commcare_cloud/argparse14.py
@@ -1,0 +1,15 @@
+from __future__ import print_function
+from __future__ import absolute_import
+import pkgutil
+import sys
+import os
+
+for filepath in sys.path:
+    if os.path.isdir(filepath) and 'argparse.py' in os.listdir(filepath):
+        argparse = pkgutil.get_importer(filepath).find_module('argparse').load_module('argparse')
+else:
+    import argparse
+
+ArgumentParser = argparse.ArgumentParser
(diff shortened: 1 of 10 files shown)


commcare-cloud/commcare_cloud/commands/command_base.py
```diff
@@ -0,0 +1,19 @@
...
+abstractclassmethod = abc.abstractmethod
+
+
+@six.add_metaclass(abc.ABCMeta)
+class CommandBase(object):
+    command = None
+    help = None
+
+    @abstractclassmethod
+    def make_parser(cls, parser):
```
millerdev: I think it would be cleaner if these class methods were normal instance methods. I used `classmethod` in [`Lookup` & friends](https://github.com/dimagi/commcare-cloud/blob/dd63996ffe6d58194a25cc42ce650f004c3acb66/commcare-cloud/commcare_cloud/commands/inventory_lookup/inventory_lookup.py#L14) because they needed a reference to the parser, but in hindsight it feels kinda dirty to do it that way because it modifies global state (the class object). I think it would be cleaner if [`commcare_cloud.main`](https://github.com/dimagi/commcare-cloud/blob/dd63996ffe6d58194a25cc42ce650f004c3acb66/commcare-cloud/commcare_cloud/commcare_cloud.py#L42-L47) were written like this:
```py
    commands = {}
    for StandardArg in STANDARD_ARGS:
        # note: make_parser became or is called by __init__ here
        cmd = StandardArg(subparsers.add_parser(
            cmd.command,
            help=cmd.help,
            aliases=cmd.aliases,
        ))
        commands[cmd.command] = cmd
        for alias in cmd.aliases:
            commands[alias] = cmd

    args, unknown_args = parser.parse_known_args()
    commands[args.command].run(args, unknown_args)
```

commcare-cloud/commcare_cloud/commands/command_base.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@six.add_metaclass(abc.ABCMeta)
+class CommandBase(object):
+    command = None
+    help = None
+    aliases = ()
+
+    @abstractclassmethod
+    def make_parser(cls, parser):
```
millerdev: I think it would be cleaner if these class methods were normal instance methods. I used `classmethod` in [`Lookup` & friends](https://github.com/dimagi/commcare-cloud/blob/dd63996ffe6d58194a25cc42ce650f004c3acb66/commcare-cloud/commcare_cloud/commands/inventory_lookup/inventory_lookup.py#L14) because they needed a reference to the parser, but in hindsight it feels kinda dirty to do it that way because it modifies global state (the class object). I think it would be cleaner if [`commcare_cloud.main`](https://github.com/dimagi/commcare-cloud/blob/dd63996ffe6d58194a25cc42ce650f004c3acb66/commcare-cloud/commcare_cloud/commcare_cloud.py#L42-L47) were written like this:
```py
    commands = {}
    for StandardArg in STANDARD_ARGS:
        # note: make_parser became or is called by __init__ here
        cmd = StandardArg(subparsers.add_parser(
            cmd.command,
            help=cmd.help,
            aliases=cmd.aliases,
        ))
        commands[cmd.command] = cmd
        for alias in cmd.aliases:
            commands[alias] = cmd

    args, unknown_args = parser.parse_known_args()
    commands[args.command].run(args, unknown_args)
```

dannyroberts:
Okay, I like that suggestion, and I went ahead with it. Is this more or less what you had in mind?

commcare-cloud/commcare_cloud/argparse14.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import print_function
```
snopoke: A comment explaining this file would be good.

dannyroberts: 👍 also cleaned this up a bit

dannyroberts:
Just checked all main workflows after making a few fixes. @millerdev want to take a final pass?

PR #1309, opened 2018-01-25
Changed ICDS couch username and pass

emord:
What a terrifying change

Not going to be a useful review since this is just a vault file. so fyi @snopoke @mkangia


czue:
:fearful: :pray:

PR #1311, opened 2018-01-26
Promote commcare-cloud README to the top level

dannyroberts:
- move previous README into ansible/
- edit lightly to make more appropriate as a repo README


dannyroberts:
Hey @millerdev, I split up the first commit like you suggested, but it doesn't seem to have done much to the way the final diff is displayed. If nothing else, at least reviewing commit by commit should now be easier.

As for the links issue, that's a good point and something I hadn't really given enough thought too. It would be nice to be able to search all outgoing links on confluence, but unfortunately I seem to only be able to do that one page at a time, so it's hard to get a list of references to fix. It may be better to accept some amount of brokenness with the understanding that I will be doing a sweep of our documentation and fixing the links I do find. I took a pass at our most up-to-date docs on using ansible through commcare-cloud (https://confluence.dimagi.com/display/internal/Managing+CommCare+HQ+Environments) and it looks like the references there were easily fixable. A lot of the other pages that reference or used to reference this (old) readme are now deprecated, particularly https://confluence.dimagi.com/display/internal/Ansible, so I think this is overall not going to be a huge issue.

millerdev:
👍 

> I split up the first commit like you suggested, but it doesn't seem to have done much to the way the final diff is displayed.

Yeah, the overall diff on this PR will be the same, but if you're running commands like `git gui blame ansible/README.md` or  `git gui blame README.md` you can navigate history for each file beyond the moves. I'm not sure it would have been smart enough to do that when both moves were done in a single commit.

PR #1312, opened 2018-01-29
Modifications in elasticesearch snapshot script

NitigyaS:
- Added Flag to change days of backup to keep. 

- Added check snapshot status , Delete function was throwing error since create snapshot was not completed. 
 `elasticsearch.exceptions.TransportError: TransportError(503, u'ConcurrentSnapshotExecutionException[[pna_es_snapshot:pna_es_snapshot_2017_12_23] another snapshot is currently running cannot delete]')`

### DIFF
--- a/ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
+++ b/ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
@@ -1,4 +1,5 @@
 #!/usr/bin/python
+import time
 from elasticsearch import Elasticsearch
 from elasticsearch.client import SnapshotClient
 from datetime import datetime, date
@@ -10,6 +11,10 @@ def get_snapshot_client():
     snapshot_client = SnapshotClient(es)
     return snapshot_client
 
+def check_snapshot(client):
+    today = date.today()
+    stats =  client.status('{{ es_repository_name }}', '{{ es_repository_name }}_{year}_{month}_{day}'.format(year=today.year, month=today.month, day=today.day))
+    return stats['snapshots'][0]['state']
 
 def create_snapshot(client):
     today = date.today()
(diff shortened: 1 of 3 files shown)


ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
```diff
@@ -18,16 +23,31 @@ def create_snapshot(client):
 
 def delete_old_snapshots(client):
     snaps = client.get('{{ es_repository_name }}', '_all')
-    three_days_ago = (datetime.utcnow() - timedelta(days=3)).isoformat()
-    old_snapshots = [s['snapshot'] for s in snaps['snapshots'] if s['start_time'] < three_days_ago]
+    number_of_days_ago = (datetime.utcnow() - timedelta(days={{elastcsearch_backup_days|default(3)}} )).isoformat()
```
NitigyaS: To change the days to keep elasticsearch backup of

ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
```diff
@@ -18,16 +23,31 @@ def create_snapshot(client):
...
+    ld_snapshots = [s['snapshot'] for s in snaps['snapshots'] if s['start_time'] < number_of_days_ago]
     for snap in old_snapshots:
         client.delete('{{ es_repository_name }}', snap)
 
 
 def main():
     client = get_snapshot_client()
+# Create Today's Snapshot
     create_snapshot(client)
-    delete_old_snapshots(client)
```
NitigyaS: Since snapshot completion takes some Time this delete call fails. So added a check to see if snapshot creation completed successfully or not.

ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
```diff
@@ -18,16 +23,31 @@ def create_snapshot(client):
...
+            # Delete Previous Snapshots
+            delete_old_snapshots(client)
+	    exit(0)
+        else:
+            # Wait
+            time.sleep(300)
+        itteration += 1
+    # Exit after 5 tries.
+    print "Snapshot Not deleted, Since Current Snapshot not completed."
+    exit(1)
```
javierwilson: jsut to be clear, so ES doesn't allow us to delete a snapshot as long as there's a snapshot operation running? is that it?

NitigyaS: Yes. refer: https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html

ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
calellowitz: I think now that deletes will start to work we should reconsider how long we set this. ES backups are very lightweight since they only store each document once per version. Since we pretty much never delete documents and only a small percentage are ever changed (the majority are forms which are basically static), extra days do not increase storage costs very much. Maybe a month makes sense as a default, to give us a little more flexibility in restoring.
@javierwilson @NitigyaS

javierwilson: I agree... and S3 is also very cheap. Also, Amazon has a policy that allows you to move old S3 files to Glacier, right @NitigyaS  ?  Perhaps older than 3-days but younger than 30-days should go to glacier?

@calellowitz this code allows us to set elastcsearch_backup_days in public vars, so we can set that there, and have 30 as a default in default common vars (i think it's groups vars all, unless it has been moved)

ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
javierwilson: "Only one snapshot process can be executed in the cluster at any time." :(

javierwilson:
Yes a commnet, perhaps we should use group vars all to store defaults instead of using |default ? I always wondered if this should be our policy since it allows fast glance a the defaults.

ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
calellowitz: @javierwilson 
I am not sure that glacier plan would work. The snapshots are not complete, they are incremental, so pushing them to glacier does not buy us much in terms of storage savings, but doubles our storage usage since everything is in glacier and s3. We might as well just leave them all in s3

calellowitz:
> Yes a commnet, perhaps we should use group vars all to store defaults instead of using |default ? I always wondered if this should be our policy since it allows fast glance a the defaults.

I think is a great idea

PR #1314, opened 2018-01-30
Enabled backup flags

NitigyaS:
Added Backup Flags for local backup.


environments/swiss/public.yml
```diff
@@ -37,10 +37,10 @@ nginx_combined_cert_value: '{{ ssl_secrets.certs.swiss_commcarehq_org }}'
 
 ssh_allow_password: True
 
-backup_blobdb: False
-backup_postgres: False
+backup_blobdb: True
+backup_postgres: dump
```
javierwilson: unless the DB is really small, I think we're better off with 'pllain'

PR #1317, opened 2018-01-30
update icds certs and priv keys for letsencrypt

mkangia:


javierwilson:
I trust you on this one.  I remember you mentioned on a channel you also check-ed this with ansible, so I think we're good.

mkangia:
@javierwilson i could not check it myself but did take help from emord to confirm. He said there were some whitespace changes that were expected and there was some change about challenge directory being removed so we didnt commit the changes.

PR #1318, opened 2018-01-31
Added AWS Access key and secret

NitigyaS:


NitigyaS:
Yeah @calellowitz , I was not sure is swiss servers are really in Switzerland. Could you please confirm if it's in European region.

calellowitz:
@NitigyaS the swiss server is definitely in switzerland. I do not know what there hosting requirements are but we clarify those before choosing an aws center to back up to. Specifically, I have a feeling they do not want that data out of Europe but they also may not want it out of switzerland.

PR #1319, opened 2018-01-31
added local backup plays

NitigyaS:
- es_local_repo variable added to enable or disable local backups. 
- the local backup will occur only if the remote s3 backup is false. 
- the default value of es_local_repo is false, set in group_vars/all.yml

### DIFF
--- a/ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
+++ b/ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
@@ -23,14 +23,32 @@ def create_snapshot(client):
 
 def delete_old_snapshots(client):
     snaps = client.get('{{ es_repository_name }}', '_all')
-    number_of_days_ago = (datetime.utcnow() - timedelta(days={{elastcsearch_backup_days|default(30)}} )).isoformat()
+    number_of_days_ago = (datetime.utcnow() - timedelta(days={{elastcsearch_backup_days|default(3)}} )).isoformat()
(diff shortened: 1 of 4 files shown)


ansible/roles/elasticsearch/templates/create_es_snapshot.py.j2
```diff
@@ -23,14 +23,32 @@ def create_snapshot(client):
...
 
 def main():
     client = get_snapshot_client()
+
+
+
+{% if es_local_repo and not backup_es%}
+    # Create Local Repository if es_local_repo enabled
+    create_snapshot_repo(client)
+{% endif %}
```
dannyroberts: Took me a moment to understand what's going on here—looks like we're templating a python file that then gets called as part of a cron job here: https://github.com/dimagi/commcare-cloud/blob/26753de1fb693f9655fa650a71d6c232744a33c2/ansible/roles/elasticsearch/tasks/main.yml#L142. It feels weird to me to be templating a python file, instead of parameterizing based on sys.args, and then passing in those args in the line in the cron file (which can be templated). But I could also see how passing in a large number of vars via the command line could be _more_ error-prone.

So I think my two cents would be to gather all of the templating at the bottom of this file where main is called, and pass the values into main as parameters. Does that make sense?

Finally, I realize you didn't introduce this pattern in this PR. I just wanted to voice the suggestion—totally fine not to make this PR block on it.

javierwilson: I have the same feeling, like templating may not be the right move, thank you for actually raising the issue, we're also doing it in other scripts.  I think this is something we should change in the near future.

PR #1321, opened 2018-01-31
adding restrict domain setting to pna

jmtroth0:
https://manage.dimagi.com/default.asp?268879


environments/pna/public.yml
```diff
@@ -265,3 +265,4 @@ localsettings:
   XFORMS_PLAYER_URL: "http://127.0.0.1:4444/"
   ENABLE_PRELOGIN_SITE: False
   ENTERPRISE_MODE: True
+  DOMAIN_CREATION_RESTRICTED: True
```
dannyroberts: Are this and `RESTRICT_DOMAIN_CREATION` above supposed to match? Looks like maybe you decided to change the name at some point but missed a spot.

jmtroth0: Ugh, yes

PR #1322, opened 2018-01-31
Fix couch2 datadog

emord:
Fixes the couchdb2 datadog deploy and I created a separate PR to remove the custom datadog check from our repo.

Main issue is [here](https://github.com/dimagi/commcare-cloud/commit/a5276d6fec65bd06f0fa4d29d4b8dfa3ce411326)

@snopoke maybe this will be useful for figuring out hte replication issues we've had?

buddy @nickpell


ansible/roles/datadog/tasks/remove_integrations.yml
```diff
@@ -12,7 +12,7 @@
     path: /etc/dd-agent/checks.d/couchdb2.py
     state: absent
   notify: restart datadog
-  when: inventory_hostname != couchdb2_first_host or not DATADOG_INTEGRATIONS_ENABLED|default(DATADOG_ENABLED)
+  when: True
```
snopoke: could just remove the `when` clause completely

emord: Oops. yeah. I had that around to get the old one removed when I was deploying. I think we can just remove this whole block

PR #1324, opened 2018-01-31
Allow multiple celery queues to be consumed by one process

dannyroberts:
These changes shouldn't have any effect, except on ICDS where `repeat_record_queue` will start being processed on main celery. (It's currently not being processed at all.)


fab/fab/operations/supervisor.py
```diff
@@ -97,7 +98,7 @@ def set_celery_supervisorconf():
         num_workers = params.get('num_workers', 1)
 
         params.update({
-            'queue': queue,
+            'comma_separated_queue_names': comma_separated_queue_names,
```
snopoke: Should `celery_periodic` and `flower` be removed from this list before adding it, or maybe raise an exception if those appear grouped with others?

fab/fab/services/templates/supervisor_celery_worker.conf
```diff
@@ -1,16 +1,16 @@
...
-command=/bin/bash {{ code_current }}/services/supervisor/{{ environment }}_celery_bash_runner{% if celery_params.optimize %}_optimized{% endif %}.sh --queues={{ celery_params.queue }} --events --loglevel=INFO --hostname={{ host_string }}_{{ celery_params.queue }}_{{ celery_params.worker_num }} {% if celery_params.pooling == 'prefork' %} --autoscale={{ celery_params.concurrency }},0 -Ofair --maxtasksperchild={{ celery_params.max_tasks_per_child }}{% endif %}{% if celery_params.pooling == 'gevent' %} -P gevent --concurrency={{ celery_params.concurrency }}{% endif %} --without-gossip
+command=/bin/bash {{ code_current }}/services/supervisor/{{ environment }}_celery_bash_runner{% if celery_params.optimize %}_optimized{% endif %}.sh --queues={{ celery_params.comma_separated_queue_names }} --events --loglevel=INFO --hostname={{ host_string }}_{{ celery_params.queue }}_{{ celery_params.worker_num }} {% if celery_params.pooling == 'prefork' %} --autoscale={{ celery_params.concurrency }},0 -Ofair --maxtasksperchild={{ celery_params.max_tasks_per_child }}{% endif %}{% if celery_params.pooling == 'gevent' %} -P gevent --concurrency={{ celery_params.concurrency }}{% endif %} --without-gossip
```
gcapalbo: the hostname will need to be updated as well

fab/fab/services/templates/supervisor_celery_worker.conf
```diff
@@ -1,16 +1,16 @@
-[program:{{ project }}-{{ environment }}-celery_{{ celery_params.queue }}_{{ celery_params.worker_num }}]
+[program:{{ project }}-{{ environment }}-celery_{{ celery_params.comma_separated_queue_names }}_{{ celery_params.worker_num }}]
```
gcapalbo: probably worth deploying this once to staging to make sure supervisor program names can have commas

fab/fab/services/templates/supervisor_celery_worker.conf
dannyroberts: aha, good catch. I think I missed it in my find-all because it's on the same line.

fab/fab/operations/supervisor.py
dannyroberts: I think it's actually fine for `celery_periodic` to be grouped with the others? (Its presents will also make celery _beat_ get auto-created, but I think it's fine for the queue itself to be processed in the same process as other queues.) Maybe I'm wrong; that was my understanding.

`flower` though, yes, since that's not really a queue at all. It feels like the right time to check this is when checking in code changes, not when fab deploying, so I'm thinking of writing some validation for that file and putting it in a travis test, rather than making it fail here. Thoughts?

fab/fab/services/templates/supervisor_celery_worker.conf
dannyroberts: Good call. I assumed it would, but you know what happens when you assume...

dannyroberts: This worked ✅

fab/fab/operations/supervisor.py
dannyroberts: I implemented the travis test I was suggesting, and am overall pretty happy with it. I opted to have it test up to our current practices (including the mistake `main` "queues" that don't do anything), leaving room to tighten things up in the near future, but at least documenting the current state.

PR #1325, opened 2018-02-01
enable process agent

snopoke:
https://www.datadoghq.com/blog/live-process-monitoring


snopoke:
Deployed to prod and ICDS. Not going to proactively deploy anywhere else for now.

PR #1328, opened 2018-02-02
overwrite default HQ_INSTANCE

nickpell:


emord:
How does HQ_INSTANCE differ from SERVER_ENVIRONMENT

nickpell:
For prod, ```HQ_INSTANCE == 'www'``` and ```SERVER_ENVIRONMENT == 'production'```

@calellowitz Is there any risk that making this update is going to mess with analytics?  See [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/analytics/templates/analytics/initial/kissmetrics.html#L7) and [here](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/analytics/utils.py#L18)

nickpell:
@orangejenny See above comment

orangejenny:
There are two `HQ_INSTANCE` values in `settings.py`: one that was [implemented for ICDS](https://github.com/dimagi/commcare-hq/commit/7dab805e2174a7b402765420d33ef9d32f08baf1) but is no longer used and one that's a property of `ANALYTICS_CONFIG` and is used for analytics.

I think the right thing to do is just to delete the first one from `settings.py`.

nickpell:
thanks @orangejenny: https://github.com/dimagi/commcare-hq/pull/19397

PR #1332, opened 2018-02-05
Added celery queue to PNA.

emord:
@nickpell

### DIFF
--- a/environments/pna/app-processes.yml
+++ b/environments/pna/app-processes.yml
@@ -9,6 +9,8 @@ celery_processes:
   '*':
     main:
       concurrency: 1
+    celery:
+      concurrency: 1
     background_queue:
       concurrency: 1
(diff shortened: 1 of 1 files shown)


emord:
@dannyroberts I think you might be interested in this as you've been looking into celery issues. You were mentioning that some queues were overridden in local settings to have different values. Does this seem like a case of that?

dannyroberts:
On slack (#celery) @gcapalbo was saying that it looks like in `pna` and `l10k` what should be the `celery` queue is mistakenly referred to in `app-processes.yml` as `main`, which results in the `celery` queue just doing not getting run (conversation starts here: https://dimagi.slack.com/archives/C929CP26B/p1517439713000403). (`main` is not a real anything.) He also suggested purging the `celery` queue right before deploying this.

PR #1333, opened 2018-02-05
increase ICDS slow log threshold

snopoke:
Current long queries are around 600ms. We could just turn it off but this seemed like a good first move


snopoke:
Rolled out

PR #1337, opened 2018-02-09
allow specifying CELERY_CASE_UPDATE_RULE_QUEUE by environment

nickpell:
Used with https://github.com/dimagi/commcare-hq/pull/19395

Open to suggestions about the right concurrency for the queue on production.

### DIFF
--- a/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -295,6 +295,10 @@ CELERY_REMINDER_RULE_QUEUE = '{{ localsettings.CELERY_REMINDER_RULE_QUEUE }}'
 CELERY_REMINDER_CASE_UPDATE_QUEUE = '{{ localsettings.CELERY_REMINDER_CASE_UPDATE_QUEUE }}'
 {% endif %}
 
+{% if localsettings.get('CELERY_CASE_UPDATE_RULE_QUEUE') %}
+CELERY_CASE_UPDATE_RULE_QUEUE = '{{ localsettings.CELERY_CASE_UPDATE_RULE_QUEUE }}'
+{% endif %}
+
 {% if localsettings.get('CELERY_REPEAT_RECORD_QUEUE') %}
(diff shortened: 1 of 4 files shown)


ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -295,6 +295,10 @@ CELERY_REMINDER_RULE_QUEUE = '{{ localsettings.CELERY_REMINDER_RULE_QUEUE }}'
 CELERY_REMINDER_CASE_UPDATE_QUEUE = '{{ localsettings.CELERY_REMINDER_CASE_UPDATE_QUEUE }}'
 {% endif %}
 
+{% if localsettings.get('CELERY_CASE_UPDATE_RULE_QUEUE') %}
+CELERY_CASE_UPDATE_RULE_QUEUE = '{{ localsettings.CELERY_CASE_UPDATE_RULE_QUEUE }}'
```
emord: I think that we never want to do this anymore. @dannyroberts

environments/production/app-processes.yml
```diff
@@ -63,6 +63,9 @@ celery_processes:
     saved_exports_queue:
       concurrency: 4
       max_tasks_per_child: 1
+    case_update_rule_queue:
```
emord: We'd want this on other envs right?

nickpell: Not sure.  For other environments the task will use the background_queue.  I'd be interested in seeing if this helps on production before rolling out the change to other environments, but I don't have strong opinions on this.

gcapalbo:
@nickpell if you just want to split this just for prod for now, you can update all other environments to have one worker that consumes from both `background_queue,case_rule_queue`, and only have different worker entries for `background_queue` and `case_rule_queue` in production.

that way, you don't have to worry about overriding in localsettings. [here's](https://github.com/dimagi/commcare-cloud/blob/13353c4a91d710c74261fc444ed7038207499993/environments/staging/app-processes.yml#L9) an example of a worker definition that consumes from multiple queues

nickpell:
thanks @gcapalbo - opened new PR https://github.com/dimagi/commcare-cloud/pull/1338

PR #1344, opened 2018-02-12
 allow referencing host groups in app_processes.yml

snopoke:


fab/fab/tests.py
```diff
@@ -0,0 +1,17 @@
...
+
+@parameterized(get_available_envs())
+def test_app_process_referenced_hosts(env_name):
+    with open(get_default_app_processes_filepath()) as f:
+        app_processes_json = yaml.load(f)
+    with open(get_app_processes_filepath(env_name)) as f:
+        app_processes_json.update(yaml.load(f))
+
+    check_and_translate_hosts(env_name, app_processes_json['celery_processes'])
+    check_and_translate_hosts(env_name, app_processes_json['pillows'])
```
snopoke: I tried adding this to `test_environments.py` but couldn't since the root `fab` folder isn't a python package and making it a package breaks the imports in the fabfile.

dannyroberts: I see, I was gonna ask. I'm fine with this for now, but I think it may be worth putting both `check_and_translate_hosts` and a portion of `test_environments.py` into something like `commcare_cloud.environment_schemas`, and eventually switch all access to `app_processes.yml` to be through classes/functions in that file. Then, the tests will be pretty simple wrappers around the check method(s)/function(s) in that file, and importing them will be easy (because they'll live in the `commcare_cloud` package).

We'll can save that for another day.

PR #1345, opened 2018-02-13
Replace Restart with start

NitigyaS:


calellowitz:
What happens if kafka is already running? If the answer is nothing, I think this will prevent us from using this playbook for configuration changes which we likely do.

NitigyaS:
if Kafka is already running ansible will not run commands unless necessary.

calellowitz:
Right so what happens if we make a change to one of the configuration properties files? and then deploy the kafka playbook. Will those take effect now that we no longer restart kafka?

NitigyaS:
Yes. You are right. I think I need a better solution such as reload or using handlers. Because it will restart Kafka even if nothing has changed. Or can we leave it just like this if restarting Kafka does not cause any issue?

PR #1348, opened 2018-02-13
Continue work on commcare-cloud-bootstrap

dannyroberts:
The first step here isn't documented, but this now essentially works.

Create your spec:
```
cp commcare-cloud-bootstrap/specs/{example,droberts}_spec.yml
# edit your spec to match your own AWS configuration
vim commcare-cloud-bootstrap/specs/droberts_spec.yml
```

Create a provision a new environment named `banana` based on the spec:
```
commcare-cloud-bootstrap privision commcare-cloud-bootstrap/specs/droberts_spec.yml --env banana
```
Look at the EC2 instances associated with this env
```
commcare-cloud-bootstrap show banana
```

Apply our basic user configuration to the new machines (so you can login with your own private key, as all following ansible steps require):

```
commcare-cloud banana bootstrap-users
```
And then deploy the whole stack!

```
commcare-cloud banana deploy-stack --skip-check --quiet
```
There are a few places that this script will halt with an error, but if you start it back up where you left off using `--start-at-task` it'll continue without error. Guessing these are just timing issues (waiting for a service to start up, etc.), but I'll be looking into these next to make this process run smoothly from start to finish.

# restarting EC2 instances

You may want to have a throwaway env live for a couple days without actually paying for a couple days; so it'll be common to shut down the EC2 instances over night, for example. If you've restarted your EC2 instances, they'll have new public IPs, which matters because we ssh into them using their public IPs. So after restarting them, you'll need to update the inventory file to have the new public IPs:

```
commcare-cloud-bootstrap reip banana
```

and then you can run the after-reboot command:
```
commcare-cloud banana after-reboot
```


commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -124,27 +174,83 @@ def ask_aws_for_instances(env, aws_config, count):
...
+    hosts = []
+    for reservation in describe_instances['Reservations']:
+        for instance in reservation['Instances']:
+            hosts.append(
+                Host(public_ip=instance['PublicIpAddress'],
+                     private_ip=instance['PrivateIpAddress']))
+    return hosts
+
+
+def get_inventory_from_file(env):
```
snopoke: Why can't you use the ansible inventory parser?

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -156,31 +262,35 @@ def poll_for_aws_state(env, instance_ids):
...
             for instance in instances
             if instance['InstanceId'] in instance_ids
         }
 
 
 def save_inventory(env, inventory):
-    inventory_file_contents = yaml.safe_dump(inventory.to_json(), default_flow_style=False)
+    j2 = jinja2.Environment(loader=jinja2.FileSystemLoader(os.path.dirname(__file__)))
+    template = j2.get_template('inventory.ini.j2')
+    inventory_file_contents = template.render(inventory=inventory)
```
snopoke: Ansible does support YAML inventories - just wondering if you considered that option.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: I could have, and I considered it, but ansible inventory doesn't distinguish between a "group" that is really just naming a single host and a group that assigns a role; the inventory parser that comes with ansible would just give me back a bunch of groups and the flat list of hosts (by their `inventory_hostname` (here `host.private_ip`), not by the name of their single-host group (here `host.name`). I found parsing directly from the file that organizes these in a specific way to be easier than working backwards from the flattened group: hosts structure ansible would give me.

I definitely did go down that route first, and I wasn't liking it.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Yeah I did consider that, in fact the first versions of this _were_ written in the YAML format. I found it kind of bad to work with, and then there were some ways in which ansible wasn't reading it correctly, possibly because of the `.ini` ending it assumes. Anyway, while that was certainly fixable, I decided to switch to the ini-like syntax. The code for it generating the yaml format was getting worse and worse as I added and changed things, so I opted to refactor the code to work in terms of `Group`s and `Host`s as they're now defined, and then mapping them to and from the file format at the last minute. It's true that even with this refactor, I could be mapping to and from the YAML format, but I like that it's very obvious what's going on when you look at the `inventory.ini.j2` template, rather than having to carefully read the code that assembles it. Of course, that does trade of with the custom parsing code above; but my thought is that it's clear what that function intends to do, even if you don't read every line.

Not sure if these are convincing reasons, but I did like what I ended up with a lot better than what I had before. If the custom parsing or anything else becomes a problem, I'm happy to switch this back to YAML in the future.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -38,73 +49,112 @@ class AwsConfig(jsonobject.JsonObject):
...
+                                   .format(role, allocation.from_))
                 if allocation.from_ in incomplete:
                     continue
-                inventory.all.children[name] = AnsibleInventoryGroup(hosts={
-                    host: {}
-                    for host in inventory.all.children[allocation.from_].hosts.keys()[:allocation.count]
-                })
+                # This is kind of hacky because it does a string sort
+                # on strings containing integers.
+                # Once we have more than 10 it'll start sorting wrong
```
esoergel: I don't think that's true - it splits on `[0-9]+`, so consecutive integers are grouped together:
```python
In [2]: alphanumeric_sort_key('foo123bar')
Out[2]: ['foo', 123, 'bar']
```

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -38,73 +49,112 @@ class AwsConfig(jsonobject.JsonObject):
...
+                host_names = sorted(
+                    inventory.all_groups[allocation.from_].host_names,
+                    key=alphanumeric_sort_key,
+                )[:allocation.count]
```
esoergel: I'm having trouble following this part - might come ping you in person.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -124,27 +174,83 @@ def ask_aws_for_instances(env, aws_config, count):
             '--tag-specifications', 'ResourceType=instance,Tags=[{Key=env,Value=' + env + '}]',
         ]
         aws_response = subprocess.check_output(cmd_parts)
-        with open('ask_aws_for_instances.json', 'w') as f:
```
esoergel: Was this just broken before?

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -38,73 +49,112 @@ class AwsConfig(jsonobject.JsonObject):
...
         instance_ip_addresses = poll_for_aws_state(env, instance_ids)
         if instance_ip_addresses:
             break
 
-    ip_addresses_by_host = dict(zip(all_hosts, instance_ip_addresses.values()))
+    hosts_by_name = {}
+
+    for host, (public_ip, private_ip) in zip(inventory.all_hosts, instance_ip_addresses.values()):
+        host.public_ip = public_ip
+        host.private_ip = private_ip
```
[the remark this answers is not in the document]
esoergel: Also, are these guaranteed to be in the same order?

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -156,31 +262,35 @@ def poll_for_aws_state(env, instance_ids):
...
         f.write(inventory_file_contents)
     print('inventory file saved to {}'.format(inventory_file),
           file=sys.stderr)
 
 
 def copy_default_vars(env, aws_config):
-    template_dir = os.path.join(VARS_DIR, 'dev')
+    vars_dir = VARS_DIR
+    template_dir = os.path.join(vars_dir, '.commcare-cloud-bootstrap')
     new_dir = ENVIRONMENTS_DIR
```
esoergel: What's going on with this guy?

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -189,15 +299,72 @@ def copy_default_vars(env, aws_config):
...
+        inventory = get_inventory_from_file(args.env)
+        update_inventory_public_ips(inventory, new_hosts)
+        save_inventory(args.env, inventory)
+
+
+STANDARD_ARGS = [
+    Provision,
+    Show,
+    Reip,
+]
```
esoergel: This is a cool pattern

ansible/roles/users/tasks/main.yml
```diff
@@ -45,14 +45,9 @@
...
+  when: dev_users.absent
+
 - name: Copy sudoers file
   become: yes
   template: src=sudoers.j2 dest=/etc/sudoers.d/cchq validate='visudo -cf %s' owner=root group=root mode=0440
-
-# ansible used to create this file
-# but for consistency with prod machines, we changed to 'cchq'
-- name: remove cchq_sudoers
-  file: state=absent path=/etc/sudoers.d/cchq_sudoers
```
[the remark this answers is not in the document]
dannyroberts: Yeah this is to clean up stuff from envs we deployed years ago, and the cleanup has long been run.

ansible/vars/.commcare-cloud-bootstrap/private.yml
```diff
@@ -0,0 +1,45 @@
...
+#  STRIPE_PUBLIC_KEY:
+#  SENTRY_PUBLIC_KEY:
+#  SENTRY_PRIVATE_KEY:
+#  SENTRY_PROJECT_ID:
+#  SENTRY_QUERY_URL:
+#  SENTRY_API_KEY:
+  TOUCHFORMS_API_PASSWORD: 'TsyasEivKf4GtbQX'
+  TOUCHFORMS_API_USER: 'touchforms@example.com'
+
+ansible_sudo_pass: NkCl9ul5xtOeZGPx7JvXnaF9
```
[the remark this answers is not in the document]
dannyroberts: These are made up environments not meant to actually house any data or anything. These were randomly generated and then hard coded just here. These environments are really just for us to be able to test (like a travis test or something), so I'm not worried about them using passwords that are publicly known. (Other than the passwords to the AWS account being used, of course!)

ansible/vars/.commcare-cloud-bootstrap/public.yml
```diff
@@ -0,0 +1,318 @@
...
+#  SMS_QUEUE_ENABLED: ''
+#  STAGING_DOMAINS:
+#  STATIC_ROOT:
+#  STRIPE_PRIVATE_KEY: "{{ localsettings_private.STRIPE_PRIVATE_KEY }}"
+#  STRIPE_PUBLIC_KEY: "{{ localsettings_private.STRIPE_PUBLIC_KEY }}"
+  TOUCHFORMS_API_PASSWORD: "{{ localsettings_private.TOUCHFORMS_API_PASSWORD }}"
+  TOUCHFORMS_API_USER: 'touchforms@example.com'
+  XFORMS_PLAYER_URL: "http://{{ groups.touchforms.0 }}:4444"
+  FORMPLAYER_URL: "http://{{ groups.touchforms.0 }}:8181"
+#  ENABLE_DRACONIAN_SECURITY_FEATURES: yes
```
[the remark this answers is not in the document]
dannyroberts: Yeah it was copied from the `dev/` variables, and then modified.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -38,73 +49,112 @@ class AwsConfig(jsonobject.JsonObject):
...
     from_ = jsonobject.StringProperty(name='from')
 
 
-class AnsibleInventory(jsonobject.JsonObject):
-    all = jsonobject.ObjectProperty(lambda: AnsibleInventoryGroup)
+# Inventory
+
+class Inventory(StrictJsonObject):
+    all_hosts = jsonobject.ListProperty(lambda: Host)
+    all_groups = jsonobject.DictProperty(lambda: Group)
```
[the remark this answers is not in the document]
dannyroberts: Yeah, this lets you refer to a class that hasn't been defined yet. (It's defined below.)

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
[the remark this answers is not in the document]
dannyroberts: This is what defines how they go together; `instance_ip_addresses` is just the raw inventory of machines that AWS says are available, and this assigns them to hosts.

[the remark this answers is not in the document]
dannyroberts: About refactoring; I think you're right that this function could be a bit better separated into chunks. I think I'll wait to refactor after merging this.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -38,73 +49,112 @@ class AwsConfig(jsonobject.JsonObject):
...
 
-    for group_name, group in inventory.all.children.items():
-        for host_name, host_vars in group.hosts.items():
-            host_vars['ansible_host'] = str(ip_addresses_by_host[host_name])
+    for i, host_name in enumerate(inventory.all_groups['kafka'].host_names):
+        hosts_by_name[host_name].vars['kafka_broker_id'] = i
+        hosts_by_name[host_name].vars['swap_size'] = '2G'
+
+    for host_name in inventory.all_groups['elasticsearch'].host_names:
+        hosts_by_name[host_name].vars['elasticsearch_node_name'] = host_name
```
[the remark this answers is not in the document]
dannyroberts: That's a good point. I'll do this in a separate PR

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -38,73 +49,112 @@ class AwsConfig(jsonobject.JsonObject):
...
     copy_default_vars(env, spec.aws_config)
 
 
+def alphanumeric_sort_key(key):
+    """
+    Sort the given iterable in the way that humans expect.
+    Thanks to http://stackoverflow.com/a/2669120/240553
+    """
+    import re
+    convert = lambda text: int(text) if text.isdigit() else text
```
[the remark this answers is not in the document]
dannyroberts: Feels like this (and maybe the above block?) should go in a separate update_ip_addresses function.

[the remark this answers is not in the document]
dannyroberts: Yeah, but I didn't have to worry about that for what we're doing here because what I'm sorting is always lowercase.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -7,15 +7,26 @@
...
 
-class Spec(jsonobject.JsonObject):
+class StrictJsonObject(jsonobject.JsonObject):
+    _allow_dynamic_properties = False
+
+
+# Spec
+
+
+class Spec(StrictJsonObject):
```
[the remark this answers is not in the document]
dannyroberts: 👍

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Yeah you're right. I can get rid of this now that I added `alphanumeric_sort_key`. Thanks

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Okay, a comment might be in order here :)

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Yeah, it was. This whole file is something I've just been working on and PRing occasionally, so I must have just left it broken.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Oh and I forgot to say, ansible inventory files as far as i can tell are not actually valid ".ini" files. Python's `ConfigParser` (made for ini files) fails on parsing them.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Hmm, yeah the place where it's used below in the comment doesn't really make sense anymore, and we no longer need to create this dir because it's already there from when we create it when we make the inventory. Also, I'm creating the wrong dir (`/environments` instead of `/environments/<env>`).

dannyroberts: I'll fix this in a followup PR. (Already changed this a bit in some secret new commits I've been working on.)

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Thanks :) I started doing this in `commcare-cloud` and then copied that pattern here.

dannyroberts:
Ethan just gave me a literal thumbs up in person, and added that it had all the weight of a digital thumbs up.

dannyroberts:
@snopoke, still open to hearing your thoughts; still working on some more commits so anything we decide I can add there.

PR #1359, opened 2018-02-14
remove jhoffman from users

nickpell:
https://manage.dimagi.com/default.asp?270453#1458515


emord:
I think this is already done in https://github.com/dimagi/commcare-cloud/pull/1357/

nickpell:
Looks like I got beat too it!

PR #1362, opened 2018-02-14
add celery queue send_report_throttled to production

nickpell:


nickpell:
✅ @sravfeyn

environments/production/app-processes.yml
```diff
@@ -69,6 +69,9 @@ celery_processes:
     background_queue:
       concurrency: 4
       max_tasks_per_child: 1
+    send_report_throttled:
+      concurrency: 2
+      max_tasks_per_child: 1
```
gcapalbo: we might need to keep an eye on memory usage on hqcelery2 since we've added some new workers to it recently. i think this should be ok because these tasks will only be active during a certain time of day and then it would scale down to 0 processes during periods of inactivity, but it may start to get harder to add more workers to this machine

nickpell: 👍 I'll keep an eye out

PR #1365, opened 2018-02-15
add emm-gateway stub files

czue:
I used this to deploy our user accounts to `emm.dimagi.com`. Is unclear to me that it belongs here, but seems like a similar thing had been done for motech and captain, so figured I'd throw it out there and see what the experts had to say.

### DIFF
--- a/environments/emm-gateway/inventory.ini
+++ b/environments/emm-gateway/inventory.ini
@@ -1,0 +1,2 @@
+[emm-gateway]
+emm.dimagi.com
--- a/environments/emm-gateway/public.yml
+++ b/environments/emm-gateway/public.yml
@@ -1,0 +1,2 @@
+SITE_HOST: 'emm.dimagi.com'
+deploy_env: emm-gateway
--- a/environments/emm-gateway/public.yml
+++ b/environments/emm-gateway/public.yml
@@ -1,2 +1,3 @@
 SITE_HOST: 'emm.dimagi.com'
 deploy_env: emm-gateway
+cchq_user: web


czue:
yep, currently just using a (locally) modified version of `update-users` to bootstrap accounts, keys and sudo permissions. was super nice to be able to set up all the accounts with very little work, though also agree it's weird to have this here.

javierwilson:
i wrote this when i deployed the jenkins machines: https://github.com/dimagi/ansible-minimal/blob/master/bootstrap.yml
not elegant, but it did save me a lot of time and allowed me to run roles in the repo formerly known as commcarehq-ansible.
i just added an example of how to use it.

czue:
@javierwilson that's quite a cool approach and I like the idea of not polluting the ansible repo with random one-off non-hq servers.

i've already achieved what I wanted to achieve with the `emm-gateway` project, and so I'm inclined to not invest too much energy into hardening this approach. so does it sound right that we should:

1. close this PR
2. (eventually) standardize a non-HQ server bootstrap process that expands on the work done in https://github.com/dimagi/ansible-minimal?

point 2 feels like something that could be done at any point in the future - possibly ideally the next time someone on the @dimagi/devops team encounters this use case again.

dannyroberts:
@javierwilson cool I hadn't seen that. I like that approach. Another thing that might be worth doing is letting you set <env> to any filepath so you can specify an inventory/vars outside just what's in the `environments/` dir.

snopoke:
I guess another way to make better use of our roles would be to publish them to ansible galaxy and just include them as requirements.

czue:
(closed since I think this will be handled in other ways)

PR #1370, opened 2018-02-16
rotate production cloudant key

dannyroberts:
https://manage.dimagi.com/default.asp?269141


dannyroberts:
This has now been deployed to production FYI

PR #1373, opened 2018-02-16
allow specifying a different environment set

dannyroberts:
If you want to use environments outside this repo's `environments/` (eventually a main use-case), you can now at least do it super explicitly with

```bash
$ COMMCARE_CLOUD_ENVIRONMENTS='../outside-environments' commcare-cloud ...
```

Let me know what you think of this simple approach.


czue:
I dig it. :+1:

PR #1375, opened 2018-02-19
add jekyll config for doc site

dannyroberts:
this should get github pages working


esoergel:
This is cool.  Curious about the decision to use github pages over readthedocs - I guess it gives us more control over the output?  Could you see us moving the HQ docs to github pages as well?  Do you know if there's any support for something like sphinx, or would we just store any built files in the repo?

PR #1377, opened 2018-02-19
 generate known_hosts and allow commcare-cloud to use it

javierwilson:
add-ssh-keys.yml should be played for each environment when we add hosts and commit the changes known_hosts, in futre plays if available commcare-cloud will use it.


ansible/add-ssh-keys.yml
```diff
@@ -0,0 +1,54 @@
...
+    known_hosts:
+      name: "{{ item.item }}"
+      key: "{{ item.stdout }}"
+      state: "present"
+      path: "{{ ssh_known_hosts_file }}"
+    with_items: "{{ ssh_known_host_results.results }}"
+    tags:
+      - ssh
+
+  - name: For each host, ssh-copy-id my ssh public keys to the host
```
dannyroberts: I think we don't step, right? Since we handle ssh keys in a central place already? Maybe I'm misunderstanding what this step does.

javierwilson: yeah, my bad, forgot to simply delete it from this gist

esoergel:
This is awesome, thanks, Javier!

millerdev:
Wow, this is great!

PR #1383, opened 2018-02-20
Added data device

NitigyaS:
This directive was not present for hqstandby0. I wonder how it sat up correctly.

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -124,6 +124,7 @@ hqstandby1.internal-va.commcarehq.org
 [hqstandby1:vars]
 hot_standby_master=hqdb1.internal-va.commcarehq.org
 replication_slot=standby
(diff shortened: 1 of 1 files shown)


emord:
> Should we add this for hqstandby0 too? Or is that a moot point now?

Curious about any follow up to that question. If it isn't added, then will it correctly mount if there's a reboot? (there was a similar issue on the proxy recently)

NitigyaS:
Yeah. I must add it just to avoid a reboot mishap.

javierwilson:
I think this is only used when first formatting and encrypting and adding the mount point to fstab.  So reboot won't affect. Proxy issue is different, That mount was not added to fstab, I did that manually, also working on a PR that addes "extra_mount_points" per host to do custom stuff like add a disk to /var/log, etc.

PR #1389, opened 2018-02-21
always add '--become' if specified

snopoke:
`--become-user` does not imply `become`

buddies @millerdev @esoergel


dannyroberts:
Thanks, great! I think I misunderstood the behavior in exactly the way you described when I first wrote this :)

PR #1393, opened 2018-02-22
remove small inaccuracy

dannyroberts:


dannyroberts:
going to self merge since it's a tiny thign

PR #1396, opened 2018-02-23
ICDS: Reorganize celery workers

gcapalbo:
I noticed that some workers are being denied the appropriate amount of cpu usage they need to finish their tasks in a timely manner. This change makes celery0 the machine that holds workers whose tasks are less cpu intensive, so that they should always be able to access cpu when they need it.

@snopoke @emord

### DIFF
--- a/environments/icds-new/app-processes.yml
+++ b/environments/icds-new/app-processes.yml
@@ -8,6 +8,11 @@ celery_processes:
   None:
     repeat_record_queue:
   'celery0':
+    # The celery workers which have less cpu intensive tasks
+    # are meant to go on this machine. Try to avoid putting
+    # ucr_indicator_queue, reminder_case_update_queue, or
+    # pillow_retry_queue workers here so that the workers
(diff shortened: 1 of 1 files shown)


snopoke:
great, thanks. Build failure unrelated:
```
[ERROR]: failed to download the file: Failed to validate the SSL certificate
for github.com:443. Make sure your managed systems have a valid CA certificate
installed. You can use validate_certs=False if you do not need to confirm the
servers identity but this is unsafe and not recommended. Paths checked for this
platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs,
/usr/share/ca-certificates/cacert.org, /etc/ansible. The exception msg was:
("bad handshake: Error([('SSL routines', 'ssl3_read_bytes', 'tlsv1 alert
protocol version')],)",).
```

emord:
probably due to https://github.com/blog/2507-weak-cryptographic-standards-removed ?

PR #1402, opened 2018-02-26
Datadog integration

NitigyaS:


ansible/plugins/callback/datadog_callback.yml
```diff
@@ -0,0 +1 @@
+api_key: "{{DATADOG_API_KEY}}"
```
snopoke: how does this get populated?

ansible/plugins/callback/datadog_callback.py
```diff
@@ -0,0 +1,315 @@
...
+        # Only send an event when the task has changed on the host
+        if res.get('changed'):
+            event_text, module_name_tag = self.format_result(res)
+            self.send_task_event(
+                'Ansible task changed on "{0}"'.format(host),
+                alert_type='success',
+                text=event_text,
+                tags=[module_name_tag],
+                host=host,
+            )
```
snopoke: these task events don't seem very useful since they don't include any info about the task at all

ansible/plugins/callback/datadog_callback.yml
NitigyaS: At present, I am manually doing it. Will use vault in future.

ansible/plugins/callback/datadog_callback.py
NitigyaS: Datadog -> events -> Search (Ansible) Give lot of information . (Who ran which playbook on which server and what changed)

snopoke: Maybe I'm not looking in the right place but I don't see that much info:
image

NitigyaS: Yes, you are right... That's the same. Sorry,  I think I got confused with some error that I saw on my terminal. 
screen shot 2018-02-26 at 4 07 11 pm

snopoke: It would be possible to add more info in if we wanted to. There are plenty of examples here: https://github.com/ansible/ansible/tree/devel/lib/ansible/plugins/callback

I think of most use would be:
* date time
* user running playbook
* playbook that was run
* other options (limit, tags etc)
* any changes that were made (optional)

NitigyaS:
screen shot 2018-02-28 at 11 56 09 am

Issue Pending : 
1) Data is not being pushed to datadog hostname since Ansible has no way to know the hostname used by datadog agent on the server. 
2) Taking the API_KEY from vault.

ansible/plugins/callback/datadog_callback.py
```diff
@@ -0,0 +1,326 @@
...
+        if getattr(self, 'play', None):
+            if tags is None:
+                tags = []
+            tags.append('play:{0}'.format(self.play.name))
+        self._send_event(
+            title,
+            alert_type=alert_type,
+            text=text,
+            tags=tags,
+            host=host,
```
snopoke: should this `host` also have the same treatment as the others? 

Also for a few others below

NitigyaS: All the functions used send_event and send_metrics, So I have modified only those two.

dannyroberts: Offlined with @NitigyaS. I just misunderstood the previous comment.

ansible/plugins/callback/datadog_callback.py
```diff
@@ -0,0 +1,329 @@
...
+
+    def _get_hostname(self, host):
+        return self.hostvars[str(host)]['hostname']
+
+    def _generate_datadog_hostname(self, host=None):
+        try:
+            if host is not None:
+                hostname = self._get_hostname(host)
+                host = str(hostname) + "." +self.variables['internal_domain_name']
+        except Exception as e:
```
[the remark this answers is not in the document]
snopoke: @NitigyaS this comment still applies. Catching `Exception` is a bit of an antipattern since it can mask exceptions that you might really care about. As @dannyroberts says it would be better to catch the actual exception that could get raised.

PR #1405, opened 2018-02-27
use git@ instead of https to clone commcare-cloud repo

dannyroberts:
seems to work better for me on my current network in india


snopoke:
Will this work for people without a github account / github keys?

dannyroberts:
Oh. I guess not. I'll close this out. Annoying that from the DSI office the https uri doesn't work 🤷‍♂️

PR #1408, opened 2018-02-27
formplayer JAR on AWS

snopoke:
@wpride @javierwilson


javierwilson:
@snopoke Since we're downloading a JAR file, we don't need the extra steps, also we don't really need archive_dbs.s anymore to remove since we manage this with cron files, I think.
We only need the unzip for staging if we keep using a zip  file there.

javierwilson:
Closing in favor of https://github.com/dimagi/commcare-cloud/pull/1410

PR #1410, opened 2018-02-27
using formplayer JAR on AWS

javierwilson:
not fully tested but this should replace https://github.com/dimagi/commcare-cloud/pull/1408/


dannyroberts:
Hey @javierwilson I think unless you're more up to date than me about this that we do still need `archive_dbs.sh`. On the other PR (https://github.com/dimagi/commcare-cloud/pull/1408#issuecomment-368987988) you said "we don't really need archive_dbs.s anymore to remove since we manage this with cron files, I think". My understanding is that you're correct that we manage this with cron files; however, what the line in the cron file runs is the very `archive_dbs.sh` file you're removing: https://github.com/dimagi/commcare-cloud/blob/6645fddff47459e8590c788ba69eb91fd0c7a820/ansible/roles/formplayer/tasks/main.yml#L25. In other words, my understanding is that this file is still in use, and that nothing has changed that would make it no longer part of the process. Let me know if I'm missing a change that's happened that makes this file irrelevant.

javierwilson:
Thanks for pointing that out! We should deploy using ansible, no need to include it in the formplayer release.  I'll take care of it.

snopoke:
Still need to remove `archive_dbs.sh` from the formplayer repo

PR #1411, opened 2018-02-28
update to Ansible 2.4.3

dannyroberts:
No idea if this will fix the

```
 [ERROR]: failed to download the file: Failed to validate the SSL certificate
for github.com:443. Make sure your managed systems have a valid CA certificate
installed. You can use validate_certs=False if you do not need to confirm the
servers identity but this is unsafe and not recommended. Paths checked for this
platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs,
/usr/share/ca-certificates/cacert.org, /etc/ansible. The exception msg was:
("bad handshake: Error([('SSL routines', 'ssl3_read_bytes', 'tlsv1 alert
protocol version')],)",).
```

error, but it's just a patch release so it "can't" hurt.

### DIFF
--- a/ansible/requirements.txt
+++ b/ansible/requirements.txt
@@ -1,4 +1,4 @@
-ansible==2.4.1
+ansible==2.4.3
 netaddr
 dnspython
 passlib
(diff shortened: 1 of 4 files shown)


javierwilson:
i merged master to my PR, it also passed travis tests now :)

dannyroberts:
Great! Maybe if @emord was right about https://github.com/blog/2507-weak-cryptographic-standards-removed, ansible had been using some outdated encryption protocol for HTTPS by default in the previous version or something. 🤷‍♂️

PR #1414, opened 2018-02-28
test init.sh script in travis

dannyroberts:
install using `source control/init.sh` rather than duplicating just the important bits


snopoke:
One other thing. The script has a user input prompt with a 30s timeout. Maybe worth doing that during build.

snopoke:
https://github.com/dimagi/commcare-cloud/pull/1456

PR #1416, opened 2018-02-28
support additional mount points

javierwilson:
In some clusters you may need to add an additional disk, specially if there's no support for LVM and need to mount it, currently, this process is manual, creating potential issues like the one when the proxy machine was rebooted and /mnt/storage not mounted.  This provides a fix for that, Ideally, we should migrate to LVM. @dannyroberts

### DIFF
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -129,4 +129,6 @@
   tags: cron
 
 - include: mount.yml
-  tags: mount
+  tags:
+    - mount
+    - after-reboot
--- a/ansible/roles/common/tasks/main.yml
+++ b/ansible/roles/common/tasks/main.yml
@@ -124,5 +124,9 @@
   tags:
     - antivirus
     - after-reboot
-- include: cron.yml 
+
+- include: cron.yml
   tags: cron
+
+- include: mount.yml
+  tags: mount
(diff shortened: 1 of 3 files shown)


emord:
@dannyroberts I think that this is for the rackspace issue from a month or so ago. Look in your email for subject: "Retro: proxy0 restart caused nginx downtime"

snopoke:
Will this get run with the `after-reboot` tag?

javierwilson:
@snopoke it does now, thanks for reminding me.

ansible/roles/common/tasks/mount.yml
```diff
@@ -0,0 +1,12 @@
...
+  mount:
+    src: "{{ item.partition }}"
+    path: "{{ item.dest }}"
+    fstype: "{{ item.fstype }}"
+    state: mounted
+  with_items: "{{ mount_points }}"
+  when: mount_points is defined
+  tags:
+    - mount
+    - after-reboot
```
snopoke: are these tags necessary since the tags are re-applied in the include

ansible/roles/common/tasks/main.yml
```diff
@@ -124,5 +124,11 @@
   tags:
     - antivirus
     - after-reboot
-- include: cron.yml 
+
+- include: cron.yml
   tags: cron
+
+- include: mount.yml
```
snopoke: Sorry for post merge comment but `include` is deprecated. You should use `import_tasks`

PR #1419, opened 2018-03-01
make commcare-cloud-bootstrap more fab-ready

dannyroberts:
Work on this is still on-going, but this is in a stable state and it'd be nice to get the ancillary changes and fixes I've made here in.


ansible/vars/.commcare-cloud-bootstrap/private.yml
```diff
@@ -22,6 +22,7 @@ localsettings_private:
   DROPBOX_SECRET: ''
   EMAIL_LOGIN: ''
   EMAIL_PASSWORD: ''
+  FORMPLAYER_INTERNAL_AUTH_KEY: 'QkGxkCZ2cfT20gI901l8bxzNgspibBnx741me7Soa38BsjiPmmrAEuLw4SM5LBOQG8PQo'
```
millerdev: not clear what this file is used for. does it matter that a "private" file (including an ansible sudo password, which was not added in this PR) is actually public?

dannyroberts: This file is analogous to (and originally copied from) https://github.com/dimagi/commcare-cloud/blob/master/ansible/vars/dev/dev_private.yml. These passwords are just some preset fillers to make it possible to bootstrap a throwaway env. It's intentionally not encrypted with vault since there aren't any actual secrets in here, just mock "secrets". Is that what you were asking?

dannyroberts: This file is analogous to (and originally copied from) https://github.com/dimagi/commcare-cloud/blob/master/ansible/vars/dev/dev_private.yml. These passwords are just some preset fillers to make it possible to bootstrap a throwaway env. It's intentionally not encrypted with vault since there aren't any actual secrets in here, just mock "secrets". Is that what you were asking?

millerdev: Presumably it's mostly safe to use publicly available "secrets" as in dev_private.yml because the places where those secrets are used are private: they're on a local machine on a private network (unless you're connected to coffee shop WIFI and have ports open to the public). However, presumably this file is used to bootstrap servers on the cloud, right? Maybe they're always behind a VPN, and therefore effectively private? I'm not sure. If the servers are publicly accessible it might make sense to make a utility to generate a set of local randomized secrets that can then be used to bootstrap the cloud rather than publishing these secrets on github.

dannyroberts: I guess I'm just not really seeing the issue here. Maybe I should clarify that knowledge of these passwords in no way grants ssh access or anything to these machines, and that database access is over private IPs only anyway. And then also on top of that, access to anything on these machines gives an attacker very little, since there's no real data and nothing to steal or maliciously destroy.

We should certainly revisit this if and when this script is used for anything remotely production-like, at which point your suggestion of autogenerating secrets would be a good tack to take.

millerdev: I wasn't concerned that these secrets were granting SSH access to servers. Rather, I was thinking of them in the same way as a superuser password for a production database, for example, even though there is usually no (known) publicly accessible way to use such a password to gain direct access to the database without additional credentials such as VPN passwords and SSH keys. The reason details like this are kept private is in case some vulnerability exploit circumvents those extra levels of security.

Also, I'm not concerned that you are using these for production services right now. However, this repository will eventually be used by third parties to bootstrap their own servers. Do you think it's possible an uneducated third party following the path of least resistance may inadvertently use these "secrets" to initialize their production database? If that's even remotely likely then we should design the system to make that harder than doing the right thing.

All of these suggestions are more in the *food for thought* category than immediately actionable.

PR #1422, opened 2018-03-02
Add location for data dumps

mkangia:
Just adds the nginx location for data dumps for enikshay on enikshay and india.

All files are in place and this is tested.

### DIFF
--- a/environments/softlayer/public.yml
+++ b/environments/softlayer/public.yml
@@ -267,3 +267,9 @@ localsettings:
   XFORMS_PLAYER_URL: "http://{{ groups.touchforms.0 }}:4444/"
   ENIKSHAY_PRIVATE_API_USERS: {'BI': 'ppia-bi21', 'GU': 'ppia-gu21', 'MH': 'ppia-mh14'}
   ENIKSHAY_PRIVATE_API_PASSWORD: "{{ localsettings_private.ENIKSHAY_PRIVATE_API_PASSWORD }}"
+
+site_locations:
+  - name: /data_dumps/export/enikshay/public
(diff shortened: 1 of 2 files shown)


environments/softlayer/public.yml
```diff
@@ -267,3 +267,9 @@ localsettings:
   XFORMS_PLAYER_URL: "http://{{ groups.touchforms.0 }}:4444/"
   ENIKSHAY_PRIVATE_API_USERS: {'BI': 'ppia-bi21', 'GU': 'ppia-gu21', 'MH': 'ppia-mh14'}
   ENIKSHAY_PRIVATE_API_PASSWORD: "{{ localsettings_private.ENIKSHAY_PRIVATE_API_PASSWORD }}"
+
+site_locations:
+  - name: /data_dumps/export/enikshay/public
+    try_files: "$uri $uri/index.html $uri/ =404"
+    auth_basic: '"Restricted Content"'
+    auth_basic_user_file: '/etc/nginx/.htpasswd_enikshay_public'
```
kaapstorm: not `.htpasswd_softlayer_public`?

kaapstorm: (I might just be missing context. "data dumps for enikshay on enikshay and india" does suggest that we keep enikshay data dumps on softlayer, in which case `...enikshay...` seems right.)

mkangia: the data dumps have been generated on enikshay.in but also kept on softlayer so that they are available if needed after enikshay.in is closed. so `enikshay` here is for the project and not for the server/environment.

PR #1415, opened 2018-02-28
environment schemas

dannyroberts:
Move checks that are currently only in tests into `commcare_cloud.environment` so they can be run from anywhere.

A bit of a followup to https://github.com/dimagi/commcare-cloud/pull/1344

PR #1423, opened 2018-03-03
Fix proxy definition

emord:
@dannyroberts I think this was introduced https://github.com/dimagi/commcare-cloud/pull/1415 but there were a lot of changes and I don't have an understanding of why this broke. Basically `inventory.groups['all'].hosts` is an empty list.

I think the real/ideal fix is not defining supervisor configs during deploy. I have an idea that I'll look to PR in the next week

@czue @nickpell should fix the issue we were talking about the past couple of days


emord:
succesfully deployed so self merging

PR #1424, opened 2018-03-05
Define supervisor services in ansible

emord:
@dannyroberts This is what I was talking about in #1423 I think that this makes sense but curious to hear your opinion.

This feels like a better place for this type of code. In my mind fab should only deploy changes to the code, not changes to the services we're running. This change also means that we can see diffs of services when they are changed.

I also think that these definitions could likely be cleaned up, but I was going for an MVP where I didn't change any functionality or the app processes yaml. There also needs to be a task to copy the old services to the new directory like we have for localsettings

buddy @czue


ansible/roles/commcarehq/templates/supervisor_pillowtop.conf.j2
```diff
@@ -0,0 +1,23 @@
+{% for pillow_name, params in pillows.get(ansible_host).items() %}
```
dannyroberts: Still have to remove the corresponding file from fab, right?

ansible/roles/commcarehq/tasks/main.yml
```diff
@@ -105,3 +105,16 @@
   tags:
     - localsettings
     - hq-localsettings
+
+- include_vars: "../../../../environments/{{ deploy_env }}/app-processes.yml"
```
dannyroberts: I see why this was the easiest way to do this; at the same time I can see a couple drawbacks at least long-term. One is that this only works for our envs, but not if someone's passing in `COMMCARE_CLOUD_ENVIRONMENTS` to use their own envs dir. The other is that if one of these templates relies on `environmental-defaults/app-processes.yml`, that won't get included here.

I think the longer-term solution is for `commcare-cloud` to drop a freshly generated `.combined-vars.yml` file in the env right before calling ansible, and include that file in addition to (or instead of) `public.yml`. Taking that extra step will will give us total control over how ansible variables get set and decouples that from the config format, letting us define configs based on what's the most sensical for the maintainer of the environment to write rather than what ansible needs in order to be able to run.

I think if this works (i.e. if none of this requires the environmental-defaults) then I'm happy to merge this in now, and then I'll implement that that variables change in a separate PR.

snopoke:
big :+1: for moving the services to ansible, this will also speed up deploy by quite a bit

czue:
:clap: this is great. crazy we were still doing this on deploy/.

looks like @dannyroberts did a thorough pass but let me know if you want a second review!

ansible/roles/commcarehq/templates/supervisor_pillowtop.conf.j2
emord: https://github.com/dimagi/commcare-cloud/pull/1424/commits/ab446ad1486210dd5d5ca401b8ef7907caaf4980

ansible/roles/commcarehq/tasks/main.yml
emord: Totally agree this isn't correct long term. I cringed a bit when figuring out the correct number of `../`, but wanted to get something out there so you guys could review/we could come to consensus

Neither the pillowtop nor the form feed supervisor require that defaults file, so this should be fine for this PR.

It sounds like you have a much better grasp of where you're taking this, so I think that I'll leave that for you to do.

emord:
> How do you want to do the rollout? Do you want to port each service type over to ansible one by one and PR them individually, or are you going to do them all in this branch and then merge all at once?

I was originally going to PR them all at the same time and change the fab file to copy the entire configuration folder, but I think that this is in a state that should work, so it seems like we can do this piecewise. happy to default to your preference though

fab/fab/operations/supervisor.py
```diff
@@ -147,19 +147,15 @@ def show_periodic_server_whitelist_message_and_abort(env):
...
-    # on the 'deploy' machine, db!
-    # So you need to explicitly test to see if all_hosts is empty.
-    if env.all_hosts and _check_in_roles(ROLES_PILLOWTOP):
-        _rebuild_supervisor_conf_file(
-            'make_supervisor_pillowtop_conf',
-            'supervisor_pillowtop.conf',
-            {'pillow_env_configs': get_pillow_env_config()}
-        )
+def copy_pillowtop_supervisorconf():
+    services_dir = "/services/supervisor/"
```
dannyroberts: nit: does this make for a double slash when combined with the lines below?

dannyroberts: It occurs to me that we may simply want to host these config files in `/home/cchq/www/<env>/services` or something like that, since the coupling with the code release doesn't really make sense anymore. Do you agree, and do you want to do that now (as you move these to ansible) or in a later refactor?

emord: Totally agree. I was a little hesitant to add that to this PR as that also means we need to change the supervisor configuration to look in a couple of places. Wanted to keep the parts moving around here small since we've had some issues deploying recently

emord:
this breaks on non pillow machines at the moment so not merging

emord:
Went through a lot of iteration. I've gotten most of the supervisor configs out. all except for celery, django and formplayer as those require variables from enviromental defaults. Can work on deploying this tomorrow

dannyroberts:
Awesome. And since this seems like kind of the next step for you, I went ahead and pushed a branch implementing pretty simply what I'd been envisioning for loading app-processes.yml info into ansible in the correct format: https://github.com/dimagi/commcare-cloud/compare/dmr/generated_yml?expand=1. Feel free to just cherry-pick it into this PR. You may need to add other things to that file in the same vein, but what's implemented there I think should work as-is.

ansible/deploy_commcarehq.yml
```diff
@@ -20,3 +20,27 @@
     - role: tmpreaper
       tags: ['commcarehq', 'celery_hq', 'tmpreaper', 'cron']
       tmpreaper_time: '2d'
+
+- name: Pillows Supervisor Config
+  hosts: pillowtop
+  tasks:
+    - include_tasks: roles/commcarehq/tasks/pillowtop.yml
```
snopoke: should probably be `import_tasks`. Then you also wouldn't need to put the `tags` in the tasks files

snopoke: Unless that breaks the `when` checks on the tasks

fab/fab/services/templates/supervisor_elasticsearch.conf
```diff
@@ -1,9 +0,0 @@
-[program:{{ project }}-{{ environment }}-elasticsearch]
-command=/usr/bin/autossh -M0 -v -L 127.0.0.1:9200:127.0.0.1:9200 {{ es_endpoint }}
-user={{ sudo_user }}
-autostart=true
-autorestart=true
-stdout_logfile={{ log_dir }}/elasticsearch.log
-redirect_stderr=true
-stderr_logfile={{ log_dir }}/elasticsearch_error.log
```
snopoke: was this was just unused?

fab/fab/services/templates/supervisor_handle_survey_actions.conf
```diff
@@ -1,12 +0,0 @@
...
-command={{ virtualenv_current }}/bin/python manage.py handle_survey_actions
-user={{ sudo_user }}
-numprocs=1
-autostart=true
-autorestart=true
-stdout_logfile={{ log_dir }}/handle_survey_actions.log
-redirect_stderr=true
-stderr_logfile={{ log_dir }}/handle_survey_actions.error.log
-startsecs=0
-stopwaitsecs=10
```
snopoke: also just unused?

emord: This wasn't unused, but a lot of these were using the exact same template with minor variations. I combined them into a single template https://github.com/dimagi/commcare-cloud/pull/1424/commits/479f0b045c7753fdd4ea59015624b6052ffb18a7#diff-0bce1acd60cfc963a489f95f234323eeR18

fab/fab/services/templates/supervisor_elasticsearch.conf
emord: Yep

emord:
Ran into some issues deploying this on friday as we reference servers in different ways in app processes across environments. Next step is to change those to use one of `ansible_host` or `ansible_hostname` instead those or '*'

emord:
Sorry for the delay here. I tested this on every env (except PNA, how do we get ansible to respect the non standard port?)

> Looking good. I know this might complicate things but we'll need a way to remove service files from hosts as well e.g. if we move it from one to another.

I believe https://github.com/dimagi/commcare-cloud/pull/1424/commits/a36ffd98048f5cdd2e0659ca778706683bf775db should work for this. Thanks for bringing it up. I almost forgot

@dannyroberts I switched over to `inventory_hostname` as you suggested and everything worked. Should we change all instances of `ansible_hostname` to that?

> The other is that if one of these templates relies on environmental-defaults/app-processes.yml, that won't get included here.

This moves all of the services that don't require this file out of the deploy. I haven't touched any of the ones that do require variables from that. However I don't think that restriction exists anymore after https://github.com/dimagi/commcare-cloud/pull/1424/commits/df59b107c58098c4ce4946da815a188cf43a84ea 

I think the deploy process will look like this:

1. `cchq <env> deploy-stack.yml --tags=supervisor` (adds new configuation file)
2. `cchq <env> deploy-stack.yml --tags=services` (adds services files)
3. `fab <env> deploy` (deploy will stop/update/start supervisor)

dannyroberts:
Does it work to combine the two tags into a single commmand:

`cchq <env> deploy-stack --tags=supervisor,services`

and can we alias that to

`cchq <env> update-app-processes`

and can that also (somehow) reload/restart supervisord (unless you pass in `--no-reload` or something)?

Thoughts for another PR maybe

dannyroberts:
@emord Yeah there aren't a lot of places where `ansible_hostname` is used in our code. It refers to the first part of the `.`-separated FQDN in theory (though possibly unreliable in practice—e.g. not sure what it does with IP addresses), so yeah I can't think of a good use for it. Only places I see it are ansible/roles/ecryptfs/tasks/main.yml and `DEPLOY_MACHINE_NAME`, and I don't really understand what each of those is trying to do, but I also don't get why it'd use `ansible_hostname`.

dannyroberts:
> how do we get ansible to respect the non standard port?
You can use `ansible_port` to get the port, which we do in a couple places. Inventory parser figures this out if you use `:` e.g. `196.207.230.171:1337`:
```
cchq pna run-shell-command all "echo {{ inventory_hostname }} {{ ansible_port }}"

196.207.230.171 | SUCCESS | rc=0 >>
196.207.230.171 1337
```

emord:
@dannyroberts unless you would like to complete this, I'm happy to fully deploy it on Wednesday/Thursday when I get back. I'd like for at least one of us to be around boston time to make sure deploys go smoothly and backstop any issues that pop up

emord:
- [ ] staging
- [ ] www
- [ ] india
- [ ] enikshay
- [ ] pna
- [ ] swiss
- [ ] ICDS

(side note: holy crap this is a lot of envs....)

emord:
Will deploy to enikshay tomorrow

ansible/group_vars/all.yml
```diff
@@ -3,9 +3,11 @@ www_dir: "{{ cchq_home }}/www"
 www_home: "{{ www_dir }}/{{ deploy_env }}"
 code_home: "{{ www_home }}/current"
 log_home: "{{ www_home }}/log"
-service_home: "{{ code_home }}/services"
+legacy_service_home: "{{ code_home }}/services"
+service_home: "{{ www_home }}/services"
 virtualenv_home: "{{ code_home }}/python_env"
 encrypted_root: "/opt/data"
+project: "commcare-hq"
```
mkangia: Curious to why we have a variable to store "commcare-hq"? Is it even needed? It's hard coded at quite some places in the codebase. I don't think we will call this anything apart from `commcare-hq` ever. This adds a bit of confusion by doing this by two different ways.

ansible/roles/commcarehq/tasks/celery.yml
```diff
@@ -0,0 +1,73 @@
...
+      management_command: run_reminder_queue
+    - dest_name: supervisor_queue_schedule_instances.conf
+      process_name: queue_schedule_instances
+      management_command: queue_schedule_instances
+    - dest_name: supervisor_handle_survey_actions.conf
+      process_name: handle_survey_actions
+      management_command: handle_survey_actions
+  tags:
+    - services
+  register: celery_reminder_queue_files
```
mkangia: It's may be my incorrect understanding, but isn't `reminder_queue`, `sms_queue` etc that are not set via `set_celery_supervisorconf` in `set_supervisor_config` not related to celery and just management commands that need to be run periodically? Just confused by the naming here with `celery_`

emord: Yeha they aren't technically celery queues, which I didn't realize until I was writing htis PR. Historically they've been specified under "celery" in the app-processes.yml file though, so I kept that convention here

PR #1429, opened 2018-03-05
fix encrypted root override

snopoke:


environments/softlayer/inventory.ini
```diff
@@ -15,7 +15,7 @@
 
 [db0:vars]
 datadog_integration_cloudant=true
-encrypted_root=/opt/data
```
dannyroberts: Can you explain why this wasn't working? Does it have to do with the subtle `kafka_log_dir` change below?

snopoke: It was just a variable precedence issue. Variables loaded from the vars files override anything that's in the inventory so the options were:

1. remove the var from `public.yml` and only specify it in the inventory file
2. change the var name and check if it's already defined in the vars file 

Option 2 seemed like the better option.

For reference: https://docs.ansible.com/ansible/latest/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

environments/softlayer/public.yml
```diff
@@ -124,10 +124,10 @@ KSPLICE_ACTIVE: yes
 AMQP_HOST: "{{ groups.rabbitmq.0 }}"
 AMQP_NAME: commcarehq
 
-kafka_broker_id: 0
-kafka_log_dir: '/opt/data/kafka'
+encrypted_root: "{{ encrypted_root_override|default('/opt/data/ecrypt') }}"
```
millerdev: Is that really the recommended default? I thought we had annoying issues with envs where `/opt/data/ecrypt` was used instead of `/opt/data`.

dannyroberts: @millerdev this is for the softlayer env, where all but a few machines use `/opt/data/ecrypt`. Now understanding this PR a bit better though, I have some thoughts about how I'd do this differently to highlight that `/opt/data` is really the preferred one—even though in this particular env its also the less common one.

dannyroberts:
@snopoke thanks for the explanation, that makes sense to me now.

If I were to do essentially this same PR, I would have done sort of the inverse of this. Rather than specifying `encrypted_root`/`encrypted_root_override` on `db0` only and using the `public.yml` to have it default to `/opt/data/ecrypt`, I would leave it out of `public.yml` entirely, leave it empty for `db0` and set it to `/opt/data/ecrypt` on _all other_ machines in the inventory individually. Even though that seems more repetitive and verbose, I think it's actually more explicit. It's saying "this machine's encrypted drive is different from the default, this machine's encrypted drive is different from the default, this machine's encrypted drive is different from the default, ...", rather than making the normal one seem like the odd one out.

snopoke:
@dannyroberts applied the change here: https://github.com/dimagi/commcare-cloud/pull/1437

Note that I'm not sure what you meant by `leave it out of public.yml entirely, leave it empty for db0`. You can't do both of those. Either it must be defined in `public.yml` or it must be defined at the host level.

PR #1437, opened 2018-03-06
remove 'encrypted_root' var from public.yml files

snopoke:
Changes based on feedback from https://github.com/dimagi/commcare-cloud/pull/1429#issuecomment-370682229

Validated this config with the following playbook:
```
- name: test
  hosts:
    - postgresql
    - couchdb
    - couchdb2
    - redis
    - elasticsearch
    - rabbitmq
    - zookeeper
    - riakcs
    - formplayer
    - pg_standby
  tasks:
    - shell: mount | grep ecryptfs | cut -d" " -f 1
      register: mount
      check_mode: no
    - assert: { that: "encrypted_root == mount.stdout"}
    - debug: msg="{{encrypted_root}} == {{mount.stdout}}"
```

### DIFF
--- a/environments/softlayer/public.yml
+++ b/environments/softlayer/public.yml
@@ -124,7 +124,7 @@ KSPLICE_ACTIVE: yes
 AMQP_HOST: "{{ groups.rabbitmq.0 }}"
 AMQP_NAME: commcarehq
 
-encrypted_root: "{{ encrypted_root_override|default('/opt/data/ecrypt') }}"
+encrypted_root: "{{ encrypted_root_override|default('/opt/data') }}"
 
 kafka_broker_id: 0
 kafka_log_dir: "{{ encrypted_root }}/kafka"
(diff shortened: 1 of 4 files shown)


environments/softlayer/public.yml
```diff
@@ -124,7 +124,7 @@ KSPLICE_ACTIVE: yes
 AMQP_HOST: "{{ groups.rabbitmq.0 }}"
 AMQP_NAME: commcarehq
 
-encrypted_root: "{{ encrypted_root_override|default('/opt/data/ecrypt') }}"
+encrypted_root: "{{ encrypted_root_override|default('/opt/data') }}"
```
dannyroberts: This ties into your issue comment here: https://github.com/dimagi/commcare-cloud/pull/1429#issuecomment-370766306

I was suggesting (though I didn't state it clearly) that we drop `encrypted_root_override`, define `encrypted_root` on each host where it's not `/opt/data`, _not_ mention it in `public.yml`, and then let the default defined in `ansible/group_vars/all.yml` (`/opt/data`) shine through for `db0`.

snopoke: ok, that sounds fair. I wasn't aware of the value in `all.yml`: 26ca09d

PR #1438, opened 2018-03-06
use environment object

dannyroberts:
also adds validates fab-settings.yml config against a schema (both realtime and in tests)

### DIFF
--- a/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
+++ b/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
@@ -13,9 +13,8 @@ import re
 import jinja2
 import yaml
 import jsonobject
-from commcare_cloud.environment.paths import get_inventory_filepath, \
-    get_public_vars_filepath, get_vault_vars_filepath, ENVIRONMENTS_DIR, \
-    get_app_processes_filepath
+
+from commcare_cloud.environment.main import get_environment
 
 TEMPLATE_DIR = os.path.join(os.path.dirname(__file__), 'environment')
 j2 = jinja2.Environment(loader=jinja2.FileSystemLoader(TEMPLATE_DIR))
@@ -86,16 +85,17 @@ class Group(StrictJsonObject):
     vars = jsonobject.DictProperty()
 
 
-def provision_machines(spec, env=None):
-    if env is None:
-        env = u'hq-{}'.format(
+def provision_machines(spec, env_name=None):
+    if env_name is None:
+        env_name = u'hq-{}'.format(
             ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(7))
         )
-    inventory = bootstrap_inventory(spec, env)
-    instance_ids = ask_aws_for_instances(env, spec.aws_config, len(inventory.all_hosts))
+    environment = get_environment()
+    inventory = bootstrap_inventory(spec, env_name)
+    instance_ids = ask_aws_for_instances(env_name, spec.aws_config, len(inventory.all_hosts))
 
     while True:
-        instance_ip_addresses = poll_for_aws_state(env, instance_ids)
+        instance_ip_addresses = poll_for_aws_state(env_name, instance_ids)
         if instance_ip_addresses:
             break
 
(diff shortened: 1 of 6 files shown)


commcare-cloud/commcare_cloud/environment/schemas/fab_settings.py
```diff
@@ -0,0 +1,24 @@
+import jsonobject
+
+GitUriProperty = jsonobject.StringProperty
+TimezoneProperty = jsonobject.StringProperty
```
snopoke: what's the purpose of overriding these?

commcare-cloud/commcare_cloud/environment/schemas/fab_settings.py
```diff
@@ -0,0 +1,24 @@
...
+    _allow_dynamic_properties = False
+    sudo_user = jsonobject.StringProperty()
+    default_branch = jsonobject.StringProperty()
+    home = jsonobject.StringProperty()
+    project = jsonobject.StringProperty()
+    code_repo = GitUriProperty()
+    timing_log = jsonobject.StringProperty()
+    keepalive = jsonobject.IntegerProperty()
+    ignore_kafka_checkpoint_warning = jsonobject.BooleanProperty()
+    acceptable_maintenance_window = jsonobject.ObjectProperty(lambda: AcceptableMaintenanceWindow)
```
snopoke: what's the difference between this and `ObjectProperty(AcceptableMaintenanceWindow)`? I've seen both in use.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -86,16 +85,17 @@ class Group(StrictJsonObject):
...
-    if env is None:
-        env = u'hq-{}'.format(
+def provision_machines(spec, env_name=None):
+    if env_name is None:
+        env_name = u'hq-{}'.format(
             ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(7))
         )
-    inventory = bootstrap_inventory(spec, env)
-    instance_ids = ask_aws_for_instances(env, spec.aws_config, len(inventory.all_hosts))
+    environment = get_environment()
```
snopoke: shouldn't this have `env_name` as a parameter?

commcare-cloud/commcare_cloud/environment/paths.py
```diff
@@ -17,51 +16,35 @@ def __init__(self, env_name):
...
-    return os.path.join(REPO_BASE, 'environments', env_name, 'app-processes.yml')
+        return os.path.join(REPO_BASE, 'environmental-defaults', 'app-processes.yml')
 
+    @memoized_property
+    def fab_settings_yml(self):
+        return os.path.join(ENVIRONMENTS_DIR, self.env_name, 'fab-settings.yml')
 
-def get_default_app_processes_filepath():
-    return os.path.join(REPO_BASE, 'environmental-defaults', 'app-processes.yml')
+    @memoized_property
```
snopoke: do these really need to be memoized? They don't seem to be very expensive

commcare-cloud/commcare_cloud/environment/schemas/fab_settings.py
dannyroberts: This lets you define `AcceptableMaintenanceWindow` below rather than above this class. (At the time this class is being created `AcceptableMaintenanceWindow`, defined below, does not yet exist. `jsonobject` allows you to use this convention to work around that, emulating Django models, I believe.)

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Sure should!

commcare-cloud/commcare_cloud/environment/paths.py
dannyroberts: They definitely don't need to be memoized for performance reasons. It also doesn't _hurt_ performance, so I'm opting them to use these decorators to signal to the reader that these are lazily computed immutable properties. I like to signal things like that where I can, just to narrow down what the reader has to think about when debugging later. (I.e. they never have to consider the thought that maybe the value of `REPO_BASE` changed between calls or something.) If that's not how you read it though, maybe it's ineffective at serving that purpose.

commcare-cloud/commcare_cloud/environment/schemas/fab_settings.py
dannyroberts: Signals to the reader what values types are supposed to be here. Makes relatively explicit, for example, the validation code someone would write if they wanted to tighten up the schema even more. I've done this in other schemas and usually been happy I have when I'm revisiting the code to remember what various properties are/look like.

dannyroberts:
@snopoke thanks for the feedback. Responded, let me know what you think

commcare-cloud/commcare_cloud/environment/schemas/fab_settings.py
snopoke: 👍

commcare-cloud/commcare_cloud/environment/paths.py
snopoke: that makes sense but isn't the way I think about `memoized`. What about aliasing it to `lazy_immutable_property`?

PR #1439, opened 2018-03-07
Changed configuration for netvault user

NitigyaS:
**Changes**
1) Changing auth-method to `trust` for Netvault user. 
2) Added directive `hot_standby_feedback = on`
     I am concerned if it could have a negative impact on our Master nodes when a backup is happening.
     According to https://www.postgresql.org/docs/9.6/static/runtime-config-replication.html `but can cause database bloat on the primary for some workloads`

### DIFF
--- a/ansible/roles/postgresql/templates/pg_hba.conf.j2
+++ b/ansible/roles/postgresql/templates/pg_hba.conf.j2
@@ -7,8 +7,8 @@ local   all             postgres                                peer
 
 {% if postgres_users.get('netvault') %}
 # Netvault required configuration
-host    all             netvault        127.0.0.1/32            md5
-local   all             netvault                                md5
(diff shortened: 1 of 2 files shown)


javierwilson:
I would like a dev to have an opinion here, or Simon since he's dealt with how we use the standby to distribute load. We should be sure this won't bite us.  Perhaps try to test it first?

NitigyaS:
I have enable it on pg_ucrstandby yesterday It resulted in the successful backup. But I am not sure if it caused anything to the master.

snopoke:
I'm happy to try this out - we should just monitor the master nodes. When do the backups happen?

NitigyaS:
I am not sure when the backups will happen. maybe we can tell NIC to take them in off Peak hours. 
If its okay can we do a deploy today itself so that I can push it ASAP.

PR #1440, opened 2018-03-07
getting rid of 'include'

javierwilson:
include is deprecated and "confusing" according to docs.
i have not tested riakcs playbook, but hanlders are supposed to be allowed to import tasks, so it should work.

### DIFF
--- a/ansible/deploy_cleanup.yml
+++ b/ansible/deploy_cleanup.yml
@@ -5,7 +5,7 @@
   vars_files:
     - roles/shared_dir/vars/main.yml
   tasks:
-    - include: roles/shared_dir/tasks/cleanup_client.yml
+    - import_tasks: roles/shared_dir/tasks/cleanup_client.yml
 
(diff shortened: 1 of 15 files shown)


dannyroberts:
Nice Do you have a rule for when you chose `import_tasks` vs. `include_tasks`?

snopoke:
One big caveat to what I said above is that **For static imports, the parent task options will be copied to all child tasks contained within the import.**. This applies to things like `when` and `tags`.

dannyroberts:
Thanks for the summary. I'd looked over some of those docs trying to understand it and this was a lot more helpful.

Sounds like as a rule of thumb: "use `include_tasks` with `when` or when the filename itself is dynamic, otherwise use `import_tasks`. If you really know what you're doing and want to use `import_tasks` with `when` for the less intuitive (but reliable) behavior, leave a comment."

Sound right?

PR #1441, opened 2018-03-07
shut up if api_key is not set

NitigyaS:


ansible/plugins/callback/datadog_callback.py
```diff
@@ -254,6 +254,8 @@ def v2_playbook_on_play_start(self, play):
             return
         try:
             api_key = self.variables['secrets']['DATADOG_API_KEY']
+            if api_key == '':
+                self.disabled = True
```
dannyroberts: I think `DATADOG_API_KEY` is often `None` (rather than `''`), so maybe just check for `if not api_key:`?

ansible/plugins/callback/datadog_callback.py
```diff
@@ -254,6 +254,8 @@ def v2_playbook_on_play_start(self, play):
             return
         try:
             api_key = self.variables['secrets']['DATADOG_API_KEY']
+            if api_key == '' or not api_key:
```
snopoke: can be shortened to just `if not api_key:`

Issue #1243, opened 2018-01-05
Add command for managing service state


dannyroberts:
```bash
commcare-cloud softlayer service riakcs start
```

would be equivalent to

```bash
commcare-cloud softlayer run-module riakcs service 'name=riak state=started' -b
commcare-cloud softlayer run-module riakcs service 'name=riak-cs state=started' -b
commcare-cloud softlayer run-module stanchion service 'name=stanchion state=started' -b
```

Could also have an option to only do one specific service,

```bash
commcare-cloud softlayer services riakcs start --only=riak
```

or on one specific machine(Note: This is not being added. One can pass --limit to achieve this.)

```bash
commcare-cloud softlayer services riakcs start --host=riak0
```

To cover -
- [x] proxy
- [x] riakcs
- [x] celery
- [x] pillowtop
- [x] couchdb2
- [x] redis
- [x] elasticsearch
- [x] webworkers
- [x] postgresql -> includes pgbouncer
- [x] kafka -> includes zookeeper
- [x] rabbitmq
- [x] touchforms -> formplayer
- [x] formplayer -> formplayer-spring
- [x] pg_standby -> includes pgbouncer

PR #1452, opened 2018-03-07
[WIP] Mk add manage services

mkangia:
[Issue](https://github.com/dimagi/commcare-cloud/issues/1243)


commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,84 @@ def run(self, args, unknown_args):
         args.playbook = 'deploy_stack.yml'
         unknown_args += ('--tags=users',)
         AnsiblePlaybook(self.parser).run(args, unknown_args)
+
+
+class ManageServices(AnsibleOptions, _AnsiblePlaybookAlias):
```
dannyroberts: I generally try to name the class the same thing as the command, so by that convention this ought to be called just `Service`.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,84 @@ def run(self, args, unknown_args):
...
+    """
+    example usages
+    1. To restart riak and stanchion only for riakcs
+       only option can be skipped to restart all services which are a part of riakcs
+       i.e riak, riak-cs and stanchion in that order
+        commcare-cloud staging service riakcs restart --only riak stanchion riak-cs
+    2. To start services under proxy i.e nginx
+       You can still pass ansible options which are
+       --user, --become, --become_user
+        commcare-cloud staging service proxy start -b
```
dannyroberts: I would push for _not_ allowing `--user`, `--become`, `--become_user`, etc. or at least not requiring them for proper functioning; rather I think this command itself should know which user to run each command under and pass in `--become`, etc., accordingly. (I think in every case it's just `root`, so it would mean always passing in `--become`.)

You can do that with `args.become = True` before the call to `RunAnsibleModule`. In addition if you take this suggetion the way I'm imaging, you won't need the `AnsibleOptions` change.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,84 @@ def run(self, args, unknown_args):
...
+            '--only',
+            nargs='+',
+            help=(
+                "Specific services to act on for the inventory group. "
+                "Can mention multiple separated with empty spaces. "
+                "Example Usage: --only riak stanchion"
+            )
+        )
+
+    def for_proxy(self, args, unknown_args):
```
dannyroberts: I'd give these methods a verb phrase name, maybe something like `run_for_proxy`

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,84 @@ def run(self, args, unknown_args):
...
+        action = args.action
+        if action == 'start':
+            state = 'started'
+        elif action == 'restart':
+            state = 'restarted'
+        elif action == 'stop':
+            state = 'stopped'
```
dannyroberts: As a next step, can add a `status` action that uses run-shell-command to run `service status` directly.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,84 @@ def run(self, args, unknown_args):
...
+        args.module_args = "name=nginx state=%s" % state
+        RunAnsibleModule(self.parser).run(
+            args,
+            unknown_args
+        )
+
+    def for_riakcs(self, args, unknown_args):
+        tags = []
+        action = args.action
+        args.playbook = "restart_riakcs.yml"
```
dannyroberts: I'd still push for these all to be in a directory together (`ansible/service_playbooks/restart_riakcs.yml` or something).

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,84 @@ def run(self, args, unknown_args):
...
         AnsiblePlaybook(self.parser).run(args, unknown_args)
+
+
+class ManageServices(AnsibleOptions, _AnsiblePlaybookAlias):
+    """
+    example usages
+    1. To restart riak and stanchion only for riakcs
+       only option can be skipped to restart all services which are a part of riakcs
+       i.e riak, riak-cs and stanchion in that order
+        commcare-cloud staging service riakcs restart --only riak stanchion riak-cs
```
dannyroberts: nit: most of these commands follow the convention that would have us do `--only riak,stanchion,riak-cs` i.e. comma-separated rather than space-separated. Ansible `--limit` for example uses that convention.

mkangia:
@dannyroberts Master Danny.. made the changes on the PR.
Two things:
1. Blocker: is there a way to run multiple shell commands. For checking status for multiple services at once i need to run shell command multiple times. I can possibly to try to merge them into a single shell command but it wont work since we might need different shell command for different hosts. The only way i could run multiple shell command was by commenting [this](https://github.com/dimagi/commcare-cloud/blob/master/commcare-cloud/commcare_cloud/commands/ansible/run_module.py#L146) since that causes exit after first shell command.

2. One ToDo pending but not blocking.

mkangia:
@dannyroberts Covered more services. Updated the issue to mark the services covered.
The ones left are all supervisor managed services.. So would probably need some ground work to set things up for them.

dannyroberts:
> 1. Blocker: is there a way to run multiple shell commands using RunShellCommand? For checking status for multiple services at once i need to run shell command multiple times. The only way i could run multiple shell command was by commenting [this](https://github.com/dimagi/commcare-cloud/blob/master/commcare-cloud/commcare_cloud/commands/ansible/run_module.py#L146) since that causes exit after first shell command.
I can possibly try to merge them into a single shell command but it wont work since we might need different shell command for different hosts.

I'd be happy for you to refactor the control code so that `run` returns an integer and the control code (`main()` function) calls `exit` on the return value of `run` `or 0`. Then if you change all `run` commands that call `exit` to return the exit code instead, you give other commands more flexibility. You can do this refactor in a separate PR if you'd like.

dannyroberts:
> 2. In terms of adding the option to limit to hosts by using --host. Can we just let the user pass limit instead. Looks like ansible already is checking that.
For example: 
`commcare-cloud staging service riakcs stop --only=stanchion --limit=proxy0`
Gives an error
`ERROR! Specified hosts and/or --limit does not match any hosts`
But this
`commcare-cloud staging service riakcs stop --only=stanchion --limit=proxy0,stanchion`
works.

Yeah, that sounds great! Both less code to manage and more consistent with other commands. Good suggestion

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,198 @@ def run(self, args, unknown_args):
...
+                self.ANSIBLE_MODULE_FOR_SERVICE.get(service, service),
+                state)
+            RunAnsibleModule(self.parser).run(
+                args,
+                unknown_args
+            )
+
+    def run_for_services(self, service_group, args):
+        if args.only:
+            return args.only.split(',')
```
dannyroberts: Would be good to have this check that it's a subset of `self.SERVICES[service_group]`. Otherwise you could pass in any random word here and it'd cause pretty unintuitive behavior or at least probably give a much worse error message.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,198 @@ def run(self, args, unknown_args):
...
+                    tags.append("%s_%s" % (action, service))
+            if tags:
+                unknown_args.append('--tags=%s' % ','.join(tags),)
+        unknown_args.extend(['--extra-vars', "desired_state=%s desired_action=%s" % (state, action)])
+        # ToDo: use this with when in the playbook instead of tags
+        # currently its running riak even when just riak-cs is ran
+        # but skips riak when running for only stanchion
+        # as of now it looks like
+        # --extra-vars 'desired_services=['"'"'riak-cs'"'"', '"'"'stanchion'"'"']'
+        unknown_args.extend(['--extra-vars', "desired_services=%s" % run_for_services])
```
dannyroberts: I can't see where the `desired_services` is used anywhere in the playbooks. Was this from an older commit you've since changed away from / can you just remove these lines at this point?

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,198 @@ def run(self, args, unknown_args):
...
+                    tags.append("%s_%s" % (action, service))
+            if tags:
+                unknown_args.append('--tags=%s' % ','.join(tags),)
+        unknown_args.extend(['--extra-vars', "desired_state=%s desired_action=%s" % (state, action)])
+        # ToDo: use this with when in the playbook instead of tags
+        unknown_args.extend(['--extra-vars', "desired_services=%s" % run_for_services])
+        AnsiblePlaybook(self.parser).run(args, unknown_args)
+
+    def run_for_pg_standby(self, args, unknown_args):
+        self.run_ansible_module_for_service_group('pg_standby', args, unknown_args, inventory_group="pg_standby")
```
dannyroberts: This is the only place we pass in the `inventory_group` arg? If so, I don't see what it gets us, since I think if we left it `None`, the default behavior would arrive at the same answer. If I'm not missing something, I'd suggest removing this argument from the method.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,198 @@ def run(self, args, unknown_args):
...
+        elif service_group == "stanchion":
+            if not args.only:
+                args.only = "stanchion"
+            self.run_for_riakcs(args, unknown_args)
+        elif service_group == "es":
+            self.run_for_es(args, unknown_args)
+        elif service_group == "kafka":
+            self.run_for_kafka(args, unknown_args)
+        elif service_group == "pg_standby":
+            self.run_for_pg_standby(args, unknown_args)
```
dannyroberts: Other than "es" (which uses a different implementation) these other "special" cases look really similar to the default case. What's preventing them from using the `run_ansible_module_for_service_group` method like the others?

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
dannyroberts: I see, you do that check below, in `ensure_permitted_only_options`. There might be a way to reorganize this so the control flow is clearer, but we can leave that for another day.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -229,3 +235,198 @@ def run(self, args, unknown_args):
...
+            args.module = 'service'
+            args.module_args = "name=%s state=%s" % (
+                self.ANSIBLE_MODULE_FOR_SERVICE.get(service, service),
+                state)
+            RunAnsibleModule(self.parser).run(
+                args,
+                unknown_args
+            )
+
+    def run_for_services(self, service_group, args):
```
dannyroberts: This should be called something like `get_subservices`; the current name is misleading because this method doesn't really "run" anything.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: @dannyroberts i could add choices when calling `add_arguments` and that would work if we do space separated arguments like `--only riak stanch` but then had to do this manual check later since we moved to comma-separated. Open to better ideas.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: @dannyroberts i have this ToDo above this which shares where this can be used. I want to replace the playbook to use `desired_services` instead of relying on tags. Currently facing a challenge with how its making this as a command line argument so need to look into that.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: @dannyroberts 
Since `pg_standby` also checks for `postgresql` it would end up checking for inventory group for service `postgresql` when calling `get_inventory_group_for_service`. As of now it probably would work without this `inventory_group ` argument since `get_inventory_group_for_service` will fall back to service group which is `pg_standby`, but i still chose to do this to make it an explicit override from the normal behaviour.
Fine to remove if you think its not needed or wont be needed.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: @dannyroberts 
You're right, they are similar. I can extract `run_ansible_playbook_for_service_group` for riakcs, kafka and stanchion. Can't reuse `run_ansible_module_for_service_group` since that runs a module.

mkangia:
@dannyroberts thanks for the review Danny.. made required changes, replied to comments and also took a shot at allowing multiple shell commands by implementing some version of what you suggested in [commit](https://github.com/dimagi/commcare-cloud/pull/1452/commits/2822212787b1c52feb1bccad04b076d589c5e812)

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
dannyroberts: Ah, okay. Well I think what you would do in the command line is `... --extra-vars '{"desired_services": ["riak-cs", "stanchion"]}' ...`—i.e. if you need any part of it to be JSON and not just a string value then I believe you have to pass in the entire arg as a JSON dict.

If you don't end up using this in this PR, I suggest removing it and cleaning things up first, and then refactoring in a later PR. Leaving code in from an idea that didn't end up panning out is in my opinion (and I think generally recognized as) an anti-pattern: if it's not working now, then it's unlikely that having it in the code for later when/if you do decide to use it will be more helpful than starting from scratch looking at your previous code as a reference, and it's all saved in the commit history. So I'd still suggest removing this now, even if you think you're going to come back to it soon.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
dannyroberts: OK. I'm not sure I totally understand, but if you have a good reason in mind for keeping it, I'm fine with that. If we end up not needing it, it's easy to remove later as well.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
dannyroberts: Ah, right, sounds like I stated the wrong function, but sounds like you understood what I was trying to get at. Glad you were able to simplify a bit!

dannyroberts:
@mkangia This is looking really awesome! Thanks for the quick turnaround times on my different comments and suggestions. Made some more comments on the now-outdated changes, but I think this is the only one where I'm potentially asking for a change: https://github.com/dimagi/commcare-cloud/pull/1452#discussion_r174119819.

This is getting super close! Looking forward to getting access to these commands!

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: @dannyroberts i could not get this done.. struggled due to my inexperience with ansible. Will get to this later after learning a bit more about ansible. 
removed this section.

mkangia:
@dannyroberts this is ready now..

PR #1454, opened 2018-03-08
tmp'ly remove production hqdjango0.internal-va.commcarehq.org

dannyroberts:
This PR is just to document the current state of things on production. We'll either actually remove or swap back in by the end of the day.


dannyroberts:
@snopoke on slack#server-fires
> reboot worked
> added hqdjango0 back to proxy

PR #1455, opened 2018-03-08
added pgsynclog in inventory

NitigyaS:


emord:
Will we have a standby as well?

NitigyaS:
Yes. But it's not provisioned yet.

PR #1457, opened 2018-03-08
Two more tweaks for a smooth deploy-stack

dannyroberts:
Getting closer to a smooth `deploy-stack`; in the last run I still had the following issues:

- What seemed like a short network connectivity error made me have to restart the process (at task `jython : Install Jython`)
- When I ran fab the way it failed made me notice supervisor wasn't properly set up, and running with `--tags=supervisor` fixed it—odd error both because that doesn't make a lot of sense and because this didn't happen in a previous run (yesterday)
- after trying again, I was then unable to connect to the celery machine. This has happened a couple times before, weirdly always to the celery machine

will keep investigating, but wanted to get these two fixes in anyway.

### DIFF
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -105,11 +105,3 @@
   pip:
     name: virtualenv-clone
     version: 0.2.6
-
-- name: Install npm (only) a second time
-  npm: name="{{ item.name }}" version="{{ item.version }}" state=present global=yes
-  become: yes
-  with_items:
-    - {name: 'npm', version: '3.6.0'}
-  tags:
-    - npm
--- a/ansible/roles/common_installs/tasks/main.yml
+++ b/ansible/roles/common_installs/tasks/main.yml
@@ -78,9 +78,10 @@
   npm: name="{{ item.name }}" version="{{ item.version }}" state=present global=yes
   become: yes
   with_items:
+    - {name: 'n', version: '1.3.0'}
+    # must come after 'n', otherwise the wrong version of npm ends up installed on first pass
     - {name: 'npm', version: '3.6.0'}
     - {name: 'less', version: '2.7.2'}
-    - {name: 'n', version: '1.3.0'}
     - {name: 'bower', version: '1.5.3'}
     - {name: 'uglify-js', version: '2.6.1'}
   tags:
@@ -104,3 +105,11 @@
   pip:
     name: virtualenv-clone
     version: 0.2.6
+
+- name: Install npm (only) a second time
+  npm: name="{{ item.name }}" version="{{ item.version }}" state=present global=yes
+  become: yes
+  with_items:
+    - {name: 'npm', version: '3.6.0'}
+  tags:
+    - npm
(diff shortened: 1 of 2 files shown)


ansible/roles/postgresql/tasks/main.yml
```diff
@@ -90,12 +90,16 @@
...
+- name: Restart postgresql
   become: true
-  service: name=postgresql state=started args="{{ postgresql_version }}"
+  service: name=postgresql state=restarted args="{{ postgresql_version }}"
   tags: after-reboot
   register: start_postgresql
 
+- name: Reload postgresql
+  become: true
+  service: name=postgresql state=reloaded args="{{ postgresql_version }}"
```
snopoke: why would you need a reload immediately after a restart?

ansible/roles/common_installs/tasks/main.yml
```diff
@@ -104,3 +105,11 @@
   pip:
     name: virtualenv-clone
     version: 0.2.6
+
+- name: Install npm (only) a second time
```
snopoke: comment would be useful explaining why

dannyroberts: Oops, I meant to either do the above or this. When I tested just now, I thought I was testing without this task. Let me run again without it and see if this task is necessary.

ansible/roles/postgresql/tasks/main.yml
dannyroberts: I don't know, but my distinct impression after some testing (and the reason I added this) is that the following commands (`Create PostgreSQL users` specifically) breaks without this. I may have conflated this change with changing `started` to `restarted`, so let me just check again without this.

dannyroberts: OK, that's definitely what happened; made two changes at once without realizing it and took the wrong one as the operative change. Thanks for the sharp eyes!

ansible/roles/postgresql/tasks/main.yml
```diff
@@ -90,9 +90,9 @@
   tags: postgresql_conf
   notify: Reload postgres
 
-- name: start postgresql
```
emord: Why did this change from start to restart? I think that this will get included everytime we run a postgres command which means that postgres may restart even if that's unecessary.

Also if it's intended to restart should we change the register below to `restart_postgresql` ?

dannyroberts: I had to change it to restart because on the first pass, I believe by this point it was already started, so `state=started` did nothing; however since the configuration files just changed above, it's important that postgresql is restarted, or else the following command fails.

I see what you're saying about this forcing a restart. Maybe if I just make sure that it gets reloaded/restarted if any of the files change, this will still work and not have the same negative effect. (The `notify` technique above doesn't cut it because that only does it after the playbook is over. Maybe I should just remove those as well then.)

snopoke: You can force the handlers to run with:
```
- meta: flush_handlers
```

dannyroberts: Oh, that's great! Thanks, Simon.

dannyroberts: https://github.com/dimagi/commcare-cloud/pull/1467

PR #1460, opened 2018-03-09
fix newrelic-java role (for formplayer)

dannyroberts:
noticed this when running deploy_formplayer.yml


snopoke:
Not as part of this PR but it would be nice to get rid of this too

dannyroberts:
@snopoke you're saying we should replace these tasks with ones that simply uninstall newrelic-java if it's there (etc.)?

snopoke:
Yea, I'm pretty sure nobody is looking at the data in NewRelic so we should kill it. Added it here: https://trello.com/c/KMeYJ7Lt/43-new-relic

PR #1464, opened 2018-03-09
Add four workers for the dashboard reports

emord:
https://github.com/dimagi/commcare-hq/pull/19738

@proteusvacuum @dannyroberts


gcapalbo:
celery 3 and 4 usually run high on cpu. i've noticed anecdotally that celery 5, 6, and 7 are usually a bit lower.  might be better to put these workers there

emord:
@gcapalbo those are tapped out on memory though. Going to go with this for now and will be talking with dashboard team to make this task much smaller

PR #1465, opened 2018-03-09
Include queue for softlayer

emord:
follow up for https://github.com/dimagi/commcare-cloud/pull/1464

Didn't realize that we statically check this. pretty cool @dannyroberts 

also adding for softlayer so that QA works


emord:
(self merging since it would block future deploy of icds)

PR #1467, opened 2018-03-12
fix postgresql restart behavior on --tags=postgresql

dannyroberts:
use the pattern of flushing notify handlers rather than explicitly restarting

### DIFF
--- a/ansible/roles/postgresql/tasks/main.yml
+++ b/ansible/roles/postgresql/tasks/main.yml
@@ -90,16 +90,7 @@
   tags: postgresql_conf
   notify: Reload postgres
 
-- name: Restart postgresql
-  become: true
-  service: name=postgresql state=restarted args="{{ postgresql_version }}"
-  tags: after-reboot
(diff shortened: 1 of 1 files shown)


ansible/roles/postgresql/tasks/main.yml
```diff
@@ -90,16 +90,7 @@
...
-  become: true
-  service: name=postgresql state=restarted args="{{ postgresql_version }}"
-  tags: after-reboot
-  register: start_postgresql
-
-- name: wait after restarting postgres
-  when: start_postgresql.changed
-  wait_for:
-    port: '{{ postgresql_port }}'
+- meta: flush_handlers
```
snopoke: should you keep the wait?

dannyroberts: I tested it and I didn't have to, so I figured I'd remove and add back if it's ever a problem. I originally added it because I misunderstood what the problem was and thought it would help (it didn't).

PR #1469, opened 2018-03-13
Added Error Handling

NitigyaS:


ansible/roles/pg_backup/templates/plain/create_postgres_dump.sh.j2
```diff
@@ -24,18 +24,27 @@ ARGS="$ARGS | gzip > $DIRECTORY.gz"
...
-EXITCODE=$?
+
+if [[ $? -ne 0 ]] ; then
+    echo "Postgres Backup : pg_basebackup Failed"
+    exit(1)
+fi
 
 if [ -d "$DIRECTORY" ]; then
-    tar -czf --remove-files "${DIRECTORY}.tar.gz" "${DIRECTORY}"
+    tar -czf "${DIRECTORY}.tar.gz" "${DIRECTORY}" && rm -rf "${DIRECTORY}"
```
emord: What's the benefit of using `rm -rf` instead of `--remove-files?

NitigyaS: ohh I first try --remove-files It was throwing some error. Than I thought it is better to delete the directory only if tar completed with exit status 0

PR #1470, opened 2018-03-13
run django checks after migration

snopoke:
see https://github.com/dimagi/commcare-hq/pull/19766

### DIFF
--- a/fab/fab/operations/db.py
+++ b/fab/fab/operations/db.py
@@ -94,6 +94,7 @@ def migrate():
         sudo('%(virtualenv_root)s/bin/python manage.py sync_finish_couchdb_hq' % env)
         sudo('%(virtualenv_root)s/bin/python manage.py migrate_multi --noinput' % env)
 
+
 @roles(ROLES_DEPLOY)
 def set_in_progress_flag(use_current_release=False):
     venv = env.virtualenv_root if not use_current_release else env.virtualenv_current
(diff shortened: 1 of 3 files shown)


fab/fab/operations/db.py
```diff
@@ -93,6 +93,7 @@ def migrate():
     with cd(env.code_root):
         sudo('%(virtualenv_root)s/bin/python manage.py sync_finish_couchdb_hq' % env)
         sudo('%(virtualenv_root)s/bin/python manage.py migrate_multi --noinput' % env)
+        sudo('%(virtualenv_root)s/bin/python manage.py check --deploy' % env)
```
esoergel: I'm not sure `migrate` is the appropriate place for this, as this could end up checking a bunch of stuff.

snopoke: Yea, agree. I'll move it.

snopoke: 021a381

fab/fab/operations/db.py
```diff
@@ -94,7 +94,6 @@ def migrate():
         sudo('%(virtualenv_root)s/bin/python manage.py sync_finish_couchdb_hq' % env)
         sudo('%(virtualenv_root)s/bin/python manage.py migrate_multi --noinput' % env)
 
-
```
dannyroberts: Nit: this seems like an unintentional/irrelevant change.

PR #1475, opened 2018-03-14
Add service management for pending services

mkangia:
This adds service management for 
- [x] celery
- [x] webworkers
- [x] formplayer -> formplayer-spring
- [x] touchforms -> formplayer

Follow up on [earlier PR](https://github.com/dimagi/commcare-cloud/pull/1452)
[Issue](https://github.com/dimagi/commcare-cloud/issues/1243)


commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -319,6 +331,11 @@ def run_status_for_service_group(self, service_group, args, unknown_args):
         for service in self.services(service_group, args):
             if service == "redis":
                 args.shell_command = "redis-cli ping"
+            elif service == "celery":
+                shell_command = "supervisorctl status"
+                if args.worker_name:
+                    shell_command += " %s" % ' '.join(args.worker_name.split(','))
```
dannyroberts: I forgot to say this from the last PR, but in general we like to use `{}` and `.format(` rather than `%s` and `%`. (See for example https://stackoverflow.com/questions/12382719/python-way-of-printing-with-format-or-percent-form#12382738)

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -310,19 +318,133 @@ def make_parser(self):
...
+        app_processes_config = environment.translated_app_processes_config
+        return app_processes_config.celery_processes
+
+    def get_celery_config(self, args):
+        environment = get_environment(args.environment)
+        project_name = environment.fab_settings_config.project
+        """
+        This picks the logic from set_celery_supervisorconf to produce the full worker names
+        as declared in the template supervisor_celery_worker.conf
+        {{ project }}-{{ environment }}-celery_{{ celery_params.comma_separated_queue_names }}_{{ celery_params.worker_num }}
```
dannyroberts: Careful, this `environment` variable here is not the same as `args.environment` (sorry!). In fab, it gets populated from the `environment` property of the AppProcessesConfig. As you can see they are very often the same:
```
$ grep environment environments/*/app-processes.yml
environments/enikshay/app-processes.yml:environment: enikshay
environments/icds-new/app-processes.yml:environment: icds
environments/icds/app-processes.yml:environment: icds
environments/l10k/app-processes.yml:environment: l10k
environments/pna/app-processes.yml:environment: pna
environments/production/app-processes.yml:environment: production
environments/softlayer/app-processes.yml:environment: softlayer
environments/staging/app-processes.yml:environment: staging
environments/swiss/app-processes.yml:environment: swiss
```
but different for `icds-new`. `environment.translated_app_processes_config.environment` should get you what you're looking for.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -310,19 +318,133 @@ def make_parser(self):
...
     def get_inventory_group_for_service(self, service, service_group):
         return self.INVENTORY_GROUP_FOR_SERVICE.get(service, service_group)
 
+    @staticmethod
+    def _get_celery_processes(env):
+        environment = get_environment(env)
+        app_processes_config = environment.translated_app_processes_config
+        return app_processes_config.celery_processes
+
+    def get_celery_config(self, args):
```
dannyroberts: It looks like the only thing you use from `args` is `args.environment`, so a really simple refactor would be to have this function accept `env` instead, and I think that would be an improvement in that it's more explicit and confined, so easier to reason with.

I would also recommend, though it's a little more work, going one step further and making this (and other methods like `_get_celery_processes`) accept an environment object (from `get_environment(...)`) rather than the env name.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -310,19 +318,133 @@ def make_parser(self):
...
+        exit_code = 0
+        service = "celery"
+        if action == "status" and not args.worker_name:
+            args.shell_command = "supervisorctl %s" % action
+            args.inventory_group = self.get_inventory_group_for_service(service, args.service_group)
+            exit_code = RunShellCommand(self.parser).run(args, unknown_args)
+        else:
+            try:
+                workers_by_host = self.get_workers_to_work_on(args)
+            except Exception:
```
dannyroberts: We try not to use `except Exception` where another more explicit option exists. What do you intend for this except to catch?

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -310,19 +318,133 @@ def make_parser(self):
...
+        :return:
+        celery_worker_config: a worker name mapped to list for full worker names per host
+        {'submission_reprocessing_queue':
+          {'hqcelery0.internal-va.commcarehq.org':
+            ['commcare-hq-production-celery_submission_reprocessing_queue_0']
+          }
+        },
+        """
+        celery_worker_config = {}
+        celery_processes = self._get_celery_processes(args.environment)
```
dannyroberts: if we're working with the environment object (as I suggested above), then you won't really need this function at all. You can just use

```
celery_processes  = environment.translated_app_processes_config.celery_processes
```

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: oh ho.. this is a bit confusing..
i am going to take icds-new example

so `icds-new` is the environment
where `icds` should be called the `environment_environment` or `environment_name` or `commcare_environment`?

also `env` is not always an `Environment` object. This is similar to what we have in HQ were domain in code is at times the name of the domain or the `Domain` object. So we try to use project for `Domain` object but its not reliable still. 
Can we avoid that here by using `env_name` for the `name` like `icds-new`, and `env_obj` for `Environment` class object and `env` for the `fab.api.env` module.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: in method `get_workers_to_work_on` in case the worker name passed as the argument in not found, we need this to return exit code 1. So one way i thought of doing this was to raise an exception in that method and then catch that here. I am going to make this not generic.
Another way of doing this is to check if that method returns no workers.. going to think more about it.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: i think its a useful extraction to get celery processes out from `env` since its available directly on args. I am guessing this would be happening at more places than 1.  Can move it to `utils` if that makes more sense?

dannyroberts:
Build failed because of an import error. I would suggest moving the function into maybe `commcare-cloud/commcare_cloud/environment/schemas/app_processes.py` and then importing it into fab, rather than the other way around.

dannyroberts:
I should also reiterate that this is going to cause some conflicts with https://github.com/dimagi/commcare-cloud/pull/1424, so make sure you're thinking about that (sorry for the poor timing on that!)

mkangia:
@dannyroberts 

> Build failed because of an import error. I would suggest moving the function into maybe commcare-cloud/commcare_cloud/environment/schemas/app_processes.py and then importing it into fab, rather than the other way around.

Moved it to ansible helpers since this would eventually move to ansible after Emord's work on this.

> I should also reiterate that this is going to cause some conflicts with #1424, so make sure you're thinking about that (sorry for the poor timing on that!)

I had a quick sync up with emord about the changes in this PR. He has not made any changes to things happening via `set_celery_supervisorconf` which this PR takes care of.. So it should not conflict but he would need to consider changes being done here when he works on it, and so now he is aware of what i am doing in this PR.

mkangia:
@dannyroberts 
I think this PR now does what its intended to do for `celery`. 

I was quite confused about other supervisor set ups like `set_sms_queue_supervisorconf` where i confused them with actually sms_queues. So for ex:
On staging there is a 
`commcare-hq-staging-sms_queue` and also
`commcare-hq-staging-celery_sms_queue_0` where first one is just a supervisor setup running a management command but not managed by celery whereas the second one is actually via celery(@emord helped point this out to me, thanks Emord).
So i think we are good for managing celery here, apart from any reviews still left to address.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: this got removed in a later commit.

fab/fab/operations/supervisor.py
```diff
@@ -107,7 +107,8 @@ def set_celery_supervisorconf():
             conf_destination_filename = 'supervisor_celery_worker_{}_{}.conf'.format(
                 comma_separated_queue_names, worker_num)
 
-            worker_name = get_celery_worker_name(params['comma_separated_queue_names'],
+            worker_name = get_celery_worker_name(env.env_name,
```
dannyroberts: @mkangia I just noticed this: `env.env_name` isn't the right variable here, even though most of the time it'll have the same value. What you need is `environment.meta_yml.deploy_env`, which I believe in fabric is already set on the fabric `env` variable as `env.deploy_env`.

dannyroberts: @mkangia I just noticed this: `env.env_name` isn't the right variable here, even though most of the time it'll have the same value. What you need is `environment.meta_config.deploy_env`, which I believe in fabric is already set on the fabric `env` variable as `env.deploy_env`.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -377,21 +378,21 @@ def get_workers_to_work_on(self, args):
...
+        workers_by_host = defaultdict(set)
         if args.worker_name:
             for worker_name in args.worker_name.split(','):
                 if worker_name not in celery_config:
                     puts(colored.red(
                         "%s not found in the list of possible worker names, %s" % (
                             worker_name, celery_config.keys()
                         )))
-                    raise
+                    raise IncorrectOptions
```
dannyroberts: Not following that closely, but since the convention is to raise _instances_ of exception objects, I'd guess you want `IncorrectOptions()` here.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -278,6 +280,7 @@ class Service(_AnsiblePlaybookAlias):
         'pg_standby': ['postgresql', 'pgbouncer'],
         'celery': ['celery'],
         'webworkers': ['webworkers'],
+        'formplayer': ['formplayer', 'formplayer-spring']
```
dannyroberts: I would actually push for these two to be defined separately like this:

```
'formplayer': ['formplayer-spring'],
'touchforms': ['formplayer'],
```

Why? These supervisor services are kind of poorly named as a historical accident. What we now call "formplayer" is backed by the supervisor service named `formplayer-spring`, because what we used before, touchforms, was already being run under the name `formplayer`.

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -260,7 +260,12 @@ class Service(_AnsiblePlaybookAlias):
         commcare-cloud staging service celery status
         commcare-cloud staging service celery restart
         commcare-cloud staging service celery restart --only=saved_exports_queue,pillow_retry_queue
-        commcare-cloud production service celery restart --only=logistics_reminder_queue,saved_exports_queue
+    6. Check status or act on web workers
+        commcare-cloud staging service webworkers status
+        commcare-cloud staging service webworkers restart
+    7. Check status or act on formplayer and formplayer-spring
+        commcare-cloud staging service formplayer status
+        commcare-cloud staging service formplayer restart --only=formplayer-spring
```
dannyroberts: if you separate touchforms and formplayer like I suggested, remember to change this example.

fab/fab/operations/supervisor.py
```diff
@@ -106,10 +112,15 @@ def set_celery_supervisorconf():
             conf_destination_filename = 'supervisor_celery_worker_{}_{}.conf'.format(
                 comma_separated_queue_names, worker_num)
 
+            worker_name = get_celery_worker_name(env.env_name,
```
dannyroberts: yeah in all of these places where you're using `env.env_name` (fab) or `args.environment` (commcare-cloud) to generate the supervisor service name, you should use `env.deploy_env` (fab) or `environment.meta_config.deploy_env` (commcare-cloud) instead.

fab/fab/operations/supervisor.py
mkangia: @dannyroberts thanks for flagging.. i think this is correct.
taking example of `icds-new` here, we would want to pass `icds-new` here and not `icds`. The methods i have added, for example `get_celery_worker_name` will call `get_environment` using this env_name passed and hence this looks right to me. Let me know if that does not look right.

i am still confused about now the naming is working.. can you share your experience with that in the README or add some documentation so that one does not trip with this..

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: oh thats quite confusing.. thanks for clearing that out. Will update.

fab/fab/operations/supervisor.py
mkangia: responded earlier [here](https://github.com/dimagi/commcare-cloud/pull/1475#discussion_r176038699)

commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
mkangia: Cool, this is now replaced with `assert`, thanks for flagging though..

fab/fab/operations/supervisor.py
dannyroberts: I'm fairly sure that in the case of `icds-new`, we want the worker name to have `icds` (not `icds-new`) in order to preserve current behavior. Additionally, env_name here is _whatever the user typed into the command line_, so for `icds` where you can use `commcare-cloud icds ...` or `commcare-cloud icds-new ...`, you don't want the outcome to depend on which of those two you typed.

Here is where the `'environment'` param that is used in the current celery template gets set:
https://github.com/dimagi/commcare-cloud/blob/b23dfc47b9790a6ce61428b314dd2c21ac22c95a/fab/fab/operations/supervisor.py#L306
As you can see, it's set to the `deploy_env` value.

I'm not exactly sure where to document, but in general `deploy_env` is used for everything except datadog and email subject lines, which use `env_monitoring_id` (which is usually the same thing, but different for `icds-new`). Both of these are defined in `meta.yml`. In general, we _never_ use `env.env_name` (fab) / `args.environment` (ansible) for anything other than locating the environment configuration, for the reason given above.

dannyroberts: And thanks for asking, in the course of describing that, I realized there's at least one change I could make to reduce one bit of confusion, and that's to change `args.environment` to `args.env_name` in commcare-cloud so that fab and commcare-cloud call the same value the same thing.

dannyroberts: Aha. Offlined with @mkangia. The part I was missing is that `get_celery_worker_name` and related functions only use `env.env_name` to look up the environment (which is exactly how it's supposed to be used); it's not used directly in the service name.

Only comment now is that in the meantime `environment.translated_app_processes_config.environment` was moved to `environment.meta_config.deploy_env`, so @mkangia will make that change.

mkangia: thanks @dannyroberts .. 

i think the confusion how things are at fab and commcare-cloud.. is it
1. possible to have it at a single place?
2. follow the same naming convention at both places?

also again taking example of icds-new
icds-new would be the name of the environment
whereas icds would the deploy environment?

i still find that a little confusing but may be thats just a matter of being familiar.. can we use inventory name/inventory instead of environment name? when we talk environment i think one(at least I) always think of this as icds and icds-new.

PR #1481, opened 2018-03-15
Modification to use Versioing in S3

NitigyaS:
Upload Files with the same name. 
This will cause them to have different versions in s3. if versioning is enabled.

### DIFF
--- a/ansible/roles/backups/templates/backup_snapshots.py.j2
+++ b/ansible/roles/backups/templates/backup_snapshots.py.j2
@@ -3,7 +3,7 @@ import sys
 import boto3
 
 if len(sys.argv) == 2:
-    bucket = {{postgres_snapshot_bucket}}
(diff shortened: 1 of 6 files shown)


ansible/roles/backups/templates/backup_snapshots.py.j2
```diff
@@ -13,6 +13,11 @@ else:
     raise Exception('Usage is backup_snapshots.py filename <s3_bucket> [ <endpoint> ]')
 
 filename = sys.argv[1]
+# This will give daily / weekly
+upload_filename = sys.argv[1].split('_')[1]
```
snopoke: this relies on the format of the input filename which isn't obvious from outside this script. It might be better to pass in the upload filename as an additional param.

NitigyaS:
Can we merge this? 
I have changed the minimum back up to 2 days. or should I make it 1?

PR #1484, opened 2018-03-15
add ability to run on control machine instead

dannyroberts:
add --control before any command to run it on the control machine instead

What do you think?


proteusvacuum:
I do believe this is the almost the One-Click-Deploy-from-Control dream! `cchq production fab deploy --control`

PR #1487, opened 2018-03-16
Refactor environment name variables

dannyroberts:
There were three variables that had the same meaning or similar meaning with important (but poorly stated) distinctions: `environment` in `app-processes.yml` and `deploy_env` and `env_name` in `public.yml`. Turns out that `environment` and `deploy_env` were not only always the same in all environments, but they also had exactly the same semantic meaning and were used for the same thing; `deploy_env` was used in ansible and `environment` was used in fab. `env_name` had a different meaning—it's the key used for email headings and the environment name in datadog—but had the same value in all environments except `icds`. (In fact, the way that was done is that it was only declared in `icds`, and every time it was used in ansible, `deploy_env` was manually provided as a backup. I've made the following changes:

- these variables now live in a new `meta.yml` file rather than `public.yml` and `app-processes.yml`
- `deploy_env` is used throughout (including in fab); `environment` is now dead
- `env_name` has been renamed to `env_monitoring_id`—this is a clearer indication of what it's for and why it's a different variable from `deploy_env`

May seem small, but it's a nice step towards getting these two to work off a shared variable set and shared vocabulary. A bit of a followup on this exchange a while ago: https://github.com/dimagi/commcare-cloud/pull/1279#issuecomment-358652959

**Update**: realized this outgrew the way we've been running travis tests, so as a bonus in this PR I've switched over to having travis run its test command using `commcare-cloud`, and as a bonus from _that_, I added a `validate-environment-settings` command that was helpful to me during this process.


dannyroberts:
✅

fab/fab/fabfile.py
```diff
@@ -222,16 +223,19 @@ def development():
...
     Example command:
 
         fab development awesome_deploy \
-        --set inventory=/path/to/commcare-cloud/ansible/inventories/development,environment=dev
+        --set inventory=/path/to/commcare-cloud/ansible/inventories/development,deploy_env=dev
 
     """
-    load_env('development')
+    # not sure if this still works, it was broken for a while without anyone saying anything
+    # so guessing we can delete this task altogether? ~ Danny
```
snopoke: probably doesn't get used very often (or ever)

dannyroberts: awesome, good to know.

PR #1495, opened 2018-03-19
Changed redis node and var appendonly

NitigyaS:


environments/production/public.yml
```diff
@@ -121,6 +121,7 @@ redis_syslog_enabled: 'no'
 redis_database_save_times: []
 redis_max_clients: 10000
 redis_maxmemory_policy: allkeys-lru
+redis_appendonly: 'yes'
```
emord: Why is this changing?

NitigyaS: By default data in redis resides only in memory and not persistent. by enabling redis also writes data in an append-only file. If the server reboots it will play these logs again.

snopoke: longstanding outcome of a retro from a while back that we should persist redis data to avoid issues if a restart happens

emord: :+1: in that case, should this be the default?

PR #1497, opened 2018-03-19
dev setup cleanup

snopoke:
Tried to test something on vagrant and found it pretty tricky without the correct files in place. 

I copied the environment files from the bootstrap env. Not loving the duplication so very open to alternatives.


snopoke:
@dannyroberts I spent a bit more time on this and think this is now in a much better state for local testing

ansible/README.md
```diff
@@ -14,38 +14,34 @@ and the other shortcut commands it provides.
...
-If you run into issues starting vagrant, see the troubleshooting section at the bottom.
+This will setup 4 VMs:
+* 192.168.33.14 (control)
+* 192.168.33.15 (app1)
+* 192.168.33.16 (db1)
+* 192.168.33.17 (proxy1)
+
+See `environments/development` the configuration of this testing environment.
 
-The `./scripts/reset-vms` command can be run at any time, possibly with a subset of the
```
millerdev: What's the equivalent way to do this now that this PR has been merged? I think it's important to be able to easily reset vagrant machines in order to test deployment from a clean state.

snopoke:
```
vagrant destroy [vm]
vagratn up
```

snopoke: To add more detail - the main thing that script was doing was setting up the auth but that's now done during Vagrant provisioning.

PR #1499, opened 2018-03-19
use newer github api library so deploy fails fast when api key doesn't have the right scope

nickpell:
This addresses https://manage.dimagi.com/default.asp?266529 by making sure the deploy fails fast instead of silently failing to create the release tag, and later failing hard when the release tag is not found on github.

If the github api key doesn't have the scope ```public_repo```, then the deploy fails right at the start with a 401 error.

Example release generated with new code: https://github.com/dimagi/commcare-hq/releases/tag/2018-03-19_23.08-staging-deploy

### DIFF
--- a/fab/fab/utils.py
+++ b/fab/fab/utils.py
@@ -79,13 +79,9 @@ class DeployMetadata(object):
             )))
         tag_name = "{}-{}-deploy".format(self.timestamp, self._environment)
         msg = "{} deploy at {}".format(self._environment, self.timestamp)
-        repo.create_git_tag_and_release(
-            tag=tag_name,
-            tag_message=msg,
-            release_name=tag_name,
-            release_message=msg,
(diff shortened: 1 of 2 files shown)


fab/fab/utils.py
```diff
@@ -66,8 +66,8 @@ def tag_commit(self):
 
         pattern = ".*-{}-.*".format(re.escape(self._environment))
         github = _get_github()
-        repo = github.repository('dimagi', 'commcare-hq')
-        for tag in repo.tags(self._max_tags):
+        repo = github.get_organization('dimagi').get_repo('commcare-hq')
```
emord: @nickpell fyi this is where the deploy fails with a 400 on ICDS (specifically `get_organization`) Still haven't figured this one out. Tried login/password, and a few different API keys

PR #1501, opened 2018-03-20
Added correct options

NitigyaS:


ansible/setup_pg_standby.yml
```diff
@@ -61,8 +61,7 @@
           -h {{ hot_standby_master }} \
           -U {{ postgres_users.replication.username }} \
           -D {{ postgresql_home }} \
-          -S {{ replication_slot }} \
-          -R -v --xlog-method=stream 2>&1
+          -P -v --xlog-method=stream 2>&1
```
dannyroberts: Looks like you're removing the `-S <replication slot>` option and replacying `-R` with `-P`. Can you describe this change? Particular questions to answer:
- what motivated the change
- does it fix something that was broken, change functionality, or simply do the same thing before with a shorter command?

etc.

NitigyaS: pg_basebackup do not provide -S flag. So, I changed it to hqstandby0. The branch does not need to be merged with the master. it was only for the scheduled downtime.

dannyroberts: If `pg_basebackup` doesn't allow the `-S` flag, then don't we want to merge this? I don't have a lot of context here. Can you describe more fully how what you used this for today differs from what this playbook is for, just so I can understand?

NitigyaS: Context: Production used 9.4 and all other environment uses 9.6. Present playbook runs fine for 9.6 but for 9.4 `pg_basebackup` commands will fail since different version of `pg_basebackup` uses different flags.
Solution: 
1) Write logic to use different pg_basebackup flags for different Postgres versions. 
2) Use a temporary branch to create hqstandby1 . Instead of merging it. (We will get rid of 9.4 soon)

I proceeded with solution 2.

dannyroberts: Great, thank you @NitigyaS, very helpful. That clears things up for me!

PR #1502, opened 2018-03-20
Make auditcare configurable

emord:
@snopoke I've seen some talk about taking this off of ICDS and this should do it, but I'm skeptical it will solve the issue. I tried to find correlation with any of the couchdb metrics we have in datadog, but couldn't find anything in particular.

My feeling is that https://github.com/dimagi/commcare-hq/blob/je/comp-feeding-forms-agg-script/corehq/ex-submodules/pillowtop/processors/form.py#L18 introduces too much churn on the users db but I have nothing to prove it other than at the end of the users DB is 15 GB and is then compacted to ~2 GB


snopoke:
@emord I've rolled this out

PR #1503, opened 2018-03-20
use redis role from galaxy

snopoke:
This would require restarting redis on any env it's run on but that needs to happen anyway when we add storage.


ansible/group_vars/redis.yml
```diff
@@ -1 +1,12 @@
 datadog_integration_redisdb: true
+redis_version: 4.0.8
+redis_install_dir: /opt/redis
+redis_dir: "{{ encrypted_root }}/redis"
+redis_local_facts: false
+redis_service_name: redis-server
+redis_install_dir: "/opt/redis/{{ redis_version}}"
```
emord: Double defined `redis_install_dir`.

snopoke: aab2b4c

snopoke:
This will update the existing init script to point to the new redis lib and also updates `/usr/bin` so all future processes will use the new version.

PR #1510, opened 2018-03-21
Split service command into its own file

dannyroberts:
Preserved history using this trickery: https://stackoverflow.com/a/44036771/240553


mkangia:
👍

PR #1514, opened 2018-03-22
datadog check for serverup

snopoke:
I want to test this out as a possible replacement for pingdom.

I've split out the `heartbeat` check from the others since that's usually the one that's flaky and has a lot of false positives.

Combine with https://github.com/dimagi/commcare-hq/pull/19880

### DIFF
--- a/ansible/roles/datadog/templates/http.yaml.j2
+++ b/ansible/roles/datadog/templates/http.yaml.j2
@@ -8,5 +8,6 @@ instances:
     days_warning: 28
     days_critical: 14
     timeout: 20
-  - name: {{ env_monitoring_id }} Heartbeat
-    url: https://{{ SITE_HOST }}/serverup.txt?only=heartbeat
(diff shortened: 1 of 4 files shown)


ansible/roles/datadog/templates/http.yaml.j2
```diff
@@ -0,0 +1,12 @@
+init_config:
+
+instances:
+  - name: {{ env_monitoring_id }}
+    url: https://{{ SITE_HOST }}/serverup.txt
+    disable_ssl_validation: false
+    check_certificate_expiration: true
+    days_warning: 28
+    days_critical: 14
```
jmtroth0: I probably don't know what this means, but this seems like a lot of days?

snopoke: this is how much many days before the cert expires that the alert will go from `warning` to `critical`

PR #1515, opened 2018-03-22
incorrect usages of ipaddr

snopoke:
`ipaddr` filter can test for IP addresses and convert between various formats. If the value passed in isn't an IP address it returns false: https://docs.ansible.com/ansible/latest/playbooks_filters_ipaddr.html

These usages seem incorrect, at best they are a noop and at worse the convert a hostname like `hqdb0.internal....` to `False`.


snopoke:
The other uses look legitimate.

PR #1520, opened 2018-03-22
Remove old blob deletion pillows

emord:
I thought I had already done this, but back in May I removed these from the HQ code https://github.com/dimagi/commcare-hq/pull/15972

### DIFF
--- a/commcare-cloud-bootstrap/environment/app-processes.yml.j2
+++ b/commcare-cloud-bootstrap/environment/app-processes.yml.j2
@@ -22,11 +22,7 @@ pillows:
       num_processes: 1
     AppDbChangeFeedPillow:
       num_processes: 1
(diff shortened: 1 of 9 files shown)


emord:
Now realize that this worked when deploying in fab because it generated these configs in the current code's directory. Going to merge to get rid of some erorrs after deploying #1424

millerdev:
Are we no longer deleting unreferenced blobs? I think I'm changing my opinion on that, given the recent disk space issues we've had.

emord:
@millerdev we haven't had those pillows running since May. I'm having difficulty finding any context as to why we deleted them but I recall talking about it with you at some point. If we want those to continue, we'd also need to revert the old HQ PR

millerdev:
OK, thanks. I'll put this on my backlog to review.

PR #1522, opened 2018-03-22
Datadog rotated their key

benrudolph:
https://help.datadoghq.com/hc/en-us/articles/360000886852-Agent-APT-Repo-Key-Rotation

hit this today when deploying the datadog role

PR #1524, opened 2018-03-23
update expired apt key

snopoke:


snopoke:
https://github.com/dimagi/commcare-cloud/pull/1522

PR #1526, opened 2018-03-23
Added cchq command to update services.

emord:
Following up with an alias from my last PR. Happy to change the name if something else makes sense.

### DIFF
--- a/commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -239,8 +239,8 @@ class UpdateUsers(_AnsiblePlaybookAlias):
         return AnsiblePlaybook(self.parser).run(args, unknown_args)
 
 
-class UpdateServices(_AnsiblePlaybookAlias):
-    command = 'update-services'
+class UpdateSupervisorConfs(_AnsiblePlaybookAlias):
+    command = 'update-supervisor-confs'
     help = (
         "Updates the supervisor configuration files for services required by "
         "CommCare. These services are defined in app-processes.yml."
--- a/commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -237,3 +237,16 @@ class UpdateUsers(_AnsiblePlaybookAlias):
         args.playbook = 'deploy_stack.yml'
         unknown_args += ('--tags=users',)
         return AnsiblePlaybook(self.parser).run(args, unknown_args)
+
+
+class UpdateServices(_AnsiblePlaybookAlias):
+    command = 'update-services'
+    help = (
+        "Updates the supervisor configuration files for services required by "
+        "CommCare. These services are defined in app-processes.yml."
+    )
+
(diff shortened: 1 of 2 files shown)


commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -237,3 +237,16 @@ def run(self, args, unknown_args):
         args.playbook = 'deploy_stack.yml'
         unknown_args += ('--tags=users',)
         return AnsiblePlaybook(self.parser).run(args, unknown_args)
+
+
+class UpdateServices(_AnsiblePlaybookAlias):
```
mkangia: nit: can this be named better? @dannyroberts would probably have better suggestions but i think when we talk about Service it would include everything like riakcs or webworker that does not work with supervisor. Can this be names just `UpdateSupervisorConf` or `UpdateAppProcessesConf`. 
Might want to update the tag `services` as well since that is quite generic.

dannyroberts: Hmm, yeah it's a delicate balance here, and I see both sides of it. On the one hand, "update services" could mean a number of things, and on the other hand `update-app-services`, which is at least a little more specific, might be an annoying thing to remember when typing the command. (I say "app services" because webworkers _does_ happen to be run by supervisor, as do all of the other things that run off of the commcare-hq code base, whereas most/all of our third party services are run using `service`.)

I also thought maybe call it `update-app-services` with an alias to `update-services`—but that just leaves people with the vague impression that those two might do different things, so that's not great either. Right now I'm leaning towards calling it `update-services`—just because I'm okay with this tool being very "commcare-hq-"-centric and having that be implied—even though `update-app-services` is the more explict/technically correct choice. I do wish I had a more certain opinion here though.

mkangia: Thanks @dannyroberts ..
keeping things a bit vague/not definite would eventually cause confusion later.. like i shared the example of domain in hq code base over a PR earlier. At places its just a name and at places its a `Domain` object. I usually go for definite naming even if its long (from my habit working on Ruby on Rails). 
I just think using Service in the name without using `conf` would add confusion with the `service` we refer to that includes non hq services as well. Plus i am not sure if this command would be used as often that we compromise on the naming to keep it short.

I like the idea of using an alias to keep it shorter. `UpdateAppServicesConf` or `UpdateAppProcessesConf`with command `update-app-services-conf`/ `update-app-processes-conf` with alias `update-app-conf` at least puts the clarity in the class name.
(also, this is non blocking for the PR)
(just a coding style i believe, feel free to not discuss this any further if you think `update-services` is fine)

dannyroberts: Offlined with @emord a bit, and he pointed out that this doesn't actually reload supervisor, just updates the files. Maybe for now we should go with a more literal name like `update-supervisor-confs` until we've integrated this more into a process that can get its own more conceptual name.

dannyroberts:
@emord still planning to change the name and have this merged? Happy to take over that small change if you're focus right now is elsewhere, just wanted to make sure you're on board.

emord:
@dannyroberts Sorry lost track of this. updated

PR #1528, opened 2018-03-23
Stop deploying to postgres servers

emord:
I don't think there's much of a reason to deploy code to postgres as none of them need to run HQ code. May also cut down on deploys to ICDS since all of those servers need to go through the proxy and that could be a bottleneck.


fab/fab/fabfile.py
```diff
@@ -254,7 +253,7 @@ def env_common():
     # if no server specified, just don't run pillowtop
     pillowtop = servers.get('pillowtop', [])
 
-    deploy = servers.get('deploy', servers['postgresql'])[:1]
+    deploy = servers.get('deploy', servers['webworkers'])[:1]
```
emord: May be nicer to have an explicit deploy machine specified in the inventory, but this seems good enough for now

dannyroberts:
Took me a minute to understand what's happening so I'll state my understanding in case it's wrong. This PR change the machine on which management commands that are run during deploy are run from a `postgresql` machine to a `webworkers` machine. The motivation for this is that it has the effect of not needing to have the commcare-hq codebase on `postgresql` machines, which may speed up deploy a little. (Am I right in thinking this would just remove the need to update code on _one_ postgres machine, as currently all the others already don't need it?)

snopoke:
@dannyroberts although what you say is true (we only need HQ source on one PG machine) we still deploy it to all PG machines (most of our envs have more than one PG machine now and ICDS has 8)

dannyroberts:
> [...] we still deploy it to all PG machines
Brilliant! Thanks for fixing, @emord

emord:
@dimagi/team-commcare-hq FYI in case you missed this. You can no longer run management commands on db machines as they won't be updated. You must use a webworker celery or pillowtop machine

snopoke:
@emord should we just delete the releases from the DB machines to make it impossible for anyone to make that mistake?

emord:
Deleted releases for prod, softlayer and ICDS. I think all the others also run other service which need the code directories

PR #1530, opened 2018-03-26
remove 'config' repo dependency

snopoke:
And some cleanup.

buddy @jmtroth0


ansible/roles/nginx/tasks/install.yml
```diff
@@ -150,7 +125,7 @@
     group: root
   when: not fake_ssl_cert and tableau_nginx_combined_cert_value is defined
 
-- name: Copy icds tableau SSL Key (from vault)
+- name: Copy icds tableau SSL Key
```
emord: I think that this can be removed now

dannyroberts:
Yeah this is great, thanks @snopoke 🎉

snopoke:
@emord re using `with_items`, I thought about doing that but didn't have a good solution for figuring out which were 'enabled'

PR #1533, opened 2018-03-27
Ns/fix swiss issue

NitigyaS:
1) Added Flag to enable versioning in scripts
2) Set Flag to true in all.yml
3) Set Flag to false for swiss

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -140,4 +140,4 @@ es_snapshot_bucket: "dimagi-{{ deploy_env }}-es-snapshots"
 aws_region: None
 # this reads "'s3.{{ aws_region }}.amazonaws.com' if aws_region else None"
 aws_endpoint: '{{ aws_region and "s3." + aws_region + ".amazonaws.com" }}'
-aws_versioning_enable: true
+aws_versioning_enabled: true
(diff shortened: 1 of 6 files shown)


ansible/roles/backups/templates/create_blobdb_backup.sh.j2
```diff
@@ -4,12 +4,18 @@ DAYS_TO_RETAIN_BACKUPS=$2
 HOSTNAME=$(hostname)
 TODAY=$(date +"%Y_%m_%d")
 BACKUP_FILE="blobdb_${BACKUP_TYPE}_${TODAY}.tar.gz"
+UPLOAD_NAME="blobdb_${BACKUP_TYPE}_${HOSTNAME}.tar.gz"
+
+{% if not aws_versioning_enable%}
+#Reset if versioning is not enabled
+UPLOAD_NAME="${BACKUP_FILE}"
+{% endif %}
```
snopoke: These would be a bit clearer if it was just an `if else`
```
{% if not aws_versioning_enable%}
UPLOAD_NAME="${BACKUP_FILE}"
{% else %}
UPLOAD_NAME="blobdb_${BACKUP_TYPE}_${HOSTNAME}.tar.gz"
{% endif %}
```

ansible/group_vars/all.yml
```diff
@@ -140,3 +140,4 @@ es_snapshot_bucket: "dimagi-{{ deploy_env }}-es-snapshots"
 aws_region: None
 # this reads "'s3.{{ aws_region }}.amazonaws.com' if aws_region else None"
 aws_endpoint: '{{ aws_region and "s3." + aws_region + ".amazonaws.com" }}'
+aws_versioning_enable: true
```
dannyroberts: I think this variable should be named `aws_versioning_enabled` (with a d).

snopoke:
@NitigyaS is this done?

NitigyaS:
yes

PR #1534, opened 2018-03-27
Added datadog_integration_http.

NitigyaS:
Datadog is skipping configs 
in https://github.com/dimagi/commcare-cloud/blob/ee1692663eddb6e10870cbf8c05f7991c1abfd0a/ansible/roles/datadog/tasks/add_integrations.yml#L38
since variable `datadog_integration_http` was not defined.

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -93,6 +93,7 @@ datadog_integration_riakcs: false
 datadog_integration_zk: false
 datadog_integration_jmx: false
 datadog_integration_vmware: false
+datadog_integration_http: false
 
 root_email: commcarehq-ops+root@dimagi.com
 server_email: commcarehq-noreply@dimagi.com


NitigyaS:
I removed the condition `item is defined` because this play was skipped . Ran the --check mode and got the error `datadog_integration_http is not defined` . the error was resolved on adding the variable.

NitigyaS:
I believe it broke when we added the check to remove Pingdom. since earlier it went smoothly on hqstandby0 and hqstandby1 .

snopoke:
Not sure what you're referring to re Pingdom.

NitigyaS:
Are we using HTTP check for that purpose right ? maybe I am wrong that was a guess

snopoke:
oh, I get what you're saying now. I thought you were saying we'd removed some pingdom related thing from this repo.

PR #1537, opened 2018-03-27
specify staging control but skip it in ansible runs

dannyroberts:
Followup to https://github.com/dimagi/commcare-cloud/pull/1517 that still lets us run `cchq staging ssh control`, `cchq --control staging fab deploy`, etc.

### DIFF
--- a/environments/staging/inventory.ini
+++ b/environments/staging/inventory.ini
@@ -46,5 +46,9 @@ hqriak00-staging.internal-va.commcarehq.org
 hqriak01-staging.internal-va.commcarehq.org
 
 [control]
+control.internal-va.commcarehq.org
 
 [pg_standby]
+
+[ansible_skip:children]
+control
--- a/commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/commcare-cloud/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -11,7 +11,7 @@ from commcare_cloud.commands.ansible.helpers import (
     get_user_arg)
 from commcare_cloud.commands.command_base import CommandBase
 from commcare_cloud.commands.shared_args import arg_inventory_group, arg_skip_check, arg_quiet, \
-    arg_branch, arg_stdout_callback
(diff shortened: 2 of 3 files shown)


environments/staging/inventory.ini
```diff
@@ -46,5 +46,9 @@ hqriak00-staging.internal-va.commcarehq.org
 hqriak01-staging.internal-va.commcarehq.org
 
 [control]
+control.internal-va.commcarehq.org
 
 [pg_standby]
+
+[ansible_skip:children]
```
emord: Do you see this being used elsewhere? It feels like this adds a bit of complexity for something that's fairly unnecessary.

Any reason to not solve this by making a different staging server control?

dannyroberts: Yeah, I don't know, but I could totally see a third party setting up a number of different monoliths for different customers and using the control machine for all of them. Not that that's great security-wise. But yeah, I don't know. Just for the specific problem at hand, this seemed like an easier solution to me than setting up a new server and telling everyone to set up either control machine envs again.

dannyroberts:
I should also say, though it's a small thing, that this approach does technically solve a problem in `after-reboot` behavior, that if you specified `--limit` there (in addition to the required `inventory_group` positional arg), one or the other of those two would arbitrarily win; now it does the intersection of the two, which though rare, is almost certainly what someone would mean if they did that.

environments/staging/inventory.ini
emord: > setting up a new server

I was originally thinking just reusing an already used server, not provisioning a new one. (we reuse kafka0 as the "control" on ICDS, so there's precedent)

Anyways, I'm likely not the person to be commenting on this since I'm still stuck in my ways and only use cchq for ansible deploying stack

dannyroberts:
Ah, I see. Well, I'm going to go with this since I foresee this working well enough, but if it becomes awkward and/or if excluding this way causes bizarre behavior of some sort, then happy to revert and go the route you suggest.

PR #1538, opened 2018-03-28
Ce/airflow redux

calellowitz:
@snopoke
I redid the PR without the postgres playbook changes, using our existing pg nodes. I think I addressed all the feedback except about the config file. I also added a fab role to update airflow code, and fixed up the supervisor files. Let me know how it looks to you


ansible/group_vars/all.yml
```diff
@@ -135,10 +168,12 @@ supervisor_http_password: "{{ secrets.SUPERVISOR_HTTP_PASSWORD | default(None) }
...
 blobdb_snapshot_bucket: dimagi-{{ deploy_env }}-blobdb-backups
 couchdb_snapshot_bucket: dimagi-{{ deploy_env }}-couch-backups
 postgres_snapshot_bucket: dimagi-{{ deploy_env }}-postgres-backups
 es_snapshot_bucket: "dimagi-{{ deploy_env }}-es-snapshots"
-aws_region: None
-# this reads "'s3.{{ aws_region }}.amazonaws.com' if aws_region else None"
-aws_endpoint: '{{ aws_region and "s3." + aws_region + ".amazonaws.com" }}'
+aws_endpoint: "s3.{{ aws_region }}.amazonaws.com"
+
+airflow_secret_key: '{{ secrets.AIRFLOW_SECRET_KEY | default(None) }}'
```
calellowitz: I left this like that because it is how all the other secrets are done in this file. Happy to change to something else if that makes sense to you. I think having this as the string 'None' will break airflow since it expects a fernet key, which probably makes this a safe thing to do.

ansible/roles/airflow/templates/airflow.cfg.j2
```diff
@@ -0,0 +1,350 @@
+[core]
```
calellowitz: I am not sure if these are all required or if we can be eliminate the ones we do not set, and they seemed well organized (they are currently organized thematically by default). I agree it's hard to see which ones we are controlling and which we are not though, so I am curious if you have alternative organizational options you would recommend.

calellowitz:
This included more changes than expected (reverting changes to master that happened since my original work because of the way I applied the changes), so I am adding a do not merge tag, until I can revert those.

ansible/deploy_commcarehq.yml
```diff
@@ -20,49 +21,3 @@
...
-    - shell: ls -1 {{ service_home|quote }}
-      register: contents
-      tags: services
-    - name: "Remove supervisor configs that are no longer needed"
-      file:
-        path: "{{ service_home }}/{{ item }}"
-        state: absent
-      with_items: contents.stdout_lines
-      when: item not in supervisor_files
-      tags: services
```
snopoke: I don't think these should be removed

ansible/deploy_db.yml
```diff
@@ -36,7 +37,7 @@
     - hostTo: "{{ hot_standby_server }}"
     - state: 'present'
   tasks:
-    - import_tasks: roles/setup_auth_keys.yml
+    - include: roles/setup_auth_keys.yml
```
snopoke: `include` is deprecated, why didn't `import_tasks` work? If it needs to be dynamic then you should use `include_tasks`

ansible/deploy_db.yml
```diff
@@ -131,7 +132,7 @@
   roles:
     - {role: kafka, tags: 'kafka'}
 
-- import_playbook: deploy_riakcs.yml
+- include: deploy_riakcs.yml
```
snopoke: similar to above about deprecated `include`

ansible/deploy_stack.yml
```diff
@@ -22,7 +22,7 @@
 - import_playbook: deploy_mailrelay.yml
 - import_playbook: deploy_tmpreaper.yml
 - import_playbook: deploy_etckeeper.yml
-# migrate must happen before first fab, and also before touchforms can create its special user
-- import_playbook: migrate_on_fresh_install.yml
```
snopoke: probably shouldn't be removed

ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+  git:
+    repo: "{{ airflow_repository }}"
+    dest: "{{ airflow_code_dir }}"
+    version: "{{ airflow_version }}"
+    recursive: yes
+    accept_hostkey: yes
+    update: yes
+  become: true
+
+
```
snopoke: nit: double space

ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+- name: install pip requirements
+  become: true
+  pip:
+    requirements: "{{ airflow_code_dir }}/requirements.txt"
+    virtualenv: "{{ airflow_virtualenv }}"
+    chdir: "{{ airflow_code_dir }}"
+
+- name: chown clone
```
snopoke: probably not necessary if you specify `become_user: "{{ cchq_user }}"` on some of the tasks above

ansible/roles/airflow/templates/airflow.cfg.j2
snopoke: I think it depends how airflow handles this file. If you leave an option out will it use a default? If that's the case then I'd suggest removing anything that's not being set by us + any comments that aren't necessary

ansible/roles/airflow/templates/airflow.cfg.j2
```diff
@@ -0,0 +1,350 @@
...
+s3_log_folder =
+
+# The executor class that airflow should use. Choices include
+# SequentialExecutor, LocalExecutor, CeleryExecutor
+executor = LocalExecutor
+
+# The SqlAlchemy connection string to the metadata database.
+# SqlAlchemy supports many different database engine, more information
+# their website
+sql_alchemy_conn = postgresql://{{ postgres_users.commcare.username }}:{{ postgres_users.commcare.password }}@127.0.0.1:5432/airflow
```
snopoke: db host should be parameterized and port should be 6432 (or parameterized)

Possibly take the same approach as formplayer: https://github.com/dimagi/commcare-cloud/blob/760932dcebaa8e31bfd82fb7c8e4a53d700dad97/ansible/roles/formplayer/templates/application.properties.j2#L31

fab/fab/operations/airflow.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import absolute_import
+
+from fabric.api import roles, sudo, env
+from fabric.context_managers import cd
+
+from ..const import ROLES_AIRFLOW
+
+@roles(ROLES_AIRFLOW)
+def deploy_airflow():
```
snopoke: should be `update_airflow` based on call in `deploy_airflow` task

fab/fab/operations/supervisor.py
```diff
@@ -42,6 +43,8 @@ def set_supervisor_config():
     set_djangoapp_supervisorconf()
     set_formsplayer_supervisorconf()
     set_formplayer_spring_supervisorconf()
+    set_airflow_scheduler_supervisorconf()
+    set_airflow_webserver_supervisorconf()
```
snopoke: I think these can be moved to ansible

ansible/deploy_db.yml
```diff
@@ -11,6 +11,7 @@
     - riakcs
     - formplayer
     - pg_standby
+    - airflow
```
snopoke: I don't think having the DB on the same VM is necessary - even if it is this should be accomplished in the inventory file by putting the airflow host in the `postgresql` group

ansible/group_vars/all.yml
```diff
@@ -143,3 +143,12 @@ aws_region: None
 # this reads "'s3.{{ aws_region }}.amazonaws.com' if aws_region else None"
 aws_endpoint: '{{ aws_region and "s3." + aws_region + ".amazonaws.com" }}'
 aws_versioning_enabled: true
+
+airflow_secret_key: '{{ secrets.AIRFLOW_SECRET_KEY | default(None) }}'
+airflow_flask_key: '{{ secrets.AIRFLOW_FLASK_KEY | default(None) }}'
```
snopoke: This will end up as a string `'None'`

ansible/roles/airflow/templates/airflow.cfg.j2
```diff
@@ -0,0 +1,67 @@
...
+# This path must be absolute
+dags_folder = {{ airflow_home }}/dags
+
+# The folder where airflow should store its log files
+# This path must be absolute
+base_log_folder = {{ airflow_home }}/logs
+
+# The executor class that airflow should use. Choices include
+# SequentialExecutor, LocalExecutor, CeleryExecutor
+executor = LocalExecutor
```
snopoke: docs recommend `CeleryExecutor` for production usage, do we have plans to move to that later?

ansible/roles/airflow/templates/airflow.cfg.j2
```diff
@@ -0,0 +1,67 @@
...
+# The maximum number of active DAG runs per DAG
+max_active_runs_per_dag = 1
+
+# Whether to load the examples that ship with Airflow. It's good to
+# get started, but you probably want to set this to False in a production
+# environment
+load_examples = False
+
+# Where your Airflow plugins are stored
+plugins_folder = /Users/calellowitz/airflow/plugins
```
snopoke: `/Users/calellowitz`

ansible/group_vars/all.yml
calellowitz: That seems to be what all the other credentials in this file do. Happy to go a different route if there is a preferable one, but as a default that seems at worst harmless.

ansible/roles/airflow/templates/airflow.cfg.j2
calellowitz: As we scale, we probably. For now it does not seem worth it as all the tasks can probably be run locally, especially since the db machine does all the heavy lifting since most of the work is raw sql, and the added complexity of celery does not buy us much unless we have other workers.

calellowitz:
@snopoke I think I got to your questions and would love to merge this as not having it on master is causing problems with the staging setup whenever someone else ansible deploys there.

calellowitz:
@snopoke I removed the do not merge label. want to get it in to avoid further merge confilcts (especially in the vault file). This will have no effect if there are no airflow machines in the inventory, but will prevent staging changes from breaking the setup.

src/commcare_cloud/ansible/roles/airflow/tasks/main.yml
```diff
@@ -0,0 +1,60 @@
...
+
+- name: install pip requirements
+  become: yes
+  become_user: "{{ cchq_user }}"
+  pip:
+    requirements: "{{ airflow_code_dir }}/requirements.txt"
+    virtualenv: "{{ airflow_virtualenv }}"
+    chdir: "{{ airflow_code_dir }}"
+
+- name: copy airflow settings
```
dannyroberts: Ran into this during a search. I think (at least eventually) this task should get run on `update-config`, which also updates `localsettings.py` and formplayer's `application.properties`.

snopoke: I agree with some of that but I also think that it's useful to have a command to update all settings - for instance if you change the IP of a service that they all use.

Perhaps the best approach would be to make the `update-config` command able to update the all or each one individually.

calellowitz: Yeah, I am not opposed to including this in `update-config` as long as a dedicated way still exists. In the past we have replaced those methods with just adding it to `update-config`, so there is no longer an officially supported way to deploy just localsettings for example, which is probably the dominant case.  In the same way we separate `service restart` to happen by service, even though sometimes you want to do all of them, I think it makes sense to maintain the ability to update single configs.

I more meant to call out what I saw as a general trend of grouping of tasks that are conceptually similar but unlikely to be needed at the same time into single commands, as a replacement for the individual ones.

PR #1539, opened 2018-03-28
fix commcare-cloud-bootstrap env

dannyroberts:
just keeping things up to date with recent changes

### DIFF
--- a/ansible/roles/commcarehq/templates/supervisor_pillowtop.conf.j2
+++ b/ansible/roles/commcarehq/templates/supervisor_pillowtop.conf.j2
@@ -1,4 +1,4 @@
-{% for pillow_name, params in app_processes_config.pillows.get(inventory_hostname).items() %}
+{% for pillow_name, params in app_processes_config.pillows.get(inventory_hostname, {}).items() %}
 {% for num_process in range(params.get('start_process', 0), params.get('start_process', 0) + params.get('num_processes', 1)) %}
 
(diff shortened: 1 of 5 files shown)


ansible/roles/commcarehq/templates/supervisor_pillowtop.conf.j2
```diff
@@ -1,4 +1,4 @@
-{% for pillow_name, params in app_processes_config.pillows.get(inventory_hostname).items() %}
+{% for pillow_name, params in app_processes_config.pillows.get(inventory_hostname, {}).items() %}
```
emord: bootstrapping doesn't have any pillows? seems like that would be a broken installation

dannyroberts: Oh, well no I noticed this when solving the problem in the other commit. I think we should either do this or do `app_processes_config.pillows[inventory_hostname]`; if figured you were going for lenient when you wrote this so I went in that direction, but I'm fine going more explicitly strict as well.

emord: oh. yeah. that makes sense. I don't know why I used `.get` at all now. I did go through a lot of iteration on that other PR so it's possible I added it to fix a bug, but it's no longer necessary.

I generally prefer to fail fast and hard

dannyroberts: Okay, I'll make this small change on this PR and then merge

PR #1542, opened 2018-03-29
Refactor active_sites

dannyroberts:
A crazy amount of repetitive code was involved in declaring for each envs which sites to install, even though most of it is totally predictable. (Check number of lines added/removed!) Refactored to use logic instead, and specify any remaining sites in a simple list called `special_sites` that's empty by default.

I also changed it to systematically _uninstall_ files for all sites with the exceptions of those that go through `custom_sites.yml`.

While this is mostly a refactor, I did make some changes when I noticed envs were signed up for sites they shouldn't actually be getting, etc. Hopefully that's pretty clear in each commit.

Tested in check mode on a production, staging, swiss, and icds


ansible/deploy_proxy.yml
```diff
@@ -70,7 +102,7 @@
         hosts_group: "{{ 'riakcs_new' if 'BLOB_DB_MIGRATING_FROM_S3_TO_S3' in localsettings else 'riakcs' }}"
         riakcs_proxy_port: "{{ temp_riakcs_proxy_port }}"
         site_config: {name: riakcs, vars_file: riakcs}
-      when: "active_sites.riakcs == True
+      when: "groups.get('riakcs')
```
snopoke: should this not also mimic the `riakcs site` task above and use `set_site_present`?

ansible/deploy_proxy.yml
```diff
@@ -96,7 +128,7 @@
         tasks_from: custom_site
```
snopoke: Seems like we could get rid of this task file since it's just duplicating work done elsewhere. I know I created that file (https://github.com/dimagi/commcare-cloud/commit/b2c3b5a458c4e0aa3492ad93e0779afbeb356336) but I don't see why we can't treat them the same as the other `special_sites`.

I think it was like this because they use a separate inventory group but I don't see the necessity for that.

ansible/deploy_proxy.yml
```diff
@@ -96,7 +128,7 @@
...
       vars:
         site_config: {name: cas_ssl, vars_file: cas_ssl}
-      when: active_sites.cas_ssl == True
+      when: "'cas_ssl' in special_sites"
```
emord: How different are these from one another?

dannyroberts: I'm not sure I understand the question, so I'll answer a question I do understand, and maybe there'll be some overlap.

The impetus for the change from `active_sites` to `special_sites` was to (1) make sites whose presence could be logically determined not have to be declared and (2) to make the remaining sites declareable by simply mentioning their name in a list, rather than having to mention _all_ sites as dict keys with the value being either true or false. So

For example, for enikshay it goes from
```
active_sites:
  cchq_ssl: True
  cchq_www_redirect_secure: False
  cchq_http: True
  cchq_www_redirect_insecure: False
  cchq_http_j2me: True
  commtrack_ssl: False
  commtrack_http: False
  tableau: False
  icds_tableau: False
  wiki: False
  wiki_http: False
  motech: False
  motech2: False
  riakcs: True
  enikshay_ssl: True
```
to
```
special_sites:
  - enikshay_ssl
```
because `enikshay_ssl` is the only thing that was (2) `True` and (1) not automatically determinable.

ansible/deploy_proxy.yml
dannyroberts: Okay, it's good to hear that because I was scratching my head trying to understand why that needed a different task file. I think I'll address this in a followup PR, since each iteration on this stuff takes a bit of manual testing.

ansible/deploy_proxy.yml
dannyroberts: Ah! I think you're right. I originally decided that since I didn't understand the details behind this, I just wouldn't change the current behavior. However, I'm just noticing that the following task ("drop riak temp") has exactly the opposite logic (okay well, I think `groups.get('riakcs') and` should actually be `not groups.get('riakcs') or (...)`, in which case it _would_ be the exact opposite), so it's pretty clearly what's intended.

I'll switch this and the aforementioned following task to use `set_site_present` in a followup PR. Thanks for pointing out!

ansible/deploy_proxy.yml
dannyroberts: Offlined with @emord. To summarize:
- He left off half his question—the thing he was asking about what why we have `enikshay_ssl`, `cas_ssl`, and `pna_ssl` and why they can't share a single template (and have differences determined by variables)
- I looked at this question a little bit when doing this PR and determined they were different enough that refactoring them to share a template wasn't going to be the most straightforward thing to do and would require some individual focus
- Those files are for handling requests to "original" DNS names—so cas.commcarehq.org, etc. and my understanding is that each of these envs has a different relationship to its "original" dns name (for example, CAS still handles form submissions etc at that address, whereas maybe the other envs don't), so the template, if they did share one, would definitely have a fair bit of branching. (It's TBD in my head whether it would end up being cleaner in the end or not.)
- emord pointed out that enikshay's winding down soon and at that point there'd just be two anyway (which means either it matters less, or it's easier to refactor, or both)

That about summarizes our conversation, and it was confusing enough to me when I was looking at it earlier that I decided to document here as well.

PR #1543, opened 2018-03-29
Followup to #1542 PR feedback

dannyroberts:
Follow to https://github.com/dimagi/commcare-cloud/pull/1542.

- @millerdev just want to make sure you have a chance to review the logic in the commit touching riakcs—longer note in the commit comment
- @NitigyaS can you take a look at the commit removing swiss_ssl? More context in the commit comment


ansible/deploy_proxy.yml
```diff
@@ -125,48 +114,39 @@
...
-
-- name: SWISS Proxy
-  hosts: swiss_proxy
-  tasks:
-    - include_role:
-        name: nginx
-        tasks_from: custom_site
-      vars:
-        site_config: {name: swiss_ssl, vars_file: swiss_ssl}
-      when: "'swiss_ssl' in special_sites"
```
snopoke: is this site no needed at all?

ansible/deploy_proxy.yml
```diff
@@ -125,48 +114,39 @@
...
         name: nginx
-        tasks_from: custom_site
+        tasks_from: set_site_present
       vars:
+        site_present: "{{ 'cas_ssl' in special_sites }}"
         site_config: {name: cas_ssl, vars_file: cas_ssl}
-      when: "'cas_ssl' in special_sites"
 
 - name: Enikshay Proxy
   hosts: enikshay_proxy
```
snopoke: I still think that these can be incorporated into the `install other sites` task

dannyroberts: Just happened to see this and forget if I ever responded off-channel, but basically the thing that makes this not totally trivial is that these are run on hosts `enikshay_proxy`/`cas_proxy`/`pna_proxy` whereas the `install other sites` is  run on `hosts: proxy`. I didn't dig into it after noticing that; maybe there's enough of a system that this can be simplified and it just didn't happen to pop out at me.

snopoke: right that makes sense, thanks for the follow up

PR #1548, opened 2018-03-30
remove hosts: in task file

dannyroberts:
can only go in playbooks. Noticed it was giving a warning about these being ignored.


emord:
Oops sorry about that.

PR #1550, opened 2018-03-31
Move postgresql_dbs to postgresql.yml

dannyroberts:
This adds validation checks for `postgresql_dbs`, lets `commcare-cloud` fill in some boilerplate with default values, and lets you specify the inventory host-group name ("`hqdb2`") instead of either writing out the `inventory_hostname` (e.g. the IP address) or writing an ansible expression that did the lookup.

Checked that if I run `... ansible-playbook deploy_db.yml --tags=postgresql` and `... update-config` on master, there's no diff when deploying from this branch
- [x] icds-new
- [x] softlayer
- [x] production
- [x] staging
- [x] swiss


environments/production/public.yml
```diff
@@ -62,50 +62,9 @@ pgbouncer_pool_timeout: 1
 pgbouncer_pool_mode: transaction
 pg_query_stats: True
```
emord: Should these other pg/pgbouner config options get moved over?

dannyroberts: I think I do eventually plan to move these postgresql-related config options into `postgresql.yml`, but I wanted to focus on the most challenging part first, and then all move the smaller things into place once that's settled. I have a couple more things in mind before I do that, so I'll do leave it for a followup PR if that's okay.

dannyroberts:
> Looks reasonable to me. Do you plan to break more settings away from public.yml?

In general, it would be nice for settings to be broken down into files based on e.g. the service they primarily effect, and to have each file passed through fairly strict validation. Since these files are sort of the "interface" that others outside the organization (as well as others on our team) will use to run `commcare-cloud`, it's important that we be able to define and control the interface more carefully than we have been, and I think this kind of thing will help that.

dannyroberts:
I added a few more commits, and I've tested this PR pretty thoroughly. I'll just wait for a final review as a formality.

commcare-cloud/commcare_cloud/environment/constants.py
```diff
@@ -0,0 +1,9 @@
+import jsonobject
+
+
+class _Constants(jsonobject.JsonObject):
+    formplayer_db_name = jsonobject.StringProperty(default='formplayer')
+    postgresql_db_name = jsonobject.StringProperty(default='commcarehq')
```
snopoke: how about calling this `commcare_main_db_name`?

environments/development/postgresql.yml
```diff
@@ -0,0 +1,11 @@
+DEFAULT_POSTGRESQL_PASSWORD: commcarehq
+DEFAULT_POSTGRESQL_USER: commcarehq
+SEPARATE_SYNCLOGS_DB: True
```
snopoke: If this is `True` surely there has to be a db in `postgresql_dbs` with a django alias of `synclogs`? I think this should be `False`. 

Also in travis and bootstrap config.

Seems worth validating. Maybe also worth making a constant out of it like `formplayer` and `commcarehq`

dannyroberts: Ah, yes, I got these backwards. Thanks!

dannyroberts: > Maybe also worth making a constant out of it like `formplayer` and `commcarehq`
I've got a little more refactoring to do here so that these are all treated more similarly in general, but that's a good point. I'll try and treat this constant names like the others.

PR #824, opened 2017-07-06
Set up enikshay ansible configs

mkangia:
Just raising the first draft for feedback/review

cc: @snopoke

PR #1554, opened 2018-04-02
don't deploy commcarehq code on postgresql machines

dannyroberts:


ansible/deploy_commcarehq.yml
```diff
@@ -6,7 +6,6 @@
    - celery
    - proxy
    - pillowtop
-   - postgresql
    - couchdb2_proxy
```
emord: Could this be removed? Also wonder if formplayer could eventually be removed. It should only need to run the jar file right?

dannyroberts: Yeah these are good points. Offlined a bit with @emord. `couchdb2_proxy` was added here in this commit: https://github.com/dimagi/commcare-cloud/pull/824/commits/e3e1327a1c9b54aed00285e83da5b13d24933905, and as @emord pointed out, it looks like the issue is more that fab fails when commcarehq files aren't on `couchdb2_proxy` than that the files weren't on `couchdb2_proxy`.

As for being on `formplayer`, I think that was originally an artifact of how we were deploying it, but I agree that can probably change pretty easily. I made this issue so I wound't forget to follow up on this.

PR #1562, opened 2018-04-04
more consistent and concise group tags

snopoke:


snopoke:
yea, it's just the sorting which should prevent having lots of diffs every time this get's run

PR #1570, opened 2018-04-05
Make postgresql config db aware

dannyroberts:
This currently leaves it in a state that has both the old and new postgresql_dbs and asserts in test that they generate the same output.

~Next step is to delete the old, but replace the current assurance with a more long-term test that validates the light but still somewhat involved transformation, so that people aren't afraid to refactor it later.~ 

~Just wanted to get an idea of people's reaction to the new format.~

Added tests that freeze a subset of our env config and test that `postgresql.yml` generates the correct `postgresql_dbs` variable. This test was intentionally written on top of the commit just before deleting the original `postgresql_dbs` variable form `postgresql.yml` and then merged back into this branch, so that I know that it passes both before and after that transition. I decided to freeze a subset of all 9 environments in `environments/` just to overdo it a bit while this is being actively developed. Later on, if we find these tests more cumbersome than helpful, I'm fine deleting most of them and just keeping one or two representative examples to run the test against.


commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -1,39 +1,102 @@
 import jsonobject
+import re
+
+import six
+
+from commcare_cloud.environment.constants import constants
+
+
+def alphanum_key(key):
+    """adapted from https://stackoverflow.com/a/2669120/240553"""
```
snopoke: Would be useful to add a line saying what this is for: "Used to generate a natural sorting keys"

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -1,39 +1,102 @@
...
-        for db in self.postgresql_dbs:
+        for db in self.generate_postgresql_dbs():
             if db.host is None:
                 db.host = self.DEFAULT_POSTGRESQL_HOST
             elif db.host != '127.0.0.1':
                 db.host = environment.translate_host(db.host, environment.paths.postgresql_yml)
 
+    def generate_postgresql_dbs(self):
+        return filter(None, [
+            self.dbs.main, self.dbs.synclogs,
```
snopoke: `main` could never be `None` right?

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -45,7 +108,70 @@ class DBOptions(jsonobject.JsonObject):
...
     django_alias = jsonobject.StringProperty()
-    shards = jsonobject.ListProperty(int, exclude_if_none=True)  # [start, end] pair
     django_migrate = jsonobject.BooleanProperty(default=True)
     query_stats = jsonobject.BooleanProperty(default=False)
     create = jsonobject.BooleanProperty(default=True)
+
+    @classmethod
+    def wrap(cls, data):
+        self = super(DBOptions, cls).wrap(data)
+        if re.match('^{{\s*commcarehq_main_db_name\s*}}$', self.name):
```
snopoke: these don't seem necessary anymore

environments/icds-new/postgresql.yml
```diff
@@ -1,66 +1,64 @@
...
+        shards: [824, 926]
+        host: pgshard3
+      p10:
+        shards: [927, 1023]
+        host: pgshard3
+  custom:
+    - django_alias: icds-ucr
+      name: commcarehq_icds_aggregatedata
+      host: pgucr
+      query_stats: True
```
emord: Perhaps a note should be added about `django_migrate: True` here? It's an uncommon scenario, but necessary for this.

Also maybe custom should always have `django_migrate` default to `False` as that's the most common

environments/staging/postgresql.yml
```diff
@@ -1,33 +1,24 @@
-postgresql_dbs:
-  - django_alias: default
-    name: "{{commcarehq_main_db_name}}"
+USE_PARTITIONED_DATABASE: True
```
emord: Thoughts on defaulting this to `True` and only specifying `False` when necessary? I think only swiss has this as false and if we could move that, we could get rid of some branching logic in our form processor code to make it a bit cleaner and also reduce our tests and remove the special partitioned travis job.

@snopoke This feels a bit tech-debty thoughts on having "moving swiss to partitioned db" on that backlog?

snopoke: Sure. Feel free to put it on there.

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
dannyroberts: Nope, it's `required=True`. If you specify `main` is null with

```
main:
```
or
```
main: null
```
then you'll get this error: `BadValueError: Property main is required.` If you (1) leave it out or (2) set it to `{}`
then it takes on the (recursive) default value (which is a `MainDBOptions` with all the default values).

environments/icds-new/postgresql.yml
dannyroberts: Good idea about leaving a comment and maybe an explicit `django_migrate: True` here. I'm a bit torn about giving `custom` it's own different defaults. While I agree that both in current usage and in future usage I can think of `django_migrate` is more likely to be `False` than `True`, I think it's also confusing to have a different default for `custom`, which is meant to be the sort of most manual/explicit way to add databases when they don't fit into predetermined categories. I think in that spirit, I'm going to make `django_migrate` required but have no default, meaning you have to be explicit about it for every custom db. The thinking there is that since it's custom, there's no inference we can reliably make about it, and either assumption could create a bit of a mess, so we'll just force you to think about it for each db. That work? Thanks for the suggestion.

emord: Sounds like a good compromise to me. Sounds like this leaves less room for error

environments/staging/postgresql.yml
dannyroberts: Okay, good suggestion. I switched the default to True, and then also took the liberty of renaming the variable to `SEPARATE_FORM_PROCESSING_DBS` (but leaving it `USE_PARTITIONED_DATABASE` in `localsettings.py`).

dannyroberts:
OK, added a test that mimics what I was doing to test this as I was going along, so I'm pretty confident this doesn't change the ultimate `postgresql_dbs` var that's given over to ansible in any material way. (Order changes, since it was previously arbitrary.)

So I think this is good to go, pending final review. In a followup PR, I'll now move over all the other postgres-related vars, which should be pretty straightforward.

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -117,5 +122,34 @@ class SynclogsDBOptions(DBOptions):
...
+    django_alias = 'proxy'
+
+
+class PartitionDBOptions(DBOptions):
     shards = jsonobject.ListProperty(int, exclude_if_none=True)  # [start, end] pair
+
+
+class StrictPartitionDBOptions(PartitionDBOptions):
+    def validate(self, *args, **kwargs):
+        assert re.match(r'p\d+', self.django_alias)
```
millerdev: should this say `re.match(r'p\d+$'`? as is it will match values like `p1_hey_there`

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -1,39 +1,108 @@
...
             if db.host is None:
                 db.host = self.DEFAULT_POSTGRESQL_HOST
             elif db.host != '127.0.0.1':
                 db.host = environment.translate_host(db.host, environment.paths.postgresql_yml)
 
+    def generate_postgresql_dbs(self):
+        return filter(None, [
+            self.dbs.main, self.dbs.synclogs,
+        ] + (
+            self.dbs.form_processing.get_db_list() if self.dbs.form_processing else [None]
```
millerdev: nit: why `[None]` instead of `[]`?

.travis/environments/travis/postgresql.yml
```diff
@@ -1,11 +1,8 @@
...
-postgresql_dbs:
-  - name: commcarehq_ucr
-    django_alias: ucr
-    django_migrate: False
-  - name: "{{commcarehq_main_db_name}}"
-    django_alias: default
-  - name: "{{ formplayer_db_name }}"
+dbs:
+  synclogs: null
+  form_processing: null
```
snopoke: what's the reason for having this `null` as well as specifying `SEPARATE_X_DB`. They seem to communicate same information. It is just a check to make sure that's the intention?

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -45,7 +114,65 @@ class DBOptions(jsonobject.JsonObject):
...
+                                     key=lambda db: alphanum_key(db.django_alias))
+
+
+class FormProcessingProxyDBOptions(DBOptions):
+    name = constants.form_processing_proxy_db_name
+    django_alias = 'proxy'
+
+
+class PartitionDBOptions(DBOptions):
+    shards = jsonobject.ListProperty(int, exclude_if_none=True)  # [start, end] pair
```
snopoke: not for this PR but could be worth validating the shard config too. We do validate it in HQ (https://github.com/dimagi/commcare-hq/blob/d8fdc0e5b2f7a3200ea743da60be8f808ddd8a60/corehq/sql_db/config.py#L74-L93) but would be nice to know it's valid at this stage already.

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
dannyroberts: Yes I did mean that. I thought match implied `^` and `$`, but it looks like it just implies `^`. Go figure. Thanks!

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
dannyroberts: No good reason. I was thinking of `None` as the value of `self.dbs.form_processing`, but agree your suggestion is better. I'll change to `[]`.

.travis/environments/travis/postgresql.yml
dannyroberts: Yeah, it was originally just sort of a bridge, but I'm torn as to whether removing the explicit `SEPARATE_*` variables makes this better or just more error prone, so I decided to leave it as is and wait for inspiration. I guess I could make it only look at the `SEPARATE_*` variables and make `synclogs`/`form_processing` default to `null`, but then if you specified something there you'd still want it to fail, so that doesn't seem like much of a gain either.

I'm inclined to keep it as it is, but I'm not so attached to it and wouldn't mind having something better.

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
dannyroberts: Yeah good point. When I do it, I'm inclined to copy-paste-modify rather than try to make the two codebases have this in some shared library. I think it's not only just easier, but also the "right" thing to do, in that this code isn't meant to change, so if you _are_ changing something in the code in CommCare HQ, it's not clear that you also want to be changing it in commcare-cloud—for example if we change the format a bit, it may require a two-step rollout where the application code accepts a new format and then the config management tool produces the new format some time later when we're sure everyone's up to date and/or based on some kind of versioning scheme we don't have yet. Thus in my estimation, having the code be shared is as likely to create coupling we then have to work around as it is to save us work maintaining two copies of similar code. Let me know what you think here.

Anyway, I'll do this shortly in a followup PR.

dannyroberts:
> The test fixtures seems like quite a lot to maintain though looking at it again I don't think it's meant to be kept in sync with current envs. I guess fine for now but perhaps they should rather be named more like test cases?

Yeah that's right, these are definitely frozen snapshots and not meant to be kept up to date with their respective environments of origin. When changes are made to the `dbs` => `postgresql_dbs` spec, then these tests will have to be updated, which is also the point.

Good point on the naming. I'm going to rename these all to `2018-04-04-<env>-snapshot`. Ugly names for envs, but since they're not actually envs, that's a good visual differentiator—which will also be nice when searching by directory/filename in the editor.

commcare-cloud/commcare_cloud/environment/schemas/postgresql.py
snopoke: Yea, agree with that sentiment. Seems fine for now.

PR #1571, opened 2018-04-05
Added playbook to automate letsencrypt certificate

NitigyaS:
What it does.
1) Uses Certbot to get the certificate and uses HTTP-01 challenge. 
2) Uses SITE_HOST and NO_WWW_SITE_HOST variables.  
3) Dumps the certificate on the screen so the operator can copy paste it to vault file. 
4) copy the certs instead of creating a soft link (Soft link will be overridden by the deploy_proxy.yml )

What needs to be done. 
1) Allow using domains other than SITE_HOST which will require parameters to be passed as extra_vars.

### DIFF
--- a/ansible/letsencrypt_cert.yml
+++ b/ansible/letsencrypt_cert.yml
@@ -1,5 +1,5 @@
 ---
-# commcare-cloud <env> ansible-playbook letsencrypt_cert.yml -e SITE=[example.com] -e nginx_conf_file=[nginx_conf_file] -e subdomain=[subdomain]
+# commcare-cloud <env> ansible-playbook letsencrypt_cert.yml
 - hosts: proxy
   become: yes
   tasks:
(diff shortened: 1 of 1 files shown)


NitigyaS:
One thing I also need to add is
Prompt / fail if the certificate is older than 7 days

NitigyaS:
yes, I tried using letsencrypt and [openssl](https://docs.ansible.com/ansible/2.4/openssl_certificate_module.html) first. But I was getting some weird errors in OpenSSL and to fulfill the challenge using letsencrypt module I needed to provide the static domain name string in the challenge variable. while in our case domain name is a variable and that got quite complicated. So I switched to this method.

ansible/letsencrypt_cert.yml
```diff
@@ -0,0 +1,100 @@
...
+    - name: Install Certbot
+      apt:
+        name: "{{item}}"
+        state: present
+      with_items:
+        - software-properties-common
+        - certbot
+      ignore_errors: '{{ ansible_check_mode }}'
+
+    - name: Create Challenge Directory
```
snopoke: can't `certbot` do all of this?

NitigyaS: yeah, Certbot can do it in the complete automated way. But I do not want it to temper with our nginx configurations.

NitigyaS:
Thanks . I was thinking of using `certbot renew` . And we can remove the certs from our cron. So it do not over write that by mistake. This playbook can be used to setup it for first time.

PR #1577, opened 2018-04-08
make supervisor dependency of commcarehq not common

snopoke:
No reason why all machines have to have supervisord


dannyroberts:
I believe formplayer (and touchforms) also need supervisord?

dannyroberts:
Although maybe those roles somewhat erroneously require `commcarehq`. Might be a two wrongs makes a right kind of situation.

snopoke:
@dannyroberts good point:https://github.com/dimagi/commcare-cloud/pull/1596

PR #1587, opened 2018-04-10
Changed Version

NitigyaS:
changed version

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -53,7 +53,7 @@ postgres_s3: True
 
 aws_region: 'us-east-1'
 postgresql_backup_days: 1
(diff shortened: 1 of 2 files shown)


dannyroberts:
I think you have to change this in `postgresql.yml` now. Sorry, I should have pinged you on those PRs!

environments/production/public.yml
```diff
@@ -53,7 +53,7 @@ postgres_s3: True
 
 aws_region: 'us-east-1'
 postgresql_backup_days: 1
-postgresql_version: '9.4'
+#postgresql_version: '9.6'
```
NitigyaS: Is this be commented out or will it be over ride by the var in postgres.yml

PR #1590, opened 2018-04-10
Consolidate localsettings_private boilerplate

dannyroberts:
Move ~all the boilerplate settings out of `localsettings` and into `all.yml`. You'll notice I had to change some vault files too. These are cases where there was a `localsettings_private` variable that was defined but not used previously, and now in order to retain the same behavior it had to be removed, since the boilerplate now automatically pulls from `localsettings_private`.

In all live environments (minus enikshay), I checked that this does result in any `cchq <env> update-config` changes.
- [x] production
- [x] staging
- [x] icds-new
- [x] softlayer
- [x] swiss
- [x] pna


ansible/roles/formplayer/templates/application.properties.j2
```diff
@@ -49,8 +49,12 @@ spring.jpa.hibernate.ddl-auto
 
 smtp.host={{ localsettings.EMAIL_SMTP_HOST }}
 smtp.port={{ localsettings.EMAIL_SMTP_PORT }}
-smtp.username={{ localsettings.EMAIL_LOGIN }}
-smtp.password={{ localsettings.EMAIL_PASSWORD|default('dummy') }}
+smtp.username={{ EMAIL_LOGIN }}
+{% if EMAIL_PASSWORD or EMAIL_PASSWORD == 'None' %}
```
millerdev: `EMAIL_PASSWORD == 'None'` seems redundant here. Is there some strange ansible evaluation going on that would make the string `'None'` evaluate to false?

dannyroberts: Oops, I meant to remove this, sorry.

This and a couple other variables were set in various vault files to `None`, which yaml interprets not as python `None`, but as the string `'None'`. I ended up changing these to use `null` as clearly intended, but apparently forgot to remove this workaround I played with when I was trying to figure out what was going on. I'll remove.

PR #1598, opened 2018-04-12
include from __future__ import * in localsettings

nickpell:
Tested on staging - no issues observed yet.  Including ```division``` and ```print_function``` is a little paranoid, but I'd rather air on the side of caution because ```modernize``` won't pass over this file.

@dimagi/py3


millerdev:
I can't wait for the future, hit me baby! 
```py
from __future__ import *
```

kaapstorm:
@millerdev about that: https://www.youtube.com/watch?v=n7nFEnFtvCM

PR #1602, opened 2018-04-13
automate couchdb migration

snopoke:
still a few rough edges on this but probably not going to do a lot more right now.


commcare-cloud/commcare_cloud/environment/main.py
```diff
@@ -37,10 +38,36 @@ def check(self):
         self.proxy_config
         self.create_generated_yml()
 
+    @property
+    def env_name(self):
```
dannyroberts: I'm almost tempted to call this `get_env_name_for_logging` so no one is ever tricked into using this value instead of `deploy_env` for anything other than that. I'm not going to be militant about it, but do think that would reduce future errors.

commcare-cloud/commcare_cloud/environment/paths.py
```diff
@@ -15,57 +15,60 @@
...
-        self.environments_dir = environments_dir
+        self.environments_dir = environments_dir or ENVIRONMENTS_DIR
+
+    def get_env_file_path(self, filename):
+        return os.path.join(self.environments_dir, self.env_name, filename)
 
     @lazy_immutable_property
     def public_yml(self):
-        return os.path.join(self.environments_dir, self.env_name, 'public.yml')
+        return self.get_env_file_path('public.yml')
```
dannyroberts: Thanks this is nicer :)

dannyroberts:
Also, can you give me some background on the change, what's motivating it now particularly, etc., what your end goal is? (Or point to a previous place you wrote that that I missed.)

commcare-cloud/commcare_cloud/environment/main.py
snopoke: Good point, I've switched to using `deploy_env` and removed this altogether: 310b5a4

snopoke:
@dannyroberts sorry for the lack of context. I used this last weekend to move couchdb to a new set of nodes. This was based off the migration doc here: https://docs.google.com/document/d/1Xt5Ut6ZPjf0ZhbW6pFGUlm9pDP5IHvHzuSJnzSez2ag/edit#

Basically I wanted to be able to automate as much of the migration as possible and the `couchdb-cluster-admin` tool didn't have enough info about the setup to do that so it seemed like doing it in `commcare-cloud` was the right place and using the `couchdb-cluster-admin` as a tool library.

My end goal is to have the tool be able to do the following:
* migrate from one cluster to another (where clusters are in separate envs)
* extend / reconfigure an existing cluster

This does still need some more work to polish it up and some docs but considering it makes some changes to core `commcare-cloud` pieces I thought it would be to get it in sooner rather than later.

snopoke:
@dannyroberts this is what I did to migrate couch: https://github.com/dimagi/commcare-cloud/pull/1602/files#diff-3577629c58eaa2b079cf242124ad6fab

dannyroberts:
Nice, well, pretty awesome stuff!

PR #1617, opened 2018-04-16
Add sumologic_logs_queue to prod and softayer

proteusvacuum:
I added this to the repeat records queue as this also uses gevent, and does similar "forwarding stuff"


proteusvacuum:
Based on this comment, I think this is the right thing to do: https://github.com/dimagi/commcare-hq/pull/19767#discussion_r174094038

PR #1620, opened 2018-04-17
Ns/icds new backup request

NitigyaS:
Added Postgresql nodes in netvault . Because we do not have all standby running.

### DIFF
--- a/environments/icds-new/inventory.ini
+++ b/environments/icds-new/inventory.ini
@@ -890,7 +890,7 @@ formplayer0
 [netvault_backup:children]
 pg_standby
 couchdb2
-postgresql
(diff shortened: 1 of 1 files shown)


snopoke:
Don't you think we should setup standby nodes?

NitigyaS:
Yes, I think we should.  NIC is not taking backups of all shards as of now and for those that they are doing that's only half / corrupted backup. Do we have enough storage to provision standbys for all shards?

snopoke:
Yes we should if we delete the old couch and PG nodes

NitigyaS:
Closing since I have setup Standbys.

PR #1621, opened 2018-04-17
monitor postgresql receiver process

snopoke:


ansible/roles/datadog/tasks/add_integrations.yml
```diff
@@ -44,6 +44,7 @@
     - {"name": "nginx", "enabled": "{{ datadog_integration_nginx }}"}
     - {"name": "pgbouncer", "enabled": "{{ datadog_integration_pgbouncer }}"}
     - {"name": "postgres", "enabled": "{{ datadog_integration_postgres }}"}
+    - {"name": "process", "enabled": "{{ datadog_integration_postgres_receiver }}"}
```
dannyroberts: Can you remind me how this name ("process") gets used? On the face of it, I wouldn't see "process" and think "postgresql receiver process"

snopoke: Yea, it's not idea. As far as I can tell Datadog requires the file to be called `process.yaml`. I thought about separating `name` and `filename` but that could lead to multiple integrations writing to `process.yaml`.

PR #1624, opened 2018-04-17
standby reads for users and accounting

calellowitz:
@snopoke 
As discussed in https://github.com/dimagi/commcare-hq/pull/20201 this adds load balancing for users and accounting since they get hit a lot, but don't necessarily need to be up to date. I can continue to add more as I look through the apps.


snopoke:
@calellowitz I assume you'll merge this when ready

calellowitz:
yeah, i was waiting for the HQ pr to be merged first

PR #1629, opened 2018-04-18
command to manage downtime

snopoke:
Every time I have to do maintenance I want this.

buddy @gbova

### DIFF
--- a/commcare-cloud/commcare_cloud/commands/ansible/downtime.py
+++ b/commcare-cloud/commcare_cloud/commands/ansible/downtime.py
@@ -72,7 +72,7 @@ def start_downtime(environment, ansible_context, args):
             print_downtime(downtime)
         go_down = ask("Do you want to continue?")
     else:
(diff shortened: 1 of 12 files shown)


commcare-cloud/commcare_cloud/commands/ansible/downtime.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def run(self, args, unknown_args):
+        environment = get_environment(args.env_name)
+        environment.create_generated_yml()
+        ansible_context = AnsibleContext(args)
+
+        if args.action == 'start':
+            start_downtime(environment, ansible_context, args)
+
+        if args.action == 'check':
```
dannyroberts: When might you use check?

snopoke: I had thought it could be useful but can't really think of a good use case right now so I've removed it.

mkangia:
omg!

PR #1636, opened 2018-04-20
Add command to dump data

mkangia:
@dannyroberts 
this is how i thought of adding the ability to dump domain data.
One can choose to use this command to dump data as they want.
The documentation in the class shares example of what all this provides.

We can also choose to support these options in HQ but even then this would stay unchanged here in `commcare-cloud` so skipping adding more flexibility in HQ.

Corresponding PR in HQ https://github.com/dimagi/commcare-hq/pull/20339

ToDo:
- [ ] Merge naming for dumpers and exporters

### DIFF
--- a/commcare-cloud/commcare_cloud/commands/ansible/data_dump.py
+++ b/commcare-cloud/commcare_cloud/commands/ansible/data_dump.py
@@ -1,116 +1,27 @@
 from commcare_cloud.commands.inventory_lookup.inventory_lookup import DjangoManage
-from commcare_cloud.environment.main import get_environment
 
 
 class DumpData(DjangoManage):
     """
-    Single place to support taking data dumps for a domain from various databases.
     It's recommended to always run this in a tmux session by passing the option --tmux because
     the dumping process can take quite long for domains with large amounts of data
-    commcare-cloud softlayer dump-data enikshay all --tmux
-    You can also run things in parallel by running commands with different options like
-    1. commcare-cloud softlayer dump-data enikshay sql --tmux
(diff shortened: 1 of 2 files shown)


mkangia:
@snopoke the intention was to present a better interface via `commcare-cloud` and add the ability for someone to manage their own cluster via `commcare-cloud` without knowing about the commands to be run in commcare-hq. 
Dumping a projects data would be something that would be needed when a project would want to migrate to different servers or just taking dumps like we did for enikshay and thats where this would come in handy.

dannyroberts:
@mkangia I'm still trying to understand what happens here. Is there anything here that we couldn't do by making a nice `dump_data` management command, which had basically the same interface as this and did the right thing for each database and object model? (I also don't totally get the interface, so I think I would have feedback about it once I understand it better.)

If there's nothing here that can't be done in a management command, I would prefer to move all of that out of here and into that management command. Then we could either run

```
commcare-cloud <env> django-manage dump_data <args>
```

or if we do decide to get fancy and add a special command for it (not sure here)

```
commcare-cloud <env> dump-data <args>
```

Essentially my preference would be to push as much of this logic as possible into the management command.

mkangia:
@snopoke @dannyroberts any thoughts on this?

snopoke:
I agree with @dannyroberts that this should rather be in CommCare. We already have a mechanism for calling management commands so if there was a command like this in CommCare we could just call it from commcare-cloud.

mkangia:
Cool. Thank you @snopoke. Will update.

mkangia:
Thanks @snopoke

PR #1639, opened 2018-04-20
Get to `pip install commcare-cloud`

dannyroberts:
This PR goes a long way to making it so others can `pip install commcare-cloud`—possibly everything except actually uploading it to pypi.

In addition you can now install all the python deps by just cloning the repo, entering it, entering a new virtualenv, and running

```
pip install -e .
```
if you want your install to update as you update your checked out code or
```
pip install .
```
if you want your install to be frozen to that point in time. (This is more like pypi install; it installs from a wheel.)

To install the ansible galaxy dependencies, you can then run
```
manage-commcare-cloud install
```
Notably, this installs them _in the virtualenv_ rather than in the previous location (`~/.ansible/roles`). Also, nothing in the install depends on `~/.commcare-cloud` anymore, 

To generate a script to source from your `~/.bash_profile` to get the global install experience w/ bash completion, run

```
manage-commcare-cloud configure
```

There may be a couple more things to do here in terms of smoothing the transition for our team, but I think this is a much cleaner setup.

This also got me thinking about the different ways we want to manage installs, and I think they are meaningfully different:

1. dev install: probably the egg install, where your install changes as your checked-out code changes. Ideally this is as independent from how someone is managing their virtualenv (what it's named, whether it was created with virtualenvwrapper or not, etc.) as possible.
2. third party client install: install from pypi
3. commcare-cloud install: installed in a very consistent way across all machines, totally self-managed by the script, everything from the creation of the virtualenv and what it's named, to where the code lives.

Right now 1 and 3 are the same, both managed by the control machine init script. I'd like to continue in the direction that this PR starts, by making the pip install process simple and comprehensive, and then being okay with having those 3-4 lines included both in the README instructions and the control init script. Thoughts?


czue:
don't really feel like I have the context to do this justice but conceptually it all sounds great!

millerdev:
I like the direction you're taking this. I've recently had issues I could not easily resolve with ansible-galaxy requirements, and it sounds like those should be fixed by this PR. The `~/.commcare-cloud` directory has annoyed me for a long time. Several times I got into a state where the symlinks were no longer correct and I had to manually delete them and setup new ones. Maybe there was an automated way to do that and I just didn't know about? Anyway, I'm happy to see that go.

src/commcare_cloud/environment/paths.py
```diff
@@ -4,11 +4,24 @@
...
-FABFILE = os.path.join(REPO_BASE, 'fabfile.py')
+
+def get_virtualenv_path():
+    return os.path.dirname(sys.executable)
+
+
+def get_virtualenv_site_packages_path():
+    for filepath in sys.path:
+        if filepath.startswith(os.path.dirname(get_virtualenv_path())) and filepath.endswith('site-packages'):
+            return filepath
```
millerdev: A more conventional way to get the site-packages path is outlined here: https://stackoverflow.com/a/122340/10840
```py
from distutils.sysconfig import get_python_lib
print(get_python_lib())
```

src/commcare_cloud/environment/paths.py
```diff
@@ -4,11 +4,24 @@
...
 from memoized import memoized_property, memoized
 
-REPO_BASE = os.path.expanduser('~/.commcare-cloud/repo')
-ANSIBLE_DIR = os.path.join(REPO_BASE, 'ansible')
-ENVIRONMENTS_DIR = os.environ.get('COMMCARE_CLOUD_ENVIRONMENTS', os.path.join(REPO_BASE, 'environments'))
-FAB_DIR = os.path.join(REPO_BASE, 'fab')
-FABFILE = os.path.join(REPO_BASE, 'fabfile.py')
+
+def get_virtualenv_path():
+    return os.path.dirname(sys.executable)
```
millerdev: I think this should be, although not sure it's actually needed, given the comment below on `get_virtualenv_site_packages_path`.
```py
return os.environ["VIRTUAL_ENV"]
```

src/commcare_cloud/environment/paths.py
dannyroberts: Oh, nice! Thanks for pointing that out. Updated

src/commcare_cloud/environment/paths.py
dannyroberts: Replaced `get_virtualenv_site_packages_path` with the built-in (thanks!), but I still do need to keep this function around, because it's used in a couple other places.

I didn't know about `VIRTUAL_ENV`, but that's actually exactly what I was trying to work _around_ with this way of fetching that dir. That env variable is set when the user is has entered the virtual environment, but _not_ set when the user runs an executable within the virtualenv `bin` directory directly.

Also, I'm realizing from your comment that this function is poorly named. It returns the path to the `bin` directory, so I'm going to name it accordingly.

millerdev: A bit of context helps so much! Maybe add a docstring to this function that explains why it's not using `os.environ["VIRTUAL_ENV"]` (because it's not expected to be set).

millerdev:
> I did put one new thing in ~/.commcare-cloud ... would you advocate for doing this differently?

Hmm, I guess it depends on how this ages? If it breaks again and is not easy to fix then I might get annoyed with it. If it's something I never need to change then it's fine.

One thing I'm not wild about is this:
```sh
export PATH=$PATH:/Users/droberts/.virtualenvs/commcare-cloud/bin
```
That will change the user's default Python interpreter (and pip and anything else in there) to the one in that virtualenv, which is probably not what they want. I'm not sure what the best alternative is though.

dannyroberts:
> That will change the user's default Python interpreter (and pip and anything else in there) to the one in that virtualenv, which is probably not what they want. I'm not sure what the best alternative is though.

I thought about this, and because I (intentionally) put this new bin directory at the _end_ of the `PATH`, my understanding is that it won't overshadow anything you already have elsewhere, but it will make new executables available:

```
droberts$ echo $PATH
/Users/droberts/.rbenv/shims:/opt/local/bin:/opt/local/sbin:/opt/local/bin:/opt/local/sbin:/opt/local/bin:/usr/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/opt/X11/bin:/usr/local/git/bin:/usr/local/go/bin:/opt/lessc/bin:/usr/local/sbin:/Users/droberts/gpg-scripts:/Users/droberts/.virtualenvs/commcare-cloud/bin
droberts$ which pip
/usr/local/bin/pip
droberts$ which python
/usr/local/bin/python
droberts$ which ansible-galaxy
/usr/local/bin/ansible-galaxy
droberts$ which commcare-cloud
/Users/droberts/.virtualenvs/commcare-cloud/bin/commcare-cloud
droberts$ which cchq
/Users/droberts/.virtualenvs/commcare-cloud/bin/cchq
droberts$ which manage-commcare-cloud
/Users/droberts/.virtualenvs/commcare-cloud/bin/manage-commcare-cloud
```

You can see (1) the virtualenv bin directory is at the end of the `PATH` (2) `pip`, `python`, and `ansible-galaxy` all point to the non-virtualenv one, even though these also exist within the virtualenv, and (3) `commcare-cloud`, `cchq`, and `manage-commcare-cloud` all refer to the executables within the virtualenv (because they don't exist elsewhere).

dannyroberts:
@dimagi/team-commcare-hq 

Hey everyone, next time you want to use `commcare-cloud` (i.e. deploy, etc.), after you update the code you will have to run `source control/init.sh` again, or if you're on the control machine you can just run `update-code` and maybe log out and log back in again to make sure it worked.

If you want to update it manually (if there's any trouble with the init script locally for example, you can run (in your virtualenv)

```
pip uninstall commcare-cloud
pip install -e .
manage-commcare-cloud install
manage-commcare-cloud configure
```

Please let me know of any problems you have along the way so that I can fix them before anyone else has them.

proteusvacuum:
For `zsh` users, you should add
```
autoload -U bashcompinit
bashcompinit
```
before `source ~/.commcare-cloud/load_config.sh` in your config files.

src/commcare_cloud/environment/paths.py
```diff
@@ -1,14 +1,30 @@
...
+    This is meant to work even when the user isn't inside a virtualenv,
+    but is directly running an executable that lives in the virtualenv bin,
+    so `os.environ["VIRTUAL_ENV"]` does not work here.
+
+    """
+    return os.path.dirname(sys.executable)
+
+
+PACKAGE_BASE = os.path.realpath(os.path.join(os.path.dirname(__file__), '..'))
+ANSIBLE_ROLES_PATH = os.path.realpath(os.path.join(get_python_lib(), '.ansible/roles'))
```
snopoke: @dannyroberts can you explain why we chose to put the roles path in the virtual env?

dannyroberts: I think it was under this idea
> This PR goes a long way to making it so others can pip install commcare-cloud—possibly everything except actually uploading it to pypi.
that we should package it in such a way that we could eventually push it to pypi.

Imagining commcare-cloud were in pypi, and were installed via
```
pip install commcare-cloud
```
we would then want to be able to run `manage-commcare-cloud install` and `manage-commcare-cloud configure` and have the effect of those things be contained to the virtualenv directory as well (since we would no longer have a commcare-cloud source directory to play with).

PR #1641, opened 2018-04-23
Check shell

kaapstorm:
This change allows this script to play nicely with shells other than BASH (which is great and all, but not quite [as great as zsh](https://github.com/robbyrussell/oh-my-zsh))

@emord @dannyroberts

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -1,10 +1,16 @@
-#! /bin/bash
+# This script must be run from BASH with "source control/init.sh".
 
-if [[ $_ == $0 ]]
+if [ "$_" = "$0" ]
 then
     echo "Please run this script as follows:"
     echo "    $ source /path/to/repo/control/init.sh"
     exit 1
+fi
+
+if [ "$0" != "bash" ]
+then
+    echo "Please run this script from the BASH shell."
+    return
 fi
 
 


control/init.sh
```diff
@@ -1,12 +1,18 @@
-#! /bin/bash
+# This script must be run from BASH with "source control/init.sh".
```
kaapstorm: A sourced script douesn't need a shebang because it runs in the shell you're in.

control/init.sh
```diff
@@ -1,12 +1,18 @@
...
 
-if [[ $_ == $0 ]]
+if [ "$_" = "$0" ]
```
kaapstorm: Makes this `if` statement POSIX compliant so it doesn't error if you're not in BASH.

control/init.sh
```diff
@@ -1,12 +1,18 @@
...
 then
     echo "Please run this script as follows:"
     echo "    $ source /path/to/repo/control/init.sh"
     exit 1
 fi
 
+if [ "$0" != "bash" ]
+then
+    echo "Please run this script from the BASH shell."
+    return
```
kaapstorm: `return` not `exit` because if the script is sourced then `exit` will exit your shell and not just this script.

snopoke:
build failed, not sure if related

control/init.sh
millerdev: ❤️ my future self thanks you for thinking ahead on that one.

emord:
Interesting. I don't recall having any issues running this locally, but maybe I ran an old version or did something weird to get it to work in ZSH

dannyroberts:
Underlying error in the travis failure is
```
$ source control/init.sh
Please run this script from the BASH shell.
```

PR #1640, opened 2018-04-23
Revert "Revert "standby reads for users and accounting""

calellowitz:
Reverts dimagi/commcare-cloud#1637
@snopoke @emord 
after https://github.com/dimagi/commcare-hq/pull/20360 gets in

PR #1643, opened 2018-04-23
Fix formplayer database definition

emord:
@calellowitz This was concatenating the two main databases in the formplayer properties.

@dannyroberts not sure if you have a better idea of how to filer these sorts of configs

(going to go ahead and merge since ICDS localsettings are currently broken

### DIFF
--- a/ansible/roles/formplayer/vars/main.yml
+++ b/ansible/roles/formplayer/vars/main.yml
@@ -3,7 +3,7 @@ formplayer_data_dir: "{{ encrypted_root }}/formplayer"
 # formplayer_sentry_dsn: ..defined in secret file
 # formplayer_db_name: ..defined in secret file
 hq_main_db_host: "{% for config in postgresql_dbs -%}
-  {%- if config.name == commcarehq_main_db_name -%}
+  {%- if config.name == commcarehq_main_db_name and config.django_migrate -%}
     {{ config.host }}
   {%- endif -%}
 {%- endfor %}"


ansible/roles/formplayer/vars/main.yml
```diff
@@ -3,7 +3,7 @@ formplayer_data_dir: "{{ encrypted_root }}/formplayer"
 # formplayer_sentry_dsn: ..defined in secret file
 # formplayer_db_name: ..defined in secret file
 hq_main_db_host: "{% for config in postgresql_dbs -%}
-  {%- if config.name == commcarehq_main_db_name -%}
+  {%- if config.name == commcarehq_main_db_name and config.django_migrate -%}
```
dannyroberts: You might be able to get away with `config.django_alias == 'default'`? (Not sure how this other new db is configured, so I'm not sure.) It's a little sad because kind of what you would like to do here is simply

```
hq_main_db_host: {{ postgresql_dbs_by_name.main.host }}
```
where `postgresql_dbs_by_name` is just the `dbs` variable we've already defined in `postgresql.yml` (but with the defaults from the schema filled in). We already have the information we want in this exact format, but we throw it away before it gets to ansible. We could pretty easily pass it in as well, so that we have essentially the same information in two different formats being passed to ansible, though there are annoying things about that as well.

Those are my thoughts on the topic. If you want to keep it the way it is for now, that's fine, but I think just throw in a comment to say why it's set this way and what it's trying to do.

emord: > Not sure how this other new db is configured, so I'm not sure

This came up because we just added load balancing for some of the main db's apps. It's passed in here: https://github.com/dimagi/commcare-cloud/pull/1640/files#diff-a3e25b2e7f211966578413179699ab81R85

It would be nice if we could also use the standby here, but I think that may be out of scope.

What I'm trying to do here is "the main db that isn't a standby", but I didn't see a better way to do it. Something I tried to do was "first db with this name" but you can't easily use a "break" statement in jinja and I also didn't want this to be the standby if the standby happens to be in the list before the main db

Does that give you a better idea of the issue?

PR #1646, opened 2018-04-24
Update docs

dannyroberts:
There's more to come in the coming days, but I thought I'd get started with what I did yesterday.

Note also that in the course of documenting I noticed some holdovers in how defaults were defined and changed them to fit into a more centralized schema approach.

### DIFF
--- a/src/commcare_cloud/fab/operations/supervisor.py
+++ b/src/commcare_cloud/fab/operations/supervisor.py
@@ -80,8 +80,8 @@ def set_celery_supervisorconf():
             )
             continue
 
-        pooling = params.get('pooling', 'prefork')
-        max_tasks_per_child = params.get('max_tasks_per_child', 50)
+        pooling = params['pooling']
+        max_tasks_per_child = params['max_tasks_per_child']
         num_workers = params.get('num_workers', 1)
 
         params.update({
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/fab/operations/supervisor.py
```diff
@@ -80,8 +80,8 @@ def set_celery_supervisorconf():
             )
             continue
 
-        pooling = params.get('pooling', 'prefork')
-        max_tasks_per_child = params.get('max_tasks_per_child', 50)
+        pooling = params['pooling']
+        max_tasks_per_child = params['max_tasks_per_child']
```
gcapalbo: is `params` now an instance of `CeleryOptions`?

dannyroberts: It's currently getting generated here https://github.com/dimagi/commcare-cloud/blob/e0e133bfa4263d168659b494975525d6cda545e9/src/commcare_cloud/fab/fabfile.py#L158 and then set on envs with `env.update(vars)` a few lines below.

once it gets to these lines I'm changing, params is essentially a `CeleryOptions` converted to json (via `.to_json()`).

PR #1651, opened 2018-04-24
adding manage role to deploy just to webworkers[0]

jmtroth0:
@dannyroberts @esoergel  I've mentioned this to both of you, but it seems like a small nicety so we aren't setting up a release on every machine just to run a management command. 

Let me know if you hate this pattern though

`w=1` is definitely the way to go for this review


esoergel:
Nice!  I like this pattern.  Might want to add a comment about how to call it somewhere, or maybe an alias that provides the parameter or something.  Code looks fine, but I'm not as familiar with this codebase, so I'll let Danny do a pass.

src/commcare_cloud/fab/fabfile.py
```diff
@@ -550,13 +560,14 @@ def unlink_current():
...
 
 
-def copy_release_files():
-    execute(release.copy_localsettings)
-    execute(release.copy_tf_localsettings)
-    execute(release.copy_formplayer_properties)
-    execute(release.copy_components)
-    execute(release.copy_node_modules)
-    execute(release.copy_compressed_js_staticfiles)
+def copy_release_files(manage=False):
```
snopoke: The name of this arg is a bit meaningless here. Maybe rename to something like `full_cluster=True`?

src/commcare_cloud/fab/fabfile.py
```diff
@@ -221,6 +221,7 @@ def env_common():
     all = servers['all']
     proxy = servers['proxy']
     webworkers = servers['webworkers']
+    manage = webworkers[0]
```
emord: thoughts on making this `servers.get('manage', webworkers[0])` in case we want it to be overridden? (for example on ICDS, webworkers[0] is also the proxy and already has some high memory, so may want to do it somehwere else

src/commcare_cloud/fab/fabfile.py
```diff
@@ -383,14 +385,20 @@ def prepare_offline_deploy():
...
+@task
+def setup_release(keep_days=0, full_cluster=True):
     """
     Setup a release in the releases directory with the most recent code.
     Useful for running management commands. These releases will automatically
     be cleaned up at the finish of each deploy. To ensure that a release will
     last past a deploy use the `keep_days` param.
 
     :param keep_days: The number of days to keep this release before it will be purged
+    :param manage: If True, only setup on webworkers[0] where the command will be run
```
emord: this doesn't match the paramater for the function

src/commcare_cloud/fab/fabfile.py
```diff
@@ -383,33 +385,47 @@ def prepare_offline_deploy():
...
+@task
+def setup_release(keep_days=0, full_cluster='True'):
     """
     Setup a release in the releases directory with the most recent code.
     Useful for running management commands. These releases will automatically
     be cleaned up at the finish of each deploy. To ensure that a release will
     last past a deploy use the `keep_days` param.
 
     :param keep_days: The number of days to keep this release before it will be purged
+    :param keep_cluster: If True, only setup on webworkers[0] where the command will be run
```
dannyroberts: I think this is supposed to say `full_cluster`

src/commcare_cloud/fab/fabfile.py
```diff
@@ -386,11 +386,11 @@ def prepare_offline_deploy():
 
 @task
 def setup_limited_release(keep_days=1):
-    setup_release(keep_days, False)
+    setup_release(keep_days, 'False')
```
esoergel: nit: I'd name this parameter so you can read this line and know what `'False'` means - in general I'd almost always prefer to name boolean arguments, unless it's clear from the function name what it's for.

src/commcare_cloud/fab/fabfile.py
```diff
@@ -221,7 +221,7 @@ def env_common():
     all = servers['all']
     proxy = servers['proxy']
     webworkers = servers['webworkers']
-    manage = webworkers[0]
+    manage = servers.get('manage', webworkers[0])
```
esoergel: Is the idea here that the inventory file can specify a preferred "manage" machine?

jmtroth0: Yeah, emord's [suggestion](https://github.com/dimagi/commcare-cloud/pull/1651/files/5d8d4bb11edfbacb0f70e5647ab3556860dbc045#diff-4b38b47cd9db17e5e8e82f5424318f83] since on icds apparently webworkers:0 is already a bit over taxed

src/commcare_cloud/fab/fabfile.py
jmtroth0: Seems reasonable. My standard also generally includes if the function definition is right next to it, but it's probably still better to name it anyway

src/commcare_cloud/fab/fabfile.py
```diff
@@ -383,33 +385,47 @@ def prepare_offline_deploy():
 
 
 @task
-def setup_release(keep_days=0):
+def setup_limited_release(keep_days=1):
+    setup_release(keep_days, full_cluster='False')
```
dannyroberts: nit: I bristle a little at passing string values to be parsed as booleans between functions internally, versus having them accept booleans in the first place. I realize that the outermost later (the @task) has to accept string arguments, but I'd prefer to see those explicitly take a strings, and convert them immediately and call the actual doer function which doesn't concern itself with parsing string arguments. So here I guess that would look like

```python
def parse_int_or_exit(int_string, default):
     if int_string is None:
         return default
     try:
         return int(int_string)
     except ValueError:
         print(red("Unable to parse '{}' into an integer".format(keep_days)))
         exit()


def setup_limited_release(keep_days=None):
    _setup_release(
        keep_days=parse_int_or_exit(keep_days, default=1),  # I think you actually wanted a default of 1 not 0?
        full_cluster=False
    )


def setup_release(keep_days=None):
    _setup_release(
        keep_days=parse_int_or_exit(keep_days, default=0),
        full_cluster=True
    )


def _setup_release(keep_days=0, full_cluster=True):
    # all the stuff, with no string parsing

File: src/commcare_cloud/fab/fabfile.py [hunk 8]
```diff
@@ -383,33 +385,47 @@ def prepare_offline_deploy():
...
         print(red("Unable to parse '{}' into an integer".format(keep_days)))
         exit()
 
+    try:
+        full_cluster = strtobool(full_cluster)
+    except ValueError:
+        print(red("Unable to parse '{}' into a boolean".format(full_cluster)))
+        exit()
+    if not full_cluster and not keep_days:
+        keep_days = 1
```
dannyroberts: If we go the route of the example code block in my comment above, then we won't need to have this other sort of secondary default here.

src/commcare_cloud/fab/fabfile.py
```diff
@@ -383,33 +385,47 @@ def prepare_offline_deploy():
 
 
 @task
-def setup_release(keep_days=0):
+def setup_limited_release(keep_days=1):
+    setup_release(keep_days, full_cluster='False')
+
+
+@task
+def setup_release(keep_days=0, full_cluster='True'):
```
dannyroberts: Going along with the example above, I think it'd be fine to _only_ let people access the `full_cluster` functionality via the `setup_limited_release` task, rather than allowing them to either do that or pass in `:full_cluster=yes` or whatever in the command line.

src/commcare_cloud/fab/fabfile.py
```diff
@@ -221,6 +221,7 @@ def env_common():
     all = servers['all']
     proxy = servers['proxy']
     webworkers = servers['webworkers']
+    manage = servers.get('manage', webworkers[0])
```
dannyroberts: I might go with a consistent "django manage" throughout. So calling the role above `ROLES_DJANGO_MANAGE`, and calling this server group variable  `django_manage` and the optional ansible inventory group accessed on this line `django_manage` as well.

src/commcare_cloud/fab/fabfile.py
dannyroberts: I like the pattern. For it to play nice with our other tools, this should be the default for the `DjangoManage` and `Tmux` commands as well

src/commcare_cloud/fab/fabfile.py
jmtroth0: Seems fine and updated, but at some point perhaps celery, webworkers, and pillowtop should be updated too

src/commcare_cloud/fab/fabfile.py
```diff
@@ -382,34 +384,59 @@ def prepare_offline_deploy():
...
     Setup a release in the releases directory with the most recent code.
     Useful for running management commands. These releases will automatically
     be cleaned up at the finish of each deploy. To ensure that a release will
     last past a deploy use the `keep_days` param.
 
     :param keep_days: The number of days to keep this release before it will be purged
-
-    Example:
-    fab <env> setup_release:keep_days=10  # Makes a new release that will last for 10 days
+    :param full_cluster: If True, only setup on webworkers[0] where the command will be run
```
snopoke: I think this should say `If False`

PR #1653, opened 2018-04-25
use deploy_env instead of env_name for fab env.env_name

snopoke:
The only env where `deploy_env` differs from the name of the env (as in the folder name) is `icds-new`.

I've checked on ICDS and the supervisor files all use `icds` in the process name so this switch won't have any effect.


snopoke:
realized this is wrong

PR #1654, opened 2018-04-25
Migrating UCR to hqdb3

NitigyaS:


snopoke:
@emord I'd like to do that but as a separate step

PR #1655, opened 2018-04-25
add num_workers to celery config

snopoke:
I did a bit of testing it doesn't appear that the `CeleryOptions` class every wraps the raw dict. I played with it a bit but couldn't figure out how to get the wrapping to work (without manually wrapping each one).


dannyroberts:
Interesting. Well, that certainly explains some previous behavior that stumped me. I'm guessing I was wrong when I wrote it and

```
celery_processes = jsonobject.DictProperty(jsonobject.DictProperty(CeleryOptions))
```

is silently just equivalent to

```
celery_processes = jsonobject.DictProperty(dict)
```

PR #1656, opened 2018-04-25
refactor services into classes

snopoke:
easiest review by commit

Needs a bit more testing but mostly good to go


src/commcare_cloud/commands/ansible/service.py
```diff
@@ -1,328 +1,444 @@
...
+            SERVICES_BY_NAME[name]
+            for name in args.services
+        ]
+
+        ansible_context = AnsibleContext(args)
+        exit_codes = []
+        for service_cls in services:
+            service = service_cls(environment, ansible_context)
+            exit_codes.append(service.run(args.action, args.limit))
+        return max(exit_codes)
```
dannyroberts: if exit_codes is `[0, -1]` this will return `0`, but I think the right answer is to return `-1`. Maybe only append non-zero statuses?

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -1,328 +1,444 @@
...
+            help="The services to run the command on"
+        )
+        self.parser.add_argument(
+            'action', choices=ACTIONS,
+            help="What action to take"
+        )
+        self.parser.add_argument('--limit', help=(
+            "Restrict the hosts to run the command on. Use 'help' action to list all option."
+        ))
+        self.parser.add_argument('--only', help=(
```
dannyroberts: I can't find where this arg is used.

src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -122,3 +119,25 @@ def run_action_with_check_mode(run_check, run_apply, skip_check, quiet=False, al
...
+
+
+def get_celery_workers(environment):
+    for host, queues in environment.app_processes_config.celery_processes.items():
+        if not host or host == 'None':
+            continue
+        for comma_separated_queue_names, config in queues.items():
+            for queue in comma_separated_queue_names.split(','):
+                for worker_num in range(config.get('num_workers', 1)):
+                    process_name = get_celery_worker_name(environment, comma_separated_queue_names, 0)
```
mkangia: should this be `worker_num` instead of 0 for getting the name?

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -1,328 +1,444 @@
...
+    @abstractproperty
+    def inventory_groups(self):
+        raise NotImplementedError
+
+    def __init__(self, environment, ansible_context):
+        self.environment = environment
+        self.ansible_context = ansible_context
+
+    def run(self, action, host_pattern=None, process_pattern=None):
+        if action == 'help':
```
mkangia: Nice add. Does this make `service` command behave differently than others? 
Would be nice if all commands keep the same interface.

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -1,328 +1,444 @@
...
-            exit_code = self.run_supervisor_action_for_service_group(service_group, args.action, args, unknown_args)
-        return exit_code
-
-    def services(self, service_group, args):
-        if service_group != 'celery' and args.only:
-            return args.only.split(',')
+            run_on = self.get_inventory_group_for_sub_process(process_pattern)
+            hosts = self.environment.inventory_manager.get_hosts(run_on)
+            if not hosts:
+                raise NoHostsMatch
```
mkangia: Don't clearly understand the intention here? trying to handle this earlier?
Ansible already raises an error if we run something and there are no corresponding hosts to run it on. I have not checked what `supervisorctl` returns in such a case.

src/commcare_cloud/commands/ansible/service.py
snopoke: slightly, you still get the standard help if you use `-h` but this gives you more based on the env and options you've selected which I don't think argparser can do.

src/commcare_cloud/commands/ansible/service.py
snopoke: If there are no hosts then there's no need to call ansible at all. This just fails early and allows us to output a nice error message

src/commcare_cloud/commands/ansible/service.py
snopoke: I did some experiments on getting argparser to manage this help and it can be done but it's not pretty and doesn't output very well.

src/commcare_cloud/commands/ansible/service.py
snopoke: thanks: 85d3b0d

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -1,328 +1,443 @@
...
+        host_pattern = host_pattern or ','.join(self.inventory_groups)
+        service_args = 'name={} state={}'.format(self.service_name, STATES[action])
+        return self._run_ansible_module(host_pattern, 'service', service_args)
 
-    def get_inventory_group_for_service(self, service, service_group):
-        return self.INVENTORY_GROUP_FOR_SERVICE.get(service, service_group)
 
-    @staticmethod
-    def get_celery_workers_to_work_on(args):
+class MultiAnsibleService(SubServicesMixin, AnsibleService):
```
snopoke: @dannyroberts @mkangia I was thinking about removing this class. It's only used RiakCS and Kafka but the same functionality can be accomplished by passing multiple services to the command line and it adds quite a lot of code:

I also don't think it's that common that you want to manage these services as a single unit. 

```
# current
cchq <env> service riakcs restart
cchq <env> service riakcs restart --only riakcs

# If we split them up
cchq <env> service riakcs riak stanchion restart
cchq <env> service riakcs restart
```

snopoke:
@mkangia I've added some usage examples which get printed in the help: eeab115

Regarding the complexity, the reason I started this is that I wanted to add pillowtop to the list of managed services and struggled to find all the places I needed to change in order to do that. I've now [added it](e0d583a). I've also extracted out the only real pieces of functionality into 2 functions which I've added tests for. The rest is really just the service definitions which are only a few lines each.

Are there specific pieces that you think are overly complex?

dannyroberts:
merged master and fixed the test that makes sure the command docs are up to date with the command help text (and this is exactly the kind of situation they were meant to help out with :))

src/commcare_cloud/commands/ansible/service.py
dannyroberts: There are certainly ways to change the help text, with only minor hacks 😉. Here are some examples of where we already do this
- https://github.com/dimagi/commcare-cloud/blob/abf7a894f8d1ae0ff53f4eb6f183159783ad81cd/src/commcare_cloud/commands/fab.py#L16
- https://github.com/dimagi/commcare-cloud/blob/abf7a894f8d1ae0ff53f4eb6f183159783ad81cd/src/commcare_cloud/commands/ansible/ansible_playbook.py#L38

You could probably just use the second method (with `add_to_help_text`) to append what's now printed with `help` to the the default help text. Maybe that's what you already figured out, and I agree making that pretty is probably an unbounded effort

snopoke:
@dannyroberts updated

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -1,328 +1,540 @@
...
+class Kafka(MultiAnsibleService):
+    name = 'kafka'
+    service_process_mapping = {
+        'kafka': ('kafka-server', 'kafka'),
+        'zookeeper': ('zookeeper', 'zookeeper')
+    }
+
+
+class Postgres(MultiAnsibleService):
+    name = 'postgres'
```
dannyroberts: Should this say 'postgresql' to match the inventory host name? Or did you change this intentionally?

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -1,328 +1,540 @@
...
+    service_process_mapping = {
+        'postgres': ('postgresql', 'postgresql'),
```
dannyroberts: I think in this change `pg_standby` is dropped. It was originally its own separate service group I think, which would now look something like this:

```
class PgStandby(MultiAnsibleService):
    name = 'pg_standby'
    service_process_mapping = {
        'postgresql': ('postgresql', 'pg_standby'),
        'pgbouncer': ('pgbouncer', 'pg_standby')
    }
```

src/commcare_cloud/commands/ansible/service.py
snopoke: I did change this intentionally since it's a few less characters to type and since it also isn't actually postgresql anymore but a combined service. Fine to change back though if you don't think it's a good change

docs/commcare-cloud/commands/index.md
```diff
@@ -415,9 +415,11 @@ Manage services.
 
 Usage examples:
    cchq <env> service postgresql status
+   cchq <env> service riakcs restart --only riak,riakcs
    cchq <env> service celery help
    cchq <env> service celery restart --limit <host>
    cchq <env> service celery restart --only <queue-name>
```
dannyroberts: I think you need to remove this line for command doc tests to pass (although there seems to be another real test failure)

snopoke: :+1: also had to fix something else. OK, I'm really done now. Going to bed.

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -1,328 +1,551 @@
...
+    @property
+    def inventory_groups(self):
+        return [
+            inventory_group
+            for service, inventory_group in self.service_process_mapping.values()
+        ]
+
+    @property
+    def managed_services(self):
+        return [name for name in self.service_process_mapping]
```
mkangia: nit: `self.service_process_mapping.keys()` ?

src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -122,3 +120,54 @@ def run_action_with_check_mode(run_check, run_apply, skip_check, quiet=False, al
...
+    for host, pillows in environment.app_processes_config.pillows.items():
+        for name, params in pillows.items():
+            start = params.get('start_process', 0)
+            num_processes = params.get('num_processes', 1)
+            for num_process in range(start, start + num_processes):
+                process_name = "commcare-hq-{deploy_env}-pillowtop-{pillow_name}-{num_process}".format(
+                    deploy_env=environment.meta_config.deploy_env,
+                    pillow_name=name,
+                    num_process=num_process
+                )
```
mkangia: There was some work in progress for moving pillowtop out of fab to ansible by Emord when the service management was being added. Adding pillowtop was avoided at that moment so that this naming convention for the process name is not at two different places. Like for celery `get_celery_worker_name` was added and was then made the single location for celery worker name.
I think we should consider having something similar for pillowtop as well so that any changes are reflected at both places.

mkangia:
thanks for the changes @snopoke 

>  I wanted to add pillowtop to the list of managed services and struggled to find all the places I needed to change in order to do that

Hmm.. pillow top was kept on hold for [this](https://github.com/dimagi/commcare-cloud/pull/1656#discussion_r184626545) when we were adding this command. Not sure why it was difficult to find places to change but things are definitely looking cleaner/better now and should be easy to make changes/additions in future. Thanks for the clean up :)

PR #1657, opened 2018-04-25
Start documenting commands

dannyroberts:
This also includes basic nosetests that will warn us if commands, their help text, or their aliases are missing from this file, as just a very basic nudge to help us keep them in sync.

You can review the main addition in a prettier form here: https://github.com/dimagi/commcare-cloud/blob/d27b1ca6bf315c481f44fece73b984843125b4be/docs/commcare-cloud/commands/index.md


dannyroberts:
@czue Yeah, that definitely occurred to me, but I couldn't immediately think of a way to do that cleanly. The help text on each argument in particular is frustratingly inaccessible at runtime. There are potential ways around that, and then I'd have to work out how to piece things together so that the doc generation process is convenient and not burdensome. But I do think that'd be a nice place to end up.

snopoke:
Nice, could do with a TOC

dannyroberts:
Thanks @snopoke https://github.com/dimagi/commcare-cloud/pull/1662. I didn't realize it was this easy.

PR #1661, opened 2018-04-26
Revert 1658 revert 1651 jr/single release

jmtroth0:
This time with better testing... i.e. `cchq staging fab deploy` works, as well as just `cchq staging fab setup_<limited_>releases`. sorry @biyeun 

Apparently I didn't account for how the decorators worked with execute and just running setup_releases and setup_limited_releases wasn't the best test. The last commit (edit: f2dc175) is the one that changes things. The others are cleaning

tagging everyone who commented on #1651 --> @snopoke @dannyroberts @esoergel @emord 
buddy: @gcapalbo

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -281,7 +281,7 @@ def preindex_views():
     Creates a new release that runs preindex_everything. Clones code from
     `current` release and updates it.
     """
-    setup_release()
+    _setup_release()
     db.preindex_views()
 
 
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/fab/fabfile.py
```diff
@@ -382,34 +384,59 @@ def prepare_offline_deploy():
     offline_ops.prepare_formplayer_build()
 
 
+def parse_int_or_exit(val):
+    try:
+        return int(val)
+    except ValueError:
```
gcapalbo: this may also raise `TypeError` for example if `val` is `None`

src/commcare_cloud/fab/fabfile.py
```diff
@@ -387,7 +387,7 @@ def prepare_offline_deploy():
 def parse_int_or_exit(val):
     try:
         return int(val)
-    except ValueError:
+    except ValueError, TypeError:
```
gcapalbo: this should be expressed as an explicit tuple: `(ValueError, TypeError)`

(in python 2, `except Exception, e` is the same as `except Exception as e`)

PR #1663, opened 2018-04-26
Static arguments

dannyroberts:
Pull out arguments so that they're defined statically on the command class as a list, rather than being added within `make_parser`. Any functionality from `make_parser` that was more complicated than that has been moved to the new `modify_parser` method, which only a few command classes have.

This lets me add a test that all arguments appear (somewhere) in the command docs. Not quite the dream of auto-generating docs that @czue mentioned on my last PR, but a small step in the right direction I think.

Opening for tests and review on general direction, will test more thoroughly that I didn't break any of the commands tomorrow.

### DIFF
--- a/docs/commcare-cloud/commands/index.md
+++ b/docs/commcare-cloud/commands/index.md
@@ -203,7 +203,7 @@ commcare-cloud <env> run-module all debug "msg={{ inventory_hostname }}"
 Run an arbitrary command via the Ansible shell module:
 
 ```
-commcare-cloud <env> run-shell-command <inventory_group> <shell_command>
+commcare-cloud <env> run-shell-command <inventory_group> <shell_command> [--silence-warnings]
 ```
 
 ##### `<inventory_group>`
@@ -215,6 +215,10 @@ See [`run-module`](#run-module).
 Command to run remotely.
 (Tip: put quotes around it, as it will likely contain spaces.)
 Cannot being with `sudo`; to do that use the ansible `--become` option.
+
+##### `[--silence-warnings]`
+
+Silence shell warnings (such as to use another module instead).
 
 ##### Example
 ```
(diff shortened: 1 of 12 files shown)


dannyroberts:
@czue I didn't know about the `parent` option. Looks like that's useful for one parser to "inherit" the  arguments of another parser or parsers. Did you kind of mean use that instead of e.g. https://github.com/dimagi/commcare-cloud/pull/1663/files#diff-630448bcf64108b4e08a87d74ec52e1cR36 (`arguments = ... + NON_POSITIONAL_ARGUMENTS`)? I think we'd have to reshuffle how the parser's built because from the examples it looks like `parents` needs to be built and available at the time that the parser is first instantiated.

@snopoke that's a really enticing suggestion, but I decided to start with this lower-touch thing of just duplicating some functionality with ~30 lines of logic (my `argument.name_in_docs`) and just using that to run some super basic (and incomplete) checks. I have been kind of wondering in general what the best way to do this is. It seems like it must be a really "solved" problem by now given how many well documented command line tools there are, but I haven't found a well-maintained argparse-to-markdown library. I played around this morning with what I can do with a simple custom help formatter like you mentioned. It seems like the most promising route (thanks for the suggestion), and I'll see if I get anywhere with it this afternoon.

czue:
> Did you kind of mean use that instead of e.g. https://github.com/dimagi/commcare-cloud/pull/1663/files#diff-630448bcf64108b4e08a87d74ec52e1cR36 (`arguments = ... + NON_POSITIONAL_ARGUMENTS`)?

Yeah was thinking for that type of stuff where we want to provide the same defaults to a bunch of different commands. 

Was also thinking that this must be a solved problem based on how common it seems like it should be but guess not...

mkangia:
Nice, thanks @dannyroberts.

PR #1667, opened 2018-05-01
remove unnecessary ANSIBLE_ROLES_PATH env var

dannyroberts:
clean up from https://github.com/dimagi/commcare-cloud/pull/1626#discussion_r184796942

My best guess is that @pr33thi and I were making changes around the same time, and there was an interim period where she added this to get things to work, but I think it shouldn't be necessary on the latest version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,6 @@ We also recommend that you put the following in your `~/.profile` which gives yo
 from anywhere:
 ```
 export PATH=$PATH:~/.commcare-cloud/bin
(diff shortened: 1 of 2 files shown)


dannyroberts:
Hmm. I should have probably just made the version `1.0`, because even though it's conceptually "adding a new feature", it's in practice (as in this repo and the commcare-hq) changing bad behavior that was being implicitly relied on (in both cases without the authors' knowledge). Too late now though. I'll just merge master in here once it's updated.

PR #1677, opened 2018-05-03
acceptable delay for standby db

sravfeyn:
companion for https://github.com/dimagi/commcare-hq/pull/20315


environments/icds-new/postgresql.yml
```diff
@@ -82,8 +82,10 @@ dbs:
       host: pgucrstandby
       create: False
       django_migrate: False
+      hq_acceptable_standby_delay: 3
```
snopoke: can we make this 30s

emord: Any reason you think we couldn't go higher? At the moment, the UCRs themselves would be > 30s behind in processing anyways. It seems like in that case we could even go as far as 30 minutes? at least until the cluster is more stable

PR #1679, opened 2018-05-03
use setup.py tests_require instead of test_requirements.txt

dannyroberts:


dannyroberts:
@jmtroth0 Didn't realize we weren't properly using setup.py `test_requires` when I recommended `pip install -e .[test]` earlier

PR #1680, opened 2018-05-04
optimize saved exports queues

snopoke:
Not sure why these weren't set


emord:
@mkangia that flag tells the supervisor config to use that optimized bash runner [here](https://github.com/dimagi/commcare-cloud/blob/9332bb83529a6b8c5155c964827238084dd95038/src/commcare_cloud/fab/services/templates/supervisor_celery_worker.conf#L3). otherwise it uses the bash runner from a line up in your link.

@snopoke what sort of verification do we do to decide if `-O` is safe?

mkangia:
Thanks @emord. i was expecting so find a file somewhere with name `celery_bash_runner_optimized` in hq code base but found just `celery_bash_runner.sh`

snopoke:
@emord we only do this for the `saved_exports_queue` which only runs data exports. We've only done manual verification that this is safe.

PR #1681, opened 2018-05-04
adding couch sql logfile to better place

jmtroth0:
I noticed the logging that i added in [1f3ef0dfc8](https://github.com/dimagi/commcare-hq/commit/1f3ef0dfc8833ada35cdb06263281dae6c6060b1) is not showing up in the file also created there. Looking at logfiles that are working, this pattern seemed like the successful one. 

Lemme know if there's something else that might be going on, since I don't know much about our logging scaffolding

@proteusvacuum


proteusvacuum:
worth a try!
however, I don't think this is a great place to keep logs since they are getting stored in the releases directory, which might get blown away at any time. you'll also need to know which release the log item you are looking for is located. I guess this is fine if you are running a management command from a private release, but something to look out for.

jmtroth0:
>stored in the releases directory

`LOG_HOME` is `www/<env_name/log` in this case, as far as I could tell. The commit linked above definitely put it in the releases folder, but I was hoping to avoid that this time

PR #1691, opened 2018-05-07
Keep only 1 week old backup

NitigyaS:
Reduce weekly backup on servers to 1.

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -23,6 +23,7 @@ backup_postgres: plain
 backup_es: True
 postgres_s3: True
 postgresql_backup_days: 1
+postgresql_backup_weeks: 1
 
 aws_region: 'us-east-1'
 


NitigyaS:
Giving it a second thought. I think keeping weekly backups on s3 is overkill.

dannyroberts:
Can you say more? Will we now store two historical backups for 1 day ago and 1 week ago respectively? I don't have a clear enough picture of how we'd eventually use these in an emergency to know whether I agree or disagree, so it'd be nice to just hear out the reasoning (or link to a conversation if you have it), but then I'll probably defer to you

NitigyaS:
Since we are already storing all our daily backups on s3. It contains the daily backups of 30 days and the older one in the glacier.  Therefore storing a separate weekly backup is redundant. we can also discuss the same offline

calellowitz:
@NitigyaS not sure what the state of this is, but historically the "weekly" backup was just the normal daily backup taken on sunday, which we kept for longer. Does your last comment mean we no longer send the weekly backup to aws? If so that might me we no longer maintain full backup coverage.

calellowitz:
> since it is a redundant copy of Sunday's daily backup

We are on the same page up until this point. My understanding is not that the weekly is a redundant copy of the sunday daily backup but that it **is** the sunday daily backup. We only do one backup that day, we just mark it differently so we know to keep it for longer. I think the daily backup script does not run on sundays. If that is true, we would need to upload to aws, just like all the daily backups. If your understanding is correct, then I agree with the rest of what you say.

calellowitz:
And on second look, this PR does not seem to address AWS uploads anyway, so that conversation may not be relevant to this work.

NitigyaS:
That makes perfect sense. Thanks, @calellowitz  for pointing out that on Sunday we take only one backup.
This PR is for only limiting the number of weekly backups to keep on the VM to avoid running out of space. I was asking if I should limit the backup upload but I guess it is not required at all. 
On the other hand, can I get this PR approved?

PR #1693, opened 2018-05-07
adding celery main queue to pna

calellowitz:
@snopoke @kaapstorm @NitigyaS

### DIFF
--- a/environments/pna/app-processes.yml
+++ b/environments/pna/app-processes.yml
@@ -21,6 +21,8 @@ celery_processes:
       concurrency: 1
     async_restore_queue:
       concurrency: 1
+    celery:
+      concurrency: 1
     flower: {}
   None:
     # todo: create these queues?
--- a/environments/pna/app-processes.yml
+++ b/environments/pna/app-processes.yml
@@ -21,8 +21,9 @@ celery_processes:
       concurrency: 1
     async_restore_queue:
       concurrency: 1
-    celery:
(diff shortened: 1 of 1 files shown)


calellowitz:
I suspect we want to add all the missing queues but am holding off for now

environments/pna/app-processes.yml
```diff
@@ -21,6 +21,8 @@ celery_processes:
       concurrency: 1
     async_restore_queue:
       concurrency: 1
+    celery:
+      concurrency: 1
```
esoergel: Is this not what line 9 accomplishes?
```
repeat_record_queue,celery,export_download_queue:
  concurrency: 1
```

esoergel:
Not sure whether it's related, but it looks like malt tables aren't getting built on PNA either (that normally happens via these tasks:
https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/data_analytics/tasks.py

environments/pna/app-processes.yml
calellowitz: yeah, @dannyroberts pointed that out to me. Which leaves confused how this solved the problem, or how that problem existed before, but confident that this PR should not be needed.

calellowitz:
@esoergel guessing thats the lack of a periodic queue. i can amend this pr to include one (and eliminated the redundant main queue)

PR #1702, opened 2018-05-08
Changing the way auth is done

pr33thi:
This addresses the discussion in this thread: https://github.com/dimagi/commcare-cloud/pull/1626#discussion_r184790845

1. The user no longer has to specify whether to use the pem file as a command line argument
2. If a pem file isn't available, the user is prompted to enter a root password
3. Created a `factory-ping` command specifically to ping machines to see if they are provisioned
4. `bootstrap-users` now has auth (it didn't before, which is a bug)

----
Notes to self
Questions:
- NONPOSITIONAL_ARGUMENTS aren't included in index.md?
- Is that the place to put the equivalent of a bool that defaults to false?


dannyroberts:
I just merged master, which now auto-generates the command docs. I saw from the new diffs in the auto-generated docs that `factory-ping` accepts the `--use-factory-auth`; but it also _always_ sets `args.use_factoy_auth = True`.

It's small but worth fixing I think. We either want to change the command to be called just `ping` and accept (and respect) a `--use-factory-auth` argument, _or_ remove the `--use-factory-auth` argument from `factory-ping`. I'm actually leaning towards a generic `ping` command, since I've found myself wanting to just ping all machines in steady state as well to check connectivity.

src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -35,13 +35,16 @@ def _build_env(self, args):
...
     strict_host_key_checking = environment.public_vars.get('commcare_cloud_strict_host_key_checking', True)
 
     common_ssh_args = []
-    if use_pem:
-        pem = environment.public_vars.get('commcare_cloud_pem', None)
-        if pem:
+
+    if use_factory_auth:
+        if not environment.public_vars.get('commcare_cloud_pem'):
+            common_ssh_args += ('--ask-pass',)
```
dannyroberts: I think this won't work because `--ask-pass` isn't an ssh arg, but rather an ansible arg. All this stuff gets shell-quoted and passed together to the ansible arg `--ssh-common-args`, so I think ansible won't see that you used `--ask-pass`.

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -233,6 +234,7 @@ class BootstrapUsers(_AnsiblePlaybookAlias):
     def run(self, args, unknown_args):
         environment = get_environment(args.env_name)
         args.playbook = 'deploy_stack.yml'
+        args.factory_auth = True
         public_vars = environment.public_vars
         root_user = public_vars.get('commcare_cloud_root_user', 'root')
         unknown_args += ('--tags=users', '-u', root_user)
```
dannyroberts: This stuff (adding `'-u', public_vars.get('commcare_cloud_root_user', 'root')` to the ansible args) should happen on account of `--use-factory-auth`, so this should get added in the same place that the rest of that functionality happens.

src/commcare_cloud/commands/ansible/helpers.py
dannyroberts: Given that `use_factory_auth` spans spans ssh and non-ssh arguments, I might break it out to another place. You  may want to also test whether you can pass in `--ssh-common-args` more than once, since that'll affect the implementation.

pr33thi:
> I just merged master, which now auto-generates the command docs. I saw from the new diffs in the auto-generated docs that factory-ping accepts the --use-factory-auth; but it also always sets args.use_factoy_auth = True.

How can I view those docs?

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -120,10 +121,10 @@ def ansible_playbook(environment, playbook, *cmd_args):
             return p.returncode
 
         def run_check():
-            return ansible_playbook(environment, args.playbook, '--check', *unknown_args)
+            return ansible_playbook(environment, args.playbook, args.factory_auth, '--check', *unknown_args)
```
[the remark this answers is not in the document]
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1702/commits/6cb8306c55d7df61d06fca0a9a818163e3b36f8e

src/commcare_cloud/commands/ansible/helpers.py
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1702/commits/ce8f40d15e52574a8bcad6beaf0f311cd9299b67

There's some redundant code between the two places where this function gets called that I'd like to consolidate once everything is working

src/commcare_cloud/commands/ansible/ansible_playbook.py
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1702/commits/c875390478ba7efb6492f5af4023ae5a50babc22

https://github.com/dimagi/commcare-cloud/pull/1702/commits/898af473bfe427544070dea1c554873d8842bce5

https://github.com/dimagi/commcare-cloud/pull/1702/commits/27542ff516ac6c7c35a38eca57d88a62818e4d8f

(This part is still buggy, I need to keep debugging)

dannyroberts:
> How can I view those docs?


You can rebuild the docs by running `make` in the root directory of the repo. The file it generates you then have to commit, and it's under `docs/commcare-cloud/commands/index.md`. You can look at it in any markdown viewer, or to get the full effect and closest to the way it looks at http://dimagi.github.io/commcare-cloud/commcare-cloud/commands/, you can follow the instructions at `docs/README.md`

dannyroberts:
Yeah I know I was the one who originally suggested `factory-ping`, but seeing the way it was coming together made me adjust a bit to thinking `ping` would be helpful command in general. `cchq ... ping --use-factory-auth` seems like an okay thing to have to type.

src/commcare_cloud/commands/ansible/run_module.py
```diff
@@ -141,7 +137,7 @@ def run_ansible_module(environment, ansible_context, inventory_group, module, mo
     ask_vault_pass = include_vars and public_vars.get('commcare_cloud_use_vault', True)
     if ask_vault_pass:
         cmd_parts += ('--vault-password-file=/bin/cat',)
-    cmd_parts += get_common_ssh_args(environment, use_pem)
+    cmd_parts = get_common_ssh_args(cmd_parts, environment, use_factory_auth=factory_auth)
```
dannyroberts: I'm not sure I understand why `cmd_parts` needs to get passed in; it looks to me that this change just pushes the `+=`ing to inside that function, which is (for me) harder to reason about. In addition, I _think_ I'm seeing a mistake there that makes `cmd_parts` get totally overridden when you're using factory auth (?)

src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -35,27 +35,38 @@ def _build_env(self, args):
...
     strict_host_key_checking = environment.public_vars.get('commcare_cloud_strict_host_key_checking', True)
 
     common_ssh_args = []
-    if use_pem:
-        pem = environment.public_vars.get('commcare_cloud_pem', None)
-        if pem:
-            common_ssh_args.extend(['-i', pem])
+
+    if use_factory_auth:
+        cmd_parts, common_ssh_args = add_factory_auth_cmd(environment, common_ssh_args, cmd_parts)
```
dannyroberts: Won't this clobber the `cmd_parts` that was passed in? My reading is that the passed in variable contains lots of crucial stuff about the command to be run that can't be thrown away. I think this kind of error (if I'm following what's going on here correctly) is a good demonstration of why I think it's simpler to reason about not to pass in `cmd_parts` and keep the function signature the way it was previously.

src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -35,27 +35,38 @@ def _build_env(self, args):
...
     if os.path.exists(known_hosts_filepath):
         common_ssh_args.append('-o=UserKnownHostsFile={}'.format(known_hosts_filepath))
 
-    cmd_parts = tuple()
     if common_ssh_args:
         cmd_parts += ('--ssh-common-args', ' '.join(shlex_quote(arg) for arg in common_ssh_args))
     return cmd_parts
 
 
+def add_factory_auth_cmd(environment, common_ssh_args, cmd_parts):
```
dannyroberts: Nice, nice to have the central logic broken out like this.

src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -35,27 +35,38 @@ def _build_env(self, args):
...
+    root_user = environment.public_vars.get('commcare_cloud_root_user', 'root')
+    if '-u' not in cmd_parts:
+        cmd_parts += ('-u', root_user)
+    if not environment.public_vars.get('commcare_cloud_pem'):
+        cmd_parts += ('--ask-pass',)
+    else:
+        pem = environment.public_vars.get('commcare_cloud_pem', None)
```
dannyroberts: This expression and the one in the line above one that looks like it are equivalent (i.e. `None` is the default value of the second argument), I think it'd be clearer to read if you pulled this line out above the if/else ("`pem = environment.public_vars.get('commcare_cloud_pem')`") and referenced it in the condition as well.

src/commcare_cloud/commands/ansible/run_module.py
```diff
@@ -197,3 +193,30 @@ def run(self, args, unknown_args):
...
+    def run(self, args, unknown_args):
+        ping_shell_command = 'echo {{ inventory_hostname }}'
+
+        args.module = 'shell'
+        if args.silence_warnings:
+            args.module_args = 'warn=false ' + ping_shell_command
+        else:
+            args.module_args = ping_shell_command
+        args.skip_check = True
+        args.quiet = True
```
dannyroberts: A bunch of this smacks of being copy-pasted from `RunShellCommand`. Conceptually this command (`ping`) is trying to basically run `run-shell-command 'echo {{ inventory_hostname }}'`, right? So I'd advocate for _calling_ `RunShellCommand.run` rather than _copying_ its contents.

src/commcare_cloud/commands/ansible/run_module.py
```diff
@@ -197,3 +193,30 @@ def run(self, args, unknown_args):
...
+        del ping_shell_command
```
dannyroberts: this line doesn't do anything, to my knowledge, but looks like it was maybe done in analogy to e.g. `del args.shell_command` above? The purpose of _that_ line is to remove an unnecessary argument from args, since it's (okay, somewhat sloppily, I admit) being repurposed right after as the `args` for `RunAnsibleModule`.

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -121,10 +122,10 @@ def ansible_playbook(environment, playbook, *cmd_args):
             return p.returncode
 
         def run_check():
-            return ansible_playbook(environment, args.playbook, '--check', *unknown_args)
+            return ansible_playbook(environment, args.playbook, args.use_factory_auth, '--check', *unknown_args)
```
dannyroberts: Hmm, it's kinda ugly just because of the current state of this function, but I'd actually suggest simply depending on the outer scope for the value of this variable, like we do with e.g. `public_vars` and a number of other outside variables used within `ansible_playbook`. Having a bunch of positional arguments here is a bit awkward I think.

commcare-cloud-bootstrap/bootstrap.sh
```diff
@@ -6,7 +6,7 @@ SPEC=$3
 
 commcare-cloud-bootstrap provision $SPEC --env $ENV
 while
-    commcare-cloud $ENV run-shell-command all 'echo {{ inventory_hostname }}' -u ubuntu --use-pem
+    commcare-cloud $ENV run-shell-command all 'echo {{ inventory_hostname }}' -u ubuntu --use-factory-auth
```
dannyroberts: I think this would be a great use case for the shiny new `ping` function :)

Also, `-u ubuntu` should now be unnecessary, since that's now included in `--use-factory-auth` (right?)

dannyroberts:
Re: build, looks like you need to hide pvtest from view (i.e. move it to another dir or something) and run `make` again

src/commcare_cloud/commands/ansible/run_module.py
pr33thi: Fixed it based on our convo here:
https://github.com/dimagi/commcare-cloud/pull/1702/commits/f70a5f2d756f90ed5d88422d9e4af563fdb2fe46

src/commcare_cloud/commands/ansible/helpers.py
pr33thi: Doesn't error, but updated this to make more readable here:
https://github.com/dimagi/commcare-cloud/pull/1702/commits/f70a5f2d756f90ed5d88422d9e4af563fdb2fe46

src/commcare_cloud/commands/ansible/helpers.py
pr33thi: Here: https://github.com/dimagi/commcare-cloud/pull/1702/commits/a33437069826eefc41e9073c51aec4c9cde767d0

src/commcare_cloud/commands/ansible/run_module.py
pr33thi: I'm not confident in my change in line 207 (because the user could specify something besides the default, which isn't really the point of this), do you have advice on that?
https://github.com/dimagi/commcare-cloud/pull/1702/commits/e41d302be1014e4dacd3f3ad27eaa9dfb12f3aef

src/commcare_cloud/commands/ansible/run_module.py
pr33thi: obe

src/commcare_cloud/commands/ansible/ansible_playbook.py
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1702/commits/61df7271376d9381b3d5065ead051b6a6b1cd478

commcare-cloud-bootstrap/bootstrap.sh
pr33thi: Ended up changing it to default to root if `--use-factory-auth` is specified, here:
https://github.com/dimagi/commcare-cloud/pull/1702/commits/6ca956609be7f17cff3c83fdf32ad333e3895c3c

PR #1706, opened 2018-05-09
move pg vars into groups

snopoke:
just moving vars around to reduce duplication


environments/icds-new/inventory.ini
```diff
@@ -124,9 +96,15 @@ partitions=["/dev/sdb1","/dev/sdc1"]
...
+
+[main_dbs:children]
+pgmain
+pgmainstandby1
+
+[main_dbs:vars]
 pgbouncer_max_connections=3500
 pgbouncer_default_pool=775
-postgresql_max_connections=800
+postgresql_max_connections=400
```
snopoke: `pgmain1` had 400 and `pgmainstandby1` had 800. Based on their usage 400 is more than enough

mkangia: just for better understanding,
assuming that we redirect more reads to pgmainstandby1 and hence it is receiving all/most of the reads. If that is, should it be supporting more connections than pgmain1 considering we would be having more reads than writes?
so if 400 is enough for `pgmainstandby1`, 400 is actually a nice buffered number for `pgmain1`?

snopoke: We're currently sending equal number of reads to `pgmainstandby1` for `auth`, `locations`, `accounting` apps only. Currently `pgmain1` is still getting about 50% more traffic than `pgmainstandby1`. I don't think we ever want to send all reads to the standby.

mkangia: Cool, thanks.

PR #1710, opened 2018-05-09
assert exactly one redis host

dannyroberts:
Found this commit from an abandoned branch of mine that seems like a harmless check. Easy to remove later if we want to support external redis.


dannyroberts:
Okay, interesting. I would be happy not to add this check if it gets in the way of normal migration workflows. It seems like there's some discussion on the other thread, but feel free to close this PR if you decide the constraint is too... constraining.

PR #1711, opened 2018-05-09
fix and extend deploy_control.yml

dannyroberts:
make it install commcare-cloud environment for each dev user

### DIFF
--- a/src/commcare_cloud/ansible/deploy_control.yml
+++ b/src/commcare_cloud/ansible/deploy_control.yml
@@ -1,6 +1,5 @@
 - name: Bootstrap Control
-  hosts: localhost
-  gather_facts: no
+  hosts: control
   become: true
   gather_facts: true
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/deploy_control.yml
```diff
@@ -1,6 +1,5 @@
 - name: Bootstrap Control
-  hosts: localhost
-  gather_facts: no
+  hosts: control
```
snopoke: won't this make it ssh into the control machine? or is that the point so that you can run it from elsewhere?

dannyroberts: Yeah that's the point. I wanted to be able to use this to bootstrap a control machine remotely. It should also still work from the control machine itself, it'll just be sshing into itself.

PR #1712, opened 2018-05-10
Added db1

NitigyaS:
Create db1 with lower specs then db0.
The plan is to set as standby of the db0 and the promote to master.

### DIFF
--- a/environments/softlayer/inventory.ini
+++ b/environments/softlayer/inventory.ini
@@ -74,7 +74,6 @@ proxy0
 django0
 
 [postgresql:children]
-db0
 db1
 
 [rabbitmq:children]
@@ -99,7 +98,6 @@ touch0
 formplayer0
 
 [redis:children]
-db0
 db1
 
 [elasticsearch:children]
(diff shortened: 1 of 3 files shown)


environments/softlayer/inventory.ini
```diff
@@ -96,13 +100,15 @@ formplayer0
 
 [redis:children]
 db0
+db1
```
snopoke: was just commenting here https://github.com/dimagi/commcare-cloud/pull/1710#pullrequestreview-119000245

It might be better to remove db0 completely and just deploy from this branch until ready to flip?

tests/postgresql_config/2018-04-04-softlayer-snapshot/inventory.ini
```diff
@@ -1,5 +1,5 @@
-[db0]
-10.162.36.205
+[db1]
+10.162.36.238
```
dannyroberts: I don't think you need to change this, since this is just a test based on a snapshot of how things were at the time it was created.

PR #1713, opened 2018-05-10
Added Elasticsearch Vars

NitigyaS:
Want to do it with Postgres downsizing.

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
@@ -21,7 +21,7 @@ transport.tcp.port: {{ elasticsearch_tcp_port }}
 http.port: {{ elasticsearch_http_port }}
 
 http.cors.enabled: true
-
+# read data from older 1.7.3 directory if ES is upgraded to 1.7.6 . Untill rename is done on softlayer 
 path.data: {{ elasticsearch_old_data_dir }}/data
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -22,8 +22,8 @@ http.port: {{ elasticsearch_http_port }}
 
 http.cors.enabled: true
 
-path.data: {{ elasticsearch_data_dir }}/data
-path.logs: {{ elasticsearch_data_dir }}/logs
+path.data: {{ elasticsearch_old_data_dir }}/data
```
NitigyaS: It's a quick fix. After Upgrading Softlayer. the data remained in the 1.7.3 directory. this PR enables 1.7.6 to uses the older version directory for the data. I would like to fix this by moving data from 1.7.3 to 1.7.6.

snopoke: Can we not just rename the folder?

NitigyaS: This is what I want to do but I will have to stop the elasticsearch for that. So it is just to avoid changes if someone deploys it meanwhile.

snopoke: OK, can you just add a comment with a short explanation. Thanks

PR #1714, opened 2018-05-10
using ip for server whitelist

calellowitz:
@snopoke 
@dannyroberts is it intentional that inventory names do not work and only IPs do? Fine if thats the case, was just surprised this failed


dannyroberts:
Yeah, it's intentional because the point of the `server_whitelist` is to make it really hard for celery periodic to be running on a machine without someone explicitly deciding for it to; in particular, we want copying and pasting from another environment never to work without calling that line to someone's attention, even if the servers are in inventory groups with the same name in the two different environments.

PR #1719, opened 2018-05-11
use hostnames in celery worker names

snopoke:
For environments like ICDS where host_string is an IP this will translate the IP address into the hostname as defined in the inventory file.

This makes it easier to know which host a process is running on when looking at the celery metric data:

`celery@10.247.164.40_reminder_case_update_queue_1`
becomes
`celery@celery0.internal-icds-new.commcarehq.org_reminder_case_update_queue_1`

### DIFF
--- a/environments/development/inventory.ini
+++ b/environments/development/inventory.ini
@@ -1,3 +1,21 @@
+[app1]
+192.168.33.15
+
+[app1:vars]
+hostname='app1'
+
+[db1]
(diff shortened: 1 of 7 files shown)


snopoke:
It could be possible to do that but we'd have to make some assumptions about the way the inventory file was constructed which would be hard to enforce. This also makes it explicit what the hostname is meant to be.

PR #1721, opened 2018-05-11
Add decision record for custom service elasticsearch implementation

dannyroberts:
I'm trying out a pattern called Architecture Decision Records. The first one explains the pattern (it's a stock template made by the `adr` tool that it itself mentions).

The second one is the payload of this PR, and describes the (proposed) decision to push "smarter" behavior into the `cchq <env> service elasticsearch <action>` commands. (It is marked with status "Approved" because merging this PR constitutes approving the decision.)

### DIFF
--- a/decisions/0002-back-service-elasticsearch-with-custom-implementation.md
+++ b/decisions/0002-back-service-elasticsearch-with-custom-implementation.md
@@ -1,0 +1,65 @@
+# 2. Back service elasticsearch with custom implementation
+
+Date: 2018-05-11
+
+## Status
+
+Accepted
+
+## Context
+
+The `service` command is has become a useful way for us to start, stop, restart,
+and check the status of the many service processes behind a commcare cluster.
+For the most part, these commands map very predictably to a start/stop/restart/status
+command on either [`service`](https://linux.die.net/man/8/service) or `supervisorctl`.
+
+Recently and in the past, we have found that `elasticsearch` does not stop or restart
+reliably. In particular, after running `service elasticsearch stop`,
+using `ps aux | grep elasticsearc[h]` may still show a running Elasticsearch
+process, even if `service` is reporting a stopped status. This then requires
+killing the rogue process in order for a `start` to bring up a genuinely new process.
+
+Before the addition of the `service` command, we already had the `restart-elasticsearch`
+command, which is conceptually (though currently not functionally) equivalent to
+`... service elasticsearch restart`. It does not address the above issue of rogue
+processes, but it does address some other issues, including a reminder to stop all pillows,
+as well as theoretically providing rolling (and thus no-downtime) restart,
+which we currently do not benefit from because we do not store data in duplicate.
+
+## Decision
+
+We will be changing `... service elasticsearch ...` from using simply using the default
+behavior to performing a more targeted set of actions that are guaranteed to achieve the
+stop/start/restart/status action, and we will also be deprecating the `restart-elasticsearch`
+entry point. In particular, `... service elasticsearch stop` will
+
+- stop pillows
+- guarantee the process has stopped and `kill -9` it after a period of time if it has not
+
+and `start` will
+
+- start pillows
+
+and `restart` will be functionally a `stop` followed by a `start`.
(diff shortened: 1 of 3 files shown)


decisions/0002-back-service-elasticsearch-with-custom-implementation.md
```diff
@@ -0,0 +1,65 @@
...
+Before the addition of the `service` command, we already had the `restart-elasticsearch`
+command, which is conceptually (though currently not functionally) equivalent to
+`... service elasticsearch restart`. It does not address the above issue of rogue
+processes, but it does address some other issues, including a reminder to stop all pillows,
+as well as theoretically providing rolling (and thus no-downtime) restart,
+which we currently do not benefit from because we do not store data in duplicate.
+
+## Decision
+
+We will be changing `... service elasticsearch ...` from using simply using the default
```
emord: `using simply using` should probably just be `using`

decisions/0002-back-service-elasticsearch-with-custom-implementation.md
```diff
@@ -0,0 +1,65 @@
...
+and `restart` will be functionally a `stop` followed by a `start`.
+
+Pros:
+- Someone with little context will be able to run a `... service elasticsearch ...`
+  and have it have the intended overall consequence
+- The tool will incorporate what is currently heard knowledge
+
+Cons:
+- A "power" user may think of the command as executing a very specific other command,
+  and may be taken aback by the tool trying to outsmart them.
```
emord: It's always frustrating to me when tools try to do this. I think that `service` is the exact command that runs `/etc/init.d/<service_name>` right? What do you think of having a different command like `cchq <env> safe_restart <service>`  that does these sorts of things? It makes it clear that it is a superset of whatever service management scheme we use now and in the future.

snopoke: I think this is OK given what @dannyroberts said in the next paragraph about the tool making it clear what's happening and also allowing the user to 'downgrade' to the simple version.

I also don't think having a separate command is a great idea since it would mean there are 2 commands for doing the same thing which requires a user to know which command to run in their context. It also means a user may not know about both and end up using the wrong one out of ignorance.

calellowitz:
Did this ever get implemented? I had to restart elasticsearch recently and it seems both that not only was this change never made but `service elasticsearch restart` does not work. I know we have another command to restart es `restart-elasticsearch` which prompts you to stop pillows and I was able to use that one to restart the system, but it feels like this project has been abandoned. If so, I still think we need to decide which of the two ways to restart is correct and make sure it works

dannyroberts:
This is on @pr33thi's backlog to work on. Feel free to offline with me and her if you want to weigh in, but your point about this still being an issue and dragging on is noted.

pr33thi:
Yep, I've been working on this on the side, but I can prioritize it

calellowitz:
No need to prioritize this on my behalf. I am more curious what the current recommended workflow is. Is it still `restart-elasticsearch`?

dannyroberts:
I believe it's more or less

```bash
cchq <env> service elasticsearch stop
cchq <env> run-shell-command elasticsearch 'ps aux | grep elasticsearc[h]'
# if there's any output...
cchq <env> run-shell-command elasticsearch 'ps aux | grep elasticsearc[h] | cut -c10-15 | xargs kill'
# check again until they're all dead
cchq <env> run-shell-command elasticsearch 'ps aux | grep elasticsearc[h]'
# and then finally once they are
cchq <env> service elasticsearch start
```

dannyroberts:
My impression reading the tasks is that it doesn't actually check to see
it's really stopped:
https://github.com/dimagi/commcare-cloud/blob/e0e133bfa4263d168659b494975525d6cda545e9/src/commcare_cloud/ansible/roles/elasticsearch/tasks/rolling_restart.yml#L10-L19,
it just waits 10 seconds and assumes that it's then stopped. Part of what
@pr33thi is going to do is insert logic here that checks for the processes
and kills them if they exist.

PR #1723, opened 2018-05-14
downtime tweaks

snopoke:


src/commcare_cloud/commands/ansible/downtime.py
```diff
@@ -81,19 +82,29 @@ def wait_for_all_processes_to_stop(environment, ansible_context):
...
-            options + ['a', 'w', 'k']
+            options + ['a', 'w', 'c', 'k']
         )
         if response in ('a', 'abort'):
-            downtime = get_downtime_record(environment)
-            cancel_downtime_record(environment, downtime)
-            return
+            if ask('This will start all CommCare processes again. Do you want to proceed?'):
+                downtime = get_downtime_record(environment)
+                supervisor_services(environment, ansible_context, 'start')
```
mkangia: @snopoke should this be a restart instead of start? thinking of this as a state where some processes are still running and some have stopped. Using start would mean that some of them would fail because they are already running but restart would safely just restart them.

src/commcare_cloud/commands/ansible/downtime.py
```diff
@@ -81,19 +82,29 @@ def wait_for_all_processes_to_stop(environment, ansible_context):
...
+                cancel_downtime_record(environment, downtime)
+                return
         elif response in ('w', 'wait'):
             time.sleep(30)
+        elif response in ('c', 'continue'):
+            if ask('Are you sure you want to continue with downtime even though there '
+                   'are still some processes running?'):
+                return
```
mkangia: Can you share why this option has been added? because you would just terminate the command passed?
I think fine to have an option to pass to just continue but don't think it needs to be confirmed with the user.

src/commcare_cloud/commands/ansible/downtime.py
snopoke: By this point the processes are already 'stopped' according to supervisor. Any remaining processes are either zombies or not controlled by supervisor so I don't expect any to be in the 'RUNNING' state. Also when you do `start all` supervisor will ignore any processes that are already running.

src/commcare_cloud/commands/ansible/downtime.py
snopoke: This came up during recent downtime where someone else was running a shell and wanted to keep the shell running during the downtime (as it wasn't going to be impacted by the maintenance). Pressing `Ctrl-C` has the same effect but that is much less user friendly and results in a traceback etc. Also it leaves the operator unsure as to what state the system was left in.

Having the confirmation is just to make sure that it wasn't a mistype (since the shortcut is just a single character) and also to make sure the operator knows what state the system will be left in.

mkangia: Cool, thanks for sharing the context. Makes sense.

PR #1726, opened 2018-05-14
add production hqriak35

dannyroberts:


environments/production/inventory.ini
```diff
@@ -119,6 +119,7 @@ hqriak31.internal-va.commcarehq.org
 hqriak32.internal-va.commcarehq.org
 hqriak33.internal-va.commcarehq.org
 hqriak34.internal-va.commcarehq.org
+hqriak35.internal-va.commcarehq.org
```
emord: Whats the tradeoff between adding new machines and just adding more storage?

millerdev: Adding more storage (using LVM) would be an option, although we have not used or documented that process (yet). I'm not familiar with the LVM recovery process when a physical disk fails when there are multiple disks in the logical volume, but my guess is it requires a rebuild of the entire logical volume. Theoretically the odds of volume failure increase as you add more disks. Does it increase the complexity of recovery scenarios or otherwise affect probability of data loss if we add more physical volumes to each logical volume?

One disadvantage of having very large nodes is it takes longer to replace a node in case of failure.

dannyroberts:
Deployed. The test failure is also happening on master. Will fix

PR #1736, opened 2018-05-17
Reduce concurrency of few workers.

NitigyaS:
I have halved the concurrency for SoftLayer.  But looking for some suggestions on what would be a good number.

### DIFF
--- a/environments/softlayer/app-processes.yml
+++ b/environments/softlayer/app-processes.yml
@@ -4,7 +4,7 @@ celery_processes:
   'celery0':
     celery_periodic:
       server_whitelist: '10.162.36.233'
-      concurrency: 4
+      concurrency: 2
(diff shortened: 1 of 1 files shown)


snopoke:
Is this because the CPU usage is too high or because we don't need that many workers?

NitigyaS:
CPU usage is normal. RAM usage is much higher.

NitigyaS:
@snopoke can you review this.

NitigyaS:
@gcapalbo Can you suggest which queues I can consolidate together. Do not want to club 2 high task processing queues together.

gcapalbo:
@NitigyaS for now you could try putting the `email_queue` with the `celery,export_download_queue`, and putting the `ucr_queue` and `ucr_indicator_queue` together

PR #1740, opened 2018-05-18
Enterprise branch

emord:
Defaults to deploying `origin/enterprise` to our non-dimagi hosts.

Keeping discussion on https://github.com/dimagi/commcare-hq/pull/20354

commenters on that PR: @snopoke @czue @millerdev @dannyroberts

### DIFF
--- a/src/commcare_cloud/fab/utils.py
+++ b/src/commcare_cloud/fab/utils.py
@@ -58,27 +58,42 @@ class DeployMetadata(object):
         self._last_tag = None
         self._code_branch = code_branch
         self._environment = environment
+        self._repo = None
+
+    @property
+    def repo(self):
+        if self._repo is not None:
+            return self._repo
+
+        github = _get_github()
+        self._repo = github.get_organization('dimagi').get_repo('commcare-hq')
+        return self._repo
+
+    @property
+    def last_tag(self):
+        if self._last_tag is not None:
+            return self._last_tag
+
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/fab/utils.py
```diff
@@ -147,6 +159,10 @@ def deploy_ref(self):
 
         return self._deploy_ref
 
+    @property
+    def current_ref_is_different_than_last(self):
+        return self.deploy_ref != self.last_tag.commit.sha
```
esoergel: @emord 
For infrequently deployed environments, `last_tag` is often `None`.  This is because we only check the last 100 tags:
https://github.com/dimagi/commcare-cloud/blob/3441972bdb76c0385a37faa67b1c1a4d3f8e6f95/src/commcare_cloud/fab/utils.py#L78

@gbova ran into this issue deploying PNA.  Fixes would probably be:
* Bump the number of tags to check up from 100, which I imagine was chosen arbitrarily, but is evidently too small.
* Modify this code to account for the fact that `last_tag` may sometimes be `None` (that's what I did in https://github.com/dimagi/commcare-cloud/pull/1619).  This should likely be done either way, since at least during the first deploy to an environment, there won't _be_ a previous deploy.

Also, we should probably announce this switch to `origin/enterprise` for these environments to the broader tech team (apologies if it's already been announced and I missed it).  I could imagine that causing some really confusing behavior, like if I merge a PR for PNA on a Monday before the deploy.

PR #1751, opened 2018-05-21
Push more queries to ucr standby

emord:
@dimagi/scale-team standbys should be able to handle more reads than the main DB and the main db has quite a bit higher load pretty consistently

### DIFF
--- a/environments/icds-new/postgresql.yml
+++ b/environments/icds-new/postgresql.yml
@@ -5,8 +5,8 @@ REPORTING_DATABASES:
     WRITE: icds-ucr
     READ:
       - [icds-ucr, 1]
-      - [icds-ucr-standby1, 1]
-      - [icds-ucr-standby2, 1]
+      - [icds-ucr-standby1, 2]
+      - [icds-ucr-standby2, 2]
   icds-test-ucr: icds-ucr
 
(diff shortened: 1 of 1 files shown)


emord:
https://github.com/dimagi/commcare-cloud/pull/1715/ Looks like htis was changed before we figured out where the "idle in transaction" issues were coming from

czue:
> before we figured out where the "idle in transaction" issues were coming from

I must have missed this but we figured it out? Can someone send a pointer to what it was?

emord:
@czue https://github.com/dimagi/commcare-hq/pull/20594

Much more detail: https://docs.google.com/document/d/12ta_Jo4fv_CqMe-4GyA5JFuaNoKKhcjymAc_4d_UnJ8/edit

czue:
thanks @emord. I had seen that but didn't realize we knew it was the solution.

snopoke:
Based on the metrics I don't think we've solved this issue.

emord:
Hmm. yeah. it does look like there are a few left, although they are much less.

When I looked it seems like they were all either exists/deletes or inserts which points to pillows being likely, but the code (in the adapter) looks like  it handles sessions properly (uses a context manager)

PR #1752, opened 2018-05-21
Starting supervisor with processes.

pr33thi:
Adding changes on top of these: https://github.com/dimagi/commcare-cloud/pull/1702 to make sure the tests also pass.

The weird errors we saw in travis before seemed to be transient, and went away after restarting the build a couple times. 
The reason the tests failed after that: I got rid of `failed_when: no`
The fix: make sure that supervisor was installed before running it with a `when: which_supervisord.stdout`


src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
```diff
@@ -59,8 +59,7 @@
 
 - name: stop supervisor to pick up configuration changes
   become: yes
-  command: stop supervisord
-  failed_when: no
+  service: name=supervisord state=stopped
```
dannyroberts: I'm confused why this wouldn't also need a
```
when: which_supervisord.stdout
```
Maybe `state=stopped` doesn't actually require the thing to exist at all?

commcare-cloud-bootstrap/bootstrap.sh
```diff
@@ -14,6 +14,6 @@ done
 commcare-cloud $ENV bootstrap-users --quiet --branch=$BRANCH
 commcare-cloud $ENV deploy-stack --skip-check --quiet -e 'CCHQ_IS_FRESH_INSTALL=1' --branch=$BRANCH
 
-## This next line is a temp fix to allow supervisor to actually get installed.
-commcare-cloud $ENV update-users --quiet --branch=$BRANCH --skip-check
```
dannyroberts: Do we maybe need both these lines? I think this was originally added so that the sudoers file gets updated once supervisor is installed

src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
pr33thi: When I ssh onto a machine before supervisor is installed, `stop supervisord` throws an error. But yeah, maybe having it as a service doesn't require it to exist. Should I put the `when: which_supervisord.stdout` line here for completeness?

commcare-cloud-bootstrap/bootstrap.sh
pr33thi: Yep, added it back here: https://github.com/dimagi/commcare-cloud/pull/1752/commits/2fa3c5c3f7f4838104801fcd2c9e5f35b9bcc462

src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
dannyroberts: Eh, if it didn't cause problems, I'm fine with it!

PR #1755, opened 2018-05-21
More robust deploy last commit finder

emord:
@gbova @esoergel sorry about the earlier error. 

The first commit adds a new method to find the latest commit. It defaults to using `last_tag`'s sha but otherwise it cd's into the current directory to find it there. I'm pretty sure this should never fail because we always assume that there's a current release (even on new servers ansible will create that symlink)

The second commit gives some more firm direction on what to do with the error. It tells you what branch you are deploying and how to rebuild the appropriate yaml files.

Let me know what you think and then I'll send a note to the dev team

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -167,7 +167,7 @@ def _setup_env(env_name):
     _confirm_branch(env.default_branch)
     _confirm_environment_time(env_name)
     execute(env_common)
-    _confirm_deploying_same_code()
+    execute(_confirm_deploying_same_code)
 
 
 def _confirm_branch(default_branch='master'):
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -208,11 +208,29 @@ def _confirm_deploying_same_code():
     if env.deploy_metadata.current_ref_is_different_than_last:
         return
 
+    if env.code_branch == 'master':
+        branch_specific_msg = "Perhaps you meant to merge a PR or specify a --set code_branch=<branch> ?"
+    elif env.code_branch == 'enterprise':
+        branch_specific_msg = (
+            "Have you tried rebuilding the enterprise branch (in HQ directory)? "
+            "./scripts/rebuildstaging --enterprise"
+        )
+    elif env.code_branch == 'autostaging':
+        branch_specific_msg = (
+            "Have you tried rebuilding the autostaging branch (in HQ directory)? "
+            "./scripts/rebuildstaging"
+        )
+    else:
+        branch_specific_msg = (
+            "Did you specify the correct branch using --set cod_branch=<branch> ?"
+        )
+
     message = (
-        "Whoa there bud! You're deploying the same code as was previously deployed.\n"
-        "Perhaps you meant to specify --set code_branch=<branch> or should merge some PRs\n"
+        "Whoa there bud! You're deploying {code_branch} which happens to be "
+        "the same code as was previously deployed to this environment.\n"
+        "{branch_specific_msg}\n"
         "Is this intentional?"
-    )
+    ).format(code_branch=env.code_branch, branch_specific_msg=branch_specific_msg)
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/fab/utils.py
```diff
@@ -69,6 +70,14 @@ def repo(self):
         self._repo = github.get_organization('dimagi').get_repo('commcare-hq')
         return self._repo
 
+    @property
+    def last_commit_sha(self):
+        if self.last_tag:
+            return self.last_tag.commit.sha
+
+        with cd(env.code_current):
+            return sudo('git rev-parse HEAD')
```
esoergel: Oh interesting - I didn't even think of having the actual `current` dir in scope for this kind of thing - I assumed this all ran locally before you started talking to machines.  Does this happen on one particular target machine or something?

src/commcare_cloud/fab/fabfile.py
```diff
@@ -208,11 +208,29 @@ def _confirm_deploying_same_code():
     if env.deploy_metadata.current_ref_is_different_than_last:
         return
 
+    if env.code_branch == 'master':
+        branch_specific_msg = "Perhaps you meant to merge a PR or specify a --set code_branch=<branch> ?"
```
esoergel: Do you know if the tags from failed deploys get cleaned up?  There's something about `setup_release` tags and `deploy` tags, maybe the latter only happens on successful deploy?

src/commcare_cloud/fab/utils.py
emord: Yeah. this only works because I used `execute` https://github.com/dimagi/commcare-cloud/pull/1755/files#diff-4b38b47cd9db17e5e8e82f5424318f83R170 which calls it on a remote server instead of your local machine. I believe it's happening on the one tagged 'deploy' which should also be 'web0'

src/commcare_cloud/fab/fabfile.py
emord: `deploy` only happens on successful deploys which is what last_tag looks for https://github.com/dimagi/commcare-cloud/blob/697b045adeafb192baf1ab68c6ab0dc9c9a65b50/src/commcare_cloud/fab/utils.py#L86

emord:
@dimagi/team-commcare-hq 

I recently made a change to our enterprise deploy process. From now on instead of deploying master we will be deploying a branch called "enterprise". This branch is constructed from https://github.com/dimagi/commcare-hq/blob/master/scripts/enterprise.yaml in a similar process that staging is rebuilt. Original discussion around this change is contained here: https://github.com/dimagi/commcare-hq/pull/20354

The workflow is expected to be as follows:

1. update enterprise.yaml with a recent tag from a production deploy
2. `./scripts/rebuildstaging --enterprise`
   a. if the tag you use is "old" this will warn you and provide a list of the most recent ([currently 4](https://github.com/dimagi/commcare-hq/pull/20643)) production deploy tags
3. `commcare-cloud <env> fab deploy`
  a. There is a new check that warns you if you are deploying the same code to the server, and gives you a hint of what to do based on what code branch you are deploying

I included this process in our [deploy process confluence](https://confluence.dimagi.com/display/commcarehq/Deploy+Process)

Feel free to ping me on any other questions you have or issues you run into with this process

src/commcare_cloud/fab/fabfile.py
```diff
@@ -208,11 +208,29 @@ def _confirm_deploying_same_code():
...
+            "./scripts/rebuildstaging --enterprise"
+        )
+    elif env.code_branch == 'autostaging':
+        branch_specific_msg = (
+            "Have you tried rebuilding the autostaging branch (in HQ directory)? "
+            "./scripts/rebuildstaging"
+        )
+    else:
+        branch_specific_msg = (
+            "Did you specify the correct branch using --set cod_branch=<branch> ?"
```
proteusvacuum: nit: `code_branch`

emord: https://github.com/dimagi/commcare-cloud/pull/1768

PR #1760, opened 2018-05-22
add TRANSIFEX_DETAILS to localsettings

mkangia:
Just sets the `TRANSIFEX_DETAILS` if api token is present. Just need it on softlayer as of now.


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -780,3 +780,18 @@ EULA_CHANGE_EMAIL = '{{ eula_change_email }}'
 CONTACT_EMAIL = '{{ contact_email }}'
 SOFT_ASSERT_EMAIL = '{{ soft_assert_email }}'
 DAILY_DEPLOY_EMAIL = '{{ daily_deploy_email }}'
+{% if localsettings_private.get("TRANSIFEX_API_TOKEN") %}
```
dannyroberts: We have a pretty well-followed convention of not referring to `localsettings_private` in ansible code outside of `all.yml`. Would you mind adding a line below https://github.com/dimagi/commcare-cloud/blob/99326e0ee2a5f22a94046b92665e72aa0d15d440/src/commcare_cloud/ansible/group_vars/all.yml#L172 that provides a default like the examples around it and then just testing for `{% if TRANSIFEX_API_TOKEN %}` here?

mkangia:
Thanks for the quick review @dannyroberts . That didn't feel right while adding as well. Updated.

PR #1761, opened 2018-05-23
Move more service definitions to ansible

emord:
@dannyroberts it's everything you didn't ask for. :) 

a few things that don't work yet that i note in commits. Celery's the only service still in fab though, which i'll look at soon (there's a bit of extra configuration there.) should be able to get it done before end of the week


emord:
(closing to cut down on noitification noise while I iterate to finish this up)

Issue #1633, opened 2018-04-18
Make git hook not match development/vault.yml or tests/*/vault.yml


pr33thi:

dannyroberts:
fixed here: https://github.com/dimagi/commcare-cloud/pull/1709/files#diff-6337eeb9e904b223332ed534a845b356R8

dannyroberts:
That fixed a different related problem, but not the ones in the issue title

pr33thi:
^Note: that makes sure that only filepaths ending in vault.yml get matched, still need to make it not match on development/ and tests/!

pr33thi:
Fixed in these:
https://github.com/dimagi/commcare-cloud/pull/1762
https://github.com/dimagi/commcare-cloud/pull/1775

PR #1762, opened 2018-05-23
Git-hooks: do not match on paths starting with tests or containing development

pr33thi:
To fix this issue:
https://github.com/dimagi/commcare-cloud/issues/1633

### DIFF
--- a/git-hooks/pre-commit.sh
+++ b/git-hooks/pre-commit.sh
@@ -15,7 +15,7 @@ yellow='\033[1;33m'
 # carriage return hack. Leave it on 2 lines.
 cr='
 '
-for f in $(git diff --cached --name-only | grep -E "VAULT_FILE_INCLUDE_PATTERN" | grep -E "VAULT_FILE_EXCLUDE_PATTERN")
+for f in $(git diff --cached --name-only | grep -E "$VAULT_FILE_INCLUDE_PATTERN" | grep -E "$VAULT_FILE_EXCLUDE_PATTERN")
 do
   MATCH=`grep -L $REQUIRED $f | head -n 1`
   if [ -n "${MATCH// }" ] ; then
--- a/git-hooks/pre-commit.sh
+++ b/git-hooks/pre-commit.sh
@@ -5,7 +5,8 @@
 # If not, commit will fail with an error message
 #
 # File should be .git/hooks/pre-commit and executable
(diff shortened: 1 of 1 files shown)


git-hooks/pre-commit.sh
```diff
@@ -14,7 +15,7 @@ yellow='\033[1;33m'
 # carriage return hack. Leave it on 2 lines.
 cr='
 '
-for f in $(git diff --cached --name-only | grep -E "$FILES_PATTERN")
+for f in $(git diff --cached --name-only | grep -E "FILES_PATTERN_CONTAIN" | grep -E "FILES_PATTERN_NOT_CONTAIN")
```
dannyroberts: You'll need dollar-signs here c.f. `$FILES_PATTERN`.

git-hooks/pre-commit.sh
```diff
@@ -5,7 +5,8 @@
 # If not, commit will fail with an error message
 #
 # File should be .git/hooks/pre-commit and executable
-FILES_PATTERN='(/vault.yml|.*.vault)$'
+FILES_PATTERN_CONTAIN='(/vault.yml|.*.vault)$'
+FILES_PATTERN_NOT_CONTAINS='^((?!development|^tests).)*$'
```
dannyroberts: Maybe `VAULT_FILE_INCLUDE_PATTERN` and `VAULT_FILE_EXCLUDE_PATTERN`?

git-hooks/pre-commit.sh
dannyroberts: You'll also need `-v` to exclude in the second grep

pr33thi: Does the -v say "don't match on this"? If so, I might not need that since the regex pattern I used is to exclude strings:
https://stackoverflow.com/questions/1395177/regex-to-exclude-a-specific-string-constant

dannyroberts: > Is that just how you reference variables in shell scripts?
Yup! And in the (bash) command line in general. (Pretty much anything you can do in a shell script you can also do in the command line and vice versa.)

pr33thi: Cool, did that here: https://github.com/dimagi/commcare-cloud/pull/1762/commits/8e751ace1aaa8ea00ebb60e4f01d2b6d3f8fb5bc

PR #1764, opened 2018-05-23
infer shared_drive_enabled from is_monolith

dannyroberts:
can always override in public.yml, but no need to make every env specify
when it is this predictable

### DIFF
--- a/.travis/environments/travis/public.yml
+++ b/.travis/environments/travis/public.yml
@@ -78,7 +78,6 @@ postgres_users:
 
 encrypted_root: '/opt/data'
 
-shared_drive_enabled: True
 backup_postgres: dump
 backup_blobdb: True
 postgresql_backup_master: True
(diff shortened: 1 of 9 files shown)


snopoke:
Agree with @gcapalbo but as long as we can override it should be OK. Here's a env exactly the setup @gcapalbo mentioned: https://github.com/dimagi/commcare-cloud/compare/sk/echis?expand=1

dannyroberts:
That's a good point @gcapalbo. I _think_ it's still a positive change. I could also see us being smarter about this (and still letting you override). A "smarter" rule might be if the union of groups `proxy`, `webworkers`, `celery`, `pillowtop` is one machine, then set it to true. Maybe even better would be to just set it up if an only if the `[shared_dir_host]` group has any machines under it, and then change swiss and pna to not have anything under that group.

Thoughts?

PR #1766, opened 2018-05-23
regroup pna celery queues

dannyroberts:


environments/pna/app-processes.yml
```diff
@@ -6,18 +6,18 @@ service_blacklist:
...
-    background_queue,case_rule_queue:
+    celery:
       concurrency: 2
       max_tasks_per_child: 1
-    saved_exports_queue:
-      concurrency: 1
+    background_queue,export_download_queue,saved_exports_queue:
+      concurrency: 4
       max_tasks_per_child: 1
       optimize: True
```
emord: `optimize: True` runs python in [optimized mode](https://stackoverflow.com/questions/2055557/python-optimized-mode) and `saved_exports_queue` has been manually verified (not sure on the process @snopoke ?) to not have issues with that. I don't think the same is true for `background_queue` or `export_download_queue`

I think that either optimize should be removed or saved_exports_queue should be kept separate until we know why it's optimized

dannyroberts: https://github.com/dimagi/commcare-cloud/pull/1767

PR #1767, opened 2018-05-23
remove optimize on pna joint queue

dannyroberts:


dannyroberts:
yeah, yaml doesn't care about indentation on a line containing only a comment, it seems. (Tests pass)

PR #1770, opened 2018-05-24
echis env

snopoke:
Mostly echis env specific stuff except for 0e4952102c2ce01e19a3b443a22f98fa73860b94


environments/echis/public.yml
```diff
@@ -0,0 +1,77 @@
...
+    - echis.moh.gov.et  # TODO
+    - 172.19.0.160
+    - 172.19.0.161
+    - 172.19.3.40
+    - localhost
+    - 127.0.0.1
+  CELERY_PERIODIC_QUEUE: 'celery_periodic'
+  CELERY_REMINDER_CASE_UPDATE_QUEUE: 'reminder_case_update_queue'
+  CELERY_REMINDER_RULE_QUEUE: 'reminder_rule_queue'
+  CELERY_REPEAT_RECORD_QUEUE: 'repeat_record_queue'
```
dannyroberts: I'm pretty sure these are just repeats of the defaults in commcare-hq/settings.py. Looks like I could have done more cleanup after some of the celery changes I made a while ago.

environments/echis/public.yml
```diff
@@ -0,0 +1,77 @@
...
+  CELERY_RESULT_BACKEND: 'djcelery.backends.database:DatabaseBackend'
```
dannyroberts: This seems like something that should be specified in commcare-hq/settings.py and not overridden (?). Seems to be `djcelery.backends.database:DatabaseBackend` or `database` in all envs.

environments/echis/public.yml
```diff
@@ -0,0 +1,77 @@
...
+  COUCH_CACHE_DOCS: True
+  COUCH_CACHE_VIEWS: True
+  COUCH_STALE_QUERY: 'update_after'
+  DEPLOY_MACHINE_NAME: "{{ ansible_hostname }}"
+  ELASTICSEARCH_HOST: "{{ groups.elasticsearch.0 }}"
```
dannyroberts: this and the similar lines below could probably be removed from a number of envs if we just started with a better default.

snopoke:
@dannyroberts I agree with all of that but didn't want to change it in this PR

PR #1781, opened 2018-05-25
start postgres after reboot

snopoke:
Noticed that PG doesn't get started by `after-reboot`.

Was introduced here: https://github.com/dimagi/commcare-cloud/commit/cb131fcadbddceb282f01fb6e09acb7cf46033b0

### DIFF
--- a/src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
@@ -160,7 +160,7 @@
 
 - name: Restart postgresql (after reboot)
   command: /bin/true
-  notify: Restart postgres
+  notify: Start postgres
   when: no_tags is not defined
   tags: after-reboot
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
```diff
@@ -206,11 +217,6 @@
   tags:
     - pgbouncer
 
-# Allows to excecute task only when a tag is specified: https://serverfault.com/a/748864
-- shell: /bin/true
-  changed_when: false
-  register: no_tags
-
 - name: Restarts pgbouncer (affect max open files limit)
```
dannyroberts: Interesting. How come this one didn't do it?

snopoke: this only restart pgbouncer, not postgres

PR #1782, opened 2018-05-25
clean up celery settings

snopoke:


environments/staging/public.yml
```diff
@@ -49,9 +49,6 @@ localsettings:
   BANK_NAME: "RBS Citizens N.A."
   BANK_SWIFT_CODE: 'CTZIUS33'
   CELERY_PERIODIC_QUEUE: 'celery_null'
```
snopoke: not sure why this is `celery_null`

src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -138,6 +138,7 @@ BANK_ROUTING_NUMBER_WIRE: "{{ localsettings_private.BANK_ROUTING_NUMBER_WIRE | d
 BITLY_APIKEY: "{{ localsettings_private.BITLY_APIKEY | default('') }}"
 BITLY_LOGIN: "{{ localsettings_private.BITLY_LOGIN | default('') }}"
 BOOKKEEPER_CONTACT_EMAILS: "{{ localsettings_private.BOOKKEEPER_CONTACT_EMAILS | default(None) }}"
+CELERY_RESULT_BACKEND: 'djcelery.backends.database:DatabaseBackend'
```
snopoke: I left this in here for now, wasn't clear if there was a reason it isn't in `settings.py`

environments/staging/public.yml
emord: Possibly to be sure we never run periodic queue on staging. I think there was some issue with not wanting to send emails because some prod domains were mirrored

dannyroberts:
❤️

PR #1787, opened 2018-05-25
ref pg host as 127.0.0.1 if on same machine

dannyroberts:
useful for monoliths without having to futz with the configuration


snopoke:
Why does this matter?

dannyroberts:
I found when trying to deploy to a new monolith (a test server on AWS) that in order to make connections to postgresql work I need to set `DEFAULT_POSTGRESQL_HOST: "127.0.0.1"` in postgresql.yml like is done for swiss (but not PNA because that is apparently using its public IP address internally...)

PR #1791, opened 2018-05-28
Move all service definitions to ansible

emord:
Finished up this work today, but hasn't been deployed anywhere yet.

Intentional design decisions:
- instead of a separate supervisor configuration for each celery worker, they are all defined in one file like our pillows

TODO:

- [x] add supervisorctl reread and supervisorctl update to ansible task
- [ ] clean up permissions. When I did my PR I screwed up `become` vs `become_user` use. will do it after this PR.

To deploy:
- [x] staging
- [x] swiss
- [x] pna
- [x] softlayer
- [x] prod
- [ ] icds


src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
```diff
@@ -53,8 +53,55 @@
...
+  tags:
+    - services
+  register: celery_flower_files
+
+- name: define celery beat service
+  become: cchq
+  template:
+    src: "../templates/supervisor_celery_beat.conf.j2"
+    dest: "{{ service_home }}/{{ deploy_env }}_supervisor_celery_beat.conf"
+  when: app_processes_config.celery_processes.get(inventory_hostname).celery_periodic is defined
```
emord: Historically fabric required beat to be on the same server as the periodic queue. I don't think that's necessary and I think it would limit us to only having the periodic queue on one server. I think we should be able to have multiple of those as long as there's only one beat process?

src/commcare_cloud/fab/operations/supervisor.py
```diff
@@ -1,277 +1,20 @@
...
-
-        if 'celery_periodic' in queue_names:
-            _rebuild_supervisor_conf_file(
-                'make_supervisor_conf',
-                'supervisor_celery_beat.conf',
-                {'celery_params': params}
-            )
-
-
-def show_periodic_server_whitelist_message_and_abort(env):
```
emord: I dropped this warning. I think we can decouple periodic queue and celerybeat which would solve this issue

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
snopoke: Makes sense to me. As far as I understand all `beat` does is send tasks into the queue so it shouldn't matter where the queue worker is.

src/commcare_cloud/fab/operations/supervisor.py
snopoke: We could also do this validation when we validate the environment settings.

snopoke:
build failing because of docs

src/commcare_cloud/fab/operations/supervisor.py
dannyroberts: I'm also in favor of keeping it, and validating it when we validate the environment settings. I think we still get value out of this.

environments/staging/app-processes.yml
```diff
@@ -2,7 +2,8 @@ celery_processes:
   None:
     # celery_periodic is intentionally absent from here,
     # so that no SMSes, etc. ever come from staging.
-    celery_periodic:
+    celery_periodic: {}
+    beat: {}
```
dannyroberts: Does this add celery beat to staging? As the comment above says, we don't want that. If this is a mistake (and not just me misunderstanding the semantics of this new `beat` "queue"), then this is the kind of mistake that having the `server_whitelist` check was meant to prevent.

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -289,9 +289,9 @@ def run(self, args, unknown_args):
         args.playbook = 'deploy_stack.yml'
         unknown_args += ('--tags=supervisor,services',)
         rc = AnsiblePlaybook(self.parser).run(args, unknown_args)
-        if ask("Some celery configs are still updated through fab. "
-               "Do you want to run that as well (recommended)?"):
-            exec_fab_command(args.env_name, 'update_current_supervisor_config')
+        if ask("Would you like to update supervisor to use the new configurations?"):
+            exec_fab_command(args.env_name, 'supervisorctl:"reread"')
+            exec_fab_command(args.env_name, 'supervisorctl:"update"')
```
dannyroberts: Nice! thank you :)

src/commcare_cloud/fab/operations/supervisor.py
emord: I did add https://github.com/dimagi/commcare-cloud/pull/1791/commits/bf2b8fa9032781533e28361e2fb45a2760ef0b74  in my push last night, which does validate there's only one (the one for periodic queue was already there) . are you saying we want to specify what server is hosting beat in two places so that even if you change one, we still warn if they don't match (in the hopes of catching something)?

environments/staging/app-processes.yml
emord: from my understanding, that is adding `beat` to the server `None` so it is just being explicit that we have decided not to run those queues. let me know if my understanding is mistaken

emord:
> What's the rollout plan for this? Will current deploy's continue to work once this get's merged?

Rollout plan will be like the others services I moved over.

1. Warn deployer I'm about to be changing the deploy script, so to not deploy until I've confirmed things are ok
1. Deploy services from this branch to staging
2. Double check the service definitions to make sure nothing has changed
3. Restart/update services 
4. do final check
5. iterate through other envs
6. let deployer know it's safe to deploy

By "current deploys" do you mean the current directory? They will still work.

In this commit I ignore the previous directory we used: https://github.com/dimagi/commcare-cloud/pull/1791/commits/4110b9d4b419a7b5e055d0d27fcf74774a6264b0

environments/staging/app-processes.yml
dannyroberts: Oh, sorry, I missed the none above, thanks

dannyroberts:
> build failing because of docs
you can fix tests by just running `make` locally, and then committing the changes that creates

emord:
assuming interrupt isn't awful tomorrow i'm planning to push this out  (at least to staging/prod/softlayer) tomorrow

emord:
Ran into a few issues here. mostly that our supervisor ansible script is bad. It restarts supervisord twice which means we get like 3 zombie processes for everything that is managed by it. (which is insane for celery processes

emord:
Currently deploying to ICDS. Deployed to every other env

PR #1797, opened 2018-05-29
Bump max connections to postgres on staging

proteusvacuum:


environments/staging/postgresql.yml
```diff
@@ -1,5 +1,7 @@
 override:
   pgbouncer_pool_mode: transaction
+  pgbouncer_max_connections: 400
```
snopoke: with 300 pg connections this could be much higher - like 1000 - though I doubt that's necessary for staging

environments/staging/postgresql.yml
```diff
@@ -1,5 +1,7 @@
...
+  postgresql_max_connections: 300
```
snopoke: the default is 20 so going to 300 is a huge jump, just mentioning

proteusvacuum:
This worked like a charm :+1:

PR #1798, opened 2018-05-29
Remove enterprise branch as a default

emord:
Realized today after I attempted to rebuild the enterprise branch that including pull requests is not going to be the best way forward. 

If the submitter of the pull request has pulled master since the last deploy, that pull request then includes all commits to master since then which basically undoes any use of this branch.

It may be possible to determine what commits are in each branch and then cherry-pick each commit instead of merging the pull request, but I'm open to any other suggestions.

from the original convo: @millerdev @snopoke

### DIFF
--- a/environments/icds-new/fab-settings.yml
+++ b/environments/icds-new/fab-settings.yml
@@ -3,4 +3,3 @@ acceptable_maintenance_window:
   hour_end: 7
   timezone: Asia/Kolkata
 email_enabled: False
-default_branch: enterprise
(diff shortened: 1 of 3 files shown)


snopoke:
Bummer. you could specify individual commits rather than PRs which should be OK since we're only really expecting bug fixes right?

millerdev:
I agree, it's not going to work to merge branches into the enterprise branch like we do for staging. Any change that needs to be deployed to an enterprise env will need to be rebased or cherry-picked onto the most recent release for that env. It might still be nice to maintain an "enterprise" branch, which points to the most recent enterprise release. Hotfixes would be made on this branch in addition to master.

PR #1802, opened 2018-05-31
Explicitly set formplayer database host

emord:
Currently this is the UCR database on ICDS. not sure how intentional that was, but that's at least the current state.

### DIFF
--- a/src/commcare_cloud/ansible/group_vars/all.yml
+++ b/src/commcare_cloud/ansible/group_vars/all.yml
@@ -211,5 +211,11 @@ hq_main_db_host: "{% for config in postgresql_dbs -%}
   {%- endif -%}
 {%- endfor %}"
 
+formplayer_db_host: "{% for config in postgresql_dbs -%}
+  {%- if config.name == formplayer_db_name  -%}
+    {{ config.host }}
+  {%- endif -%}
+{%- endfor %}"
+
 nofile_limit: 4096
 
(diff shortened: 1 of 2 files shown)


emord:
That's true. I was keeping this consistent with the main_db_host that formplayer and airflow already use. Will let @dannyroberts comment on if that's a direction he wants to go with the generated yaml

dannyroberts:
I think putting them in generated.yml makes good sense. In some world it'd be nice if these could be referred to as `postgresql_dbs.formplayer.host`, and then refer to what we currently call `postgresql_dbs` as `postgresql_dbs.all` (which is essentially the same info, helpfully flattened). Not sure if we like this enough go through the pain involved in a rename. The simpler option of just defining `formplayer_db_host` in `.generated.yml` would also work well I think, so I'm okay with that. If the first thing I proposed pops out as really nice, I'm happy to implement it, otherwise Simon's suggestion sounds good

snopoke:
I was also wondering about something like that but apart from these two uses by formplayer I couldn't think of anywhere else they would be used.

dannyroberts:
@snopoke yeah not sure. To make it a little more concrete I went ahead and did the usage side of the refactor to see what it'd look like. (Doesn't include the underlying change to vars that would support this refactor, just the changes to how we'd use it in the scheme i imagined.) https://github.com/dimagi/commcare-cloud/compare/dmr/example-postgresql_dbs-usage-refactor

If that looks like a nicer way to work, I don't think it's a difficult change, otherwise, happy to let this sit.

snopoke:
Looks good to me.

dannyroberts:
Cool I'll just start with this prefactor for now: https://github.com/dimagi/commcare-cloud/pull/1810

PR #1805, opened 2018-05-31
Conditionally run prove-deploy script on travis

pr33thi:


pr33thi:
Summary of the commits I made:
1. Tested with [pv-deploy], realized I hadn't swapped the order of the installs like we discussed earlier
2. Swapped the order of installs
3. Tested with [pv-deploy] again and tests passed (note: it says the tests failed next to the commit, but the required tests all passed, it only says that because I cancelled the fullstack deploy build once it was clear it was running)
4. Tested without [pv-deploy] and tests passed

.travis.yml
```diff
@@ -3,18 +3,42 @@ language: python
...
+    - TEST=main
+    - TEST=prove-deploy
+
+matrix:
+  fast_finish: true
+  allow_failures:
+    - env: TEST=prove-deploy
 
 before_install:
+  - if [[ $TEST = 'prove-deploy' ]] && git log --pretty=format:'%s %b' -n1 HEAD^2 | grep -iv '\[pv-deploy\]'; then exit 0; fi
```
snopoke: What is this line doing?

pr33thi: Yeah that's pretty unreadable, I can add in a comment, but It's only running the prove-deploy test if the commit message contains `[pv-deploy]`:

`if [[ $TEST = 'prove-deploy' ]]`: if the test being run is `prove-deploy`
`&& git log --pretty=format:'%s %b' -n1 HEAD^2` and the commit message of the commit that was merged in to master
`| grep -iv '\[pv-deploy\]'`: doesn't contain the pattern `[pv-deploy]`
`; then exit 0; fi` then exit out of here

snopoke:
ah ok. Is that a temporary measure?

pr33thi:
Yes, the end goal is for this to run every time and not be an allowed failure, but now since it isn't ready, we don't want to waste travis resources by running it every time

snopoke:
Makes sense. Thanks.

PR #1807, opened 2018-05-31
Get rid of the supervisor workaround

pr33thi:
-In bootstrap-users, the `users` role has a few steps that depend on `supervisor` to be installed, namely finding the supervisorctl path, and copying the sudoers file (which has a line that requires `supervisor` to be installed)
-Because bootstrap-users is run as root, we don’t want to add the whole supervisor installation process to it. We currently run `bootstrap-users`, and then run `deploy-stack` again but with the `—tags=supervisor` tag, to install supervisor, and then `update users`, to run the `users` role again and get the sudoers file correctly copied
-To get rid of that workaround, we renamed `users` to `bootstrap-users`, which is what is now run by `bootstrap-users` (so that step doesn’t do anything differently)
-We then created a role called `users`, which has no tasks, but depends on the role `supervisor` and the role `bootstrap-users` (in that order). This gets run during `deploy-stack` -> `deploy-common`, which has a dependency on `users`.


src/commcare_cloud/ansible/roles/users/meta/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+dependencies:
+  - {role: supervisor, tags: supervisor}
+  - {role: bootstrap-users, tags: bootstrap-users}
```
dannyroberts: nit: we have the convention of making all files end in a newline

pr33thi: Added it here
https://github.com/dimagi/commcare-cloud/pull/1807/commits/ba038c9a3d1715d008c2267567584777ea34238a

PR #1809, opened 2018-06-01
Added web10 and web11

NitigyaS:


emord:
Should the known hosts file be updated as well?

PR #1810, opened 2018-06-01
refactor generated postgresql_dbs test

dannyroberts:
Aside from tests, the only change in functionality is that the generated `postgresql_dbs` variable comes sorted by db `name`. The other changes
- sort dbs in `tests/postgresql_config/*/.generated.yml` files to match this output and
- make the test work off the actual output rather than an internal/intermediate data structure
- simplify the test now that the variable comes pre-sorted

This is a prefactor to make changes to `postgresql_dbs` and updates to the test easier to eyeball for correctness

### DIFF
--- a/tests/test_postgresql_config.py
+++ b/tests/test_postgresql_config.py
@@ -14,20 +14,14 @@ TEST_ENVIRONMENTS = os.listdir(TEST_ENVIRONMENTS_DIR)
 
 @parameterized(TEST_ENVIRONMENTS)
 def test_postgresql_config(env_name):
-    def map_json(db_list):
-        return [db.to_json() for db in db_list]
-
-    def get_normalized(db_json_list):
-        return sorted(db_json_list, key=lambda db: db['name'])
-
     env = Environment(DefaultPaths(env_name, environments_dir=TEST_ENVIRONMENTS_DIR))
 
     with open(env.paths.generated_yml) as f:
         generated = yaml.load(f)
         assert generated.keys() == ['postgresql_dbs']
 
-    expected_json = get_normalized(generated['postgresql_dbs'])
+    expected_json = generated['postgresql_dbs']
 
-    actual_json = get_normalized(env.postgresql_config.to_generated_variables()['postgresql_dbs'])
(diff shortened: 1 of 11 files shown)


tests/test_postgresql_config.py
```diff
@@ -14,20 +14,14 @@
...
         generated = yaml.load(f)
         assert generated.keys() == ['postgresql_dbs']
 
-    expected_json = get_normalized(generated['postgresql_dbs'])
+    expected_json = generated['postgresql_dbs']
 
-    actual_json = get_normalized(map_json(env.postgresql_config.generate_postgresql_dbs()))
+    actual_json = env.postgresql_config.to_generated_variables()['postgresql_dbs']
 
     assert actual_json == expected_json, "{} != {}".format(actual_json, expected_json)
```
snopoke: Isn't this just testing that the yml file was written correctly an not that the contents is actually correct?

dannyroberts: I think what this tests is that the actual variable that our code generates matches the generated test files that we've committed to the test dir of the repo. That'll mean that the code we use to generate the variable is correct only if the files we've committed to the test dir are correct. I think that's a reasonable pattern, and basically means that the most important thing to validate in this PR is that the changes I made to the generated files are actually correct, i.e. they are in fact simply resorting the existing list by the `name` key, and otherwise not making any changes. That's unfortunately difficult to see by inspection of the diff. Any testing suggestions you have that would make you more confident in this change? Making room to discuss that kind of concern was sort of one of the motivations behind making this almost no-op PR its own change, so definitely open to suggestions.

snopoke: Right sorry, I wasn't reading this fully and assumed that the test was also generating the file before it read it.

snopoke: Just had a thought. If we named the expected file to something different than the real file then it would be obvious that the contents couldn't be the real file?
```
with open(env.paths.get_env_file_path('expected_generated_postgresql.yml') as f:
         generated = yaml.load(f)
```

Just a thought - not a big deal at all.

PR #1813, opened 2018-06-04
two factor rollout

snopoke:
Rollout:
- [x] staging
- [x] swiss
- [x] softlayer
- [x] pna


environments/pna/public.yml
```diff
@@ -98,6 +98,8 @@ localsettings:
   REMINDERS_QUEUE_ENABLED: False
   SMS_GATEWAY_URL: 'http://gw1.promessaging.com/sms.php'
   SMS_QUEUE_ENABLED: False
+  TWO_FACTOR_SMS_GATEWAY: 'corehq.apps.hqwebapp.two_factor_gateways.Gateway'
+  TWO_FACTOR_CALL_GATEWAY: 'corehq.apps.hqwebapp.two_factor_gateways.Gateway'
```
czue: would we ever change these values? just wondering whether they could be moved to a base file if they are always going to be the same across environments (and maybe overridden in `dev_settings`)

emord: I would be pro making this the default in localsettings.py.j2 and allowing an override in environment settings. Looks like this won't be live on ICDS and I'm not sure if everyone uses dev_settings. (I don't, but that's probably just because I already had the settings i wanted before that was introduced

snopoke: This is already live on ICDS. I can change it so that there is a bit less duplication.

snopoke: Update here: 94335cd

Note that I made this false since you need to configure the gateway on the UI for this to work. Also this will probably change soon so that the config all happens in localsettings.

PR #1814, opened 2018-06-04
[WIP] Added Barman

NitigyaS:
- Added Barman role.
- Modified setup_biderictional_auth.yml to support the user with different names. eg. (barman , Postgres)
It requires following variable to be setup to add barman
```
[barman]
barman_serv

[barman_source]
serv1
serv2
```
```
BACKUP_TO_BARMAN = True
```


src/commcare_cloud/ansible/roles/postgresql/templates/pg_hba.conf.j2
```diff
@@ -5,6 +5,11 @@ local   all             postgres                                peer
 
 # TYPE  DATABASE        USER            ADDRESS                 METHOD
 
+{% if postgres_users.get('barman') and groups['barman'].0 %}
```
snopoke: I think a better check would be `len(groups['barman'])`

src/commcare_cloud/ansible/roles/setup_auth_keys.yml
```diff
@@ -18,13 +18,13 @@
...
   authorized_key:
-    user: "{{username}}"
+    user: "{{userTo}}"
     state: "{{ state }}"
     key: "{{ lookup('file', '/tmp/id_rsa.tmp') }}"
-  delegate_to: "{{ inventory_hostname }}"
+  delegate_to: "{{ hostTo }}"
   ignore_errors: "{{ ansible_check_mode }}"
 
 - name: Deleting temp files
```
snopoke: This file is also referenced here: https://github.com/dimagi/commcare-cloud/blob/e0e133bfa4263d168659b494975525d6cda545e9/src/commcare_cloud/ansible/deploy_db.yml#L39

environments/development/public.yml
```diff
@@ -172,3 +177,5 @@ commcare_cloud_remote_user: vagrant
 commcare_cloud_root_user: ubuntu
 commcare_cloud_strict_host_key_checking: no
 commcare_cloud_use_vault: no
+
+BACKUP_TO_BARMAN: True
```
snopoke: don't think this is used anywhere

src/commcare_cloud/ansible/roles/barman/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+## Barman Config
+barman_configuration_file: "/etc/barman.conf"
+barman_user: "barman"
+barman_configuration_files_directory: "/etc/barman.d"
+barman_home: "{{encrypted_root}}/barman"
+barman_log_directory: "{{encrypted_root}}/barman/log"
+barman_log_file: "{{ barman_log_directory }}/barman.log"
+barman_log_level: "INFO"
+
+barman_users:
```
snopoke: I don't think this is used either

src/commcare_cloud/ansible/roles/barman/templates/streaming-pg.conf.j2
```diff
@@ -0,0 +1,7 @@
+[{{ item }}]
+description =  "{{ item }} (Streaming-Only)"
+conninfo = "host={{ groups[item].0 }} user={{ postgres_users.barman.username }} dbname=postgres password={{ postgres_users.barman.username }}"
```
snopoke: `item` here should already be the host: https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#magic-variables-and-how-to-access-information-about-other-hosts

NitigyaS: I used `host` instead of `groups[item].0` but it is giving the hostname in my case its `db1`. I wanted to configure the IP address of the server here. 
I also tried using `hostvars[host]['ansible_eth0']['ipv4']['address']` but I got error `AnsibleUndefinedVariable: 'dict object' has no attribute 'ansible_eth0'` So I used the above way.

NitigyaS:
@dannyroberts I am getting an error in Travis `AssertionError: Disallowed variables in /home/travis/build/dimagi/commcare-cloud/environments/development/public.yml: set(['backup_to_barman'])` . Could please suggest how I can rectify the same.

src/commcare_cloud/ansible/roles/barman/templates/streaming-pg.conf.j2
snopoke: You could try `hostvars[item]['inventory_hostname']`

NitigyaS: this is also giving `db1` not the IP.

NitigyaS: Thanks worked like charm.

snopoke:
The build failure is because you've put `backup_to_barman` in the postgres role defaults which means that you should override it in `postgresql.yml`. I think it's not really necessary to have `backup_to_barman` at all since you infer it from `len(groups['barman'])`. 

You could also check `inventory_hostname in groups['barman_source']` if you only wanted to apply the PG changes to listed source hosts (though probably not really necessary to do that.

src/commcare_cloud/ansible/roles/barman/templates/streaming-pg.conf.j2
```diff
@@ -0,0 +1,7 @@
+[{{ item }}]
+description =  "{{ item }} (Streaming-Only)"
+conninfo = "host={{ hostvars[item]['inventory_hostname'] }} user={{ postgres_users.barman.username }} dbname=postgres password={{ postgres_users.barman.username }}"
```
snopoke: This should be the same as `item` now that we've fixed the inventory

src/commcare_cloud/ansible/roles/barman/tasks/cron.yml
```diff
@@ -0,0 +1,26 @@
...
+
+- name: Create cron job for backups
+  cron:
+    user: "{{ barman_user }}"
+    name: "barman backup server {{ item }}"
+    cron_file: barman_backup
+    job: "barman backup {{ item }} >> {{ barman_log_directory }}/{{ item }}.log"
+    state: present
+    minute: "0"
+    hour: "0"
```
snopoke: Should make this `"{{ nadir_hour|default(0) }}"`

src/commcare_cloud/ansible/roles/barman/tasks/install.yml
```diff
@@ -0,0 +1,38 @@
...
+- name: Add PostreSQL apt repository
+  apt_repository: repo='deb http://apt.postgresql.org/pub/repos/apt/ {{ ansible_lsb.codename }}-pgdg main' state=present
+  register: add_postgres_repo
+
+- name: Add PosgreSQL apt key
+  apt_key: url=https://www.postgresql.org/media/keys/ACCC4CF8.asc state=present
+
+- name: Update package list
+  apt: update_cache=yes
+  when: add_postgres_repo
```
snopoke: this and the previous 2 are included in `common_installs` role

src/commcare_cloud/ansible/roles/barman/templates/streaming-pg.conf.j2
```diff
@@ -0,0 +1,12 @@
+{% if hostname is defined %}
```
snopoke: shouldn't this be `if 'hostname' in hostvars[item]` or `if hostvars[item].get('hostname')`

src/commcare_cloud/ansible/setup_barman.yml
```diff
@@ -0,0 +1,5 @@
+- name: Barman
+  hosts: barman
+  become: true
+  roles:
+    - {role: barman}
```
snopoke: could we include the ssh auth setup here like we do in `deploy_db.yml` for setting up standby server auth

snopoke:
some conflicts with master

environments/icds-new/inventory.ini
```diff
@@ -927,3 +927,6 @@ kafka0
 
 [mailrelay:children]
 web0
+
+[django_manage:children]
```
snopoke: what's this for?

PR #1815, opened 2018-06-04
Pv/unblock g2

pr33thi:
https://manage.dimagi.com/default.asp?277145
- Changed all user groups to be lists
- Created a g2 user group and added their ssh keys
- Separated out the `dimagi` and `dimagi-qa` group (so that `dimagi-qa` only has Nick Clark)
- Enforce users to be a list here https://github.com/dimagi/commcare-cloud/pull/1815/commits/70e7dba646145367e1057992b79c44e017f34d79
- Make `users_config` combine all the user lists here: https://github.com/dimagi/commcare-cloud/pull/1815/commits/d415ca543d64cea6b2107b940771ff743cf31261


src/commcare_cloud/environment/main.py
```diff
@@ -105,9 +105,16 @@ def proxy_config(self):
...
-            users_json = yaml.load(f)
-        return UsersConfig.wrap(users_json)
+        user_groups_from_yml = self.meta_config.users
+        absent_users = []
+        present_users = []
+        for user_group_from_yml in user_groups_from_yml:
+            with open(self.paths.get_users_yml(user_group_from_yml)) as f:
+                user_group_json = yaml.load(f)
+                present_users += user_group_json['dev_users']['present']
+                absent_users += user_group_json['dev_users']['absent']
```
emord: do the QA & g2 users need to be added here?

dannyroberts: Implicit in my interpretation of your question is the idea that "qa users" and "g2 users" are a different type of user. You can tell me if that's not how you meant it, but I'll address that point and say that there's only one type of user our ansible is aware of, and it calls all of them "dev_users". The breakdown into `dimagi`, `dimagi-qa`, and `g2` is for our own ability to organize which user accounts and ssh keys get added to (or removed from) which environments. For example, staging gets `dimagi`, `dimagi-qa`, `g2`; production gets `dimagi` and `g2`; most of our custom envs get just `dimagi`.

src/commcare_cloud/environment/main.py
```diff
@@ -105,9 +105,16 @@ def proxy_config(self):
...
+        all_users_json = {'dev_users': {'absent': absent_users, 'present': present_users}}
```
dannyroberts: Something I thought of here is can we validate that there's no overlap between `absent_users` and `present_users`? With the extra level of expressiveness that we get from this change, we could also inadvertently mess that up, and until we have a well defined behavior for what that would mean, I think it's good to just fail and say that we need to fix our users files before continuing.

(Make sure that wherever this validation happens, a failure in it results in a `nosetests` failure as well as a `cchq <env> validate-environment-settings` failure.)

src/commcare_cloud/environment/main.py
dannyroberts: Nit: I might dedent these two lines to show that they do not depend on the file (`f`) being open (whereas `yaml.load(f)` does depend on the file being open).

src/commcare_cloud/environment/schemas/meta.py
```diff
@@ -5,5 +5,5 @@ class MetaConfig(jsonobject.JsonObject):
     _allow_dynamic_properties = False
     deploy_env = jsonobject.StringProperty(required=True)
     env_monitoring_id = jsonobject.StringProperty(required=True)
-    users = jsonobject.StringProperty(required=True)
+    users = jsonobject.ListProperty(required=True)
```
dannyroberts: I left a comment directly on the commit: https://github.com/dimagi/commcare-cloud/commit/70e7dba646145367e1057992b79c44e017f34d79#r29236121

src/commcare_cloud/environment/main.py
emord: Ah. yeah that makes sense. somehow skipped over the extra users varialbe in environments. Thanks

pr33thi:
@dannyroberts 
Dedented lines: https://github.com/dimagi/commcare-cloud/pull/1815/commits/0893c4e3eaf06db6d7fdce56f1ac377bb93608a6
Enforce unicode: https://github.com/dimagi/commcare-cloud/pull/1815/commits/dea77acf4c09ce32d92be4ba3f98149543176218

PR #1816, opened 2018-06-04
troubleshooting docs

dannyroberts:
As questions come up and we help people troubleshoot, might be worth adding our suggestions here so that we end up with a bit more of a knowledge bank we can point people to.

### DIFF
--- a/docs/basics/troubleshooting.md
+++ b/docs/basics/troubleshooting.md
@@ -1,0 +1,56 @@
+# Troubleshooting first time setup
+
+## My site is showing "Internal Server Error"
+
+If you are seeing a blank screen with just the words "Internal Server Error" on it,
+it means that the django webworker process is not reporting as "down",
+but still failing to bootstrap fully.
+(If you are seeing a more elaborate 500 page, then that is an issue with a single request,
+but ususally does not indicate a more pervasive problem with the site's ability to receive and handle requests in general.)
+Often this is because it is unable
+using a backing service (such as PostgreSQL, CouchDB, or Redis).
+This in term can fall into a number of categories of issue:
+1. the service is down
+2. the service is up, but unreachable due to network configuration
+3. the service is up and reachable, but is blocking the connection for a permissions-related reason
+   (auth is wrong, IP is not whitelisted, etc.)
+4. the service is up, reachable, and accepting the connection,
+   but is failing due to other problems such as misconfiguration (e.g. a low max connection limit),
+   or other problem such as out of memory errors.
+
+### Further diagnosis
+You can start by backing up and simply checking which of the application services are reporting as up
+by running
+
+```bash
+commcare-cloud <env> service commcare status
+```
+
+You will likely find the `django` process is reporting as RUNNING.
+Some other processes if affected by a similar issue may (or may not) be reporting as FATAL "exited too quickly".
+
+To dig into a particular error, you can log into the machine and tail one of the logs:
+
+```bash
+commcare-cloud <env> ssh webworkers:0
+$ tail -n100 /home/cchq/www/<env>/log/django.log
+```
+
+or, if you do not want to figure out where a particular log lives, you can run the command on all machines
+(allowing that it'll fail on any machine that doesn't contain that particular log):
+
+```
+commcare-cloud <env> run-shell-command all 'tail -n100 /home/cchq/www/<env>/log/django.log'
+```
+
+### Steps to fix
+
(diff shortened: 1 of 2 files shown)


docs/basics/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+## My site is showing "Internal Server Error"
+
+If you are seeing a blank screen with just the words "Internal Server Error" on it,
+it means that the django webworker process is not reporting as "down",
+but still failing to bootstrap fully.
+(If you are seeing a more elaborate 500 page, then that is an issue with a single request,
+but ususally does not indicate a more pervasive problem with the site's ability to receive and handle requests in general.)
+Often this is because it is unable
+using a backing service (such as PostgreSQL, CouchDB, or Redis).
+This in term can fall into a number of categories of issue:
```
snopoke: `tern`

docs/basics/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+
+## My site is showing "Internal Server Error"
+
+If you are seeing a blank screen with just the words "Internal Server Error" on it,
+it means that the django webworker process is not reporting as "down",
+but still failing to bootstrap fully.
+(If you are seeing a more elaborate 500 page, then that is an issue with a single request,
+but ususally does not indicate a more pervasive problem with the site's ability to receive and handle requests in general.)
+Often this is because it is unable
+using a backing service (such as PostgreSQL, CouchDB, or Redis).
```
snopoke: ES, Raik, pgbouncer

docs/basics/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+1. the service is down
+2. the service is up, but unreachable due to network configuration
+3. the service is up and reachable, but is blocking the connection for a permissions-related reason
+   (auth is wrong, IP is not whitelisted, etc.)
+4. the service is up, reachable, and accepting the connection,
+   but is failing due to other problems such as misconfiguration (e.g. a low max connection limit),
+   or other problem such as out of memory errors.
+
+### Further diagnosis
+You can start by backing up and simply checking which of the application services are reporting as up
```
snopoke: `backing up` might be taken to mean 'taking a back up'. Just make it "You can start by simply checking...."

docs/basics/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+You will likely find the `django` process is reporting as RUNNING.
+Some other processes if affected by a similar issue may (or may not) be reporting as FATAL "exited too quickly".
+
+To dig into a particular error, you can log into the machine and tail one of the logs:
+
+```bash
+commcare-cloud <env> ssh webworkers:0
+$ tail -n100 /home/cchq/www/<env>/log/django.log
```
snopoke: I discovered that you can also do this with supervisorctl which doesn't require you finding the right log file:
```
sudo supervisorctl tail -f <process name>
```

docs/basics/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+4. the service is up, reachable, and accepting the connection,
+   but is failing due to other problems such as misconfiguration (e.g. a low max connection limit),
+   or other problem such as out of memory errors.
+
+### Further diagnosis
+You can start by backing up and simply checking which of the application services are reporting as up
+by running
+
+```bash
+commcare-cloud <env> service commcare status
```
snopoke: Before even doing this I'd recommend running the `check_services` Django management command which will straight away point you to a service if that's the issue rather than having to figure it out from the logs.

docs/basics/troubleshooting.md
millerdev: turn?

docs/basics/troubleshooting.md
millerdev: "unable using" -> "unable to connect to"

docs/basics/troubleshooting.md
snopoke: :man_facepalming:

docs/basics/troubleshooting.md
dannyroberts: Whoa, that's a great tip right there!

dannyroberts:
Okay I updated the text to (I think) incorporate all of the feedback. @snopoke, let me know if you're happy with the more prominent place I gave `check_services` and the treatment of checking commcare status and log diving as the final section below it.

PR #1822, opened 2018-06-05
test tweaks

snopoke:


dannyroberts:
(unintentionally restarted the last build)

.travis/tests.sh
```diff
@@ -1,10 +1,12 @@
 #!/bin/bash
 
+set -ve
+
 if [[ ${TEST} = 'main' ]]
 then
 
     test_syntax() {
-        ansible-playbook -i .travis/environments/travis/inventory.ini ansible/deploy_stack.yml --syntax-check
+        ansible-playbook -i .travis/environments/travis/inventory.ini src/commcare_cloud/ansible/deploy_stack.yml --syntax-check
```
dannyroberts: The test failure is that now we rely on `commcare-cloud` putting virtualenv's `site-packages/.ansible` on the ansible roles path. I think we can fix this and make it more future-proof by following the pattern from below and running

```
        cp .travis/environments/travis/private.yml .travis/environments/travis/vault.yml
        COMMCARE_CLOUD_ENVIRONMENTS=.travis/environments commcare-cloud travis deploy-stack --branch=FETCH_HEAD  --skip-check --quiet --syntax-check
```

or alternatively setting this environment variable manually before the current ansible-playbook command

```
ANSIBLE_ROLES_PATH=$VIRTUAL_ENV/lib/python2.7/site-packages/.ansible ansible-playbook ...
```
(I like the first option better

dannyroberts:
Reading between the lines, it looks like for a while now our test failures haven't actually been bubbling through? That's kind of frighting if so—thanks for catching!

snopoke:
Ag, this is now failing on generating the localsettings file: `AnsibleError: Unexpected templating type error occurred on .... Undefined is not JSON serializable`

I can't see the issue, even with debugging. @dannyroberts could you take a quick look and see if you can spot it please.

dannyroberts:
@snopoke I tracked it down to https://github.com/dimagi/commcare-cloud/pull/1831 which fixes an actual problem—so basically in the process of reenabling test you caught a real bug that when through while tests were disabled.

snopoke:
Sweet thanks, and I wasn't on latest master which is why I didn't see it.

snopoke:
@dannyroberts any suggestions for fixing this one:
```
ansible -h
34c34
<                         (default=[u'/home/travis/.ansible/plugins/modules',
---
>                         (default=[u'~/.ansible/plugins/modules',
```

Looks like ansible may have changed their code to expand the `~` in the default path?

.travis/tests.sh
```diff
@@ -1,27 +1,30 @@
...
         COMMCARE_CLOUD_ENVIRONMENTS=.travis/environments commcare-cloud travis deploy-stack --branch=FETCH_HEAD  --skip-check --quiet --tags=commcarehq
         sudo python -m py_compile /home/cchq/www/travis/current/localsettings.py
     }
 
     test_help_cache() {
-        diff <(ansible -h) commcare-cloud/commcare_cloud/help_cache/ansible.txt
-        diff <(ansible-playbook -h) commcare-cloud/commcare_cloud/help_cache/ansible-playbook.txt
+        diff <(ansible -h | sed "s#$HOME#\~#g") src/commcare_cloud/help_cache/ansible.txt
+        diff <(ansible-playbook -h | sed "s#$HOME#\~#g") src/commcare_cloud/help_cache/ansible-playbook.txt
     }
```
dannyroberts: Hey @snopoke so actually I think this test can be removed entirely. `make` now auto-generates these files (https://github.com/dimagi/commcare-cloud/blob/master/Makefile#L3-L7) so `./tests/test_command_docs.sh` (which runs `make` and then sees if there's a diff) is already testing this now.

I guess I would have noticed and removed this test back when I added that, if it had been failing the build!

snopoke:
Woo hoo, finally build passed.

dannyroberts:
🎉 thanks for finding and fixing!

PR #1827, opened 2018-06-06
quote env vars

snopoke:
I discovered that on ICDS without the quotes the env vars
were getting cut off


snopoke:
I think our best option may be to add all hosts to `/etc/hosts` on all VMs and then use a domain prefix for the `no_proxy` value: ie.
```
# /etc/hosts
10.247.164.83 couch3.internal-icds-new.commcarehq.org couch3

# /etc/environment
no_proxy=internal-icds-new.commcarehq.org,localhost,127.0.0.1,....
```

This would also require using hostnames everywhere instead of IP addresses e.g. in localsettings

emord:
> This would also require using hostnames everywhere instead of IP addresses e.g. in localsettings

Is there no way to use an IP range?

PR #1828, opened 2018-06-06
add couchdb2 service

snopoke:
Having both of these is bound to lead to some confusion. I was thinking of printing out a deprecation warning if you use `couchdb` instead of `couchdb2`.

What do you think?


dannyroberts:
Printing some kind of warning when `couchdb` service is used would be useful, since in most cases people will use it intending to manage `couchdb2`.

dannyroberts:
For test failure, I think you just need to run `make`

emord:
I think that the only env that uses couch 1 is swiss. Could we prioritize upgrading that and removing the extra role?

PR #1829, opened 2018-06-06
Add AWS monolith

dannyroberts:
This includes both changes to mostly automate setting up a monolith on AWS, as well as a new monolith test environment, `64-test`, so called because it has 64GB RAM and it's a test environment. This new environment is live at https://64-test.commcarehq.org/


commcare-cloud-bootstrap/specs/example_monolith.yml
```diff
@@ -0,0 +1,53 @@
...
+  rabbitmq:
+    from: proxy
+  kafka:
+    from: proxy
+  zookeeper:
+    from: proxy
+  celery:
+    from: proxy
+  pillowtop:
+      from: proxy
```
snopoke: nit: some of the indents don't match

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
+# For a real deployment
+# should be e.g. 'www.commcarehq.org'
```
snopoke: looking at how this is used I don't think it should ever by `www.commcarehq.org`. It mostly gets used to set machine hostnames

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+
+etc_hosts_lines: []
+etc_hosts_lines_removed: []
+
+touchforms_enabled: false
+
+riak_backend: "bitcask"
+riak_ring_size: 64
+
+ansible_user_password_sha_512: "{{ secrets.ANSIBLE_USER_PASSWORD_SHA_512 }}"
```
snopoke: not related to this PR but I think we could dynamically generate this now: https://stackoverflow.com/questions/19292899/creating-a-new-user-and-password-with-ansible#29910294

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+
+dev_users:
+  present:
+    - dmiller
+    - droberts
+    - jwilson
```
snopoke: jwilson has been removed

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+etc_hosts_lines_removed: []
+
+touchforms_enabled: false
+
+riak_backend: "bitcask"
+riak_ring_size: 64
+
+ansible_user_password_sha_512: "{{ secrets.ANSIBLE_USER_PASSWORD_SHA_512 }}"
+
+dev_users:
```
snopoke: I thought these got moved outside of 'public.yml'

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+elasticsearch_cluster_name: 'deves'
+
+# uncomment to only allow access to eth1
+#ufw_private_interface: eth1
+
+# uncomment and change to the control machine IP when using ufw
+# this ensures that ssh access from the control machine is always allowed regardless of other firewall rules
+#control_machine_ip: 127.0.0.1
+
+ECRYPTFS_PASSWORD: "{{ secrets.ECRYPTFS_PASSWORD }}"
```
snopoke: duplicate of what's in `all.yml`

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+
+KSPLICE_ACTIVE: no
+KSPLICE_ACTIVATION_KEY: "{{ secrets.KSPLICE_ACTIVATION_KEY }}"
+
+DATADOG_ENABLED: True
+DATADOG_API_KEY: "{{ secrets.DATADOG_API_KEY }}"
+DATADOG_APP_KEY: "{{ secrets.DATADOG_API_KEY }}"
```
snopoke: these are also in `all.yml`

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+
+CLOUDANT_CLUSTER_NAME: null
```
snopoke: not necessary to specify this

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+
+# This dumps the dummy keystore, but allows the ansible script to continue, should be overridden for real deploys
+#keystore_file: 'DimagiKeyStore'
+
+backup_es: False
+
+postgres_users:
+  commcare:
+    username: commcarehq
+    password: commcarehq
```
snopoke: any reason not to use strong passwords?

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+    privs:
+      -
+        privs: SELECT
+        objs: ALL_IN_SCHEMA
+  replication:
+      username: 'hqrepl'
+      password: 'hqrepl'
+      role_attr_flags: 'LOGIN,REPLICATION'
+
+encrypted_root: '/opt/data'
```
snopoke: same as default

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+DATADOG_API_KEY: "{{ secrets.DATADOG_API_KEY }}"
+DATADOG_APP_KEY: "{{ secrets.DATADOG_API_KEY }}"
+
+CLOUDANT_CLUSTER_NAME: null
+
+
+# This dumps the dummy keystore, but allows the ansible script to continue, should be overridden for real deploys
+#keystore_file: 'DimagiKeyStore'
+
+backup_es: False
```
snopoke: nit: group with other backup settings

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+postgres_s3: False
+nadir_hour: 18
+
+nofile_limit: 65536
+
+supervisor_http_enabled: True
+supervisor_http_username: "{{ secrets.SUPERVISOR_HTTP_USERNAME }}"
+supervisor_http_password: "{{ secrets.SUPERVISOR_HTTP_PASSWORD }}"
+
+mailrelay: 'relay.example.com'
```
snopoke: remove

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+
+couch_dbs:
+  default:
+    host: "{{ groups.couchdb2.0 }}"
+    port: "{{ couchdb2_proxy_port }}"
+    name: commcarehq
+    username: "{{ localsettings_private.COUCH_USERNAME }}"
+    password: "{{ localsettings_private.COUCH_PASSWORD }}"
+    is_https: False
+  users:
```
snopoke: Don't think you need to specify the `users` DB separately, you can just leave it out.

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+
+TOUCHFORMS_API_USER: 'touchforms@example.com'
+
+localsettings:
+  ALLOWED_HOSTS:
+    - localhost
+    - 127.0.0.1
+    - "{{ SITE_HOST }}"
+  CELERY_PERIODIC_QUEUE: 'celery_periodic'
+  CELERY_RESULT_BACKEND: 'database'
```
snopoke: can remove celery settings

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+  DEPLOY_MACHINE_NAME: "{{ inventory_hostname }}"
+  ELASTICSEARCH_HOST: "{{ groups.elasticsearch.0 }}"
+  ELASTICSEARCH_PORT: '9200'
```
snopoke: can remove ES settings

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+
+  EMAIL_SMTP_HOST: email-smtp.us-east-1.amazonaws.com
+  EMAIL_SMTP_PORT: 587
+  EMAIL_USE_TLS: yes
+  FULLSTORY_ID:
+  HQ_INSTANCE: 'dev'
+  J2ME_ADDRESS: ''
+  KAFKA_URL: "{{ groups.kafka.0 }}:9092"
+  LOCAL_PILLOWS:
```
snopoke: do you want local pillows?

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+      - name: 'kafka-ucr-main-custom-2'
+        class: 'corehq.apps.userreports.pillow.ConfigurableReportKafkaPillow'
+        instance: 'corehq.apps.userreports.pillow.get_kafka_ucr_static_pillow'
+        params:
+          ucr_division: 'af'
+  REDIS_DB: '0'
+  REDIS_HOST: "{{ groups.redis.0 }}"
+  REDIS_PORT: '6379'
+  STATIC_CDN: ""
+  FORMPLAYER_URL: "http://{{ groups.formplayer.0 }}:8181"
```
snopoke: I think this should be `"/formplayer"`

environments/64-test/public.yml
```diff
@@ -0,0 +1,177 @@
...
+master_list_email: droberts+master-list@dimagi.com
+report_builder_add_on_email: droberts+sales@dimagi.com
+eula_change_email: droberts+eula-notifications@dimagi.com
+contact_email: droberts+info@dimagi.com
+soft_assert_email: droberts+commcarehq-ops+soft_asserts@dimagi.com
+daily_deploy_email: null
+
+commcare_cloud_root_user: ubuntu
+commcare_cloud_strict_host_key_checking: no
+commcare_cloud_pem: /Users/droberts/.ssh/commcarehq_testing.pem
```
snopoke: is this right?

src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -183,6 +183,7 @@ ECRYPTFS_PASSWORD: "{{ secrets.ECRYPTFS_PASSWORD | default(None) }}"
 KSPLICE_ACTIVATION_KEY: "{{ secrets.KSPLICE_ACTIVATION_KEY | default(None) }}"
 NEW_RELIC_KEY: "{{ secrets.NEW_RELIC_KEY | default(None) }}"
 
+# use https://v-punk.com/automate-password-changes-with-ansible/ to generate hash
```
snopoke: It's actually much easier now that ansible has a `password_hash` filter: https://stackoverflow.com/questions/19292899/creating-a-new-user-and-password-with-ansible#29910294

src/commcare_cloud/fab/operations/supervisor.py
```diff
@@ -44,8 +44,9 @@ def set_supervisor_config():
 
 def _get_celery_queues():
     queues = {}
-    queues.update(env.celery_processes.get(env.get('host_string').split(':')[0], {}))
-
+    inventory_hostname = env.ccc_environment.inventory_hostname_map[env.host_string]
+    queues.update(env.celery_processes.get(inventory_hostname))
+    print(queues)
```
snopoke: print

dannyroberts:
Thanks for the review! As you probably guessed, these files were auto generated from files that were originally copied from the development environment, so later today when I fix these, I'll probably do it by updating those templates and then regenerating the file(s).

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -53,6 +53,16 @@ class AwsConfig(StrictJsonObject):
...
     subnet = jsonobject.StringProperty()
+    data_volume = jsonobject.DictProperty(exclude_if_none=True)
+    boot_volume = jsonobject.DictProperty(exclude_if_none=True)
+
+    @classmethod
+    def wrap(cls, data):
+        if 'data_volume' in data:
+            assert data['data_volume']['DeviceName'] == '/dev/xvdb'
+        if 'boot_volume' in data:
+            assert data['boot_volume']['DeviceName'] == '/dev/sda1'
```
millerdev: Is it our convention to use `/dev/xvdb` and `/dev/sda1`, or do AWS machines always use those devices for data and boot (that would be surprising to me)? If it's a convention, is it documented somewhere?

Unconditionally overwriting these values here seems like it could result in unexpected outcomes if `AwsConfig` is ever used in a context where other devices are being used.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -90,14 +100,17 @@ class Group(StrictJsonObject):
...
+def provision_machines(spec, env_name=None, create_machines=True):
     if env_name is None:
         env_name = u'hq-{}'.format(
             ''.join(random.choice(string.ascii_lowercase + string.digits) for _ in range(7))
         )
     environment = get_environment(env_name)
     inventory = bootstrap_inventory(spec, env_name)
-    instance_ids = ask_aws_for_instances(env_name, spec.aws_config, len(inventory.all_hosts))
+    if create_machines:
+        instance_ids = ask_aws_for_instances(env_name, spec.aws_config, len(inventory.all_hosts))
```
millerdev: It seems odd to me that a function called `provision_machines` in a module named `commcare_cloud_bootstrap`, both fairly generic and could be expected to work in many hosting contexts, would make specific references to AWS. Is it expected that this module will only ever be used to bootstrap commcare clouds on Amazon servers?

I see there were references to AWS here before this PR, so this comment probably does not apply only to the changes being made here.

Thinking out loud a bit more: I suppose we're writing this for AWS first, and then maybe someday if we ever use it on some hosting provider other than Amazon we'll abstract out the references to AWS around a more generic API. Is it true that right now it's too early to know what the generic API would look like, and so it's easier to just reference AWS even though it's overly specific?

environments/64-test/public.yml
dannyroberts: Yeah, well, this is just what I used to bootstrap it; no one else needs this file to work on this env. Since it doesn't make much of a difference I think I'm just going to leave it

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: This code really has no current calling/purpose besides being used for the provable deploy, and for setting up test monoliths like this one. For that reason, I'm less concerned with things being general than I am with things working for a few predefined use-cases.

For aws EC2 instances treat `/dev/sda1` as the boot volume, so this assert I don't expect to change. (The reason I have it there at all is because in the spec file, the YAML for each disk is exactly the JSON you'd send via the `aws` command line tool, so I thought I'd just make it keep that form and check that the device is set correctly.)

For the `data_volume`, it's true this could be anything. I'd only tested it on `/dev/xvdb`, and didn't see a use for setting it to something else, so I thought I'd just have it tell you to set it to that if you had it set to anything else (or left unset), possibly avoiding an issue later with untested configuration. For now I'll just leave comments for each of these asserts with context, saying that if you ever want to, you could probably relax the `data_volume` one.

dannyroberts: Actually, I'm going to just remove the `/dev/xvdb` assert. I think it's more confusing than helpful.

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
dannyroberts: Yeah, so right now this is set up only for me and Preethi to use for what we're doing right now, and is intentionally not being presented as a more general purpose tool. Some of the things it does are sort of test beds for actual tools we'll probably want to build, such as the ability to generate environment files with reasonable presets, etc.; when we want to roll out those tools, we'll probably pull out code from here and put it somewhere appropriate and accessible, document it, etc.

For now, this is code for automating a particular setup involving AWS. Depending on what we actually use it for in the future, I'm sure it'll morph into something else, but I didn't want to code to a general problem that we're not yet facing, and am not even sure we'll stick with this versus replacing some of it with a commonly used tool like terraforms or something.

environments/64-test/public.yml
dannyroberts: Nice! Yeah we should use that

dannyroberts:
Alright, I think I've addressed all the comments! Let me know if I missed anything

dannyroberts:
Oops, some unaddressed comments were hidden, couple more commits on the way

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
millerdev: Wow, I don't know how I overlooked what's really going on here. My comment must have been confusing. I wrote

> Unconditionally overwriting these values...

Clearly these are asserts, and not assignments, and that makes my comment completely invalid. Sorry!

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
millerdev: > I didn't want to code to a general problem that we're not yet facing

👍 makes sense

environments/64-test/public.yml
dannyroberts: Thanks, good catch. I just moved this config for all the "example" environments (development, travis, bootstrap) into the private/vault file, and changed the password to something "passwordy" so its easier to find and replace with a real password. (Longer term we're probably going to want to be auto-generating these passwords for a new env, but this smaller change puts is more in line with the current convention.)

For this current env (64-test), we're just going to be doing some loadtesting and then probably deleting the actual machine. For just that, I'm fine leaving the weak password here, especially because of how strict postgresql IP whitelisting is (so even if there's some terrible network misconfiguration, someone will have to ssh into the machine before being able to access postgresql). If we reuse the config behind this env, I'll update vault to have a real password for this.

dannyroberts:
@snopoke Okay now I think I've responded to everything

PR #1830, opened 2018-06-06
idempotent couchdb

snopoke:
This is an effort to make the couchdb role idempotent. Unfortunately I don't think it's quite possible without forking the galaxy role that we use since by the time test tasks fire the DB will already be running and have generated some of the values for itself.

The issue is that couchdb updates the config files which means that if you re-run the ansible role it will wipe out those values. To make this idempotent we need to make sure that ansible writes all the values into the config files and not couchdb.


snopoke:
Going to merge this for now until we can get it in the galaxy role

PR #1832, opened 2018-06-07
Quickstart docs (to be continued...)

dannyroberts:


docs/basics/0002-installation.md
```diff
@@ -0,0 +1,88 @@
...
+
+(Did that work? Only type the text starting after the `$` in these examples.)
+
+You should see a prompt like
+
+```bash
+admin@control.example.com:~$
+```
+
+Run this command to verify that you are in the home directory of the `commcare-cloud` user.
```
snopoke: I think you mean the `admin` user

docs/basics/0002-installation.md
```diff
@@ -0,0 +1,88 @@
...
+
+```bash
+admin@control.example.com:~$ pwd
+/home/commcare-cloud
```
snopoke: `/home/admin`

docs/basics/0002-installation.md
```diff
@@ -0,0 +1,88 @@
...
+docs                      Makefile      scripts       Vagrantfile
+```
+
+If you see something like
+
+```bash
+ls: cannot access commcare-cloud: No such file or directory
+```
+
+then the `git clone` command did not run correctly
```
snopoke: Add: "Make sure you have git installed and run it again with `--verbose` to give more logging output"?

docs/index.md
```diff
@@ -44,6 +44,7 @@ an understanding of the system's architecture and each of the system's underlyin
 
 ## System installation basics
 
+- [Getting started guide](basics)
```
millerdev: Should this link to `basics/index.md`? As is it doesn't work from https://github.com/dimagi/commcare-cloud/blob/369706f1596e57800f9a29b3db43acdb6043b48f/docs/index.md#system-installation-basics

dannyroberts: Oh interesting. I was writing this with the site (http://dimagi.github.io/commcare-cloud/) in mind, and in that context it works fine (i.e. the corresponding page on my local test server, http://127.0.0.1:4000/basics/, does in fact read from the that `index.md` file). Do you think it's important to make it work in the browsing-docs-code-on-github context?

millerdev: Ah, makes sense.

> Do you think it's important to make it work in the browsing-docs-code-on-github context?

I guess not since the link in github code context takes you to a directory listing, and it's fairly obvious you want to start with `index.md`

PR #1836, opened 2018-06-11
combine swiss queues

emord:
the swiss server always has high memory usage, so figured we should combine these queues so kafka stops dying during deploy


emord:
@dannyroberts I didn't remove async restore and ucr queues for performance reasons, just for clarity. I don't think they'll be able to accidentally turn on async queues and UCRs would be broken anyways because the UCR pillows aren't running on this server either.

We can add these back to an existing queue or separate queue if they ever need those features, but it felt more clear to me to only run what's needed

PR #1838, opened 2018-06-12
adding django manage group to environments and deploying hq to it

jmtroth0:
@dannyroberts


jmtroth0:
this only affects prod and softlayer right now, since they are both in decent use and of a similar structure, as well as because i'm running the migration regularly on both of them.

i can definitely put off softlayer for now to make sure this works as expected though

PR #1839, opened 2018-06-12
use django manage

jmtroth0:
can merge once #1838 is out and the code has been deployed to prod and softlayer

### DIFF
--- a/environments/development/inventory.ini
+++ b/environments/development/inventory.ini
@@ -98,3 +98,6 @@ swap_size=100M
 
 [control:vars]
 hostname='control'
+
+[django_manage:children]
+webworkers
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -111,7 +111,7 @@ class Tmux(_Ssh):
     arguments = (
         Argument('server', help="""
             Server to run tmux session on.
-            Use '-' to for default (webworkers:0)
+            Use '-' to for default (django_manage, which depends on env)
         """),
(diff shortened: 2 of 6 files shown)


jmtroth0:
@dannyroberts @emord updated this to work on (i think) all envs and for all cchq commands. let me know if i missed anything

environments/development/inventory.ini
```diff
@@ -98,3 +98,6 @@ swap_size=100M
 
 [control:vars]
 hostname='control'
+
+[django_manage:children]
+webworkers
```
emord: Would this be multiple machines if someone spun up webworkers?

dannyroberts: It would by my eye, which I'd suggest we get around by using `django_manage:0` below to pick off the first one.

jmtroth0: yeah, i'll add `django_manage:0` to the places i'm selecting them

dannyroberts:
I think you can fix the build by running `make` and then committing the changed files. (If there are any untracked files, make sure you don't add those! All changes `make` makes should be to existing files.)

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -111,7 +111,7 @@ class Tmux(_Ssh):
     arguments = (
         Argument('server', help="""
             Server to run tmux session on.
-            Use '-' to for default (webworkers:0)
+            Use '-' to for default (django_manage:0, which depends on env)
```
dannyroberts: I think you can just have this say `Use '-' for default (django_manage:0)`, fixing the typo in the original. (Remember to run `make` again!)

PR #1841, opened 2018-06-14
Add seconds to the dateformat

pr33thi:
This means deploys can be attempted within the same minute, which will make testing faster. It didn't look like it would break anything because it's used everywhere deploys are referenced.

### DIFF
--- a/src/commcare_cloud/fab/const.py
+++ b/src/commcare_cloud/fab/const.py
@@ -41,7 +41,7 @@ ROLES_AIRFLOW = ['airflow']
 
 RELEASE_RECORD = 'RELEASES.txt'
 KEEP_UNTIL_PREFIX = 'KEEP_UNTIL__'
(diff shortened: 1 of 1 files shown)


dannyroberts:
@snopoke do you think this change will have any bad effects? Does other code depend on the format of these directories? (This will also change the name in the github tag.)

@pr33thi another way to solve this would be to have it ask you if you want to just delete the conflicting one as long as it's not `current` (?)

snopoke:
Curious where this is an issue. 

I think this will cause a problem with `clean_releases` since it will attempt to parse old release folders using the new date format.

pr33thi:
@snopoke I made this because I was running deploys one after another in the process of testing my other branch, and kept getting an error because, as long as it was within a minute, the release was under the same name as an existing release.

@dannyroberts  I'll try that solution and close this PR

PR #1848, opened 2018-06-19
use pip-tools for requirements

dannyroberts:


dannyroberts:
I think this is going to be too annoying.

PR #1850, opened 2018-06-20
Ns/icds certbot

NitigyaS:


src/commcare_cloud/ansible/roles/nginx/tasks/install.yml
```diff
@@ -138,62 +138,44 @@
...
-    content: "{{ cas_nginx_combined_cert_value }}"
+  file:
+    src: "/etc/letsencrypt/live/{{CAS_SITE_HOST}}/fullchain.pem"
     dest: "{{ ssl_certs_dir }}/{{ cas_nginx_ssl_cert }}"
-    mode: 0400
-    owner: root
-    group: root
-  when: not fake_ssl_cert and cas_nginx_combined_cert_value is defined
+    state: link
+  when: not fake_ssl_cert and CAS_SITE_HOST | default(None) and letsencrypt_cchq_ssl
```
snopoke: why the need to check `letsencrypt_cchq_ssl`

NitigyaS: Thanks for catching it Simon. That defenetly makes sense not to use it. 
There is still some work left in this PR. Sorry I did not marked it [WIP]

PR #1859, opened 2018-06-21
Staging riakcs => aws

dannyroberts:
This just sets up writes to go to AWS S3. Reads still fall back to riakcs if they're not in S3.


environments/staging/public.yml
```diff
@@ -36,6 +36,10 @@ couch_dbs:
 
 DROPBOX_APP_NAME: 'CommCareHQFiles - Staging'
 
+s3_blob_db_enabled: 'true'
```
snopoke: I think this should just be `true` (without the quotes)

src/commcare_cloud/ansible/deploy_proxy.yml
```diff
@@ -106,6 +106,7 @@
         hosts_group: "{{ 'riakcs_new' if 'BLOB_DB_MIGRATING_FROM_S3_TO_S3' in localsettings else 'riakcs' }}"
         riakcs_proxy_port: "{{ temp_riakcs_proxy_port }}"
         site_config: {name: riakcs, vars_file: riakcs}
+      when: "'riakcs_new' in groups"
```
millerdev: It's not needed yet, but once this migration is complete (and there is no riak cluster) we'll want `when: "'riakcs' in groups"` above on the `riakcs site` task (line 91).

src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -76,6 +77,7 @@ s3_blob_db_url: "http://{{ groups['proxy'][0] }}:{{ temp_riakcs_proxy_port if 'B
 # - remove `localsettings.TEMP_RIAKCS_PROXY: True` from environments/$ENV/public.yml
 # - deploy proxy (new=8080)
 old_s3_blob_db_url: "http://{{ groups['proxy'][0] }}:{{ riakcs_proxy_port }}"
+old_s3_blob_db_s3_bucket: 'blobdb'
```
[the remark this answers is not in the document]
dannyroberts: because now we're in the realm of these being Amazon S3 buckets and S3 buckets are global namespace within all of Amazon S3 customers. So calling our bucket "blobdb" doesn't make sense for S3. We're not changing it for the old one (riakcs) for staging, but the new one (S3) has a custom bucket name:
```
+s3_blob_db_s3_bucket: 'dimagi-commcare-staging-blobdb'
```

environments/staging/public.yml
dannyroberts: Yeah I wasn't totally sure about this, but the line in `all.yml` is

```
s3_blob_db_enabled: "{{ 'true' if groups.get('riakcs') else '' }}"
```

Maybe I'm lapsing on my ansible syntax knowledge; my impression is that this would leave this value with one of the _strings_ `'true'` or `''`. Rather than digging into it, I just followed the pattern. Looking into it now, it seems like that value is only ever used for its truthiness, so

```
s3_blob_db_enabled: "{{ True if groups.get('riakcs') else False }}"
```

should do just as well, and then I can change this line to

```
s3_blob_db_enabled: yes
```
like you suggested

src/commcare_cloud/ansible/deploy_proxy.yml
dannyroberts: Okay nice thinking. I'll add that now.

dannyroberts: Oh, you know what @millerdev? I think that this condition should be added to the `site_present` variable rather than a `when` clause: `set_site_present` both adds _and removes_ sites, based on the value of `site_present`. Using a `when` clause will at best _not add_ a site, but won't actively remove it when values change.

Along those lines, you'll see that "riakcs site" above already has `site_present: "{{ groups.get('riakcs') }}"`, so I think the issue behind your comment is already taken care of.

I'm going to go ahead and follow that pattern, adding `groups.get('riakcs_new')` to the `site_present` variable in this "riakcs new" task, and removing the `when`.

src/commcare_cloud/ansible/group_vars/all.yml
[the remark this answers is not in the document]
snopoke: :+1: I missed that this was getting overridden in the env vars

environments/staging/public.yml
millerdev: I thought there were cases where true/false values got converted to a string (`'True'` or `'False'`), making them both truthy in future evaluations. Using `'true'` and `''` evades this since any non-empty string is truthy and an empty string is falsey. You can ignore my input here if we know the issue with Ansible has been fixed or if I'm just mis-remembering (e.g., if no one else has ever observed that issue in this context).

dannyroberts: Interesting. I think that in this context they do evaluate correctly to booleans. The test I came up with was to run `cchq swiss update-config --branch=dmr/riakcs-to-aws` (in just check mode). Since there it's supposed to be set to `False`, if it's not getting set correctly it'll be interpreted as truthy and then that would change some values in localsettings.py; when I ran it I didn't see any related changes, so I'm interpreting that as implying we're safe here.

dannyroberts:
got verbal approval from @millerdev on the last commit, merging

PR #1860, opened 2018-06-22
Addition of 2 pillows

NitigyaS:
@snopoke what should I put in the app-processes.yml


snopoke:
@emord can you fill in the processes you want to run on these 2?

emord:
Split up the case to elasticsearch pillow to 36. We're ok right on all pillows but this will allow better processing when we finish up the case reindex work

emord:
Looks like these weren't deployed to (and they aren't in the pillowtop group):

```
jemord@pillow10:~$ ls /home/cchq
datadog
```

was this intentional?

NitigyaS:
Sorry for missing it. Added in https://github.com/dimagi/commcare-cloud/pull/1896.

PR #1863, opened 2018-06-22
Start lazy migration riak=>s3 on production

dannyroberts:


dannyroberts:
Recording here for posterity, looks like this may have caused noticeable increase in site-wide average request timings.

screen shot 2018-07-09 at 1 29 21 pm

PR #1865, opened 2018-06-25
increasing redis memory

calellowitz:
@emord cc: @snopoke 
After some investigation I believe the constant logouts on prod are caused by redis running out of allowed memory and evicting old sessions. This PR increases allowed memory for redis from 4 to 6gb on prod, where it currently resides on its own 8gb machine. I set the memory to this using runtime config changes on friday, but it was reset over the weekend. It did appear to max out before that so it is not clear this is high enough (perhaps 7 is better to try since nothing else is on the machine). Based on this, at some point, we will also probably need a new bigger redis machine (note for @dannyroberts while planning aws migration), but hopefully this will delay that.

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -46,6 +46,8 @@ couch_dbs:
 
 TWO_FACTOR_GATEWAY_ENABLED: True
 
+redis_maxmemory: 6gb
+
 localsettings:
   ALLOWED_HOSTS:
     - www.commcarehq.org


emord:
> Based on this, at some point, we will also probably need a new bigger redis machine

At some point we were looking at redis clustering. Maybe we'd just need to two sharded redis nodes>

snopoke:
Interesting, this is not what I was expecting - how did you come to this conclusion?

calellowitz:
I was suspicious because the logouts did not appear to follow a regular
 time schedule, so I checked the next time I was logged out and my session
did not appear to exist in redis. After that, I looked at the redis
dashboards in datadog and memory had been at the configured limit for the
past two weeks and key evictions seemed pretty high. I switched it Friday
and stayed logged in all weekend, and verified that no key evictions had
happened. Today, I was logged out this morning but it looked like both the
memory setting had been reset and evictions were occurring.

dannyroberts:
This caused the disk to fill up on the redis machine with `temp-rewriteaof-*.aof` files, which made redis respond to most (all?) requests with an error. More details in a retro to come. Just noting here so we don't re-deploy this!

proteusvacuum:
Also, it doesn't seem like this part actually happened:
> but it was reset over the weekend

Running `CONFIG GET *` showed:
```
17) "maxmemory"
18) "6442450944"
```

snopoke:
So what's the outcome of this? Do we need a cache VM with more memory?

calellowitz:
> Do we need a cache VM with more memory?

I still believe this is the cause of the logout issues, and given that it looks like we cannot deploy this without increasing the size of the vm, I would say yes. I do not have a sense of how difficult it is to move redis to a new vm (or whether it requires downtime), so am not sure on timeframes or prioritziation.

dannyroberts:
One option is to prioritize moving to AWS ElastiCache sooner, which we'll be doing as part of the AWS migration anyway. Otherwise it's just setting up a new redis machine, rsync'ing, topping off with a little downtime and switching over.

PR #1870, opened 2018-06-26
Update public.yml and inventory.ini.j2

pr33thi:
1. Add to the inventory.ini.2 template and entry that will add to inventory.ini:
```
[<host>:vars]
hostname='<host>'
```
for each host

2. Add an internal domain name so that this task does not get skipped here: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/common/tasks/main.yml#L74

Reason: In order for kafka to start up, the hostname needs to be in `/etc/hosts`, and this was getting skipped because `internal_domain_name` was set to `null` here: https://github.com/dimagi/commcare-cloud/commit/d46527f156e4837c59ede9c7a47eba86b093c842

### DIFF
--- a/commcare-cloud-bootstrap/environment/inventory.ini.j2
+++ b/commcare-cloud-bootstrap/environment/inventory.ini.j2
@@ -11,4 +11,10 @@
 {%- endfor %}
 {% endfor %}
 
+{% for _, group in inventory.all_groups.items()|sort %}
+[{{ group.name }}:vars]
+hostname='{{ group.name }}'
+{{ host_name }}
+{%- endfor %}
+
 [control]
--- a/commcare-cloud-bootstrap/environment/public.yml
+++ b/commcare-cloud-bootstrap/environment/public.yml
@@ -1,4 +1,4 @@
-internal_domain_name: null
+internal_domain_name: 'fullstack-deploy.commcarehq.org'
 
 riak_backend: "bitcask"
 riak_ring_size: 64


commcare-cloud-bootstrap/environment/inventory.ini.j2
```diff
@@ -11,4 +11,10 @@
 {%- endfor %}
 {% endfor %}
 
+{% for _, group in inventory.all_groups.items()|sort %}
+[{{ group.name }}:vars]
+hostname='{{ group.name }}'
+{{ host_name }}
+{%- endfor %}
+
```
dannyroberts: Oh, you know what? I think there was a better way to do this that wouldn't cause https://github.com/dimagi/commcare-cloud/pull/1883. There are already places in `commcare_cloud_bootstrap.py` where we add variables, and then those are added to this template in the line above
```
{%- for key, value in machine.vars.items()|sort %} {{ key }}={{ value }}{% endfor %}
```

pr33thi: I jus tested it again without my change, and the lines that looks like 
```
[celery:vars]
hostname='celery'
```
don't show up. I can look into why that is!

pr33thi: Oh yep, that works.. I'm not sure why it didn't work the first time I had this, maybe I just didn't pull from the latest master or something.

pr33thi: https://github.com/dimagi/commcare-cloud/pull/1884

PR #1871, opened 2018-06-26
Normalize inventories for production and staging

dannyroberts:
put each host into its own group and form other groups from these host groups

The impetus for this is so we can add ec2 instances as they come, in a strategy I'm writing up that I think will make things as simple as possible. (Regardless of that, I do think it's nice to have these files in basically the same format as our other inventories.) This doesn't do anything to use IP addresses instead of DNS, and I'm think I'll just let that become irrelevant with the AWS migration.


environments/staging/inventory.ini
```diff
@@ -1,53 +1,90 @@
-[proxy]
+[proxy0-staging]
```
snopoke: why the `-staging` suffix? We don't use that anywhere else.

dannyroberts: I just left it to match the host name in `proxy0-staging.internal-va.commcarehq.org`. Happy to get rid of it, since I don't think there's any practical implications and it's maybe cleaner.

PR #1872, opened 2018-06-27
Add a django_manage group which contains the webworker

pr33thi:
This will make the `django_manage` group show up in the inventory.ini file, with the webworkers0 machine in the group when it is autogenerated during the fullstack deploy.

I also changed the couchdb proxy group to use the proxy machine, which already needs to have nginx running (so no longer 2 different machines running nginx). This might have made it easier to fix a bug @dannyroberts and I looked at earlier today.

Without this, we get an error when this ansible task is run: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/shared_dir/tasks/setup_host.yml#L72

### DIFF
--- a/.travis/spec.yml
+++ b/.travis/spec.yml
@@ -12,6 +12,9 @@ settings:
 
 allocations:
   proxy: 1
+  couchdb2_proxy:
+    count: 1
+    from: proxy
   webworkers: 1
   postgresql: 1
   rabbitmq: 1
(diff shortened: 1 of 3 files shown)


pr33thi:
Closing because this has some reverts in the commit history that I don't want. Reopened with just the good stuff here: https://github.com/dimagi/commcare-cloud/pull/1879

proteusvacuum:
Next time you can do a rebase and force-push to this branch

pr33thi:
Good point!

PR #1873, opened 2018-06-27
update icds app download limits

snopoke:
@dimagi/scale-team


emord:
Do we actually need to rate limit these? I think they're just served from the nginx cache right?

Also we may need to increase the cache size if we ad dmore languages

snopoke:
https://github.com/dimagi/commcare-cloud/pull/1878

PR #1878, opened 2018-06-28
remove rate limiting from app downloads 

snopoke:
And add cache locking.

Docs for cache locking: https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cache_lock


snopoke:
@emord yes I would like to add most of these 'features' to prod at some point including the rate limiting.

PR #1879, opened 2018-06-28
Add a django_manage group which contains the webworker

pr33thi:
This is the same as this pr: https://github.com/dimagi/commcare-cloud/pull/1872, but I noticed that it had some reverts in the commit history that I don't want. This contains only the changes I wanted to make in 1 commit.
=======Description copy pasted from other ticket=======
This will make the django_manage group show up in the inventory.ini file, with the webworkers0 machine in the group when it is autogenerated during the fullstack deploy.

I also changed the couchdb proxy group to use the proxy machine, which already needs to have nginx running (so no longer 2 different machines running nginx). This might have made it easier to fix a bug @dannyroberts and I looked at earlier today.

Without this, we get an error when this ansible task is run: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/shared_dir/tasks/setup_host.yml#L72

### DIFF
--- a/.travis/spec.yml
+++ b/.travis/spec.yml
@@ -12,7 +12,13 @@ settings:
 
 allocations:
   proxy: 1
+  couchdb2_proxy:
+    count: 1
+    from: proxy
   webworkers: 1
+  django_manage:
+    count: 1
+    from: webworkers
   postgresql: 1
   rabbitmq: 1
   kafka: 1
@@ -30,7 +36,4 @@ allocations:
     count: 1
     from: riakcs
   couchdb2: 1
-  couchdb2_proxy:
-    count: 1
-    from: couchdb2
   pg_standby: 0
--- a/commcare-cloud-bootstrap/specs/example_spec.yml
+++ b/commcare-cloud-bootstrap/specs/example_spec.yml
@@ -9,7 +9,13 @@ aws_config:
 
 allocations:
   proxy: 1
+  couchdb2_proxy:
+    count: 1
+    from: proxy
   webworkers: 1
+  django_manage:
+    count: 1
+    from: webworkers
   postgresql: 1
   rabbitmq: 1
   kafka: 1
@@ -27,7 +33,4 @@ allocations:
     count: 1
     from: riakcs
   couchdb2: 1
-  couchdb2_proxy:
-    count: 1
-    from: couchdb2
   pg_standby: 0


commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -300,33 +300,34 @@ def get_inventory_from_file(environment):
...
-                    private_ip, variables = line_groups[0], line_groups[1:]
-                    variables = dict(var.strip().split('=') for var in variables)
-                    public_ip = variables.pop('ansible_host')
-                    host = Host(name=current_host_name, private_ip=private_ip, public_ip=public_ip,
-                                vars=variables)
-                    inventory.all_hosts.append(host)
-                elif state == 'parsing-group':
-                    host_name = line.strip()
-                    current_group.host_names.append(host_name)
+                group_line_match = re.match(r'^\[\s*(.*)\s*\]\s*$', line)
```
dannyroberts: what happened here? Seems like this entire section just got indented an extra 4 spaces with no other change?

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -300,33 +300,34 @@ def get_inventory_from_file(environment):
...
+                        current_group_name = section_name[:-len(':children')]
+                        current_group = Group(name=current_group_name)
+                        inventory.all_groups[current_group_name] = current_group
+                    else:
+                        state = 'parsing-host'
+                        current_host_name = section_name
                 else:
-                    raise ValueError('Encountered items outside a section')
+                    if state == 'parsing-host':
+                        if 'ansible_host' in line:
```
dannyroberts: I think this is the other change. Wasn't this a change from the PR we decided not to make?

pr33thi: I just reverted 2 commits which I accidentally pushed on this branch:
https://github.com/dimagi/commcare-cloud/pull/1879/commits/ae91419f29e9994ba480baefa47975ef87c6ea06
https://github.com/dimagi/commcare-cloud/pull/1879/commits/9d186dfab6cfb6c3241c8e0cfc6eef798a68d82a

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
pr33thi: I just reverted 2 commits which I accidentally pushed on this branch:
https://github.com/dimagi/commcare-cloud/pull/1879/commits/ae91419f29e9994ba480baefa47975ef87c6ea06
https://github.com/dimagi/commcare-cloud/pull/1879/commits/9d186dfab6cfb6c3241c8e0cfc6eef798a68d82a

PR #1881, opened 2018-06-28
Add docs for troubleshooting couchdb2

dannyroberts:


docs/basics/troubleshooting.md
```diff
@@ -75,3 +75,114 @@ or, you can use the output from the status command above and run it through the
...
+|  | ⇩ |  |  |
+| CouchDB Node | `[couchdb2]` | `couchdb2` | 15984 | `couchdb2_port` |
+
+
+#### How to confirm the issue
+
+To confirm the issue, that django processes cannot reach CouchDB, run
+
+```bash
+commcare-cloud <env> django-manage check_services
```
snopoke: could add `couch` as a param to only check couchdb which is much faster than checking everything

snopoke: Side note - we should standardize these names (couch, couchdb, postgres, postgresql)

docs/basics/troubleshooting.md
```diff
@@ -75,3 +75,114 @@ or, you can use the output from the status command above and run it through the
...
+all services will be running on the same machine.
+
+
+Requests to CouchDB are made over HTTP,
+and are normally routed the following way:
+1. They start at the originator of the request,
+   such as a Django web worker
+2. They are made to port 25984 on host `[couchdb_proxy]`,
+   which is served by the `nginx` web server, acting as a load balancer.
+3. `nginx` passed them through to one of the `couchdb2` nodes
```
pr33thi: nit: passed --> passes

docs/basics/troubleshooting.md
dannyroberts: Thanks will. Yeah, I was noticing that too :/

PR #1886, opened 2018-07-02
on == True in yaml

emord:
Some days I think about how YAML is way nicer than JSON to use, and then this kind of thing happens and I have no idea why anyone would want this.

Also @snopoke it looks like deploy_proxy wants to change the app cache's owner:

```
 {
-    "owner": 33,
+    "owner": 1022,
     "path": "/home/cchq/www/icds/app_downloads/cache"
 }

```

I know you just worked on that. any ideas?

I hotfixed this on ICDS


snopoke:
Thanks for this. Re the ownership change I think nginx changes the ownership to `www-data`. We should probably just set it to that in the first place.

PR #1894, opened 2018-07-03
Add aws staging machines, commented out

dannyroberts:


environments/staging/inventory.ini
```diff
@@ -1,27 +1,49 @@
...
+# hqproxy1
 
 [hqdjango0]
 hqdjango0-staging.internal-va.commcarehq.org swap_size=1G
 
 [hqdjango1]
 hqdjango1-staging.internal-va.commcarehq.org swap_size=1G
 
+[hqdjango2]
+10.200.10.178 swap_size=1G hostname=hqdjango2-staging
```
snopoke: Should we keep the host names the same?

dannyroberts: That's a fair question. I chose to add `-staging` simply because for the rackspace VMs with DNS entries, my understanding is that it takes everything up to the first `.` and uses that as the hostname, so those would have `-staging`.

I also considered using a different naming convention that highlights the different location, and starting over with the numbering, but nothing clicked, and I wasn't sure if `awsdjango0` (`awsdjango0-staging`?) was better enough to merit change, and so without having a strong new convention I greatly preferred, I opted to stick to the current naming system as meticulously as possible.

dannyroberts: Actually, let me form that as a question. Do you think a new convention setting the infrastructure apart would be useful?

Things I could imagine wanting to encode here are: environment, hosting provider, and of course role and index. Altogether that might be `aws-django0-staging`. (Would production then be `aws-django0-production`? I think since it's no longer a thing we're intending to type out much, it might be worth being that explicit.) Maybe we don't want `aws` in there at all, since it'll soon be irrelevant. Open to your thoughts, both whether it's worth it at all, and then if so what convention is better.

dannyroberts:
@snopoke bump if you want to entertain the naming convention discussion, otherwise happy to merge as-is.

dannyroberts:
Okay, thanks. I went with `web0`, `rabbit0`, etc. naming convention for aws hosts.

PR #1898, opened 2018-07-05
Ns/security updates

NitigyaS:
Used this ansible role https://github.com/weareinteractive/ansible-apt. 
by default, it adds only security updates. 
I also want suggestions on adding the packages we do not want to upgrade to on hold.


src/commcare_cloud/ansible/setup_security_updates.yml
```diff
@@ -0,0 +1,18 @@
...
+    - postgresql-9.6
+    - postgresql-9.6-plproxy
+    - postgresql-client-9.6
+    - postgresql-client-9.6
+    - postgresql-contrib-9.6
+    - postgresql-server-dev-9.6
+    - rabbitmq-server
+    - riak
+    - riak-cs
+    - zookeeper
```
snopoke: I don't think we need to blacklist the postgres clients. Otherwise this looks like a good list.

src/commcare_cloud/ansible/setup_security_updates.yml
```diff
@@ -0,0 +1,18 @@
+---
+- hosts: all
+  become: yes
+  roles:
+   - security_upgrades
```
snopoke: You need to add `- src: weareinteractive.apt` to requirements.yml. Also this should be `weareinteractive.apt` I think

NitigyaS:
@snopoke Do you think that adding default variable in our playbook would be a good idea to increase the readability of the play. The role I have used uses multiple ansible default variable which changes the configuration files of `unattended-packages` in ways that are not clear by just reading `setup_security_updates.yml`

snopoke:
@NitigyaS do you mean adding the variables even if they are the same as the defaults? Could do I guess but only the relevant ones.

NitigyaS:
yes, not all only those which changes the default configuration files on `unattended-packages`

dannyroberts:
What's the rollout plan for this and are there any current blockers?

NitigyaS:
It will be rolled out on staging this week. No blockers left.

PR #1902, opened 2018-07-09
postgresql dbs usage refactor

dannyroberts:
Followup to https://github.com/dimagi/commcare-cloud/pull/1802#issuecomment-393881302


dannyroberts:
Ok, did a --check run on staging and prod, looking good

src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
```diff
@@ -261,10 +261,9 @@
     ansible_ssh_pipelining: true
   postgresql_ext:
     name: plproxy
-    db: "{{item.name}}"
+    db: "{{postgresql_dbs.form_processing.proxy.name}}"
```
snopoke: Will these still work on environments like `swiss` where `postgresql_dbs.form_processing` is `null`?

dannyroberts: Hmm, now that I'm looking deeper, I'm not sure what the original intention here was. It seems to me the important thing isn't whether it's a monolith, but whether `SEPARATE_FORM_PROCESSING_DBS` is `true`, but then also if `SEPARATE_FORM_PROCESSING_DBS` is `false` then you don't need to install plproxy extention at all. The conclusion I'm coming to is that the right thing to do is to just remove `or is_monolith|bool` from the `when` condition. Does that seem right, @snopoke?

snopoke: Yes that seems right to me.

snopoke: @dannyroberts I think my original question still stands

dannyroberts: So I think yes, after my last fixes, this should now work on swiss. I'll outline my reasoning so you can highlight if you think something's missing.

I changed the `when` condition here to
```
when: postgresql_dbs.form_processing and postgresql_dbs.form_processing.proxy.host == inventory_hostname
```

On environments like swiss, where `postgresql_dbs.form_processing` is `null`, this task will get skipped. Since [form_processing should be None if and only if SEPARATE_FORM_PROCESSING_DBS is False](https://github.com/dimagi/commcare-cloud/blob/081803513b313f70b4562cc0f38b6badcab53be1/src/commcare_cloud/environment/schemas/postgresql.py#L145), that's the behavior we want: we don't want to install plproxy on envs that aren't using partitioned form processing.

Sound right?

snopoke: Yes that sound right and sorry for labouring the issue here but I still just want to check that accessing `postgresql_dbs.form_processing.proxy.host` won't raise an error if `postgresql_dbs.form_processing` is null. Will Jinja swallow the `AttributeError`?

dannyroberts: If it got to this line and saw `postgresql_dbs.form_processing.proxy.host`, and `postgresql_dbs.form_processing` was `null`, then you're right that it would fail—ansible doesn't swallow attribute errors. The reason I'm not concerned about it is that two lines below we have `when: postgresql_dbs.form_processing and ...` so it'll never get here if `postgresql_dbs.form_processing` is `null`. I confirmed that this runs fine on swiss, so I think my assumptions about `when` causing the evaluation in this line to be skipped are correct. Does that help? Feel free to offline with me if I'm missing your point.

snopoke: Thanks Danny, we're all good. I discovered that the version I was looking at still had the when condition as `postgresql_dbs.form_processing.proxy.host == inventory_hostname` without the `postgresql_dbs.form_processing and`  part.

It seems when you click on a file link form the PR overview page it takes you to the version where the comment was first made.

PR #1904, opened 2018-07-09
Fix postgres auth error

pr33thi:
I was getting a postgres auth error when the `MIGRATE DB` task was getting run. 2 things caused it:
1. The default password was stored in `secrets.POSTGRES_USERS.commcare.username`, which wasn't getting populated because many settings yml files weren't organized like this. I reorganized the settings files so that these variables exist.
2. There was a username and password defined in `postgres.yml`, which was overriding the username/pw defined in `private.yml`. I deleted the username and password in `postgres.yml` because they shouldn't be defined there anyway, and probably were just left there as an artifact of some sort.


dannyroberts:
Awesome, thanks for all the digging, @pr33thi! I think I'm going to go ahead and see if I can safely get rid of the `DEFAULT_POSTGRESQL_USER` etc vars. No reason to keep around variables that you should never set!

PR #1905, opened 2018-07-09
Remove DEFAULT_POSTGRESQL_{USER,PASSWORD}

dannyroberts:


.travis/environments/travis/private.yml
```diff
@@ -34,4 +34,3 @@ localsettings_private:
   SECRET_KEY: "JHX1y7iG0bQaVICcBXrQYw1fcJcomC24p6lxAfKMG0B5g7D5"
   TOUCHFORMS_API_PASSWORD: 'TsyasEivKf4GtbQX'
   TOUCHFORMS_API_USER: 'touchforms@example.com'
-
```
pr33thi: (Minor) Needs a newline at end of file

src/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -53,15 +53,16 @@ def validate_shards(shard_ranges_by_partition_name):
     assert _is_power_of_2(num_shards), \
         'Total number of shards must be a power of 2: {}'.format(num_shards)
 
+DEFAULT_POSTGRESQL_USER = "{{ secrets.POSTGRES_USERS.commcare.username }}"
```
pr33thi: How come you moved this here?

snopoke: Just to make it a constant

.travis/environments/travis/private.yml
dannyroberts: I think this previously ended with _two_ newlines and I deleted one of them. (Github will put a "No newline at end of file" message if there's no newline.)

src/commcare_cloud/environment/schemas/postgresql.py
dannyroberts: Yeah the idea was to make it not part of the schema anymore so that people can't declare it in `postgresql.yml`. Pulled out here, it's just a constant that I can use later in the code.

PR #1908, opened 2018-07-10
Added documentation for letsencrypt

NitigyaS:


docs/howto/automate_ssl.md
```diff
@@ -0,0 +1,47 @@
...
+  become: yes
+  file:
+    src: "/etc/letsencrypt/live/{{ENV_SITE_HOT}}/privkey.pem"
+    dest: "{{ ssl_keys_dir }}/{{ env_nginx_ssl_key  }}"
+    state: link
+
+  when: not fake_ssl_cert and ENV_SITE_HOST | default(None)
+
+
+```
```
dannyroberts: What is this block of code? What is the reader supposed to do when they get to this point?

docs/howto/automate_ssl.md
```diff
@@ -0,0 +1,57 @@
...
+---
+
+## How to setup.
+To setup the automate renewal we can use commcare-cloud command after making following changes.
+* Add `letsencrypt_cchq_ssl: True` variable in `proxy.yml`
+
+* Add configuration to create symlink to `privkey.pem` and `fullchain.pem` in `src/commcare_cloud/ansible/roles/nginx/tasks/install.yml`
+
+Following variables are predefined
+* ENV_SITE_HOT: Site for which certificate is renewed
```
dannyroberts: ENV_SITE_HOT => ENV_SITE_HOST

docs/howto/automate_ssl.md
```diff
@@ -0,0 +1,57 @@
...
+### Monitoring
+The expiry of certificates is monitored via external monitoring tools.
+
+---
+
+## How to setup.
+To setup the automate renewal we can use commcare-cloud command after making following changes.
+* Add `letsencrypt_cchq_ssl: True` variable in `proxy.yml`
+
+* Add configuration to create symlink to `privkey.pem` and `fullchain.pem` in `src/commcare_cloud/ansible/roles/nginx/tasks/install.yml`
```
dannyroberts: "Add configuration" => "Add the following configuration"

docs/howto/automate_ssl.md
dannyroberts: @NitigyaS I think I'm still wondering about this.

Are you suggesting that the user copy and paste this block into their local `src/commcare_cloud/ansible/roles/nginx/tasks/install.yml` and then run `deploy_proxy.yml`? I hadn't originally interpreted that way, but I think that makes sense now.

I should also say, if I stumbled upon these instructions, I would think that that's a very strange thing for me to have to do! Is there a reason we can't just put these in the playbook to begin with?

This doesn't have to be a blocker—I'm pro documenting weird things we have to do rather than not documenting just because they're weird. If it's too big a change, we can leave it like this and take it on later on.

snopoke: I think this only applies if you're creating a new nginx site which for the most part wouldn't be necessary. I originally thought this doc was for 3rd parties who wanted to configure lets encrypt but now it feels more like its an 'internal' doc that documents how letsencrypt is configured.

NitigyaS: >Are you suggesting that the user copy and paste this block into their local.

I do agree that adding this block to playbook is something that we should avoid. problem with this is that our playbook `src/commcare_cloud/ansible/roles/nginx/tasks/install.yml` was designed in this way from the beginning, which includes a site-specific block for each server. I would like to replace that with a single task and that shouldn't be too time taking. I will keep it in the backlog. 

> this only applies if you're creating a new nginx site

This block addition is required for those cases in which user is making a switch from Third Party SSL to Letsencrypt SSL. If we are configuring up a new site the user will be adding this block anyway.

PR #1911, opened 2018-07-11
update UCR database settings

snopoke:
Pulled from https://github.com/dimagi/commcare-cloud/pull/1707

These seem pretty safe to start with.

buddy @jmtroth0

### DIFF
--- a/environments/icds-new/inventory.ini
+++ b/environments/icds-new/inventory.ini
@@ -66,14 +66,15 @@ pgucrstandby1
 pgbouncer_max_connections=3500
 pgbouncer_default_pool=775
 postgresql_max_connections=800
-postgresql_work_mem=32MB
-postgresql_shared_buffers=8GB
-postgresql_effective_cache_size=24GB
+postgresql_work_mem=3495kB
+postgresql_shared_buffers=16GB
+postgresql_effective_cache_size=48GB
 postgresql_checkpoint_segments=128
 postgresql_checkpoint_completion_target=0.9
 postgresql_maintenance_work_mem = 2GB
 postgresql_wal_buffers=16MB
 postgresql_default_statistics_target = 500
+postgresql_max_wal_size = 1GB
 
 [pgmain1]
 10.247.164.85


snopoke:
@emord I think we can go ahead. Let's try without downtime.

emord:
I'll try this either tonight after agg script, or tomorrow in that case

snopoke:
👍

emord:
@snopoke @dannyroberts @NitigyaS Not going to deploy this today. There's a combination of this PR not actually changing these values and unapplied changes to postgres. I believe the unapplied changes are ok to deploy, but would like you guys to verify/resolve that this is expected. Also I haven't tried this deploy on anything other than the UCR databases, so there's probably more discrepancies

A. unapplied changes to postgres (truncated):

```
TASK [postgresql : Make new postgresql directory] ******************************
--- before
+++ after
@@ -1,4 +1,4 @@
 {
-    "mode": "0700",
+    "mode": "0755",
     "path": "/opt/data/postgresql"
 }

changed: [10.247.164.25]
TASK [postgresql : Create wal archive directory] *******************************
--- before
+++ after
@@ -1,4 +1,4 @@
 {
     "path": "/opt/data/postgresql/wal_archive",
-    "state": "absent"
+    "state": "directory"
 }

changed: [10.247.164.25]

(removal of several lines in pg_hba.conf)
```

B. This only changes work_mem and max_wal_size. I think that's because there are overrides in the environment level: https://github.com/dimagi/commcare-cloud/blob/sk/icds-pg-tune1/environments/icds-new/postgresql.yml#L23 Changing this is going to mean changing all postgres instances becuase postgresql_effective_cache_size is set for all of the servers in the inventory

C. Failure

```
fatal: [10.247.164.25]: FAILED! => {"changed": false, "msg": "pgbouncer process not presently configured with monit", "name": "pgbouncer", "state": "monitored"}
...ignoring
```

snopoke:
@emord updated to fix the issues you noticed

src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
```diff
@@ -71,15 +71,6 @@
 - name: assert replication slot config correct
   assert: { that: "replication_slot is not defined or replication_slot in hostvars[hot_standby_master]['postgresql_replication_slots']"}
 
-- name: Create wal archive directory
```
emord: should this be actively removed from servers where it still exists?

src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
```diff
@@ -43,7 +43,7 @@
     path: '{{ postgresql_dir_path }}'
     owner: postgres
     group: postgres
-    mode: 0755
```
emord: I think this will need to be deployed everywhere?

environments/icds-new/postgresql.yml
```diff
@@ -24,9 +24,7 @@ override:
   pgbouncer_pool_mode: transaction
   pgbouncer_pool_timeout: 1
   pgbouncer_reserve_pool: 5
-  postgresql_effective_cache_size: 16GB
```
emord: Should we plan time/downtime to deploy to all postgres servers. I believe that this changes all servers configurations

src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
snopoke: I suppose so. I'll update to remove it.

src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
snopoke: This will already be the case everywhere since PG won't start if the permissions are wrong.

environments/icds-new/postgresql.yml
snopoke: hmm, yes I see it will so let's schedule some downtime.

src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
snopoke: d970298

src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
emord: I don't believe that's true as the standbys currently have `755`:

```
TASK [postgresql : Make new postgresql directory] ******************************
ok: [10.247.164.25]
--- before
+++ after
@@ -1,4 +1,4 @@
 {
-    "mode": "0755",
+    "mode": "0700",
     "path": "/opt/data/postgresql"
 }

changed: [10.247.164.84]
--- before
+++ after
@@ -1,4 +1,4 @@
 {
-    "mode": "0755",
+    "mode": "0700",
     "path": "/opt/data/postgresql"
 }

changed: [10.247.164.24]
```

snopoke: Interesting, in the past when I've tried starting PG with 0755 I get this:
```
FATAL: data directory "/path/to/cluster" has group or world access
DETAIL: Permissions should be u=rwx (0700).
```

Either way I don't think there's harm in changing it to `0700`

environments/icds-new/inventory.ini
```diff
@@ -66,14 +66,15 @@ pgucrstandby1
...
-postgresql_effective_cache_size=24GB
+postgresql_work_mem=3495kB
+postgresql_shared_buffers=16GB
+postgresql_effective_cache_size=48GB
 postgresql_checkpoint_segments=128
 postgresql_checkpoint_completion_target=0.9
 postgresql_maintenance_work_mem = 2GB
 postgresql_wal_buffers=16MB
 postgresql_default_statistics_target = 500
+postgresql_max_wal_size = 1GB
```
emord: may want to consider upping this again. still getting

```
checkpoints are occurring too frequently (27 seconds apart)",,"Consider increasing the configuration parameter ""max_wal_size"".",,,,,,,""
```

up from 4 seconds

snopoke: Turns out this is the default. I'll bump it to 5GB.

snopoke:
Any reason not to merge this now?

environments/icds-new/inventory.ini
snopoke: https://github.com/dimagi/commcare-cloud/pull/2029/files#diff-0688e9dafa291b12fa97b468ab69dfa1R77

calellowitz: Which machine was that warning on? I think it is fine to increase it, but I think based on the volume of data we load nightly into the UCR machine, we will likely see it no matter how large the size is. That was my takeaway when I looked into this for the warehouse machine. Although, they also suggested I set it to the full ram of the machine, so I would not worry about this configuration, but maybe also not worry about those messages.

snopoke: I'm not too concerned about the warning itself, more about the performance implications of it.

PR #1915, opened 2018-07-12
Run travis off pinned versions

dannyroberts:
This will let us loosen up pinned versions in setup.py, while still working off of pinned versions.

As new libraries get updates, we can run

```
pip-compile --upgrade
```
to re-collect all libraries and create a new requirements.txt file with libraries pinned to the latest matching version. We can then PR that change to see if it breaks travis, and (until we trust our tests more) do some manual testing. While upgrading libraries is still a pain, this at least gives us a sane way to do it when we have to, whereas pinning versions in setup.py tends to cause version conflicts as other libraries update their required versions.


pr33thi:
Cool! What's the workflow for when one would run this? And what's the strategy for if we intentionally want to use a version of some package besides the latest?

snopoke:
What's the benefit of doing this? Couldn't this lead to the situation where a library gets an update the breaks something but we don't know about it because we've pinned the versions in tests?

dannyroberts:
@snopoke I think the idea is

1. to keep version _ranges_ in `setup.py` when they're needed. So, we can pretty liberally put min verions, and if we find that a new update breaks something, we add a max version in `setup.py` as well.
2. to recompile requirements regularly and PR requirements updates separately.

This lets us stay up to date on libraries without creating the version conflicts that often come with pinned versions in `setup.py`, and also without having version issues cause random test failures in unrelated PRs. Instead, by regularly recompiling the requirements file and PRing, we can uncover those failures separately.

This change was motivated by an actual version conflict I was experiencing, as well as a desire to follow best practice in python versioning. This process is something I saw articulated as a best-practice for versioning dependencies of a python library at the last Django Boston meetup.

dannyroberts:
> Cool! What's the workflow for when one would run this? And what's the strategy for if we intentionally want to use a version of some package besides the latest?

In this case we can still pin it, or preferably give it a range.

PR #1917, opened 2018-07-12
Add requirements for fixing <2.7.9 ssl issue

dannyroberts:
@pr33thi here's my plan for keeping the fix for the python ssl issue in code, once https://github.com/dimagi/commcare-cloud/pull/1915 is merged.


requirements.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile --output-file requirements.txt setup.py
+#    pip-compile --output-file requirements.txt setup.py requirements-python-lt-2.7.9-ssl-issue.in
```
pr33thi: what's this doing?

dannyroberts: This file is auto-generated by running `pip-compile --output-file requirements.txt setup.py requirements-python-lt-2.7.9-ssl-issue.in`, and it includes the command that generated in this comment so you know how to regenerate it. Does that answer your question?

pr33thi: Specifically I meant what is the addition of `requirements-python-lt-2.7.9-ssl-issue.in` doing?

dannyroberts: Ah, okay. So I'll interpret your question in two different ways and try to answer each. If neither's the way you meant it, we can try another round :)

_What is `pip-compile`'s argument structure, and how is the addition of this last argument (`requirements-python-lt-2.7.9-ssl-issue.in`) interpreted?_ Here's `pip-compile`'s help output:
```
$ pip-compile --help
Usage: pip-compile [OPTIONS] [SRC_FILES]...
[...]
  -o, --output-file TEXT          Output file name. Required if more than one
                                  input file is given. Will be derived from
                                  input file otherwise.
[...]
```
So here `--output-file requirements.txt` is one of the `OPTIONS` just saying where to put the output (this file). The other two arguments (`setup.py` and `requirements-python-lt-2.7.9-ssl-issue.in`) are `SRC_FILES`. `pip-compile` allows source files of a few different formats. 
- One of them is the `setup.py` that is the standard for python packaging (putting things on pypi, etc.), where it reads dependencies from the [`install_requires`](https://github.com/dimagi/commcare-cloud/blob/dda476fdddcd15305baaf61b16648ae5f336392b/setup.py#L27-L49) argument of the base `setup(...)` call.
- Another is a "`.in`" file which lists requirements in the `requirements.txt` format, i.e. one requirement per line, where each line could be appended to `pip install` to install that dependency, which is the format that `pip install -r <file>` accepts as well. It's called an '`.in`' file because unlike `requirements.txt`, it doesn't contain the entire dependency tree with versions pinned; instead it contains just the direct dependencies, with versions unpinned, or restricted to a version range if deemed necessary.

_Why is `requirements-python-lt-2.7.9-ssl-issue.in` in a separate file instead of just adding those as dependencies to `setup.py`?_ I wanted to make it clear that these dependencies were being added to fix this bug, which only affects some environments, and not because anything in the code calls these libraries directly. Although I'm not sure if there's some standard convention for doing this (my guess knowing how fragmented python versioning & packaging tools are is that there isn't), I chose this as hopefully the most approachable and explicit way of separating it out.

In the end, `pip-compile` takes the list of requirements from both commcare-cloud's `setup.py` file and the `requirements-python-lt-2.7.9-ssl-issue.in` file—each of which has (or will have) loosely defined version ranges rather than fully pinned versions—reconciles them, and then dumps a pip-freeze of the result to requirements.txt, in which every requirement is pinned to a specific version and every requirement in the dependency tree (not just the direct dependencies) is included.

pr33thi: That makes sense, thanks for the thorough answer!

PR #1919, opened 2018-07-13
Remove duplicate repeat_record_queue from swiss

emord:
Looks like we've been running an extra queue since february https://github.com/dimagi/commcare-cloud/commit/ef6c0eaa4bcbc9926013120e3d30c656dfb47614#diff-02338665690f88220a8d96f9c210ea2d

@dannyroberts fyi, not sure if we want to add something that checks running the same queue twice on one server


dannyroberts:
Oops, thanks. Great idea: https://github.com/dimagi/commcare-cloud/issues/1921

PR #1923, opened 2018-07-16
Move some pillows around

emord:
Doing this to get the queues down. will revert it back before EOD


src/commcare_cloud/ansible/deploy_commcarehq.yml
```diff
@@ -46,6 +46,12 @@
     - include_tasks: roles/commcarehq/tasks/celery.yml
       tags: services
 
+- name: Pillowtop Supervisor Config
+  hosts: pillowtop
+  tasks:
+    - include_tasks: roles/commcarehq/tasks/pillowtop.yml
+      tags: services
```
emord: Looks like I accidentally removed this in a merge commit.

PR #1925, opened 2018-07-16
Remove pycrypto dependency

dannyroberts:
added here https://github.com/dimagi/commcare-cloud/commit/68aca9bd04ee4fc28774c449e4d784b19a25613e#r29726053
for reasons not obvious or documented and that I no longer remember

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,6 +36,7 @@ paramiko==2.4.1           # via ansible, fabric
 passlib==1.7.1
 pyasn1==0.4.3             # via ndg-httpsclient, paramiko
 pycparser==2.18           # via cffi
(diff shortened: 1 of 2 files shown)


dannyroberts:
I couldn't find where pycrypto is needed, but if it _is_ then [apparently](https://github.com/dlitz/pycrypto/issues/253#issuecomment-388683806) the library to use nowadays is [pycryptodome](https://github.com/Legrandin/pycryptodome) in drop-in replacement mode, so if we find we need it we should replace with that.

dannyroberts:
Okay, turns out we need it. Added it in the last commit; hopefully now it's better documented why we're including it.

PR #1907, opened 2018-07-10
Added documentation for encrypted tmp directory.

NitigyaS:

PR #1929, opened 2018-07-17
Add documentation for #1791

emord:
Tried to give roughly the same amount of info as https://github.com/dimagi/commcare-cloud/pull/1907

### DIFF
--- a/docs/changelog/0002-supervisor-service-definitions.md
+++ b/docs/changelog/0002-supervisor-service-definitions.md
@@ -14,6 +14,9 @@ configuration files. This change decouples the process definitions from code.
 All services are now defined separately from a code deploy and outside of our
 the directory where code runs.
 
+Important Note: Any code deploys before running the following command will cause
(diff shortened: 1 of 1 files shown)


snopoke:
Thanks.  One quick question, would it be a problem if code deploy continued to happen without running this command?

emord:
If you code deploy from master before running this command then celery workers & formplayer would be down. (other services should have already been moved) so most things should continue working, but the `check_services` would indicate it being down.

snopoke:
I think that's something that's important to highlight in the changelog - I'd tried to include that with the 'dependant commcare version'  but in this case that won't help. Maybe we should add another field like `Affects CommCare deploy` or something?

emord:
I added a note about this in the details. It wasn't clear how prominent to make it. If this were a versioned utilitity I would say we could change the major version number, but not clear to me how to make a note of it when it's always expected to run from master

PR #1931, opened 2018-07-17
better formatting

sravfeyn:
This was a todo from eCHIS training. 

@dannyroberts 

Example before/after 
Before http://dimagi.github.io/commcare-cloud/commcare-cloud/commands/#run-module
After
screen shot 2018-07-17 at 1 00 15 pm

### DIFF
--- a/docs/commcare-cloud/commands/index.md
+++ b/docs/commcare-cloud/commands/index.md
@@ -14,16 +14,14 @@ commcare-cloud [--control]
                ...
 ```
 
-## Positional Arguments
+*Positional Arguments*
 
-### `<env>`
-
+- `<env>`
 server environment to run against
 
-## Optional Arguments
+*Optional Arguments*
 
-### `--control`
-
+- `--control`
 Run command remotely on the control machine.
 
 You can add `--control` _directly after_ `commcare-cloud` to any command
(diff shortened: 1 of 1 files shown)


dannyroberts:
Nice, @sravfeyn. I like the new look. I think something you missed is that these docs are auto-generated, so that they stay up-to-date with changes to commands and the help text we add to them.

The downside is that it's somewhat tricky to change formatting: it's possible, it'll just take staring at [this file](https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/manage_commcare_cloud/make_docs.py) for a long time. To run the autogeneration, just run `make`.

So, I suggest you start from here, where you've already made the formatting changes you want, and then run `make`. You'll see a big diff on this file, since it'll have been autogenerated back to what it was. Then start playing around with `make_docs.py`. If you want, you can run something like `watch -n 5 make` to rebuild every 5 seconds (if there are changes). As you make changes, the goal is to slowly make the diff smaller and smaller, until your modified `make_docs.py` generates exactly the file that you submitted here. You may find that in the process that there are things that are too tricky to get right, and you may even find some mistakes with your originally reformatting (as I did when I first did this).

Good luck, and let me know if you want to sync about this!

PR #1940, opened 2018-07-18
copy files command

snopoke:
This is mostly extracting functionality out of the couchdb migration command so that I can be used on it's own.

This should be useful for the upcoming ICDS migration where we'll likely need to copy data to all the DB machines from a 'backup' location.

Can be reviews as a whole or by commit


dannyroberts:
Also, having flashbacks of copy-pasting dozens of rsync commands from the last ICDS migration 😵 This is going to be a killer tool to have, in a good way.

PR #1942, opened 2018-07-20
Upgrade pip on fab deploy

dannyroberts:
On current deploy I had problems with the pip version being too old. I think just always bringing it to the latest version is a reasonable thing to do.


emord:
Wouldn't it work to just increase https://github.com/dimagi/commcare-hq/blob/master/requirements/requirements.in#L67 ?

PR #1946, opened 2018-07-20
Use supervisor 3.3.4 instead of 3.0b

dannyroberts:
which on ubuntu 14.04 requires installing from pip

Curious what you all think about this. We've been on 3.0b forever, but it seems like with the occasional issues we have with supervisor not properly killing this, it'd be nice to rule out just being on an old version [May 2013](https://pypi.org/project/supervisor/3.0b2/).

The immediate ticket that prompted this was https://manage.dimagi.com/default.asp?276826. Not sure if we've seen that before (celerybeat process not getting killed on restart, resulting two celerybeat processes, resulting in duplicate period tasks), but it seems to be happening regularly now.

### DIFF
--- a/src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
@@ -1,5 +1,9 @@
 ---
 
+- name: remove supervisor 3.0b
+  apt: name=supervisor state=absent
+  become: yes
+
 - name: find supervisord path
   shell: 'which supervisord'
   register: which_supervisord
@@ -7,10 +11,19 @@
   failed_when: which_supervisord.rc != 0 and which_supervisord.rc != 1
   check_mode: no
 
-- name: install supervisor
-  apt: name=supervisor state=present
+- name: install supervisor 3.3.4
+  pip: name=supervisor state=present version=3.3.4
   become: yes
   when: not which_supervisord.stdout
+
+- name: link supervisor
+  file:
+    path: /usr/bin/{{ item }}
+    state: link
+    src: /usr/local/bin/{{ item }}
+  with_items:
+    - supervisord
+    - supervisorctl
 
 - name: retry finding supervisord path
   shell: 'which supervisord'


emord:
What's the test plan for this? It does seem like we've had more issues with supervisor recently, so I'm pro updating.

dannyroberts:
I ran this on staging Friday, so I was going to see if any issues came up there. Once it's undergone the following things without any problems:

- [ ] a normal deploy
- [ ] start and stop services
- [ ] restart supervisord

I'm pretty comfortable updating it. (It helps that it's the same major version.)

dannyroberts:
No new issues that I've seen during my testing so I'm going to merge it. (I say "new" because I saw some of the same problems we're seeing where if a machine's already low on memory then restarting processes doesn't properly kill the old ones.)

src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
```diff
@@ -1,17 +1,30 @@
...
   shell: 'which supervisord'
   register: which_supervisord
   changed_when: false
   failed_when: which_supervisord.rc != 0 and which_supervisord.rc != 1
   check_mode: no
 
-- name: install supervisor
-  apt: name=supervisor state=present
+- name: install supervisor 3.3.4
+  pip: name=supervisor state=present version=3.3.4
```
emord: @dannyroberts it looks like the staging machines are consistently missing meld3 which is a dependency of this library. I checked swiss (deployed today) and it does exist there. Do you recall any issues that you had when deploying this?

emord: FWIW the output of running this command is:

```
Requirement already satisfied (use --upgrade to upgrade): supervisor==3.3.4 in /usr/local/lib/python2.7/dist-packages
Downloading/unpacking meld3>=0.6.5 (from supervisor==3.3.4)
  Downloading meld3-1.0.2-py2.py3-none-any.whl
Installing collected packages: meld3
Successfully installed meld3
```

meaning the correct version of supervisor was installed, but without the meld3 dependency

PR #1948, opened 2018-07-23
Ns/redis prod

NitigyaS:
Because Rackspace does not allow to change the server's class and we are already using the biggest server in _General Purpose class_. 
I have created a new server with configuration `2 core and 15 GB RAM` in memory-optimized class.


environments/production/public.yml
```diff
@@ -11,6 +11,7 @@ elasticsearch_cluster_name: 'prodhqes-1.x'
 #elasticsearch_memory: '4096m'
 #elasticsearch_node_name: '???'
 
+redis_maxmemory: 4gb
```
snopoke: shouldn't this be 8?

PR #1950, opened 2018-07-23
Add list-dbs command

dannyroberts:
G2 kept asking to verify which dbs on which machines should be copied, and I noticed we didn't really have a source of truth that showed the information in a decisive format. (`postgresql.yml` for instance doesn't show the actual names of the dbs, which are inferred; `.generated.yml` is an embarrassing place to ask someone to look, and kind of a mess to pull this out from.)


snopoke:
Need to update docs

dannyroberts:
Thanks Simon. Taste of my own medicine.

PR #1951, opened 2018-07-24
Reduce workers and concurrency for softlayer celery

mkangia:
Celery machine of softlayer is running full on memory.
Since there is not much activity on softlayer it should be fine to have less processes running since we wont need high number of tasks to be run in parallel.
i think its mainly the "number of workers" which is going to reduce the load by reducing the number of processes running.
Reduced concurrency at places to avoid memory spikes, just in case they are multiple tasks required of a worker.

### DIFF
--- a/environments/softlayer/app-processes.yml
+++ b/environments/softlayer/app-processes.yml
@@ -26,10 +26,10 @@ celery_processes:
       optimize: True
     reminder_case_update_queue:
       pooling: gevent
-      concurrency: 1
+      concurrency: 5
     reminder_queue:
       pooling: gevent
-      concurrency: 2
+      concurrency: 5
     reminder_rule_queue:
       concurrency: 1
       max_tasks_per_child: 1
--- a/environments/softlayer/app-processes.yml
+++ b/environments/softlayer/app-processes.yml
@@ -18,7 +18,7 @@ celery_processes:
     ucr_indicator_queue:
       concurrency: 1
     celery,export_download_queue:
-      concurrency: 4
+      concurrency: 2
       max_tasks_per_child: 5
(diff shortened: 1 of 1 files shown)


environments/softlayer/app-processes.yml
```diff
@@ -18,28 +18,26 @@ celery_processes:
...
       max_tasks_per_child: 5
     saved_exports_queue:
       concurrency: 2
       max_tasks_per_child: 1
       optimize: True
     reminder_case_update_queue:
       pooling: gevent
-      concurrency: 5
-      num_workers: 2
+      concurrency: 1
```
calellowitz: hey this changes 2 workers (each with concurrency of 5, but via gevent not forked processes) to one worker with concurrency 1 (effective concurrency of 10 -> 1). same thing with the `reminder_queue` that seems a pretty dramatic reduction. I also do not believe there is a large memory overhead in gevent processes (although could be wrong). Might be better start by moving it to one worker with concurrency 5 instead of both slashing the number of workers and their concurrency

mkangia: Cool, Updated.

environments/softlayer/app-processes.yml
```diff
@@ -27,19 +27,17 @@ celery_processes:
...
-      num_workers: 2
     reminder_rule_queue:
       concurrency: 1
       max_tasks_per_child: 1
     sms_queue:
       pooling: gevent
       concurrency: 10
     async_restore_queue:
-      concurrency: 8
+      concurrency: 4
```
snopoke: Can drop this to 1 - don't think there are any projects using this at all

Issue #1928, opened 2018-07-17
Autogenerate changelog index


dannyroberts:
We'd like to automate the generation of [docs/changelog/index.md](https://github.com/dimagi/commcare-cloud/pull/1927/files#diff-8126fa4ec84c1259fc7a6ea789a1ac53), which should contain a link to all the numbered change records in `docs/changelog/`.

### Checklist

Some details to consider
- [ ] skip `docs/changelog/0000-changelog.md`
- [ ] skip `docs/changelog/index.md` itself
- [ ] changes should be listed _newest first_ so 0002 should show up above 0001.

### Testing

To test manually, you should be able to run

```
rm docs/changelog/index.md
make
git diff
```
and see no `git diff` output (for `docs/changelog/index.md` at least).

### Guidance

In order to implement, follow the pattern set by the autogeneration of `docs/commcare-cloud/commands/index.md`.

I'll trace through the thread of figuring out how that file is generated.

#### `Makefile`

First, if you search for `docs/commcare-cloud/commands/index.md` in the codebase, and you'll find it in the `Makefile`.

Cutting out irrelevant lines we have:

```Makefile
all : src/commcare_cloud/help_cache/ansible.txt src/commcare_cloud/help_cache/ansible-playbook.txt docs/commcare-cloud/commands/index.md
[...]
docs/commcare-cloud/commands/index.md : src/commcare_cloud/* src/commcare_cloud/*/* src/commcare_cloud/*/*/*
	manage-commcare-cloud make-docs > docs/commcare-cloud/commands/index.md
[...]
clean:
	rm docs/commcare-cloud/commands/index.md src/commcare_cloud/help_cache/ansible.txt src/commcare_cloud/help_cache/ansible-playbook.txt
```

(Here's a [refresher on `Makefile`s](http://www.cs.colby.edu/maxwell/courses/tutorials/maketutor/), and googling "makefile" should bring up lots of helpful results if that doesn't have the right level of context.)

So the important things here are

1. It's listed as a dependency for `all` at the top, which is what gets run when you run `make` without args.
2. It's passed to the `rm` command under `clean`, which gets run when you run `make clean`.
3. There's a section for building it, listing all the files it depends on, with the command you use to build it given under (after a `\t` TAB character).

#### `manage-commcare-cloud` command

That takes us to `manage-commcare-cloud make-docs`, which is the command in the `Makefile` that's used to build `docs/commcare-cloud/commands/index.md`.
[...]

PR #1954, opened 2018-07-24
Autogenerate changelog

pr33thi:
https://github.com/dimagi/commcare-cloud/issues/1928


src/commcare_cloud/manage_commcare_cloud/make_docs.py
```diff
@@ -194,3 +193,27 @@ def do_print_help(parser):
...
+
+    def make_parser(self):
+        pass
+
+    def run(self, args, unknown_args):
+        j2 = jinja2.Environment(loader=jinja2.FileSystemLoader(os.path.dirname(__file__)), keep_trailing_newline=True)
+
+        def do_print_help(parser):
+            """Capture output of parser.print_help()"""
+            string_io = StringIO()
```
nickpell: I'd recommend using ```io.StringIO``` here.  It will help with eventual py3 compatibility.

src/commcare_cloud/commcare_cloud.py
```diff
@@ -148,9 +148,67 @@ def make_parser(available_envs, formatter_class=RawTextHelpFormatter,
     return parser, subparsers, commands
 
 
+def make_changelog_parser():
```
snopoke: This name is confusing. `make_command_parser` returns an argument parser whereas this function compiles the changelog.

I'd rename this to something like `compile_changelog`

src/commcare_cloud/commcare_cloud.py
```diff
@@ -148,9 +148,67 @@ def make_parser(available_envs, formatter_class=RawTextHelpFormatter,
...
+    # Parse the contents of the changelog dir
+    changelog_contents = []
+    files_to_ignore = ['0000-changelog.md', 'index.md']
+    changelog_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'docs', 'changelog')
+    sorted_files = _sort_files(changelog_dir)
```
snopoke: could pass `files_to_ignore` to `_sort_files` to have them excluded from the start

src/commcare_cloud/commcare_cloud.py
```diff
@@ -148,9 +148,67 @@ def make_parser(available_envs, formatter_class=RawTextHelpFormatter,
...
+    for change_file_name in sorted_files:
+        if change_file_name not in files_to_ignore:
+            try:
+                with open(os.path.join(changelog_dir, change_file_name)) as change_file:
+                    change_context = ''
```
snopoke: Moving this code (that processes the file) into a separate function would make this more readable.

src/commcare_cloud/manage_commcare_cloud/make_docs.py
```diff
@@ -194,3 +193,27 @@ def do_print_help(parser):
...
+            parser.print_help(file=string_io)
+            return string_io.getvalue()
+
+        j2.filters['print_help'] = do_print_help
+
+        changelog_contents = make_changelog_parser()
+
+        template = j2.get_template('changelog.md.j2')
```
snopoke: This file is missing

src/commcare_cloud/manage_commcare_cloud/make_docs.py
```diff
@@ -194,3 +193,27 @@ def do_print_help(parser):
...
+    def run(self, args, unknown_args):
+        j2 = jinja2.Environment(loader=jinja2.FileSystemLoader(os.path.dirname(__file__)), keep_trailing_newline=True)
+
+        def do_print_help(parser):
+            """Capture output of parser.print_help()"""
+            string_io = StringIO()
+            parser.print_help(file=string_io)
+            return string_io.getvalue()
+
+        j2.filters['print_help'] = do_print_help
```
snopoke: I don't think you'll need this filter for this command

docs/commcare-cloud/commands/index.md
```diff
@@ -277,9 +277,9 @@ authenticate using the pem file (or prompt for root password if there is no pem
   --list-hosts          outputs a list of matching hosts; does not execute
                         anything else
   -M MODULE_PATH, --module-path=MODULE_PATH
-                        prepend colon-separated path(s) to module library
-                        (default=[u'~/.ansible/plugins/modules',
-                        u'/usr/share/ansible/plugins/modules'])
+                        prepend colon-separated path(s) to module library (def
+                        ault=[u'~/.ansible/plugins/mo
+                        dules', u'/usr/share/ansible/plugins/modules'])
```
snopoke: how come these changed?

pr33thi: Oops, reverted this.

If you're curious why it happened, my `make` command generated a different linebreak than what was was on master, which danny fixed here: https://github.com/dimagi/commcare-cloud/commit/5b02b4cc8de1737a6fe6183cef06419713e2b9b9

pr33thi:
@snopoke the file that was originally make to test the command docs also tests this doc, I renamed it to reflect this here: https://github.com/dimagi/commcare-cloud/pull/1954/commits/8b48375c07037ecc99ad714987a177fea517933d

Makefile
```diff
@@ -17,8 +17,11 @@ src/commcare_cloud/help_cache/ansible-playbook.txt:
 docs/commcare-cloud/commands/index.md : src/commcare_cloud/* src/commcare_cloud/*/* src/commcare_cloud/*/*/*
 	manage-commcare-cloud make-docs > docs/commcare-cloud/commands/index.md
 
+docs/changelog/index.md : src/commcare_cloud/* src/commcare_cloud/*/* src/commcare_cloud/*/*/*
```
dannyroberts: To the right of the `:` should go the files that this build step depends on. `docs/commcare-cloud/commands/index.md` above depends on or could possibly depend on pretty much everything under `src/commcare-cloud` (thought I probably could have been more specific about it): at least `src/commcare_cloud/commcare_cloud.py`, `src/commcare_cloud/commands/`, and `src/commcare_cloud/manage_commcare_cloud/`.

Here, I think we depend on really just `src/commcare_cloud/manage_commcare_cloud/*` (which is covered by what you put here) but then also, importantly, `docs/changelog/*-*.md`, since these are essentially the input files used in `make-changelog-index`.

Currently if I run `make`, then add another file to `docs/changelog/`, and then run `make` again, nothing is done on the second run, whereas the desired behavior would be for `docs/changelog/index.md` to get rebuilt.

src/commcare_cloud/commcare_cloud.py
```diff
@@ -150,9 +150,70 @@ def make_parser(available_envs, formatter_class=RawTextHelpFormatter,
...
+            this_changelog = {'filename': change_file_name,
+                              'context': change_context,
+                              'date': change_date,
+                              'summary': change_summary,
+                              'action_required': change_action_required}
+
+        changelog_contents.append(this_changelog)
+    except (AttributeError, AssertionError):
+        print("Error parsing the file {}.".format(change_file_name))
+        sys.exit(1)
```
dannyroberts: There are a couple try/except antipatterns going on here as I see it:
- `try`/`except` blocks with a large (many-line) `try` block—ideally you just have a few lines in the `try` block so that it's quite clear what errors you're planning to catch
- Catching a very general error type (`AttributeError`), which could result in catching errors other than the ones you had in mind
- Catching an error and exiting when simply _not_ handling the error would do just as well or better
- Blanket suppression of error messages

That last one is practically the most important I think. The point of the `try`/`except` here as I understand it is basically to catch any type of error that might happen so it doesn't bubble up (i.e. `except Exception` would be the more explicit way to do this). But why don't we want the error to bubble up? Let's say I just added a new file, and I run `make` (or I forget to run `make` and then see this error in travis), and then all I see is `Error parsing the file docs/changelog/foo.md`. What's my next step? I'll probably search the code for that, and then remove the `try`/`except` here and run it again so that I can see the actual error! Why not just remove the try except to begin with?

One answer I can think of is that then the error output might not include the name of the problematic file. The solution to this would be to do a blanket catch (`except Exception`) print out the message like you do here, but then have a bare `raise` statement (to reraise the previous error) rather than exit. That way they see your message _and_ the traceback that'll help them identify the exact problem quicker.

To avoid the long `try` block, I'd suggest removing it here entirely, but then above where this function is called have

```
try:
    _parse_changelog_file(changelog_contents, changelog_dir, change_file_name)
except Exception:
    print("Error parsing the file {}.".format(change_file_name))
    raise
```

docs/changelog/index.md
```diff
@@ -11,6 +11,12 @@ an action on your part will be marked "_action optional_".
...
-### **2018-06-11** [Added encrypted temporary directory](0001-add-encrypted-tmp.md) (_action required_)
+### **2018-07-13** [Changelog](0002-supervisor-service-definitions.md) (_action required_) 
+
+There are several CommCare specific processes that are defined in supervisorconfiguration files. This change decouples the process definitions from code.
+
+### **2018-06-11** [Added encrypted temporary directory](0001-add-encrypted-tmp.md) (_action required_) 
 
 Some of the CommCare processes make use of temporary files to store client data (such as data exports) so in order to keep that data protected we have modified the setup to use an encrypted temporary directory.
+
+
```
dannyroberts: Nit: Trailing newlines (see note I made in the template)

docs/changelog/index.md
```diff
@@ -11,6 +11,12 @@ an action on your part will be marked "_action optional_".
 
 ## Changelog
 
-### **2018-06-11** [Added encrypted temporary directory](0001-add-encrypted-tmp.md) (_action required_)
+### **2018-07-13** [Changelog](0002-supervisor-service-definitions.md) (_action required_) 
+
+There are several CommCare specific processes that are defined in supervisorconfiguration files. This change decouples the process definitions from code.
+
+### **2018-06-11** [Added encrypted temporary directory](0001-add-encrypted-tmp.md) (_action required_)
```
dannyroberts: Nit: Trailing single space on this line (which is what creates the diff here)

src/commcare_cloud/manage_commcare_cloud/changelog.md.j2
```diff
@@ -0,0 +1,17 @@
...
+newest first.
+
+Changes that will require an action from anyone choosing
+to update on or after the date listed will be maked "_action required_".
+Those which are worth taking a look at but which may or may not require
+an action on your part will be marked "_action optional_".
+
+
+## Changelog
+{% for changelog_entry in changelog_contents %}
```
dannyroberts: You can use `{%-` to mean trim whitespace (the newline) before and `-%}` to mean trim whitespace after. You can use these two things together to address unwanted whitespace in the output.

src/commcare_cloud/manage_commcare_cloud/changelog.md.j2
```diff
@@ -0,0 +1,17 @@
...
+### **{{ changelog_entry['date'] }}** [{{ changelog_entry['summary'] }}]({{ changelog_entry['filename'] }}) {% if changelog_entry['action_required'] %}(_action required_) {%endif %}
```
dannyroberts: This line (before endif) ends in a space, which means the line in the output ends in an unwanted space.
I would replace ` {%endif %}` here with a newline followed by `{%- endif %}`, which will trim away that newline.

src/commcare_cloud/manage_commcare_cloud/manage_commcare_cloud.py
```diff
@@ -7,14 +7,15 @@
...
-from commcare_cloud.manage_commcare_cloud.make_docs import MakeDocs
+from commcare_cloud.manage_commcare_cloud.make_docs import MakeDocs, MakeChangelog
 
 COMMAND_TYPES = [
     Install,
     Get,
     ListVaultKeys,
     Configure,
     MakeDocs,
+    MakeChangelog
```
dannyroberts: nit: a trailing comma here will make diffs cleaner next time. No need to update this now, just noting that as a nice diff-friendly practice.

src/commcare_cloud/commcare_cloud.py
```diff
@@ -150,9 +150,70 @@ def make_parser(available_envs, formatter_class=RawTextHelpFormatter,
     return parser, subparsers, commands
 
 
+def compile_changelog():
```
dannyroberts: I would move all of this code plus the `MakeChangelog` to a new file `manage_commcare_cloud/make_changelog.py`. This file, which is the entrypoint for `commcare-cloud` tool, doesn't feel like an appropriate place to put the logic behind autogenerating a piece of documentation.

As for moving `MakeChangelog` (and this logic) to its own file rather than sharing it with `make_docs.py`, that's just to emphasize that while these two commands are _analogous_ (they both autogenerate a piece of documentation), they don't actually share much if any of their implementation logic and are actually quite separate from one another.

Makefile
snopoke: I noticed this after merging and fixed it with a few other makefile changes

docs/changelog/index.md
pr33thi: There shouldn't be a space below the header? There is one in the file I was going off of (line 15 here): https://github.com/dimagi/commcare-cloud/pull/1927/files#diff-8126fa4ec84c1259fc7a6ea789a1ac53

dannyroberts: The space I'm talking about is _in_ this line, the last character before the newline.

pr33thi: ooooh I was interpreting your comment as a trailing newline. got it.

pr33thi: https://github.com/dimagi/commcare-cloud/pull/1979/commits/a516bc3342f8e0bded4bfa2bb1e07dba8169d99b

src/commcare_cloud/commcare_cloud.py
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1979/commits/14ec7abd7521f11bb0e6c6fbb6bd15ca9eb25794

src/commcare_cloud/commcare_cloud.py
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1979/commits/2e3be14fcdaa6f31038d928256d52baa33f630ef

One thing I noticed with this:
The printout only shows up when running `manage-commcare-cloud make-changelog-index`, and not when running `make` (because in that case, the printout is part of the output, so it gets written to `index.md` ☹️ Is that acceptable? I would imagine the workflow is to run the `manage-commare-cloud` command before actually running make, right?)

src/commcare_cloud/manage_commcare_cloud/changelog.md.j2
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1979/commits/a516bc3342f8e0bded4bfa2bb1e07dba8169d99b

src/commcare_cloud/manage_commcare_cloud/changelog.md.j2
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1979/commits/a516bc3342f8e0bded4bfa2bb1e07dba8169d99b

src/commcare_cloud/manage_commcare_cloud/manage_commcare_cloud.py
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1979/commits/93abc0c36d3ee7f0bf594951438519f8d795e00a

src/commcare_cloud/manage_commcare_cloud/changelog.md.j2
pr33thi: https://github.com/dimagi/commcare-cloud/pull/1979/commits/dc835ad45ce608c7b6948272a3195f999da72319

PR #1957, opened 2018-07-25
add more saved export workers

calellowitz:
@mkangia @NitigyaS 
To deal with the saved export woes on prod

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -70,7 +70,7 @@ celery_processes:
       concurrency: 8
       max_tasks_per_child: 5
     saved_exports_queue:
-      concurrency: 4
+      concurrency: 6
       max_tasks_per_child: 1
       optimize: True
   hqcelery4.internal-va.commcarehq.org:


environments/production/app-processes.yml
```diff
@@ -70,7 +70,7 @@ celery_processes:
       concurrency: 8
       max_tasks_per_child: 5
     saved_exports_queue:
-      concurrency: 4
+      concurrency: 6
```
mkangia: Is the machine capable of doing more processes in parallel?
is it possible to add a worker to a different machine instead and keep concurrency of 4 there as well. This would help stretch to 8 without burdening this machine further?

calellowitz: Current memory usage on that machine is a little below 50% and cpu usage seems to be about 30%. I think it can hand the extra two processes, and perhaps even four more to get to eight, although I wanted to start small. I am not sure we have another celery machine with less load than this, but I don't think it has to be on this machine if there is a better candidate.

mkangia: Sounds good. Though i think 6 would be better than 6 since the queue is continuously blocked at times, with consecutive exports blocking the worker.. but good to try with 6 if you think.

PR #1959, opened 2018-07-25
Fix disconnects in fullstack deploy

pr33thi:
1. Make an option to not run the `record_successful_release` task (which often errors during fullstack deploy and isn't actually necessary)
2. Remove non-required celery processes which seemed to be taking up too much memory.

### DIFF
--- a/commcare-cloud-bootstrap/environment/app-processes.yml.j2
+++ b/commcare-cloud-bootstrap/environment/app-processes.yml.j2
@@ -9,7 +9,7 @@ celery_processes:
       concurrency: 1
     sms_queue, export_download_queue:
       concurrency: 1
-#    reminder_queue,submission_reprocessing_queue,ucr_indicator_queue,ucr_queue:
+#    reminder_queue,submission_reprocessing_queue,ucr_indicator_queue,ucr_queue, celery_periodic:
 #      concurrency: 1
     flower: {}
 pillows:
--- a/commcare-cloud-bootstrap/environment/app-processes.yml.j2
+++ b/commcare-cloud-bootstrap/environment/app-processes.yml.j2
@@ -1,16 +1,14 @@
 celery_processes:
   {{ celery_host.name }}:
     beat: {}
-    celery_periodic:
-      concurrency: 4
-#    async_restore_queue,background_queue,case_rule_queue,celery,export_download_queue:
-#      concurrency: 1
-#    email_queue,reminder_case_update_queue:
-#      concurrency: 1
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/fab/fabfile.py
```diff
@@ -579,8 +579,9 @@ def _deploy_without_asking():
         execute(check_servers.perform_system_checks)
         execute_with_timing(release.update_current)
         silent_services_restart()
-        execute_with_timing(release.record_successful_release)
-        execute_with_timing(release.record_successful_deploy)
+        if skip_record == 'no':
```
nickpell: Should this be ```if skip_record == 'yes'```?  Or am I not understanding how this code works?

nickpell: Whoops, I see it now - double negative.

commcare-cloud-bootstrap/environment/app-processes.yml.j2
```diff
@@ -1,15 +1,13 @@
...
-      concurrency: 4
-    async_restore_queue,background_queue,case_rule_queue,celery,export_download_queue:
+    background_queue,case_rule_queue,celery,export_download_queue:
       concurrency: 1
     email_queue,reminder_case_update_queue:
       concurrency: 1
-    reminder_rule_queue,reminder_queue,repeat_record_queue,saved_exports_queue:
+    reminder_rule_queue,repeat_record_queue,saved_exports_queue:
       concurrency: 1
-    sms_queue,submission_reprocessing_queue,ucr_indicator_queue,ucr_queue:
```
dannyroberts: Is it worth adding these back in commented-out form? Since this has already been used for spinning up test servers, it'd be nice if at least all you had to do to get these back was uncomment some lines.

pr33thi: Can do. Is there any reason that these were split up the way that they were? If it was just for visual reasons, I can put all the commented out ones in the same line.

dannyroberts: We have it configured so that multiple queues can be consumed by the same process—so putting queue names together on the same lime, comma-separated, makes them all get processed in one process. We do that mainly to conserve memory.

So yeah, I think putting the commented out ones all on the same line comma separated is a reasonable idea. Then if someone uncomments them, they get a relatively memory-conserving default.

pr33thi: cool, fixed here:
https://github.com/dimagi/commcare-cloud/pull/1959/commits/3b905872d069604b7c8196677340e9b9547748bd

PR #1961, opened 2018-07-26
Add enikshay-reference environment

proteusvacuum:


environments/enikshay-reference/app-processes.yml
```diff
@@ -0,0 +1,64 @@
...
+#    UpdateUserSyncHistoryPillow:
+#      num_processes: 1
+#    UserCacheInvalidatePillow:
+#      num_processes: 1
+#    UserGroupsDbKafkaPillow:
+#      num_processes: 1
+#    UserPillow:
+#      num_processes: 1
+#    XFormToElasticsearchPillow:
+#      num_processes: 1
```
esoergel: Are you leaving this in here as a way of documenting that we're intentionally not running them?

proteusvacuum: this was all auto-generated by `commcare-cloud-bootstrap`, but yeah, mostly leaving it as documentation.

proteusvacuum:
@pr33thi I ran this last week some time, and it mostly worked fine.

environments/enikshay-reference/app-processes.yml
pr33thi: Yeah, that's commented out in the template file because it only gets used for provable full stack deploy (before this pr anyway), and having a bunch of pillows leads to the memory getting filled up on the `t2.micro` aws instances we use. And they aren't necessary for proving that the deploy worked, so they are commented out. Because this env is only for reference, I'm assuming that it doesn't matter that they're commented out here too

PR #1965, opened 2018-07-26
Username in settings should be a list

pr33thi:
Must have missed this one for this pr: 
https://github.com/dimagi/commcare-cloud/pull/1815

Noticed it because it gets hit when running fullstack deploy from travis.

### DIFF
--- a/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
+++ b/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
@@ -67,7 +67,7 @@ class AwsConfig(StrictJsonObject):
 
 
 class Settings(StrictJsonObject):
-    users = jsonobject.ListProperty(unicode, default='dimagi', required=True)
+    users = jsonobject.ListProperty(unicode, required=True)
 
 
 class Allocation(StrictJsonObject):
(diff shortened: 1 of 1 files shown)


commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -67,7 +67,7 @@ def wrap(cls, data):
 
 
 class Settings(StrictJsonObject):
-    users = jsonobject.StringProperty(default='dimagi')
+    users = jsonobject.ListProperty(unicode, default='dimagi', required=True)
```
gcapalbo: does the default value also need to be a list or it's ok to leave as is?

pr33thi: Good point. I don't see why the usernames shouldn't be specified so I removed the default entirely: https://github.com/dimagi/commcare-cloud/pull/1965/commits/6dfdf01e88cddac32b6e374b8884854e224494eb

Is that being used somewhere that I'm not seeing @dannyroberts ?

pr33thi:
I was barking up the wrong tree with this, closing this pr and opening this one with the right fix:
https://github.com/dimagi/commcare-cloud/pull/1967

PR #1967, opened 2018-07-26
The username setting should be a string at this level

pr33thi:


dannyroberts:
I think you're other PR was also a reasonable approach for what it's worth, though I agree it needed a bit more work before it'd work (like editing `meta.yml.j2` to match), and this is a more direct solution to the problem you're seeing.

pr33thi:
@dannyroberts agreed

PR #1970, opened 2018-07-26
Update the ami to match the provable deploy ami

pr33thi:


.travis/spec.yml
```diff
@@ -1,6 +1,6 @@
 aws_config:
   pem: /home/travis/.ssh/commcarehq_testing.pem
-  ami: ami-7dce6507
+  ami: ami-e317729c
```
dannyroberts: Can you link me to where you got his AMI? I might be searching in the wrong place, but didn't find it in `commcare-cloud` anywhere

pr33thi: hmm I'm 100% sure it existed at some point on https://cloud-images.ubuntu.com/locator/ec2/
but I don't see it there anymore. I'll play around with different ones

pr33thi: Ok, I changed it to match the one in example_spec.yml, which I just tested and it works:
https://github.com/dimagi/commcare-cloud/pull/1970/commits/f556d71dfa7082036f15c89315be4873474a5a90

PR #1972, opened 2018-07-27
Ns/db verification

NitigyaS:
Brand new PR


NitigyaS:
This function requires a `db-tools.py` script to be present on the servers.
Why I useed the script instead of in line function. 

1.  I tried using the command `sudo -iu postgres psql -l -A -F, -t` with `paramiko` but though it was executing fine on vagrant VM but, on staging and other env it was halted and misbehaving
```
ansible@hqdb1-staging:~$ ps aux | grep psql
root     15846 90.5  0.0  85980  2856 pts/4    Ss+  08:47   0:01 sudo -iu postgres psql -l -A -F, -t
postgres 15847  0.0  0.0 103620  3784 pts/4    S+   08:48   0:00 /usr/lib/postgresql/9.6/bin/psql -l -A -F, -t
ansible  15859  0.0  0.0  11992   916 pts/6    S+   08:48   0:00 grep --color=auto psql
```
Note the high cpu usage of 90%

2. I think using a script at server will allow me to run more complex queries on postgres if require in future, Instead of passing them as inline function and writing in commcare-cloud codebase.

NitigyaS:
Output: 
```
(ansible) Dimagis-MacBook-Pro:repo nsharma$ cchq staging list-databases
Vault Password:
hqdb1-staging.internal-va.commcarehq.org:
    Expected Databases:
        - airflow
        - commcarehq
        - commcarehq_p1
        - commcarehq_p2
        - commcarehq_p3
        - commcarehq_p4
        - commcarehq_p5
        - commcarehq_proxy
        - commcarehq_synclogs
        - commcarehq_ucr
        - formplayer
    Additional Databases:
        - custom_ucr_database
        - commcarehq_icds_testdata
        - CUSTOM_DATABASES
        - alphacchq
        - cchqcelerytest
        - hqblank
        - pact_prod
        - sentry
        - test_commcarehq
```

src/commcare_cloud/ansible/roles/devops_scripts/templates/db-tools.py.j2
```diff
@@ -0,0 +1,58 @@
...
+        usage()
+        sys.exit(2)
+
+# Parse GetOpts and execute the desired functionality
+    for opt, arg in opts:
+        if opt in ('-l', '--list-all'):
+            list_all_database()
+        else:
+            usage()
+            sys.exit(2)
```
snopoke: do we really need this script? Is there any reason we can't run the command directly instead of via a script? From the shell you can run: `psql -l -A -F, -t` which will output the list of databases in CSV format.

src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -0,0 +1,122 @@
+import collections
+import paramiko
+from commcare_cloud.commands.command_base import CommandBase, Argument
+from commcare_cloud.commands.inventory_lookup.getinventory import get_instance_group
+from commcare_cloud.environment.main import get_environment
+
+
+class PrivilegedCommands():
```
snopoke: I assume you're doing this instead of running it with ansible so that you can get the command output correct?

I went through a similar issue before and settled on using Fabric to run the command which has a higher level api than paramico: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/commands/migrations/copy_files.py#L197

You could adapt the code there to get the output as follows:
```
out = StringIO()
sudo(command, stdout=out)
```

src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -0,0 +1,122 @@
...
+                    if database not in dbs_expected_on_host[host_address]:
+                        print(" " * 8 + "- " + database)
+        else:
+            # Print Default
+            for host_address in dbs_expected_on_host.keys():
+                print(host_address + ":")
+                print(" " * 4 + "Expected Databases:")
+                for database in dbs_expected_on_host[host_address]:
+                    print(" " * 8 + "- " + database)
+
```
snopoke: Quite a lot of duplication in the output logic. Could be clean up do something like this (also you can use `clint.textui` to do indenting and more:
```python
    dbs_expected_by_host = get_expected_dbs()
    dbs_present_by_host = None
    if args.compare:
        dbs_present_by_host = get_present_dbs()
    
    for host_address, dbs_expected in dbs_expected_by_host.items():
        dbs_present = dbs_present_by_host.get(host_address, {}) if dbs_present_by_host is not None else None
        puts(colored.blue(host_address + ":"))
        with indent():
            puts("Expected Databases:")
            with indent():
                for database in dbs_expected:
                    puts("- " + database)
                    dbs_present and dbs_present.pop(database, None)

            if dbs_present:
                puts(colored.orange("Additional Databases:"))
                with indent():
                    for database in dbs_present:
                        puts("- " + database)
```

src/commcare_cloud/ansible/roles/devops_scripts/templates/db-tools.py.j2
NitigyaS: Thanks for reviewing. I answered this in the first comment.

src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -0,0 +1,115 @@
...
+                for database in dbs_present_in_host[host_address]:
+                    if database not in dbs_expected_on_host[host_address]:
+                        print(" " * 8 + "- " + database)
+
+    @staticmethod
+    def get_present_dbs( args):
+        dbs_present_in_host = collections.defaultdict(list)
+        args.server = 'postgresql'
+        ansible_username = 'ansible'
+        command = "sudo -iu postgres python /usr/local/sbin/db-tools.py  --list-all"
```
snopoke: the command doesn't need to include `sudo` since that's handled by fabric. Since you want to run this as the `postgres` user you'll need to pass that to fabric too: `sudo(command, user='postgres')`

snopoke:
need to update docs for build

PR #1973, opened 2018-07-27
Remove block from pg_conf

NitigyaS:
Will deploy on staging before rolling out.


NitigyaS:
I wanted to do it in steps. I wanted to first remove this config as this gives a false impression that we are whitelisting at PostgreSQL level. Will throw another PR for adding whitelisting in pgbouncer after checking and testing it on staging.

PR #1979, opened 2018-07-30
Autogenerate changelog fixes

pr33thi:
https://github.com/dimagi/commcare-cloud/pull/1954#discussion_r205833635

Addressing @dannyroberts's comments from that PR^^^

### DIFF
--- a/src/commcare_cloud/manage_commcare_cloud/make_changelog.py
+++ b/src/commcare_cloud/manage_commcare_cloud/make_changelog.py
@@ -1,0 +1,86 @@
+from __future__ import print_function
+from __future__ import absolute_import
+
+import jinja2
+
+import os
+import re
+from commcare_cloud.commands.command_base import CommandBase
+
+
+def compile_changelog():
+    # Parse the contents of the changelog dir
+    changelog_contents = []
+    files_to_ignore = ['0000-changelog.md', 'index.md']
+    changelog_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', '..', 'docs', 'changelog')
+    sorted_files = _sort_files(changelog_dir)
+    for change_file_name in sorted_files:
+        if change_file_name not in files_to_ignore:
+            try:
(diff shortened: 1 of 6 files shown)


src/commcare_cloud/manage_commcare_cloud/make_changelog.py
```diff
@@ -0,0 +1,86 @@
...
+    changelog_contents = []
+    files_to_ignore = ['0000-changelog.md', 'index.md']
+    changelog_dir = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', '..', 'docs', 'changelog')
+    sorted_files = _sort_files(changelog_dir)
+    for change_file_name in sorted_files:
+        if change_file_name not in files_to_ignore:
+            try:
+                _parse_changelog_file(changelog_contents, changelog_dir, change_file_name)
+            except Exception:
+                print("Error parsing the file {}.".format(change_file_name))
```
dannyroberts: Answering https://github.com/dimagi/commcare-cloud/pull/1954#discussion_r206258018 here
> 2e3be14
> 
> One thing I noticed with this:
> The printout only shows up when running manage-commcare-cloud make-changelog-index, and not when running make (because in that case, the printout is part of the output, so it gets written to index.md ☹️ Is that acceptable? I would imagine the workflow is to run the manage-commare-cloud command before actually running make, right?)

I'm guessing when you see this happen now, the actual traceback (coming from `raise`) does get printed to the screen? And yet this line doesn't, right? If I'm following along correctly, that's because the traceback is going to stderr and this print is going to stdout. To print to stderr you can do `print(..., file=sys.stderr)`

pr33thi: https://github.com/dimagi/commcare-cloud/pull/1979/commits/5ceffb0e03f0a7b5569178dbf2db2008e63d7bdf

dannyroberts:
Thanks!

PR #1980, opened 2018-07-30
Add curly braces to new localsettings variable

pr33thi:


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -119,7 +119,7 @@ EMAIL_BACKEND='django.core.mail.backends.console.EmailBackend'
 
 EMAIL_SUBJECT_PREFIX = '[{{ env_monitoring_id }}] '
 
-ENABLE_SOFT_ASSERT_EMAILS = localsettings.ENABLE_SOFT_ASSERT_EMAILS|default(True)
+ENABLE_SOFT_ASSERT_EMAILS = {{ localsettings.ENABLE_SOFT_ASSERT_EMAILS|default(True) }}
```
dannyroberts: Did you check that this outputs something like `True` or `False` that results in valid python? (I.e. if it output `true`, for example, then it wouldn't be valid python.

pr33thi: I actually have no idea how to test this besides running the full stack deploy, which I'm currently doing. what's the better way?

dannyroberts: Yeah a much better way! You can run `cchq <env> update-config`

dannyroberts: > Is this what you mean?
It's not; I meant to check that this interpolation _outputs_ `True` or `False`—something you can check by running `cchq <env> update-config` (in check mode) and seeing what it produces in the diff. (Or if there's no diff because you already ran it, checking the localsettings.py file that ended up on the server.)

pr33thi: Ah ok, there are multiple lines that look like:
```
-ENABLE_SOFT_ASSERT_EMAILS = localsettings.ENABLE_SOFT_ASSERT_EMAILS|default(True)
+ENABLE_SOFT_ASSERT_EMAILS = True
```
Which is enough evidence to say that this worked?

pr33thi: And it wouldn't have worked if it set that value to a non-bool, which would happen if the file that the localsettings.py.j2 template got the value from had it set to a non-boolean?

PR #1983, opened 2018-07-30
Change soft assert from no to valid python

pr33thi:
To work with this code:
https://github.com/dimagi/commcare-cloud/pull/1980#pullrequestreview-141720653


environments/echis/public.yml
```diff
@@ -61,7 +61,7 @@ localsettings:
   EMAIL_SMTP_HOST: 'smtp.gmail.com'
   EMAIL_SMTP_PORT: 587
   EMAIL_USE_TLS: yes
-  ENABLE_SOFT_ASSERT_EMAILS: no
+  ENABLE_SOFT_ASSERT_EMAILS: False
```
dannyroberts: These two are the same. (`no` is a boolean literal equivalent to `False` in YAML: see https://docs.ansible.com/ansible/2.4/YAMLSyntax.html)

pr33thi: Ok, in that case I'll close out this PR

PR #1986, opened 2018-07-31
updates to copy command

snopoke:
* setup auth between target and source hosts
* allow specifying target user
* allow excluding files from source


kaapstorm:
Is the idea behind separating "prepare", "copy" and "cleanup" so that you can copy (using different plans) between the same hosts that have been prepared?

snopoke:
@kaapstorm that's a good question which I hadn't thought much about. In this case it's probably not completely necessary though it does give the user a bit more flexibility in terms of how this get's used. For example if they wanted to run the copy multiple times to 'top' up the target folder.

PR #1989, opened 2018-07-31
Added new server for pgsynclog

NitigyaS:
new pgsynclog1


environments/production/postgresql.yml
```diff
@@ -19,7 +19,7 @@ dbs:
     host: hqdb3
     query_stats: True
   synclogs:
-    host: pgsynclog
+    host: pgsynclog1
```
dannyroberts: Won't this make processes start reading and writing to `pgsynclog1`? Sounds like we can do that until `pgsynclog1` is promoted to master

environments/production/inventory.ini
```diff
@@ -1,3 +1,4 @@
+
```
dannyroberts: oops?

environments/production/postgresql.yml
NitigyaS: It will be merge and deployed only after I promote it. There is a ticket for the same.

PR #1991, opened 2018-07-31
Travis encryption

pr33thi:
Changing how travis secrets are encrypted


dannyroberts:
Hey so looking at the errors, this is actually a great success! Provable deploy runs until it hangs for 10 minutes (which isn't as good as running until it hits 50 minutes, but is probably a sign that a machine becomes unresponsive due to memory or something), and the main job fails because the secrets directory, which is committed to the repo, is getting removed, and part of the test is running `make` and then checking if there's a git diff. In this case it fails because the removal of `.travis/secrets/` results in a git diff.

From 
```
(commcare-cloud) francine:commcare-cloud droberts$ ansible-vault decrypt --help
Usage: ansible-vault decrypt [options] file_name

encryption/decryption utility for Ansible data files

Options:
...
  --output=OUTPUT_FILE  output file name for encrypt or decrypt; use - for
                        stdout
...
```
I can see that there's an `--output` option on `ansible-vault` decrypt, and you can use it to specify where to decrypt the file to, rather than doing it in place. A nice perk to doing that is that you also then don't have to do a separate `mv` for each file; you can just list the right location for it to end up right there in the decryption command.

pr33thi:
@dannyroberts It looks like this is good to merge?
I'll then:
-add the vault pw to keepass
-delete every env variable here: https://travis-ci.org/dimagi/commcare-cloud/settings, besides VAULT_PASSWORD

dannyroberts:
Oh shit! (In a good way.) Looks like in addition to encryption working, the deploy also managed to get really far along!

dannyroberts:
And yup, that sounds like a good plan!

PR #1992, opened 2018-08-01
Update Kissmetrics API key for staging

orangejenny:
@nickpell


orangejenny:
Self-merging, as there's no practical way to review this and if there were relevant tests they would have broken already (there's no data coming through to the kiss sandbox).

PR #1994, opened 2018-08-01
Refactor deploy_stack.yml

pr33thi:


src/commcare_cloud/ansible/deploy_http_proxy.yml
```diff
@@ -11,14 +11,10 @@
         options: 'no-query no-digest'
     pri_domain_name: "{{ internal_domain_name }}"
   roles:
-    - role: squid
-      tags:
-        - 'squid'
+    - {role: squid, tags: 'squid', when: http_proxy_address|default(false)}
```
snopoke: I don't think this needs the `when` condition since it will only get applied to hosts in the `squid` group which mostly doesn't exist.

pr33thi: What do you mean? Is it not important to do things to the `squid` group?

dannyroberts: @snopoke searching our environments for `[squid]`, I'm seeing that _no_ environments set it. Should we just remove squid-related code from here entirely?

commcare-cloud-bootstrap/bootstrap.sh
```diff
@@ -14,4 +14,4 @@ done
 commcare-cloud $ENV bootstrap-users --quiet --branch=$BRANCH
 commcare-cloud $ENV deploy-stack --skip-check --quiet -e 'CCHQ_IS_FRESH_INSTALL=1' --branch=$BRANCH
 
-commcare-cloud $ENV fab deploy:confirm=no,skip_record=yes --show=debug --set ignore_kafka_checkpoint_warning=true --branch=$BRANCH
+commcare-cloud $ENV fab deploy:confirm=no,skip_record=yes --show=debug --set ignore_kafka_checkpoint_warning=true
```
dannyroberts: Maybe this is from a mistake during the merge? I think you do want `--branch` here, right?

src/commcare_cloud/ansible/deploy_http_proxy.yml
pr33thi: Merging this but gentle reminder that this question is here! @snopoke

commcare-cloud-bootstrap/bootstrap.sh
pr33thi: Looks like file was in conflict, fixed here: 
https://github.com/dimagi/commcare-cloud/pull/1994/commits/a2add1d0e7ebf16ecc5ad5002da3b1768206cfa4

PR #1998, opened 2018-08-02
Create host_file if it doesn't exist

pr33thi:


src/commcare_cloud/commands/fab.py
```diff
@@ -56,6 +56,10 @@ def run(self, args, unknown_args):
             env = get_environment(args.env_name)
             fab_args.extend(['--disable-known-hosts',
                              '--system-known-hosts', env.paths.known_hosts])
+        # Create known_hosts file if it doesn't exist
+        known_hosts_file = os.path.join(os.path.dirname(__file__), '..', '..', '..', 'environments', args.env_name, 'known_hosts')
```
dannyroberts: See how in the line above we use `env.paths.known_hosts`? I think you should use that here too instead of reconstructing the path.

PR #2000, opened 2018-08-02
Prep for separate env

dannyroberts:
Pulled out from https://github.com/dimagi/commcare-cloud/pull/1858


src/commcare_cloud/manage_commcare_cloud/test_environments.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class TestEnvironments(CommandBase):
+    command = 'test-environments'
+    help = "Run test environments"
+
+    def run(self, args, unknown_args):
+        import nose
+        nose.runmodule('commcare_cloud.manage_commcare_cloud',
+                       argv=['manage-commcare-cloud', '-v'])
```
snopoke: what does this actually do?

.travis/tests.sh
```diff
@@ -16,8 +16,15 @@ then
         sudo python -m py_compile /home/cchq/www/travis/current/localsettings.py
     }
 
+    test_dimagi_environments() {
+        git clone https://github.com/dimagi/commcare-environments.git
```
snopoke: maybe just clone with depth = 1

.travis/tests.sh
```diff
@@ -16,8 +16,15 @@ then
...
+        ln -s commcare-environments/environments environments
```
snopoke: based on your comments in https://github.com/dimagi/commcare-environments/pull/3 I think you could just clone the repo to this location directly

.travis/tests.sh
snopoke: Not necessary to address this now but just noting that if we make a breaking change in this repo that requires a change to the env files then we'd need a way to change this to pull from a branch in order to make the build pass (or something like that).

dannyroberts: Yeah I agree with that, and I think that also highlights a sort of fundamental problem that versioning is here to solve—making this painful for us will also force us (to some extent at least) to version in a way that's mindful of others. For example, if we find ourselves needing to point this to a branch in a PR and then merge the branch in environments and then reset this to master, wait for build and merge, that's a sign that we should be creating a backwards compatibility or deprecation window so that these two repo's changes can be loosely coordinated rather than tightly coordinated. I imagine it'll also force other good patterns to the surface that I haven't yet imagined.

.travis/tests.sh
dannyroberts: Oh yeah, thanks. Not sure how this test passed since commcare-environments/environments doesn't exist... Will investigate.

dannyroberts: Oh I see, since `environments/` still exists here, this link line is interpreted as "link commcare-environments/environments to environments/environments". I think for this PR I'm going to fix but then comment out these lines, since I'm not removing `environments/` from here just yet. That way the PR that removes environments will do less (hopefully just uncomment these lines), which will make the coordination easier.

src/commcare_cloud/manage_commcare_cloud/test_environments.py
dannyroberts: Part of the point of this PR is to separate the tests that are based on the dimagi `environments/` file from everything else. This line is the from-python way of doing the equivalent of running

```
nosetests commcare_cloud.manage_commcare_cloud -v
```
from the command line. (Since tests themselves now live in `src/commcare_cloud/manage_commcare_cloud/tests/`)

dannyroberts:
@pr33thi This PR goes along with changes described here https://github.com/dimagi/commcare-environments/pull/3. It sets us up to be closer to being able to remove `environments/` from this repo entirely and checking out our own environments from https://github.com/dimagi/commcare-environments instead, which will allow external users the opportunity to do the same in a first-class supported way.

src/commcare_cloud/manage_commcare_cloud/test_environments.py
snopoke: :+1: that's what I thought but was a bit confused for some reason.

PR #2001, opened 2018-08-02
Define the riak secrets

pr33thi:
Set a riak password rather than waiting for them to be generated by riak. Reason: I want the variable to persist over multiple ansible-playbook command calls because I'm currently working on parallelizing a bunch of this.

### DIFF
--- a/src/commcare_cloud/ansible/group_vars/all.yml
+++ b/src/commcare_cloud/ansible/group_vars/all.yml
@@ -54,8 +54,8 @@ old_s3_blob_db_s3_bucket: 'blobdb'
 
 # To use these vars without deploying the riak control machine
 # add riakcs_s3_access_key and riakcs_s3_secret_key to the ansible secret config directory
-s3_blob_db_access_key: "{{ riakcs_s3_access_key if riakcs_s3_access_key else (hostvars[riakcs_control]['riak_key'] if s3_blob_db_enabled else '') }}"
-s3_blob_db_secret_key:  "{{ riakcs_s3_secret_key if riakcs_s3_secret_key else (hostvars[riakcs_control]['riak_secret'] if s3_blob_db_enabled else '') }}"
+s3_blob_db_access_key: "{{ secrets.S3_BLOB_DB_ACCESS_KEY | default(None) }}"
+s3_blob_db_secret_key:  "{{ secrets.S3_BLOB_DB_SECRET_KEY | default(None) }}"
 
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -54,8 +54,8 @@ old_s3_blob_db_s3_bucket: 'blobdb'
 
 # To use these vars without deploying the riak control machine
 # add riakcs_s3_access_key and riakcs_s3_secret_key to the ansible secret config directory
-s3_blob_db_access_key: "{{ riakcs_s3_access_key if riakcs_s3_access_key else (hostvars[riakcs_control]['riak_key'] if s3_blob_db_enabled else '') }}"
-s3_blob_db_secret_key:  "{{ riakcs_s3_secret_key if riakcs_s3_secret_key else (hostvars[riakcs_control]['riak_secret'] if s3_blob_db_enabled else '') }}"
```
millerdev: I think this change is not necessary, and it may mess up some existing environments. But you should be able to set `secrets.RIAKCS_S3_ACCESS_KEY` and `secrets.RIAKCS_S3_SECRET_KEY` in `commcare-cloud-bootstrap/environment/private.yml`, and those values will be used here:

https://github.com/dimagi/commcare-cloud/blob/c308db83baaa2576778f4bd22c91aae4d99ce1ff/src/commcare_cloud/ansible/group_vars/all.yml#L167-L168

pr33thi: That makes sense, and just to make sure I have it straight: your proposed change would be functionally the same as this (for new environments), but would make sure that old environments still work?

millerdev: correct

pr33thi: Ok cool, and do you think it would make sense for me to try to get the existing environments to use what I have defined here, just so this doesn't get too messy with outdated variables?

millerdev: Followed up on slack.

dannyroberts:
Outcome of the conversation yesterday was
- We can't set credentials in the vault file until they've been created in riak-cs (because you can't tell riak-cs what you want the key and secret to be; it has to create them for you)
- There's no stable way around this, so we're stuck with the status quo for now

pr33thi:
@dannyroberts just flagging that we still need some resolution to this to make this work for in the parallelized fullstack deploy

PR #1999, opened 2018-08-02
Separate out deploy_stack playbooks in bootstrap.sh

pr33thi:
Separate out deploy_stack playbooks in bootstrap.sh

To be merged in afer this: https://github.com/dimagi/commcare-cloud/pull/1994

PR #2002, opened 2018-08-03
Move logic from deploy_dbs.yml into separate playbooks

pr33thi:
To be merged in after this:
https://github.com/dimagi/commcare-cloud/pull/1999
^^ Note, accidentally merged in the wrong order, here's the revert: https://github.com/dimagi/commcare-cloud/pull/2008

Also, I have a dumb question. I am merging this in to that branch^ but the diff definitely shows things that are in that branch. Any idea why?


src/commcare_cloud/ansible/deploy_db.yml
```diff
@@ -1,194 +1,18 @@
...
-
-- name: Kafka
-  hosts: kafka
-  become: true
-  environment:
-    JMX_PORT: 9999
-  roles:
-    - {role: kafka, tags: 'kafka'}
 
+- import_playbook: update_apt_cache.yml
```
gcapalbo: i don't see a new file added for this one in this pr, is this adding something new to this list of steps that wasn't here before?

also i see you created a new file for `common_db_machine_setup.yml` in this pr but i don't see a reference it from here. is that intentional?

gcapalbo:
left one comment otherwise looks good to me. might also want to have @dannyroberts take a pass

dannyroberts:
> Also, I have a dumb question. I am merging this in to that branch^ but the diff definitely shows things that are in that branch. Any idea why?

Looks like one of these two was rebased, so while the commit messages (and for the most part the diffs) of each commit is the same, they are different commits with different commit hashes. (That's what happens when you rebase, because rebasing involves changing the parent commit of the first commit, and that then has to cascade all the way down.)

src/commcare_cloud/ansible/deploy_db.yml
pr33thi: Oops, copy and paste error. Put in the correct playbook name: https://github.com/dimagi/commcare-cloud/pull/2002/commits/5a9be422713688b010090e338628e8626247ebcf

dannyroberts:
Hey does https://github.com/dimagi/commcare-cloud/pull/2038 include this PR now?

pr33thi:
@dannyroberts yes, so I'm going to close this PR

PR #2004, opened 2018-08-06
Add service check to the end of bootstrap.sh

pr33thi:


commcare-cloud-bootstrap/bootstrap.sh
```diff
@@ -15,3 +15,8 @@ commcare-cloud $ENV bootstrap-users --quiet --branch=$BRANCH
 commcare-cloud $ENV deploy-stack --skip-check --quiet -e 'CCHQ_IS_FRESH_INSTALL=1' --branch=$BRANCH
 
 commcare-cloud $ENV fab deploy:confirm=no,skip_record=yes --show=debug --set ignore_kafka_checkpoint_warning=true --branch=$BRANCH
+
+commcare-cloud $ENV django-manage check_services
+
+proxy=$(grep -A1 '\[pvtest-proxy-0\]' environments/pvtest/inventory.ini | tail -n 1| awk '{print $2}' | awk -F'=' '{print $2}')
```
gcapalbo: this might be intentional, but did you mean to leave `pvtest` hardcoded in there?

commcare-cloud-bootstrap/bootstrap.sh
```diff
@@ -15,3 +15,8 @@ commcare-cloud $ENV bootstrap-users --quiet --branch=$BRANCH
...
+curl https://${proxy}/serverup.txt --insecure
```
gcapalbo: did you also want to check celery with `?heartbeat` or is that intentionally excluded here?

commcare-cloud-bootstrap/bootstrap.sh
pr33thi: That was a mistake, fixed here: https://github.com/dimagi/commcare-cloud/pull/2004/commits/60ddfa12ec01b235616b64be7c3ec53f36425ed2

commcare-cloud-bootstrap/bootstrap.sh
pr33thi: I'm not familiar with that syntax, but`django-manage check_services` on line 19 checks celery, so I think this is covered?

gcapalbo: i see, yeah it looks like check_services is including the heartbeat check too.

actually, is serverup.txt checking anything that check_services isn't already checking? although i suppose it's also checking that it can respond to an http request

pr33thi: Yeah exactly, its just to check that the webserver is up and responding

PR #2019, opened 2018-08-10
vault key rotation

sravfeyn:
@dannyroberts 
Haven't updated other envs, because either they are test or external


sravfeyn:
Also updated keepass, will be good if you can confirm that you can decrypt with new entries

emord:
@sravfeyn any reason not to update PNA and swiss as well?

PR #2023, opened 2018-08-14
Add request length to nginx logs

emord:
@dimagi/devops Looking for some feedback

> $request_length - request length (including request line, header, and request body)

https://nginx.org/en/docs/http/ngx_http_log_module.html

This could be useful for current interrupt ticket to see how large of a request a mobile worker is trying to send. It seems like it could be useful in the future, but not sure if we have a preferred way to add these, or a reason to not add it.


NitigyaS:
Seems a good idea if it helps in long run. I have one concern, but I am not sure how frequently we use access logs. Since `request_length` includes all headers, body etc. It might cause `access_logs` to fill much more quickly. We might need to tweak our log rotation for these logs or we can also consider enabling/disabling it followed by Nginx reload if we want to debug.

PR #2025, opened 2018-08-14
Add analytics queue

emord:
Adding for an upcoming HQ PR

### DIFF
--- a/docs/changelog/0004-add-analytics-queue.md
+++ b/docs/changelog/0004-add-analytics-queue.md
@@ -1,0 +1,20 @@
+# 4. Add queue for analytics tasks
+
+**Date:**  2018-08-16
+
+**Optional per env:** Yes
+
+**Dependant CommCare version:** None
+
+## Change Context
+Tasks for analytics reporting have been separated into a new analytics celery queue.
+
+## Details
+All analytics tasks will be sent to analytics_queue.
+
+## Steps to update
+1. Run the following to update the supervisor configuration:
+
+```bash
+commcare-cloud <env> update-supervisor-confs
+```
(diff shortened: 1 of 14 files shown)


emord:
Right now they would get dropped which is a pattern that we follow for some other queues (where required=False)

Do you think that we should add a queue/add it to the background queue on the enterprise environments to process those tasks if they're accidentally called? One benefit of this way is that there's no way we can collect data on those environments, even accidentally

emord:
Did some checking and there certainly will be some tasks that are queued on enterprise environments. I've pushed changes for these to be in all environments.

@snopoke This adds a celery queue for analytics tasks. This is a change that shouldn't affect any (known) environments running. Is it something that should go under the changelog? (seems small enough to not be required)

docs/changelog/0004-add-analytics-queue.md
```diff
@@ -0,0 +1,20 @@
+# 4. Add queue for analytics tasks
+
+**Date:**  2018-08-16
+
+**Optional per env:** Yes
+
+**Dependant CommCare version:** None
```
snopoke: can you update this with the merge commit on the PR which requires it: https://github.com/dimagi/commcare-hq/commit/a507757628bc5c087fd1badc0145e39c5bf790ae

docs/changelog/0004-add-analytics-queue.md
```diff
@@ -0,0 +1,20 @@
+# 4. Add queue for analytics tasks
+
+**Date:**  2018-08-16
+
+**Optional per env:** Yes
```
dannyroberts: I think this isn't optional, since the new queue is a required queue in app_processes. If we want to keep it that way other than changing this to no, I think we also need to note that `analytics_queue` needs to be added to the environment.

emord: Oops. yeah you're right. I had intially had it as optional because required=False, but changed it due to cal's

PR #2038, opened 2018-08-17
Set up the fullstack deploy for parallelizing tasks.

pr33thi:
There's a lot going on in here, but the idea is that I refactored code so that `bootstrap.sh` is just a series of playbooks

The big thing that I had to do to make this possible is change up the logic of the riak secret. Before, it would get stored as a variable within an ansible command, but now I am changing it to get written to the vault file, the bulk of this logic is in `src/commcare_cloud/ansible/deploy_riakcs.yml` and `src/commcare_cloud/commands/ansible/run_module.py`

This currently works locally in the fullstack deploy, because the vault.yml file is unencrypted. I'm currently working on getting it to work for all other envs (where the vault file is encrypted) hence the do not merge tag.


commcare-cloud-bootstrap/bootstrap.sh
```diff
@@ -3,20 +3,39 @@ set -e
...
-
 commcare-cloud $ENV bootstrap-users --quiet --branch=$BRANCH
-commcare-cloud $ENV deploy-stack --skip-check --quiet -e 'CCHQ_IS_FRESH_INSTALL=1' --branch=$BRANCH
-
+commcare-cloud $ENV ansible-playbook update_apt_cache.yml $PLAYBOOK_TAGS
+commcare-cloud $ENV ansible-playbook deploy_common.yml $PLAYBOOK_TAGS
+commcare-cloud $ENV ansible-playbook deploy_lvm.yml $PLAYBOOK_TAGS
+commcare-cloud $ENV ansible-playbook deploy_db.yml $PLAYBOOK_TAGS
+commcare-cloud $ENV ansible-playbook deploy_riakcs.yml $PLAYBOOK_TAGS --skip-tags=delete_riak_secret
+cchq pvtest update-riak
```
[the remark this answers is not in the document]
pr33thi: https://github.com/dimagi/commcare-cloud/pull/2038/commits/c0302783a36ab262b05aa515d825df29a52dcff0

src/commcare_cloud/ansible/deploy_riakcs.yml
```diff
@@ -154,7 +154,27 @@
   hosts: riakcs_control_host
   roles:
     - role: riakcs/create_admin_user
-      when: riak_key is undefined or riak_secret is undefined
+#      when: riak_key is undefined or riak_secret is undefined
```
millerdev: Curious why this was removed? The purpose of this is to only create a new user when credentials do not already exist, which should only be on initial cluster setup.

Will this change cause a new user to be created each time this task is run on riakcs hosts?

src/commcare_cloud/commands/ansible/run_module.py
```diff
@@ -203,3 +205,34 @@ def run(self, args, unknown_args):
         args.shell_command = 'echo {{ inventory_hostname }}'
         args.silence_warnings = False
         return RunShellCommand(self.parser).run(args, unknown_args)
+
+
+class UpdateRiakSecrets(CommandBase):
+    command = 'update-riak'
```
millerdev: When I saw this above, the command name `update-riak` did not clearly communicate what aspect of riak was being updated. What do you think of renaming this to `update-riak-secrets`?

src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -54,8 +54,8 @@ old_s3_blob_db_s3_bucket: 'blobdb'
 
 # To use these vars without deploying the riak control machine
 # add riakcs_s3_access_key and riakcs_s3_secret_key to the ansible secret config directory
-s3_blob_db_access_key: "{{ riakcs_s3_access_key if riakcs_s3_access_key else (hostvars[riakcs_control]['riak_key'] if s3_blob_db_enabled else '') }}"
-s3_blob_db_secret_key:  "{{ riakcs_s3_secret_key if riakcs_s3_secret_key else (hostvars[riakcs_control]['riak_secret'] if s3_blob_db_enabled else '') }}"
+s3_blob_db_access_key: "{{ secrets.S3_BLOB_DB_ACCESS_KEY | default(None) }}"
+s3_blob_db_secret_key:  "{{ secrets.S3_BLOB_DB_SECRET_KEY | default(None) }}"
```
snopoke: doesn't this already use the value from vault if it exists? In `all.yml` there is this:
```
riakcs_s3_access_key: "{{ secrets.RIAKCS_S3_ACCESS_KEY | default(None) }}"
```

millerdev: I think @snopoke is correct, this change is not necessary.

millerdev: I think you should write `secrets.RIAKCS_S3_ACCESS_KEY` and `secrets.RIAKCS_S3_SECRET_KEY` in the vault rather than `secrets.S3_BLOB_DB_ACCESS_KEY` and `secrets.S3_BLOB_DB_SECRET_KEY`. They will be assigned to the proper variable in `all.yml`:

https://github.com/dimagi/commcare-cloud/blob/df9ab5d1fc1705c92ebe617fbf26dee8ce658090/src/commcare_cloud/ansible/group_vars/all.yml#L171-L172

pr33thi: Offlined with Daniel that keeping both `s3_blob_db_secret_key` and `riakcs_s3_access_key ` is more robust for within an ansible session (mostly writing this here for my own memory). Updated here: https://github.com/dimagi/commcare-cloud/pull/2038/commits/21ffe5dbc4feafdfc1eae0c4deba3aab58189868

src/commcare_cloud/commands/ansible/run_module.py
pr33thi: Updated here https://github.com/dimagi/commcare-cloud/pull/2038/commits/5d32c8de9264f96efa4a76de1e0674fa19a3f6be

src/commcare_cloud/ansible/deploy_riakcs.yml
pr33thi: Commented out for debugging, uncommented here: https://github.com/dimagi/commcare-cloud/pull/2038/commits/5edde79f49ec74bfface323491976d3b22b36294

millerdev:
👍 Looking good

dannyroberts:
> Would it make your life easier if we could say riak is not a recommended/supported backend and use S3 or file system by default?

@emord I think this would definitely make (all of!) our lives easier if we could say this, but I don't think we can yet, even if ICDS gets off riak and onto the CSP's object store. That's because we're committed to being deployable fully on open source technologies; on a practical level it means we can tell any new self-hosted deployment that they can start on a single large server and scale up to a dozen or so servers without having to buy into a proprietary backend.

dannyroberts:
(resolved conflict by deleting `docs/commcare-cloud/commands/index.md` and running `make`)

dannyroberts:
> What exactly is the full stack deploy for?

The idea is to have a test that runs the process of deploying to a new set of VMs, basically following the instructions that we'd give third parties for setting up a new environment and testing that they work. We're including riak here (even though it's not needed for a monolith) just because it's the more complicated of the two and currently still is required for a medium-large fully OSS cluster.

pr33thi:
This has been open for a while and we're not planning on continuing with this at least for now, closing

Issue #2009, opened 2018-08-08
Fix service command for celery beat


dannyroberts:
The `celery` service for the `service` command gets the name of the celery beat process wrong. (I think it's `...celery_beat` not `...celery_beat_0` but you can check that by finding it in the output of a `sudo supervisorctl status` command.)

Example:
```
$ cchq pna service celery status

196.207.230.171 | SUCCESS | rc=0 >>
No such process commcare-hq-pna-celery_beat_0
commcare-hq-pna-celery_celery_0  RUNNING    pid 32540, uptime 1:28:56
commcare-hq-pna-celery_repeat_record_queue_0 RUNNING    pid 32551, uptime 1:28:56
commcare-hq-pna-celery_background_queue,export_download_queue,saved_exports_queue_0 RUNNING    pid 32542, uptime 1:28:56
commcare-hq-pna-celery_submission_reprocessing_queue,ucr_queue,async_restore_queue,email_queue,case_rule_queue_0 RUNNING    pid 32543, uptime 1:28:56
commcare-hq-pna-celery_celery_periodic_0 RUNNING    pid 21491, uptime 10:55:24
commcare-hq-pna-celery_flower    RUNNING    pid 32547, uptime 1:28:56
```

(emphasis on the first line) or the more specific

```
$ cchq pna service celery status --only=beat

196.207.230.171 | SUCCESS | rc=0 >>
No such process commcare-hq-pna-celery_beat_0
```

pr33thi:
Output to find the correct service name: commcare-hq-pna-celerybeat
```
ansible@pnaserver1:~$ sudo supervisorctl status
[sudo] password for ansible:
commcare-hq-pna-celery_background_queue,export_download_queue,saved_exports_queue_0 RUNNING    pid 6050, uptime 1 day, 20:07:53
commcare-hq-pna-celery_celery_0  RUNNING    pid 5972, uptime 1 day, 20:07:54
commcare-hq-pna-celery_celery_periodic_0 RUNNING    pid 6577, uptime 1 day, 20:07:52
commcare-hq-pna-celery_flower    RUNNING    pid 5518, uptime 1 day, 20:07:54
commcare-hq-pna-celery_repeat_record_queue_0 RUNNING    pid 6440, uptime 1 day, 20:07:52
[...]

PR #2039, opened 2018-08-17
Change the name of the celery beat service

pr33thi:
Fixes https://github.com/dimagi/commcare-cloud/issues/2009

Changing the name of the beat service itself so it matches the pattern of the other services.

After this is merged, I'll go through each environment and 
- stop the beat service
- run `cchq <env> update-supervisor-confs`
- start the beat service


src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_celery_beat.conf.j2
```diff
@@ -1,4 +1,4 @@
-[program:{{ project }}-{{ deploy_env }}-celerybeat]
+[program:{{ project }}-{{ deploy_env }}-celery_beat_0]
```
emord: Do you also want to change the log names? I'm not sure if there are any other automated tasks that would break for them

gcapalbo:
how come we're renaming this? the workers have an `_n` prefix because there can be multiple workers running on the same machine which consume from the same queue, but we only ever need to run one instance of celery beat.

src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_celery_beat.conf.j2
pr33thi: https://github.com/dimagi/commcare-cloud/pull/2039/commits/8ba2051711e33c47529c3205ebb17a403dc3ddfd
That was the only place I could find `celerybeat` in the codebase

pr33thi:
@gcapalbo It currently breaks here:
https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/commands/ansible/helpers.py#L174
I figured hardcoding in a `_0` would be better than putting in a different check for the 1 service that's named differently, but am open to not doing it this way!

gcapalbo:
On Fri, Aug 17, 2018 at 3:15 PM, pr33thi <notifications@github.com> wrote:

> @gcapalbo <https://github.com/gcapalbo> It currently breaks here:
> https://github.com/dimagi/commcare-cloud/blob/master/
> src/commcare_cloud/commands/ansible/helpers.py#L174
> I figured hardcoding in a _0 would be better than putting in a different
> check for the 1 service that's named differently, but am open to not doing
> it this way!
>
I see, in that case I don't feel too strongly either way

pr33thi:
@snopoke updated changelog here: https://github.com/dimagi/commcare-cloud/pull/2039/commits/d932a4bb3fffce358e62f53fb83aba2c2cc04497

emord:
Should flower also be changed? https://github.com/dimagi/commcare-cloud/blob/2080f99a16e97fffff29460c79b07042216f87c9/src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_celery_flower.conf.j2#L1

pr33thi:
@emord `flower` is already handled here: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/commands/ansible/helpers.py#L161

dannyroberts:
@pr33thi I originally intended by https://github.com/dimagi/commcare-cloud/issues/2009 to have just the service command change to know the correct name. I see that the name "celerybeat" is inconsistent with the naming scheme; even flower, the other fake "queue" that doesn't ever have multiple (and so doesn't need the number) is called "celery_flower" (with that "celery_") prefix. I think the best options are

- Change "beat" to do the same thing that flower currently does, resulting in a queue named "...celery_beat" (with the underscore, but without the `_0`)
- or accept a little branching logic to construct the correct name, and then include a comment about the irregular name.

I'm okay with either; the first leaves us in a slightly nicer state, but is more of a pain to implement, and in a world where others are running this themselves, would likely cause problems on their side if they didn't realize they needed to run this upgrade process. Since we're not yet in that world, I'm okay with updating the name, but agree with giovanni that we should do what we do for flower (as you linked) and exclude the `_0`.

pr33thi:
(FYI, that push was just to fix up merge conflicts, this isn't ready to be looked at)

pr33thi:
@snopoke @emord @dannyroberts 
This is ready for review again! I did not rebase since there have been comments on past commits that I didn't want to disappear.

I got rid of the changelog, because I'm now just changing the way the non-standard name is handled, which doesn't really seem worthy of documentation outside of this PR

src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -171,6 +171,8 @@ def get_celery_workers(environment):
 def get_celery_worker_name(environment, comma_separated_queue_name, worker_num):
     environment_environment = environment.meta_config.deploy_env
     project = environment.fab_settings_config.project
+    if comma_separated_queue_name == 'beat':
+        return "{project}-{environment}-celerybeat".format(project=project, environment=environment_environment)
```
snopoke: why is `celerybeat` this better than `celery_beat` (which is what it would be without this if statement)?

pr33thi: The name is defined as `celerybeat` herre: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_celery_beat.conf.j2#L1

And modifying it there would require modifying it on all envs, so I was going with this as a simpler solution.

PR #2040, opened 2018-08-17
Generate new passwords in commcare-cloud-bootstrap

pr33thi:
https://trello.com/c/mL2fpRPS/130-generate-new-passwords-in-commcare-cloud-bootstrap

- Generate passwords instead of hardcoding them
- Get rid of references to touchforms in the vault.yml file

### DIFF
--- a/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
+++ b/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
@@ -384,7 +384,7 @@ def save_vault_yml(environment):
         return str(uuid.uuid4())
     j2.globals.update(generate_uuid=generate_uuid)
     template = j2.get_template('private.yml.j2')
-    vault_file_contents = template.render(my_pw="elephant")
+    vault_file_contents = template.render()
     vault_file = environment.paths.vault_yml
     with open(vault_file, 'w') as f:
         f.write(vault_file_contents)
--- a/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
+++ b/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
@@ -7,6 +7,7 @@ import string
 import subprocess
 import sys
 import shutil
+import uuid
 
 import re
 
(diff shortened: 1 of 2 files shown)


commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -378,6 +379,18 @@ def save_inventory(environment, inventory):
           file=sys.stderr)
 
 
+def save_vault_yml(environment):
+    def generate_uuid():
+        return str(uuid.uuid4())
+    j2.globals.update(generate_uuid=generate_uuid)
+    template = j2.get_template('private.yml.j2')
+    vault_file_contents = template.render(my_pw="elephant")
```
snopoke: what's `my_pw` for? I don't see it referenced in the template.

commcare-cloud-bootstrap/environment/private.yml.j2
```diff
@@ -1,35 +1,33 @@
 secrets:
-  ECRYPTFS_PASSWORD: 'xxBvYiMZ3YVIVkFR'
+  ECRYPTFS_PASSWORD: '{{ generate_uuid() }}'
   KSPLICE_ACTIVATION_KEY:
   DATADOG_API_KEY: null
   DATADOG_APP_KEY: null
   AMQP_USER: "{{ deploy_env }}_worker"
```
snopoke: this will get rendered as `"_worker"` unless you escape the `{{` or you pass `deploy_env` into the jinja context

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
pr33thi: Oops, I was using that for debugging, removed here: https://github.com/dimagi/commcare-cloud/pull/2040/commits/89f1837f78a6f308372f720bc3019df22cb82667

commcare-cloud-bootstrap/environment/private.yml.j2
pr33thi: Didn't notice that, I think it currently doesn't work in master either:
https://github.com/dimagi/commcare-cloud/pull/2040/commits/3a10da86b133453d9a2353d33d332fbc38ff604a

dannyroberts: My reading of the code here is that it _did_ use to work on master, because before this PR, `private.yml` was copied directly into the new env, and then only later during execution, `ansible`/`ansible-playbook` would plug in the value for `{{ deploy_env }}` here. When you change a text file to a .j2 file, you have to start by escaping all the `{{`s, i.e. you have to replace  `{{` with `{{ '{{' }}`.

That said, I think the solution you landed on will totally work—it's just plugging in the value now so it's hardcoded in the resultant `vault.yml` file inside the env—and is more readable. It results in something different landing in the env `vault.yml` file, but I'd say the change is an improvement.

pr33thi: got it

PR #2041, opened 2018-08-20
drop supervisor logs kept to 5

snopoke:
Should help swiss disks which have 14.7GB of logs. Also don't think we use these much.

### DIFF
--- a/src/commcare_cloud/ansible/roles/supervisor/templates/supervisord.conf.j2
+++ b/src/commcare_cloud/ansible/roles/supervisor/templates/supervisord.conf.j2
@@ -21,7 +21,7 @@ password={{ supervisor_http_password }}               ; (default is no password 
 [supervisord]
 logfile=/tmp/supervisord.log ; (main log file;default $CWD/supervisord.log)
 logfile_maxbytes=20MB        ; (max main logfile bytes b4 rotation;default 50MB)
-logfile_backups=10           ; (num of main logfile rotation backups;default 10)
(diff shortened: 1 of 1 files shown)


docs/changelog/0004-support-multiple-kafak-brokers.md
```diff
@@ -0,0 +1,29 @@
+# 4. Support multiple Kafka borkers
```
snopoke: oops, this wasn't meant to be in this PR.

src/commcare_cloud/ansible/roles/supervisor/templates/supervisord.conf.j2
```diff
@@ -21,7 +21,7 @@ password={{ supervisor_http_password }}               ; (default is no password
 [supervisord]
 logfile=/tmp/supervisord.log ; (main log file;default $CWD/supervisord.log)
 logfile_maxbytes=20MB        ; (max main logfile bytes b4 rotation;default 50MB)
-logfile_backups=10           ; (num of main logfile rotation backups;default 10)
+logfile_backups=5            ; (num of main logfile rotation backups;default 10)
```
emord: fyi This is going to restart supervisor, and I've had issues with that leaving zombie processes around before.

snopoke: The other option is to make it configurable per env. I've already rolled this out to swiss - but that also included a supervisor upgrade since that hadn't been rolled out yet.

snopoke:
yes

PR #2042, opened 2018-08-20
Upgrade to couchdb 1.7.2

emord:
Add this to the list of things shouldn't be done in a production instance. Will send out a more full retro for why I did it.

@dimagi/devops This is necessary because the package for 1.6.1 is no longer around.


emord:
retro sent

nickpell:
Are there any implications for ECHIS?  I'm curious how this change impacts a full deploy.

emord:
@nickpell They're on couchdb 2. Swiss is the only instance where couch 1.x is still running

PR #2044, opened 2018-08-21
CSV inventory plugin

snopoke:
See docs and examples in diff for how it works and expected format.

Also see here for another example: https://docs.google.com/spreadsheets/d/1WlamplK6thSqbPt2fMWsoHkPwOan9kdzN63uwu_G7Zo/edit#gid=403315111

### DIFF
--- a/src/commcare_cloud/ansible/ansible.cfg
+++ b/src/commcare_cloud/ansible/ansible.cfg
@@ -11,5 +11,10 @@ fact_caching_timeout = 21600
 
 timeout = 30
 
+inventory_plugins = ./plugins/inventory
+
 [ssh_connection]
 pipelining = True
+
+[inventory]
+enable_plugins = ini, csv
--- a/src/commcare_cloud/ansible/plugins/inventory/csv.py
+++ b/src/commcare_cloud/ansible/plugins/inventory/csv.py
@@ -8,7 +8,6 @@ __metaclass__ = type
 
 DOCUMENTATION = '''
     inventory: csv
-    version_added: "2.4"
     short_description: Uses CSV files and inventory source.
     description: |
         - CSV file based inventory, lines are split into groups by a blank line.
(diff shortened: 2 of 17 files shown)


src/commcare_cloud/ansible/plugins/inventory/csv.py
```diff
@@ -0,0 +1,195 @@
...
+                    row_groups.append(current_group)
+                    current_group = []
+            else:
+                current_group.append(line)
+        if current_group:
+            row_groups.append(current_group)
+        return row_groups
+
+
+def conv_str2value(item_type, item):
```
snopoke: The `ini` parser uses [ast.literal_eval](https://docs.python.org/2/library/ast.html#ast.literal_eval) to parse var values - I could switch to that but thought that being explicit was better. Open to opinions.

src/commcare_cloud/ansible/plugins/inventory/csv.py
```diff
@@ -0,0 +1,195 @@
...
+                item_type, name = key.split('.') if '.' in key else ('S', key)
+                name = name.split(' ')[1]
+                vars[name] = conv_str2value(item_type, val)
+        return vars
+
+    def _parse_groups(self, rows):
+        for row in rows:
+            group = row['group']
+            self.inventory.add_group(group)
+            var_name, type_, raw_val = row['var'], row['type'], row['val']
```
calellowitz: nit: you use `item_type` above to refer to the same thing (which i generally like better than the trailing underscore). consistency would be nice.

calellowitz:
That was supposed to be an approval not a change request

emord:
I think the tests are failing due to docs not being made?

src/commcare_cloud/ansible/ansible.cfg
```diff
@@ -11,5 +11,10 @@ fact_caching_timeout = 21600
 
 timeout = 30
 
+inventory_plugins = ./plugins/inventory
```
emord: For me this only works if i'm already in the `src/commcare_cloud/ansible` directory. I was trying to run `cchq icds-cas x` but this plugin would not be found

emord:
fyi I added a commit to check for the new inventory paths

src/commcare_cloud/ansible/ansible.cfg
snopoke: odd, it should be relative to this file (same as callback plugins)

snopoke: I this should fix it: https://github.com/dimagi/commcare-cloud/pull/2050/files#diff-9af49c7720ecc05ddbc9456b1c90b73bR10

PR #2048, opened 2018-08-21
don't require specifying files list

snopoke:
(on top of https://github.com/dimagi/commcare-cloud/pull/2047)


src/commcare_cloud/commands/migrations/copy_files.py
```diff
@@ -149,12 +149,15 @@ def prepare_file_copy_scripts(target_host, soucre_file_configs, script_root):
     files_for_node = []
     for config in soucre_file_configs:
```
pr33thi: [nit] `soucre_file_configs` > `source_file_configs`

PR #2050, opened 2018-08-21
icds cas deploy

emord:
This is based off of sk/icds-cloud & sk/csv-inventory. This PR is going into sk/icds-cloud

The additions made here are:

1. removing SMS from app processes check because we will not want to run that until we go live
2. removing the app processes check because we do not have all machines yet which causes some translations to fail
3. removing public_ip from ALLOWED_HOSTS because that does not exist. (I also had trouble creating tha variable in the new CSV format, but may have just done it incorrectly)
4. defined the remote user as wcdadmin as that's the user given to us
5. Updated the lvm config for couch0. Note that I'm not positive this is correct, so it would be great for some on @dimagi/devops to check this


snopoke:
merged master and updated PR to merge into `master`

environments/icds-cas/public.yml
```diff
@@ -124,7 +124,7 @@ localsettings:
     - 127.0.0.1
     - '{{ SITE_HOST }}'
     - '{{ CAS_SITE_HOST }}'
-    - '{{ hostvars[groups.proxy.0].public_ip }}'
+      # - '{{ hostvars[groups.proxy.0].public_ip }}' TODO: hostvars seems to not work?
```
snopoke: I don't think any hosts have this var set

emord: erg. I realize what I did now after seeing https://github.com/dimagi/commcare-cloud/pull/2050/commits/6da8fe85ba921e3b136569490f6da1ed2bd6bd35

I didn't fully add the comma to all lines... totally missed htat yesterday. thanks for fixing

src/commcare_cloud/commcare_cloud.py
```diff
@@ -2,9 +2,14 @@
 from __future__ import print_function
 from __future__ import absolute_import
 
-import inspect
 import os
 
+# set ansible config before other imports so that it's present when `ansible.constants` is imported
+if 'ANSIBLE_CONFIG' not in os.environ:
+    ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
+    os.environ['ANSIBLE_CONFIG'] = os.path.join(ROOT_DIR, 'ansible/ansible.cfg')
```
dannyroberts: I think it's a bit cleaner (but essentially the same) to do it as is done in this line: https://github.com/dimagi/commcare-cloud/blob/72013de466e0ed8328e7b2deed5929c0e6f68cb7/src/commcare_cloud/commands/ansible/helpers.py#L31

I think additionally, if it's done here, then it probably doesn't need to be done there, since it uses the current `os.environ` as its starting point. 

Another pattern we could follow is to put this in another file and then import it like `from commcare_cloud.patch_environ import patch_environ; patch_environ()`. (In loose analogy to e.g. `from gevent import monkey; monkey.patch_all()`.) I think it's a bit nicer, but I don't feel super strongly about it and won't feel bad if you decide not to go with it.

dannyroberts: Oh, and I realize the line I suggested going off of depends on an import (ANSIBLE_DIR) whereas yours doesn't. I didn't write this down anywhere but I've pretty intentionally made `commcare_cloud.paths` import next to nothing so that it's pretty much always safe to import without circular dependencies or (in this case) having to think about import order and import side effects.

environments/icds-cas/inventory/postgresql.csv
```diff
@@ -0,0 +1,54 @@
...
+pgshard0standby,100.71.184.47,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS00,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard0,,standby0
+pgshard1standby,100.71.184.51,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS01,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard1,,standby0
+pgshard2standby,100.71.184.57,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS02,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard2,,standby0
+pgshard3standby,100.71.184.49,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS03,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard3,,standby0
+pgshard4standby,100.71.184.60,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS04,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard4,,standby0
+pgshard5standby,100.71.184.52,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS05,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard5,,standby0
+pgshard6standby,100.71.184.58,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS06,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard6,,standby0
+pgshard7standby,100.71.184.61,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS07,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard7,,standby0
+pgshard8tandby,100.71.184.56,pg_standby,,shard_standby_dbs,MUMGCCWCDPRDSS08,"[""/dev/vdb""]","[""/dev/vdb1""]",,pgshard8,,standby0
```
emord: This is missing an s in standby. left it alone for now, but we may want to change it

environments/icds-cas/public.yml
```diff
@@ -1,5 +1,5 @@
 internal_domain_name: 'internal-icds-cloud.commcarehq.org'
-cluster_ip_range: '10.247.164.0/23'  # TODO: cluster ip range
+cluster_ip_range: '100.71.184.0/22'  # TODO: cluster ip range
```
snopoke: this isn't actually correct since we have 3 ranges

src/commcare_cloud/commcare_cloud.py
snopoke: Nice, that's definitely cleaner: https://github.com/dimagi/commcare-cloud/pull/2056

environments/icds-cas/postgresql.yml
```diff
@@ -0,0 +1,113 @@
+DEFAULT_POSTGRESQL_HOST: pgmain0
+REPORTING_DATABASES:
+  ucr: ucr
+  icds-ucr:
+    WRITE: icds-ucr
+    READ:
+      - [icds-ucr, 1]
```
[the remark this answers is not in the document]
snopoke: I don't want to remove the main UCR DB from reads completely since our query code will fall back to it if the replication delay on the standby's gets too high but I've increased the weighting of the standbys: 3f3330a

PR #2053, opened 2018-08-22
BROKER_URL -> CELERY_BROKER_URL

nickpell:
Must deploy localsettings changes before [upgrading to celery 4.x](https://github.com/dimagi/commcare-hq/pull/21588)


emord:
Does 3.1.25 support CELERY_BROKER_URL? Should we have something that says: `BROKER_URL = CELERY_BROKER_URL` for the short term/make reverts easy?

nickpell:
I like this approach, since future cleanup is very straightforward and it doesn't complicate the already-complicated HQ PR.

PR #2054, opened 2018-08-22
enable password obfuscation on softlayer

mkangia:
Enabling password obfuscation on india server as well since ICDS master app is on softlayer for which [we have enabled obfuscation](https://github.com/dimagi/commcare-cloud/pull/2051) and is blocking using of app on india server.

### DIFF
--- a/environments/softlayer/public.yml
+++ b/environments/softlayer/public.yml
@@ -109,6 +109,7 @@ localsettings:
   TRANSFER_SERVER: 'nginx'
   WS4REDIS_CONNECTION_HOST: "{{ groups.redis.0 }}"
   XFORMS_PLAYER_URL: "http://{{ groups.touchforms.0 }}:4444/"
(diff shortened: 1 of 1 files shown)


environments/softlayer/public.yml
```diff
@@ -109,6 +109,7 @@ localsettings:
   TRANSFER_SERVER: 'nginx'
   WS4REDIS_CONNECTION_HOST: "{{ groups.redis.0 }}"
   XFORMS_PLAYER_URL: "http://{{ groups.touchforms.0 }}:4444/"
+  OBFUSCATE_PASSWORD_FOR_NIC_COMPLIANCE: True
```
millerdev: Seems like this should be a domain-level feature flag since it introduces a fair amount of additional complexity for authentication, and any bugs could affect domains that do not need or want this.

PR #2056, opened 2018-08-23
patch environment with ansible.cfg location

snopoke:
See discussion here: https://github.com/dimagi/commcare-cloud/pull/2050#discussion_r212009107

### DIFF
--- a/src/commcare_cloud/patch_environ.py
+++ b/src/commcare_cloud/patch_environ.py
@@ -1,0 +1,8 @@
+import os
+
+
+def patch_environ():
+    if 'ANSIBLE_CONFIG' not in os.environ:
+        # this must happen before other imports so that it's present when `ansible.constants` is imported
+        from commcare_cloud.environment.paths import ANSIBLE_DIR
+        os.environ['ANSIBLE_CONFIG'] = os.path.join(ANSIBLE_DIR, 'ansible.cfg')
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/patch_environ.py
```diff
@@ -0,0 +1,19 @@
...
+
+def patch_environ():
+    if 'ANSIBLE_CONFIG' not in os.environ:
+        from commcare_cloud.environment.paths import ANSIBLE_DIR
+        constants_module = 'ansible.constants'
+        if constants_module in sys.modules:
+            puts(colored.red(
+                "\nSettings in 'ansible.cfg' have not been applied. "
+                "'ANSIBLE_CONFIG' environment variable must be set before the '{}' module is imported.\n"
+            ).format(constants_module))
```
dannyroberts: Is this essentially to check the import order, i.e. to check that this function is being run before `ansible.constants` is imported? If so good thinking, I'm sure we'll be glad to have that. In addition, we could make the tests (and really any run of `commcare-cloud`) fail by adding `exit(-1)` or something here, so that an accidental change breaking this could never make it to master, even if it wasn't tested locally.

snopoke: That would be great. I had to move where this get's called to `commcare_cloud/__init__.py`. I also removed the `ANSIBLE_CONFIG` var in `.travis.yml`.

PR #2057, opened 2018-08-23
full support for separate envs in copy files command

snopoke:


src/commcare_cloud/commands/migrations/copy_files.py
```diff
@@ -253,19 +268,71 @@ def teardown_auth(plan, environment, ansible_context):
...
+    run_ansible_module(env, ansible_context, host, 'fetch', fetch_args, True, None, None)
+
+
+def _set_auth_key(env, host, user, ansible_context, remove=False):
+    state = 'absent' if remove else 'present'
+    args = "user={} state={} key={{{{ lookup('file', '{}') }}}}".format(user, state, ID_RSA_TMP)
+    run_ansible_module(env, ansible_context, host, 'authorized_key', args, True, None, None)
+
+
+class PrivilegedCommand():
```
snopoke: will replace this with the one from https://github.com/dimagi/commcare-cloud/pull/1972 once it's merged

src/commcare_cloud/commands/migrations/copy_files.py
```diff
@@ -253,19 +268,71 @@ def teardown_auth(plan, environment, ansible_context):
...
-        run_ansible_playbook(
-            environment, 'setup_bidirectional_auth.yml', ansible_context, skip_check=True,
-            unknown_args=[
-                '-e', 'hostA={}'.format(target_host),
-                '-e', 'hostB={}'.format(source_host),
-                '-e', 'usernameA=root',
-                '-e', 'usernameB={}'.format(source_user),
-                '-e', 'action={}'.format(action),
-                '-e', 'bidirectional=False',
-            ])
```
snopoke: had to split this up since some tasks could be running on different envs

PR #2058, opened 2018-08-23
run commands for celery 4.x

nickpell:
To update, run ```cchq <env> update-config``` and ```cchq <env> update-supervisor-confs```.

Prior to merging, deploy https://github.com/dimagi/commcare-hq/pull/23319 to all environments:
- [x] staging

- [x] production

- [x] softlayer

- [x] swiss

- [x] pna

- [x] ICDS

Must be deployed before https://github.com/dimagi/commcare-hq/pull/21588


src/commcare_cloud/ansible/roles/commcarehq/templates/celery_bash_runner.sh.j2
```diff
@@ -44,7 +44,8 @@ HOSTNAME_ARG+=".${TIMESTAMP}_timestamp"
 HOSTNAME_PARTS=(${HOSTNAME_ARG//=/ })
 HOSTNAME=${HOSTNAME_PARTS[1]}
 
-{{ virtualenv_home }}/bin/python {{ item.get('python_options', '') }} {{ code_home }}/manage.py celery worker ${HOSTNAME_ARG} ${ARGS}  &
+source {{ virtualenv_home }}/bin/activate
```
emord: Why is this being sourced? I'm hesitant to make this change if it's not necessary, because I believe it would affect the entire script/shell. Before when it's not sourced, it should only affect the command (same for supervisor changes)

nickpell: Thanks, updated

nickpell:
Tested on staging, except for celery beat.  That's going to have to be tested in a production environment, but I've validated that the command works locally.

gcapalbo:
@nickpell fyi i think this might be what's causing https://github.com/dimagi/commcare-hq/pull/21712#issuecomment-418440452

nickpell:
@gcapalbo Thanks for pointing that out - I was able to reproduce the error in a toy-example by trying to schedule a task within a task.  I haven't gotten to the bottom of why that's happening though.

dannyroberts:
This and https://github.com/dimagi/commcare-hq/pull/21588 look like good changes, but looks like maybe they got stuck when some issues came up? What's the status?

nickpell:
@dannyroberts I got blocked because starting a task from within a task would fail on staging.  I believe there is some bootstrapping missing from the new setup but I never got to the bottom of the problem.  I think it's something that I could solve in an afternoon with a second pair of eyes.

src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -398,6 +391,13 @@ CACHES = {
     'redis': redis_cache,
 }
 
+CELERY_BROKER_URL = BROKER_URL = '{{ BROKER_URL }}'
+CELERY_RESULT_BACKEND = redis_cache['LOCATION']
```
nickpell: @dannyroberts Figured now would be the right time to switch over to using redis as the results backend, so we don't have to change the backend more than once.  I've tested that this works locally and on staging.

nickpell: If this looks like a reasonable change to you, then I think I will break this PR into two pieces with two changelogs - one for localsettings changes (no CCHQ min version, max version prior to celery 4.x upgrade, brief service interruption when backend changes - to be performed on weekend), and one for bash script changes (min version: https://github.com/dimagi/commcare-hq/pull/23319, max version prior to celery 4.x upgrade, no service interruption - can run anytime).

nickpell: separate PRs are so the localsettings changes don't get run accidentally during peak hours, and so I can run the bash script updates sooner to avoid making too many changes at once

nickpell:
Closing in favor of https://github.com/dimagi/commcare-cloud/pull/2680 and https://github.com/dimagi/commcare-cloud/pull/2693

PR #2059, opened 2018-08-23
Rewrite local_pillows config

emord:
This should make the pillow config diffs we deploy for ICDS less flexible, but much more readable

### DIFF
--- a/.travis/environments/travis/public.yml
+++ b/.travis/environments/travis/public.yml
@@ -93,11 +93,13 @@ localsettings:
         class: 'corehq.apps.userreports.pillow.ConfigurableReportKafkaPillow'
         instance: 'corehq.apps.userreports.pillow.get_kafka_ucr_static_pillow'
         params:
+          topics: ['case-sql']
           ucr_division: '09'
       - name: 'kafka-ucr-main-custom-2'
         class: 'corehq.apps.userreports.pillow.ConfigurableReportKafkaPillow'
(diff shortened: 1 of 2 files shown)


sravfeyn:
Curious why do this, since the readability is available in the `public.yml` and the final localsettings.py?

emord:
@sravfeyn This is for readability in the diff when you deploy. Currently it looks like:

```
- LOCAL_PILLOWTOPS = json.loads('reallly long string no one can read')
+LOCAL_PILLOWTOPS = json.loads('another really long string no one can read')
```

This will make that diff similar to what we see in the yaml file

.travis/environments/travis/public.yml
```diff
@@ -93,12 +93,16 @@ localsettings:
...
         params:
-          ucr_division: '09'
+          topics: ['case-sql']
+          include_ucrs:
+            - 'ucr2'
       - name: 'kafka-ucr-main-custom-2'
         class: 'corehq.apps.userreports.pillow.ConfigurableReportKafkaPillow'
         instance: 'corehq.apps.userreports.pillow.get_kafka_ucr_static_pillow'
         params:
-          ucr_division: 'af'
```
snopoke: is this no longer used?

emord: This was only used as a stop gap when we got really behind on UCRs. it's still supported in the pillow, but we could probably remove it. It's not super useful anymore since we have a way to include/exclude specific UCR (for ICDS) and partitioning so we can split up pillows that way

PR #2062, opened 2018-08-24
icds-cas pillow & celery configurations

emord:
@dimagi/scale-team 

Being fairly conservative here. We should still have plenty of RAM and CPU after this. main changes are form partitions 16 -> 24 and case partitions 36 -> 60. 

Going to start a doc about how I think scaling to 250k works with some current code changes that are being worked on.

### DIFF
--- a/environments/icds-cas/app-processes.yml
+++ b/environments/icds-cas/app-processes.yml
@@ -191,6 +191,39 @@ celery_processes:
       concurrency: 4
 pillows:
   'pillow0':
+    # Catch all server for low volume pillows
+    AppDbChangeFeedPillow:
+      num_processes: 1
+    ApplicationToElasticsearchPillow:
+      num_processes: 1
+    CacheInvalidatePillow:
+      num_processes: 1
+    DefaultChangeFeedPillow:
+      num_processes: 1
+    DomainDbKafkaPillow:
+      num_processes: 1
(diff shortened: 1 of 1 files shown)


environments/icds-cas/app-processes.yml
```diff
@@ -44,7 +44,7 @@ celery_processes:
   'celery1':
     ucr_queue:
       concurrency: 4
-      max_tasks_per_child: 5
+      max_tasks_per_child: 1
```
snopoke: why this change?

emord:
doc as mentioned: https://docs.google.com/document/d/1ET9Gm24CuuYlLvuWhcrhyjkHlKp6EY-nsgLui6Oww1Y/edit#

environments/icds-cas/app-processes.yml
emord: It's not a necessary change, but left a comment about why in the commit https://github.com/dimagi/commcare-cloud/pull/2062/commits/1c7ee7d78407bea740303610bd0fe2e4a635b858

> There isn't a lot of activity on these queues and it's likely a high risk for memory leaks due to loading many documents

snopoke:
@emord I'll let you roll out these changes

emord:
already deployed

PR #2064, opened 2018-08-28
Make test_autogen.docs.sh fail loudly if any part of it fails

dannyroberts:
currently if running `make` fails, it will only fail later incidentally
because the previous `make clean` created a diff. Now any failing command in this script
will bubble up directly


dannyroberts:
Adding you @emord because of your helpful work reviewing similar PRs on commcare-hq recently.

PR #2066, opened 2018-08-28
ubuntu 18.04 changes

snopoke:
Started looking at this in some downtime that I had.

These changes get us through `deploy_common`

### DIFF
--- a/src/commcare_cloud/ansible/roles/common_installs/tasks/bionic_apt_installs.yml
+++ b/src/commcare_cloud/ansible/roles/common_installs/tasks/bionic_apt_installs.yml
@@ -1,0 +1,8 @@
+- name: Install common packages (bionic)
+  apt: name="{{ item }}" state=present
+  become: yes
+  with_items:
+    - software-properties-common
+    - npm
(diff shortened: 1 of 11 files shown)


src/commcare_cloud/ansible/roles/common_installs/tasks/bionic_apt_installs.yml
```diff
@@ -0,0 +1,8 @@
+- name: Install common packages (bionic)
+  apt: name="{{ item }}" state=present
+  become: yes
+  with_items:
+    - software-properties-common
+    - npm
+    - python-minimal
+    - python3-pip
```
emord: why do we need to install python 3's pip?

snopoke: ansible is using python3 on the nodes since that's what 18.04 comes with standard and so if we need any libs for ansible we need pip3 to install them. For example ansible needs `psycopg2` installed in order to do PG tasks like create users.

PR #2067, opened 2018-08-28
Allow ssh key by user

emord:
@dimagi/scale-team


snopoke:
Does this file change much between versions of OpenSSH? Guessing not really
so then I'm OK with it.

The other nice thing about the other way is it's easy to see where we're
deviating from the defaults. To solve that you could make vars for the
variables we change to show they're non-defaults. Don't feel strongly about
that though.

emord:
Quickly seeing how we differ from defaults is something I hadn't thought about, and it does seem useful. Going to keep this PR in this state since it's already implemented, but I'll look into using `blockinfile`/this pattern for future work

PR #2069, opened 2018-08-29
Use monit for some services

emord:
Haven't fully tested this yet, but I think this bit us during the CSP migration dry run today. 

Currently if you use `service <service> stop` to stop a service it will come back within a minute because monit will restart it. This PR should change that so that the service will not come back until you explicitly start it.

### DIFF
--- a/src/commcare_cloud/commands/ansible/service.py
+++ b/src/commcare_cloud/commands/ansible/service.py
@@ -30,6 +30,8 @@ STATES = {
     'status': 'status'
 }
 
+MONIT_MANAGED_SERVICES = ['postgresql', 'pgbouncer', 'redis', 'couchdb2']
+
 
 @attr.s
 class ServiceOption(object):
@@ -181,7 +183,21 @@ class SupervisorService(SubServicesMixin, ServiceBase):
         :param process_pattern: process pattern from the args or None
         :return: dict mapping tuple(hostname1,hostname2,...) -> [process name list]
         """
-        raise NotImplemented
+        raise NotImplementedError
+
+
+def _service_status_helper(service_name):
+    if service_name in MONIT_MANAGED_SERVICES:
+        return 'monit status {}'.format(service_name)
+
+    return 'service {} status'.format(service_name)
+
+
+def _ansible_module_helper(service_name):
+    if service_name in MONIT_MANAGED_SERVICES:
+        return 'monit'
+
+    return 'service'
 
 
 class AnsibleService(ServiceBase):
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/ansible/service.py
```diff
@@ -245,7 +261,7 @@ def execute_action(self, action, host_pattern=None, process_pattern=None):
 
         def _change_state(service_name, run_on, action=action):
             service_args = 'name={} state={}'.format(service_name, STATES[action])
-            return self._run_ansible_module(run_on, 'service', service_args)
+            return self._run_ansible_module(run_on, _ansible_module_helper(service_name), service_args)
```
dannyroberts: i see this is kind of clever. Basically this works because the `monit` ansible module and the `service` ansible module have similar interfaces. It's unlikely to cause us problems, but I might have preferred minor code duplication to treating service and monit modules as an interchangeable interface, which only happens to work here. Something like this

```
    if service_name in MONIT_MANAGED_SERVICES:
            service_args = 'name={} state={}'.format(service_name, STATES[action])
            return self._run_ansible_module(run_on, 'monit', service_args)
    else:
            service_args = 'name={} state={}'.format(service_name, STATES[action])
            return self._run_ansible_module(run_on, 'service', service_args)
```
in the helper function would better reflect that these are two different module calls with potentially different interfaces. (For example if we need to pass a special argument to monit to change the behavior in some way, we can then just edit that one.) Anyway, not a big deal/necessarily actionable, just trying to articulate what felt funny to me about this pattern.

dannyroberts:
yeah this is great

src/commcare_cloud/commands/ansible/service.py
emord: Yeah. I'm not married to the approach I took. I was hoping to make the `monit` services a subclass of `AnsibleService` but the multi service configurations made that a bit more difficult because couch has one service monitored by monit and another by service. Changing service configurations to fix that seemed like a larger change that wasn't worth it which is why I went the way I did.

Your approach seems fine as well. I haven't had to maintain/change this code before so I'd default to your judgement on the way to write it. Feel free to change it as you see fit

snopoke:
I've noticed a few issues with this:
* the monit commands always have `0` return code (unless there's an error)
  * when running `start` if the service didn't start you won't be able to tell
  * when running `status` you have to read the output to tell if the service is running instead of being able to look at the success / failure of the ansible call which is much easier since success is green and failure is red
* `monit status X` shows the status for all services managed by monit on the host and is quite verbose

I think the approach we should take is to issue a separate command to activate / deactivate the service in monit

PR #2072, opened 2018-08-29
Added hostname in /etc/host and change in datadog.conf

NitigyaS:


src/commcare_cloud/ansible/roles/common/tasks/main.yml
```diff
@@ -60,7 +60,16 @@
   become: yes
   hostname:
     name: "{{ hostname }}"
```
snopoke: could change this to `{{ host_alias|default(hostname) }}`

src/commcare_cloud/ansible/roles/datadog/templates/datadog.conf.j2
```diff
@@ -38,3 +38,7 @@ dogstreams: {% for parser in datadog_parsers -%}
     {%- endif -%}
 {%- endfor %}
 {% endif %}
+
+{% if host_alias is defined %}
+hostname: {{ hostname }}.{{ internal_domain_name }}
```
snopoke: missed this but it should be `host_alias` here not `hostname`

snopoke: https://github.com/dimagi/commcare-cloud/pull/2077

PR #2073, opened 2018-08-29
cleanup replication slots

snopoke:
@dimagi/scale-team FYI


src/commcare_cloud/ansible/ansible.cfg
```diff
@@ -12,6 +12,7 @@ fact_caching_timeout = 21600
 timeout = 30
 
 inventory_plugins = ./plugins/inventory
+library = ./library
```
snopoke: This change is what caused the updates to the docs

dannyroberts:
That should do it. Not the prettiest, but tell me what you think

PR #2075, opened 2018-08-29
Autogenerate changelog entries

dannyroberts:
All changes to `docs/changelog/*.md` (besides removing `0000-changelog.md`) are because due to minor formatting inconsistencies between these .md files and thus unavoidable. Changes include:

- `docs/changelog/0*.md` are now built from `changelog/0*.yml`
- `docs/changelog/index.md` are also now built directly from the yaml files
- As a result, all files in `docs/changelog` are now autogenerated, and `make clean` results in that dir being empty
- the rebuild semantics are basically correct (it constructs the list of md files it expects based on the `changelog/*.yml` files, so running `make` only rebuilds the ones that have changed)
- changelog yaml format is validated against a schema


src/commcare_cloud/manage_commcare_cloud/changelog.md.j2
```diff
@@ -1,18 +1,24 @@
...
+{% if changelog_entry.max_commcare_version -%}
+## CommCare Version Dependency
+CommCare versions beyond this commit require this change to function correctly:
+[{{ changelog_entry.max_commcare_version[:8] }}](https://github.com/dimagi/commcare-hq/commit/{{ changelog_entry.max_commcare_version }})
+{% endif %}
 
-## Changelog
-{%- for changelog_entry in changelog_contents %}
+{%- if changelog_entry.min_commcare_version -%}
+## CommCare Version Dependency
```
dannyroberts: nit (i think): If both max and min are defined, you'd get this same heading twice. I think I'd rather Always have this heading, and then have a phrase like "This change is not known to be dependent on any particular version of CommCare HQ."

dannyroberts:
Thanks for the review and improvements on dependent commcare versions. I'll probably merge once the tests pass, unless you want to comment on my last change. I expect this code to have a tail, so followup PRs welcome.

src/commcare_cloud/manage_commcare_cloud/changelog.md.j2
snopoke: Never mind - I see you already did that.

snopoke:
:candy: :camel:

PR #2077, opened 2018-08-30
alt_hostname support

snopoke:
This allows us to use an alternate hostname to the one
that is the primary hostname of the VM for internal purposes
such as email, Datadog etc.

This need for this arose because the hosting provider required
specific hostnames which were not easily human readable.


environments/icds-cas/public.yml
```diff
@@ -154,7 +154,7 @@ localsettings:
   COUCH_CACHE_VIEWS: True
   COUCH_STALE_QUERY: 'update_after'
   DAYS_TO_KEEP_DEVICE_LOGS: 45
-  DEPLOY_MACHINE_NAME: "{{ ansible_hostname }}"
+  DEPLOY_MACHINE_NAME: "{{ alt_hostname|default(ansible_hostname) }}"
```
NitigyaS: Maybe I missed, But I do not see where `alt_hostname` is defined.

snopoke: It's in the CSV inventory plugin: https://github.com/dimagi/commcare-cloud/pull/2077/files#diff-a4ffb19691318917608e6abbff958e45R164

snopoke: A bit sneaky putting it there - we could also just define a column in the inventory file and add it there but it would basically duplicate the first column

NitigyaS: Thanks.

PR #2081, opened 2018-08-30
Update riak keys

emord:
Had to update the riak keys because I recreated the data directories after the aborted migration

@dimagi/scale-team


emord:
errrrggg. will need to regenerate again. copied wrong things in vualt file....

PR #2085, opened 2018-08-31
clearer delineation of commands

snopoke:
@dannyroberts what do you think about this to make it easier to differentiate between command content.

Other thoughts I had were to have one page per command instead of one page for all.

buddy @Rohit25negi


dannyroberts:
Looks pretty reasonable!

I think as long as the index is still searchable with browser's "find in page", we could have one page per command. Don't have a strong sense of whether it'd be an improved experience or not.

PR #2086, opened 2018-08-31
Move terraform code into commcare-cloud

dannyroberts:
This commit just subtree merges https://github.com/dimagi/g2-iac-commcarehq. I'm working on another branch where I refactor this to be many thousands of lines shorter but deduplicating copy-pasted code. I mostly wanted to PR this separately
- so that you can see and review the actual changes I'm making in the other branch separately
- so that you have an opportunity to weigh in on the decision to move this stuff here.

Open question is still how to plug this into our current convention of keeping all environment specific things in environments. I'm imagining we'll eventually put the environment specific code in the individual environments dirs, but it's not currently set up for that.


dannyroberts:
(We talked about this in the devops meeting where it was verbally approved)

PR #2088, opened 2018-08-31
Revert "Use hubspot sandbox credentials on staging"

orangejenny:
Reverts dimagi/commcare-cloud#2079

Using the generic credentials causes an obnoxious "Sign up for email updates" pop up to show up on staging during signup. Let's remove them, now that testing for https://github.com/dimagi/commcare-hq/pull/21694 is done.

@gcapalbo @jmtroth0


orangejenny:
updated localsettings on staging

PR #2092, opened 2018-09-02
New riak keys

emord:
@dimagi/scale-team we're going to be so good at setting up new riak clusters after this


snopoke:
superseeded by https://github.com/dimagi/commcare-cloud/pull/2093

PR #2096, opened 2018-09-03
Fix production amazon s3 keys

dannyroberts:
RIAKCS_S3_ACCESS_KEY and RIAKCS_S3_SECRET_KEY were improperly set to local riak cs credentials
when they should be the credentials to amazon s3


dannyroberts:
Best hypothesis at this point is they got mis-set in https://github.com/dimagi/commcare-cloud/commit/998c03f67a452d60e287d22283d3e809aa93a71b, but I haven't tried to get a readable from that commit to verify.

dannyroberts:
I only ran it with --check to make sure this got rid of the undesired diff; I haven't applied anything

snopoke:
hmm, very confused by this. must have been a late night change. Sorry

PR #2099, opened 2018-09-04
copy key to working dir instead of /tmp

snopoke:


sravfeyn:
Wonder how it worked on riak when emord did it yesterday.

snopoke:
As I mentioned on slack, the file is owned by someone else which means you can't overwrite it:
```
(icds) skelly@patch0:~$ ls -la /tmp/id_rsa.tmp 
-rw-rw-r-- 1 nsharma nsharma 408 Sep  1 04:49 /tmp/id_rsa.tmp
``

PR #2113, opened 2018-09-05
Apply security updates for python crypto libraries

dannyroberts:
cryptography: https://nvd.nist.gov/vuln/detail/CVE-2018-10903
pycryptodome: https://nvd.nist.gov/vuln/detail/CVE-2018-15560

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ cffi==1.11.5              # via bcrypt, cryptography, pynacl
 chardet==3.0.4            # via requests
 clint==0.5.1
 couchdb-cluster-admin==0.4.2
-cryptography==2.2.2
+cryptography==2.3.1
 datadog==0.2.0
 decorator==4.3.0          # via datadog
 dimagi-memoized==1.1.0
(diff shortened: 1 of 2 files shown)


dannyroberts:
@emord This repo was the pioneer! Here's the original PR from mid-July https://github.com/dimagi/commcare-cloud/pull/1926, and we've come a long way since: https://github.com/dimagi/commcare-cloud/blob/299c62d2d173ec5f9e6002749998b5ebe521cddb/Makefile#L42.

To preempt a reasonable next question, `pip-compile` supports both the `requirements.in` format we use in commcare-hq as well as the `setup.py` format we use here. While the original Django Boston talk I took the pattern from suggested using a `requirements.in` and then autogenerating `setup.py` `install_requires` from it, this (using compiling from `setup.py` directly) seemed simpler to me, and so far it's been working well.

emord:
> pip-compile supports both the requirements.in format we use in commcare-hq as well as the setup.py

Thanks. That's the part i was missing

PR #2114, opened 2018-09-05
Add staging-aws-test environment

dannyroberts:
I think though eventually I'm going to merge `staging-aws-test` back into just `staging`, I wanted to incorporate these changes into master before continuing, since it's already a fair amount. It's probably easiest to do commit-by-commit.

`staging-aws-test` is a live environment (currently at https://34.196.111.115/) we stood up to vet the AWS setup but that we'll probably spin down and redo piecewise for the real staging migration.


environments/staging-aws-test/public.yml
```diff
@@ -0,0 +1,93 @@
+riak_backend: "leveldb"
+riak_ring_size: 64
+
+DATADOG_ENABLED: False
+elasticsearch_endpoint: '{{ groups.elasticsearch.0 }}'
+elasticsearch_cluster_name: 'staginges'
+elasticsearch_version: 1.7.6
+elasticsearch_download_sha256: 78affc30353730ec245dad1f17de242a4ad12cf808eaa87dd878e1ca10ed77df.
```
Rohit25negi: nit: is the dot at the end is intended or a typo?

dannyroberts: Interesting! I hadn't noticed this at all (the whole file was just copied from staging), but searching for the same variable on other envs, I'm seeing that a lot of them have a trailing period. I'm not sure if that came from an initial error or not, but it doesn't seem to cause a problem. @NitigyaS do you know if that period (`.`) has a particular meaning, or if it's just a typo that ended up not causing a problem?

src/commcare_cloud/ansible/roles/postgresql/tasks/set_up_dbs.yml
```diff
@@ -0,0 +1,67 @@
...
+    ansible_ssh_pipelining: true
+  postgresql_user:
+    name: "{{ item.username }}"
+    password: "{{ item.password }}"
+    role_attr_flags: "{{ item.role_attr_flags|default('CREATEDB') }}"
+    state: present
+    port: "{{ postgresql_port }}"
+    login_host: "{{ db_is_remote and postgresql_host or omit }}"
+    login_user: "{{ db_is_remote and postgres_users.root.username or omit }}"
+    login_password: "{{ db_is_remote and postgres_users.root.password or omit }}"
```
snopoke: what's the significance of `omit` in theres?

src/commcare_cloud/ansible/roles/postgresql/tasks/set_up_dbs.yml
```diff
@@ -0,0 +1,67 @@
+- name: Create PostgreSQL users
+  become: yes
+  become_user: "{{ db_is_remote and 'ansible' or 'postgres' }}"
```
snopoke: how does this work for AWS? Where do these tasks get run?

src/commcare_cloud/ansible/roles/postgresql/tasks/set_up_dbs.yml
```diff
@@ -0,0 +1,67 @@
...
+  vars:
+    ansible_ssh_pipelining: true
+  postgresql_user:
+    name: "{{ item.username }}"
+    password: "{{ item.password }}"
+    role_attr_flags: "{{ item.role_attr_flags|default('CREATEDB') }}"
+    state: present
+    port: "{{ postgresql_port }}"
+    login_host: "{{ db_is_remote and postgresql_host or omit }}"
+    login_user: "{{ db_is_remote and postgres_users.root.username or omit }}"
```
snopoke: is this `root` user one that we create?

src/commcare_cloud/ansible/roles/postgresql/tasks/set_up_dbs.yml
dannyroberts: `omit` is an ansible feature we rarely use, but when you need it it's super useful: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#omitting-parameters.

Basically if something evaluates to

```
<module>:
  arg1: 'value'
  arg2: omit
```
that's the same as if you hadn't mentioned `arg2` at all:
```
<module>:
  arg1: 'value'
```
In this line, I use it because neither `login_password: null` or `login_password: ""` have the same meaning as just not mentioning `login_password` in the first place. In the case of postgresql on an SSHable VM, we run as the `postgres` linux user and do not need `login_username` and `login_password`; in the case of a remote db (RDS), we can't run as a linux user, so we have to provide a `postgresql` user and password (the `root` user above).

src/commcare_cloud/ansible/roles/postgresql/tasks/set_up_dbs.yml
snopoke: Is that an actual RDS machine or just a VM we keep around for running tasks against RDS instances? Just wondering if there's a reason we can't just run them from the control machine (local)?

I was thinking we could standardize to run the same workflow on all our PG nodes but I think the issue is that when we install PG we don't get a root user which we can perform operations with. Is that right?

src/commcare_cloud/terraform/staging/terraform.tf
```diff
@@ -86,7 +86,7 @@ module "commcarehq" {
     {
       server_name           = "proxy1-staging"
       server_instance_type  = "t2.medium"
-      subnet_index          = 0
+      subnet_index          = 4
```
snopoke: I think this will conflict with your other terraform changes

src/commcare_cloud/ansible/roles/postgresql/tasks/set_up_dbs.yml
dannyroberts: In option (B) above, where `db_is_remote = true`, this gets run _from_ `django_manage` (see https://github.com/dimagi/commcare-cloud/pull/2114/files#diff-8a7f836d9437d4c444262eb4453d6971R28), _"against"_ an RDS instance—"against" in the sense that `login_host` in the tasks below gets set to the DNS address of the RDS instance.

> Just wondering if there's a reason we can't just run them from the control machine (local)

this was my first thought too I think back when I was first playing around with this, but I found it tricky to get that to work—it wasn't accepting local as an option or something. Possibly related to how we add `--limit=!ansible_skip` to every command, which I think silently also excludes local for some reason. My memory's hazy so this may not be the exact right story.

> I was thinking we could standardize to run the same workflow on all our PG nodes but I think the issue is that when we install PG we don't get a root user which we can perform operations with. Is that right?

Yeah that's exactly right. I though the same thing and went through the experiment of trying to get this to _always_ run remotely, to standardize the workflow, but then realized that the instances we create using this ansible role don't have a root user, and creating one before running this role would at least partially defeat the purpose, since this file is all about creating the right user set anyway. So I settled on this way of having different auth styles—using privileged linux user for a non-remote db, and using root user/password for a remote db.

src/commcare_cloud/terraform/staging/terraform.tf
dannyroberts: Yeah quite possibly. It'll be a very easy conflict to fix though. I'd originally done this change on this branch, but then I cherry-picked it onto https://github.com/dimagi/commcare-cloud/pull/2125 as well, after which I continued with other formatting changes to make it more readable.

environments/staging-aws-test/meta.yml
```diff
@@ -1,6 +1,7 @@
 deploy_env: staging
 env_monitoring_id: staging
 users:
-  - dimagi
-  - dimagi-qa
+#  - dimagi
+#  - dimagi-qa
+  - droberts
   - g2
```
millerdev: this can be removed

environments/staging-aws-test/meta.yml
```diff
@@ -0,0 +1,7 @@
+deploy_env: staging
+env_monitoring_id: staging-aws-test
+users:
+#  - dimagi
+#  - dimagi-qa
+  - droberts
+  - g2
```
millerdev: `g2` can be removed.

environments/staging-aws-test/meta.yml
```diff
@@ -0,0 +1,7 @@
+deploy_env: staging
+env_monitoring_id: staging-aws-test
+users:
+#  - dimagi
+#  - dimagi-qa
+  - droberts
```
millerdev: `droberts` seems like a value you were using for testing. Is that still necessary?

dannyroberts: I'll probably remove soon, but it's still useful for now.

environments/staging-aws-test/meta.yml
dannyroberts: Yeah I'll check with Lucina and remove from all environments once they're done

PR #2115, opened 2018-09-05
set forks in config

snopoke:
Rather than setting this programatically just set in config.
This also has the effect of applying to runs of the 'ansible' command
and not just 'ansible_playbook'.

Increased count from 15 to 50.


dannyroberts:
Nice, using the built-in functionality :)

PR #2123, opened 2018-09-06
switching billing email to sales@

jmtroth0:
https://manage.dimagi.com/default.asp?282645#1526181

@emord 

@gbova fyi in case you want to change this solution at all when you get back


emord:
From ticket:

> This should fix it on the next deploy

This won't get fixed on the next deploy. You'll need to deploy this with commcare-cloud to production and restart services `cchq production update-config`

jmtroth0:
>This won't get fixed on the next deploy.

Huh. I guess I figured it would update localsettings.py on deploy.

I'm deploying anyway, so to be clear, the sequence is just 1) in commcare-cloud, run `update-code`, 2) deploy, 3) cchq production update-config?

nickpell:
Add step (4) restart services

Alternatively, switch the order of (2) and (3) so when services restart at the end of deploy, they will read from the updated localsettings.py

PR #2124, opened 2018-09-06
Update staging terraform

dannyroberts:
- update proxy to use public subnet
- bring pre-existing RDS instance under terraform's control

### DIFF
--- a/src/commcare_cloud/terraform/modules/commcarehq/main.tf
+++ b/src/commcare_cloud/terraform/modules/commcarehq/main.tf
@@ -40,12 +40,14 @@ variable "proxy_servers" {
 
 
 locals {
-  subnet_options  = ["${module.network.subnet-a-app-private}",
-                     "${module.network.subnet-b-app-private}",
-                     "${module.network.subnet-c-app-private}",
-                     "${module.network.subnet-a-public}",
-                     "${module.network.subnet-b-public}",
-                     "${module.network.subnet-c-public}"]
+  subnet_options = {
+    private-a = "${module.network.subnet-a-app-private}"
+    private-b = "${module.network.subnet-b-app-private}"
+    private-c = "${module.network.subnet-c-app-private}"
+    public-a = "${module.network.subnet-a-public}"
+    public-b = "${module.network.subnet-b-public}"
+    public-c = "${module.network.subnet-c-public}"
+  }
 }
 
 module "servers" {
--- a/src/commcare_cloud/terraform/modules/commcarehq/main.tf
+++ b/src/commcare_cloud/terraform/modules/commcarehq/main.tf
@@ -101,35 +101,50 @@ module "Redis" {
 #  # dns_domain       = "${var.dns_domain}"
 #}
 
-# If an RDS instance or cluster is needed, uncomment this.
-#module "database" {
-#  source                         = "../rds"
-#  vpc_id                         = "${module.network.vpc-id}"
-#  environment                    = "${var.environment}"
-#  dns_zone_id                    = "${var.dns_zone_id}"
-#  dns_domain                     = "${var.dns_domain}"
-#  azs                            = "${var.azs}"
-#  rds_storage                    = "${var.rds_storage}"
-#  rds_engine                     = "${var.rds_engine}"
-#  rds_engine_version             = "${var.rds_engine_version}"
-#  rds_instance_type              = "${var.rds_instance_type}"
-#  rds_database_name              = "${var.rds_database_name}"
-#  rds_username                   = "${var.rds_username}"
-#  rds_storage_type               = "${var.rds_storage_type}"
-#  rds_backup_window              = "${var.rds_backup_window}"
-#  rds_backup_retention           = "${var.rds_backup_retention}"
-#  rds_maintenance_window         = "${var.rds_maintenance_window}"
-#  rds_auto_minor_version_upgrade = "${var.rds_auto_minor_version_upgrade}"
-#  rds_multi_az                   = "${var.rds_multi_az}"
-#  rds_storage_encrypted          = "${var.rds_storage_encrypted}"
-#  rds_port                       = "${var.rds_port}"
-#  environment                    = "${var.environment}"
-#  rds_password                   = "${var.rds_password}"
-#  rds_prevent_destroy            = "${var.rds_prevent_destroy}"
-#  rds_instance_count             = "${var.rds_instance_count}"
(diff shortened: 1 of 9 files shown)


src/commcare_cloud/terraform/modules/commcarehq/main.tf
```diff
@@ -101,35 +103,50 @@ module "Redis" {
...
+  vpc_security_group_ids = ["${module.network.rds-sg}"]
+
+  maintenance_window = "${local.rds["maintenance_window"]}"
+  backup_window      = "${local.rds["backup_window"]}"
+  backup_retention_period = "${local.rds["backup_retention"]}"
+
+  # DB subnet group
+  subnet_ids = ["${module.network.subnet-a-db-private}",
+                "${module.network.subnet-b-db-private}",
+                "${module.network.subnet-c-db-private}"]
```
snopoke: are these supposed to appear in the list of subnets in `subnet_options`?

src/commcare_cloud/terraform/staging/terraform.tf
```diff
@@ -27,57 +26,57 @@ module "commcarehq" {
...
       server_name           = "web0-staging"
       server_instance_type  = "t2.micro"
-      subnet_index          = 0
+      subnet                = "private-a"
     },
     {
       server_name           = "web1-staging"
       server_instance_type  = "t2.micro"
-      subnet_index          = 1
+      subnet                = "private-b"
```
snopoke: why is this and then next two on different subnets?

src/commcare_cloud/terraform/staging/terraform.tf
```diff
@@ -27,57 +26,57 @@ module "commcarehq" {
...
       server_name           = "pgproxy0-staging"
       server_instance_type  = "t2.medium"
-      subnet_index          = 0
+      subnet                = "private-a"
     },
     {
       server_name           = "control-staging"
       server_instance_type  = "t2.medium"
-      subnet_index          = 3
+      subnet                = "public-a"
```
snopoke: according to pattern above, index 0 = public-a. This was index 3 but it's also public-a. Is that correct?

src/commcare_cloud/terraform/staging/terraform.tf
```diff
@@ -86,7 +85,14 @@ module "commcarehq" {
     {
       server_name           = "proxy1-staging"
       server_instance_type  = "t2.medium"
-      subnet_index          = 0
+      subnet                = "public-b"
```
snopoke: should this be public-a (for index 0)?

src/commcare_cloud/terraform/staging/terraform.tf
dannyroberts: Yeah so I think there's a large amount of arbitrariness here, but I'm just following the setup G2 left us with when I don't have a specific reason to change it. The only part that's not arbitrary and does serve a purpose, as far as I can tell, is that web0 and web1 are in different _availability zones_ (zones a & b). Availability zones are structured to be two completely parallel systems within the same data center so that a system-wide outage can only possibly affect a single availability zone. We don't have enough redundancy make the application robust against complete outage of a single AZ, but with more machines (esp. redundant proxy) you can get closer to that.

That said, there's no downside to having machines on different AZs, so even if it doesn't fully protect us, it could conceivably be helpful in the case of partial service interruption in a single AZ, and it doesn't cost us anything.

src/commcare_cloud/terraform/modules/commcarehq/main.tf
dannyroberts: No this is a different set of subnets altogether. Not totally sure why G2 set the DB subnets to be distinct from the app subnets, but again, I don't think it hurts. I don't have a strong enough sense of AWS networking (or networking best practices in general) to say that it's pointless, so for now I'm just leaving it as they set it up.

src/commcare_cloud/terraform/staging/terraform.tf
```diff
@@ -27,57 +26,57 @@ module "commcarehq" {
...
       server_name           = "web1-staging"
       server_instance_type  = "t2.micro"
-      subnet_index          = 1
+      subnet                = "private-b"
     },
     {
       server_name           = "celery0-staging"
       server_instance_type  = "t2.large"
-      subnet_index          = 1
+      subnet                = "private-b"
```
NitigyaS: a subnet can not span multiple availability zones. That means `celery0-staging` is also in a different availability zone than machines in `private-a` and `private-c`. 
Maybe there is something I am missing here but, I think this and so do `pillow0-staging` should be in the same availability zone as other machines?

src/commcare_cloud/terraform/staging/terraform.tf
```diff
@@ -27,57 +26,57 @@ module "commcarehq" {
...
       server_name           = "kafka0-staging"
       server_instance_type  = "t2.medium"
-      subnet_index          = 0
+      subnet                = "private-a"
     },
     {
       server_name           = "es0-staging"
       server_instance_type  = "t2.micro"
-      subnet_index          = 0
+      subnet                = "private-a"
```
NitigyaS: Shouldn't elasticsearch have a separate subnet like we are doing in postgresql.
Or at least different than applications subnet.

src/commcare_cloud/terraform/modules/network/main.tf
```diff
@@ -290,7 +290,6 @@ resource "aws_security_group" "g2-access-sg" {
   }
```
NitigyaS: it should be `dimaig-access-sg` instead `g2-access-sg`

NitigyaS: Suggestion on the private route table.
can we mark the private route table as default so all the subnets created in future will go directly to the default route table (private) or you see any downfall of that?

src/commcare_cloud/terraform/staging/terraform.tf
NitigyaS: I always thought AZ as a datacenter. They do provide very less latency (connected by low latency links ) but latency is even less if VMs are in the same AZ (datacenter). 
as you mentioned in another comment, AZs main purpose is to provide high availability so that if AZ1 fails we can shift the load to the machines in AZ2.

dannyroberts:
Scheduled a sync for Monday with @NitigyaS to review the basic network design. I'm merging this for now as-is to get some other changes out, and I'll update with outcomes from our discussion next week.

PR #2126, opened 2018-09-07
update datadog monitors

snopoke:
Not sure if `commcare-cloud` is the right place for this but putting it here for now.

Needs some docs.


proteusvacuum:
Build fail looks related:
```
======================================================================
ERROR: tests.test_jinja_templates.test_jinja_templates
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/dimagi/commcare-cloud/tests/test_jinja_templates.py", line 16, in test_jinja_templates
    jinja_env.parse(template.read(), filename=path)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/jinja2/environment.py", line 493, in parse
    self.handle_exception(exc_info, source_hint=source)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "./src/commcare_cloud/manage_commcare_cloud/monitors/notification_block.j2", line 2, in template
    {{#is_match "<<env_key>>" "<<env>>"}} <<alert_channel>> {{/is_match}}
TemplateSyntaxError: unexpected char u'#' at 48
```

snopoke:
I'm having conflicting thoughts about whether this is actually useful to have. I wanted it now so that I could update the notification block on a bunch of monitors without having to edit each individually but I can't really think of another use for it and that task doesn't happen often.

@dannyroberts @proteusvacuum any thoughts on that?

dannyroberts:
Wow if I'm understanding this right, as a tool that keeps some/most/all of our datadog monitor/alert configurations in code, I think that's pretty great, whether or not this is specifically the right place to keep it. How about we sync on it briefly today and hopefully come up with the right place for it to live.

dannyroberts:
I merged master, tweaked this diffs to be cleaner, and updated all of the check configs to match what's on datadog or at least only contain desirable diffs that are the result of renaming india, adding opsgenie to prod, etc. and essentially represent a manual mistake someone (me) made.

I also made an attempt at fixing the test failure. If the tests passes, I vote to get this in so we can start using it and improving it, something I'm happy to pick up under platform reliability.

dannyroberts:
And I forgot to state a premise of this whole thing which is: @snopoke this is a really great tool that even in its first-pass state could provide a lot of value. It already caught a number of alerts that may have gotten silenced in the softlayer => india switchover and other inconsistencies, and with a little work this could be a really great workflow.

dannyroberts:
I made these followup PRs that add functionality and move all our monitors into code:
- https://github.com/dimagi/commcare-cloud/pull/2771
- https://github.com/dimagi/commcare-cloud/pull/2772
- https://github.com/dimagi/commcare-cloud/pull/2773

PR #2134, opened 2018-09-10
Simplify supervisord installation

dannyroberts:
potentially fixes bug that caused supervisor executable path to be
missing from the conf, which made it impossible to start supervisord
(using upstart at least) until upgrade-users had been run a second time


dannyroberts:
@millerdev approved this over my shoulder

PR #2143, opened 2018-09-11
Add terraform command and update networking

dannyroberts:
This PR contains a number of separate things, so my apologies.* The best way to review is commit by commit. Some notes:

- Towards the beginning (other than the first commit) the commits are mostly about refactoring and changing the network settings either to fix something that wasn't working or to move to a network design I discussed with @NitigyaS. We were aiming for something that takes advantage of some basic subnet restrictions without getting us tangled in something overly complicated or restrictive.
- The last few commits add a `terraform` command that autogenerates the entrypoint file, assembles the files you need into a temp directory (`<env>/.generated-terraform`), and then runs `terraform <extra args>` from that directory. Having that will make it a lot easier to roll out both to the team and to other environments.

*When I'm working with actual remote changes on AWS, it's hard to do other work in a side branch, since then I'll see reverse diffs from the other branch, so I've fallen into doing all of my terraform work on a single longer-lived branch. As the changes I'm making become less all-encompassing, I'm hoping to PR smaller things at a time.


src/commcare_cloud/terraform/modules/network/main.tf
```diff
@@ -160,6 +160,22 @@ resource "aws_security_group" "proxy-sg" {
   }
 }
 
+resource "aws_security_group" "ssh" {
+  name   = "ssh-sg-${var.env}"
+  vpc_id = "${aws_vpc.main.id}"
+
+  ingress {
+    from_port   = 22
```
NitigyaS: I think should be `0` from all ports.

src/commcare_cloud/terraform/modules/network/main.tf
```diff
@@ -327,11 +327,15 @@ resource "aws_security_group" "rds" {
   vpc_id = "${aws_vpc.main.id}"
 
   ingress {
-    from_port =         "5432"
-    to_port =           "5432"
-    protocol =          "tcp"
-    cidr_blocks =       ["${var.vpc_begin_range}.10.0/24"]
-    ipv6_cidr_blocks =  ["::/0"]
+    from_port = "5432"
```
NitigyaS: this should be `0` since the application can connect from any port on VM to 5432 of RDS.

src/commcare_cloud/terraform/modules/openvpn/main.tf
```diff
@@ -5,7 +5,7 @@ resource aws_instance "server" {
   instance_type          = "${var.vpn_size}"
   subnet_id              = "${var.instance_subnet}"
   key_name               = "g2-access"
-  vpc_security_group_ids = ["${var.vpc-all-hosts-sg}","${var.g2-access-sg}","${aws_security_group.openvpn-access-sg.id}"]
```
NitigyaS: `openvpn-access-sg` is different from `ssh-sg`?

dannyroberts: This module is actually not yet used. I think G2 originally added it thinking they'd use it to set up our openvpn server, but that ended up getting set up in a way that's not captured here.

I'm still trying to figure out how we want to do VPNs for multiple environments so it's on my list to either use this or remove it from the code.

src/commcare_cloud/environment/main.py
```diff
@@ -123,8 +123,11 @@ def users_config(self):
 
     @memoized_property
     def terraform_config(self):
-        with open(self.paths.terraform_yml) as f:
-            config_yml = yaml.load(f)
+        try:
+            with open(self.paths.terraform_yml) as f:
+                config_yml = yaml.load(f)
+        except IOError:
+            return None
```
millerdev: Does the thing that calls this check if the result is `None`?

In general, I'm not really a fan of this pattern of error handling since it forces the caller to inspect the returned object before using it. The alternatives all probably involve more work though, so I guess it's reasonable to do it this way.

dannyroberts: Yeah the annoying thing here is that this needs to be pickleable (since this object needs to be picklable). The calling code doesn't check this currently, so on an env that doesn't use terraform (all of them but the test one currently) this will give a traceback to this point if you run `cchq <env> terraform ...`. Not the greatest, but for the level of maturity we're currently at with this feature I think it's roughly appropriate.

dannyroberts:
Thanks all! Makin' progress!

PR #2145, opened 2018-09-12
Use spawnvp instead of execvp

emord:
This allows calling two fab commands in a row.

https://docs.python.org/2/library/os.html

execvp:

> These functions all execute a new program, replacing the current process; they do not return

spawnvp waits until it returns

@dannyroberts any reason that we were using execvp originally? I haven't tested this out with other commands, but this is what I would have expected behavior to be. 

Also note that the python docs suggest using `subprocess` here. I tried that for a second, but got an error about not being able to find `fab`. I didn't look deeply into it, but that would probably be even better

### DIFF
--- a/src/commcare_cloud/commands/fab.py
+++ b/src/commcare_cloud/commands/fab.py
@@ -70,4 +70,4 @@ def exec_fab_command(env_name, *extra_args):
     ) + tuple(extra_args)
     cmd = ' '.join(shlex_quote(arg) for arg in cmd_parts)
     print_command(cmd)
-    os.execvp('fab', cmd_parts)
+    os.spawnvp(os.P_WAIT, 'fab', cmd_parts)
--- a/src/commcare_cloud/commands/fab.py
+++ b/src/commcare_cloud/commands/fab.py
@@ -70,4 +70,4 @@ def exec_fab_command(env_name, *extra_args):
     ) + tuple(extra_args)
     cmd = ' '.join(shlex_quote(arg) for arg in cmd_parts)
     print_command(cmd)
-    subprocess.call(cmd_parts)
+    return subprocess.call(cmd_parts)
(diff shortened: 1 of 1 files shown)


dannyroberts:
Thanks for looking into this. Originally the reason I went the route of replacing the current process (with `execvp`) instead of spawning a new one (with `subprocess.*`) is that when I was trying it out I could not for the life of me replicate natural I/O forwarding to the subprocess. For SSH, for example, the user needs to be able to interact with the `ssh` program directly and interactively, and mediating that through the original process using `.communicate` and the like were just not working out for me. (For `fab` it's less dramatic but still important—we need to be able to `y` through various prompts, etc.) When I stumbled upon this (replacing the process) as a solution, I didn't love that we couldn't then do something with the return code, but didn't have an immediate use-case for that and figured I'd go with it for the time being.

Do you know if `spawnvp` handles I/O forwarding seamlessly the way we'd like it to? My level of understanding around this is definitely at the search stuff on google until it works level, so I'm by no means confident what I have here is the only or best solution.

As for the subprocess not being able to find `fab`, I think in the past we've been able to fix that by using `shell=True`, which seems to preserve more about the parent context, like where to find executables and stuff including in the virtualenv. (Again though, I didn't have luck with getting the I/O behavior I wanted from `subprocess.*`.)

emord:
Thanks for all of that background. 

I can say that I successfully ran `update-supervisor-conf` on ICDS and was able to successfully go through both prompts without an issue. It restarted all processes as i would have expected.

I believe that that satisfies the behavior you're looking for, but my understanding of this is also at the "search stuff on google until it works level"

dannyroberts:
Oh, wow, you're right! And furthermore, `return subprocess.call(cmd_parts)` works exactly as I'd described I wanted, which is both great and mildly embarrassing. I guess we should just go with that throughout!

emord:
@dannyroberts Admittedly didn't test this, but I believe this is the change that you mentioned working

dannyroberts:
Great work @emord

PR #2148, opened 2018-09-13
Add from-scratch staging env in separate AWS account using terraform

dannyroberts:
These seem to be all of the changes we need to make `cchq staging terraform apply` run cleanly. It's not clear to me whether the rackpace routing stuff works, but I'll dig into it tomorrow (and make sure it still works on the staging-aws-test env as well).


dannyroberts:
This should be ready to review

PR #2150, opened 2018-09-14
add extra config with auth_cache_size

snopoke:
Seems like the default isn't big enough: 
image

http://docs.couchdb.org/en/stable/config/auth.html#couch_httpd_auth/auth_cache_size


snopoke:
This didn't work, even with very large numbers for `auth_cache_size` (100000)

snopoke:
I think it might be because we're using the admin user for all requests which isn't stored in the `_users` database.

emord:
fixed in 2.3.0

https://github.com/apache/couchdb/issues/1746

PR #2160, opened 2018-09-19
WIP; Get more visibility into network issues on TCL cloud

sravfeyn:


src/commcare_cloud/ansible/roles/datadog/templates/tcp_check.yaml.j2
```diff
@@ -0,0 +1,22 @@
...
+init_config:
+
+instances: 
+# debug; Todo
+{{ hostvars[groups['webworkers'][0]].ansible_hostname }}
+{% for tcp_check in datadog_tcp_checks_list %}
+  {% if tcp_check.from_host_groups | intersect(group_names) | length > 0 %}
+    {% for to_group in tcp_check.to_host_groups %}
+      {% for to_host in groups[to_group] %}
+        {% if hostvars[to_host].ansible_hostname == inventory_hostname %}
```
sravfeyn: I am running into an error which I couldn't nail down - `ansible_hostname` is unavailable as an attribute of `hostvars[to_host]` for some reason. As a test I added a check above line 8, which evaluates correctly, but fails on line 13. Do you spot any issues here?

PR #2163, opened 2018-09-20
Add more pillows

mkangia:
on pillow 3
added 2 processes for `kafka-ucr-main`
added 1 for `FormSubmissionMetadataTrackerPillow`

on pillow 0 
added one for `kafka-ucr-main` since one pillow was removed from it earlier [here](https://github.com/dimagi/commcare-cloud/pull/2162/files)

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -94,13 +94,23 @@ pillows:
     XFormToElasticsearchPillow:
       num_processes: 2
     FormSubmissionMetadataTrackerPillow:
-      num_processes: 2
-    kafka-ucr-main:
+      start_process: 0
       num_processes: 1
+      total_processes: 2
+    kafka-ucr-main:
+      start_process: 0
+      num_processes: 2
+      total_processes: 7
   hqpillowtop2.internal-va.commcarehq.org:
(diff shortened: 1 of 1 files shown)


environments/production/app-processes.yml
```diff
@@ -93,6 +93,10 @@ pillows:
       num_processes: 4
     XFormToElasticsearchPillow:
       num_processes: 2
+    FormSubmissionMetadataTrackerPillow:
+      num_processes: 2
+    kafka-ucr-main:
+      num_processes: 1
```
emord: To have the same pillow on different servers there are a couple of other parameters that need to be specified:

https://github.com/dimagi/commcare-cloud/blob/8aee79f74fd25b1d960c0d807eb3e82c9bcc04e7/environments/icds-cas/app-processes.yml#L239-L242

emord:
looks a bit more complicated than needed, but it looks ok to me

mkangia:
Did so due to resourcing,
kafka-ucr-main takes quite a lot of RAM so could not add 3 to pillow3 itself. Added one to pillow0 instead since it had some room and avoided increasing form meta on the same machine to avoid conflict with ucr-main

PR #2165, opened 2018-09-20
Update staging aws network

dannyroberts:


dannyroberts:
Yes it would mean that, but I still need to troubleshoot the site-to-site connection to rackspace and document how to connect to the VPN for our staging aws account

dannyroberts:
Okay, verified that the site-to-site connection is working. VPN is up and I know the process to add new users which I'll document, but I'll merge this in the meantime.

PR #2169, opened 2018-09-23
haproxy setup for couchdb2

snopoke:


src/commcare_cloud/ansible/roles/haproxy/templates/service.j2
```diff
@@ -0,0 +1,92 @@
...
+{% set _ = entry.append(item.service.haproxy_backend_nodes | count | string) %}
+    {{ entry | join(' ') }}
+{% endfor %}
+
+{% for host_name in item.service.haproxy_backup_nodes|default([]) %}
+{% if hostvars[host_name] is defined %}
+{%   set ip_addr = hostvars[host_name].internal_ipv4.address %}
+{% endif %}
+{% set entry = [] %}
+{% set _ = entry.append("server") %}
```
millerdev: Interesting use of mutating side-effects. Could it be structured like this instead?
```
{% set entry = [
    "server",
    ((host_name.name | default(host_name)) | string),
    ...
] %}
{{ entry | join(' ') }}
```
or even better
```
{{ [
    "server",
    ((host_name.name | default(host_name)) | string),
    ...
] | join(' ') }}
```

src/commcare_cloud/ansible/roles/thp/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+- stat:
+    path: /etc/default/grub
+    register: uses_grub
+
+- stat:
+    path: /sys/kernel/mm/transparent_hugepage/enabled
+  register: thp_enabled
```
millerdev: Inconsistent indentation (also below)

src/commcare_cloud/ansible/roles/haproxy/templates/service.j2
snopoke: interesting, that would read better. Most of this role is copied from https://github.com/openstack/openstack-ansible-haproxy_server as noted in the README.

I'll test this out and update.

snopoke:
rolled this out to ICDS

PR #2171, opened 2018-09-24
Add new webworker

mkangia:
Adding web worker to prod http://manage.dimagi.com/default.asp?283306

was asked to add two but pushing back if we need actually need 2. So starting with one here.


mkangia:
@proteusvacuum this was not ready yet. Need to add the server's public key under known hosts.
@NitigyaS can you add the other file you mentioned as well?

PR #2172, opened 2018-09-24
Bump JS versions

proteusvacuum:
@orangejenny @jmtroth0


proteusvacuum:
@jmtroth0 our docs already don't tell you to install any specific version (https://github.com/dimagi/commcare-hq/#installing-bower). You are probably running the version locally that was current when you installed HQ. i.e. no change needed locally.

proteusvacuum:
tested a deploy on staging and all went well

PR #2177, opened 2018-09-25
add 2 more processes for export download queue

mkangia:


mkangia:
Merging before build. Seems safe enough.

PR #2181, opened 2018-09-26
Rekey DimagiKeyStore

dannyroberts:
Thanks @NitigyaS for noticing yesterday. I bumped into this too and went ahead and rotated DimagiKeyStore file for each environment where it existed.


dannyroberts:
Nope, that's good. I used `ansible-vault rekey` to do this, so the content shouldn't have changed (and it would have failed if I put in the wrong old password).

PR #2187, opened 2018-09-27
Mk move phonelogs

mkangia:


src/commcare_cloud/environment/constants.py
```diff
@@ -6,6 +6,7 @@ class _Constants(jsonobject.JsonObject):
     formplayer_db_name = jsonobject.StringProperty(default='formplayer')
     ucr_db_name = jsonobject.StringProperty(default='commcarehq_ucr')
     synclogs_db_name = jsonobject.StringProperty(default='commcarehq_synclogs')
+    phonelogs_db_name = jsonobject.StringProperty(default='commcarehq_phonelogs')
```
mkangia: Saw these things in other yml files like "synclogs_db_name" but those files were "generated" and looked like picked from the default value here, so didn't add at other places.

mkangia:
Thanks @snopoke 
addressed both changes

PR #2188, opened 2018-09-27
Terraform production

dannyroberts:
Setting up production instance on AWS. (Already deployed.)

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -1,0 +1,48 @@
+account_alias: commcare-production
+state_bucket: dimagi-terraform
+state_bucket_region: "us-east-1"
+region: "us-east-1"
+environment: "production"
+azs:
+  - "us-east-1a"
+  - "us-east-1b"
+  - "us-east-1c"
+vpc_begin_range: "10.202"
+
+vpn_connections:
+  - name: rackspace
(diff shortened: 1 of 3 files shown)


environments/production/terraform.yml
```diff
@@ -0,0 +1,48 @@
...
+    bgp_asn: 65000
+    amazon_side_asn: 7224
+
+external_routes:
+  - cidr_block: "172.24.16.0/22"
+    gateway_id: "vgw-05c8c22d05e33c5bf"
+  - cidr_block: "172.24.32.0/22"
+    gateway_id: "vgw-05c8c22d05e33c5bf"
+
+key_name: droberts
```
snopoke: what is this for?

dannyroberts:
Note to self: I've followed the terraform/README.md up `Gain temporary SSH access via your own IP`. (i.e. openvpn is not yet set up). Those directions should be sufficient to set it up if somehow that becomes a priority while I'm offline.

environments/production/terraform.yml
dannyroberts: This is the public key that originally gets installed on new EC2 instances so I have access to bootstrap the machine. Here `droberts` means "pull the public key from `_authorized_keys/droberts.pub`.

PR #2189, opened 2018-09-27
Set kafka api version and stop load balancing accounting

emord:


emord:
https://github.com/dimagi/commcare-hq/pull/21951

the only places where NoBrokersAvailable is thrown should be skipped by setting the kafka version

PR #2196, opened 2018-09-28
Add one more process for kafka-ucr-main

emord:
Just temporarily until the queue goes back to 0.


mkangia:
I suggest we keep it this way for some time.. When form submissions are high, there is a little backlog..

PR #2198, opened 2018-09-28
New pg machine with more resources to replace hqdb2 

sravfeyn:
@mkangia


environments/production/known_hosts
```diff
@@ -5,32 +5,28 @@
...
 hqcelery0.internal-va.commcarehq.org,172.24.32.13 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBLtaIJBzAQz8VQGPXpLlS+8Z+JadUvaEklgx6fOzZPg0HqOgF9TMXFi7dgJEIkbr/n3vYFfHg67Yaa1moNGpWt8=
 hqcelery0.internal-va.commcarehq.org,172.24.32.13 ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAICK/myLm0lapD3ufeTR4fRy2Ssw6yYe+J4TjhNvbg2AP
 hqcelery0.internal-va.commcarehq.org,172.24.32.13 ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDNwysSUb2RlX8/C8dx0HVwZou1KTkvF+aDAAgbjGIUEVsDKvnfY+6S9uxOQ3tXPNJYWYMl1608vRkVdf8JmOEOiTuFS+YaRJhln4IfxMhlBMZqjvmwhii6Nnskm2TCFZSG77+kV7PJqKMuP7tkHZZbgE1qKFMCASXUdRhmNWfc001p3UWBMclKzFW0zx8EH/K4RYznd/I6zx7Bn04pAqvZOOfTRHlfgkSO7zh848da6AJqIHur4ogMRnwC+ciUZrDChIIIwoQkPe12/f7aGLUlc1o9FsUSS6wvfoSVsU24yzIjE0Wj4RtUEbs9sDMknbOJC0TOH7FLyYnzU2dENNVN
 hqcelery1.internal-va.commcarehq.org,172.24.32.14 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBM09leVaWCkVHKuhcUxbCRSYBdUOEnQMV2dqXuM7of34EdIgwyHdM41epuCtRQ0JqpCA5pdf43yAgF8wUojKdjA=
-hqcelery1.internal-va.commcarehq.org,172.24.32.14 ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINKX5WIFLCIepkO5gYDE70v0OnkKHa/o+bFqh3lk2KZh
```
mkangia: i would expect only new entry for the db machine in this file and not others.
@snopoke saw something similar on Danny's PR, do you know why this is and if its okay?

snopoke: I've noticed differences if I run the command locally or on the control machine

environments/production/inventory.ini
```diff
@@ -81,6 +87,7 @@ hqdb0
 hqdb1
 hqdb2
 hqdb3
+hqdb5
```
emord: what about hqdb4?

sravfeyn: It was an oversight. Didn't care much since this was going to be hqdb2 after the migration

emord: what do you mean it's going to be hqdb2? long term plan was just splitting the current database I thought?

sravfeyn: From the doc, I gathered that we wanted to increase the resources on hqdb2 by migrating all of it to a flavour where we can add resources. If that's not the case, then we should probably just destroy this VM and restart, because this VM has double the resources of hqdb2 which it probably doesn't need if we are splitting hqdb2 and since this is not resizable as well.

PR #2200, opened 2018-09-28
Cache HQ multimedia in nginx

emord:
@proteusvacuum this will cache all media from apps on nginx's disk for 24 hours since they were last accessed.


emord:
did a quick calc and it would have eliminated about 20k requests to gunicorn over the past day, assuming the cache didn't run out:

```
(production) jemord@hqproxy0:/home/cchq/www/production/log$ grep multimedia production_commcare-nginx_access.log |  awk '{print $10}' | grep multimedia | sort | uniq | wc -l
4370
(production) jemord@hqproxy0:/home/cchq/www/production/log$ grep multimedia production_commcare-nginx_access.log |  awk '{print $10}' | grep multimedia | sort  | wc -l
23612
```

PR #2201, opened 2018-09-28
Add hqdjango7

proteusvacuum:
This machine is based on an image of `hqdjango4` instead of building it from scratch.

I want to see if this machine goes down in the same way that our other new workers have been. If it doesn't, it might mean there is something wrong with our bootstrapping scripts, or something that is happening when building our environment from scratch.


proteusvacuum:
this machine is now running

PR #2202, opened 2018-10-01
Quick Status Check playbook to find issues in ICDS

NitigyaS:
Simple checks that will tell what is wrong on ICDS

### DIFF
--- a/src/commcare_cloud/ansible/status_check.yml
+++ b/src/commcare_cloud/ansible/status_check.yml
@@ -38,6 +38,16 @@
     file:
       path: /opt/data/ansible-test
       state: touch
+    register: opt_data_readable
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/status_check.yml
```diff
@@ -0,0 +1,84 @@
...
+    register: opt_data_readable
+    tags:
+    - opt_data
+
+  - name: Remove the test file
+    become: yes
+    file:
+      path: /opt/data/ansible-test
+      state: absent
+    when: opt_data_readable
```
snopoke: I think this should be `opt_data_readable is success`. Same for `/home/cchq` below

PR #2214, opened 2018-10-03
remove giovanni/clark's keys

sravfeyn:
@mkangia


environments/_users/dimagi.yml
```diff
@@ -25,3 +24,4 @@ dev_users:
   absent:
     - astone
     - wpride
+    - gcapalbo
```
emord: I think that nclark needs to be removed frrom dimagi-qa

sravfeyn: It was already in the absent list https://github.com/dimagi/commcare-cloud/blob/master/environments/_users/dimagi-qa.yml#L4

Am I missing any other place?

emord: thanks for confirming. misread that before

PR #2218, opened 2018-10-03
Add new "current" form pillows

emord:
Adding some pillows with a more recent offset. i'll monitor this over the next couple of days and write up something based on the perfomance seen


emord:
@snopoke I did

PR #2219, opened 2018-10-04
Riak fixes

pr33thi:
There were a couple places where the riak-related variables being passed into different playbooks weren't being recognized in those playbooks.

@calellowitz 
code buddy @jmtroth0

### DIFF
--- a/src/commcare_cloud/ansible/deploy_riakcs.yml
+++ b/src/commcare_cloud/ansible/deploy_riakcs.yml
@@ -119,12 +119,6 @@
         name: riak
         drop_conf: "{{ backend_vars.drop_conf }}"
         advanced_conf: "{{ backend_vars.advanced_conf }}"
-        data_dirs:
-          - "{{ riak_data_dir }}"
-          - "{{ riak_ring_dir }}"
-          - "{{ riak_data_root_leveldb }}"
-          - "{{ riak_data_root_bitcask if _riak_backend == 'bitcask' else '' }}"
-          - "{{ riak_data_root_blocks if _riak_backend == 'leveldb' else '' }}"
         when: _force_riak_config | bool
 
 - name: Start Riak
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/riakcs/configure/tasks/main.yml
```diff
@@ -46,6 +51,9 @@
     mode=0640
   notify: "restart {{ name }}"
 
+- set_fact:
+    advanced_conf: "{{ _backend_vars.bitcask.advanced_conf }}"
```
calellowitz: This will hardcode the bitcask backend for all environments, right? I do not think we want that as many of our environments use a different backend.

src/commcare_cloud/ansible/deploy_riakcs.yml
```diff
@@ -119,12 +119,6 @@
         name: riak
         drop_conf: "{{ backend_vars.drop_conf }}"
         advanced_conf: "{{ backend_vars.advanced_conf }}"
-        data_dirs:
-          - "{{ riak_data_dir }}"
-          - "{{ riak_ring_dir }}"
-          - "{{ riak_data_root_leveldb }}"
-          - "{{ riak_data_root_bitcask if _riak_backend == 'bitcask' else '' }}"
-          - "{{ riak_data_root_blocks if _riak_backend == 'leveldb' else '' }}"
         when: _force_riak_config | bool
```
calellowitz: Is it possible these were not getting passed through just because this conditional was evaluating to false? If you run the original playbook with this as true, does it work? It may not be a bug in the playbook, but in the way commcare-cloud uses it

pr33thi:
Closing this PR in favor of [this one](https://github.com/dimagi/commcare-cloud/pull/2223).

PR #2220, opened 2018-10-04
Add command to fill inventory from AWS

dannyroberts:
Very rudimentary pattern here to allow templating an inventory and filling in IP addresses from what's on AWS. Helpful to be able to use the machine's name and have it fill in the address, and for when you need to rebuild a machine and have it change the IP for you.

### DIFF
--- a/src/commcare_cloud/commands/terraform/aws.py
+++ b/src/commcare_cloud/commands/terraform/aws.py
@@ -28,54 +28,81 @@ def aws_cli(environment, cmd_parts):
         check_output(cmd_parts, env={'AWS_PROFILE': environment.terraform_config.aws_profile}))
 
 
+def get_aws_resources(environment):
+    config = environment.terraform_config
+
+    # Private IP addresses
+    private_ip_query = aws_cli(environment, [
+        'aws', 'ec2', 'describe-instances',
+        '--filter', "Name=tag-key,Values=Name", "Name=tag-value,Values=*",
+        "Name=instance-state-name,Values=running",
+        "Name=tag-key,Values=Environment",
+        "Name=tag-value,Values={}".format(config.environment),
+        "--query",
+        "Reservations[*].Instances[*][Tags[?Key=='Name'].Value, NetworkInterfaces[0].PrivateIpAddresses[0].PrivateIpAddress]",
+        "--output", "json",
+        "--region", config.region,
+    ])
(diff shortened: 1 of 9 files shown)


src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -28,45 +28,81 @@ def aws_cli(environment, cmd_parts):
...
+    help = "Fill inventory.ini.j2 using existing AWS resources"
+
+    def run(self, args, unknown_args):
+        environment = get_environment(args.env_name)
+        resources = get_aws_resources(environment)
+        with open(environment.paths.inventory_ini_j2) as f:
+            template_string = f.read()
+
+        for name, address in resources.items():
+            template_string = template_string.replace('{{ ' + name + ' }}', address)
```
pr33thi: So if this is to replace the parts of the inventory.ini.j2 file that are nested in curly braces, how come some of the groups are hardcoded ie https://github.com/dimagi/commcare-cloud/pull/2220/files#diff-b1724c5eb58f38c295df121214624b47R91 ?

src/commcare_cloud/terraform/modules/commcarehq/main.tf
```diff
@@ -78,7 +78,7 @@ resource "aws_eip" "proxy" {
 module "Redis" {
   source               = "../elasticache"
   create               = "${lookup(local.redis, "create", true)}"
-  cluster_id           = "${var.environment}-redis"
+  cluster_id           = "redis-${var.environment}"
```
pr33thi: how come this is changing?

dannyroberts: I noticed I had named this backwards from how other resources are named, so I decided to change it to match. I noticed this when I was adding the code to look up the redis endpoint directly from AWS.

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: There are a lot of lines that are commented out in the staging inventory file right now, like the one you pointed out, that are essentially placeholders for machines that don't exist yet. (The IP addresses that are there are IP of machines that existed at some point when I was testing some things a couple months back, but have since been destroyed.)

pr33thi: Sounds good! A comment to that effect could be useful

PR #2222, opened 2018-10-05
Better integrate our OpenVPN setup

dannyroberts:
This PR ended up including a number of rearrangements, but because I've been running provable-deploy, I'm fairly certain this doesn't introduce bugs for the general use case.

In addition to managing openvpn user accounts via ansible—which means for example we won't have to remember to remove VPN users as a separate step when someone leaves—this PR also adds a new command, `cchq <env> openvpn-activate-user <username>`, that sets the VPN password temporarily to the ansible user password, so that the user can log on to the VPN and change their password themselves. It runs `passwd -x 1 <username>` with the intention of forcing the user to reset their password after a day. I'm not sure whether that (1) actually prevents them from using the VPN after a day, though I think it should, and whether that (2) continues to force password reset policy of _every day_ after the user sets their password (I hope not!).


dannyroberts:
FWIW, provable deploy ran until it hit the 50 minute mark, which is the closest thing we have to passing for that test currently. It got a little further than it did on master https://github.com/dimagi/commcare-cloud/pull/2227.

environments/production/inventory.ini
```diff
@@ -331,3 +331,10 @@ hqdjangomanage0.internal-va.commcarehq.org
 
 [django_manage:children]
 hqdjangomanage0
+
+[openvpn]
+10.202.20.194  # ansible_host=18.235.223.2
```
pr33thi: Is the comment checked in on purpose?

dannyroberts: Yeah it is. It's the machine's public IP address, which you occasionally have to use if you get locked out, but is primarily part of the first-time set up of the VPN VM. It's autogenerated by the aws-fill-inventory script. See the comment here for how it's intended to be used in first-time setup: https://github.com/dimagi/commcare-cloud/pull/2222/files#diff-3873a0715ac8c9224b7b1992e21bf79cR107

src/commcare_cloud/terraform/README.md
```diff
@@ -93,12 +93,24 @@ Now you will be able to SSH into the VM.
...
 
-### Run the ovpn-init script
+Finally, make sure to run
+
+```bash
+cchq <env> aws-fill-inventory
+```
+
+which will auto-generate an `[openvpn]` section to your inventory.ini.
+For this to work, make sure you're using then inventory templating style. If you aren't,
```
pr33thi: [nit] `then` -> `the`

src/commcare_cloud/terraform/README.md
```diff
@@ -118,33 +130,47 @@ To give others SSH access to the VPN machine
...
 
 Then run
 ```
-# printf "[openvpn]\n<vpn-public-ip> subdomain_name=<vpn-subdomain-name>" > openvpn-tmp
-cchq <env> ansible-playbook deploy_openvpn.yml --skip-check -vvv -i openvpn-tmp
+cchq <env> ansible-playbook openvpn_playbooks/create_openvpn_cert.yml --skip-check -vvv
 ```
 
-### Create a user from the Admin Web UI
+### Enable PAM in the web Admin UI
```
pr33thi: Can you put what PAM means somewhere here?

PR #2224, opened 2018-10-05
Update the list of queues that cannot share a process

pr33thi:
I'll describe how I found this error which I think will explain what the error is.
[This task](https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml#L39) kept getting skipped on this `when` clause

So I added in this debug statement right before that task: `- debug: msg="{{ app_processes_config.celery_processes.get(inventory_hostname) }}"`, and the output was:
```
ok: [10.0.0.23] => {
    "msg": {
        "background_queue,case_rule_queue,celery,analytics_queue": {
            "concurrency": 1,
            "max_tasks_per_child": 50,
            "num_workers": 1,
            "optimize": false,
            "pooling": "prefork"
        },
        "beat": {
            "concurrency": 1,
            "max_tasks_per_child": 50,
            "num_workers": 1,
            "optimize": false,
            "pooling": "prefork"
        },
        "email_queue,reminder_case_update_queue": {
            "concurrency": 1,
            "max_tasks_per_child": 50,
            "num_workers": 1,
            "optimize": false,
            "pooling": "prefork"
        },
        "flower": {
            "concurrency": 1,
            "max_tasks_per_child": 50,
            "num_workers": 1,
            "optimize": false,
            "pooling": "prefork"
        },
        "reminder_rule_queue,repeat_record_queue,saved_exports_queue": {
            "concurrency": 1,
            "max_tasks_per_child": 50,
            "num_workers": 1,
            "optimize": false,
            "pooling": "prefork"
        },
        "sms_queue,export_download_queue": {
            "concurrency": 1,
            "max_tasks_per_child": 50,
            "num_workers": 1,
            "optimize": false,
            "pooling": "prefork"
        }
    }
}
```

What I got from that is that all the queues that have their own tasks in celery.yml cannot be grouped with other queues.

This fix breaks a bunch of tests, because this is currently being violated in every env specified in danny's comment!

@dannyroberts 
code buddy @jmtroth0

### DIFF
--- a/.travis/environments/travis/app-processes.yml
+++ b/.travis/environments/travis/app-processes.yml
@@ -3,9 +3,13 @@ celery_processes:
     celery_periodic:
     async_restore_queue,background_queue,case_rule_queue,celery,export_download_queue:
     email_queue,reminder_case_update_queue,analytics_queue:
-    reminder_rule_queue,reminder_queue,repeat_record_queue,saved_exports_queue:
-    sms_queue,submission_reprocessing_queue,ucr_indicator_queue,ucr_queue:
+    reminder_rule_queue,repeat_record_queue,saved_exports_queue:
+    ucr_indicator_queue,ucr_queue:
     flower:
+    submission_reprocessing_queue:
+    sms_queue:
+    reminder_queue:
+
   celery:
 pillows:
   pillowtop:
--- a/commcare-cloud-bootstrap/environment/app-processes.yml.j2
+++ b/commcare-cloud-bootstrap/environment/app-processes.yml.j2
@@ -1,17 +1,16 @@
 celery_processes:
   {{ celery_host.name }}:
     beat: {}
+    sms_queue: {}
+    flower: {}
     background_queue,case_rule_queue,celery,analytics_queue:
       concurrency: 1
-    email_queue,reminder_case_update_queue:
+    email_queue,reminder_case_update_queue,export_download_queue:
       concurrency: 1
     reminder_rule_queue,repeat_record_queue,saved_exports_queue:
       concurrency: 1
-    sms_queue,export_download_queue:
-      concurrency: 1
 #    reminder_queue,submission_reprocessing_queue,ucr_indicator_queue,ucr_queue,celery_periodic:
 #      concurrency: 1
-    flower: {}
 pillows:
   {{ pillowtop_host.name }}:
     kafka-ucr-main:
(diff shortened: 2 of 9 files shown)


pr33thi:
@dannyroberts 
Yep, I updated it, but your comment still stands! I'll share on #celery .

dannyroberts:
@calellowitz I think this is about only things that piggyback off the celery config, but aren't actually run by celery, like "sms_queue" which isn't a celery queue but rather a management command run with `manage.py run_sms_queue`.

calellowitz:
@dannyroberts I see now. That makes sense that they can't share a process if they are just management commands. I am not even sure how that would be possible, but why are they part of `celery_processes` if they are not in fact celery processes.

dannyroberts:
@calellowitz I agree with the implication of your question, which is that they shouldn't be. This is something that goes back a long time, definitely before `commcare-cloud`, and before we had really attempted to break these things out nicely. Presumably at the time, this was one of the only ways to declare where individual processes should run, and whoever was adding these queues piggybacked off it.

That's clearly not a great way to do things, so thanks for underscoring that, as it's something I had sort of just gotten used to. I'm going to make an issue to separate these out.

dannyroberts:
Any progress here?

I think for the list of issues I pulled from the error
> 64-test reminder_queue
> development reminder_queue
> enikshay-reference reminder_queue
> pna submission_reprocessing_queue
> swiss sms_queue

none of the three mentioned "queues" (reminder_queue, submission_reprocessing_queue, sms_queue) is required ([example](https://github.com/dimagi/commcare-cloud/blob/b6b44c4dd26bdf0ac4fc19f1a81be162e0209414/src/commcare_cloud/environment/schemas/app_processes.py#L82)). So I think the easiest fix here, which should have no effect on what's actually happening on those environments, where the mentioned queues are effectively disabled, would be to simply remove those queues from the app_processes.yml.

pr33thi:
@dannyroberts thanks for grepping for the queues it errored on! 
PR: https://github.com/dimagi/commcare-cloud/pull/2307

pr33thi:
FYI, the tests are passing, let me know if this is good to merge!
(@dannyroberts @calellowitz)

pr33thi:
Btw, the tests have passed, let me know if this is good for merging!
(@dannyroberts @calellowitz)

PR #2229, opened 2018-10-08
Use cleaner ansible calls

dannyroberts:
This makes run-module and ansible-playbook commands use cleaner shell calls that allow those commands to act interactively. This allows us for example to put prompts in ansible playbooks. It's also just a more straightforward subprocess call.

We previously "had" to pass in the vault password via stdin because I couldn't think of a way to pass it in as an environment variable. This uses the same idea of passing in an executable to `--ansible-password-file`, but not the executable prints out the value of `ANSIBLE_VAULT_PASSWORD` rather than the output of stdin (which is what we were using `/bin/cat` to do).


dannyroberts:
Yes, I would _love_ to do that at some point!

PR #2235, opened 2018-10-09
[pv-deploy] Call provable deploy running for 45 minutes a success

dannyroberts:
I think this nice built-in `timeout` should do what we were discussing earlier, about killing provable deploy after 45 minutes without failing the build, but still failing if it exits earlier with a non-zero status code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ env:
 
 matrix:
   fast_finish: true
-  allow_failures:
-    - env: TEST=prove-deploy
 
 before_install:
   # The next line exits out of the prove-deploy test if the last commit that was merged in doesn't contain [pv-test]
--- a/.travis/tests.sh
+++ b/.travis/tests.sh
@@ -39,7 +39,17 @@ then
     bootstrap() {
         ssh-keygen -f ~/.ssh/id_rsa -N "" -q
         cp ~/.ssh/id_rsa.pub .travis/environments/_authorized_keys/travis.pub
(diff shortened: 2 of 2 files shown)


dannyroberts:
If after this runs, I'll push a second commit, changing provable-deploy to _not_ be an allowed failure

dannyroberts:
The build is failing with (real, I think) error:
```
TASK [andrewrothstein.couchdb : build...] **************************************
fatal: [10.0.0.134]: FAILED! => {"changed": true, "cmd": ["make", "release"], "delta": "0:00:00.541629", "end": "2018-10-09 15:20:54.412530", "msg": "non-zero return code", "rc": 2, "start": "2018-10-09 15:20:53.870901", "stderr": "ERROR: OTP release R16B03 does not match required regex 17|18|19|20\nERROR: compile failed while processing /usr/local/couchdb2/src/apache-couchdb-2.2.0/src/config: rebar_abort\nmake: *** [couch] Error 1", "stderr_lines": ["ERROR: OTP release R16B03 does not match required regex 17|18|19|20", "ERROR: compile failed while processing /usr/local/couchdb2/src/apache-couchdb-2.2.0/src/config: rebar_abort", "make: *** [couch] Error 1"], "stdout": "==> config (compile)", "stdout_lines": ["==> config (compile)"]}
```
Will go ahead now and make this not an allowed failure. The next build should fail for that reason.

orangejenny:
@dannyroberts Aha, got it - way to go :)

PR #2236, opened 2018-10-09
Use cheaper and better AWS t3 instance class

dannyroberts:
and always use cpu_credits = "unlimited"

This lets us elastically pay for more CPU usage when it's needed with no cap. If things are well calibrated this shouldn't cost very much but will save us from bursty performance issues; and if it's expensive one month we can upgrade the machine type for more provisioned CPU. (The change to t3 is just upgrading to the new generation which is both better and cheaper.)

### DIFF
--- a/src/commcare_cloud/terraform/modules/servers/main.tf
+++ b/src/commcare_cloud/terraform/modules/servers/main.tf
@@ -8,6 +8,9 @@ resource aws_instance "server" {
   key_name                = "${var.key_name}"
   vpc_security_group_ids  = ["${var.security_group_options[lookup(var.servers[count.index], "network_tier")]}"]
   source_dest_check       = false
+  credit_specification {
+    cpu_credits = "unlimited"
+  }
   root_block_device {
     volume_size           = "${lookup(var.servers[count.index], "volume_size")}"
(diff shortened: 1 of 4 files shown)


src/commcare_cloud/terraform/modules/servers/main.tf
```diff
@@ -8,6 +8,9 @@ resource aws_instance "server" {
   key_name                = "${var.key_name}"
   vpc_security_group_ids  = ["${var.security_group_options[lookup(var.servers[count.index], "network_tier")]}"]
   source_dest_check       = false
+  credit_specification {
```
pr33thi: What does this mean?

dannyroberts: Yeah `cpu_credits` can be either "unlimited" or "standard"; when it's "standard" your instance type comes with a certain number of CPU bust credits that'll it'll use when necessary until they run out; with "unlimited", it'll use them as necessary until they run out, and then start charging $.05/hr for extra usage of burst CPU. See https://aws.amazon.com/ec2/instance-types/t3/ (and t2 has the same concept) for more info.

PR #2237, opened 2018-10-09
Key iam users by name in terraform

dannyroberts:
and add an AWS key pair for each user.

This change has already been applied to staging and production. It'll make it so that deleting or adding a user doesn't shift everything off by one (common terraform problem when you build things from a list, incredibly!)

### DIFF
--- a/src/commcare_cloud/commands/terraform/entrypoint.tf.j2
+++ b/src/commcare_cloud/commands/terraform/entrypoint.tf.j2
@@ -108,10 +108,12 @@ module "commcarehq" {
   }
 
   account_alias = {{ (account_alias or "")|tojson }}
-
-  users = [
-  {%- for user in (users if manage_users else []) %}
-    {username = {{ user.username|tojson }}},
-  {%- endfor %}
-  ]
 }
+
+{%- for user in (users if manage_users else []) %}
+module "iam_user__{{ user.username }}" {
+  source = "../modules/iam/user"
+  username = {{ user.username|tojson }}
+  administrators_iam_group = "${module.commcarehq.administrators_iam_group}"
+}
+{%- endfor %}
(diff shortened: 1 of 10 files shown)


src/commcare_cloud/commands/terraform/entrypoint.tf.j2
```diff
@@ -108,10 +107,17 @@ module "commcarehq" {
   }
 
   account_alias = {{ (account_alias or "")|tojson }}
+}
```
pr33thi: is there a stray unclosed bracket somewhere above?

src/commcare_cloud/commands/terraform/entrypoint.tf.j2
```diff
@@ -108,10 +107,17 @@ module "commcarehq" {
...
-    {username = {{ user.username|tojson }}},
-  {%- endfor %}
-  ]
+{%- for user in (users if manage_users else []) %}
+resource "aws_key_pair" {{ user.username|tojson }} {
+  key_name = {{ user.username|tojson }}
+  public_key = {{ user.public_key|tojson }}
+}
+
+module "iam_user__{{ user.username }}" {
```
pr33thi: Why 2 underscores?

dannyroberts: Eh, I just figured it would be a nice way to indicate separation of the static part of the name and the dynamic part of the name.

src/commcare_cloud/commands/terraform/entrypoint.tf.j2
dannyroberts: Nope, this is just the diff making it look more complicated than it is. Imagine that this line was a white `}` with out a +, and this line https://github.com/dimagi/commcare-cloud/pull/2237/files#diff-38c55dbcbe9e3c73f5b5c18b14b5aba1R122 were a green line with a plus. Then it might make more sense.

PR #2239, opened 2018-10-09
Create AWS production Elasticsearch machines

dannyroberts:
without yet deploying elasticsearch to them.

Already created the machines, will work with Farid tomorrow to deploy ES to them and have them join production ES cluster using zone awareness to put a replica of each shard aws (rather than distributing shards haphazardly between the two data centers). That'll be the most painless test of whether the VMs between the two clusters can cluster together, which I'm fairly confident they'll be able to without major issue.


environments/production/inventory.ini
```diff
@@ -188,6 +188,32 @@ hqes7.internal-va.commcarehq.org elasticsearch_node_name=hqes7
 [hqesmaster0]
 172.24.32.67 hostname=hqesmaster0 elasticsearch_node_name=hqesmaster0
 
+[es0]
+10.202.41.22 hostname=es0-production elasticsearch_node_name=es0 ec2=yes
```
proteusvacuum: any reason not to follow the `hqes` convention?

dannyroberts: This came from a discussion here https://github.com/dimagi/commcare-cloud/pull/1894#discussion_r200327569, where we decided to go with these simpler names (`es0`, `web0`, etc.) for the groups, and `{group}-{env}` for the hostnames. We're basically using the fact that we're going to recreate every machine in the environment for the AWS migration to also switch to this naming convention.

proteusvacuum: great, thanks for the explanation.

environments/production/inventory.ini
```diff
@@ -188,6 +188,32 @@ hqes7.internal-va.commcarehq.org elasticsearch_node_name=hqes7
...
+
+[es1]
+10.202.41.96 hostname=es1-production elasticsearch_node_name=es1 ec2=yes
```
proteusvacuum: I think we'll want to set the `elasticsearch_node_zone` on all of the ES nodes (new and old) before deploying this:

https://github.com/dimagi/commcare-cloud/blob/7809a83a88ff77461aad3bbcba851409b2922733/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2#L10-L14

dannyroberts: Yup, I agree. That's the next step that I wanted to discuss in more detail with you this afternoon

dannyroberts: But I can add that piece of it now

dannyroberts: Oh, ha. Jokes. That's already in the PR under `[es_data_aws:vars]`

proteusvacuum: hah, oops, missed that  :mag:

PR #2241, opened 2018-10-09
Add AWS staging proxy

dannyroberts:
Not yet fully deployed, but will use this to replace current staging proxy tomorrow.


dannyroberts:
I also had to update the control machine and change IP addresses (turned out I'd set the boot disk way to small and it filled up making the machine unresponsive)

pr33thi:
@nickpell FYI, this was the fix to both of us being unable to ssh onto the staging control machine!

PR #2244, opened 2018-10-10
Make host_group_aliases.yml result in ok, not changed

dannyroberts:
Another small thing I noticed that was annoying me. I recently added these tasks to dynamically create the `all_commcarehq` and `commcarehq` host groups so that playbooks can reference these and avoid copy-and-paste, but these in-memory tasks were always showing up as "changed". This will make them show up as "ok" instead.


dannyroberts:
Welcome! And in case it got lost from lack of context: this file is an ansible playbook (one that gets included in other ansible playbooks), more "code" than "config" in my opinion despite the .yml format.

orangejenny:
> more "code" than "config" in my opinion

interesting

PR #2246, opened 2018-10-10
Streamline deploy control

dannyroberts:
@pr33thi this is from our work on Monday—I think I just never PR'd.


pr33thi:
And this is because deploy_control had a bunch of extra stuff, so we pared it down to just what was necessary right?

dannyroberts:
Yup!

PR #2247, opened 2018-10-10
Include proxy.public_ip in ALLOWED_HOSTS

dannyroberts:
This comes up every time we're switching proxies, since you'll want to test on the public IP before flipping the DNS

### DIFF
--- a/environments/staging-aws-test/public.yml
+++ b/environments/staging-aws-test/public.yml
@@ -40,10 +40,8 @@ s3_blob_db_s3_bucket: 'dimagi-commcare-staging-test-blobdb'
 
 localsettings:
   ALLOWED_HOSTS:
-    - staging.commcarehq.org
     - j2mestaging.commcarehq.org
     - testserver
-    - '{{ hostvars[groups.proxy.0].public_ip }}'
   ANALYTICS_DEBUG: True
   ANALYTICS_LOG_LEVEL: "debug"
   BANK_ADDRESS: { 'first_line': "1 Citizens Drive", 'city': "Riverside", 'region': "RI", 'postal_code': "02915" }
--- a/environments/icds-cas/public.yml
+++ b/environments/icds-cas/public.yml
@@ -130,7 +130,6 @@ s3_blob_db_s3_bucket: 'blobdb'
 localsettings:
   ALLOWED_HOSTS:
     - '{{ CAS_SITE_HOST }}'
-    - '{{ hostvars[groups.proxy.0].public_ip }}'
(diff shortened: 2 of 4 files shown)


environments/staging-aws-test/public.yml
```diff
@@ -40,10 +40,8 @@ s3_blob_db_s3_bucket: 'dimagi-commcare-staging-test-blobdb'
 
 localsettings:
   ALLOWED_HOSTS:
-    - staging.commcarehq.org
```
orangejenny: Why does this one also get removed?

pr33thi:
What do you mean by 
> flipping the DNS
?

dannyroberts:
> What do you mean by flipping the DNS?

I mean changing the DNS record (e.g. "staging.commcarehq.org") from pointing to the public IP of the old proxy machine to the public IP of the new proxy machine. People sometimes use the word "flip" for things like this, perhaps in analogy to flipping a light switch.

environments/staging-aws-test/public.yml
dannyroberts: It could have been removed before this PR as well—this isn't the actual value of SITE_HOST for this environment, so it was unnecessary (but harmless) to include (and SITE_HOST is automatically included, so that's uncessary to include too).

pr33thi:
I approved but in retrospect I think I still don't follow what's happening.. can I get a high level summary? What is `ALLOWED_HOSTS` for?

dannyroberts:
@pr33thi `ALLOWED_HOSTS` is a django setting https://docs.djangoproject.com/en/2.1/ref/settings/#allowed-hosts.

For any django server, there may be multiple ways to "access" it over http. For example, I could log into the machine and run `curl localhost:9010`. Or I could from another machine run `curl 10.x.x.x:9010`, where "10.x.x.x" is the machine's imaginary IP address. Or I could curl the _proxy_'s IP address, and it'll pass that to the web worker. In any of these situations, that thing "localhost", "10.x.x.x", the proxy's IP address, etc. will show up as the `Host` header in the http request. Django checks that header against ALLOWED_HOSTS: if it's in it, it continues with the request; if it's not, it response with status 400 Bad Request.

Normally of course, we make the request to e.g. https://www.commcarehq.org, in which case the `Host` header will be `www.commcarehq.org`. But if we want to be able to debug, it's useful to be able to make requests before the DNS entry is set up, or before we've changed it to point to the new location. So, the idea is to put anything that we could reasonably expect one of us to want to make the request to (localhost, private IP, public IP, DNS, etc.) to help us debug.

PR #2251, opened 2018-10-10
Fix rolling restart

dannyroberts:
@pr33thi there were a number of problems with the implementation as it was:

- A restart would stop _all_ the nodes, and then start _all_ the nodes, in that order, causing lots of downtime, instead of stopping and starting each node in sequence, which wouldn't cause downtime
- it didn't respect `--limit` (but didn't complain either) making it so there's no way to restart only a subset of nodes
- (minor) it was a bit chatty and asked for too many `y`s to keep going at different points

### DIFF
--- a/src/commcare_cloud/commands/ansible/service.py
+++ b/src/commcare_cloud/commands/ansible/service.py
@@ -325,12 +325,15 @@ class Elasticsearch(ServiceBase):
                     "\nStart will start pillows and start elasticsearch "
                     "\nRestart is a stop followed by a start.\n Continue?", strict=False):
                 return 0  # exit code
-            if action == 'stop' or action == 'restart':
+            if action == 'stop':
                 self._act_on_pillows(action='stop')
                 self._run_rolling_restart_yml(tags='action_stop')
-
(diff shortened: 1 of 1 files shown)


dannyroberts:
By passing both tags in, it runs every task in the playbook, since every task has one tag or the other. At a higher level if there were nodes es0, es1, and es2, then previously the steps were:

```
stop pillows
stop es0
# right now 1 node is down, which is fine assuming there's 1 replica
stop es1
# right now 2 nodes are down!
stop es2
# right now all nodes are down!
start es0
# right now 2 nodes are still down!
start es1
# right now only 1 node is down again, so we're ok in theory
start es2
start pillows
```

Now the steps are

```
stop pillows
stop es0
start es0
# es0 is restarted, with no downtime assuming there's 1 replica
stop es1
start es1
stop es2
start es2
start pillows
```
The "rolling" in in a "rolling restart" refers to restarting them one at a time. If the cluster is set to keep 1 replica, that means it can handle any 1 node being down at a time, so a rolling restart is then a no-downtime restart, which is the motivation.

PR #2255, opened 2018-10-11
Allow specifying AZs other than a, b, c in terraform

dannyroberts:
I started doing this because I thought it would be useful for ES, but then realized it was kind of pointless to prioritize. (Thanks Farid!)

Still worth PRing because it's still a good change. The only part that makes it incomplete is the part I commented on, but I think even with that blemish, this change is still an improvement.

### DIFF
--- a/src/commcare_cloud/terraform/modules/commcarehq/main.tf
+++ b/src/commcare_cloud/terraform/modules/commcarehq/main.tf
@@ -16,6 +16,7 @@ module "network" {
   vpc_begin_range   = "${var.vpc_begin_range}"
   env               = "${var.environment}"
   azs               = "${var.azs}"
+  az_codes          = "${var.az_codes}"
   vpn_connections   = "${var.vpn_connections}"
   vpn_connection_routes = "${var.vpn_connection_routes}"
   external_routes   = "${var.external_routes}"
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/terraform/modules/commcarehq/main.tf
```diff
@@ -24,15 +25,18 @@ module "network" {
...
+    app-private-c = "${lookup(module.network.subnets-app-private, "c", "")}"
+    app-private-f = "${lookup(module.network.subnets-app-private, "f", "")}"
+    db-private-a = "${lookup(module.network.subnets-db-private, "a", "")}"
+    db-private-b = "${lookup(module.network.subnets-db-private, "b", "")}"
+    db-private-c = "${lookup(module.network.subnets-db-private, "c", "")}"
+    db-private-f = "${lookup(module.network.subnets-db-private, "f", "")}"
+    public-a = "${lookup(module.network.subnets-public, "a", "")}"
+    public-b = "${lookup(module.network.subnets-public, "b", "")}"
+    public-c = "${lookup(module.network.subnets-public, "c", "")}"
+    public-f = "${lookup(module.network.subnets-public, "f", "")}"
```
dannyroberts: As an example of accepting a new zone, 'f', I added new lines here. It'd be nice, and I _think_ possible but annoying to do in terraform, to have this map get computed from the other variables (like `az_codes`), but this works, is easy to add to if we need to use it in a pinch, and is still more flexible than it was before at least, requiring _less_ of a change to support a new AZ.

pr33thi:
AZ = Availability Zone?

dannyroberts:
Yeah sorry, AZ = availability zone. AWS jargon.

PR #2257, opened 2018-10-11
create separate process for icds_aggregation_queue on softlayer

nickpell:
ICDS aggregation tasks are causing a celery backup on softlayer:

https://app.datadoghq.com/dash/141098/celery-overview?live=true&page=0&is_auto=false&from_ts=1539208017772&to_ts=1539294417772&tile_size=m&fullscreen=false&tpl_var_environment=softlayer

### DIFF
--- a/environments/softlayer/app-processes.yml
+++ b/environments/softlayer/app-processes.yml
@@ -42,7 +42,8 @@ celery_processes:
       concurrency: 2
       max_tasks_per_child: 1
     icds_aggregation_queue:
-      concurrency: 1
+      pooling: gevent
+      concurrency: 2
     flower: {}
 pillows:
   'pillow1':
--- a/environments/softlayer/app-processes.yml
+++ b/environments/softlayer/app-processes.yml
@@ -38,9 +38,11 @@ celery_processes:
       concurrency: 10
     async_restore_queue:
       concurrency: 1
-    background_queue,case_rule_queue,icds_dashboard_reports_queue,icds_aggregation_queue,analytics_queue:
+    background_queue,case_rule_queue,icds_dashboard_reports_queue,analytics_queue:
       concurrency: 2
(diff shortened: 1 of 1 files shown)


environments/softlayer/app-processes.yml
```diff
@@ -38,9 +38,11 @@ celery_processes:
       concurrency: 10
     async_restore_queue:
       concurrency: 1
-    background_queue,case_rule_queue,icds_dashboard_reports_queue,icds_aggregation_queue,analytics_queue:
+    background_queue,case_rule_queue,icds_dashboard_reports_queue,analytics_queue:
       concurrency: 2
       max_tasks_per_child: 1
+    icds_aggregation_queue:
+      concurrency: 1
```
calellowitz: I think this actually needs to be at least two, since one main task spawns a bunch of subtasks.

calellowitz: This could also be a gevent queue since it is mostly I/O bound.

nickpell: Could you elaborate more on these two points?

calellowitz: For the ICDS aggregation, there is one main task, that spawns a large number of subtasks and waits for them to complete. If the queue only has concurrency 1, then those subtasks will never get started, and the main task will never complete resulting in a deadlock. We need at least concurrency two so that one slot can be held by the main tasks, and the other can be taken by the sub tasks.

The second comment was just once this queue had concurrency > 1 that instead of making it a prefork queue (the defualt), we could do concurrency via gevent which tends to work better in I/O heavy tasks (as opposed to prefork which is preferred for cpu heavy tasks)

nickpell: That all makes sense, thanks for explaining to me

PR #2263, opened 2018-10-14
Add easy way to call other commcare-cloud commands

dannyroberts:
`commcare_cloud('staging', 'fab', 'deploy')` is equivalent to `suprocess.call(['commcare-cloud', 'staging', 'fab', 'deploy'])`, except that it doesn't require forking a new process (which is in the case of python, _very_ expensive in terms of time).

Will test further but I think the general concept works.

I converted one usage just as an example, but if we like the pattern, I'll convert others as well. There's also some cleanup I can do with the way  `print_command` gets called in other places.

### DIFF
--- a/src/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/src/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -200,8 +200,11 @@ class UpdateConfig(CommandBase):
     This includes django `localsettings.py` and formplayer `application.properties`.
     """
 
+    arguments = (shared_args.BRANCH_ARG,)
+
     def run(self, args, unknown_args):
-        return commcare_cloud(args.environment, 'ansible-playbook', 'deploy_localsettings.yml', tags='localsettings', *unknown_args)
+        return commcare_cloud(args.environment, 'ansible-playbook', 'deploy_localsettings.yml',
+                              tags='localsettings', branch=args.branch, *unknown_args)
 
 
 class AfterReboot(_AnsiblePlaybookAlias):
--- a/src/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/src/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -203,7 +203,7 @@ class UpdateConfig(CommandBase):
     arguments = (shared_args.BRANCH_ARG,)
 
     def run(self, args, unknown_args):
-        return commcare_cloud(args.environment, 'ansible-playbook', 'deploy_localsettings.yml',
+        return commcare_cloud(args.env_name, 'ansible-playbook', 'deploy_localsettings.yml',
                               tags='localsettings', branch=args.branch, *unknown_args)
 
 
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -199,9 +201,7 @@ class UpdateConfig(_AnsiblePlaybookAlias):
     """
 
     def run(self, args, unknown_args):
-        args.playbook = 'deploy_localsettings.yml'
-        unknown_args += ('--tags=localsettings',)
-        return AnsiblePlaybook(self.parser).run(args, unknown_args)
+        return commcare_cloud(args.environment, 'ansible-playbook', 'deploy_localsettings.yml', tags='localsettings', *unknown_args)
```
snopoke: Can you articulate why you think this is better than the way we're doing it now?

dannyroberts: I think it's clearer and cleaner, and sticks to the public interface of a command. I've often found myself annoyed that if I want a command to be an alias for another bash command, then easy, I just call it with `subprocess.call`; but if I want a command to be an alias for another `commcare-cloud` command, then I have to go through this dance of reconstructing what a call to that command class's `run` method, which requires passing in a parser that doesn't really get used (so we arbitrarily use the one we have on hand), and `Namespace` object as `args`, so we modify the one we have on hand and use it.

This shortcut lets us call the command as we would from the command line, but with just a touch of syntactic sugar (but not enough to cause a syntactic blood sugar spike).

I found myself wanting this again for a new command, and feeling averse to doing the little dance again, so I figured I'd write it.

That's how I came to it. Does any of that resonate?

dannyroberts:
Found minor change I needed to make to have it play nicely with my (beloved) `--control` arg, and fixed a deprecation warning. Minor changes, but I've been scolded for sneaking changes in post review before, so requesting a final pass :)

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -190,18 +192,19 @@ def run(self, args, unknown_args):
...
     """
 
+    arguments = (shared_args.BRANCH_ARG,)
+
     def run(self, args, unknown_args):
-        args.playbook = 'deploy_localsettings.yml'
-        unknown_args += ('--tags=localsettings',)
-        return AnsiblePlaybook(self.parser).run(args, unknown_args)
+        return commcare_cloud(args.env_name, 'ansible-playbook', 'deploy_localsettings.yml',
+                              tags='localsettings', branch=args.branch, *unknown_args)
```
millerdev: Why did you decide to handle the `--branch` argument this way, but didn't add any of the other `ansible-playbook` args?

https://github.com/dimagi/commcare-cloud/blob/9639f0cde5e081318c1405d3270fac8819fc30cb/src/commcare_cloud/commands/ansible/ansible_playbook.py#L38-L43

When I run `cchq production update-config --help` I only see documentation for two arguments:
```
optional arguments:
  -h, --help
  --branch BRANCH
```
even though others are applicable and will be passed through to the underlying `ansible-playbook` command if provided.

I suppose one (not very elegant) way to work around this would be to include a note in the help string that all optional arguments for the `ansible-playbook` command are also accepted.

PR #2265, opened 2018-10-15
Added inventory for postgresql

NitigyaS:
Added new PostgreSQL standbys. 
@snopoke Kindly check if UCR standby is from the master or they need to be cascading.


snopoke:
I think we should make them cascade otherwise we're going to add too much load to the master

PR #2271, opened 2018-10-15
Soft delete es5-production by replacing with a tiny machine

dannyroberts:
I manually terminated this machine when it acted up, and deleting machines from this config is surprisingly annoying. Working on a better way to do this that's more sustainable.


dannyroberts:
Never mind, I'm going to fold this into a PR i'm about to make.

dannyroberts:
https://github.com/dimagi/commcare-cloud/pull/2272

PR #2273, opened 2018-10-16
Add terraform-migrate-state command

dannyroberts:
This PR adds a state migration system for terraform. This is something that should exist as a generic tool for terraform, but terraform is still not that mature, and it doesn't seem to exist yet.

Terraform assigns each resource an address so that it can map it back to the code.
However, often when you change the code, the addresses no longer map to the same place.
For this, terraform offers the `terraform state mv <address> <new_address>` command,
so you can tell it how existing resources map to your new code.

This is a tedious task, and often follows a very predictable renaming pattern.
So, this change introduces the beginnings of a system for specifying the renaming rule
and then having the state migrations automated, to reduce tedium and human error.


src/commcare_cloud/commands/terraform/terraform_migrate_state.py
```diff
@@ -0,0 +1,256 @@
...
+
+        wrapped as as a RemoteMigrationState object.
+        """
+
+        temp_filename = tempfile.mktemp()
+        try:
+            command = ['aws', 's3', 'cp', self.s3_uri, temp_filename]
+            print_command(command)
+            try:
+                subprocess.check_output(command, env=self.env_vars, stderr=subprocess.STDOUT)
```
snopoke: why not use `boto` here?

dannyroberts:
Okay will do. There are parts that should be easily testable, especially around the logic of applying moves which is "pure" and not dependent on interacting with the network or anything.

src/commcare_cloud/commands/terraform/terraform_migrate_state.py
dannyroberts: Yeah I probably should have. It's not currently in the install list, and I just didn't want to drop my headspace to figure out to do that and look up `boto3`'s API. But you're right, I should switch to that, and it should be pretty easy, so will do.

src/commcare_cloud/commands/terraform/terraform_migrate_state.py
```diff
@@ -0,0 +1,256 @@
...
+
+from commcare_cloud.alias import commcare_cloud
+from commcare_cloud.cli_utils import ask, print_command
+from commcare_cloud.commands.command_base import CommandBase, CommandError
+from commcare_cloud.environment.main import get_environment
+
+
+class TerraformMigrateState(CommandBase):
+    command = 'terraform-migrate-state'
+    help = "Apply unapplied state migrations in commcare_cloud/commands/terraform/migrations"
```
dannyroberts: @snopoke in terms of docs, I'm thinking just a better (and much longer) help message here would probably be the place to put it. Is that along the lines of what you were thinking?

snopoke: That's fine though I think we could do with some more general AWS docs to explain the overall setup.

dannyroberts:
OK! Got back to this after a couple weeks of fightfighting. Responded to the requests. While it's not the _most_ thoroughly tested thing, I added a basic test that would prevent a refactor from breaking the basic core planning logic, and most everything else is the glue that doesn't lend itself as nicely to being tested, and is obvious if it's not working. (And I'm going to be the only person using it for the first stretch of its existence, and I'm sure there are tweaks to come.)

snopoke:
```

  File "/home/travis/virtualenv/python2.7.14/bin/commcare-cloud", line 11, in <module>
    load_entry_point('commcare-cloud', 'console_scripts', 'commcare-cloud')()
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pkg_resources/__init__.py", line 572, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2752, in load_entry_point
    return ep.load()
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2405, in load
    return self.resolve()
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2411, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/travis/build/dimagi/commcare-cloud/src/commcare_cloud/commcare_cloud.py", line 20, in <module>
    from commcare_cloud.commands.terraform.terraform_migrate_state import TerraformMigrateState
  File "/home/travis/build/dimagi/commcare-cloud/src/commcare_cloud/commands/terraform/terraform_migrate_state.py", line 7, in <module>
    import boto3
ImportError: No module named boto3
```

dannyroberts:
Thanks, fixed!

src/commcare_cloud/commands/terraform/migrations/0000_example.py
```diff
@@ -0,0 +1,11 @@
+def get_new_resource_address(environment, old_resource_address):
+    """
+    Specify migration by specifying for a given old address what the new address should be
+
+    All addresses will be / should be given in the format used in `terraform state list`.
+
+    The migration is then essentially (in peudocode, and with better edge-case management):
```
pr33thi: [nit] typo `peudocode`

pr33thi:
>each resource an address so that it can map it back to the code.
However, often when you change the code, the addresses no longer map to the same place

Is this referring to the ordering of the machines in terraform.yaml or something else?

dannyroberts:
Reordering machines in terraform is one example of how addresses can get renamed, although I'm planning on addressing that particular issue in a different way: https://github.com/dimagi/commcare-cloud/pull/2375.

I'm talking about moving where a resource is declared in the terraform code or changing its name. For example, if something used to be declared in a module and then it's moved to the top level, it would now have a new address, so you'd have to remap the saved state (which is saved in a shared S3 bucket) so that the existing resources will map to the right new address

src/commcare_cloud/commands/terraform/migrations/0000_example.py
dannyroberts: Thanks

PR #2274, opened 2018-10-16
Fixed Ansible not noticing new line after condition

NitigyaS:
It messed up the SSH config and resulted in SSH server failed and not reachable. 

```
-UsePAM yes
-
-Match User unixadmin
+UsePAM yesMatch User unixadmin

```


src/commcare_cloud/ansible/roles/ssh/templates/sshd_config.j2
```diff
@@ -88,6 +88,7 @@ Subsystem sftp /usr/lib/openssh/sftp-server
 UsePAM yes
 
 {%- for user in ssh_allow_password_users %}
```
snopoke: you could just delete the `-` in `{%-`

PR #2276, opened 2018-10-16
Hardcode 8 gunicorn workers for production

dannyroberts:
The AWS ones are currently running at 2 even because of the default CPU calculation. They're on t3.large machines which have "burstable" CPU, but present to the OS as having two CPUs. I'm honestly not sure what that translates to, but I think it's worth upping it to 8 workers and seeing the effect.

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -2,7 +2,8 @@ newrelic_djangoagent: False
 newrelic_javaagent: True
 datadog_pythonagent: True
 django_command_prefix: '{{ virtualenv_home }}/bin/ddtrace-run '
-
+gunicorn_workers_static_factor: 8
+gunicorn_workers_factor: 0
 celery_processes:
   hqcelery0.internal-va.commcarehq.org:
     beat: {}


millerdev:
Would be interesting to know how AWS expects "burstable" CPUs to be used. Is it a thing that an application can dynamically change based on some load indicator (and how is that expected to happen)? Or is it a thing that is expected to be tweaked externally to the app itself (seems unlikely, given AWS attitude about automation).

dannyroberts:
Here's what AWS has to say about the t3 instance class and burstable CPU: https://aws.amazon.com/ec2/instance-types/t3/. I should clarify that to my knowledge it will always present to the OS as having 2 CPUs, so we won't be able to parallelize above that at the OS level; I think it's about the amount performance you get out of that virtual CPU.

millerdev:
If I'm understanding this correctly, what you did in this PR will allocate a higher number of gunicorn workers per CPU: OS will always report 2, but overriding that to 8 means we'll have 4x gunicorn workers per CPU on these hosts. The theory is that doing this will more effectively use burstable CPU power in times of high load. Does that all sound right?

Is there any chance that doing this will cause us to be billed more when load is low? I'm not sure how to quantify that. Maybe billed more than we would be without this PR and yet experience the same level of performance?

dannyroberts:
I don't think it'll cause us to be charged when load is _low_, but it very well might cause us to be charged when load is _high_. It takes some trial and error and calibration to figure out how to best use bursting credits. You can read more about it in places like https://aws.amazon.com/ec2/instance-types/t3/, but basically during low usage you accumulate cpu credits that you can use when you need them (for free), and then (if you're on "unlimited", which these machines are) you can continue to use high CPU as needed but you pay for it by the hour.

The current configuration may not be optimal for getting the most throughput per dollar, but that's something we can continue to optimize.

PR #2278, opened 2018-10-16
[pv-deploy] Add andrewrothstein.couchdb back to galaxy requirements

dannyroberts:
prove-deploy failures seem to be related to couch, so I wanted to run this with part of https://github.com/dimagi/commcare-cloud/pull/2215 rolled back to see if it's related.

### DIFF
--- a/src/commcare_cloud/ansible/requirements.yml
+++ b/src/commcare_cloud/ansible/requirements.yml
@@ -1,6 +1,9 @@
 ---
 - version: v1.1.4
   name: andrewrothstein.couchdb-cluster
+- src: git+https://github.com/andrewrothstein/ansible-couchdb.git
+  version: f0e026e73b8bf6f4ee5cc710f8fdbcdeb7418b2b
+  name: andrewrothstein.couchdb
 - src: git+https://github.com/debops-contrib/ansible-etckeeper.git
   version: 29c721de3df965ff355cbed242db813f70b811a2
   name: debops-contrib.etckeeper


dannyroberts:
@sravfeyn yeah it did, but check out my last comment I made probably while you were posting. The "normal" build is passing, but the "prove-deploy" build, the special one that only runs as a CRON job or when you start your commit message with `[pv-deploy]` failed. It was previously failing on the `[andrewrothstein.couchdb : build...]` you were seeing fail, but now it's getting further than that and failing at a later step. So this PR is a strict improvement even though this special optional test is still failing.

sravfeyn:
Ha, cool. Thanks for the note.

src/commcare_cloud/ansible/requirements.yml
```diff
@@ -1,6 +1,9 @@
 ---
 - version: v1.1.4
   name: andrewrothstein.couchdb-cluster
+- src: git+https://github.com/andrewrothstein/ansible-couchdb.git
+  version: f0e026e73b8bf6f4ee5cc710f8fdbcdeb7418b2b
+  name: andrewrothstein.couchdb
```
snopoke: Why can't we just `couchdb_ver` instead of pinning to an old version?

dannyroberts: We probably can! I actually didn't understand what I was doing so much as I was undoing a part of https://github.com/dimagi/commcare-cloud/pull/2215 that looked to me to be suspiciously related.

Is the change you're suggesting basically
- undo this commit (don't include andrewrothstein.couchdb as an explicit dependency here)
- add `couchdb_ver: '2.1.1'` to `all.yml`
? (I got `2.1.1` from the default for that variable in the version pinned in this PR (f0e026e73b8bf6f4ee5cc710f8fdbcdeb7418b2b): https://github.com/andrewrothstein/ansible-couchdb/blob/f0e026e73b8bf6f4ee5cc710f8fdbcdeb7418b2b/defaults/main.yml#L2)

PR #2209, opened 2018-10-02
configurable per env if haproxy or nginx is used for couch proxy

snopoke:
Once fully rolled out we can remove nginx references

Some conflicts with https://github.com/dimagi/commcare-cloud/pull/2207 which I'll resolve once that's merged

PR #2281, opened 2018-10-17
[pv-deploy] Fix couchdb_use_haproxy undefined

dannyroberts:
this issue appears in prove-deploy test: https://api.travis-ci.org/v3/job/442292654/log.txt

I think this full stack deploy bug was introduced in this PR https://github.com/dimagi/commcare-cloud/pull/2209, and I think this fixes it without breaking the positive use case. Running a prove-deploy build to confirm the former.

I'm pretty sure that `public.yml` will win if it's defined there (https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) and that the other interpolations in this file that use `couchdb_use_haproxy` will use the overridden value from `public.yml` and not the default value just because it's defined in the same file, but would appreciate someone if you could double check my logic @snopoke so I don't by mistake cause issues where it _is_ used (like ICDS).

### DIFF
--- a/.travis/tests.sh
+++ b/.travis/tests.sh
@@ -45,15 +45,16 @@ then
         cp ~/.ssh/id_rsa.pub .travis/environments/_authorized_keys/travis.pub
         (COMMCARE_CLOUD_ENVIRONMENTS=.travis/environments \
             timeout 45m \
-            bash commcare-cloud-bootstrap/bootstrap.sh hq-${TRAVIS_COMMIT} ${BRANCH} .travis/spec.yml)
-        rc=$?
-        if [[ "${rc}" = 124 ]]
-        then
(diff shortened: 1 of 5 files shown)


dannyroberts:
First run was a success (in terms of fixing the issue I'd mentioned), and revealed another error down the line. Added a commit hopefully fixing that as well; running another prove-deploy

snopoke:
Thanks, I have always found it a bit odd that we put nginx logs in `/home/cchq`

dannyroberts:
@snopoke Ha, yeah true, me too. There's enough stuff (and dev muscle memory) built around that that we probably won't change that until we make some bigger change

dannyroberts:
Provable deploy test passes! Good to merge once I get an OK on review

PR #2283, opened 2018-10-17
[pv-deploy] Set couch_ver rather than pin to old galaxy requirement version

dannyroberts:
Followup to suggestion on https://github.com/dimagi/commcare-cloud/pull/2278


dannyroberts:
Still gets past the original error that the last PR was meant to fix, so this is good to go. (Actually fixing the rest of the prove-deploy build here: https://github.com/dimagi/commcare-cloud/pull/2281)

PR #2288, opened 2018-10-18
attempt to speed up

snopoke:
This works and is a bit faster than the previous version
but not enough to make it worth while.

@dannyroberts I'm wondering if there isn't a way we can improve this, for ICDS we can incorporate it into the CSV inventory but not so for the other envs.


dannyroberts:
Very nice! Did you test this somehow? Worth testing with [pv-deploy]?

dannyroberts:
> but not enough to make it worth while.

It's a lot shorter though and i think more readable, so I like it if only for that reason

snopoke:
@dannyroberts not any shorter if I reformat the line...

PR #2289, opened 2018-10-18
Make ES Head work for developers on AWS

dannyroberts:
by exposing all db ports to those logged in via the VPN


esoergel:
Sweet.  I approve of the concept, though I don't think I have enough context to meaningfully review the implementation.  Happy to offline if no one else does either.

src/commcare_cloud/terraform/modules/network/main.tf
```diff
@@ -253,7 +253,12 @@ resource "aws_security_group" "db-private" {
     from_port   = 0
     to_port     = 0
     protocol    = "-1"
-    cidr_blocks = ["${aws_subnet.subnet-app-private.*.cidr_block}"]
+    cidr_blocks = [
+      // Allow traffic from app-private subnets
+      "${aws_subnet.subnet-app-private.*.cidr_block}",
+      // Allow traffic coming from the VPN, i.e. from dev machines, on all ports
+      "${var.openvpn_ip}/32",
```
czue: can we not use only the ports that ES uses?

dannyroberts: We could, but I figured this is a general problem we're going to have with other services. That said, I think the services that this security group is effectively for are:
- es
- kafka
- rabbitmq
and I don't know of any workflows we have where it's useful to access the kafka or rabbit ports from your laptop. (I don't even know what those ports are off the top of my head, which at least confirms that _I_ don't have a workflow around those two.) I think unless either of you disagree, I can go ahead and change this to target ES more specifically. (Easiest thing is just to open 9200 from this IP address on all "db-private" machines including kafka and rabbitmq, which won't have anything there anyway, so I don't see it being an issue. I'll leave a comment to that effect as well.)

czue: works for me

dannyroberts:
Okay that should take care of it!

PR #2292, opened 2018-10-19
give a default value to a setting used in couch role

sravfeyn:
@snopoke


src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -43,6 +43,7 @@ temp_riakcs_proxy_port: 8081
 riakcs_control: "{{ groups.get('riakcs')[0] | default('riakcs-is-disabled') }}"
 zookeeper_client_port: 2181
 formplayer_port: 8181
+couchdb_use_haproxy: False
 couchdb2_port: 15984
 couchdb2_proxy_port: "{{ 35984 if couchdb_use_haproxy else 25984 }}"
```
sravfeyn: Can I get a confirmation that this gets overridden by the var defined in the `public.yml`?

PR #2294, opened 2018-10-19
improving string

jmtroth0:
@sravfeyn buddy: @Rohit25negi 

this actually has the output i wanted

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
+++ b/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
@@ -61,7 +61,7 @@ smtp.to.address=commcarehq-ops+formplayer@dimagi.com
 
 // Redis for locking
 {% if is_redis_cluster %}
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
```diff
@@ -61,7 +61,7 @@ smtp.to.address=commcarehq-ops+formplayer@dimagi.com
 
 // Redis for locking
 {% if is_redis_cluster %}
-redis.clusterString={{ groups.redis_cluster_master | map('extract', hostvars, ['address']) | join(':6379,') }}:6379
+redis.clusterString={{ groups.redis_cluster_master | join(":6379,") }}:6379
```
sravfeyn: Might want to replace `6379` with the var `REDIS_PORT`

sravfeyn: Might want to replace `6379` with the var `REDIS_PORT`

PR #2297, opened 2018-10-22
Set couch data reduce limit to False for Softlayer

Rohit25negi:
http://manage.dimagi.com/default.asp?284632

@snopoke @jmtroth0

### DIFF
--- a/environments/softlayer/public.yml
+++ b/environments/softlayer/public.yml
@@ -49,7 +49,7 @@ couch_dbs:
     password: "{{ localsettings_private.COUCH_PASSWORD }}"
     is_https: False
 
-
+couchdb_reduce_limit: False
 couchdb_cluster_settings:
   q: 8
   r: 1


jmtroth0:
Yeah, I'm not sure I totally understand what's happening here. 

http://docs.couchdb.org/en/1.6.0/couchapp/ddocs.html#redfun

Any chance you could give a bit more detail about why this is necessary (1) for all views/this view in particular and (2) only on softlayer?

czue:
@jmtroth0 

> Any chance you could give a bit more detail about why this is necessary (1) for all views/this view in particular and (2) only on softlayer?

Not really a satisfying answer, but my sense is that it's likely due to the nonstandard reduced referenced in [the readme](https://github.com/dimagi/commcare-hq/blob/master/corehq/couchapps/exports_forms_by_app/README.md), and the rate at which it reduces will be a function of the data in the  DB so it can manifest itself only in certain envs.

Rohit25negi:
thanks @czue  for explaining that. Continuing Cory's answer, as reduce function was not able to fit the large data into its threshold. therefore this code skip the data reduction limit check as it feel better meantime. Perhaps the better solution for long term would be to make a standard reduce function.
fyi @jmtroth0

snopoke:
This came up on ICDS first and I reviewed the reduce function in question at that time but didn't find anything 'wrong' with it: https://github.com/dimagi/commcare-hq/blob/7eab528641daa72357bea08c0e1eb80d6b7a480d/corehq/couchapps/exports_forms_by_app/views/view/reduce.js

My theory as to why this only happens on softlayer and ICDS is because of the large ICDS apps but I wouldn't be surprised if we started to see it on prod at some point with the large l10k apps.

PR #2301, opened 2018-10-23
Increase RAM (and CPU threads) on production AWS webworkers

dannyroberts:
From 4 vCPU (provisioned), 8GB RAM (c5.xlarge @ $0.17/hr)
To 8 vCPU (burstable), 32GB RAM (t3.2xlarge @ $0.3328/hr)


dannyroberts:
AWS is temporarily out of t3.2xlarge instances in availability zone us-east-1a, so closing this for now and replacing with something else.

PR #2305, opened 2018-10-23
Revert "attempt to speed up"

dannyroberts:
Reverts dimagi/commcare-cloud#2288

Turns out this has the side effect of leaving the group with different membership when you use --limit, and in particular the group not existing at all if none of the --limit hosts are members. Since you seemed lukewarm about the change anyway, I'll suggest just reverting.

### DIFF
--- a/src/commcare_cloud/ansible/host_group_aliases.yml
+++ b/src/commcare_cloud/ansible/host_group_aliases.yml
@@ -2,21 +2,25 @@
   hosts: all
   tasks:
     - name: Create all_commcarehq group alias
-      group_by: key=all_commcarehq
-      when: "'openvpn' not in group_names"
+      add_host:
+        hostname: "{{ item }}"
+        groups: all_commcarehq
+      with_items: "{{ groups['all'] }}"
+      when: "item not in groups['openvpn']|default([])"
       changed_when: no
     - name: Create commcarehq group alias
-      group_by: key=commcarehq
-      when: |
-        ('webworkers' in group_names) or
-        ('formplayer' in group_names) or
-        ('celery' in group_names) or
-        ('proxy' in group_names) or
-        ('pillowtop' in group_names) or
-        ('couchdb2_proxy' in group_names) or
-        ('airflow' in group_names) or
-        ('django_manage' in group_names)
+      add_host:
+        hostname: "{{ item }}"
+        groups: commcarehq
+      with_items:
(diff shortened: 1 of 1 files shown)


snopoke:
Fine to revert but curious why that's bad if hosts get removed when using limit? That's the point of limit isn't it?

dannyroberts:
Yeah, we of course want limit to limit the set of hosts tasks get run on, but we don't want it to change group membership. Think for example of something like writing to the shared dir conf. If we run with `--limit <shared_dir_host>`, we still want to be able to access `groups['webworkers']`, etc. and have it give the same value as when we're not limiting machines. Same with `groups['commcarehq']`: we don't want the value of that to change depending on which machines are `--limit`ing the run to.

millerdev:
@dannyroberts it sounds like you're pointing out a general problem with the use of `--limit` in conjunction with use of `groups[...]`. It seems like that issue would apply anywhere where `--limit` may be used and `groups` is referenced in certain ways. Is there a best practice to be gleaned from this? Maybe don't use `groups` in certain ways, ever, because it might not do what you expect if someone runs the play with `--limit`?

dannyroberts:
@millerdev I don't think there's a general problem with `--limit` and `groups`. My understanding and my experience is that `groups` keeps the same values regardless of what's passed into `--limit`, for the groups defined in the inventory. The 'commcarehq' and 'all_commcarehq' groups, which are generated dynamically within the playbook, should have that consistency as well, and the PR which this PR reverts violated that principle (i.e. value of `--limit` was improperly affecting value of `groups['commcarehq']`).

Have you experienced something to the contrary? I.e. have you experienced changing the value of `--limit` to change the value of `groups[...]` in some other situation?

millerdev:
I'm glad to hear it's only limited to this PR, and not a general problem with referencing `groups` and using `--limit`. However, there still may be a best practice emerging here, specifically, don't do the kind of thing that was done in that PR. I'm not sure how to characterize it, but it looks like use of `group_by` may have surprising outcomes when `--limit` is used. Maybe you can phrase it more coherently than that?

dannyroberts:
Yeah I think the take away is that it's better to use `add_host` + `with_items` than it is to use `group_by` + `when` (which then doesn't mention things excluded by --limit). A side note is that both are kind of clunky and slow, and it'd be nice to have another option that put this group more on the level of the ones statically defined in the inventory and didn't require starting every playbook with a callout to generate them.

PR #2306, opened 2018-10-23
Add the pillow staging machine to terraform and inventory.ini

pr33thi:
Once this is merged in, I'll run:

`cchq production terraform init`
`cchq production terraform --skip-secrets apply`
`cchq production aws-fill-inventory`


environments/staging/inventory.ini.j2
```diff
@@ -71,7 +71,7 @@ hqdb1
 # celery0
 
 [hqpillow0]
-hqpillow0-staging.internal-va.commcarehq.org
+# {{ hqpillow0-staging }} hostname=hqpillow0-staging ec2=yes
```
dannyroberts: We're dropping the "hq" prefix on new aws machines. But also this machine is already outline a couple lines down, so no need to change here.

environments/staging/terraform.yml
```diff
@@ -46,6 +46,11 @@ servers:
     network_tier: "db-private"
     az: "a"
     volume_size: 80
+  - server_name: "hqpillow0-staging"
```
dannyroberts: Should be called `pillow0-staging`

environments/staging/inventory.ini.j2
```diff
@@ -71,7 +71,7 @@ hqdb1
...
 
 [pillow0]
 # {{ pillow0-staging }} hostname=pillow0-staging ec2=yes
```
dannyroberts: You'll want to uncomment this and then run

```
cchq staging aws-fill-inventory
```
(after the machine has been created through `cchq staging terraform apply`) to get the real IP copied to `inventory.ini`

dannyroberts: I just read your description more closely, so I now realize this comment wasn't that helpful, since you're planning on doing this as the following step.

pr33thi: Ok cool, and that's what I should be doing. Cool
(Btw, I sent you some slacks, letting you know here because you're snoozed there but seem to be checking this)

pr33thi:
@dannyroberts this is ready for review again

PR #2308, opened 2018-10-24
make unused pillowtop machines into webworkers

snopoke:
I want to see if adding more webworkers fixes the issues with redis etc. like it did on prod.

These pillow machines are currently unused.


millerdev:
Is it expected that these machines will be converted back to pillow machines? It will probably be fine to do so, but we do not test our ansible automation thoroughly on repurposed machines. It's possible that configuration details from previous uses of the machine could carry over, which could result in unexpected operational characteristics. This comment is not necessarily actionable, but I wanted to bring it up as food for thought.

snopoke:
Yes that is expected. Since these are both 'commcare' machines there is very little that is different between them.

PR #2311, opened 2018-10-24
Fix deploy-stack --first time and use in prove-deploy

dannyroberts:


dannyroberts:
prove-deploy build passed 🎉

PR #2313, opened 2018-10-24
Move all production webworkers to aws

dannyroberts:


environments/production/terraform.yml
```diff
@@ -120,6 +120,16 @@ servers:
...
     volume_size: 600
+  - server_name: "web10-production"
+    server_instance_type: t3.xlarge
+    network_tier: "app-private"
+    az: "b"
+    volume_size: 40
+  - server_name: "web11-production"
+    server_instance_type: t3.xlarge
+    network_tier: "app-private"
+    az: "b"
```
pr33thi: When do you do "a" vs "b"?

dannyroberts: 'a' and 'b' refer to different availability zones. Think of them as different buildings in a data center with different power sources and different internet connections: if something bad happens to one availability zone it (hopefully) doesn't affect other availability zones. That's the idea—it lets you design things in such a way that you have redundancy in different AZs and temporary outages in one AZ don't bring your site down.

Now, we're far from this being the case for us—for that to be the case we'd need each service to be redundant across more than one AZ—but it doesn't hurt, and I figure we might as well get started with webworkers, since eventually we'd have to do it with all services to make it work.

In other words, I could have just as well put all these new machines in zone 'a' and practically it wouldn't have been that different, but if we did this with all types of services and we did it right, it could potentially protect us from certain types of provider (aws) outages.

PR #2324, opened 2018-10-26
Send submissions to a separate group of webworkers

dannyroberts:
Okay, this is deployed now. Settled on splitting it 50/50, because even though `receiver/` gets less than a third of requests, it's responsible for more than half of the requests that take longer than 2 minutes. So I figured giving allotting half the webworkers for just submissions seemed like a reasonable place to start.

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -49,10 +49,10 @@ web0
 web1
 web2
 web3
-web4
-web5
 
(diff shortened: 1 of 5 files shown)


dannyroberts:
@snopoke I think that's a good idea, but I just wanted to start this as a test and see what the data said. I agree that it makes sense to move all mobile endpoints to the "mobile_webworkers".

PR #2329, opened 2018-10-30
set redis client timeout to 5 mins

snopoke:
This will make redis kill client connections that are idle for 5 mins. This will prevent the buildup of redis connections we've been seeing (though won't fix the root cause).


snopoke:
Notes:
```
$ cchq $env run-shell-command redis -b "redis-cli config set timeout 300"
$ cchq $env run-module redis lineinfile -b "dest=/etc/redis/6329.conf regexp='^timeout ' line='timeout 300'"

PR #2330, opened 2018-10-30
Set up couchdb2 on AWS in production

dannyroberts:
This is all deployed except the last commit, to use the local db for users and domains databases. I'll test that before deploying it of course, but @snopoke I just wanted to ask you to double check the format for a gut check. I've already replicated the domains and users databases from cloudant to the local one.


environments/production/public.yml
```diff
@@ -40,7 +40,7 @@ formplayer_purge_time_spec: '8d'
 KSPLICE_ACTIVE: yes
 
 
-ufw_private_interface: eth1
+ufw_private_interface: "{{ ec2|default(false) and 'eth0' or 'eth1' }}"
```
millerdev: nit: `'eth0' if ec2|default(false) else 'eth1'` is the standard convention for inline conditionals

PR #2331, opened 2018-10-30
Make some tasks never report "changed"

dannyroberts:
because they just run a shell commmand to read something

Just a few annoying little things I found while doing other stuff.


pr33thi:
I noticed this too, good call

PR #2332, opened 2018-10-30
Production rabbitmq

dannyroberts:
This PR adds machines that are set up but not hooked up to anything yet. The point is to deploy to them to make sure they're up to date, and then do the cutover as a separate step.


environments/production/inventory.ini
```diff
@@ -165,12 +169,34 @@ hqcelery3.internal-va.commcarehq.org
 [hqcelery4]
 hqcelery4.internal-va.commcarehq.org
 
+
+[celery0]
+10.202.12.202 hostname=celery0-production ec2=yes
+
+[celery1]
+10.202.12.214 hostname=celery1-production ec2=yes #datadog_integrations=celery
```
millerdev: can we keep `datadog_integrations=celery` turned on here?

dannyroberts: I was planning on turning it on once we shut off the other ones. Rackspace hqcelery1 still has this integration, and I didn't know what would happen if we turned it on for both

millerdev: Oh, I was thinking this was `datadog_integration_tcp_check`, which is different. Never mind.

PR #2334, opened 2018-10-30
Add an es node to staging

pr33thi:
Running `aws-fill-inventory` also updated the ip address of the celery and pillow machine, any ideas why that might have happened?

@dannyroberts 
code buddy @sravfeyn


pr33thi:
Oh and to answer the question in the pr description, it's because I added the new machine to the middle of the list of machines in terraform.yml and not the end, and that made terraform shift everything below the new machine by one. Next time I'll read the diff more closely  👍

PR #2336, opened 2018-10-31
Migrate production celery and rabbitmq

dannyroberts:
This PR follows up on #2332.

This has all been deployed: all new tasks are going to the new RabbitMQ (rabbit0) and picked up by the new celery workers (celery[0-4]). The old ones are still up to go through the backlog, and I'm leaving up a bit in case some process continued sending there before it died. I'll check and remove the old rabbit & celery machines tomorrow.

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -136,8 +136,8 @@ hqtouch0.internal-va.commcarehq.org
 10.202.42.98 hostname=rabbit0-production ec2=yes
 
 [rabbitmq:children]
-hqtouch0
 rabbit0
+hqtouch0
 
 [rabbitmq:vars]
 swap_size=2G
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -158,7 +158,7 @@ kafka_broker_id=1
 hqcelery0.internal-va.commcarehq.org
 
 [hqcelery1]
-hqcelery1.internal-va.commcarehq.org datadog_integrations=celery
+hqcelery1.internal-va.commcarehq.org
 
 [hqcelery2]
 hqcelery2.internal-va.commcarehq.org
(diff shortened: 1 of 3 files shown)


dannyroberts:
> I assume you updated localsettings so that all new tasks go to the new rabbit (I see you changed the order in the rabbitmq group think you must have done so).

Yup, I did. Updated localsettings for all machines except the old celery workers—their rabbitmq is the old one. That way they continue processing any remaining things that trickle into that queue.

environments/production/inventory.ini
```diff
@@ -276,13 +276,13 @@ es_data
...
-10.202.10.19 hostname=couch0-production ec2=yes
+10.202.40.211 hostname=couch0-production ec2=yes
 
 [couch1]
-10.202.11.235 hostname=couch1-production ec2=yes
+10.202.41.83 hostname=couch1-production ec2=yes
 
 [couch2]
-10.202.12.181 hostname=couch2-production ec2=yes
+10.202.42.85 hostname=couch2-production ec2=yes
```
millerdev: Were these intended to be part of this PR?

environments/production/inventory.ini.j2
```diff
@@ -174,7 +174,7 @@ hqcelery4.internal-va.commcarehq.org
 {{ celery0-production }} hostname=celery0-production ec2=yes
 
 [celery1]
-{{ celery1-production }} hostname=celery1-production ec2=yes #datadog_integrations=celery
+{{ celery1-production }} hostname=celery1-production ec2=yes
```
millerdev: Did you intend to enable this rather than deleting it?

environments/production/inventory.ini
dannyroberts: Ooops, I didn't, but it's a change that's also on master from #2337. This happened when I ran `aws-fill-inventory`, and did reflect the correct IPs. (So while it's unrelated to the other changes and didn't have a place in this PR or that commit per se, and was a mistake, there's nothing I need to undo.)

environments/production/inventory.ini.j2
dannyroberts: Embedded the reason in the commit message:
> It turns own that datadog_integrations=celery doesn't do anything,
and that it always picks the first celery machine.

So after trial and error I realized having this variable set here was useless and that I needed to move celery1 to the top of the list instead.

PR #2347, opened 2018-11-01
Redis cluster in ICDS

sravfeyn:
Only merge after confirmation of successful switch.


sravfeyn:
This is now deployed, so it's ready to be merged. cc @pr33thi

PR #2351, opened 2018-11-01
Remove old celery and rabbit from prod inventory

dannyroberts:
We've been on the new setup for a day+ now, and the old queues have been stopped for a while.

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -25,7 +25,6 @@ celery_processes:
     async_restore_queue:
       concurrency: 2
       max_tasks_per_child: 5
-  hqcelery0.internal-va.commcarehq.org: {}
   celery1:
     flower: {}
     reminder_queue:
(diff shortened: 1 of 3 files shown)


millerdev:
I'm not sure if we keep logs around when machines are removed (I think we do not). Celery logs seem especially useful for troubleshooting things like SMS failures. Would it be possible to grab a copy of `/home/cchq/www/production/logs` before these machines are deleted for good?

Longer term: would be nice to have a log aggregation service (datadog?) so we wouldn't have to worry about this when machines are decommissioned.

dannyroberts:
Okay, they now live in `/home/cchq/www/production/log/archive/log` of the similarly named aws machine. (I.e. hqcelery0 -> celery0, etc.)

dannyroberts:
@pr33thi good question. I don't think so, because these are one-off locations where I copied some logs just this one time on production, and not part of a system where people can expect to find something in other situations.

PR #2352, opened 2018-11-02
Mk minimize softlayer

mkangia:
[Source ticket](https://manage.dimagi.com/default.asp?285065)

Celery machine on softlayer has no memory to even do a location upload,
so this is to reduce the memory usage till we move it to AWS

My understanding is that activity on softlayer is minimal hence combining the queues and reducing the concurrency should be the right to do anyway

best review commit by commit

### DIFF
--- a/environments/softlayer/app-processes.yml
+++ b/environments/softlayer/app-processes.yml
@@ -3,10 +3,17 @@ newrelic_javaagent: True
 celery_processes:
   'celery0':
     beat: {}
-    celery_periodic:
-      concurrency: 2
+    flower: {}
+    reminder_queue:
+      pooling: gevent
+      concurrency: 5
     submission_reprocessing_queue:
       concurrency: 1
+    sms_queue:
+      pooling: gevent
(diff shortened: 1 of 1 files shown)


snopoke:
I suppose this is ok though we can also just increase the size of the VM.

mkangia:
Nitigya mentioned there is a plan to move softlayer to AWS so i guess we can increase the VM configuration then. 
Plus, since there isn't much activity it should be fine to club queues together, though only after this change we can see how much it impacts.
pinged @dannyroberts as well

mkangia:
Deployed to softlayer,
a minor drop in RAM usage observed
https://app.datadoghq.com/dash/141098/celery-overview?live=true&page=0&is_auto=false&from_ts=1541442576693&to_ts=1541446176693&tile_size=m&tpl_var_environment=softlayer&fullscreen=181277394

tried a task and that finished this time which failed earlier due to memory constraints. So the machine has a little breathing space now but might not suffice if case any heavy task is initiated.

PR #2354, opened 2018-11-02
add old redis instance as ws4redis

sravfeyn:


sravfeyn:
This is deployed on staging already

PR #2355, opened 2018-11-02
 Add redis0 as ws4 redis

sravfeyn:


sravfeyn:
This is deployed on ICDS

PR #2356, opened 2018-11-02
standbys for main

NitigyaS:


snopoke:
Should also modify `postgresql.yml` `LOAD_BALANCED_APPS`

PR #2357, opened 2018-11-02
Update inventory files

pr33thi:
- Changes were made directly to inventory.ini here, added those changes to inventory.ini.j2
- Running aws-fill-inventory to capture the fact that the airflow hosts should be uncommented

@dannyroberts 
code buddy @sravfeyn

### DIFF
--- a/environments/staging/inventory.ini.j2
+++ b/environments/staging/inventory.ini.j2
@@ -100,10 +100,12 @@ hqredis2-staging.internal-va.commcarehq.org
 hqredis0
 hqredis1
 hqredis2
-# Redis old single node
-# hqdb1
+hqdb1
 # Amazon ElastiCache
 # {{ redis-staging }}:6379
+
+[ws4redis:children]
+hqdb1
 
 [redis_cluster_master:children]
 hqredis0
--- a/environments/staging/inventory.ini.j2
+++ b/environments/staging/inventory.ini.j2
@@ -145,9 +145,6 @@ hqriak01-staging.internal-va.commcarehq.org
 hqriak00
 hqriak01
 
-[hqairflow0]
-hqairflow0-staging.internal-va.commcarehq.org
-
 [airflow0]
 {{ airflow0-staging }} hostname=airflow0-staging ec2=yes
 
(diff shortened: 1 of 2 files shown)


environments/staging/inventory.ini.j2
```diff
@@ -100,11 +100,13 @@ hqredis2-staging.internal-va.commcarehq.org
...
 hqredis1
 hqredis2
-# Redis old single node
-# hqdb1
+hqdb1
 # Amazon ElastiCache
 # {{ redis-staging }}:6379
 
+[ws4redis:children]
+hqdb1
```
dannyroberts: Ah so, was this the cause of the problem we saw on staging yesterday?

pr33thi: Yeah, this was the actual fix: https://github.com/dimagi/commcare-cloud/pull/2354
But it didn't update the j2 files so I did it here

pr33thi: np! Yeah, this applies to every environment. The template has variable values for the ip addresses (ie:
```
[airflow0]
{{ airflow0-staging }} hostname=airflow0-staging ec2=yes
```
)
`aws-fill-inventory` fills in those variables based on the ip addresses in aws, so it's easier to keep the inventory files up to date (as opposed to having to manually paste in the ip addresses or trust that whatever is check is up to date)

sravfeyn: I have just tried to do this and found that it needs a terraform config. Where can I get the config from? (Also assuming this process is for AWS based envs)

pr33thi: Hm, I was not able to replicate that but @dannyroberts just made changes to this process and can probably help more.

dannyroberts: Yeah the `inventory.ini.j2` / `aws-fill-inventory` process is only for AWS based envs.

It's now possible to do this without talking to AWS by running `cchq <env> aws-fill-inventory --cached`, which runs off the `aws-resources.yml` file that lives in the environment (alongside the inventory and the other files). If you're not adding a new aws machine (but just editing the template to change group membership, add non-aws machines, remove aws machines, edit host variables, etc.) then this is the easiest way to do it.

I still owe you and the rest of the team an email outlining these new processes, so I'll be sending that out shortly. That'll include what to do once you _do_ need the ability to create a new machine on AWS, etc.

Thanks for your patience with all the processes in flux right now. They've slowly been coming together, and I hope it'll be a bit more consistent going forward.

PR #2358, opened 2018-11-02
Rate-limit sauti-1

dannyroberts:
This PR should only affect prod, but includes a logically related refactor to move `site_locations` to `proxy.yml` where its schema can be enforced.

Rates arbitrarily proposed here are 1r/s for restores and 10r/s for submissions, but I think that's probably very conservative. (It's 1/6 of the rate-limiting for ICDS.)

Here's the diff that'd go to the nginx conf:

```diff
+map $http_authorization $digest_submission {
+  ""     $server_name;
+  default     "";
+}

+limit_req_zone $digest_submission zone=restore:10m rate=1r/s;
+limit_req_zone $digest_submission zone=submissions:10m rate=10r/s;

 proxy_cache_path /home/cchq/www/production/hq_media/cache levels=1:2 keys_zone=hq_media_cache:10m
                  max_size=2g inactive=24h use_temp_path=off;
@@ -113,6 +120,22 @@
             proxy_next_upstream_tries 1;
             proxy_redirect http://www.commcarehq.org https://www.commcarehq.org;
               }
+  location /a/sauti-1/phone/restore {
+            proxy_pass http://production_commcare;
+                    client_body_buffer_size 512k;
+            proxy_buffers 8 64k;
+            proxy_buffer_size 64k;
+            proxy_next_upstream_tries 1;
+                    limit_req zone=restore burst=100 nodelay;
+      }
+  location /a/sauti-1/receiver {
+            proxy_pass http://production_commcare_submissions;
+                    client_body_buffer_size 512k;
+            proxy_buffers 8 64k;
+            proxy_buffer_size 64k;
+            proxy_next_upstream_tries 1;
+                    limit_req zone=submissions burst=100 nodelay;
+      }
```

### DIFF
--- a/environments/production/proxy.yml
+++ b/environments/production/proxy.yml
@@ -23,3 +23,35 @@ special_sites:
   - wiki_http
   - motech
   - motech2
+
+limit_req_zones:
+  - name: "restore"
+    zone: "$digest_submission"
+    size: "10m"
+    rate: "1r/s"
+  - name: "submissions"
+    zone: "$digest_submission"
+    size: "10m"
+    rate: "10r/s"
+
+site_locations:
+  - name: "/a/sauti-1/phone/restore"
+    balancer: "{{ deploy_env }}_commcare"
+    proxy_next_upstream_tries: 1
+    proxy_buffers: 8 64k
+    proxy_buffer_size: 64k
+    client_body_buffer_size: 512k
+    limit_reqs:
+      - zone_name: "restore"
+        burst: 100
+        nodelay: True
+  - name: "/a/sauti-1/receiver"
+    balancer: "{{ deploy_env }}_commcare_submissions"
+    proxy_next_upstream_tries: 1
+    proxy_buffers: 8 64k
+    proxy_buffer_size: 64k
+    client_body_buffer_size: 512k
+    limit_reqs:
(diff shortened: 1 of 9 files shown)


environments/production/proxy.yml
```diff
@@ -23,3 +23,35 @@ special_sites:
...
+
+limit_req_zones:
+  - name: "restore"
+    zone: "$digest_submission"
+    size: "10m"
+    rate: "1r/s"
+  - name: "submissions"
+    zone: "$digest_submission"
+    size: "10m"
+    rate: "10r/s"
```
snopoke: this seems high - doesn't look like we go much over 10r/s globally on prod

environments/production/proxy.yml
```diff
@@ -23,3 +23,35 @@ special_sites:
...
+
+site_locations:
+  - name: "/a/sauti-1/phone/restore"
```
snopoke: any reason not to use a regex and match all domains?

dannyroberts: This PR is for intentionally rate-limiting just the one domain. (There's still a separate thing I want to do which is to send restores to the new split-out "mobile_webworkers".)

PR #2360, opened 2018-11-02
Add pillow0-production

dannyroberts:
to replace 3 machines with 1 bigger machine,
4x the size, for a total increase of 33% on both RAM and CPU


environments/production/app-processes.yml
```diff
@@ -89,6 +89,7 @@ celery_processes:
       concurrency: 6
       max_tasks_per_child: 5
 pillows:
+  pillow0: {}
```
pr33thi: Isn't the rackspace one one supposed to be `{}`'d?

mkangia:
is it okay to put all on one? 
I am not sure how this would work if we put all pillows on single machine and they start conflicting in terms of resources. Just like we try to put a single service on a machine, it would make sense to consider something on similar lines for pillows.

Someone with a better understanding of which pillows demand, which kind of resource RAM/Network/IO might be able to weigh on this better.

mkangia:
Also,
thanks for getting to this. We need this :)

environments/production/app-processes.yml
millerdev: I think that will be done in a separate step. Looks like the machine is just being added here, but no work has been assigned to it yet.

dannyroberts:
@mkangia I think that this change shouldn't meaningfully affect resource contention. Importantly, this one new machines has more resources dedicated to it (RAM & CPU) than the current three combined. It's in principle possible for one process to use up all the resources and cause problems for the other ones, but when I looked at the resource usage in datadog for these machines, it didn't seem like that would be a problem. On the plus side, it allows the OS to figure out how to distribute its resources among all the processes as we get different bursts of activities in the different change feeds. But I think overall, it won't be much of a difference, and all things being equal fewer machines is slightly easier to manage than more machines.

environments/production/app-processes.yml
dannyroberts: Yeah that's right. I want to merge this so it gets fab deploy'd this afternoon, and then afterwards I can move the pillows over to this new machine in a separate changeset

mkangia:
okay @dannyroberts , thanks.

snopoke:
@mkangia I don't see this being a problem provided the virtualization layer handles the CPU allocation correctly.  We should make sure that the file limit is set high enough to cater for all the outgoing network connection to the databases etc.

mkangia:
Cool. thanks @snopoke

dannyroberts:
@snopoke is this the "file limit" you're referring to? https://github.com/dimagi/commcare-cloud/blob/d8dc18a66f7c5f1b5e694ed30e5ecd9da499b7e7/environments/production/public.yml#L24. (If not, which var is it?) If so, that var is currently set to the same thing on prod as it is on ICDS for comparison. Do you have an idea for whether that's a high-enough value?

snopoke:
That is what I was referring to. That should definitely be high enough.

PR #2361, opened 2018-11-05
break down saved exports

mkangia:
Saved exports is often stuck and needs a kick. Trying to avoid that by breaking it down into multiple workers instead, to avoid high concurrency by a single worker but keeping the overall number of concurrent tasks same.
This also helps avoiding complete stop on processing in case the queue is stuck.

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -75,7 +75,8 @@ celery_processes:
       concurrency: 8
       max_tasks_per_child: 5
     saved_exports_queue:
-      concurrency: 6
+      num_workers: 3
+      concurrency: 2
       max_tasks_per_child: 1
       optimize: True
     export_download_queue:


environments/production/app-processes.yml
```diff
@@ -75,7 +75,8 @@ celery_processes:
       concurrency: 8
       max_tasks_per_child: 5
     saved_exports_queue:
-      concurrency: 6
+      num_workers: 3
+      concurrency: 2
```
mkangia: i am not super sure that this is how we should be adding this to have multiple workers on the same machine, so requesting reviewer to confirm that.

emord: Seems fine to me. It looks like the docs just say to test it out, they don't really give any hints on how to figure out the difference in performance of workers vs concurrency http://docs.celeryproject.org/en/3.1/userguide/workers.html#concurrency

mkangia: Thanks @emord . Is this g2g?

dannyroberts: Does num_workers vs. concurrency really mean much if it's not using gevent? I thought it didn't but I don't have a deep understanding

mkangia: one more thing i noticed that i we have concurrency of say, 6. No new tasks are taken until all 6 are completed, so if 5 tasks are done and 1 is still pending, the worker does not take any more tasks which is not a good behaviour.

mkangia:
still holding off on deploy for this, waiting on approved review

environments/production/app-processes.yml
emord: @dannyroberts from the docs i linked to:

> There is even some evidence to support that having multiple worker instances running, may perform better than having a single worker

I'm not sure why that is. it does seem confusing

dannyroberts:
Heads up that pillows are now on just the one machine, but I think this diff will apply cleanly and correctly even though the pillow it's affecting is now under a different machine

mkangia:
deployed

PR #2364, opened 2018-11-05
Switch to using local couchdb2 for production auditcare and fixutres dbs

dannyroberts:
This PR is continuing to follow the same pattern as https://github.com/dimagi/commcare-cloud/pull/2340 and https://github.com/dimagi/commcare-cloud/pull/2350


pr33thi:
Which machine was being used for this before?

dannyroberts:
@pr33thi these dbs were previously hosted on cloudant, where the rest of the dbs not mentioned in this PR (and the other two) still are. This is part of an overall process of migrating from cloudant to internally hosted couchdb2

PR #2368, opened 2018-11-06
New pillow servers

emord:
@dimagi/scale-team 

Haven't pushed out the partition additions yet. @snopoke can you alert me when you remove the pillow servers from the proxy? It wasn't clear to me whether you were planning to do that tomorrow or the next day (to decouple them from the rate limit changes)


snopoke:
I removed the pillows from the proxy yesterday.

PR #2372, opened 2018-11-07
Send restores to "mobile_webworkers"

dannyroberts:
Previously only submissions go to the `[mobile_webworkers]`, now restores (and the other more minor endpoints under /phone/) go there too

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -37,7 +37,7 @@ nginx_sites:
       proxy_buffers: 8 64k
       proxy_buffer_size: 64k
       client_body_buffer_size: 512k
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -37,7 +37,7 @@ nginx_sites:
       proxy_buffers: 8 64k
       proxy_buffer_size: 64k
       client_body_buffer_size: 512k
-    - name: '~ /a/[^/]+/receiver'
+    - name: '~ /a/[^/]+/(receiver|phone)'
```
millerdev: will this match URLs like these?

- `/a/abc/receiverightnow` - for urgent deliveries
- `/a/abc/phonews` - where we publish news about Vietnamese soup

Obviously made up URLs, and probably not a big concern, but might be safer to include a trailing slash

`- name: '~ /a/[^/]+/(receiver|phone)/'`

PR #2373, opened 2018-11-07
Make public subnet accessible from other machines on the network

dannyroberts:
on all ports

This allows webworkers to make requests to the riak proxy.


dannyroberts:
This change means machines on the internal network ("VPC") will be able to talk to the proxy machines on any port. (The outside world still will still be blocked from all ports except 80 and 443.)

PR #2375, opened 2018-11-07
Template terraform servers so each is addressed by name

dannyroberts:
This will make it a lot more intuitive to add and remove servers, or to reorganize them without affecting anything, in `terraform.yml`.


src/commcare_cloud/commands/terraform/migrations/0002_name_servers.py
```diff
@@ -0,0 +1,17 @@
...
+        server_config = environment.terraform_config.servers[index]
+        return 'module.server__{}.aws_instance.server'.format(server_config.server_name)
+
+    if old_resource_address == 'module.proxy_servers.aws_instance.server':
+        server_config = environment.terraform_config.proxy_servers[0]
+        return 'module.proxy_server__{}.aws_instance.server'.format(server_config.server_name)
+
+    if old_resource_address == 'aws_eip.proxy':
+        server_config = environment.terraform_config.proxy_servers[0]
+        return 'aws_eip.{}-production'.format(server_config.server_name)
```
millerdev: Is this function supposed to return `None` if none of the conditions match? We often do not explicitly `return None` (since Python does that implicitly), but I think it's a good practice to make the intention clear.

As a rule, a function should always explicitly return some value if any return statements with a value exist in the function. This makes it easier for a code reviewer to detect if a mistake has been made.

dannyroberts: Good suggestion. Decided to roll it up into this same PR in the following two new commits. Let me know what you think

PR #2388, opened 2018-11-09
Make templated inventory verifiable

dannyroberts:
Previously to generate the inventory file for an using aws (like production and staging), you'd edit `inventory.ini.j2` and run `cchq <env> aws-fill-inventory`, which would (1) pull server info from aws and (2) plug it into the template to generate `inventory.ini`. This PR splits that into separate steps, saving the server information to a file that's committed to the repo, and then rendering the template from that file. This has two purposes:

- It's faster if you're editing the template to just rerender it rather than pull the server info every time.
- Maybe more importantly, it'll cause the build to fail if you by mistake edit `inventory.ini` directly instead of `inventory.ini.j2`, and then tell you what to do to fix it. This'll keep us from a very foreseeable kind of human error in this process.
- Finally, and also importantly, it'll allow people to make changes to the inventory and follow the process without having to have AWS credentials—which is totally fine if you're not adding or removing machines

(I'll also be sending an email out to the team shortly with updates letting people know about this and other little aws-related processes like this.)


environments/staging-aws-test/inventory.ini
```diff
@@ -86,8 +86,15 @@ remote_postgresql
 
 [redis]
 # ElastiCache
-staging-test-redis.ck3xwy.0001.use1.cache.amazonaws.com:6379
+{{ redis-staging-test }}:6379
```
dannyroberts: This is technically wrong, but also in a fake env, so it's fine to ignore this. If I need to fix it, I'll fix it later, but we'll likely destroy this env before needing to use it again.

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -94,11 +95,37 @@ def run(self, args, unknown_args):
...
+            This runs much more quickly and gives the same result, provided no changes
+            have been made to our actual resources in AWS.
+        """)
+    ]
 
     def run(self, args, unknown_args):
         environment = get_environment(args.env_name)
-        resources = get_aws_resources(environment)
+        if not os.path.exists(environment.paths.inventory_ini_j2):
+            print("Env {} not using templated inventory (inventory.ini.j2). Skipping"
```
pr33thi: If there isn't an inventory j2 file, it seems like it would need to be added back for this to work, should the print statement be updated to reflect that?

tests/test_autogen_environments.sh
```diff
@@ -0,0 +1,24 @@
...
+
+wait
+
+git diff
+git update-index -q --refresh
+if git diff-index --quiet HEAD --; then
+    # No changes
+    exit 0
+else
+    # Changes
```
pr33thi: Instead of `Changes` and `No changes` can you add a little more detail about what's changed? Might not be clear to someone who isn't familiar with `aws-fill-inventory` yet

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -94,11 +95,37 @@ def run(self, args, unknown_args):
...
+
+        if not args.cached:
+            resources = get_aws_resources(environment)
+            with open(environment.paths.aws_resources_yml, 'w') as f:
+                f.write(yaml.safe_dump(resources, default_flow_style=False))
+        else:
+            with open(environment.paths.aws_resources_yml, 'r') as f:
+                resources = yaml.load(f.read())
+
         with open(environment.paths.inventory_ini_j2) as f:
```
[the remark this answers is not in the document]
dannyroberts: It's the opposite: if you have `--cached` than it skips asking AWS for IP addresses and just reads them from `aws-resources.yml`. If you don't use `--cached`, it first asks AWS for IP addresses and dumps them into `aws-resources.yml` (which you should then commit), before going on to rendering the template.

Both render the template, but --cached skips the step of reading IP addresses from aws. The "cache" being referred to here (in `--cached`) is `aws-resources.yml`.

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: Most of our environments don't use templated inventory files, only the ones using AWS, which is staging and production. I added this here to make it easier to add the travis test which runs `commcare-cloud ${env} aws-fill-inventory --cached` for each environment; I don't expect people to try to run it for other environments, but if they do, I think this message has enough information to orient them, and if it comes up often in a situation where more information could be helpful, happy to add more then.

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -94,11 +95,37 @@ def run(self, args, unknown_args):
...
+                  .format(args.env_name))
+            return 0
+
+        if not args.cached:
+            resources = get_aws_resources(environment)
+            with open(environment.paths.aws_resources_yml, 'w') as f:
+                f.write(yaml.safe_dump(resources, default_flow_style=False))
+        else:
+            with open(environment.paths.aws_resources_yml, 'r') as f:
+                resources = yaml.load(f.read())
```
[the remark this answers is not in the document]
dannyroberts: I think if you find the little 
screen shot 2018-11-12 at 12 34 36 pm icon, you can see this (or maybe you did but had questions about what you saw).

Basically `resources` gets used as the context for rendering `inventory.ini.j2`. It's a mapping from names to ip addresses (i.e. the contents of `aws-resources.yml`; you can scroll up to see one).

tests/test_autogen_environments.sh
dannyroberts: Offlined (pulling out "fail if there's a git diff" workflow came out of this)

dannyroberts:
got verbal approval on the last few commits

src/commcare_cloud/commands/terraform/aws.py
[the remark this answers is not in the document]
pr33thi: Thanks!

src/commcare_cloud/commands/terraform/aws.py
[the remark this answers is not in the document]
pr33thi: Thanks!

PR #2396, opened 2018-11-13
persistent django db connections

snopoke:
See commits for details.

@dimagi/scale-team FYI

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -35,6 +35,7 @@ DATABASES = { {% for config in postgresql_dbs.all %}
         'HOST': '{% if config.host == inventory_hostname %}127.0.0.1{% else %}{{ config.host }}{% endif %}',
         'PORT': {{ config.port }},
         'OPTIONS': {{ config.options|to_json }},
+        'DISABLE_SERVER_SIDE_CURSORS': True,
         {% if not config.django_migrate -%}
         'MIGRATE': False,
         {% endif -%}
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -35,6 +35,8 @@ DATABASES = { {% for config in postgresql_dbs.all %}
         'HOST': '{% if config.host == inventory_hostname %}127.0.0.1{% else %}{{ config.host }}{% endif %}',
         'PORT': {{ config.port }},
         'OPTIONS': {{ config.options|to_json }},
+        'DISABLE_SERVER_SIDE_CURSORS': True,
+        'CONN_MAX_AGE': 500,
```
calellowitz: what unit is this?

snopoke: seconds

calellowitz: does that mean a db connection cannot exceed 500 seconds? I would be curious if some of our big queries, aggregation script or maybe even some big ucr dumps, are already well above that

millerdev: oof. we should fix those queries.

snopoke: no, it just means that when Django needs a connection for a new query it will check how long the connection has been open for and if longer than 500s it will close the connection and open a new one.

calellowitz: > oof. we should fix those queries.

They are the nightly aggregation queries, not any that happen when you view the dashboard. 10-15 minutes is probably fine for those

> no, it just means that when Django needs a connection for a new query it will check how long the connection has been open for and if longer than 500s it will close the connection and open a new one.

ah, cool

PR #2398, opened 2018-11-13
stop running ws4redis websockets

sravfeyn:
@snopoke It looks like it if ws4redis stops running the failure mode is that the features stop working without any 500s. The earlier failure mode of 500 error was when supposedly ws4redis connection points to a cluster. I have tested this out on staging.


sravfeyn:
Ha, it looks like `service_blacklist` is not getting used, I assumed it was used while setting up supervisor conf files. I will explore some other way or use that var to do that

sravfeyn:
Updated

sravfeyn:
I have mgirated redis on staging to single node and updated this code as well. This is now fully deployed on staging.

PR #2399, opened 2018-11-13
Upgrade PyGithub to 1.43.3

pr33thi:
Was getting deploy errors on prod because of merged with long PR names that were truncated and ended with `...`
Unsure if this fixes it, but we were on a prerelease before so seemed worth it to upgrade.
@emord 
code buddy @esoergel

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,6 +21,7 @@ couchdb-cluster-admin==0.4.2
 cryptography==2.3.1
 datadog==0.2.0
 decorator==4.3.0          # via datadog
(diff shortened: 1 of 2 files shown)


pr33thi:
@emord Unsure if this fixes it, and the changelog didn't have anything implying that this would, I've updated the PR description to reflect that

esoergel:
Offline with @pr33thi but for posterity, looks like this is an outstanding issue with logging in PyGithub:
https://github.com/PyGithub/PyGithub/issues/911
That link proposes skipping logging, which would circumvent the issue.

PR #2403, opened 2018-11-13
Add kafka0-staging

dannyroberts:
I'm going to migrate kafka/zookeeper on staging before doing it on production


environments/staging/terraform.yml
```diff
@@ -91,6 +91,13 @@ servers:
     az: "a"
     volume_size: 80
 
+  - server_name: "kafka0-staging"
+    server_instance_type: t3.medium
+    network_tier: "db-private"
```
pr33thi: Should this be app-private?

dannyroberts: Nope, app-private is for the app layer, things that run the code that we write for commcarehq; kafka is in the db layer, it's a third-party service for general use that we rely on, analogous to the way we use redis, postgres, elasticsearch, couchdb, for different purposes.

mkangia: Is there a documentation to understand what these fields and their values mean?

pr33thi: I have an explanation on this page: https://confluence.dimagi.com/display/internal/How+to+migrate+a+rackspace+machine+to+AWS
Explained in step 1 of 'Create and Provision the New AWS Machine'

mkangia: Thank you @pr33thi

PR #2406, opened 2018-11-14
AWS enforce Multi Factor Auth

dannyroberts:
While writing documentation on our AWS processes, I didn't feel great about the AWS default of storing access keys in plaintext at ~/.aws/credentials indefinitely, and I had a backlog item to force two-factor auth for all our IAM users.

Once I got started looking into it, a solution came together that satisfies both. This PR has the following changes:
- Forces 2FA for all our IAM users, including for CLI use
- Supports entering 2FA token via command line and storing the resulting session credentials
- They're still stored in ~/.aws/credentials, but now the permanent ones are pretty weak by themselves, and the strong ones are short-lived (30 minutes by default in this PR)

While a theoretical malicious program will still be able to find your strong auth, this change at least limits the period of time during which they can use it.

### DIFF
--- a/src/commcare_cloud/commands/terraform/aws.py
+++ b/src/commcare_cloud/commands/terraform/aws.py
@@ -1,3 +1,4 @@
+# coding=utf-8
 from __future__ import print_function
 
 import getpass
@@ -5,9 +6,12 @@ import json
 import os
 import subprocess
 import textwrap
+from datetime import datetime
 
 import boto3
 import yaml
+from clint.textui import puts, colored
+from memoized import memoized
 from six.moves import shlex_quote
 from six.moves import configparser
 from six.moves import input
@@ -30,8 +34,9 @@ def check_output(cmd_parts, env):
 
 
 def aws_cli(environment, cmd_parts):
+
     return json.loads(
-        check_output(cmd_parts, env={'AWS_PROFILE': environment.terraform_config.aws_session_profile}))
+        check_output(cmd_parts, env={'AWS_PROFILE': aws_sign_in(environment.terraform_config.aws_profile)}))
 
 
 def get_aws_resources(environment):
@@ -156,19 +161,22 @@ class AwsFillInventory(CommandBase):
             f.write(out_string)
 
 
+DEFAULT_SIGN_IN_DURATION_MINUTES = 30
+
+
 class AwsSignIn(CommandBase):
     command = 'aws-sign-in'
     help = """
-        Use your MFA device to "sign in" to AWS for <duration> minutes (default 30)
+        Use your MFA device to "sign in" to AWS for <duration> minutes (default {})
 
         This will store the temporary session credentials in ~/.aws/credentials
         under a profile named with the pattern "<aws_profile>:profile".
         After this you can use other AWS-related commands for up to <duration> minutes
         before having to sign in again.
-    """
+    """.format(DEFAULT_SIGN_IN_DURATION_MINUTES)
 
     arguments = [
-        Argument('--duration-minutes', type=int, default=30, help="""
+        Argument('--duration-minutes', type=int, default=DEFAULT_SIGN_IN_DURATION_MINUTES, help="""
             Stay signed in for this many minutes
         """)
     ]
@@ -177,15 +185,40 @@ class AwsSignIn(CommandBase):
         environment = get_environment(args.env_name)
         duration_minutes = args.duration_minutes
         aws_profile = environment.terraform_config.aws_profile
-        default_username = getpass.getuser()
-        username = input("Enter username associated with credentials [{}]: ".format(default_username)) or default_username
-        mfa_token = input("Enter your MFA token: ")
-        generate_session_profile(aws_profile, username, mfa_token, duration_minutes)
-        print("You're sucessfully signed in!")
-        print("You will be able to use AWS from the command line for the next {} minutes."
-              .format(duration_minutes))
-        print("To use this session outside of commcare-cloud,"
-              "prefix your command with AWS_PROFILE={}:session".format(aws_profile))
+        aws_sign_in(aws_profile, duration_minutes, force_new=True)
+
+
+@memoized
+def aws_sign_in(aws_profile, duration_minutes=DEFAULT_SIGN_IN_DURATION_MINUTES,
+                force_new=False):
+    """
+    Create a temp session through MFA for a given aws profile
+
+    :param aws_profile: The name of an existing aws profile to create a temp session for
+    :param duration_minutes: How long to set the session expiration if a new one is created
+    :param force_new: If set to True, creates new credentials even if valid ones are found
+    :return: The name of temp session profile.
+             (Always the passed in profile followed by ':session')
+    """
+    aws_session_profile = '{}:session'.format(aws_profile)
+    if not force_new \
+            and _has_valid_session_credentials(aws_session_profile):
+        return aws_session_profile
+
+    default_username = getpass.getuser()
+    username = input("Enter username associated with credentials [{}]: ".format(
+        default_username)) or default_username
+    mfa_token = input("Enter your MFA token: ")
+    generate_session_profile(aws_profile, username, mfa_token, duration_minutes)
+
+    puts(colored.green(u"✓ Sign in accepted"))
+    puts(colored.cyan(
+        "You will be able to use AWS from the command line for the next {} minutes."
+        .format(duration_minutes)))
+    puts(colored.cyan(
+        "To use this session outside of commcare-cloud, "
+        "prefix your command with AWS_PROFILE={}:session".format(aws_profile)))
+    return aws_session_profile
 
 
 def generate_session_profile(aws_profile, username, mfa_token, duration_minutes):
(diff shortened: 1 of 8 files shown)


src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -150,3 +159,116 @@ def run(self, args, unknown_args):
...
             f.write(out_string)
+
+
+DEFAULT_SIGN_IN_DURATION_MINUTES = 30
+
+
+class AwsSignIn(CommandBase):
+    command = 'aws-sign-in'
+    help = """
+        Use your MFA device to "sign in" to AWS for <duration> minutes (default {})
```
pr33thi: What does MFA mean?

dannyroberts: Multi Factor Auth

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -150,3 +159,116 @@ def run(self, args, unknown_args):
...
+             (Always the passed in profile followed by ':session')
+    """
+    aws_session_profile = '{}:session'.format(aws_profile)
+    if not force_new \
+            and _has_valid_session_credentials(aws_session_profile):
+        return aws_session_profile
+
+    default_username = getpass.getuser()
+    username = input("Enter username associated with credentials [{}]: ".format(
+        default_username)) or default_username
```
millerdev: On one machine I use regularly, the value returned by `getpass.getuser()` is not the same as my username. Would be nice to be able to set this in an environment variable or something to skip this prompt.

src/commcare_cloud/commands/terraform/aws.py
millerdev: Curious why you're using MFA and 2FA somewhat interchangeably? Does this mean we actually have the option of more factors than just 2 or could "2FA" be used everywhere "MFA" is in this PR?

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -150,3 +159,116 @@ def run(self, args, unknown_args):
...
+        SerialNumber=device_arn,
+        TokenCode=mfa_token,
+        DurationSeconds=duration_minutes * 60,
+    )['Credentials']
+    _write_credentials_to_aws_credentials(
+        '{}:session'.format(aws_profile),
+        aws_access_key_id=credentials['AccessKeyId'],
+        aws_secret_access_key=credentials['SecretAccessKey'],
+        aws_session_token=credentials['SessionToken'],
+        expiration=credentials['Expiration'],
```
millerdev: Suggestion: pass in `credentials` instead of splitting out all these arguments, making `_write_credentials_to_aws_credentials` the only piece of code that knows the structure of this data and how it is persisted.

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -150,3 +159,116 @@ def run(self, args, unknown_args):
...
+    config = configparser.ConfigParser()
+    config.read(os.path.realpath(aws_credentials_path))
+    if aws_profile not in config.sections():
+        return False
+    try:
+        expiration = datetime.strptime(config.get(aws_profile, 'expiration'), "%Y-%m-%dT%H:%M:%SZ")
+    except configparser.NoOptionError:
+        return False
+
+    return datetime.utcnow() < expiration
```
millerdev: What is the failure mode if this returns true but the credentials are actually no longer valid?

src/commcare_cloud/terraform/modules/iam/EnforceMFAPolicy.json
```diff
@@ -0,0 +1,98 @@
...
+                "iam:ListAccessKeys",
+                "iam:ListServiceSpecificCredentials",
+                "iam:ListMFADevices",
+                "iam:GetAccountSummary",
+                "sts:GetSessionToken"
+            ],
+            "Resource": "*",
+            "Condition": {
+                "BoolIfExists": {
+                    "aws:MultiFactorAuthPresent": "false"
```
millerdev: Is this the thing that does

> Forces 2FA for all our IAM users, including for CLI use

Who has permission to change this? Could I change it using my (now time-limited) credentials in `~/.aws/credentials`?

src/commcare_cloud/terraform/modules/iam/EnforceMFAPolicy.json
```diff
@@ -0,0 +1,98 @@
...
+                "iam:ListMFADevices",
+                "iam:ResyncMFADevice"
+            ],
+            "Resource": [
+                "arn:aws:iam::*:mfa/${aws:username}",
+                "arn:aws:iam::*:user/${aws:username}"
+            ]
+        },
+        {
+            "Sid": "AllowIndividualUserToDeactivateOnlyTheirOwnMFAOnlyWhenUsingMFA",
```
millerdev: Does this mean anyone can disable MFA for their own account?

dannyroberts:
> Is this now used for all access to AWS resources, including when I'm connected to the VPN and want SSH into a VM?
Sorry no, it's used for all API access to AWS, which includes the commands `aws-list`, `aws-fill-inventory`, `terraform`, and `terraform-migrate-state`. SSH into machines (and thus ansible etc) is not affected.

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: Interesting. That would probably be useful in a couple other places, one at least that I can think of. Can I make an issue for that and fix shortly in a followup?

src/commcare_cloud/terraform/modules/iam/EnforceMFAPolicy.json
dannyroberts: Yeah you could absolutely change this using your credentials once you had access to them. It's generally true that you can do all sorts of destructive things with your credentials once you have them, which is why we're putting emphasis on locking them down to a reasonable degree. If someone maliciously gained access, them lifting this policy wouldn't be at the top of the list of things I'm afraid they'd do. I hear you that there's a potential privilege escalation thing here—going from having credentials for 30 minutes to having credentials for as long as no one notices that the policy was lifted, and that's true. That might be something to revisit

src/commcare_cloud/terraform/modules/iam/EnforceMFAPolicy.json
dannyroberts: Yeah it does—but when they disable their MFA device, they'll no longer be able to enter heightened privilege mode. Without an MFA device configured—and a session created based on a token from that device—the user can only do very few things, such as look at the list of users and change their own password and MFA device.

dannyroberts: This is I think a bigger gap in my mind actually—if a someone malicious gained access to a user's access keys (but didn't have access to their MFA device i.e. phone), they could still use that to log in and change the MFA device to be their own phone, and then log in with that. I haven't tried that escalation attack, but I think it's possible, and straightforward

millerdev: The second one is pretty critical since it means the limit is easily disabled if someone malicious gains access. However, using MFA still seems better than not using it since it means local auth tokens stored on disk have an expiration date.

src/commcare_cloud/commands/terraform/aws.py
millerdev: I'm not particularly biased toward one or the other. I think the most important thing is to be consistent when we talk about it, especially in documentation.

src/commcare_cloud/terraform/modules/iam/EnforceMFAPolicy.json
dannyroberts: Okay turns out that was a false alarm—it was something that occurred to me as a potential vulnerability that I hadn't actually vetted yet. It turns out, and this makes sense, that the people who put together the policy suggestion and the docs to promote it https://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_users-self-manage-mfa-and-creds.html thought this through. I tried the attack both through the console and through the command line. From the console, I can't complete login without MFA anyway (once I've set a device), so that's the end of the game. (Going to other pages directly redirects to sign in.) From the command line, when I send a command to enable a new MFA device for my user, it says it can't because I'd have too many devices. If I try to deactivate my current one to make room for the new one, it says I don't have permissions to deactivate. Basically it does exactly the right things in these scenarios.

Sorry for the false alarm!

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: I think like I like the way it is better myself, where I keep parsing contained to the place that made the API error, and then let `_write_credentials_to_aws_credentials` have an explicit interface; having a `_write_credentials_to_aws_credentials` take as an argument something that looks like the result of an AWS STS API call I don't like as much. (Imagine another place that already had these things calling it and having to call it like `_write_credentials_to_aws_credentials({'AccessKeyId': aws_access_key_id, ...})`; not that that's likely to happen.) Anyway, I think doing it the way you suggested would be fine too, but I think I'm going to keep it the way it is.

PR #2408, opened 2018-11-15
Added cronjob to reload Nginx, made it a part of deploy_proxy.yml

NitigyaS:


src/commcare_cloud/ansible/letsencrypt_cert.yml
```diff
@@ -108,4 +110,13 @@
         hour: "0/12"
         user: root
         cron_file: certbot
-        state: absent
+
+    - name: reload nginx every 15 days
```
dannyroberts: Does letsencrypt already automatically update the cert every 15 days?

NitigyaS: It only updates in it 30 days.

PR #2413, opened 2018-11-16
Add ksampat to users for the purpose of VPN

dannyroberts:
Added ksampat with empty ssh key, which has the effect of giving him access to the productoin AWS VPN without ssh access to the machines.


mkangia:
> without ssh access to the machines.

Oooo Nice!

PR #2417, opened 2018-11-19
Fix barman setup in postgresql.conf

dannyroberts:
currently fails on undefined variable


pr33thi:
How come the provable deploy cron job didn't seem to catch this?

dannyroberts:
I wondered that myself, but I think the answer is that setting up pg barman is optional and the prove-deploy environment doesn't do it, so it never got to the offending line in the template file

PR #2418, opened 2018-11-19
Increase pillow0-production RAM to 128G

dannyroberts:
Upgrade to r5a.4xlarge

It kept occasionally maxing out on memory and killing helper processes (datadog, sshd, etc.), about once or twice a week. Was $368/mo, now is $660/mo (not including disk which is $15/mo in both cases)

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -190,10 +190,11 @@ servers:
     volume_size: 150
 
   - server_name: "pillow0-production"
-    server_instance_type: r5.2xlarge
+    server_instance_type: r5a.4xlarge
     network_tier: "app-private"
     az: "c"
     volume_size: 150
+
   - server_name: "formplayer0-production"
     server_instance_type: t3.large
     network_tier: "app-private"


mkangia:
Just curious, how did you roll this change out?

dannyroberts:
- Run `supervisorctl stop all` on the machine to allow everything to stop gracefully
- Run `cchq production terraform apply --skip-secrets`, which (because of this PR's changes to `terraform.yml`) stops the machine, updates its instance type, and starts it again
- Run `cchq production after-reboot pillow0` to remount any encrypted dirs (tmp dir in this case) and start the processes

mkangia:
> Run cchq production terraform apply --skip-secrets, which (because of this PR's changes to terraform.yml) stops the machine, updates its instance type, and starts it again

that is super sweet!
thanks @dannyroberts

environments/production/terraform.yml
```diff
@@ -190,10 +190,11 @@ servers:
     volume_size: 150
 
   - server_name: "pillow0-production"
-    server_instance_type: r5.2xlarge
+    server_instance_type: r5a.4xlarge
```
emord: why was r5 changed to r5a ? The difference I can see is that there are no ECU and its slightly cheaper. Was there a reason we weren't using r5a.2xlarge before?

dannyroberts: I think I just hadn't originally noticed r5a the first time around. What I wanted to do was just increase memory in the cheapest way possible without necessarily increasing CPU. It looked like r5a just doesn't make the same guarantees about CPU and that seemed fine.

I agree that this is worth revisiting at some point. In steady state we may not need such a large machine, or we may be able to use that capacity better to get higher throughput by tweaking the number of processes and other tuning.

PR #2420, opened 2018-11-19
Add couchproxy0-production

dannyroberts:
All traffic to couch goes through `couchdb2_proxy`, which was formerly just `couch0`. This has the unfortunate side effect of making it not okay to restart the `couch0` machine, something I'd like to do for debugging. Putting the proxy on its own machine seemed like the easiest way to go about this.

This is already rolled out


dannyroberts:
Previously:
- All HQ machines make requests to couch0
- couch0 acts as a proxy and distributes requests to couch0, couch1, couch2

Now:
- All HQ machines make requests to couchproxy0
- couchproxy0 acts as a proxy and distributes requests to couch0, couch1, couch2

Previously, I couldn't reboot the couch0 machine, because that would take down the proxy, and thus make couchdb unreachable.

Now, I can reboot the couch0 machine, because requests continue to go to couchproxy0, it notices couch0 isn't responding, and distributes all requests between couch1 and couch2 until couch0 is back up.

Of course, I can't reboot couchproxy0 without downtime now, but it has a much simpler job, and hopefully it won't be something we want to do. When it is, we just need to make a new proxy and switch things to point there.

mkangia:
Thank You @dannyroberts for the detailed write up.

PR #2341, opened 2018-10-31
Split case imports queue off of 'celery'

esoergel:
Case imports are currently in the catchall queue.  As I understand it, this PR won't actually change anything, but it'll allow us to track the queues separately and split this off elsewhere if it looks necessary.

In conjunction with https://github.com/dimagi/commcare-hq/pull/22246

Full disclosure, I have a fairly rudimentary understanding of the implications of this change - it's largely mirrored off of https://github.com/dimagi/commcare-cloud/pull/1500

PR #2422, opened 2018-11-20
Add changelog entry for addition of celery queue

esoergel:
Explain steps and rationale behind https://github.com/dimagi/commcare-cloud/pull/2341


snopoke:
I think you'll have to run `make docs` as well.

PR #2424, opened 2018-11-20
Update Ubuntu AMI

dannyroberts:
I've noticed that our EC2 instances print

```
*** System restart required ***
```
on login, on day one. I'm guessing that's because it immediately finds something to autoupdate, because we're not starting on the latest AMI for Ubuntu 14.04. This PR updates to the appropriate AMI currently listed on https://cloud-images.ubuntu.com/locator/ec2/ (Search for "us-east 14.04 amd64 hvm:ebs").

### DIFF
--- a/src/commcare_cloud/terraform/modules/server/main.tf
+++ b/src/commcare_cloud/terraform/modules/server/main.tf
@@ -16,7 +16,7 @@ resource aws_instance "server" {
     delete_on_termination = true
   }
   lifecycle {
-    ignore_changes = ["user_data", "key_name", "root_block_device.0.delete_on_termination", "ebs_optimized"]
+    ignore_changes = ["user_data", "key_name", "root_block_device.0.delete_on_termination", "ebs_optimized", "ami"]
   }
   tags {
     Name        = "${var.server_name}"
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/terraform/modules/server/main.tf
```diff
@@ -16,7 +16,7 @@ resource aws_instance "server" {
     delete_on_termination = true
   }
   lifecycle {
-    ignore_changes = ["user_data", "key_name", "root_block_device.0.delete_on_termination", "ebs_optimized"]
+    ignore_changes = ["user_data", "key_name", "root_block_device.0.delete_on_termination", "ebs_optimized", "ami"]
```
pr33thi: What does this do?

dannyroberts: Check out the commit message here: https://github.com/dimagi/commcare-cloud/pull/2424/commits/90b917dde46b809c4ae70a78be4bed843ec3a8a3

Adding a property to `ignore_changes` tells terraform that if that property on an existing resource differs from the desired value, just ignore it rather than destroying and recreating the resource. Here it means, don't destroy all of our current machines just because they were built from a different AMI.

pr33thi: makes sense, thanks

PR #2425, opened 2018-11-20
Temp: make client_max_body_size on couchdb2_proxy ridiculously high

dannyroberts:
I finally found what was blocking the cloudant to couchdb2 migration, and it was that couchdb2_proxy was blocking requests larger than 10MB, and not all but a good few _bulk_docs requests that happen during replication have been going over 10MB, getting an response from nginx, and failing to parse it as JSON, killing the replication process.

Setting the limit to something absurdly high like this fixes the problem, but we probably don't want to set the limit that high.

Pinging Simon and Emord in case you have insight into why these original numbers were chosen.

Update: the easiest way to deploy this change is
```
deploy-stack --tags=proxy --limit=couchdb2_proxy
```

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/couchdb2.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/couchdb2.yml
@@ -7,7 +7,7 @@ nginx_sites:
       port: "{{ couchdb2_port }}"
    file_name: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
    listen: "'{{ couchdb2_proxy_port }}'"
-   client_max_body_size: "16G"
+   client_max_body_size: "100M"
    upstream_port: "{{ couchdb2_port }}"
    proxy_set_headers:
    - "Host $http_host"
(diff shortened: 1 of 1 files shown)


dannyroberts:
@snopoke I lowered it to 100M which I think I'm comfortable with as a limit. I want it to be high enough that it's not interfering with the replication, but still serve as a last resort upper limit in case our application code is doing something _very_ wrong.

emord:
> Pinging Simon and Emord in case you have insight into why these original numbers were chosen.

https://github.com/dimagi/commcare-cloud/pull/775

Only reason it was 10 M instead of 1M (default?) is to fit the ICDS app in the response.

src/commcare_cloud/ansible/roles/nginx/vars/couchdb2.yml
```diff
@@ -17,7 +17,7 @@ nginx_sites:
      - name: /
        balancer: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
        proxy_redirect: ' off'
-       client_max_body_size: ' 10m'
+       client_max_body_size: '100M'
```
emord: I think this can be removed since it matches the above max body size

dannyroberts:
@emord thanks for the comments, makes sense. Just made a followup based on something your comments reminded me of: https://github.com/dimagi/commcare-cloud/pull/2441

PR #2429, opened 2018-11-21
remove CONN_MAX_AGE

snopoke:
* setting is not respected by celery < v4.2
  * https://github.com/celery/celery/pull/4292
* 'gevent' gunicorn worker class can 'leak' connections
  * https://github.com/benoitc/gunicorn/issues/996#issuecomment-93301359

After having this deployed for ~ 1 day I'm not seeing any improvement in performance and the large increase in open connections could cause issues on db hosts (e.g. max file descriptors). Given that and issues described above I think safest to revert.

reverts setting change from https://github.com/dimagi/commcare-cloud/pull/2396


snopoke:
rolled out to `prod` & `icds` (don't think anywhere else has the previous changes rolled out)

PR #2430, opened 2018-11-21
Add resources for staging postgresql on AWS

dannyroberts:
Will roll this out today before merging


environments/staging/inventory.ini
```diff
@@ -24,10 +24,19 @@ web1
 [hqdb1]
 hqdb1-staging.internal-va.commcarehq.org swap_size=1G
 
+[pgproxy0]
+10.201.40.127 hostname=pgproxy0-staging ec2=yes
+
+[rds_pg0]
+# RDS
+pg0-staging.czkracjslrn2.us-east-1.rds.amazonaws.com
```
mkangia: interesting stuff there.
Are we going to use AWS's RDS for production as well?
Just curious if there are any limitations to it?

dannyroberts: Yeah moving to RDS has for a long time been part of the plan to move to AWS. There are some limitations to it—for example we can't install any pg extensions except a short list they support, and we don't have access to the disk so we can't just rsync the data directory and call it a day. Longer term the hope is that it lowers the operational overhead while maintaining high availability. For example, RDS comes with support for multi-zone replication as well as for read replicas, as something you can just configure by changing some settings.

mkangia: Cool. Thanks @dannyroberts .
Hopefully i mis read, but they also noted somewhere that they don't allow a db shell access. Is that true?

dannyroberts: You can certainly still enter a psql shell by running

```
psql -h pg0-staging.czkracjslrn2.us-east-1.rds.amazonaws.com -p 5432 -U hqweb2postgres
```
from one of the machines in app-private (like web0) or db-private. So yeah I don't think that's a problem. Did you mean something different from that?

mkangia: i read here
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Welcome.html

```
To deliver a managed service experience, Amazon RDS doesn't provide shell access to DB instances, and it restricts access to certain system procedures and tables that require advanced privileges.
```

dannyroberts: Ah okay. Yeah in that sentence they're not referring to a "db shell" (i.e. a REPL in which you can make commands to the db service, like `psql`), they're referring to shell access to the machine, e.g. ssh access or access through a console to run bash commands directly. So there's no way to ssh in, but you can certainly use `psql` (or pgadmin, etc. if your network is set up right)

mkangia: That makes sense, thanks @dannyroberts

PR #2431, opened 2018-11-21
Add redis0-staging (to also act as shared_dir_host)

dannyroberts:
In the process of rolling this out, will merge when done


dannyroberts:
OK, deployed and ready to merge!

environments/staging/inventory.ini.j2
```diff
@@ -148,6 +147,7 @@ airflow0
 {{ control-staging }} hostname=control-staging ec2=yes
 
 [ansible_skip:children]
+rds_pg0
```
mkangia: what does this change do?

dannyroberts: This makes this endpoint (which isn't an actual ssh-able machine because it's RDS) get skipped on most ansible commands; otherwise it'll hang and then tell you it can't ssh into the rds URL (not surprising) when it times out.

mkangia: Sweet!

PR #2434, opened 2018-11-26
Enable Parallel query

NitigyaS:
To be merged after https://github.com/dimagi/commcare-cloud/pull/2433/files


src/commcare_cloud/ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -73,6 +73,12 @@ effective_cache_size = {{ postgresql_effective_cache_size }}
 enable_nestloop = {{ enable_nestloop }}
 {% endif %}
 
+# PARALLEL QUERY
+{% if postgresql_max_parallel_degree > 0 %}
```
NitigyaS: the only thing that must be done in order to take advantage of parallel query is to set max_parallel_degree to a value greater than zero. A value between 1 and 4 is recommended. The number should be set with due regard for max_worker_processes, which limits the number of background worker processes that may exist at any one time. This value includes not only parallel workers but any other background worker processes which may have been configured
https://wiki.postgresql.org/wiki/Parallel_Query

src/commcare_cloud/ansible/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -73,6 +73,12 @@ effective_cache_size = {{ postgresql_effective_cache_size }}
...
+max_parallel_degree = {{ postgresql_max_parallel_degree }}
+max_worker_processes =  {{ postgresql_max_worker_processes }}
```
emord: should these have defaults as well?

NitigyaS: It will only be set when `postgresql_max_parallel_degree` is more than 0 and in `defaults.yml` it is 0

emord: Correct. So by default if someone sets `postgresql_max_parallel_degree = 1`, but didn't explicitly set `postgresql_max_worker_processes = x` wouldn't that  result in an invalid postgres config?

NitigyaS: Ohh right! Thanks, I Will fix that.

PR #2439, opened 2018-11-27
Fix terraform for user with empty public key file

dannyroberts:
This would previously fail (harmlessly) every time you run terraform because aws rejects the empty public key


pr33thi:
Should we add a message if the user doesn't have a public key? They might not realize that this logic is being skipped

dannyroberts:
I just fixed a mistake in my commit—I meant to just not add a keypair for them if they don't have a public key, but instead I didn't create a user for them at all either. Now it still creates an IAM user but doesn't try to add a keypair.

For this I don't think there's much use trying to warn about a missing public key. You have to add a blank file in the place of the public key file in order to make things work, so if you get to this point, it's robably on purpose.

PR #2440, opened 2018-11-27
Bring production terraform.yml into line with aws

dannyroberts:
I manually increased the size of a machine before realizing it was pointless;
reducing the size is a harder operation than increasing the size,
so temporarily updating to match reality

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -155,7 +155,7 @@ servers:
     server_instance_type: m5.4xlarge
     network_tier: "db-private"
     az: "b"
-    volume_size: 4000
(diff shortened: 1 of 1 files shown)


pr33thi:
What do you mean by 
>reducing the size is a harder operation than increasing the size
?

dannyroberts:
In the AWS console (and presumably through the API as well) you can change a volume's size in place without detaching it or stopping the VM—pretty neat! However, you can only _increase_ the volume's size in place: to decrease a volume's size you have to go through a more annoying and manual process of copying the data to a smaller volume.

PR #2447, opened 2018-11-28
Add prefetch multiplier

mkangia:
Following up on https://github.com/dimagi/commcare-hq/pull/22500#issuecomment-442479071

http://docs.celeryproject.org/en/latest/userguide/configuration.html#std:setting-worker_prefetch_multiplier

Prefetch multiplier is 4 by default.
Setting it to 2 to first test a theory on how prefetch count is calculated

and then would set it to 1 to disable prefetch for saved exports

### DIFF
--- a/src/commcare_cloud/environment/schemas/app_processes.py
+++ b/src/commcare_cloud/environment/schemas/app_processes.py
@@ -17,6 +17,7 @@ class CeleryOptions(jsonobject.JsonObject):
     concurrency = jsonobject.IntegerProperty(default=1)
     pooling = jsonobject.StringProperty(choices=['gevent', 'prefork'], default='prefork')
     max_tasks_per_child = jsonobject.IntegerProperty(default=50)
+    prefetch_multiplier = jsonobject.IntegerProperty()
     num_workers = jsonobject.IntegerProperty(default=1)
     optimize = jsonobject.BooleanProperty(default=False)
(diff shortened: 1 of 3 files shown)


snopoke:
I thought the optimize flag was supposed to disable prefetch

mkangia:
@snopoke its actually enabled for all workers that don't use gevent. I am unable to check on others as of now because they don't have much activity in them.

src/commcare_cloud/environment/schemas/app_processes.py
```diff
@@ -17,6 +17,7 @@ class CeleryOptions(jsonobject.JsonObject):
     concurrency = jsonobject.IntegerProperty(default=1)
     pooling = jsonobject.StringProperty(choices=['gevent', 'prefork'], default='prefork')
     max_tasks_per_child = jsonobject.IntegerProperty(default=50)
+    prefetch_multiplier = jsonobject.IntegerProperty()
```
mkangia: setting no default value so that it's skipped when passing to celery command

mkangia:
this actually failed because we are running a old version of celery which needs this to be added as a setting 
http://docs.celeryproject.org/en/3.1/configuration.html#std:setting-CELERYD_PREFETCH_MULTIPLIER

learning now how to add that.
sorry guys

PR #2450, opened 2018-11-28
limit pna requests

jmtroth0:
@snopoke this is what I was suggesting in the email about the okhttp retries. Let me know if you think it makes sense. I also haven't tested it, which probably should be done before this goes out.

buddy: @kaapstorm

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -17,6 +17,13 @@ nginx_sites:
       path: "{{ www_home }}/hq_media/cache"
       max_size: "2g"
       inactive_time: "24h"
+   limit_req_zones:
+    - name: "restore"
+      zone: "$digest_submission"
+      rate: "6r/s"
+    - name: "submissions"
+      zone: "$digest_submission"
+      rate: "30r/s"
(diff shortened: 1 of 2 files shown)


jmtroth0:
also pinging @emord since it looks like you did a bunch of the things around this?

src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -17,6 +17,17 @@ nginx_sites:
       path: "{{ www_home }}/hq_media/cache"
       max_size: "2g"
       inactive_time: "24h"
+   mappings:
+    - name: "$digest_submission"
+      variable: "$http_authorization"
+      value: "$server_name"
+   limit_req_zones:
+    - name: "restore"
```
snopoke: this zone is unused

src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -17,6 +17,17 @@ nginx_sites:
...
+      zone: "$digest_submission"
+      rate: "6r/s"
+    - name: "submissions"
+      zone: "$digest_submission"
```
snopoke: I don't think we should use that zone as it's global. We want to restrict it per IP right? In that case we should use $binary_remote_addr

PR #2451, opened 2018-11-28
Temporarily comment out web0 from the staging inventory

pr33thi:
I have removed web0 from the rotation for staging so that it can reindex the couch databases on the new aws couch machine. It needs to be commented out for staging deploys to still work. This is a temp fix, and will be reverted after the aws couch migration is complete.

@dannyroberts
fyi @gbova  
code buddy @calellowitz

### DIFF
--- a/environments/staging/inventory.ini
+++ b/environments/staging/inventory.ini
@@ -10,15 +10,16 @@
 # Amazon EC2
 proxy1
 
-[web0]
-10.201.10.104 swap_size=1G hostname=web0-staging ec2=yes
+# Temporarily commenting this out while it's removed from the rotation for the aws migration's couchdb2 reindex
+# [web0]
+# 10.201.10.104 swap_size=1G hostname=web0-staging ec2=yes
 
 [web1]
 10.201.11.46 swap_size=1G hostname=web1-staging ec2=yes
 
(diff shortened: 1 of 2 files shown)


environments/staging/inventory.ini
```diff
@@ -10,15 +10,16 @@
 # Amazon EC2
 proxy1
 
-[web0]
-10.201.10.104 swap_size=1G hostname=web0-staging ec2=yes
+# Temporarily commenting this out while it's removed from the rotation for the aws migration's couchdb2 reindex
+# [web0]
+# 10.201.10.104 swap_size=1G hostname=web0-staging ec2=yes
```
dannyroberts: I think you actually don't want to comment this part out, because you still want it to be under django_manage, right?

dannyroberts: Mostly for yourself

pr33thi:
> Nit: commend => comment

Thanks, rebased that

environments/staging/inventory.ini
pr33thi: https://github.com/dimagi/commcare-cloud/pull/2451/commits/3df39cc1324009d95847b02fd133325ab0bec8fa

PR #2452, opened 2018-11-29
Add 5TB disk to couch*-production

dannyroberts:
I've added a 5TB non-boot volume to each of the production couch machines. (They're already added.)

`couch1-production` already has its data migrated to this new volume, and that volume is properly mounted at `/opt/data/`.
As of now (2018-11-28 @ 10:30pm ET) `couch0-production` and `couch2-production` are in the process of copying their `/opt/data` to the new volume. This is being done in a tmux under the ansible user as an rsync in a loop. Both nodes are still operational, it's just copying the data in the background. Tomorrow morning when I wake up, I'm going to (one node at a time) shut off couch, do a final rsync, reboot the machine to make sure the drives are in their natural unmounted state, move the old /opt/data out of the way (e.g. to /opt/data-backup), and then mount the new volume at /opt/data using after-reboot.

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -234,13 +234,13 @@ es_data
 
 
 [couch0]
-10.202.40.211 hostname=couch0-production ec2=yes
+10.202.40.211 hostname=couch0-production ec2=yes datavol_device=/dev/nvme1n
 
 [couch1]
 10.202.41.83 hostname=couch1-production ec2=yes datavol_device=/dev/nvme1n
 
 [couch2]
-10.202.42.85 hostname=couch2-production ec2=yes
+10.202.42.85 hostname=couch2-production ec2=yes datavol_device=/dev/nvme1n
 
 [couchdb2:children]
 couch0
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -234,13 +234,13 @@ es_data
 
 
 [couch0]
-10.202.40.211 hostname=couch0-production ec2=yes datavol_device=/dev/nvme1n
+10.202.40.211 hostname=couch0-production ec2=yes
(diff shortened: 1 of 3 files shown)


mkangia:
whats the purpose of this change @dannyroberts ?

dannyroberts:
@mkangia Currently the data's on boot volumes, and the AWS Linux AMIs we're using have a hard limit for boot partitions of 2TB; so even though the disk we've allocated is 4TB, I realized last week that we are only able to use 2TB of the disk (on each couch node). This change is to put `/opt/data` on a non-boot volume, allowing us to increase its size arbitrarily (up to whatever AWS's ridiculously high volume size limit is)

dannyroberts:
@mkangia https://en.wikipedia.org/wiki/System_partition_and_boot_partition I suppose? Basically a VM (or any computer) needs to have a disk designated as the "startup disk", where the core code need to start the operating system lives. There's also the "system" partition which is usually mounted at `/`, so like the main disk that comes with all the default stuff like `/ect` and all those other dirs. Often (as in our case) the boot and system partitions are the same partition on the same volume. In AWS in particular, the AMI (machine image) expects there to be one volume designated as the boot volume. Then, you can add other volumes that have no special function—they're just there for you to format and mount however and wherever you want. In our case, we're adding one of these—a non-boot volume—where we're going to be storing couch data.

dannyroberts:
In AWS docs I think what I've been calling the "boot volume" they call the "Root Device Volume": https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/RootDeviceStorage.html

mkangia:
Thanks much @dannyroberts, that looks like the right way of how things should be. Glad we are doing that now.

PR #2457, opened 2018-11-30
Increase RAM on celery machines

pr33thi:
This change has already been applied (I didn't apply it to celery2 because @dannyroberts did)

@dannyroberts 
code buddy @calellowitz


dannyroberts:
Thanks!

PR #2467, opened 2018-12-03
Manage RDS (postgresql) parameters in terraform

dannyroberts:
Builds off of https://github.com/dimagi/commcare-cloud/pull/2465.

The only param currently managed (after this change) is `max_connections`—everything else is ignored—but it's set up to make it very easy to manage other params, basically by just adding them to a whitelist mapping the param name to the ansible variable name. (Currently assumes it'll be one of the variables in `postgresql/defaults/main.yml`).


pr33thi:
Does this make it so that you have to do a terraform apply instead of deploy-stack to get the max_connections change in?

dannyroberts:
@pr33thi This only affects RDS, so if you're referring to the test you're referring to changing `max_connections` on postgresql (on a VM) then this doesn't affect that at all. For example what you have planned related to https://github.com/dimagi/commcare-cloud/pull/2468 isn't affected

dannyroberts:
(I just rebased onto master after merging https://github.com/dimagi/commcare-cloud/pull/2465)

PR #2468, opened 2018-12-03
Test prod without pgbouncer

pr33thi:
Reason: RDS doesn't allow us to install pgbouncer on the postgrs machines themselves, so the plan is to increase the connection limit and just have pgbouncer installed on the proxy machine. We're applying this change to production now, so we can make sure that doesn't cause any big problems. Note: I haven't applied these changes yet, but did run in check mode

@dannyroberts 
code buddy @calellowitz


pr33thi:
Just applied the changes, merging this now

PR #2012, opened 2018-08-09
Moving Authentication from postgresql to pgbouncer

NitigyaS:

PR #2475, opened 2018-12-04
Bring back hba

dannyroberts:
This is to fix the downtime we just had

### DIFF
--- a/src/commcare_cloud/ansible/roles/postgresql/templates/pg_hba.conf.j2
+++ b/src/commcare_cloud/ansible/roles/postgresql/templates/pg_hba.conf.j2
@@ -38,7 +38,7 @@ host    all     {{ postgres_users.tableau_readonly.username }}   {{ tableau_serv
 {% endif %}
 
 {% if allow_direct_postgresql_access %}
-{% for host in (groups['formplayer'] + groups['webworkers'] + groups['celery'] + groups['pillowtop'] + groups.get('django_manage', []) + groups['postgresql'] + groups.get('airflow', [])) | unique %}
+{% for host in (groups['formplayer'] + groups['webworkers'] + groups['celery'] + groups['pillowtop'] + groups.get('django_manage', []) + groups['postgresql'] + groups.get('pg_standby', []) + groups.get('airflow', [])) | unique %}
 # {{ host }}
 {% for db_name, db_list in postgresql_dbs.all|groupby('name') %}
 {% set db = db_list[0] %}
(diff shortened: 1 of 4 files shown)


src/commcare_cloud/ansible/roles/postgresql/templates/pg_hba.conf.j2
```diff
@@ -37,6 +37,22 @@ host    all             {{ postgres_users.backup.username }}   {{ db_vlan_ip_ran
 host    all     {{ postgres_users.tableau_readonly.username }}   {{ tableau_server.split(':')[0] }}/32    md5
 {% endif %}
 
+{% if allow_direct_postgresql_access %}
+{% for host in (groups['formplayer'] + groups['webworkers'] + groups['celery'] + groups['pillowtop'] + groups['django_manage']) | unique %}
```
calellowitz: I think you need to include all the other `postgres` databases (or at least the proxies). If this is going out for staging you will also need `airflow`

calellowitz: you may want to use the same list as here
https://github.com/dimagi/commcare-cloud/pull/2012/files#diff-0685ce702f395896cae62742c0a14fcdR17 (or at least be conscious about which of those you leave out)

dannyroberts: Yeah I see what you mean. I did it this way for exactly what we needed to roll out for https://github.com/dimagi/commcare-cloud/pull/2468, which doesn't take down pgbouncer or anything, but just has all application processes hit postgresql at 5432 directly. So in the immediate term I just want to say this shouldn't affect anything else, like postgres standbys or anything.

In the slightly longer term, we're about to migrate postgres to AWS (RDS), after which we won't need this. I'm tempted to say we just leave this PR here until after the RDS migration and then close it without ever merging in. I guess the error mode there is someone updates postgres confs in the next three days and by mistake undoes what I deployed with this change. So another possibility is to merge this in but also create a revert PR that we just wait to merge in after I migrate postgres to AWS RDS this weekend.

calellowitz: How does pl/proxy connect? Does this make it connect directly?

calellowitz: :+1:

PR #2476, opened 2018-12-05
simpler renaming

sravfeyn:


sravfeyn:
@snopoke 
cc @nickpell 

This is already deployed on staging

PR #2479, opened 2018-12-05
Adding new celery machinex

NitigyaS:
@emord could you add the details in app-process.yml
This is the scale up to support for 400K users. 
I am planning to get these details before everyone goes on leave.


environments/icds-cas/app-processes.yml
```diff
@@ -377,3 +382,5 @@ pillows:
   'pillow15':
     FormSubmissionMetadataTrackerPillow:
       num_processes: 16
+  'pillow21':
+  'pillow22':
```
emord: These aren't in the inventory file. Also I don't have anything to add to them at the moment. I think we should be fine with our current setup until Sravan's work to combine processors is out in January

environments/icds-cas/app-processes.yml
```diff
@@ -151,6 +151,51 @@ celery_processes:
...
+    icds_dashboard_reports_queue:
+      concurrency: 2
+  'celery15':
+    ucr_indicator_queue:
+      concurrency: 4
+    reminder_case_update_queue:
+      pooling: gevent
+      concurrency: 5
+      num_workers: 6
+    icds_dashboard_reports_queue:
```
emord: To be honest, I don't really have a system for these. I'm just copying the most common celery config as I don't know of any bottlenecks we have at the moment.

PR #2481, opened 2018-12-05
add missing standby databases

snopoke:


snopoke:
They did cause issues once I rolled out the load balancing change that referenced them

PR #2482, opened 2018-12-05
Revert "Test prod without pgbouncer"

pr33thi:
Reverts dimagi/commcare-cloud#2468

@calellowitz 
@dannyroberts 
Reverting the whole pr (not just the port changes) because the connections are maxed at 290 with pgbouncer, so the increase to 1000 would have no effect:  https://github.com/dimagi/commcare-cloud/blob/master/environments/production/postgresql.yml#L6


pr33thi:
already discussed with @calellowitz, going to go ahead and merge this

PR #2483, opened 2018-12-06
Allow pgbouncer in front of remote dbs, and configure on staging

dannyroberts:
I tried or thought about a couple different approaches, and this seemed to be by a good margin the simplest. Basically there's a new `pgbouncer_host` option on each db, and it defaults to the same machine; throughout where looking up what host to access a db on, we'll now look at `pgbouncer_host` instead of `host`. On envs we're not changing, this should create no diff.

~Haven't yet been able to test this against staging (bad internet), but should be ready for review of the concept.~

### DIFF
--- a/environments/staging/postgresql.yml
+++ b/environments/staging/postgresql.yml
@@ -1,9 +1,5 @@
 DEFAULT_POSTGRESQL_HOST: rds_pg0
 
-host_settings:
-  rds_pg0:
-    port: 5432
-
 override:
   pgbouncer_pool_mode: transaction
   pgbouncer_max_connections: 800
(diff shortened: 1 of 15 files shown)


dannyroberts:
I just rolled this out to staging, and both the diff and the effect on staging looked great! (staging was getting lots of errors that I think are essentially RDS's version of hitting max connections limit, and they went away right when I restarted services to route queries through the shared pgbouncer)

environments/staging/postgresql.yml
```diff
@@ -12,29 +8,41 @@ override:
 
 dbs:
   main:
+    pgbouncer_host: pgproxy0
```
snopoke: I think it could get confused with `plproxy` on envs that have it so maybe call it `pgbouncer0`?

dannyroberts: Ah, well on staging I'm just using the pgbouncer that's already _on_ the plproxy machine (`pgproxy0-staging`), so in this case that is what I'm doing. On production though that's a good suggestion; I'll make a separate pgbouncer0 so not all traffic's going through one machine.

I'll probably have the partition dbs continue to use pgbouncer on the plproxy machine, since anything else would require kind of useless bouncing around between machines.

PR #2484, opened 2018-12-06
Make airflow config update on update-config

dannyroberts:


src/commcare_cloud/ansible/deploy_localsettings.yml
```diff
@@ -4,3 +4,4 @@
 - import_playbook: deploy_riakcs.yml
 - import_playbook: deploy_commcarehq.yml
 - import_playbook: deploy_formplayer.yml
+- import_playbook: deploy_airflow.yml
```
calellowitz: I am not very familiar with this playbook, but the way that it is used will just run the task you tagged localsettings below?

dannyroberts: This playbook is only ever meant to be run with the `--tags=localsettings`, as is done in `update-config`. (But yeah if you ran this without filtering to a tag, it would definitely do a lot more than just update localsettings.) Even though it's semantically no different from deploy_stack.yml with `--tags=localsettings`, someone originally split this file out so that it would run a lot faster. Bit of a hack to optimize around how slow it is to run our full stack even in the degenerate case when no tags are matched.

PR #2488, opened 2018-12-06
Make our RDS postgresql param support more complete

dannyroberts:
- Now supports params for individual instances (rather than just environment-global ones)
- Now supports params that have a unit attached to them ("90MB", etc.) as opposed to just a simple int like max_connections
- Rolled it out to just one production rds instance to make sure it works for when we need in a theoretical firefighting situation (but not to all of them, because we're moving db-to-machine allocations around and I want to give RDS's "sensical" defaults a try on the new allocation first.)

### DIFF
--- a/src/commcare_cloud/commands/terraform/tests/test_postgresql_units.py
+++ b/src/commcare_cloud/commands/terraform/tests/test_postgresql_units.py
@@ -1,0 +1,51 @@
+from nose.tools import assert_equal, assert_raises
+from parameterized.test import assert_contains
+
+from commcare_cloud.commands.terraform.postgresql_units import convert_to_unit, SECOND, \
+    BLOCK_8KB, MS, convert_to_standard_unit
+
+
+def test_convert_to_unit__time():
+    # 1 minute is 60 seconds
+    assert_equal(convert_to_unit('1m', SECOND), 60)
+    assert_equal(convert_to_unit('1 m', SECOND), 60)
+
+
+def test_convert_to_unit__bytes():
+    # 12 GB is 12 * 1024 * 1024 / 8 "8kB blocks" = 1572864 "8kB blocks"
+    assert_equal(convert_to_unit('12GB', BLOCK_8KB), 1572864)
+    assert_equal(convert_to_unit('12 GB', BLOCK_8KB), 1572864)
+
+
+def test_convert_to_unit__bad_input():
+    # number part must be an int
+    with assert_raises(ValueError) as context:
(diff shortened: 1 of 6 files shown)


src/commcare_cloud/commands/terraform/tests/test_postgresql_units.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_convert_to_standard_unit__time():
+    # authentication_timeout is measured in seconds
+    assert_equal(convert_to_standard_unit('authentication_timeout', '1m'), 60)
+
+
+def test_convert_to_standard_unit__bytes():
+    # effective_cache_size is measured in 8kB blocks
+    # 11520MB = 11520 * 1024 / 8 "8kB blocks" = 1474560 "8kB blocks"
```
snopoke: I'm a bit confused why you're converting to number of 8kB blocks. I don't see that anywhere in the docs. Also I think you can just specify `11520MB` directly and PG will do the conversion.

dannyroberts: I sort of touched on it in the commit message for https://github.com/dimagi/commcare-cloud/pull/2488/commits/6cfcd7a76423678b4c04bc4ff6f4c86c5135c3f8
> Each param has its own standard unit associated with it, and RDS doesn't
allow you to pass in a string that includes the unit the way postgres
conf does; it requires you to convert to the correct standard unit
first.

but I agree it's a little confusing and surprising.

My understanding is that internally postgresql has these standard units as well, but also lets you specify your own units in postgresql.conf and then does the conversion for you. RDS for some reason forces you to do the conversion and give them just an int in the right standard unit. They cover this and tell you the standard unit for each param in https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.html#Appendix.PostgreSQL.CommonDBATasks.Parameters. (I copied the param mapping in `postgresql_units.py` straight from this table.)

dannyroberts: For context I ran into this exactly because I first just passed in a value with a unit ("11520MB") and RDS complained.

snopoke: Thanks @dannyroberts that makes sense (sorry I missed the commit message)

dannyroberts: No worries, I kind of buried the lede on this one

PR #2490, opened 2018-12-06
Add pgproxy0-production and pgbouncer0-production for RDS migration

dannyroberts:
We'll need to have these in the new setup. They're not hooked up yet,
but I want them to be ready to be when I start downtime for the
migration.


pr33thi:
will requests go like:

[pgproxy] -> [pgbouncer] -> [the pg machine/s that correspond to the request]

?

dannyroberts:
I had an exchange with @snopoke where I wrote out how I was planning to divide it up. https://github.com/dimagi/commcare-cloud/pull/2483#discussion_r239358479.

Since that PR, each db (including those on RDS) has the opportunity to declare in `postgresql.yml` where it's pgbouncer will be hosted, using the `pgbouncer_host` property.

This PR doesn't yet do that assignment, but the plan is for `form_processing.proxy` and `form_processing.partition.p[1-5]` to have `pgproxy0` as their `pgbouncer_host` and for all other dbs to have `pgbouncer0` as their `pgbouncer_host`.

The reasoning is this basically this
- it makes sense for `pgproxy0`, where the `form_processing.proxy` db (called `commcarehq_proxy`) lives, to go host pgbouncer for that db. (Where possible it makes sense to have pgbouncer on the same machine as the dbs its serving, since that just means less network traffic.)
- it makes sense for `pgproxy0` to also host pgbouncer for the `form_processing.partition.p[1-5]` dbs, because almost all traffic to those dbs already goes through the `form_processing.proxy` db, so it already has to go through `pgproxy0` anyway. Picking a different host for the partition dbs would just add network traffic without much of a benefit that I can think of.
- would be nice not to have _all_ traffic for _all_ dbs going through the same machine, `pgproxy0`, so we're going to have a separate machine, `pgbouncer0` just host pgbouncer and have that pgbouncer instance service all other dbs (main, formplayer, ucr).

That's the plan anyway. The configuration is pretty flexible so if we want, we can do something different later, like having every RDS instance have a dedicated pgbouncer machine or something. My estimate though is that this setup, with just one more machine than the bare minimum (`pgbouncer0`) will be perfectly fine for us for now.

PR #2495, opened 2018-12-07
couchdb fixes to prevent accedental shard deletion

snopoke:
@dimagi/scale-team FYI


src/commcare_cloud/commands/migrations/couchdb.py
```diff
@@ -96,8 +97,16 @@ def generate_shard_prune_playbook(migration):
     """Create a playbook for deleting unused files.
     :returns: List of nodes that have files to remove
     """
-    full_plan = {plan.db_name: plan for plan in migration.shard_plan}
-    _, deletable_files_by_node = get_node_files(migration.source_couch_config, full_plan)
+    # get shard allocation from DB directly instead of using plan in case they are different
```
emord: Should there be something here to print out a message if they are different?

snopoke: updated to add warning

PR #2497, opened 2018-12-07
Added new ES Machines

NitigyaS:


NitigyaS:
I believe this will require a downtime, Since Master node changes and all nodes will need to redirect to new master.

snopoke:
not 100% sure if it will require downtime but either way when we add nodes it rebalances the data which will make things slow so definitely safest if we have downtime

PR #2499, opened 2018-12-07
Tune production RDS for pg_restore (revert after migration)

dannyroberts:
Follows https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_BestPractices.html#CHAP_BestPractices.PostgreSQL for what you should do to speed up a `pg_restore`


dannyroberts:
I talked this over with Cal and deployed. I've made a note to revert this before bringing everything back up.

PR #2502, opened 2018-12-08
ES version changed from 1.7.3 to 1.7.6

sanjay2916:


dannyroberts:
I think this is only failing because it's a PR from a fork

PR #2509, opened 2018-12-11
Remove unused webworkers and "rightsize" the rest

dannyroberts:
I'd originally made webworkers absurdly large to rule out resource scarcity as a cause during firefighting of Oct slowness. After looking at the eye-popping numbers from Nov, I think it's time to go back to a reasonable size. Rolling these changes out will save us $1,600/mo. (The first commit just removes machines that were already stopped, so it doesn't affect anything—performance or price.)

From looking at datadog graphs (https://app.datadoghq.com/dash/510484/host-groups?tile_size=m&page=0&is_auto=false&from_ts=1541937600000&to_ts=1544529600000&live=true&tpl_var_host_group=webworkers), I made some estimates about peak resource usage on the two webworker types, and settled on the following:

hq_webworkers: 33% the CPU, 67% the RAM
mobile_webworkers: 25% the CPU, 50% the RAM

m5.2xlarge (old) and r5.xlarge (new) both have 32GB RAM, but r5.xlarge has half the CPU (4 instead of 8). Also going from 3 to 2 hq_webworkers and 6 to 3 mobile_webworkers.

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -27,15 +27,11 @@ hqproxy0
 [hq_webworkers:children]
 web0
 web1
-web2
 
 [mobile_webworkers:children]
 web6
 web7
 web8
-web9
-web10
-web11
 
 [webworkers:children]
 hq_webworkers
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -14,9 +14,6 @@ hqproxy0
 [web1]
 10.202.10.84 hostname=web1-production ec2=yes datadog_integration_tcp_check=yes
 
-[web2]
-10.202.10.238 hostname=web2-production ec2=yes
-
 [web6]
 10.202.11.98 hostname=web6-production ec2=yes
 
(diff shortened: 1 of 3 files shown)


millerdev:
Do we still have separation of submission/restore workers and other web workers? I don't think you've changed anything related to that here, but just checking.

dannyroberts:
> How come you're removing 2-5 instead of just removing the higher numbered ones? It could get confusing to have #s in the middle missing

Originally 0-5 were "hq_webworkers" and 6-11 were "mobile_webworkers", when I stopped a few hq_webworkers a while back, I decided to keep the split rather than making some of the hq_webworkers mobile_webworkers, and in this PR I'm doing the same. It's mostly just for the ability to track over time on datadog, it seemed easier if the two sets of mobile workers don't have machines hopping from one to the other. Not a huge deal either way, but historically we've been fine not keeping a pretty numbering sequence on machines.

millerdev:
I think you answered my question in the answer to @pr33thi's question. Thanks!

pr33thi:
thanks!

dannyroberts:
@pr33thi will take next steps on rolling this out, tracking here: https://trello.com/c/aFkbmMGR/11-downsize-webworkers, and will merge once that's complete.

pr33thi:
Just applied the changes, merging this now

PR #2510, opened 2018-12-11
Adding kafka in inventory

NitigyaS:
@snopoke I think some changes also need to be done in app-process.yml
as per the https://github.com/dimagi/commcare-cloud/blob/master/docs/services/kafka.md


emord:
how of much of htese servers been deployed? 

Last time kafka was partially deployed and they joined the cluster, but they didn't proactively get any partitions. This meant htat when I added partitions, the new nodes got partitions even though no pillows wervers were reading from them.

basically i just want to make sure that these nodes haven't joined the cluster yet

PR #2512, opened 2018-12-11
keep ES on Java7 until ready to roll it out

snopoke:


NitigyaS:
Hey Approving and merging this , To proceed ahead with ES on ICDS.

PR #2519, opened 2018-12-13
enable staging analytics

biyeun:
@orangejenny / @emord 

enables hubspot, kissmetrics, and drift on staging for testing analytics workflows


orangejenny:
Are these the production accounts or test accounts? I thought the kiss sandbox account was already in the vault and that we didn't have a test account for hubspot.

biyeun:
@orangejenny the kiss sandbox account wasn't in the vault, so I added it. we need both hubspot and drift to be there to trigger any forms. btw any periodic hubspot analytics sent over are always prepended with the environment

orangejenny:
> any periodic hubspot analytics sent over are always prepended with the environment

Okay, good. Definitely don't want staging data disturbing growth's analysis.

PR #2521, opened 2018-12-15
Add blob metadata migrations to change log

millerdev:
@dannyroberts @snopoke cc @nickpell

### DIFF
--- a/docs/changelog/0008-blob-metadata-migration-p1.md
+++ b/docs/changelog/0008-blob-metadata-migration-p1.md
@@ -1,0 +1,87 @@
+# Blob Metadata Migration Process - part 1
+
+This migration will consist of running a series of commands against your cluster
+to migrate blob metadata from CouchDB to SQL.
+
+## Migration steps
+
+Before executing any shell commands, set the `ENV` variable to the name of the
+CommCareHQ environment you are migrating.
+
+```sh
+ENV=your_environment_name
+```
+
+1. Start a new screen session on a server in your cluster:
+
+   ```sh
+   cchq $ENV ssh django_manage -t screen -S blobmeta
+   ```
+
+   If you get disconnected for any reason, you can reconnect with this command:
+
+   ```sh
+   cchq $ENV ssh django_manage -t screen -x blobmeta
+   ```
+
+   Do all of the following commands within this screen session
+
+2. Create a directory for migration logs
+
+   ```sh
+   su ansible
+   sudo mkdir /opt/data/blobmeta-migration-logs
+   sudo chown -R cchq:cchq /opt/data/blobmeta-migration-logs
+   exit
+   ```
+
+3. Prepare to run migration management command
+
+   ```sh
+   sudo -iu cchq
+   ENV=YOUR_ENVIRONMENT_NAME
+   cd ~/www/$ENV/current
+   source python_env/bin/activate
+   ```
+
+4. Run the management command
+
+   ```sh
+   ./manage.py run_blob_migration migrate_metadata \
+   --log-dir=/opt/data/blobmeta-migration-logs \
+   --chunk-size=1000
+   ```
+
+   You should see some output similar to this:
+
+   ```
+   Processing XXXX documents: CommCareAudio, CommCareBuild, CommCareCase, CommCareCase-Deleted, CommCareCase-Deleted-Deleted, CommCareCase-deleted, CommCareImage, CommCareMultimedia, CommCareVideo, HQSubmission, InvoicePdf, SavedBasicExport, SubmissionErrorLog, XFormArchived, XFormDeprecated, XFormDuplicate, XFormError, XFormInstance, XFormInstance-Deleted…
+   Migration log: /opt/data/blobmeta-migration-logs/migrate_metadata-blob-migration-20181024T205839Z.txt.0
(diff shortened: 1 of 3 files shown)


millerdev:
Uh, inferring from the test failure it appears that the change log needs to be encoded as escaped markdown in a yaml file? Please tell me there's a more natural way to do this. Can I generate that yaml file from the markdown I wrote? @dannyroberts

millerdev:
@dannyroberts sorry for the outburst of indignity, and thank you for writing tests to verify that change log entries have been composed properly.

I took the liberty to allow change log details and update steps to be composed directly in the markdown file (see https://github.com/dimagi/commcare-cloud/pull/2521/commits/142bdd866c007f2fff7e5ab3bc743ebb9d5f51ee). If we like this approach we can remove all `details` and `update_steps` key/value pairs from the yaml files and simplify the process a bit there. It feels more natural to me (not to mention easier to preview) to write markdown in a markdown file rather than embedding it in a yaml file that requires a build step to generate the final output.

changelog/0008-blob-metadata-part-1.yml
```diff
@@ -0,0 +1,12 @@
+title: Blob Metadata Migration - part 1
+key: blob-metadata-part-1
+date: 2018-09-24
+
+optional_per_env: no
+
+min_commcare_version: d7a5ee92d32f719e3e19c5841fdc7cecd2072985
+max_commcare_version:
```
millerdev: Hmm, it looks like these min/max revisions in the change log are just used to create notes in the change log to be observed by people doing upgrades. I had previously thought they were used to enforce a specific upgrade path through some automated process. With what I've done in this PR these fields are no longer referenced anywhere. Working on some ideas here.

dannyroberts:
Hey @millerdev, sorry for the couple days' silence here. I've been sick and recovering and slow to catch up.

I'm gathering that you had a negative experience with the system we've set up for logging changes, and dislike that the documentation is generated from a yaml file which itself uses some markdown formatting inside values, and you're wondering why the system works that way. I'm also hearing that you hope you can change the pattern rather than conform to it.

I'm going to try and provide context and motivation for the current system, and then maybe we can work together on what happens next. Looking back over my notes from the time, I found a line I wrote about this, "Autogenerate all changelog pages in the docs from yml files. This allows us to write other workflow automation off of the yml files in the future", in reference to this PR: https://github.com/dimagi/commcare-cloud/pull/2075. I think that sums up my thinking at the time, and was the output of a series of conversations with @snopoke about how we can set this up to back a "migration" system that guides `commcare-cloud` users through applying required (or optional) maintenance. (Unfortunately, I couldn't find the text of that conversation or where we had it.) We envisioned a number of the same things you noted like "I had previously thought they were used to enforce a specific upgrade path through some automated  process" (the tool would presumably do this) and "Do I need to do something else to record that it's been done for all envs" (that would presumably be tracked per-environment by the tool).

My opinion is that a system in which some of the changelog markdown files are built from the yaml files and other are their own primary source is worse than either a system where they are all built from yaml or a system where they are primary source. Even without the imagined tool in the immediate future, there are a few things that I like about generating from yaml, perhaps the most prominent of which is the ability to enforce a consistent format that is tested in the build.

Let me know what you'd like to happen next.

dannyroberts:
I may have misrepresented your proposed change. I don't think I fully understand it just looking at the code diff. Could you give me a rundown of what the new system would look like?

millerdev:
Hey, @dannyroberts I hope you're feeling better, and no worries about the delay.

I'm hoping you agree that having markdown inside a yaml file makes it rather unnatural to edit and review said markdown? I've never seen an open source project do that before, and it makes a less than ideal workflow at the point where a new change log entry is being composed. If we don't agree on that then maybe I just better shut up and conform to the system that's in place. I've certainly learned about how parts of it work through this PR. However, figuring it out has been a process.

Here's where I'm at currently:

Maybe it would make sense to build a tool to be run by a developer when creating a new change log entry? It would generate markdown that could be easily edited and reviewed _in addition to_ a subset of the existing yaml file, which could be parsed by a (currently non-existent) tool to do more automation around enforcing upgrade paths, etc. The developer tool used to create a new change log entry might work something like this:

```
$ manage-commcare-cloud add-change-log-entry \
  "New Entry Title" --key=KEY --date=DATE --optional-per-env \
  --min-rev=COMMIT --max-rev=COMMIT

Adding new change log entry files...
  changelog/NNNN-new-entry-title.yml
  docs/changelog/NNNN-new-entry-title.md

Edit the generated files to add any missing details and/or upgrade steps.
Then run `make` to update the change log index. Tests will fail if you forget this step.
```

- `--key=KEY` auto-generated from (required) title argument if not given.
- `--date=DATE` today's date unless overridden.
- `--optional-per-env` defaults to NO unless this option is provided.
- `--min-rev=COMMIT` `--max-rev=COMMIT` optional min/max git commit ref.

This tool would generate `changelog/NNNN-entry-name.yml` (used to generate the index) and partially filled out `docs/changelog/NNNN-entry-name.md` (all the top matter plus blank sections for details and upgrade steps). The generated files could then be edited to fill out any missing details, upgrade steps, etc.

Existing tests could be improved to ensure that the yaml file stays in sync with the markdown.

PR #2522, opened 2018-12-16
Add small VM for shareddir0-production

dannyroberts:


environments/production/aws-resources.yml
```diff
@@ -34,14 +34,11 @@ proxy0-production: 10.202.20.138
...
 rabbit0-production: 10.202.42.98
 redis0-production: 10.202.41.8
+shareddir0-production: 10.202.40.191
 vpn-production: 10.202.20.194
 vpn-production.public_ip: 18.235.223.2
 web0-production: 10.202.10.233
 web1-production: 10.202.10.84
-web10-production: 10.202.11.188
-web11-production: 10.202.11.153
-web2-production: 10.202.10.238
```
czue: Is this intentional? It doesn't seem to be reflected well in the PR name / commit history

dannyroberts: Oops, yes it is. This file gets updated to reflect actual aws resources when you run `cchq production aws-fill-inventory`, and these are machines that were previously deleted without that command having been run to refresh this file.

PR #2530, opened 2018-12-26
Shallow clone repo when deploying

proteusvacuum:
Not sure whether we need our whole git history in every release. Currently, each release's `.git` directory consumes over 800MB. 

Open for review label because I want to get @dannyroberts opinion on this.

### DIFF
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -183,7 +183,7 @@ def _update_code_from_previous_release():
             sudo('git remote set-url origin {}'.format(env.code_repo))
     else:
         with cd(env.code_root):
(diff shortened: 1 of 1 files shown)


dannyroberts:
I kind of thought we did this, but clearly we don't. This will be great if indeed it doesn't cause any issues. Didn't realize you were waiting on my input. I think as long as we're able to test two sequential staging deploys using this (where the second doesn't just redeploy the same code), then this looks great to me.

Just for context, I believe the way this works is it clones from the latest deploy's .git directory, and then it pulls the code updates from origin (github). I'm not sure if something from that second process itself pulls down way more than needed, but if it does, I imagine it'll be a lot slower to pull it from github than to clone it from the local filesystem. So one thing to test for is that after a full deploy, the new code dirs are indeed much smaller.

PR #2532, opened 2019-01-02
Make cert new cron job reload nginx after renewing

dannyroberts:
using the built-in hook. Will have to roll this out to every env that
uses letsencrypt before it'll take effect.

Noticed from the discussion we currently do this manually, which adds to the confusion around certs.

@NitigyaS see any problems with this?


dannyroberts:
Okay! This finally ran smoothly on staging and production

proteusvacuum:
As mentioned in the retro, but do you think you can update the documentation on confluence about this?

dannyroberts:
Thanks for the push, Farid. I just updated these two pages to
- https://confluence.dimagi.com/display/internal/SSL+Certificates
- https://confluence.dimagi.com/display/internal/Linking+SSL+certificate
to reflect the new situation.

PR #2533, opened 2019-01-02
Try to fix "serial8250: too much work for irq4"

dannyroberts:
Following AWS Suport's recommendation in
https://console.aws.amazon.com/support/v1?#/case/?displayId=5612595231&language=en.

I can't say I deeply understand what this does, but I don't think it
affects what ends up in /var/syslog.

### DIFF
--- a/src/commcare_cloud/ansible/roles/common/handlers/main.yml
+++ b/src/commcare_cloud/ansible/roles/common/handlers/main.yml
@@ -1,0 +1,3 @@
+- name: restart rsyslog
+  become: yes
+  service: name=rsyslog state=restarted
--- a/src/commcare_cloud/ansible/roles/common/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/common/tasks/main.yml
@@ -144,3 +144,7 @@
   tags:
     - mount
     - after-reboot
+
+- import_tasks: syslog.yml
+  tags:
+    - syslog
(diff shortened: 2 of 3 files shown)


snopoke:
Seeing this in the output of rsyslog after updating the config with this change:
```
Nov 14 08:40:40 MUMGCCWCDPRDWV004 rsyslogd[49982]: command 'klogConsoleLogLevel' is currently not permitted - did you already set it via a RainerScript command (v6+ config)? [v8.32.0 try http://www.rsyslog.com/e/2222 ]
```

It's tagged as `obsolete legacy` in https://www.rsyslog.com/doc/v8-stable/configuration/modules/imklog.html#consoleloglevel. Perhaps there's a better way to do this now?

dannyroberts:
Is there any relation to your PR made at the same time https://github.com/dimagi/commcare-cloud/pull/3396? I see that PR touches this setup, but it seems to only move where it is, not change it?

I'm also not really sure what to do with this error message. It seems like the directive is just being ignored? If that's the case, then maybe we should just remove this config altogether without trying to replace it with anything

snopoke:
@dannyroberts are you saying the log messages this was intended to stop are still coming through?

dannyroberts:
Nope, I wasn't trying to add any new information here, just trying to understand the point you were making and whether there's an action we should take.

If I recall correctly, this PR back in Jan was intended to fix an issue in which machines would become unresponsive. We haven't seen this issue for some time (except for when a machine runs out of memory and has its various network agents killed, which is unrelated).

PR #2539, opened 2019-01-07
Removed celery5 from roataion for resizing

NitigyaS:
Since TCL has not yer resize the servers,
We will take it out of the rotation and will add it back once they have resized it. 
According to TCL, it should take close to 2 hours.


environments/icds-cas/inventory/commcare.csv
```diff
@@ -3,7 +3,7 @@ celery0,100.71.188.5,celery,,,MUMGCCWCDPRDC000
 celery1,100.71.188.11,celery,,,MUMGCCWCDPRDC001
 celery3,100.71.188.13,celery,,,MUMGCCWCDPRDC003
 celery4,100.71.188.12,celery,,,MUMGCCWCDPRDC004
-celery5,100.71.188.7,celery,,,MUMGCCWCDPRDC005
+#celery5,100.71.188.7,celery,,,MUMGCCWCDPRDC005
```
snopoke: I don't think you need to remove it from here

PR #2549, opened 2019-01-09
update redis max memory

snopoke:
I noticed this hasn't been done since we resized the VMs.

There may be other services that also need config adjustments.


snopoke:
Closing this since there already appear to be some possible memory issues.

PR #2550, opened 2019-01-09
Make --control flag run faster

dannyroberts:
by doing less redundant setup on the machine

### DIFF
--- a/src/commcare_cloud/commcare_cloud.py
+++ b/src/commcare_cloud/commcare_cloud.py
@@ -88,7 +88,7 @@ def run_on_control_instead(args, sys_argv):
     branch = getattr(args, 'branch', 'master')
     cmd_parts = [
         executable, args.env_name, 'ssh', 'control', '-t',
-        'source ~/init-ansible && git fetch && git checkout {branch} '
+        'source ~/init-ansible && git fetch --prune && git checkout {branch} '
         '&& git reset --hard origin/{branch} && source ~/init-ansible && {cchq} {cchq_args}'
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commcare_cloud.py
```diff
@@ -88,7 +88,8 @@ def run_on_control_instead(args, sys_argv):
     branch = getattr(args, 'branch', 'master')
     cmd_parts = [
         executable, args.env_name, 'ssh', 'control', '-t',
-        'source ~/init-ansible && git checkout master && control/update_code.sh && git checkout {branch} && git reset --hard origin/{branch} && source ~/init-ansible && {cchq} {cchq_args}'
+        'source ~/init-ansible && git fetch && git checkout {branch} '
```
jmtroth0: It might be worthwhile to add the `--prune` flag on fetch

it looks like this no longer checks for changes on the current branch, but that seems fine since it's on control and it's hard reseting?

dannyroberts: Yeah if the branch is master, than this will update the master branch, etc. That's basically the optimization, skipping updating an irrelevant branch.

What's the scenario you're thinking `--prune` would help with? (I might just be ignorant of the pattern.)

jmtroth0: oh, basically just having too many git objects around. it's probably not a huge concern, but afaik old merged branches would just hang around forever as is.

dannyroberts: okay I'll add that. thanks!

PR #2556, opened 2019-01-10
Add Telangana

mkangia:


mkangia:
@millerdev these are locations/web pages added for sharing icds-cas apps CCZs and other installation files with different states. So this is where we host the latest released versions of the app for different states and they then use this to install app for fresh installs.
Each of these links has a different set of credentials to access the page provided to the the states.
Happy to share more context if you need or if you have any questions.

PR #2558, opened 2019-01-11
Add RAM on pgproxy0

millerdev:
@dannyroberts

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -226,7 +226,7 @@ servers:
     group: "kafka"
 
   - server_name: "pgproxy0-production"
-    server_instance_type: t3.medium
+    server_instance_type: t3.large
     network_tier: "db-private"
     az: "a"
     volume_size: 80


millerdev:
Do I need to schedule downtime for this or can it be deployed safely without that?

dannyroberts:
Yeah resizing a machine unfortunately requires it stop and start, which takes a couple minutes. An alternative is to create another machine and replace all references to this one with references to that one, which of course doesn't work if there's data involved. While pgproxy maintains very little state compared to the actual partition dbs, it still seems like more trouble and risk than its worth to go that route, versus scheduling 2 minutes of downtime on during a low-load period.

emord:
> While pgproxy maintains very little state compared to the actual partition dbs, it still seems like more trouble and risk than its worth to go that route, versus scheduling 2 minutes of downtime on during a low-load period

@dannyroberts Do you think its worth doing the no downtime deploy of this as practice of a useful workflow? 

from my understanding the only commcare "data" that plproxy contains are functions that are setup via configure_plproxy_cluster. is that correct?

dannyroberts:
@emord Yeah, I think I agree with your suggestion now that you say it. I see the value in that.

millerdev:
Did the no-downtime method: https://github.com/dimagi/commcare-cloud/pull/2585

I'm still getting familiar with new AWS automation, but my overall impression is that there were a lot of steps in the process. It was tedious and took a long time.

PR #2559, opened 2019-01-14
Softlayer couch migration

dannyroberts:
This is all applied already.

@snopoke couch-migrate command is pretty awesome! I noticed it assumed a downtime workflow, and but I was able to do it live by commenting out a few portions, specifically the line that stops couch around the migration and the thing that asserts that the files are all copied (including an md5 hash). This seemed to work great (couch will continue to replicate from wherever you are if the shards aren't up to date with each other).


dannyroberts:
@snopoke I had a minor question about the migration directory. There are automatically-generated "build" files; was the intention to commit those or not? Here I committed them, but that seems a bit much. If you agree I'll delete them and leave just the two-line plan.

src/commcare_cloud/commands/migrations/plays/assert_couch_files.yml
```diff
@@ -7,21 +7,26 @@
...
         fail_msg: "Shard file missing: {{ item.item }}"
       loop: "{{ shards.results }}"
+      when: inventory_hostname in files_by_node
 
     - name: assert views exist
       assert:
         that: "{{ item.stat.exists and item.stat.isdir }}"
         fail_msg: "View missing: {{ item.item }}"
       loop: "{{ views.results }}"
+      when: inventory_hostname in files_by_node
```
snopoke: is this because you moved all data to new nodes but the old nodes were still in the inventory?

src/commcare_cloud/commands/migrations/couchdb.py
```diff
@@ -335,8 +335,8 @@ def _run_migration(migration, ansible_context, check_mode):
     rsync_files_by_host = prepare_to_sync_files(migration, ansible_context)
 
     puts(colored.blue('Stop couch and reallocate shards'))
-    with stop_couch(migration.all_environments, ansible_context, check_mode):
-        execute_file_copy_scripts(migration.target_environment, list(rsync_files_by_host), check_mode)
+    # with stop_couch(migration.all_environments, ansible_context, check_mode):
```
snopoke: If we don't need this then should just remove completely

src/commcare_cloud/commands/migrations/plays/assert_couch_files.yml
dannyroberts: Yeah exactly.

src/commcare_cloud/commands/migrations/couchdb.py
dannyroberts: Yeah you're right, I didn't mean to commit this.

dannyroberts:
Sorry I wasn't clear; I meant I had locally commented out the sections that stopped couch and that asserted shard files are the same. (And then to make matters more confusing I by mistake committed part of that. I'll go back and clean that up.)

Oh for some reason I thought it was asserting not just that the files were present, but that they were the same. I got a failure but I must have misinterpreted it at the time. I will check back in to make sure the cluster is in the state I think it is, and in the meantime, I'll leave the file assertions.

PR #2560, opened 2019-01-14
Turn on termination protection for all ec2 instances

dannyroberts:
This just adds an extra layer of friction for deleting machines, so that it's less likely that the disaster scenario of by mistake deleting important machines could happen. To delete a machine, you'll have to first turn off termination protection on that machine, which you can do from the AWS console

### DIFF
--- a/src/commcare_cloud/terraform/modules/openvpn/main.tf
+++ b/src/commcare_cloud/terraform/modules/openvpn/main.tf
@@ -14,6 +14,8 @@ resource aws_instance "vpn_host" {
     reboot
     EOF
 
+  disable_api_termination = true
+
   root_block_device {
     volume_size           = 40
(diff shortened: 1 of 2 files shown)


dannyroberts:
@ctsims you had asked about this a while back, about how to prevent disaster scenarios that centralized control of creating and destroying infrastructure (i.e. terraform) opens us up to. This change makes it a lot harder to by mistake delete machines, and acts as a backstop to the human process of being careful to understand the diff of what we're applying.

dannyroberts:
Me too. I have to remind myself not to _depend_ on it though! We should still pretend like we have the ability to delete machines and use this as a backstop, because if we depend on this to block us from deleting machines then one day something will break this and it'll be a disaster.

PR #2574, opened 2019-01-19
Switch to using local couchdb2 for production commcarehq db!

dannyroberts:
This puts us entirely on local couchdb2 and entirely off cloudant.

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -47,86 +47,9 @@ run_websockets_wsgi: True
 
 couch_dbs:
   default:
-    host: commcarehq.cloudant.com
-    port: 443
+    host: "{{ groups.couchdb2_proxy.0 }}"
+    port: "{{ couchdb2_proxy_port }}"
     name: commcarehq
-    username: "{{ localsettings_private.COUCH_USERNAME }}"
-    password: "{{ localsettings_private.COUCH_PASSWORD }}"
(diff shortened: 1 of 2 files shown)


dannyroberts:
It's hard to say, but I think so. The massive reindex is certainly negatively affecting performance this weekend, but that won't be around at nearly this level in steady state of course; and currently all writes to cloudant are being replicated to the local db. So I'd say based on these two things it can almost certainly handle the indexer load, and the write load, so the only totally untested thing is the read load. Overall though, I think we're on about the same level of resources that we have on cloudant. The cluster in AWS is at 32GB RAM / 16 vCPU / 7.5 TB disk / 15000 IOPS to data disk, times 3 nodes, with a full copy of the db on each node. Perhaps the easiest change to make if the flip does cause massive issues is to reduce the number of replicas from 3 to 2, which should reduce write load by a third.

millerdev:
🎉

snopoke:
Thanks @dannyroberts. Fine to proceed from me, just wanted to check that we'd at least thought about it and were reasonably confident in proceeding.

dannyroberts:
Just noting that this is delayed for now as it's going to take some creativity and time before we're able to reindex the commcarehq db

dannyroberts:
OK we're ready to go!

PR #2575, opened 2019-01-19
Increase data volume on couchdb2 machines to 7.5 TB

dannyroberts:
disks were over 80% full at 6TB, so I further increased it.
These changes are already rolled out.

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -133,7 +133,7 @@ servers:
     az: "a"
     volume_size: 4000
     block_device:
-      volume_size: 5000
+      volume_size: 7500
     group: "couchdb2"
   - server_name: "couch1-production"
     server_instance_type: c5.4xlarge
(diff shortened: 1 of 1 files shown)


millerdev:
That's amazing that you can just resize it so easily without data loss. Do we follow the recommendation to "Before modifying a volume that contains valuable data, it is a best practice to create a snapshot of the volume in case you need to roll back your changes." (from [Modifying the Size, IOPS, or Type of an EBS Volume on Linux](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-modify-volume.html))?

dannyroberts:
Oh and there are some manual steps to get the OS to recognize the volume increase: https://confluence.dimagi.com/display/internal/How+to+modify+volume+size+on+AWS. So it does require a node restart and some downtime while you run these steps to resize the disk, but it's pretty quick.

PR #2579, opened 2019-01-21
Temporarily increase couch0-production for view indexing

dannyroberts:
I set commcarehq db to only store shards on couch0, and I'll index it there. This is following @emord's suggestion in #interrupt-team

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -128,7 +128,7 @@ servers:
     volume_size: 80
     group: "couchdb2_proxy"
   - server_name: "couch0-production"
-    server_instance_type: c5.4xlarge
+    server_instance_type: c5.18xlarge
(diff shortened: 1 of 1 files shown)


emord:
> I set commcarehq db to only store shards on couch0, and I'll index it there. 

Why not spread out one copy of the shards?

dannyroberts:
I don't know—what does that get us? I was thinking by putting on a single node, there's only one node whose resources we need to increase, and there are always two nodes unaffected by any reindexing, still available to handle request.

emord:
It gets us better distribution of load. Especially for reindexes as they use lots of disk i/o which doesn't scale with CPU/RAM

dannyroberts:
Merging this to get master in line with real state. Offlining with emord about what to do next

PR #2581, opened 2019-01-22
Change couchdb2 nginx site to be in line with production

dannyroberts:
I must have made these changes before while debugging replication, but never committed them.
This reflects what is currently on prod.

Do you see any issues with these being the new defaults? If so I can split them out to be configurable by env.

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/couchdb2.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/couchdb2.yml
@@ -7,19 +7,24 @@ nginx_sites:
       port: "{{ couchdb2_port }}"
    file_name: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
    listen: "'{{ couchdb2_proxy_port }}'"
+   proxy_http_version: "1.1"
    client_max_body_size: "{{ couchdb2_client_max_body_size }}"
+
    upstream_port: "{{ couchdb2_port }}"
    proxy_set_headers:
    - "Host $http_host"
    - "X-Real-IP $remote_addr"
    - "X-Forwarded-For $proxy_add_x_forwarded_for"
+   - 'Connection ""'
    locations:
      - name: /
        balancer: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
        proxy_redirect: ' off'
-       client_body_buffer_size: ' 512k'
+       proxy_send_timeout: 900s
+       client_body_buffer_size: ' 50m'
        proxy_max_temp_file_size: ' 0'
-       proxy_read_timeout: ' 65s'
+       proxy_read_timeout: ' 900s'
+       proxy_connect_timeout: '120s'
      - name: "~ ^/(.*)/_changes"
        balancer: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
        proxy_redirect: ' off'


src/commcare_cloud/ansible/roles/nginx/vars/couchdb2.yml
```diff
@@ -7,19 +7,24 @@ nginx_sites:
...
    listen: "'{{ couchdb2_proxy_port }}'"
+   proxy_http_version: "1.1"
    client_max_body_size: "{{ couchdb2_client_max_body_size }}"
+
    upstream_port: "{{ couchdb2_port }}"
    proxy_set_headers:
    - "Host $http_host"
    - "X-Real-IP $remote_addr"
    - "X-Forwarded-For $proxy_add_x_forwarded_for"
+   - 'Connection ""'
```
emord: What does this mean? Do you know why its necessary?

src/commcare_cloud/ansible/roles/nginx/vars/couchdb2.yml
```diff
@@ -7,19 +7,24 @@ nginx_sites:
       port: "{{ couchdb2_port }}"
    file_name: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
    listen: "'{{ couchdb2_proxy_port }}'"
+   proxy_http_version: "1.1"
```
emord: Do you remember why HTTP2 isn't ok?

src/commcare_cloud/ansible/roles/nginx/vars/couchdb2.yml
```diff
@@ -7,19 +7,24 @@ nginx_sites:
...
    locations:
      - name: /
        balancer: "{{ deploy_env }}_couchdb_{{ couchdb2_proxy_port }}"
        proxy_redirect: ' off'
-       client_body_buffer_size: ' 512k'
+       proxy_send_timeout: 900s
+       client_body_buffer_size: ' 50m'
        proxy_max_temp_file_size: ' 0'
-       proxy_read_timeout: ' 65s'
+       proxy_read_timeout: ' 900s'
```
emord: This one does concern me a bit. This sounds like it will keep a connection between HQ and couch open for a very long time when couch is slow. I'm more in the camp of failing fast (and I wouldn't even consider the original 65s fast). It seems like a really bad reduction in a theoretical couch performance SLA.

More for the future, but it seems like this may be a good place to codify an SLA. Like if couch isn't responding with 10 seconds, we consider it down and urgent to fix.

dannyroberts:
You know what, I think I remember now when I was making these changes, and I think they ended up not mattering. I'm going to close this instead and blow them away next time I deploy the couchdb2 proxy.

To your question we're not on haproxy on prod because that wasn't the default when I was setting it up (and maybe still isn't)? I was vaguely aware of the haproxy setup existing but I don't have a conception of when we're supposed to use which.

emord:
@snopoke can confirm, but I believe we want hqproxy everywhere. it's only on ICDS now, because we didn't have haproxy before the tech summet

snopoke:
I deployed HA proxy to ICDS to get better metrics and since it was the recommended proxy in the couch dos (also the proxy used by Cloudant): https://app.datadoghq.com/screen/465025/haproxy---overview?page=0&is_auto=false&from_ts=1548665280000&to_ts=1548668880000&live=true&tpl_var_scope=icds

I don't see an issue with using nginx though would be good to be consistent everywhere. Also haproxy does have much better metrics.

The reason I didn't roll this out further was that I was interested in looking at a few other options which I found out about later (such as [Traefik](https://docs.traefik.io/)) but haven't had a chance to do so. Either way I think there's advantage to using HAProxy and unlikely we're going to change to something else anytime soon (due to resource constraints).

snopoke:
I deployed HA proxy to ICDS to get better metrics and since it was the recommended proxy in the couch dos (also the proxy used by Cloudant): https://app.datadoghq.com/screen/465025/haproxy---overview?page=0&is_auto=false&from_ts=1548665280000&to_ts=1548668880000&live=true&tpl_var_scope=icds

I don't see an issue with using nginx though would be good to be consistent everywhere. Also haproxy does have much better metrics.

The reason I didn't roll this out further was that I was interested in looking at a few other options which I found out about later (such as [Traefik](https://docs.traefik.io/)) but haven't had a chance to do so. Either way I think there's advantage to using HAProxy and unlikely we're going to change to something else anytime soon (due to resource constraints).

PR #2586, opened 2019-01-23
Make migrate-couchdb more flexible

dannyroberts:
Along with this https://github.com/dimagi/couchdb-cluster-admin/pull/23, I made these changes while moving the shards for the main db to their own nodes.


src/commcare_cloud/commands/migrations/couchdb.py
```diff
@@ -53,11 +53,22 @@ class MigrateCouchdb(CommandBase):
             - commit: update database docs with new shard allocation
             - clean: remove shard files from hosts where they aren't needed
         """),
+        Argument('--no-stop', action='store_true', help="""
+            When used with migrate, operate on live couchdb cluster without stopping nodes
+        """),
+        Argument('--skip-file-check', action='store_true', help="""
+            When used with commit, do not check for the existence of expected shard files.
+            Only use this if you've verified the files are in place by other means.
```
snopoke: how would you do that?

src/commcare_cloud/commands/migrations/couchdb.py
```diff
@@ -53,11 +53,22 @@ class MigrateCouchdb(CommandBase):
             - commit: update database docs with new shard allocation
             - clean: remove shard files from hosts where they aren't needed
         """),
+        Argument('--no-stop', action='store_true', help="""
+            When used with migrate, operate on live couchdb cluster without stopping nodes
```
snopoke: Is there any danger to doing this? Like will there be any data loss if the cluster is in use during the migration?

src/commcare_cloud/commands/migrations/couchdb.py
dannyroberts: I used `watch -n10 -d cchq <env> run-shell-command 'ls -al /opt/data/couchdb2/{,.}shards/*/*'` or something like that to watch the rsyncs in progress, and it was a simple enough change that I could tell all the files were there

src/commcare_cloud/commands/migrations/couchdb.py
dannyroberts: I think if you're moving shards so that the set of new nodes and old nodes is disjoint then yes there's danger of losing data. But if you are moving it so that for every shard there's a place that it was both before and after, then whatever bit is missed on the new shards will get replicated over from the "stable pillar" shard.

Or, as in my case, if you're doing this on a database that isn't getting written to. I had stopped the replication, and there was no live process writing to this db, so it was just a matter of transferring static data.

snopoke: I think it's worth making a note of some of these points in the docs somewhere.

src/commcare_cloud/commands/migrations/couchdb.py
dannyroberts: It for some reason was taking much longer than made sense and I couldn't get it to complete. So it was that I wanted to skip it, not exactly that it "failed". Not sure if it had to do with one of the dozens of stats that ansible stats collects being really slow or something.

I can see the argument what we ought to fix whatever that issue is rather than making it easy to skip. Let me know what you think

src/commcare_cloud/commands/migrations/couchdb.py
dannyroberts: :+1:

src/commcare_cloud/commands/migrations/couchdb.py
dannyroberts: OK I optimized the file check for speed and readability and removed this option. Thanks for talking me through it, I'm much happier with this outcome

dannyroberts:
OK @snopoke I did another pass on this over the weekend while I was using it to finalize the cluster before the last piece of the cloudant migration. I think this is all set now, let me know what you think of the changes

PR #2590, opened 2019-01-24
Activate UCR queues on eCHIS

sravfeyn:


environments/echis/app-processes.yml
```diff
@@ -25,7 +24,7 @@ celery_processes:
       optimize: True
     submission_reprocessing_queue:
       concurrency: 1
-    reminder_case_update_queue:
+    reminder_case_update_queue,ucr_indicator_queue,ucr_queue:
```
emord: These are not a good fit for `gevent` as they are usually more constrained by CPU and not I/O. `gevent` queues are used for tasks that are bound by I/O

sravfeyn: Thanks! Moved it

PR #2594, opened 2019-01-25
Added cd into code directory when running django-manage

orangejenny:
@calellowitz 

fyi @mkangia This is why I was seeing `ReportConfigurationNotFoundError` when working in shell: the paths to the static report json configs are relative, so when the shell was running from `/home/my_username` it would error out.

### DIFF
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -208,7 +208,7 @@ class DjangoManage(CommandBase):
             code_dir = '/home/{cchq_user}/www/{deploy_env}/current'.format(
                 cchq_user=cchq_user, deploy_env=deploy_env)
         remote_command = (
-            '{code_dir}/python_env/bin/python {code_dir}/manage.py {args}'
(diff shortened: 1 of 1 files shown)


orangejenny:
or code buddy @millerdev

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -208,7 +208,7 @@ def run(self, args, manage_args):
             code_dir = '/home/{cchq_user}/www/{deploy_env}/current'.format(
                 cchq_user=cchq_user, deploy_env=deploy_env)
         remote_command = (
-            '{code_dir}/python_env/bin/python {code_dir}/manage.py {args}'
+            'bash -c "cd {code_dir}; {code_dir}/python_env/bin/python {code_dir}/manage.py {args}"'
```
mkangia: Do we need to mention `code_dir` for python and manage.py if we have already used `cd`?

orangejenny: good point - 7643f9381

PR #2605, opened 2019-01-31
Adding Celery2,Celery5 and Celery12  servers to icds cluster

sanjay2916:
Adding Celery2,Celery5 and Celery12  servers to icds cluster


sanjay2916:
celery2,5,and 112 added to cluster

PR #2610, opened 2019-02-04
don't monitor heartbeat on staging

snopoke:
This should get rid of the events that get triggered each time it makes a request to the URL


snopoke:
already rolled out:

image

PR #2612, opened 2019-02-04
Begin firefighting documentation

emord:
@esoergel @dannyroberts Following up from the email thread. I started converting the firefighting docs to this reo


docs/firefighting/index.md
```diff
@@ -0,0 +1,108 @@
...
+$ ./manage.py check_services
+# or go to
+https://[commcarehq.environment.path]/hq/admin/system/check_services
+```
+
+## Check that change feeds are still processing
+
+You can use this graph on [datadog](https://app.datadoghq.com/dashboard/ewu-jyr-udt/change-feeds?to_ts=1549314000000&is_auto=false&live=true&from_ts=1549227600000&tile_size=m&page=0&fullscreen_widget=185100827)
+
+# Service Information
```
emord: The table in here sucked, so I rearranged this

docs/firefighting/index.md
```diff
@@ -0,0 +1,108 @@
+This page summarizes how to do various things related to HQ.
+For a more user-friendly guide check out [Cory's brown bag on the topic](http://prezi.com/wedwm-dgqto7/firefighting-hq/).
+
+# HQ Architecture and Machines
+
+![](./hq_architecture.jpg)
```
emord: Not sure if this will actually work. I don't have ruby on my local machine, os haven't been able to test it out

docs/firefighting/index.md
```diff
@@ -0,0 +1,108 @@
...
+To stop all supervisor processes use:
+
+```bash
+$ cchq <env> fab supervisorctl:"stop all"
+```
+
+# Notes
+
+All Datadog links will be specific and private to Dimagi employees.
+If datadog releases a feature to share dashboard configurations, we will happily share configurations in this repository.
```
esoergel: I'd move this to the top

docs/firefighting/index.md
```diff
@@ -0,0 +1,108 @@
...
+
+![](./hq_architecture.jpg)
+
+# High-level System Monitoring and Alerts
+
+[HQ Vitals](https://app.datadoghq.com/dashboard/g9s-pw6-tpg/hq-vitals?to_ts=1549314000000&is_auto=false&from_ts=1549227600000&live=true&tile_size=m&page=0) - Various graphs on datadog
+
+[Datadog Alerts](https://app.datadoghq.com/monitors/manage?q=status%3A(alert%20OR%20warn%20OR%20"no%20data")) - these are also often reported on #hq-ops or #hq-ops-priority on slack
+
+https://www.commcarehq.org/hq/admin/system/ - catchall system info, contains deploy history, pillowtop info, and a bunch of other stuff
```
esoergel: HQ links are listed as variables below:
```https://[commcarehq.environment.path]/hq/admin/system/check_services```
The link to prod is ever-so-slightly more useful, but I don't really care.  Flagging in case you feel like making that consistent.

File: docs/firefighting/index.md (hunk 2 above)
Comment (millerdev) [reply docs/firefighting/index.md]: This should work—I've done something similar [here](https://github.com/millerdev/lms-ultralight/blob/master/README.md). Maybe try this:

```suggestion
![HQ Architecture and Machines](hq_architecture.jpg?raw=true "HQ Architecture and Machines")
```

emord: It turns out that what I did does work if I use the right image extension https://github.com/dimagi/commcare-cloud/blob/5f8e72de44046e927d5c0d8382ed3bb44045c8ef/docs/firefighting/index.md

emord:
> there's the possibility of missing new changes

I'm watching hte page, so i get emails whenever its updated

dannyroberts:
Looks like the build error will be fixed by merging master

emord:
Ok. I finished moving the firefighting HQ page here. I did not start on any other pages, and I did very little editing (basically kept out thing like CDN and cloudant and rackspace)

Once this is approved and merged I'll start deleting that confluence page (except for what will be internal) and look into moving some other pages over

docs/firefighting/index.md
```diff
@@ -205,9 +205,583 @@ It may happen, specially if the client crashes or has kernel oops, that a connec
...
++----------------------------------------------------------------------+------+-------+-----+-------+
+
+Key: (C) = Claimant; availability marked with '!' is unexpected
+```
+
+## Riak is Out of Memory
+
+The most common problem we have had with Riak and Riak CS is running [out of memory](https://docs.basho.com/riak/kv/2.1.4/using/repair-recovery/failure-recovery/#out-of-memory). Most of the time this is non-actionable; nodes will crash and automatically restart and operation usually returns to normal without intervention. If a node is consistently running out of memory then it is likely necessary to either allocate more memory (usually done by creating a new node with more memory) or by allocating more swap space (can affect performance).
+
+##What do do when one or more Riak nodes has a corrupt database
```
millerdev: needs space between `##` and `What`

docs/firefighting/index.md
```diff
@@ -0,0 +1,787 @@
...
+
+[pillow](https://app.datadoghq.com/dash/256236/change-feeds-pillows)
+
+[celery/celerybeat](https://app.datadoghq.com/dash/141098/celery-overview)
+
+[elasticsearch](https://app.datadoghq.com/screen/127236/es-overview)
+
+[kafka](https://app.datadoghq.com/screen/516865/kafka---overview-cloned)
+
+[riak/riak-cs](https://app.datadoghq.com/dash/175518/riak-cs-system-resources)
```
millerdev: We should rename this to "Blob DB Dashboard"

I updated it so graphs relevant to all blob db backends are now at the top, and riak specific graphs are below. Also updated the name of the dashboard. The link can be changed to https://app.datadoghq.com/dashboard/753-35q-fwt/blob-db-dashboard (although the old link still works too).

I will do this in a separate PR once this is merged.

docs/firefighting/index.md
```diff
@@ -0,0 +1,787 @@
...
+ansible@hqriak18:~$ sudo service riak start
+```
+
+Sometimes node data can become corrupt in a way that prevents the riak service from starting. In this case the node will begin to restart, but will crash while reading indexes from the disk into RAM. One example error we have seen in the past:
+
+In /var/log/riak/console.log
+```
+2016-09-28  00:02:22.350 [error] <0.638.0>@riak_kv_vnode:init:497 Failed to  start riak_cs_kv_multi_backend backend for index  1255977969581244695331291653115555720016817029120 error:  [{riak_kv_eleveldb_backend,{db_open,"Corruption: truncated record at end  of file"}}]
+``
+`
```
millerdev: Bad ending to code block. I didn't review very carefully beyond this point because the formatting is pretty messed up in the [markdown view](https://github.com/dimagi/commcare-cloud/blob/8fcd91a2958e90d2b3a6ccb514cb0d86fdab5bf9/docs/firefighting/index.md).

docs/firefighting/index.md
emord: https://github.com/dimagi/commcare-cloud/pull/2612/commits/537922107b3b71d1f31d18ebee44495be0ea3478

docs/firefighting/index.md
emord: https://github.com/dimagi/commcare-cloud/pull/2612/commits/8ea0e728fb497c29ae9ddb09ae37b41bdf8e0771

docs/firefighting/index.md
emord: Thanks for finding. I think i tracked down some other issues as well

https://github.com/dimagi/commcare-cloud/pull/2612/commits/db92b6a927fac10ced3592a0a3447f34f3ad7861
https://github.com/dimagi/commcare-cloud/pull/2612/commits/30863739a14da510a4a423d5407e42461509ed99
https://github.com/dimagi/commcare-cloud/pull/2612/commits/3f1ede6db1185481c527597ba3a51801f7f491fd

docs/firefighting/index.md
```diff
@@ -0,0 +1,788 @@
...
+User-Agent: curl/7.50.1
+Accept: */*
+
+esoergel@hqdb0:~$
+```
+
+# Tips and Tricks
+Never run that painful sequence of sudo -u cchq bash, entering the venv, cd'ing to the directory, etc., again just to run a management command. Instead, just run e.g.:
+
+`sudo -u cchq bash -c 'cd /home/cchq/www/production/current && /home/cchq/www/production/current/python_env/bin/python manage.py shell'i`
```
millerdev: nit: trailing `i` looks like a mistake.

Wonder if it would be useful to add `cchq <env> django-manage shell` in here since that's essentially equivalent (except it's always executed on the `django_manage` machine)?

esoergel: I'd be in favor of adding that (though that sounds like a separate change, out of scope of this transfer from confluence).  I do find this snippet very useful though, such as when I'm already working on a specific machine, so I'd want that as an addition, not a replacement.

docs/firefighting/index.md
emord: Didn't want to make any changes to this or establish conventions in this PR

docs/firefighting/index.md
emord: There were a lot of these sorts of changes I noticed, but wanted to keep it small as possible. I only changed to  the cchq commands for the big services table because that was easier than transferring that table to markdown.

PR #2614, opened 2019-02-05
Script to check backup status of s3 upload

NitigyaS:
Script to check backup upload status to S3

### DIFF
--- a/src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
+++ b/src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
@@ -3,6 +3,8 @@ import argparse
 import boto3
 import datetime
 
+#input environement and service from users
+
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/python
```
snopoke: This could use a comment explaining how it works. It's not clear why it's showing an alert if the key is older than 2 days. I'm assuming the same key get's overwritten each time so an old key indicates an out of date backup?

NitigyaS: it checks the last modified date, ALERT means that backup has not been uploaded for 2 days.

PR #2615, opened 2019-02-05
Ns/add new image

NitigyaS:


src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
```diff
@@ -55,8 +55,12 @@ module "server__{{ server.server_name }}" {
   secondary_volume_size = {{ server.block_device.volume_size|default(0)|tojson }}
   secondary_volume_type = {{ server.block_device.volume_type|default("")|tojson }}
 
+{% if server.group == 'celery:bionic' %}
```
NitigyaS: Not really satisfied with this. I tried adding an `image` section in the `terraform.yml` but I think it has some fixed syntax. So far it is fulfilling the short term purpose. @dannyroberts do you have any suggestions.

dannyroberts: I'm fine with this for now if it's the easiest thing. It's clearly temporary so whatever's the easiest for you is fine.

Guessing you may eventually want to extend this to be `{% if ':bionic' in server.group %}` so it works with other types of machines.

dannyroberts:
The test failure is real here, you have to run `cchq staging aws-fill-inventory`

PR #2617, opened 2019-02-05
Expand existing nodes for final couchdb cluster

dannyroberts:
Final couchdb cluster will be 8 x (c5.2xlarge, 5.3TB disk). After migration, will terminate couch[0-2].

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -154,68 +154,68 @@ servers:
 
   # couch main-db cluster: 8 nodes, one per shard
   - server_name: "couch3-production"
-    server_instance_type: c5.xlarge
+    server_instance_type: c5.2xlarge
     network_tier: "db-private"
     az: "a"
     volume_size: 80
     block_device:
-      volume_size: 4200
+      volume_size: 5300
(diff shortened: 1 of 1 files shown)


environments/production/terraform.yml
```diff
@@ -154,68 +154,68 @@ servers:
...
   # couch main-db cluster: 8 nodes, one per shard
   - server_name: "couch3-production"
-    server_instance_type: c5.xlarge
+    server_instance_type: c5.2xlarge
     network_tier: "db-private"
     az: "a"
     volume_size: 80
     block_device:
-      volume_size: 4200
+      volume_size: 5300
```
emord: Will this leave enough space for compaction of views and databases?

dannyroberts:
Yeah, it should. I calculated it off the current total disk used across all
nodes, divided by 8, divided by 78%. I believe there are old shard files
using up space, so I wanted to keep this learn for the first pass. If we
routinely go above 80%, it's pretty easy to increase disk space (but an
absolute pain to decrease disk space later if we overshoot). I also plan to
move one copy at a time so I'll have some advance warning if I undershot.

PR #2618, opened 2019-02-05
Laboratory child_health UCR

emord:
These are already written (and were present in the old NIC configurations). Deploying them so that we can see the differences


emord:
deployed

PR #2621, opened 2019-02-06
increase UCR pillow chunk size to 20 on pillow5

sravfeyn:
Thinking to roll this out and see if number of connections from pillow5 to ucr-main DB will reduce.

### DIFF
--- a/environments/icds-cas/app-processes.yml
+++ b/environments/icds-cas/app-processes.yml
@@ -298,6 +298,7 @@ pillows:
       start_process: 0
       num_processes: 9
       total_processes: 120
+      processor_chunk_size: 20
     CaseToElasticsearchPillow:
       start_process: 0
       num_processes: 9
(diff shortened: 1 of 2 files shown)


snopoke:
If possible it would be good to change a pillow where there's another host with the same config so that we can compare.

sravfeyn:
We will be able to compare with `pillow6` which has a default chunk size of 10.

emord:
> Thinking to roll this out and see if number of connections from pillow5 to ucr-main DB will reduce.

Would the number of connections drop? From what I was thinking was we would see a higher rate of changes being processed (assuming that bulking saves is more efficient)

sravfeyn:
That's true, if it is more efficient it would try to process more and open more connections, but unless it is efficient by a factor of 2 it won't open more connections than now. I think we can look at both the metrics (number of changes being processed and number of connections from pillow5) latter to see how it affects. 

We could also reduce `num_processes` if it becomes more efficient and opens more connections than desired.

sravfeyn:
This needs https://github.com/dimagi/commcare-hq/pull/23182

PR #2622, opened 2019-02-08
Remove "action required" marker from changelog index

dannyroberts:
and make the optional per env value in the changelog entry pages more explicit

Based on an email thread in which we were trying to consider what this looks like to third parties consuming release notes on the dimagi developer forum.


src/commcare_cloud/manage_commcare_cloud/changelog.md.j2
```diff
@@ -2,7 +2,12 @@
 
 **Date:** {{ changelog_entry.date }}
 
-**Optional per env:** {% if changelog_entry.optional_per_env %}Yes{% else %}No{% endif %}
+**Optional per env:** {% if not changelog_entry.optional_per_env -%}
+  _required on all environments_
+{%- else -%}
+  _only required on some environments_
```
snopoke: how would they know if they need this or not? Will it always be spelled out in the text?

dannyroberts: Yeah I think it has to be. For things not universally required, the only way they'll be able to figure it out is from the Context section, where we say who needs to apply it.

PR #2623, opened 2019-02-08
DB alias for ICDS tables to be moved

sravfeyn:
@snopoke @calellowitz @emord


environments/icds-cas/postgresql.yml
```diff
@@ -89,6 +90,11 @@ dbs:
       host: pgucr
       query_stats: True
       django_migrate: True  # the dashboard models are stored here
+    - django_alias: icds-ucr-non-dashboard
+      name: commcarehq_icds_aggregatedata
+      host: pgucr
+      query_stats: True
+      django_migrate: True  # the dashboard models are stored here
```
snopoke: this should be `False`

sravfeyn: Is that so, even when it will be a new separate DB? Aren't there any UCR app related migrations that will need to get run on this DB?

emord: This variable is for django migrations. The only reason we need django migrations for icds_reports is for the dashboard.

Any UCR migrations are handled automatically in SQLAlchemy

sravfeyn: Cool, thanks, I have updated it

sravfeyn:
@snopoke Is this good for deploy now?

PR #2626, opened 2019-02-08
Offboard jroth

gbova:
Remove jrtoh from public keys


proteusvacuum:
I think we also need to remove this file: https://github.com/dimagi/commcare-cloud/blob/7d0a2e092519a33f738f39280d69a6428ecd8b02/environments/_authorized_keys/jroth.pub

gbova:
`jroth.pub` deleted

proteusvacuum:
I'm assuming you also ran the required `commcare-cloud` commands on all envs?

gbova:
@proteusvacuum I thought those were to be run after this PR was merged (so it's next on my to-do list). Am I supposed to run those before merging?

proteusvacuum:
@gbova :+1: you can do it either before or after, actually. (you can use `--branch` in `commcare-cloud` to run code from a specific branch that isn't master).

PR #2598, opened 2019-01-28
Upgrading to Java 8 for formplayer

sanjay2916:

PR #2629, opened 2019-02-11
Fix java-upgrade-for-formplayer changelog

dannyroberts:
I shouldn't have merged https://github.com/dimagi/commcare-cloud/pull/2598,
but since I did I went ahead and fixed changlog entry to fix the build

@sanjay2916


dannyroberts:
I think once the build passes I'm going to merge this now rather than wait for review so I don't leave the build broken overnight.

PR #2634, opened 2019-02-13
group may be optional

sravfeyn:


sravfeyn:
Its unlikely that others are optional, `touchforms` is an old service. I was able to run this without issues on ICDS with this.

NitigyaS:
I think services like airflow might cause the problem on other environments. I am approving this one though

sravfeyn:
Cool, fixed it.

PR #2639, opened 2019-02-13
Start production backups at 11AM EST

dannyroberts:
Previously they started at 7PM EST and would run all night and into the late morning,
which is our peak load time.

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -33,6 +33,7 @@ backup_couch: True
 postgres_s3: True
 postgresql_backup_days: 1
 postgresql_backup_weeks: 1
+nadir_hour: 16
 
 aws_region: 'us-east-1'
 


dannyroberts:
> Are we using RDS's automated backups now?
Yeah we are
> If backups are taking this long, what's going to be our strategy in 6 months when they take > 24 hours
I guess we have a few options:
- we can have more nodes each with a smaller data size
- we can do less frequent backups
- we can back up the primary data and the views as two separate processes

I'm also hoping that doing it in low-load hours makes the backups take a lot less time

emord:
> * we can back up the primary data and the views as two separate processes

I'm confused are you trying to optimize for couch backups or postgres backups? Would it be worth having different timings for each?

dannyroberts:
Oh this PR is motivated entirely by CouchDB backups. Postgresql backups on production are done entirely by RDS off of the warm standby and don't seem to cause any issues.

emord:
Awesome. thanks for the clarification. I got confused because this variable is referenced in several backup playbooks and it is defined right after some postgres variables

PR #2643, opened 2019-02-13
Fix COUCH_REINDEX_SCHEDULE assignment in localsettings.py

dannyroberts:
I had fixed this yesterday, but forgot to PR it


dannyroberts:
I'm going to go ahead and merge this because the current state is dangerous—if someone ran `update-config` and restarted services, it might actually cause downtime.

PR #2644, opened 2019-02-14
Use built in AWS snapshots to backup production couchdb2 volumes

dannyroberts:
The tar/gzip-style backups we do are taking longer than 24h to run on production and take a _lot_ of space on the pricey SSDs. This turns off our manual backup strategy and replaces it with a policy of taking volume snapshots of couchdb2 data volumes. Currently set to once a day, saving the last 7.


dannyroberts:
I've rolled this out

PR #2645, opened 2019-02-14
default icds-ucr-non-dashboard for softlayer and icds-new

sravfeyn:
Without this [this other PR](https://github.com/dimagi/commcare-hq/pull/23199/files) would break ICDS UCRs on softlayer.

### DIFF
--- a/environments/icds-new/postgresql.yml
+++ b/environments/icds-new/postgresql.yml
@@ -1,6 +1,7 @@
 DEFAULT_POSTGRESQL_HOST: pgmain1
 REPORTING_DATABASES:
   ucr: ucr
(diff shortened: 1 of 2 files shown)


environments/icds-new/postgresql.yml
```diff
@@ -1,6 +1,7 @@
 DEFAULT_POSTGRESQL_HOST: pgmain1
 REPORTING_DATABASES:
   ucr: ucr
+  icds-ucr-non-dashboard: icds-ucr
```
sravfeyn: I know this is perhaps not needed, but updated since it exists and since UCRs have this as an env.

PR #2652, opened 2019-02-15
Reduce nginx fail_timeout to 5s on couch proxy

dannyroberts:
##### SUMMARY

was 10s. This is the amount of time for which it considers a node down
after a timeout or error. Hoping this will punish servers less for the occasional timeout
and prevent ripple effects.

##### ENVIRONMENTS AFFECTED

all

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME

CouchDB (proxy)

##### ADDITIONAL INFORMATION
Nginx conf diff for couch proxy:

```diff
     least_conn;

-  server 10.202.40.135:15984;
-  server 10.202.40.97:15984;
-  server 10.202.40.5:15984;
-  server 10.202.40.100:15984;
-  server 10.202.40.120:15984;
-  server 10.202.40.57:15984;
-  server 10.202.40.134:15984;
-  server 10.202.40.133:15984;
+  server 10.202.40.135:15984 fail_timeout=5s;
+  server 10.202.40.97:15984 fail_timeout=5s;
+  server 10.202.40.5:15984 fail_timeout=5s;
+  server 10.202.40.100:15984 fail_timeout=5s;
+  server 10.202.40.120:15984 fail_timeout=5s;
+  server 10.202.40.57:15984 fail_timeout=5s;
+  server 10.202.40.134:15984 fail_timeout=5s;
+  server 10.202.40.133:15984 fail_timeout=5s;
 }
```


src/commcare_cloud/ansible/roles/nginx/tasks/restart.yml
```diff
@@ -6,9 +6,9 @@
   check_mode: no
   failed_when: "result.rc != 0"
 
-- name: restart the nginx service
+- name: reload the nginx service
   become: yes
-  service: name=nginx state=restarted enabled=yes
+  service: name=nginx state=reloaded enabled=yes
```
dannyroberts: I'm pretty sure this is always what we want to do—that restart causes brief downtime without giving us much if anything. If any objections, I can remove this or make it its own PR

emord: Should the name of this file be changed to reload too?

dannyroberts:
@emord I think this is doing the opposite of what you're describing. From the docs:

> `max_fails=number`
> sets the number of unsuccessful attempts to communicate with the server that should happen in the duration set by the fail_timeout parameter to consider the server unavailable for a duration also set by the fail_timeout parameter. By default, the number of unsuccessful attempts is set to 1. The zero value disables the accounting of attempts. What is considered an unsuccessful attempt is defined by the proxy_next_upstream, fastcgi_next_upstream, uwsgi_next_upstream, scgi_next_upstream, memcached_next_upstream, and grpc_next_upstream directives.
>
> `fail_timeout=time`
> sets
> - the time during which the specified number of unsuccessful attempts to communicate with the server should happen to consider the server unavailable;
> - and the period of time the server will be considered unavailable.
> By default, the parameter is set to 10 seconds.

The change in this PR is basically saying, "when a server times out, take it out of rotation for 5 seconds instead of 10 seconds."

emord:
ah. gotcha. yeah that makes sense :+1: thanks for correcting me

src/commcare_cloud/ansible/roles/nginx/tasks/restart.yml
dannyroberts: done

PR #2661, opened 2019-02-19
Add confirmation_email_queue

gbova:
##### SUMMARY
Adds `confirmation_email_queue` to `celery4`, to be used _only when sending the new domain confirmation email_.

##### ENVIRONMENTS AFFECTED
Production

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
N/A

##### ADDITIONAL INFORMATION
After parallel-izing loading of the template apps for Appcues v3 [here](https://github.com/dimagi/commcare-hq/pull/23261), we are worried that the confirmation email still takes a while to reach the user's inbox and that backups in the `email_queue` will result in a poor UX for new users. Adding `confirmation_email_queue` should speed up the process even more.

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -25,8 +25,6 @@ celery_processes:
     async_restore_queue:
       concurrency: 2
       max_tasks_per_child: 5
-    confirmation_email_queue:
-      concurrency: 1
   celery1:
     flower: {}
     reminder_queue:
@@ -91,6 +89,8 @@ celery_processes:
     export_download_queue:
       concurrency: 6
       max_tasks_per_child: 5
+    confirmation_email_queue:
+      concurrency: 1
 pillows:
   pillow0:
     CaseToElasticsearchPillow:
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -25,6 +25,8 @@ celery_processes:
     async_restore_queue:
       concurrency: 2
       max_tasks_per_child: 5
+    confirmation_email_queue:
+      concurrency: 1
   celery1:
     flower: {}
     reminder_queue:


proteusvacuum:
Build fail looks related. You might need to add it here too:

https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/environment/schemas/app_processes.py#L66

dannyroberts:
Hey @gbova and @proteusvacuum, sorry for the in response since I first saw the suggestion, but it just took me a while to get around to doing enough digging to feel like I might be able to add value by weighing in.

My intuition is that our system isn't such a firehose of outbound email that we'd need more a separate email queue for registration, and yet you're seeing what looks like delays in these emails getting through the `email_queue`. Here are the other things that use email_queue:

```
corehq/apps/accounting/tasks.py
933:@task(serializer='pickle', queue="email_queue")
934-def email_enterprise_report(domain, slug, couch_user):

corehq/apps/registration/tasks.py
61:@task(serializer='pickle', queue="email_queue")
62-def send_domain_registration_email(recipient, domain_name, guid, full_name, first_name):

corehq/apps/hqadmin/tasks.py
67:@task(serializer='pickle', queue="email_queue")
68-def send_mass_emails(username, real_email, subject, html, text):

131:@task(serializer='pickle', queue="email_queue")
132-def send_abnormal_usage_alert(alert):

corehq/apps/hqwebapp/tasks.py
11:@task(serializer='pickle', queue="email_queue",
12-      bind=True, default_retry_delay=15 * 60, max_retries=10, acks_late=True)
13-def send_mail_async(self, subject, message, from_email, recipient_list,

51:@task(serializer='pickle', queue="email_queue",
52-      bind=True, default_retry_delay=15 * 60, max_retries=10, acks_late=True)
53-def send_html_email_async(self, subject, recipient, html_content,

80:@task(serializer='pickle', queue="email_queue",
81-      bind=True, default_retry_delay=15 * 60, max_retries=10, acks_late=True)
82-def mail_admins_async(self, subject, message, fail_silently=False, connection=None,
```

Now if I'm right and we're not sending like hundreds of emails a second, then I thought maybe the way we're processing the `email_queue` right now isn't optimally configured for the workload and is pointlessly leaving performance on the table somehow. Looking at the same file touched in this PR, I see

https://github.com/dimagi/commcare-cloud/blob/a0c7b89ba130b48766ba902862fa4f37002aad2e/environments/production/app-processes.yml#L14-L15

That's saying we have two python processes each synchronously handling one email at a time. I think if we just changed that to

```
    email_queue:
      pooling
      concurrency: 100
      num_workers: 2
```
which says "have two python processes each send up to 100 emails at a time using gevent i/o-based concurrency", that might be as good or better than splitting these emails out into their own queue. If it's just as good, I prefer it in that I'm trying to move us towards fewer queues with more concrete performance targets for each as opposed to proliferating queues with one for every sub-thing we care about. (If it's not just as good, then by all means we should solve the performance issue at hand with the way that works.)

Maybe you guys have thought of that and rejected it for some reason already. Let me know what you think about this other approach.

gbova:
@dannyroberts Thanks for the suggestion and such a thorough explanantion.

I hadn't rejected the idea for any reason and I think what you're saying make a lot of sense. I'm more than happy to modify `email_queue` instead of adding `confirmation_email_queue`, especially if we are trying to limit the number of queues we have. I'm going to close this PR (as well as https://github.com/dimagi/commcare-hq/pull/23291) and open a new PR with the changes you've suggested here.

PR #2666, opened 2019-02-20
Add gcastro pubkey

pr33thi:
First step of this task: https://dimagi-dev.atlassian.net/secure/RapidBoard.jspa?rapidView=5&projectKey=HI&modal=detail&selectedIssue=HI-350&quickFilter=3

I modified the format of the original pubkey following this: https://serverfault.com/questions/797044/openssh-adding-an-ssh-key-from-putty-to-authorized-keys

@dannyroberts 
@proteusvacuum


environments/swiss/meta.yml
```diff
@@ -3,4 +3,5 @@ env_monitoring_id: swiss
 users:
   - dimagi
   - fissionlabs
+  - gcastro
```
proteusvacuum: I think you need to add this user to a `.yml` file like this one:

https://github.com/dimagi/commcare-cloud/blob/master/environments/_users/dimagi.yml#L1

dannyroberts: +1

dannyroberts:
Is this a follow up to https://github.com/dimagi/commcare-cloud/pull/2538? Now looking at that PR more closely it seems it wasn't done in a super standard way (?)

kaapstorm:
> Now looking at that PR more closely it seems it wasn't done in a super standard way

I'm all for the super standard way. Apologies for approving a non-standard PR.

Are there docs on the standard way? (Seeing as I'm on Interrupt I think I'm automatically volunteered to write them if necessary. :grin: )

kaapstorm:
fwiw context:
* [HI-350](https://dimagi-dev.atlassian.net/browse/HI-350)
* [HI-171](https://dimagi-dev.atlassian.net/browse/HI-171)

dannyroberts:
@kaapstorm yeah thanks for asking. I think http://dimagi.github.io/commcare-cloud/commcare-cloud/env/#layout-of-an-environments-directory should mostly cover it, but of course if there's something important missing there, feel free to add it!

pr33thi:
Thanks for updating this while I was offline @kaapstorm

PR #2668, opened 2019-02-21
remove Yamiko as he no longer works for Dimagi

bderenzi:
##### ISSUE TYPE

- Change Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

* Before
```paste below

```

* After
```paste below

```


environments/_users/data_team.yml
```diff
@@ -3,7 +3,6 @@ dev_users:
     - bderenzi
     - mtheis
     - esu
-    - ymsosa
     - slee
     - kchen
   absent: []
```
snopoke: you should add him to the `absent` list which will actually remove his key from the server

PR #2669, opened 2019-02-21
Added citusdb role

NitigyaS:
Review Commit by Commit
##### SUMMARY

Creates A Citus DB Role.
* Supported OS = ubuntu 18.04
* Supported PosgreSQL = 11
* Supported Citus Verison = 8.1

Merging this https://github.com/dimagi/commcare-cloud/pull/2683 require changes.

##### ENVIRONMENTS AFFECTED

Development

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

Ansible : PostgreSQL / CitusDB

##### ADDITIONAL INFORMATION

- Installation of Citusdb managed separately
- Needs a separate Pgbouncer role. (Currently has a copy of its own configurations. )
- Uses PostgreSQL role to create a database.


NitigyaS:
This role is actually dependent on Postgres role only. 
the only task that is separate is the installation of citusdb (Version and OS differs) and activation of extension.
* Creation of database is used form PostgreSQL role
* pgbouncer setup is also taken from there (Once this is merged #2683 all the redundant pgbouncer code will be removed. )
* Pghashlib is not required
* PlProxy is not required

Using PostgreSQL role for the same will require heavy if and else

snopoke:
There are a number of tasks, handers and vars that have been copied from the pg role. Ideally I think the setup should be like this:

roles:
  - postgresql-base  # vanilla postgresql install
  - postgresql-commcare  # depends on postgresql-base (via role dependency). install pghashlib / any other things we want
  - postgresql-citus  # depends on postgresql-base (via role dependency). install citusdb extensions etc.

Even without extracting out postgresql-base I think  you an make citusdb depend on postgresql role since:
* plproxy tasks are already restricted to the proxy hosts
* it shouldn't be hard to limit pghashlib to only pgshard nodes (but also probably doesn't matter if it's also installed on citusdb)

PR #2670, opened 2019-02-21
add tasks to optionally include a python 3.6 virtualenv

nickpell:
##### SUMMARY

Creates a new virtualenv for python 3.6 and adds requirements.  Coexists with the python 2 virtualenv.
Currently is only run if the variable ```py3_include_venv``` is set to true, since python 3.6 does not come by default with Ubuntu 14.04.

##### ENVIRONMENTS AFFECTED

No environments are affected for now.  After the Ubuntu 18.04 migration is complete I will make these tasks standard for all environments.
##### ISSUE TYPE

- Feature Pull Request

@dimagi/py3

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml
@@ -74,6 +74,19 @@
         - slow
       when: not testing|default(False)
 
+    - name: py3 install pip requirements
+      become: true
+      with_items: "{{ commcarehq_requirements }}"
+      pip:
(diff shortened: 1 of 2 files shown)


emord:
> python 3.6 does not come by default with Ubuntu 14.04.

Thoughts on adding https://launchpad.net/~jonathonf/+archive/ubuntu/python-3.7 or https://launchpad.net/~jonathonf/+archive/ubuntu/python-3.6

nickpell:
@emord I might consider that if the Ubuntu upgrade ends up taking a while.  For testing on staging, I built python 3.6 from scratch.

nickpell:
✅

PR #2673, opened 2019-02-21
add settings for python 3 deploy

nickpell:
##### SUMMARY

Add two settings:
```py3_include_venv``` - whether or not to clone and update the python 3 virtualenv.
```py3_run_deploy``` - whether or not to invoke the management commands used for deploy in python 3.

##### ENVIRONMENTS AFFECTED

None

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

For now, none of these settings will be set to true, so there will be no effect on deploys.

@dimagi/py3

### DIFF
--- a/environments/staging/fab-settings.yml
+++ b/environments/staging/fab-settings.yml
@@ -1,2 +1,4 @@
 ignore_kafka_checkpoint_warning: True
 default_branch: autostaging
+py3_include_venv: False
+py3_run_deploy: False
--- a/src/commcare_cloud/environment/schemas/fab_settings.py
+++ b/src/commcare_cloud/environment/schemas/fab_settings.py
@@ -16,6 +16,15 @@ class FabSettingsConfig(jsonobject.JsonObject):
     ignore_kafka_checkpoint_warning = jsonobject.BooleanProperty()
     acceptable_maintenance_window = jsonobject.ObjectProperty(lambda: AcceptableMaintenanceWindow)
     email_enabled = jsonobject.BooleanProperty()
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: environments/staging/fab-settings.yml [hunk 1]
```diff
@@ -1,2 +1,4 @@
 ignore_kafka_checkpoint_warning: True
 default_branch: autostaging
+py3_include_venv: False
+py3_run_deploy: False
```
nickpell: I expect to add future settings for running pillows, celery, webworkers etc in python 3.

src/commcare_cloud/environment/schemas/fab_settings.py
```diff
@@ -16,6 +16,15 @@ class FabSettingsConfig(jsonobject.JsonObject):
     ignore_kafka_checkpoint_warning = jsonobject.BooleanProperty()
     acceptable_maintenance_window = jsonobject.ObjectProperty(lambda: AcceptableMaintenanceWindow)
     email_enabled = jsonobject.BooleanProperty()
+    py3_include_venv = jsonobject.BooleanProperty()
+    py3_run_deploy = jsonobject.BooleanProperty()
```
sravfeyn: Could these be `False` by default?

sravfeyn:
Looks good to me. Looping in @dannyroberts as well.

src/commcare_cloud/environment/schemas/fab_settings.py
nickpell: I believe that ```False``` is the default for ```BooleanProperty```.

PR #2677, opened 2019-02-22
Open RDS to traffic form proxy for websockets features

dannyroberts:
##### SUMMARY

Security groups allow you to block or allow traffic from specific IP ranges. Previously traffic from proxy to RDS (postgresql) was _not_ allowed, simply because I thought it was unnecessary. It turns out that I was wrong, that the proxy _does_ make requests directly to RDS, for our websockets feature.

As an aside, I'm not totally sure, but I'm wondering whether the proxy directly making requests to postgres is the right setup. Still I think first reenabling this to potentially fix broken websockets features ("some else is editing this form" in form builder, for example) is worth it.

##### ENVIRONMENTS AFFECTED
AWS envs: production, staging

##### ISSUE TYPE
- Bugfix Pull Request

(Not sure if this fully fixes a specific user-facing bug, but is definitely a necessary step.)
##### COMPONENT NAME

Websockets

### DIFF
--- a/src/commcare_cloud/terraform/modules/network/main.tf
+++ b/src/commcare_cloud/terraform/modules/network/main.tf
@@ -369,7 +369,12 @@ resource "aws_security_group" "rds" {
     from_port = "5432"
     to_port = "5432"
     protocol = "tcp"
-    cidr_blocks = ["${aws_subnet.subnet-app-private.*.cidr_block}", "${aws_subnet.subnet-db-private.*.cidr_block}"]
+    cidr_blocks = [
+      "${aws_subnet.subnet-app-private.*.cidr_block}",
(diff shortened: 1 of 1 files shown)


dannyroberts:
This is rolled out, can watch https://app.datadoghq.com/dashboard/g9s-pw6-tpg/hq-vitals?screenId=g9s-pw6-tpg&screenName=hq-vitals&from_ts=1550685600000&is_auto=false&live=true&page=0&tile_size=s&to_ts=1550858400000&tpl_var_environment=production&fullscreen_widget=200684876 to see whether it fixes these errors

dannyroberts:
Oh, this actually might be the wrong thing, I would need to fix the connection between the proxy and the pgbouncer machines, not this. Closing

PR #2679, opened 2019-02-22
Separate celery check

dannyroberts:
##### SUMMARY
This adds a specific http check for the celery check (serverup.txt?only=celery) to datadog.

##### ENVIRONMENTS AFFECTED
Environments that are not relying on datadog for monitoring can ignore this change.

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME

Datadog monitoring 

##### ADDITIONAL INFORMATION
Example ansible diff:
```
--- before: /etc/dd-agent/conf.d/http_check.yaml
+++ after: /Users/droberts/.ansible/tmp/ansible-local-97263sd5Pmr/tmpygx_Im/http_check.yaml.j2
@@ -11,6 +11,12 @@
     include_content: true
     tags:
       - check_type:serverup
+  - name: production Celery
+    url: https://www.commcarehq.org/serverup.txt?only=celery
+    check_certificate_expiration: false
+    timeout: 30
+    tags:
+      - check_type:celery
   - name: production Heartbeat
     url: https://www.commcarehq.org/serverup.txt?only=heartbeat
     check_certificate_expiration: false

changed: [10.202.20.138] => (item={u'enabled': True, u'name': u'http_check'})
```

### DIFF
--- a/changelog/0015-separate-celery-datadog-http-check.yml
+++ b/changelog/0015-separate-celery-datadog-http-check.yml
@@ -1,0 +1,22 @@
+title: Separate celery datadog http check
+key: separate-celery-datadog-http-check.yml
+date: 2019-02-22
+optional_per_env: yes
+min_commcare_version:
+max_commcare_version: 81325fbd9e3131c710179f7246dbe9caccb45154
+context: |
+  This adds a specific http check for the celery check (serverup.txt?only=celery)
+  to datadog.
+  Environments that are not relying on datadog for monitoring can ignore this change.
+
+details: |
+  This will result in three distinct http checks:
+    - "serverup" endpoint (high severity)
+    - "celery" endpoint (usually lower severity)
+    - "heartbeat" endpoint (usually lower severity)
+
+update_steps: |
+  1. Update datadog integrations on the proxy machine:
+  ```bash
+  commcare-cloud <env> deploy-stack --limit=proxy --tags=datadog_integrations
+  ```
--- a/docs/changelog/0015-separate-celery-datadog-http-check.md
+++ b/docs/changelog/0015-separate-celery-datadog-http-check.md
@@ -1,0 +1,28 @@
+# 15. Separate celery datadog http check
+
+**Date:** 2019-02-22
+
+**Optional per env:** _only required on some environments_
+
+
+## CommCare Version Dependency
+CommCare versions beyond this commit require this change to function correctly:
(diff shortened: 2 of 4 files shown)


changelog/0015-separate-celery-datadog-http-check.yml
```diff
@@ -0,0 +1,22 @@
+title: Separate celery datadog http check
+key: separate-celery-datadog-http-check.yml
+date: 2019-02-22
+optional_per_env: yes
+min_commcare_version:
+max_commcare_version: 81325fbd9e3131c710179f7246dbe9caccb45154
```
kaapstorm: Just confirming that it is `max_commcare_version`, and not `min_commcare_version`, that means "this change should be applied from this revision number onwards".

I might be misunderstanding, but it looks like `0005-support-multiple-kafak-brokers.yml`, `0007-reorganize-pillows.yml`, `0008-blob-metadata-part-1.yml`, and `0009-blob-metadata-part-2.yml` use `min_commcare_version` to mean this, and `0004-add-analytics-queue.yml`, `0006-new-case-importer-celery-queue.yml`, and `0012-generalize-load-case-from-fixture.yml` use `max_commcare_version` to mean the same thing. ... Or are 0005, 0007, 0008 and 0009 NOT applied after their given revision numbers?

dannyroberts: This part of the docs template for these changes is in my mind the best explanation for the difference: https://github.com/dimagi/commcare-cloud/blob/3ba62c0e12ead0147daf4ca3c77afcfe62f4c900/src/commcare_cloud/manage_commcare_cloud/changelog.md.j2#L13-L21. I myself refer back to that whenever I'm unsure

docs/changelog/0015-separate-celery-datadog-http-check.md
```diff
@@ -0,0 +1,28 @@
+# 15. Separate celery datadog http check
+
+**Date:** 2019-02-22
+
+**Optional per env:** _only required on some environments_
+
+
+## CommCare Version Dependency
+CommCare versions beyond this commit require this change to function correctly:
+[81325fbd](https://github.com/dimagi/commcare-hq/commit/81325fbd9e3131c710179f7246dbe9caccb45154)
```
dannyroberts: So @kaapstorm for example this is what `max_commcare_version` means for this change.

PR #2680, opened 2019-02-22
invoke celery directly, without management command

nickpell:
##### SUMMARY

This change is a prerequisite for upgrading to celery 4.  We currently use django_celery to invoke celery as a management command, but this is obsolete and django_celery is incompatible with celery 4.

Pulled from https://github.com/dimagi/commcare-cloud/pull/2058/

##### ENVIRONMENTS AFFECTED

Must be deployed to all environments.

##### ISSUE TYPE

- Change Pull Request

##### COMPONENT NAME

```cchq <env> update-supervisor-confs```

##### ADDITIONAL INFORMATION

Merge only after the next deploys to swiss and pna.  Requires https://github.com/dimagi/commcare-hq/pull/23319

### CODE REVIEW

File: changelog/0015-invoke-celery-directly.yml [hunk 1]
```diff
@@ -0,0 +1,17 @@
+title: Update supervisor confs to invoke celery directly
+key: invoke-celery-directly
+date: 2019-02-22
+optional_per_env: no
+min_commcare_version: f67f9c8149bd71712f457c270b03b5a55077c273
+max_commcare_version:
```
nickpell: ```max_commcare_version``` will be added retroactively after https://github.com/dimagi/commcare-hq/pull/21588 is merged.

changelog/0015-invoke-celery-directly.yml
```diff
@@ -0,0 +1,17 @@
...
+context: |
+  Upgrading to celery 4.x requires removing the dependency on
+  django-celery, which means that the celery management command
+  becomes unavailable.  This prepares for that by invoking the
+  celery command directly.
+
```
dannyroberts: Nit: adding a `detail:` field here would get rid of the bare "None" in the markdown output below. I think this would be a good place to go over some some details of the importance of the rollout sequence. Something that will call out the importance of sequencing things right and what goes wrong if you don't.

nickpell: @dannyroberts updated

PR #2684, opened 2019-02-25
Ns/seperate postgres base role

NitigyaS:
##### SUMMARY
Splits postgresql role as follows:
* postgresql_base: install and configure postgresql
* postgresql: install and configure postgresql extensions required by HQ (depends on postgresql base)
* pgbouncer: install and configure pgbouncer

##### ISSUE TYPE

- Change Pull Request

##### COMPONENT NAME
postgresql

##### ADDITIONAL INFORMATION
Work required for CitusDB role


snopoke:
@dannyroberts please could you review this

.travis/environments/travis/postgresql.yml
```diff
@@ -5,7 +5,7 @@ dbs:
   synclogs: null
   form_processing: null
 
-override:
+override_postgres:
```
dannyroberts: Should this say `postgres_override`?

environments/echis/postgresql.yml
```diff
@@ -1,6 +1,8 @@
-override:
+pgbouncer_override:
```
dannyroberts: I think it's worth keeping in mind that if once we have people we can't speak to directly managing their own instances, we have to be very careful about changing the format of the environment files. And when we do, we'll have to create a changelog outlining how someone else needs to update their environments, to fit the change, and update the docs (http://dimagi.github.io/commcare-cloud/commcare-cloud/env/#postgresqlyml) for that file.

.travis/environments/travis/postgresql.yml
snopoke: yes

environments/echis/postgresql.yml
snopoke: yes for sure. I did consider that and though about trying some other method like splitting by var prefix but that seemed a bit hacky.

I will definitely put together a changelog for this.

snopoke:
@dannyroberts I've added a changelog entry. Regarding your question about how this will help with citus, it will mean we can use the postgresql_base role to install and configure vanilla postgres and then have a separate role to install the citus extensions etc.

I think this is much cleaner than adding tasks to the existing role with when clauses.

src/commcare_cloud/ansible/roles/postgresql_base/defaults/main.yml
```diff
@@ -33,3 +33,7 @@ postgresql_slow_log_threshold: 300
 postgresql_random_page_cost: 4.0
 
 allow_direct_postgresql_access: no
+
+postgresql_apt_repo: "deb http://apt.postgresql.org/pub/repos/apt/ {{ ansible_lsb.codename}}-pgdg main"
```
NitigyaS: For citus repo and binary differs , which can be passed at role dependency
```
postgresql_version: '11'
citus_version: '8.1'
postgresql_apt_repo: "deb https://packagecloud.io/citusdata/community/ubuntu/ {{ ansible_lsb.codename}} main"
postgresql_apt_key_url: "https://repos.citusdata.com/community/gpgkey"
postgresql_binary: "postgresql-{{postgresql_version}}-citus-{{citus_version}}"```

Comment (snopoke) [reply src/commcare_cloud/ansible/roles/postgresql_base/defaults/main.yml]: the binary that you list here isn't actually a postgres binary - it's a postgres extension that just depends on postgres so if you install it, it will also install postgres.

I'm fine with makint these configurable but I don't think the citus role should change them, rather it should just set the postgresql version, allow the pg role to install postgres, and then it can install the citus extension from the repo you've listed.

Comment (NitigyaS) [reply src/commcare_cloud/ansible/roles/postgresql_base/defaults/main.yml]: Okay, that makes sense. I was confused the whole time that it is Postgres binary.

File: src/commcare_cloud/ansible/roles/citusdb/tasks/setup_worker.yml [hunk 4]
```diff
@@ -0,0 +1,29 @@
+---
```
snopoke: is it necessary to create users on the workers? I'm not clear how citus communicates with the workers. Also wondering if there needs to be provision in the pg_hba.conf file.

NitigyaS: I think these users are only used by the application, Citus should have it's own way of communicating. 
But Since we have not installed pgbouncer on workers, I think application never interacts with worker and thus these users are not very useful in that sense. 
From provision in the pg_hba.conf , do you mean adding configurations so coordinator can connect to the worker or allowing these users to connect.

snopoke: Allowing coordinator to connect to workers. In the tests I did I don't recall having to change anything but just wanted to make sure.

src/commcare_cloud/ansible/group_vars/citusdb_worker.yml
```diff
@@ -1,3 +1,4 @@
 datadog_integration_pgbouncer: false  
 datadog_integration_postgres: true
-is_pg_standby: false
\ No newline at end of file
+is_pg_standby: false
+allow_direct_postgresql_access: true
```
snopoke: this just give `du.user` access right? I don't think we need to do that

snopoke: I also think we can remove this flag since it was just used for running a test prior to moving to RDS

NitigyaS: I found this flag was being used in pg_hba.conf to enable configs. I used the same flag to enable pg_hba.conf in workers. (Otherwise all connection goes through pgbouncer which is not present on workers. )
https://github.com/dimagi/commcare-cloud/blob/3c540e43839b977ed0fb6d596651a4a9df44a479/src/commcare_cloud/ansible/roles/postgresql/templates/pg_hba.conf.j2#L40

snopoke: I still don't think we should use this flag. There shouldn't be any connections to the workers that aren't coming from the coordinator anyway.

PR #2685, opened 2019-02-25
Add AWS commcare-india account

dannyroberts:
##### SUMMARY

These are the code changes I made in order to set up the basics for the new AWS commcare-india account, which will be replacing softlayer as our host for india.commcarehq.org.

I decided to rename the "softlayer" environment to india and then symlink softlayer/ to india/ as a way to preserve it as an alternate alias. I also changed `deploy_env` and `env_monitoring_id` from `softlayer` to `india`. I thought about waiting to do that later, but figured if we were going to make the change eventually, we would actually benefit from doing it earlier so that by the time we do the rest of the migration, we already have the minor annoyance of the name switch behind us and don't have to worry about it at the same time. (We could also just continue calling it "softlayer", but the annoyance of changing seemed less than the annoyance and confusion of continuing to call it by the name of a hosting provider we were no longer using.) The primary affects of this will be (1) changing the directory structure from `/home/cchq/www/softlayer` to `/home/cchq/www/india` and (2) changing the datadog metrics to be collected with "environment:india" instead of "environment:softlayer". For (1), I will just be symlinking `/home/cchq/www/india` to `/home/cchq/www/softlayer` on all existing machines, and then let our tooling create the directory normally on any new machine going forward. For (2) I'll roll it out by running `cchq india deploy-stack --tag=datadog`

##### ENVIRONMENTS AFFECTED
"softlayer", to be called "india" going forward
##### ISSUE TYPE

- Change Pull Request
- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME

India hosting; terraform, openvpn


environments/india/public.yml
```diff
@@ -117,3 +117,6 @@ site_locations:
     try_files: "$uri $uri/index.html $uri/ =404"
     auth_basic: '"Restricted Content"'
     auth_basic_user_file: '/etc/nginx/.htpasswd_enikshay_private'
+
+commcare_cloud_root_user: ubuntu
+commcare_cloud_pem: ~/.ssh/id_rsa
```
snopoke: why did these get added? are they correct?

src/commcare_cloud/environment/schemas/terraform.py
```diff
@@ -19,7 +19,7 @@ class TerraformConfig(jsonobject.JsonObject):
     servers = jsonobject.ListProperty(lambda: ServerConfig)
     proxy_servers = jsonobject.ListProperty(lambda: ServerConfig)
     rds_instances = jsonobject.ListProperty(lambda: RdsInstanceConfig)
-    redis = jsonobject.ObjectProperty(lambda: RedisConfig)
+    redis = jsonobject.ObjectProperty(lambda: RedisConfig, default=None)
```
snopoke: why is this becoming optional?

src/commcare_cloud/terraform/README.md
```diff
@@ -164,10 +181,11 @@ To activate a user, run
 cchq <env> openvpn-activate-user <username>
 ```
 
-and then have the user (in this case, yourself) change the password with
+and then have the user (in this case, yourself)
+claim the user and set their password with
 
 ```
-cchq <env> ssh openvpn -t passwd
+cchq india openvpn-claim-user <username>
```
snopoke: shouldn't this still be `<env>`?

src/commcare_cloud/terraform/modules/openvpn/variables.tf
```diff
@@ -3,7 +3,7 @@ data "aws_ami" "openvpn_image" {
   most_recent = true
   filter {
     name   = "name"
-    values = ["OpenVPN Access Server 2.5.0-3b5882c4-551b-43fa-acfe-7f5cdb896ff1-ami-548e4429.4"]
+    values = ["OpenVPN Access Server 2.6.1-fe8020db-5343-4c43-9e65-5ed4a825c931-ami-0f5d312e085235ed4.4"]
```
snopoke: will this trigger a change on prod as well?

dannyroberts: It _would_, but in this PR I also override this in [`production/terraform.yml`](https://github.com/dimagi/commcare-cloud/pull/2685/files#diff-aca45def687ae46ebf00815b019dfef8) as well as `staging/terraform.yml`, to their current value.

The original query was no longer working, possibly because of the change of region. These updates combined made it run smoothly on india, while also not resulting in any changes to production or staging

src/commcare_cloud/terraform/README.md
dannyroberts: 👍 thanks

environments/india/public.yml
dannyroberts: Yeah this is right. These are also in `production/public.yml` (and `staging/public.yml`), and they together outline the "factory auth" for this environment. While on some environments the "factory auth" (the auth you should use to log into a brand new machine for the first time, usually to run `bootstrap-users`) is a username and a password, on AWS (and I'm sure some other providers) factory auth is done with the ssh key of the person who created the machine (logging in as the `ubuntu` user for Ubuntu AMIs). AWS lets you update a bunch of ssh public keys and then choose one to use when you create a new EC2 instance; our terraform code is set up to automatically select the key of person creating it.

src/commcare_cloud/environment/schemas/terraform.py
dannyroberts: This part of `terraform.yml` hooks into a terraform module that creates an elasticache instance, but we ended up _not_ using elasticache, at least not yet. (When migrating our cache data into elasticache ended up not looking super straightforward, I just cut bait and set redis up on a VM like we do elsewhere. We may very well get value eventually out of using elasticache, just didn't seem worth slowing down the AWS migration in general for at the time.)

This schema change basically just says if you're not using elasticache, you can just omit the `redis` section.

dannyroberts:
OK I just ran

```
cchq india run-shell-command all 'ls /home/cchq/www/softlayer && ln -s /home/cchq/www/softlayer /home/cchq/www/india' -b --become-user cchq
```

like I mentioned above to make softlayer => india renaming transition smooth.

dannyroberts:
From my side I'm all set to merge this and roll out the datadog side of it, pending approval

src/commcare_cloud/environment/schemas/terraform.py
snopoke: should it not then be called `elasticache` instead of `redis`?

dannyroberts: Great point :) I'll make a note to do that in a separate PR

PR #2686, opened 2019-02-25
Fix formplayer deploy after recent change

dannyroberts:
##### SUMMARY
fixes https://github.com/dimagi/commcare-cloud/pull/2682

In addition, I found that another assumption from when I made that PR weren't correct. I found that since possibly forever, formplayer has been restarted _twice_ on every deploy: once unnecessarily before we've re-linked the current commcare-hq release (thus not picking up any changes) and once again afterwards when we restart services. In this PR I remove the first time entirely, and make the second time only happen if we detect an actual change in the formplayer jar

##### ENVIRONMENTS AFFECTED
All
##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Formplayer deploy

### DIFF
--- a/src/commcare_cloud/fab/operations/formplayer.py
+++ b/src/commcare_cloud/fab/operations/formplayer.py
@@ -23,7 +23,8 @@ def _formplayer_jars_differ(build_dir, release_1, release_2):
     with cd(build_dir):
         with settings(warn_only=True):
             result = sudo("diff {}/libs/formplayer.jar {}/libs/formplayer.jar".format(release_1, release_2))
-    return result.return_code == 1
+    # 1 means there's a diff, 2 means one of the files doesn't exist
+    return result.return_code in (1, 2)
 
 
 @roles(ROLES_FORMPLAYER)
@@ -58,7 +59,8 @@ def build_formplayer(use_current_release=False):
         with cd(build_dir):
             sudo('ln -sfn {} current'.format(release_name))
             sudo('ln -sf current/libs/formplayer.jar formplayer.jar')
-        supervisor.restart_formplayer()
+        env.NEEDS_FORMPLAYER_RESTART = True
+
 
 
 @roles(ROLES_FORMPLAYER)
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -819,6 +819,8 @@ def silent_services_restart(use_current_release=False):
     """
     execute(db.set_in_progress_flag, use_current_release)
     if not env.is_monolith:
+        if getattr(env, 'NEEDS_FORMPLAYER_RESTART', False):
+            execute(supervisor.restart_formplayer)
         execute(supervisor.restart_all_except_webworkers)
     execute(supervisor.restart_webworkers)
 
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -412,13 +412,11 @@ def copy_localsettings(full_cluster=True):
 @parallel
 @roles(ROLES_FORMPLAYER)
 def copy_formplayer_properties():
-    sudo('mkdir -p {}'.format(os.path.join(env.code_root, FORMPLAYER_BUILD_DIR)))
-    for filename in ['application.properties', 'sentry.properties']:
-        sudo(
-            'cp {} {}'.format(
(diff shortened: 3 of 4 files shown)


src/commcare_cloud/fab/operations/formplayer.py
```diff
@@ -23,7 +23,8 @@ def _formplayer_jars_differ(build_dir, release_1, release_2):
     with cd(build_dir):
         with settings(warn_only=True):
             result = sudo("diff {}/libs/formplayer.jar {}/libs/formplayer.jar".format(release_1, release_2))
```
emord: is `diff` a good way to do this sort of thing? I thought it wasn't very good for binary files.

It might be better to see if we can upload the md5sum to s3 as well and we could download that first and not even download the jar at all

dannyroberts: I think diff is fine at it in the sense that it'll reliably have a return code of 1 if they differ and 0 if they don't. Have you heard/seen otherwise? Of course, it won't actually give a good text diff for binary files, but that's fine for our purposes.

dannyroberts: In terms of not needing to actually download the large file to get a diff, I agree that's a good thing to try.

dannyroberts:
I think tests are going to take a while to run, and wouldn’t catch an issue in this part of the code anyway, so I’m going to just merge

src/commcare_cloud/fab/operations/formplayer.py
```diff
@@ -23,7 +23,8 @@ def _formplayer_jars_differ(build_dir, release_1, release_2):
...
-    return result.return_code == 1
+    # 1 means there's a diff, 2 means one of the files doesn't exist
+    return result.return_code in (1, 2)
```
millerdev: why not `!= 0`? I believe [`diff` only returns `0` if the files are the same](https://stackoverflow.com/questions/6971284/what-are-the-error-exit-values-for-diff).

src/commcare_cloud/fab/operations/formplayer.py
emord: > I think diff is fine at it in the sense that it'll reliably have a return code of 1 if they differ and 0 if they don't.

Yeah I think it's fine from a "are these two files the same" standpoint. I haven't looked at actually diffing files recently, but I thought binary files were pretty bad for diff since they may have trouble finding anchor points for diff if there actually are changes. Though perhaps those algorithms are fairly sophisticated and can abort if it takes too long

so basically I think it's likely very fast for when these are the same, but there may be some set of differences where diff takes quite a while. I think comparing output from  `md5sum` would never have that issue

src/commcare_cloud/fab/operations/formplayer.py
dannyroberts: Yeah I think the difference between `in (1, 2)` and `!= 0` has to do with what we do when there's an unexpected error code. Ideally we fail hard if anything goes wrong, return `True` if the files are positively the different and `False` if they're positively the same. I think both the way I'm doing it and modifying to `!= 0` have the same issue which is that there are plenty of ways something could error where that error would be swallowed and simply result in one of the two possible normal values (True/False).

src/commcare_cloud/fab/operations/formplayer.py
dannyroberts: I see, I hadn't thought of the performance on binary files. I'm used to it just saying "the binary files differ" which I assumed was easy and quick for it to do.

Anyway I think reworking to use md5 sums would be very reasonable, and something I might do.

dannyroberts:
Both of you guys raise good points that I may incorporate soon. For _right_ now I'd like to sit on this and make sure that the changes I've made (whatever downstream flaws they might have) work on the next few deploys, since my original change had issues with it, I just want to make sure we're definitely in a good steady state before making other changes. I'll make a note to myself to revisit this and incorporate improvements related to these comments.

src/commcare_cloud/fab/operations/formplayer.py
millerdev: My thinking was that `!= 0` would be a more conservative change than `in (1, 2)`: it will cause a restart (return `True`) when it is maybe not necessary (due to some unknown failure mode of `diff`), but will always prevent a restart (return `False`) when `diff` reports that the files are the same.

As is, an unknown `diff` failure mode will prevent a (possibly necessary) restart.

PR #2690, opened 2019-02-26
Fix certbot renew

dannyroberts:
##### SUMMARY

The `/etc/cron.d/certbot` was a mix of what certbot puts there automatically and what we configured in ansible, leaving it with two jobs, one of which ran every 12 hours and _didn't_ restart nginx, the other of which ran every `hours: 0/12` (which I'm not sure what that means, maybe the same as just `0`, and I think it was a mistake), which would have restarted nginx but didn't always get the opportunity to run because the other one would have already renewed the cert. This put us in a situation on swiss where the cert was renewed but we hadn't restarted nginx to pick it up, leaving it serving an expired cert.

This change deletes the default certbot cron file and replaces it with one managed entirely by ansible.

##### ENVIRONMENTS AFFECTED

All

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

certbot

##### ADDITIONAL INFORMATION

```
TASK [Remove default certbot cron] *************************************************************************************************************************************************************************
--- before: /etc/cron.d/certbot
+++ after: /dev/null
@@ -1,13 +0,0 @@
-# /etc/cron.d/certbot: crontab entries for the certbot package
-#
-# Upstream recommends attempting renewal twice a day
-#
-# Eventually, this will be an opportunity to validate certificates
-# haven't been revoked, etc.  Renewal will only occur if expiration
-# is within 30 days.
-SHELL=/bin/sh
-PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
-
-0 */12 * * * root test -x /usr/bin/certbot && perl -e 'sleep int(rand(3600))' && certbot -q renew
-#Ansible: Certbot Renew
-0 0/12 * * * root test -x /usr/bin/certbot && perl -e 'sleep int(rand(3600))' && certbot -q renew --renew-hook '/etc/init.d/nginx reload'

changed: [10.202.20.138]

TASK [Add custom_certbot cron] *****************************************************************************************************************************************************************************
--- before: /etc/cron.d/custom_certbot
+++ after: /etc/cron.d/custom_certbot
@@ -0,0 +1,2 @@
+#Ansible: Certbot Renew
+0 */12 * * * root test -x /usr/bin/certbot && perl -e 'sleep int(rand(3600))' && certbot -q renew --renew-hook '/etc/init.d/nginx reload'

changed: [10.202.20.138]

```


dannyroberts:
Somehow I never merged this... 🤦‍♂️

PR #2692, opened 2019-02-26
Clean up solo_queues declaration in app_processes.py

dannyroberts:
##### SUMMARY

Just a little thing I come across sometimes. The comment here wasn't quite right for all queues there and I thought I could break it down a little more clearly. Also removed some things that weren't defined as a `CeleryProcess` and thus didn't need to be in the list.

##### ENVIRONMENTS AFFECTED
It's a no-op change.

##### ISSUE TYPE
- Comment Pull Request

##### COMPONENT NAME
Celery config


kaapstorm:
Nice clarification.

PR #2693, opened 2019-02-26
use redis as celery results backend

nickpell:
##### SUMMARY

Currently we use a postgres table defined by ```django-celery``` as the celery results backend.  We are removing this library in order to upgrade to celery 4.x, so we have to replace the results backend.  I choose redis since this is the recommended best practice, since it reduces pressure on postgres.

##### ENVIRONMENTS AFFECTED

All

##### ISSUE TYPE

- Change Pull Request

##### COMPONENT NAME

```commcare-cloud <env> update-config```

Recommend reviewing commit by commit 🐡

### DIFF
--- a/changelog/0017-use-redis-celery-results-backend.yml
+++ b/changelog/0017-use-redis-celery-results-backend.yml
@@ -1,0 +1,22 @@
+title: Update supervisor confs to invoke celery directly
+key: invoke-celery-directly
+date: 2019-02-26
+optional_per_env: no
+min_commcare_version:
+max_commcare_version:
+context: |
+  Upgrading to celery 4.x requires removing the dependency on
+  django-celery, which means that the results backend provided
+  by django-celery has to be replaced.  This change configures
+  celery to use the redis results backend instead.
+
+details: |
+  This change must be applied before the django-celery dependency
+  is removed from the commcare-hq library, otherwise celery
+  will not be able to store task results.
+
+update_steps: |
+  Update supervisor confs:
+  ```bash
+  commcare-cloud <env> update-config
+  ```
--- a/docs/changelog/0017-use-redis-celery-results-backend.md
+++ b/docs/changelog/0017-use-redis-celery-results-backend.md
@@ -1,0 +1,27 @@
+# 17. Update supervisor confs to invoke celery directly
+
+**Date:** 2019-02-26
+
+**Optional per env:** _required on all environments_
+
+
+## CommCare Version Dependency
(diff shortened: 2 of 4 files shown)


nickpell:
> Will this preserve existing results? Or will tasks that start/end around the time of the switch over be essentially orphaned, causing undesireable behavior in currently downloading exports, etc.?

It won't preserve existing results, but I did an [audit](https://docs.google.com/document/d/18OWQZnkRLypPKhpHfLAEjOybqvwWvxcnxEcj-7ElAU8/edit) of where we use the results and it shouldn't be a big deal.  To minimize effects, I'm going to deploy at a low usage time, and stop celery beforehand so there won't be ongoing processes that would risk being affected by the backend change.

nickpell:
> I'm wondering if we're okay with the occasional result getting deleted from redis before it's been used, when redis is running out of memory and starts deleting keys.

Our eviction policy is ```volatile-lru```: https://github.com/dimagi/commcare-cloud/blob/1e44b726c33f621bef7c5a7657404b4310f3717c/src/commcare_cloud/ansible/group_vars/redis.yml#L10
(see https://redis.io/topics/lru-cache)

I'm going to let @snopoke comment on the implications of our eviction policy.

Task results have a default expiry: http://docs.celeryproject.org/en/latest/userguide/configuration.html#result-expires

sravfeyn:
Throwing out an option, we could use a standalone redis instance just for this in the beginning, which can be run on the same redis machine (with main redis eviction policy being strict)

snopoke:
I do have some concerns about switching to redis particularly regarding cache eviction which is not something we monitor well. We could certainly set up a separate redis instance but we could also as easily stick with PostgreSQL which is one of the supported backends.

@nickpell where did you see that it's best practice to use Redis?

nickpell:
> @nickpell where did you see that it's best practice to use Redis?

@snopoke I've come across it in a few articles like https://khashtamov.com/en/celery-best-practices-practical-approach/, but never in official documentation.

For the reasons you mentioned I think it's best to stick with a database backend for now since it sounds like using redis robustly might add additional complexity.  It's easy to switch between results backends so we can always use redis in the future if we change our minds.

snopoke:
Thanks for the link. I agree with the article about not using a database as a broker but the author doesn't really give any reason not to use a DB as the result backend. Either way, we rarely have any issues with using the DB as the backend and since the setup is simpler for now I think it's the right choice.

nickpell:
Closing in favor of https://github.com/dimagi/commcare-cloud/pull/2698

PR #2694, opened 2019-02-26
add ability to run all python code in 3.6

nickpell:
##### SUMMARY

This change allows you to pick which process on which environments to run in python 3.  This might be useful for testing particular components of the system in isolation.

I couldn't find a way to provide default values for the settings, but that would be preferable than marking each one on each environment.

##### ENVIRONMENTS AFFECTED

None

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

```cchq <env> update-supervisor-confs```

@dimagi/py3


millerdev:
I think @dannyroberts should review this. He may know if there is a way to provide default values for the settings.

nickpell:
Going to merge now but curious to hear from @dannyroberts if there's a way to provide default values for settings.

PR #2695, opened 2019-02-27
Ns/ciuts role

NitigyaS:
##### SUMMARY

CitusDB on top of PostgresBase .

##### ENVIRONMENTS AFFECTED

development
##### ISSUE TYPE

- Feature Pull Request

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,7 +10,7 @@ require_relative './provisioning/key_authorization'
 
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.box = "ubuntu/trusty64"
+  config.vm.box = "ubuntu/bionic64"
   cchq_proxy_port = ENV.fetch("VAGRANT_CCHQ_PROXY_PORT", 8080)
   config.ssh.insert_key = false
   authorize_key config, '~/.vagrant.d/insecure_private_key'
@@ -18,6 +18,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   {
     'app1'    => '192.168.33.15',
     'db1'   => '192.168.33.16',
+    'citusdb0' => '192.168.33.30',
+    'citusdb1'   => '192.168.33.31',
+    'citusdb2'   => '192.168.33.32',
+    'citusdb3'   => '192.168.33.33',
     'proxy1' => '192.168.33.17',
   }.each do |short_name, ip|
     config.vm.define short_name do |host|
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -18,7 +18,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   {
     'app1'    => '192.168.33.15',
     'db1'   => '192.168.33.16',
-    'citusdb0' => '192.168.33.30',
     'citusdb1'   => '192.168.33.31',
     'citusdb2'   => '192.168.33.32',
     'citusdb3'   => '192.168.33.33',
--- a/environments/development/postgresql.yml
+++ b/environments/development/postgresql.yml
@@ -13,7 +13,3 @@ dbs:
   form_processing: null
   ucr:
     host: citusdb1
-  ucr:
-    host: citusdb2
-  ucr:
-    host: citusdb3
--- a/environments/development/postgresql.yml
+++ b/environments/development/postgresql.yml
@@ -11,3 +11,9 @@ postgres_override:
 dbs:
   synclogs: null
   form_processing: null
+  ucr:
+    host: citusdb1
+  ucr:
+    host: citusdb2
+  ucr:
+    host: citusdb3
--- a/src/commcare_cloud/ansible/deploy_db.yml
+++ b/src/commcare_cloud/ansible/deploy_db.yml
@@ -8,6 +8,7 @@
     - riakcs
     - formplayer
     - django_manage
+    - citusdb
   become: true
   roles:
     - {role: ecryptfs, tags: 'ecryptfs'}
@@ -16,7 +17,12 @@
     - import_tasks: roles/ecryptfs/tasks/remove-tmp.yml
       tags: ecryptfs-cleanup
 
-- import_playbook: deploy_postgres.yml
+
+- name : PostgreSQL
+  import_playbook: deploy_postgres.yml
+
+- name : CitusDB
+  import_playbook: deploy_citusdb.yml
 
 - name: Couchdb
   hosts: couchdb
--- a/src/commcare_cloud/ansible/deploy_db.yml
+++ b/src/commcare_cloud/ansible/deploy_db.yml
@@ -8,7 +8,6 @@
     - riakcs
     - formplayer
     - django_manage
-    - citusdb
   become: true
   roles:
     - {role: ecryptfs, tags: 'ecryptfs'}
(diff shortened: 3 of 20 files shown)


environments/development/postgresql.yml
```diff
@@ -11,3 +11,9 @@ postgres_override:
 dbs:
   synclogs: null
   form_processing: null
+  ucr:
+    host: citusdb1
+  ucr:
+    host: citusdb2
+  ucr:
+    host: citusdb3
```
snopoke: how come these are repeated so 3 times? Is it just to get the DB created? If that's the reason I think it would be better to have the citus role import the `set_up_dbs` task file to do the necessary setup on the workers and just leave the coordinator here. 

Also I you can't have multiple keys with the same name.

src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
```diff
@@ -1,3 +1,4 @@
+postgresql_port: 5432
```
snopoke: sucks that this is needed here. What do you think about removing it from the pgbouncer.ini since pgbouncer will use 5432 if it's not provided (and we always use the standard port)?

src/commcare_cloud/ansible/roles/citusdb/meta/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+dependencies:
+  - role: monit
+  - role: postgresql_base
+    postgresql_version: '11'
+    postgresql_process_path: /var/run/postgresql
+    postgresql_cluster_name: '11/main'
+    postgresql_shared_preload_libraries:
+     - 'citus'
```
snopoke: this will be added before citus is installed. Will that prevent PG from starting?

src/commcare_cloud/ansible/roles/citusdb/meta/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+dependencies:
+  - role: monit
+  - role: postgresql_base
+    postgresql_version: '11'
+    postgresql_process_path: /var/run/postgresql
+    postgresql_cluster_name: '11/main'
```
snopoke: why do these need to be overridden?

NitigyaS: default cluster name in `11/main` if it is not mentioned the `postgresql` fails to restart. 
in `bionic` the process path in `system service` file is `/var/run/postgresql`, again it was failing to start when i changed it to `/opt/data/postgresql`. It would be a good idea to set this on distribution type instead of passing as variable .

src/commcare_cloud/ansible/roles/citusdb/meta/main.yml
NitigyaS: I tried is to set it after installing citus, but than PostgreSQL had to be restart. It was annoying since everytime I run the playbook it was restarting/reloading as the base role was setting it empty and `citus role` was again setting it to `citus`
I didn't find any trouble while testing. It could also be because I already had citus installed in test machines.

src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
NitigyaS: will test and implement.

environments/development/postgresql.yml
NitigyaS: Are you saying to use this to create DB on coordinator and write a separate task to create DB on workers?

src/commcare_cloud/ansible/deploy_db.yml
```diff
@@ -8,6 +8,7 @@
     - riakcs
     - formplayer
     - django_manage
+    - citusdb
```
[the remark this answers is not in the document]
NitigyaS: I was confused in this one, as except PostgreSQL all are using this from here only.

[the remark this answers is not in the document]
snopoke: couchdb2 has it in deploy_couchdb2. The rest are in here because they don't have their own playbooks

environments/development/postgresql.yml
snopoke: yes

src/commcare_cloud/ansible/roles/citusdb/meta/main.yml
snopoke: ok

src/commcare_cloud/ansible/roles/citusdb/meta/main.yml
snopoke: OK, nice find. I had this issue when testing on bionic but never got to the bottom of it.

NitigyaS:
@snopoke can we merge this ?

src/commcare_cloud/ansible/roles/postgresql_base/defaults/main.yml
```diff
@@ -36,4 +36,6 @@ allow_direct_postgresql_access: no
 
 postgresql_apt_repo: "deb http://apt.postgresql.org/pub/repos/apt/ {{ ansible_lsb.codename}}-pgdg main"
 postgresql_apt_key_url: "https://www.postgresql.org/media/keys/ACCC4CF8.asc"
-postgresql_binary: "postgresql-{{ postgresql_version }}"
\ No newline at end of file
+postgresql_binary: "postgresql-{{ postgresql_version }}"
+
+postgresql_shared_preload_libraries: 'pg_stat_statements'
```
snopoke: I think this should be a list since you're using the 'join' filter

src/commcare_cloud/ansible/roles/citusdb/tasks/setup_worker.yml
```diff
@@ -0,0 +1,16 @@
...
+    name: "{{ item.name }}"
+    state: present
+    port: "{{ postgresql_port }}"
+    owner: "{{ item.user }}"
+    encoding: 'UTF-8'
+    lc_collate: 'en_US.UTF-8'
+    lc_ctype: 'en_US.UTF-8'
+    template: 'template0'
+  when: item.create and ((item.host in groups.citusdb_cordinator) and (inventory_hostname in groups.citusdb_worker ))
+  with_items: "{{ postgresql_dbs.all }}"
```
snopoke: you also need to create the `citus` extension on the worker databases

Vagrantfile
```diff
@@ -10,14 +10,18 @@ require_relative './provisioning/key_authorization'
...
-  config.vm.box = "ubuntu/trusty64"
+  config.vm.box = "ubuntu/bionic64"
   cchq_proxy_port = ENV.fetch("VAGRANT_CCHQ_PROXY_PORT", 8080)
   config.ssh.insert_key = false
   authorize_key config, '~/.vagrant.d/insecure_private_key'
 
   {
     'app1'    => '192.168.33.15',
     'db1'   => '192.168.33.16',
+    'citusdb0' => '192.168.33.30',
```
snopoke: unused

snopoke:
Also I think you need to merge with master

snopoke:
hmm, actually I see this is a PR into the other branch, then no need to merge master

src/commcare_cloud/ansible/roles/citusdb/tasks/setup_cordinator.yml
```diff
@@ -0,0 +1,14 @@
+---
+ - name: Add Workers Nodes to the Cordinator
+   become: yes
+   become_user: postgres
+   shell: psql {{ item.0.name }} -c "SELECT * from master_add_node( '{{item.1|ipaddr}}' , 5432) ; "
+  #  debug:
+  #    msg: " {{ item.0.name }} {{item.1|ipaddr}} "
```
snopoke: debug

PR #2701, opened 2019-02-28
Just in case: Revert "use python 3 for staging"

nickpell:
Reverts dimagi/commcare-cloud#2697

If staging needs to be reverted to python 2 and I am unreachable, the steps to do it are:

1) Merge this PR
2) Run ```cchq staging update-supervisor-confs```
3) Run ```cchq staging fab restart_services```
4) Run ```./manage.py flush_caches```.  This is probably not needed, but may avoid caching errors.  I'm going to do additional work to make this step unnecessary when transitioning between python versions.


nickpell:
Merging this so as to avoid interference with pending QA

PR #2713, opened 2019-03-12
Undo split prod webworkers

millerdev:
This is an attempt to fix the drop-off in form submission success rate. We're not sure why the split would have caused the drop, but at this point we're grasping at straws.

Waiting for approval before applying this change to production.

##### SUMMARY
Attempt to fix https://dimagi-dev.atlassian.net/browse/PLAT-103

##### ENVIRONMENTS AFFECTED
production

##### ISSUE TYPE
- Change Pull Request

@dannyroberts

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -22,20 +22,13 @@ proxy0
 [web9]
 10.202.10.37 hostname=web9-production ec2=yes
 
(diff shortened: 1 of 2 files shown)


millerdev:
Deployed to production immediately after merging.

PR #2716, opened 2019-03-13
Use ddtrace-run with celery

millerdev:
##### SUMMARY
Run celery processes with [`ddtrace-run`](https://docs.datadoghq.com/tracing/languages/python/)

Probable fix for https://sentry.io/organizations/dimagi/issues/918618547, based on absence of errors in trace-agent.log after deploying to celery0-production.

Will deploy to other celery machines after review/approval.

##### ENVIRONMENTS AFFECTED

production

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

commcarehq ansible role

@snopoke @dannyroberts

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -2,6 +2,7 @@ newrelic_djangoagent: False
 newrelic_javaagent: True
 datadog_pythonagent: True
 django_command_prefix: '{{ virtualenv_home }}/bin/ddtrace-run '
+celery_command_prefix: '{{ virtualenv_home }}/bin/ddtrace-run '
 gunicorn_workers_static_factor: 0
 gunicorn_workers_factor: 2
 celery_processes:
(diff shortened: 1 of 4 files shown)


millerdev:
Looks like there have been no new events on https://sentry.io/organizations/dimagi/issues/918618547/ since I deployed this to celery0 today at about 21:42 UTC, so I'm declaring this a success.

However, this may increase our costs for Datadog APM because it will not be running on celery machines in addition to web workers. I'm not sure how much difference this will make cost-wise, but APM continues to be high value and I would vote to keep this solution.

An alternative solution would be to disable the datadog trace agent on celery machines so it doesn't even try to send them.

millerdev:
This was rolled out on production just now. Celery was restarted on all hosts except for `celery0`, which was done a few days ago.

PR #2719, opened 2019-03-18
improvements for production celery stability

nickpell:
Updates include:

1) Remove ```dd-trace``` from production.  Memory issues are reproduced when ```dd-trace``` is running and you run:
```
from corehq.apps.export.tasks import saved_exports
saved_exports.delay()
```
2) Remove ```maxtasksperchild``` from all workers in prefork.  This addresses https://github.com/celery/celery/issues/5120.  This has a risk of causing memory leaks if there are many tasks in the queue and the child process is never stopped.  However, I think this is more of a theoretical risk than a practical risk and, if it were to appear, could be addressed by adding more workers so queues don't get backlogged.

3) Runs ```background_queue``` in prefork.  I switched it to run in gevent to avoid https://github.com/celery/celery/issues/5120, but this caused the bug: https://github.com/celery/celery/issues/3377.  

🐡 

After letting these changes run in production for 24 hours and if no new issues are observed, I'll remove ```maxtasksperchild``` from other environments as well.

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -2,7 +2,7 @@ newrelic_djangoagent: False
 newrelic_javaagent: True
 datadog_pythonagent: True
 django_command_prefix: '{{ virtualenv_home }}/bin/ddtrace-run '
-celery_command_prefix: '{{ virtualenv_home }}/bin/ddtrace-run '
+celery_command_prefix: ''
 gunicorn_workers_static_factor: 0
 gunicorn_workers_factor: 2
 celery_processes:
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -18,12 +18,10 @@ celery_processes:
       concurrency: 50
     ucr_queue:
       concurrency: 4
-      max_tasks_per_child: 5
     ucr_indicator_queue:
       concurrency: 2
     async_restore_queue:
       concurrency: 2
-      max_tasks_per_child: 5
   celery1:
     flower: {}
     reminder_queue:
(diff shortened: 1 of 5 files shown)


environments/production/app-processes.yml
```diff
@@ -36,14 +34,12 @@ celery_processes:
       num_workers: 3
     case_rule_queue:
       concurrency: 4
-      max_tasks_per_child: 1
```
millerdev: A low value like this looks like it might have been set to deal with memory leaks or general bad behavior when multiple tasks are run. Seems like the ones that were `1` may be worth keeping an eye on.

nickpell:
> Looks like dd-trace was recently added to celery. Just wanted to make sure you're aware of the history and are coordinating: #2716.

@millerdev and I offlined about this and agreed to test dd-trace once celery 4 is more stable.

> I think there's a big difference between, say, multiplying the number by 10 and removing it entirely so that the celery workers just never restart.

As currently used the celery workers should stop the child processes relatively frequently, since most of the time there are no tasks in the queue.  It's unclear to me how long a child process would have to be running in order to accumulate enough memory to cause a problem.  I plan on monitoring this closely and giving Interrupt instructions on what to look for during IST working hours.

environments/production/app-processes.yml
nickpell: 👍

nickpell:
Github is stale but travis test passed:
https://travis-ci.org/dimagi/commcare-cloud/builds/507885284?utm_source=github_status&utm_medium=notification

PR #2730, opened 2019-03-20
double num_workers for async_restore_queue on celery0 and celery2

nickpell:
https://github.com/dimagi/commcare-cloud/pull/2724 helped prevent ```async_restore_queue``` from getting backlogged.  I'm making this change to provide a little more buffer, and because these two machines are still underutilized.

code buddy @mkangia

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -21,7 +21,7 @@ celery_processes:
     ucr_indicator_queue:
       concurrency: 2
     async_restore_queue:
-      num_workers: 2
+      num_workers: 4
       concurrency: 2
   celery1:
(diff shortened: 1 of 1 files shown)


kaapstorm:
> these two machines are still underutilized.

When I recently bumped up the resources allocated to the repeat record queue, I used Datadog to try to get an idea of how the Celery machines are untilized. But I had to do some guessing.

Did you quantify this, @nickpell? And if so, how did you do it?

nickpell:
@kaapstorm I'm just looking at % CPU in datadog and cautiously increasing the number of workers.  Not doing anything scientific here.

PR #2733, opened 2019-03-21
Record commcare-cloud command events in datadog

millerdev:
##### SUMMARY

Send an event to datadog when a commcare-cloud command is run (if it loads the ansible vault). This will provide a basic audit trail of commcare-cloud changes applied to environments with datadog monitoring.

##### ENVIRONMENTS AFFECTED

All environments with datadog enabled and having proper datadog credentials in the `secrets` section of the corresponding ansible vault.

The following have `DATADOG_ENABLED: True` in *public.yml*:

- production
- icds
- india/softlayer
- swiss
- pna
- staging
- 64-test?

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

Events are only sent when ansible vault variables are loaded since some of those variables are needed to send the event. It would be possible to always prompt for the vault password for every commcare-cloud command, and we may want to consider that for some additional commands in the future, but I opted not to impose any extra inconvenience at this time for commands that do not load the vault.

In most cases the event will not be sent during the "check" phase of a two-phase ansible run. In this case it will not be sent until the second "apply" phase is initiated. However, there are a couple commands (couchdb migrations and the `copy-files` command) that have a check phase where this does not apply—the event will be sent immediately the first time the vault is loaded for these commands.

Note that this mechanism provides an audit trail based on trust in the person running the command. It is trivially simple to disable the code that sends the event before running a command. It is also possible to cancel the command before it has any effect, so the presence of an event in datadog is not a guarantee that the command ran to successful completion. Secure verification of commcare-cloud command execution is not the intent nor in scope for this change.

Review by commit 🐟 

@dannyroberts @snopoke

### DIFF
--- a/src/commcare_cloud/environment/main.py
+++ b/src/commcare_cloud/environment/main.py
@@ -68,6 +68,8 @@ class Environment(object):
                 with open(self.paths.vault_yml, 'r') as vf:
                     return vault.load(vf.read())
             except AnsibleVaultError:
+                if os.environ.get('ANSIBLE_VAULT_PASSWORD'):
+                    raise
                 print('incorrect password')
                 self.get_ansible_vault_password.reset_cache(self)
 
--- a/src/commcare_cloud/environment/main.py
+++ b/src/commcare_cloud/environment/main.py
@@ -28,6 +28,10 @@ class Environment(object):
     def __init__(self, paths):
         self.paths = paths
 
+    @property
+    def name(self):
+        return self.paths.env_name
+
     def check(self):
 
         included_disallowed_public_variables = set(self.public_vars.keys()) & self._disallowed_public_variables
@@ -47,7 +51,7 @@ class Environment(object):
     def get_ansible_vault_password(self):
         return (
             os.environ.get('ANSIBLE_VAULT_PASSWORD') or
-            getpass.getpass("Vault Password for '{}': ".format(self.paths.env_name))
+            getpass.getpass("Vault Password for '{}': ".format(self.name))
         )
 
     @memoized
--- a/src/commcare_cloud/environment/main.py
+++ b/src/commcare_cloud/environment/main.py
@@ -1,21 +1,19 @@
 import getpass
 import os
+from collections import Counter
 
 import yaml
+from ansible.inventory.manager import InventoryManager
+from ansible.parsing.dataloader import DataLoader
 from ansible.parsing.vault import AnsibleVaultError
+from ansible.vars.manager import VariableManager
 from ansible_vault import Vault
 from memoized import memoized, memoized_property
-from collections import Counter
 
 from commcare_cloud.environment.constants import constants
 from commcare_cloud.environment.exceptions import EnvironmentException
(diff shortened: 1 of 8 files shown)


src/commcare_cloud/environment/main.py
```diff
@@ -43,15 +50,35 @@ def check(self):
...
+        the commcare-cloud command that is currently being run.
+        """
+        self.get_vault_variables()
+        return self._get_ansible_vault_password()
+
+    def get_vault_variables(self):
+        """Get ansible vault variables
+
+        This method has a side-effect: it records a Datadog event with
+        the commcare-cloud command that is currently being run.
```
millerdev: I took the side-effect approach because it seemed easier and cleaner than sprinkling `record_command_event` calls around in various places near where vault variables or the ansible password are loaded. An argument could be made that this is too sneaky and hacky. I'm open to feedback or other ideas.

src/commcare_cloud/environment/main.py
```diff
@@ -324,6 +387,13 @@ def get_hostname(self, sshable_hostname, full=True):
         return hostname
 
 
+def setup_environment(env_name, input_argv):
```
dannyroberts: Maybe it'd be better to call this `make_argv_available_to_environment`. A name like that more clearly signals that this has a special, possibly hacky, purpose and that it's not a good place to highjack for your own use. In this case, `input_argv` is fundamentally an optional piece of info, used only for this ancillary feature; it'd probably be a bad idea to hijack this setup function to inject fundamentally necessary properties. A more what-this-does name would signal to someone who wanted to repurpose this entry point that it'd be up to them to make it a more general system, and rename accordingly, if they did want to go that route.

src/commcare_cloud/environment/main.py
```diff
@@ -77,6 +106,40 @@ def get_vault_var(self, var):
...
+
+    @contextmanager
+    def suppress_vault_loaded_event(self):
+        """Prevent "run event" from being sent to datadog
+
+        This is only effective if `self.get_vault_variables()` has not
+        yet been called outside of this context manager. If it has been
+        called then the event has already been sent and this is a no-op.
+        """
+        sent = self.did_send_vault_loaded_event
```
dannyroberts: I think in the first commit, this was a good name for the variable, but now it's being used for a connected but distinct meaning—with this suppression functionality, it might be better called `should_send_vault_loaded_event` (with the inverse value).

src/commcare_cloud/environment/main.py
millerdev: https://github.com/dimagi/commcare-cloud/pull/2733/commits/abdf3890ae99b9c5358edec015a3f4e76f94eac8 - decided to use `sys.argv` directly rather than passing `input_arv` around.

src/commcare_cloud/environment/main.py
millerdev: 1a642eacd7d314575f47bdea60e359369361726d

PR #2734, opened 2019-03-21
Use new timezone setting

emord:
##### SUMMARY

They removed (accidentally?) their support form CELERY_TIMEZONE because software is hard and we all make mistakes

code: https://github.com/celery/celery/blob/4.1/celery/app/utils.py#L122
(wrong) documentation: http://docs.celeryproject.org/en/latest/userguide/configuration.html?highlight=timezone#new-lowercase-settings

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
celery


emord:
Actually I think this may not be a good change. thoughseeing some strange behavior with beat

emord:
This should handle the changes https://github.com/dimagi/commcare-hq/blob/master/corehq/celery.py. have not figured out why we'renot seeing the correct timezone

PR #2742, opened 2019-03-26
double concurrency of saved_export_queue workers in production

nickpell:
Continuing to increase resources for saved_export_queue.  Ideally I would like this task to be scheduled for completion by midnight EST so if there are issues, I'll be awake to debug them.

Current machine utilization:

https://app.datadoghq.com/dash/host/665473534?live=false&tile_size=m&page=0&is_auto=false&from_ts=1553557381993&to_ts=1553606775000

https://app.datadoghq.com/dash/host/665473537?live=false&tile_size=m&page=0&is_auto=false&from_ts=1553557866139&to_ts=1553606806000

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -68,14 +68,14 @@ celery_processes:
       concurrency: 50
     saved_exports_queue:
       num_workers: 3
-      concurrency: 2
+      concurrency: 4
       optimize: True
   celery3:
     async_restore_queue:
       concurrency: 8
     saved_exports_queue:
       num_workers: 3
-      concurrency: 2
+      concurrency: 4
       optimize: True
     export_download_queue:
       concurrency: 4


environments/production/app-processes.yml
```diff
@@ -68,14 +68,14 @@ celery_processes:
       concurrency: 50
     saved_exports_queue:
       num_workers: 3
-      concurrency: 2
+      concurrency: 4
```
mkangia: i would usually vote for increasing the workers just because i find that more effective w/o any supporting facts.
What's your quick thought on increasing workers vs increasing concurrency here?
note: i assume that for bigger saved exports the process would consume higher memory.

nickpell: There was some chat about this in #celery.  The conclusion seemed to be that increasing concurrency provides the least memory overhead.  It does seem possible that it would have less redundancy, though, if a child process issue brings down a worker.

PR #2747, opened 2019-04-01
Added 1 Celery server with Ubuntu 18

NitigyaS:
##### SUMMARY

Add new server with bionic image on ICDS
replace celery 12

##### ISSUE TYPE

- Change Pull Request


environments/icds-cas/app-processes.yml
```diff
@@ -189,6 +189,7 @@ celery_processes:
       num_workers: 6
     icds_dashboard_reports_queue:
       concurrency: 2
+  'celery112':
```
snopoke: `112` - is that intentional?

NitigyaS: Yes.

NitigyaS: It is done to avoid any conflict with existing machine , although in this case celery12 was already out of cluster but going forward I am planning to proceed like this only.

PR #2750, opened 2019-04-02
Resize prod couch disks to target 65% full

dannyroberts:
##### SUMMARY

We've been getting oscillating disk usage warnings for about a month now, and some of the disks really are uncomfortably full. I did out what it would take for each disk to get down to 65% full (at their daily peak) and came up with these numbers. Will cost an extra $930/mo or so, but it puts us in a safer place and will make the disk warnings a stronger signal.

  | current max (%) | current disk (GB) | target max (%) | new disk (GB) | increase (GB)
-- | -- | -- | -- | -- | --
couch3 | 82 | 5500 | 65 | 6938 | 1438
couch4 | 80 | 5500 | 65 | 6769 | 1269
couch5 | 83 | 5500 | 65 | 7023 | 1523
couch7 | 77 | 6100 | 65 | 7226 | 1126
couch8 | 86 | 5800 | 65 | 7674 | 1874
couch9 | 89 | 5500 | 65 | 7531 | 2031
  | **current spend** | $3,390/mo |   | **increase spend** | $926/mo

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -142,7 +142,7 @@ servers:
     az: "a"
     volume_size: 80
     block_device:
-      volume_size: 5500
+      volume_size: 5857
     group: "couchdb2"
   - server_name: "couch4-production"
     server_instance_type: c5.4xlarge
@@ -150,7 +150,7 @@ servers:
     az: "a"
     volume_size: 80
     block_device:
-      volume_size: 5500
+      volume_size: 5714
     group: "couchdb2"
   - server_name: "couch5-production"
     server_instance_type: c5.4xlarge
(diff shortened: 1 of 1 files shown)


dannyroberts:
had good discussion with cal on slack #interrupt-team today, and am replacing this with a PR that gives us greater buffer and holds the line for alerting on disk usage at 80%.

dannyroberts:
If we were to maintain a 4% growth rate of daily peak, we would start hitting 80% disk warnings in 6 months. This gives us a nice long break from having to worry about increasing couch disk. A the same time, we're currently executing on moving more out of couch and into SQL so within the quarter this number should start going down. Since our goal is to be able to start _downsizing_ the cluster in the next 6 months, this addition to the run-rate will hopefully be short-lived.

PR #2752, opened 2019-04-04
Added Celery112 in app process

NitigyaS:
##### SUMMARY

Add process in new celery112

##### ISSUE TYPE

- Feature Pull Request


NitigyaS:
There are 2 reasons.
1) when TCL upgraded the C012 they created new machine to C112. So I also followed the same convention. 
2) I though it would be easier to know which celery machines have been upgraded by simply adding using range in 1XX

PR #2753, opened 2019-04-04
accounting should never use standby DBs

nickpell:
https://dimagi-dev.atlassian.net/browse/HI-125

Accounting should never use the standby servers, since accounting often assumes a ```.save()``` operation to have effect immediately after it is called.

We now cache the current subscription, which used to be the largest source of accounting load by a lot.


nickpell:
> Could just delete accounting block entirely.

Thanks for pointing that out.  It's updated.  @emord

nickpell:
What do I need to run to apply this change?

emord:
deploy localsettings and restart services (sync with cal before doing this as he's currently fighting dashboard celery

PR #2757, opened 2019-04-05
add rabbitmq config with warning level logging

snopoke:
##### SUMMARY
I think upgrading to Celery4 has increased the number of connections and disconnections occurring on RabbitMQ which has greatly increased the logging and is filling up the disks.

##### ENVIRONMENTS AFFECTED
all (optional)


snopoke:
It may be worth checking if Celery4 has some connection pooling that can be used as well.

PR #2760, opened 2019-04-05
Add example script for resizing disk on aws

dannyroberts:
I had put this together for a recent bit of maintenance and figured it was worth saving somewhere.


dannyroberts:
Killed the build since this isn't really a code change

PR #2778, opened 2019-04-15
add packages for weasyprint

calellowitz:
@Rohit25negi @biyeun @emord 
This adds two libraries needed for weasyprint, which is added in https://github.com/dimagi/commcare-hq/pull/23707/files


calellowitz:
That is a good point. Will post a message there, although I believe the only things that break if you don't install those are the ISSNIP monthly registers, which should not be super relevant (before installing was able to run hq locally and navigate fine)

PR #2780, opened 2019-04-16
icds rate limit

snopoke:
Update ICDS rate limiting.


snopoke:
That's correct. I don't know what % are on the new version but with the previous limits we weren't rate limiting at all.

PR #2781, opened 2019-04-16
Add celery6-production

dannyroberts:
##### SUMMARY
Add redundant processors for the 4 queues that have exceeded their
blockage threshold in the last week:
- celery
- case_import_queue
- case_rule_queue
- async_restore_queue

In addition make sure each of those has their own processor on both
machines, and change concurrency/num_workers to be split evenly across
the two machines and a little higher than previously (but not twice).

For async_restore_queue, which is already on 3 machines, leave in place but convert to have each node use a single worker with higher concurrency and distribute evenly across the 3 nodes.

##### ENVIRONMENTS AFFECTED
production

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
Celery


environments/production/known_hosts
```diff
@@ -28,12 +28,12 @@
 10.202.10.238,10.202.10.238 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBFPb3UG0OYM+f1wa1uIa0EyXcrJAUeBRWNupik11oyiZqDoMnlNodI4eoXyRDywNKCnpDB0Cq6k48DJWvr+VZyg=
 10.202.10.238,10.202.10.238 ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIL5Ufct69wD1ZIzlt8yEC45kuv6/eqBIbDaQu3Wlr/LR
 10.202.10.238,10.202.10.238 ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDAP4W5JFl2/KyWsWahW7+NfL9Z7rZUUfSjxMkeIqwGRTzBR3OXOp055GmU7IL9QrxR1JnhQ6sBD6keZ6UMohU0bnmNjHGsSDp5M3PuvPzP9RxMPQ9SHdMZGIsy4kGhV1V8kNE+gM34NmBvlP9e+fQ1XAICt2BCeAmv33fkHY+9C5SdeYqrLDVtWGBo8Z6oHQsQTTcQHMSNH8KEwPV/zrNo74095AOtrD7Xx65iromxLbC4G0JUgFvNMOvexn+6Q6CfgKQ3siVIUJ/9NhY6dCmlEDZB61cREn/YU7GknceRc4UQJK0KBQn6joZU9tatUhTExQJdwSXpDyH698Wd+1LV
-10.202.10.4, ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBHZvgnRBtcOkDc3GwJOMPsnfhclZZ5L7QApctnjvYn61Xn/hm8Bps0iF7Fcp06HNDD0St6Ka96BT40sXvRcUdb0=
```
nickpell: Any particular reason for this re-ordering?

environments/production/app-processes.yml
```diff
@@ -97,6 +99,16 @@ celery_processes:
...
       num_workers: 2
+  celery6:
+    celery:
+      concurrency: 4
+    case_import_queue:
+      concurrency: 4
+    case_rule_queue:
+      concurrency: 4
+    async_restore_queue:
+      num_workers: 3
```
nickpell: Is there any particular reason to not make this ```concurrency: 6``` with default ```num_workers: 1```?  It seemed from the #celery conversation that this configuration would result in fewer ```MainProcess``` instances and therefore less memory overhead.  Or do we think having more ```MainProcess``` instances gives us some extra form of redundancy?

environments/production/known_hosts
dannyroberts: That's just what happened when I ran `cchq production update-local-known-hosts --limit celery6`. At the end of that run, the file is sorted so that there's a standard format (to minimize diffs, ironically in this particular case).

I'm not totally sure how these might have gotten out of order, possibly someone adding 10.202.10.4 in a more manual way

environments/production/app-processes.yml
dannyroberts: You're right, there's no particular reason I chose to do it this way—I was just copying from the existing definition. I'll go with your suggestion

PR #2784, opened 2019-04-17
Fix private interface for enhanced network enabled AWS instances

dannyroberts:
##### SUMMARY

The bionic image on AWS has AWS Enhanced Networking enabled, which means among other things it calls the private interface `ens5` instead of `eth0`

##### ENVIRONMENTS AFFECTED
AWS envs using bionic (staging, production)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Bionic

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -122,7 +122,7 @@ kafka0
 10.202.12.120 hostname=celery5-production ec2=yes
 
 [celery6]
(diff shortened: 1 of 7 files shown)


dannyroberts:
I tried to standardize these settings across envs as well. I think it will be useful to refactor so that this is more automatic and DRY, but I'll leave that for another PR

PR #2785, opened 2019-04-17
Fix s3transfer version conflict by setting a lower bound to boto3

dannyroberts:
##### SUMMARY
I noticed this was causing a version conflict locally, including on a fresh install.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,8 +11,8 @@ args==0.1.0               # via clint
 asn1crypto==0.24.0        # via cryptography
 attrs==18.2.0
 bcrypt==3.1.4             # via paramiko
-boto3==1.9.38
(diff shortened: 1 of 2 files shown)


emord:
Can you give more information on how this was causing a versioning conflict? Is there an issue with pip-tools that make upgrade-requirements not resolve dependencies correctly?

dannyroberts:
Offlined with @emord.

Basically the original issue here is that if you run

```
pip install -r requirements.txt
pip install -e .[aws]  # this just installs awscli on top of the other requirements
```
which is what you need to do to use any of the aws-specific functionality, you get the version conflict. Running `make upgrade-requirements` fixes that, but unfortunately my recent attempt to do that causes test failures I haven't gotten to the bottom of, so this PR is taking a more targeted approach.

PR #2789, opened 2019-04-18
update redis config

snopoke:
update max memory and eviction policy

FYI @dimagi/scale-team

### DIFF
--- a/environments/icds-cas/public.yml
+++ b/environments/icds-cas/public.yml
@@ -28,6 +28,8 @@ swift_url: https://mowcdmum.ipstorage.tatacommunications.com/auth/v1.0/
 nofile_limit: 65536
 
 redis_appendonly: 'yes'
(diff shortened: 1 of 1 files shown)


environments/icds-cas/public.yml
```diff
@@ -28,6 +28,8 @@ swift_url: https://mowcdmum.ipstorage.tatacommunications.com/auth/v1.0/
 nofile_limit: 65536
 
 redis_appendonly: 'yes'
+redis_maxmemory: 6gb
```
emord: are there any catches in ansible that make sure this is < 1/2 the system's memory?

dannyroberts: I searched our code as well as the redis galaxy module and I don't see anything of the sort; it's only mentioned a couple times in code, and only to interpolate it in a template.

PR #2791, opened 2019-04-18
Apply pg_stat_statement settings from other environments to pgmain0-production

dannyroberts:
##### SUMMARY

Since RDS is managed through terraform, we have to set these settings separately there. For now just setting on the main db, but we can of course turn it on for others as well.

This is already applied.

##### ENVIRONMENTS AFFECTED
production

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME

RDS

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -308,6 +308,11 @@ rds_instances:
     instance_type: "db.t2.xlarge"
     storage: 500
     multi_az: true
+    params:
+      pg_stat_statements.track: all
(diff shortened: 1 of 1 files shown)


calellowitz:
@dannyroberts 
Why are we not applying this to all DBs?

dannyroberts:
We should, I was just testing it on the one out of cautiousness, and then didn't follow through. Thanks for the reminder, I'll roll it out to all of them

PR #2792, opened 2019-04-19
Creating redis1,es3 and control servers in staging environment

sanjay2916:
Creating redis and es3 servers in staging environment which has bionic os


sanjay2916:
changes are not proper and creating new PR for this

Issue #2571, opened 2019-01-17
update-supervisor-confs should should do reread/update without fab


dannyroberts:
Doing `supervisorctl reread` and `supervisorctl update` through fab causes problems and is also just too chatty. Can use the ansible module or just the shell module instead.
```
Would you like to update supervisor to use the new configurations? [y/N]y
fab -f /home/sanjay/commcare-cloud/src/commcare_cloud/fabfile.py production 'supervisorctl:"reread"'
code_branch not specified, using 'master'. You can override this with '--set code_branch=<branch>'
[10.202.10.233] Executing task '_confirm_deploying_same_code'
You can add a config file to automate this step:
    $ cp /home/sanjay/commcare-cloud/src/commcare_cloud/fab/config.example.py /home/sanjay/commcare-cloud/src/commcare_cloud/fab/config.py
Then edit /home/sanjay/commcare-cloud/src/commcare_cloud/fab/config.py
Github username (leave blank for no auth): sanjay2916
```

PR #2795, opened 2019-04-19
Streamline reloading supervisor in update-supervisor-confs

dannyroberts:
##### SUMMARY
run commands through ansible shell instead of through fab

Also now the reload also respects the `--limit` arg

Fixes https://github.com/dimagi/commcare-cloud/issues/2571

##### ENVIRONMENTS AFFECTED
None

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
`update-supervisor-confs` command

### DIFF
--- a/src/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/src/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -301,8 +301,13 @@ class UpdateSupervisorConfs(_AnsiblePlaybookAlias):
         unknown_args += ('--tags=services',)
         rc = AnsiblePlaybook(self.parser).run(args, unknown_args)
         if ask("Would you like to update supervisor to use the new configurations?"):
-            exec_fab_command(args.env_name, 'supervisorctl:"reread"')
-            exec_fab_command(args.env_name, 'supervisorctl:"update"')
+            carryover_args = []
+            if args.limit:
+                carryover_args.extend(['--limit', args.limit])
+            commcare_cloud(
+                args.env_name, 'run-shell-command', 'webworkers:celery:pillowtop:formplayer',
+                'supervisorctl reread; supervisorctl update', '-b', *carryover_args
+            )
         else:
             return rc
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -301,8 +301,13 @@ def run(self, args, unknown_args):
         unknown_args += ('--tags=services',)
         rc = AnsiblePlaybook(self.parser).run(args, unknown_args)
         if ask("Would you like to update supervisor to use the new configurations?"):
-            exec_fab_command(args.env_name, 'supervisorctl:"reread"')
-            exec_fab_command(args.env_name, 'supervisorctl:"update"')
+            carryover_args = []
+            if args.limit:
+                carryover_args.extend(['--limit', args.limit])
+            commcare_cloud(
+                args.env_name, 'run-shell-command', 'webworkers:celery:pillowtop:formplayer',
```
emord: Short term: should this include the proxy and airflow? I think the websockets process runs on the proxy.

Long term: should this list of hosts be defined elsewhere. if there are any other additions to our groups, this may be a hard place to know to update

dannyroberts: Good points.

The full list being used by fab is https://github.com/dimagi/commcare-cloud/blob/a0e74dde3231c61ea4ad351e475ba837257155af/src/commcare_cloud/fab/const.py#L9-L18, which doesn't seem to include proxy, but that doesn't mean that's right. I'll make some changes in a followup PR

emord: staticfiles is the proxy https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/fab/fabfile.py#L296

dannyroberts: Oh right, thanks

PR #2797, opened 2019-04-19
Nic cleanup

czue:
##### SUMMARY

Reconcile NIC inventory file with what exists today.

##### ENVIRONMENTS AFFECTED

Legacy NIC environment

##### ISSUE TYPE

- Change Pull Request

##### COMPONENT NAME

n/a

##### ADDITIONAL INFORMATION

not sure if the build will pass on this, but it makes it easier to run shell commands against NIC and not have to wade through a bunch of errors related to missing/dead servers

### DIFF
--- a/environments/icds-new/app-processes.yml
+++ b/environments/icds-new/app-processes.yml
@@ -151,15 +151,6 @@ pillows:
       start_process: 4
       num_processes: 4
       total_processes: 36
-  'pillow2':
-    kafka-ucr-static-cases:
-      start_process: 12
-      num_processes: 6
-      total_processes: 36
-    CaseToElasticsearchPillow:
-      start_process: 8
-      num_processes: 4
-      total_processes: 36
   'pillow3':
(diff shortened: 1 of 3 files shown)


mkangia:
There are three environments for icds as of now in repo,
Aren't two of those including icds-new obsolete?
@NitigyaS

NitigyaS:
@mkangia `icds` and `icds-cas` are alias of same  and they point to new environement, `icds-new` is actually the older NIC one. 
this PR is for NIC's inventory modification.

mkangia:
I would leave it to @NitigyaS for review since he has context on this and i think i need some more context on icds environments so i will offline with him.

czue:
Yeah @mkangia this env is obsolete but since much of the infrastructure is still online and in use by the data team it's still useful to have up to date inventory.

mkangia:
Thank @czue , i didn't know it was being used by the data team.

PR #2758, opened 2019-04-05
Revert "Added 1 Celery server with Ubuntu 18"

calellowitz:
Reverts dimagi/commcare-cloud#2747
FYI: @NitigyaS 
@emord @snopoke 
This machine is improperly configured and is blocking deploy. given that it isn't doing anything, seems easier to revert this until someone can look into it

PR #2799, opened 2019-04-22
Adding celery112 to ICDS cluster

sanjay2916:
Adding celery112 to ICDS cluster


calellowitz:
@sanjay2916 
The original issue that caused this machine to be removed in https://github.com/dimagi/commcare-cloud/pull/2758/files was never fixed. I believe I fixed it now, but in the future, please address the underlying issues before undoing a revert. ICDS deploy was blocked for a few hours today before the problem was figured out.

PR #2800, opened 2019-04-22
Remove old Elasticsearch and Redis servers in staging environment

sanjay2916:
Remove old Elasticsearch and Redis servers in the staging environment

### DIFF
--- a/environments/staging/inventory.ini.j2
+++ b/environments/staging/inventory.ini.j2
@@ -122,9 +122,9 @@ redis1
 # Amazon EC2
 airflow1
 
-[control0]
+[control]
 # Amazon EC2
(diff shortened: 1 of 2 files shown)


environments/staging/inventory.ini.j2
```diff
@@ -135,10 +122,6 @@ redis1
 # Amazon EC2
 airflow1
 
-[control]
-# Amazon EC2
-{{ control-staging }} hostname=control-staging ec2=yes
-
 [control0]
```
snopoke: should rename this to `control`

sanjay2916: We will do Simon

PR #2812, opened 2019-04-26
Make inventory template dryer

dannyroberts:
##### SUMMARY
This is motivated by the new bionic vars that we're needing, but also the amount of copy-paste-edit-edit-edit we've been doing to maintain the template.

##### ENVIRONMENTS AFFECTED
production, staging (and to a tiny extent india)

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
inventory / aws-fill-inventory

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -1,5 +1,5 @@
 [proxy0]
-10.202.20.138 hostname=proxy0-production public_ip=100.24.216.221 ec2=yes
+10.202.20.138 hostname=proxy0-production ec2=yes public_ip=100.24.216.221
 
 [proxy:children]
 proxy0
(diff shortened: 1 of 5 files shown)


dannyroberts:
@dimagi/team-commcare-hq heads up y'all!

On the AWS envs, which are *production* and *staging* (and soon to be india), the format for `inventory.ini.j2` has changed to require less repetition.

Where previously you might have had something like
```
[web6]
{{ web6-production }} hostname=web6-production ec2=yes
```

you will now have

```
{{ __web6__ }}
```

which will produce exactly the same output in `inventory.ini`. This new format detects whether the host is an Ubuntu Trusty Taur (14.04) instance, an Ubuntu Bionic Beaver (18.04) instance, or an RDS instance and adjusts the default host variables (above, `hostname=web6-production` and `ec2=yes`) accordingly.

PR #2815, opened 2019-04-26
Add web4-staging

dannyroberts:
##### SUMMARY
This required making some changes, which I've also included in the PR

##### ENVIRONMENTS AFFECTED
staging


src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml
```diff
@@ -38,6 +38,8 @@
 - name: register git repo
   stat: path="{{ code_home }}/.git"
   register: gitdir
+  tags:
```
nickpell: Could you explain why this is needed?

dannyroberts: Yeah one of the `tags: py3` tasks below checks `gitdir`. In the end I realized running py3 alone isn't that useful (it can leave you in a state where the repo has been cloned but the `python_env-3.6/` dir is empty, which makes subsequent runs skip that whole part including installing pip packages, which leaves `python_env-3.6/` forever empty), but this still felt like a small improvement worth keeping

PR #2816, opened 2019-04-27
add new links

mkangia:
Changes made on proxy0.


mkangia:
@snopoke i manually made these changes in the `sites-enabled` under nginx.
that works right?
i am assuming that if someone runs an update for nginx conf, they would see no change since this is already done

PR #2819, opened 2019-04-29
riak cache

snopoke:
2nd commit is the change: 7b167a8

For context we are going to be migrating from Riak to TCL object storage (Swift) soon and thinking that having a cache in front of it will give us better performance. 

Implementing this now to test it out.


emord:
I think the main performance boost we want is the reads from pillows that happen very soon after write, correct? It would also be cool if we could cache it on the write, but that might be harder/impossible.

From my understanding, those shouldn't need to read images from the cache. Would it be possible to only cache XML? My main concern is that usage of the AWC report on the dashboard will kill the cache since it often includes many days of images and images are so much larger than xml.

snopoke:
@emord in the limited time I had I couldn't find a way to distinguish between a request for form XML and one for images. The request URLs are homogeneous like `http://.../blobs/[uuid]`. There's no file extension or anything.

I thought perhaps we could add a header but that doesn't seem possible with the `boto3` library.

Regarding caching the POST request body, that's not possible. We can cache the response but not the request body. There are other modules for nginx like the redis module which may work but we could as well do that from the application layer.

It looks like this might be possible with [Varnish](https://stackoverflow.com/a/37587478/632517)

PR #2821, opened 2019-04-29
Install ufw on bionic

dannyroberts:
##### SUMMARY
Friday when I was adding web4-staging I also had issues with `ufw` giving an opaque error on the first few passes, then I added this and run common_installs.yml and then the `ufw` commands worked. Not sure if it was this change or the re-running of some other part that made the difference, but I don't think this could hurt.


dannyroberts:
FYI @sanjay2916

dannyroberts:
Oops I didn't mean to merge this PR—I'll ask @nickpell to review

PR #2822, opened 2019-04-29
Fix deploy control by adding the host_group_aliases.yml boilerplate

dannyroberts:
##### SUMMARY

I stumbled upon this when I was bootstrapping the new staging control machine this morning

```
cchq staging ansible-playbook deploy_control.yml
```
now proceeds smoothly.

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -50,7 +50,13 @@ if [ -z ${TRAVIS_TEST} ]; then
     source virtualenvwrapper.sh
     if [ ! -d ~/.virtualenvs/ansible ]; then
         echo "Creating ansible virtualenv..."
-        mkvirtualenv ansible
+        mkvirtualenv ansible --python $(which python2)
+    # If we by mistake are using a py3 env (unsupported), then replace with a py2 env
+    elif ~/.virtualenvs/ansible/bin/python -c 'print ""' 2> /dev/null; [ "$?" -ne "0" ]; then
+        echo "Replacing ansible py3 virtualenv with py2 virtualenv..."
+        deactivate 2> /dev/null || :  # deactivate if in a virtualenv, else ignore error
+        rmvirtualenv ansible
+        mkvirtualenv ansible --python $(which python2)
     else
         workon ansible
     fi
--- a/src/commcare_cloud/ansible/deploy_control.yml
+++ b/src/commcare_cloud/ansible/deploy_control.yml
@@ -1,3 +1,5 @@
+- import_playbook: host_group_aliases.yml
+
 - name: Bootstrap Control
   hosts: "{{ target|default('control') }}"
   become: true


dannyroberts:
FYI @dimagi/team-commcare-hq @sanjay2916 if you ever find yourself needing to bootstrap a new control machine for an environment, you can do that with

```
cchq staging ansible-playbook deploy_control.yml
```

This initializes each user's home directory so that they'll be able to get started with using `commcare-cloud` on that control machine (including through running `cchq --control ...` locally) with no further steps.

As for how this just came up, I noticed this hadn't been run on the new staging control machine, so I couldn't use the new control right away and each new person had to set themselves up on it. I ran it now, so everyone should be able to continue their workflows as if the switchover to a new control machine hadn't happened.

dannyroberts:
Oh, dear. Looks like this set everything up with python3 instead of python2, which won't work yet for commcare-cloud. Looking into it, but for now if you have issues with staging control, this may be why

dannyroberts:
Should be fixed. If you have issues when logging into the control machine and you see any py3-related messages/errors, running `update-code` should be enough to reset your virtualenv and unblock you.

control/init.sh
```diff
@@ -50,7 +50,13 @@ if [ -z ${TRAVIS_TEST} ]; then
     source virtualenvwrapper.sh
     if [ ! -d ~/.virtualenvs/ansible ]; then
         echo "Creating ansible virtualenv..."
-        mkvirtualenv ansible
+        mkvirtualenv ansible --python $(which python2)
+    # If we by mistake are using a py3 env (unsupported), then replace with a py2 env
+    elif ~/.virtualenvs/ansible/bin/python -c 'print ""' 2> /dev/null; [ "$?" -ne "0" ]; then
+        echo "Replacing ansible py3 virtualenv with py2 virtualenv..."
+        deactivate 2> /dev/null || :  # deactivate if in a virtualenv, else ignore error
+        rmvirtualenv ansible
```
millerdev: This is a destructive operation, and this script is usually run automatically on login. This seems like it could cause great inconvenience and confusion if someone has purposefully configured their virtualenv with Python 3. What do you think of displaying a loud warning that things are unlikely to work rather than deleting their virtualenv and automatically setting up a new one?

dannyroberts: Currently a py3 virtualenv doesn't work at all for commcare-cloud, which is why I felt comfortable deleting it. Presumably we will work on this in the future, and at _that_ point, I could see this tripping up whoever's working on setting us up for the switch.

Seeing as I just added this for specific problem we were having where I'd inadvertently set up the new staging control with py3 virtualenvs and needed to reset that, this code's specific reason for existing is kind of over. So yeah I'm happy to replace it with a warning and call it a day.

dannyroberts: https://github.com/dimagi/commcare-cloud/pull/2832

PR #2825, opened 2019-04-30
Update proxy1-india machine size

dannyroberts:
I meant to get this in to the previous PR (https://github.com/dimagi/commcare-cloud/pull/2824)


dannyroberts:
I don't normally do this, but this seems uncontroversial enough that I'm going to just self-merge to keep master in line with what's deployed

PR #2826, opened 2019-04-30
Fix staging inventory template and render it

dannyroberts:
##### SUMMARY

Build wasn't failing when `aws-fill-inventory` fails because of a poorly formatted `inventory.ini.j2`; now it will. I also then fixed the staging inventory template and rendered it.


dannyroberts:
@sanjay2916 FYI

dannyroberts:
The error is on master, so I'm just going to merge this and fix that separately.

PR #2829, opened 2019-04-30
test no read to ucr master

calellowitz:
##### SUMMARY

Test PR to see if stopping reads from the main UCR db affects pillow throughput. I believe this was initially suggested by @snopoke. This was already rolled out for testing after a conversation with @ctsims 
FYI: @emord 

##### ENVIRONMENTS AFFECTED

ICDS

##### ISSUE TYPE

- Change Pull Request


environments/icds-cas/postgresql.yml
```diff
@@ -5,7 +5,6 @@ REPORTING_DATABASES:
   icds-ucr:
     WRITE: icds-ucr
     READ:
-      - [icds-ucr, 1]
```
emord: if all of these standbys are behind, would the dashboard and UCRs not load since there wouldn't be any relevant READ databases? Looks like the "acceptable delay" is 30 min at the moment. Not sure how close we get to that

PR #2831, opened 2019-05-01
Force travis to use Ubuntu trusty (14.04) to fix tests

dannyroberts:
I don't see it in the docs anywhere, but at some point today it looks
like Xenial (16.04) became the default, which started breaking tests.

The build error we've been having all day is:

```
TASK [supervisor : Add Upstart Conf] *******************************************
skipping: [localhost]
TASK [supervisor : Add Systemd Conf] *******************************************
skipping: [localhost]
TASK [supervisor : stop supervisor to pick up configuration changes] ***********
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Could not find the requested service supervisord: host"}
	to retry, use: --limit @/home/travis/build/dimagi/commcare-cloud/src/commcare_cloud/ansible/deploy_stack.retry
```


dannyroberts:
I cancelled the second build. The first build passed! Merging this since currently all builds fail

PR #2832, opened 2019-05-01
Print loud warning instead of auto replacing a py3 virtualenv

dannyroberts:
##### SUMMARY

Since after this we won't auto-replace py3 virtualenvs on the control machine, I ran manual cleanup on staging to remove all py3 envs. (Since I know that they were all created in error, rolled out by mistake and then discovered, while working on https://github.com/dimagi/commcare-cloud/pull/2822.)

To clean up I ran
```
cchq staging run-shell-command control 'ls /home/*/.virtualenvs/ansible/bin/python3 | sed "s|/bin/python3||" | xargs rm -rf' -b
```

and then I also ran

```
cchq staging run-shell-command control 'rm -rf /home/droberts/.virtualenvs/ansible' -b
cchq --control staging ping all
```

just to test the experience, and as I predicted and hoped, the script properly recreates the missing virtualenv with no complaint, so it's a seamless experience.

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -51,12 +51,18 @@ if [ -z ${TRAVIS_TEST} ]; then
     if [ ! -d ~/.virtualenvs/ansible ]; then
         echo "Creating ansible virtualenv..."
         mkvirtualenv ansible --python $(which python2)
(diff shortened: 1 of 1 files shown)


millerdev:
Thanks

PR #2836, opened 2019-05-02
add per IP rate limit

snopoke:
Adding this to try and combat the retry storm that we get
during peak hours on ICDS

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
@@ -21,6 +21,10 @@ nginx_sites:
       zone: "$server_name"
       size: "10m"
       rate: "200r/s"
+    - name: "clients"
+      zone: "$binary_remote_addr"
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -21,6 +21,10 @@ nginx_sites:
       zone: "$server_name"
       size: "10m"
       rate: "200r/s"
+    - name: "clients"
+      zone: "$binary_remote_addr"
```
emord: This will negatively affect trainings if they're on wifi and sharing an IP. We had this before https://github.com/dimagi/commcare-cloud/pull/646 and https://github.com/dimagi/commcare-cloud/pull/663 says we've dropped it because we can't do this.

do we need to clear it with app team/CPMU ?

snopoke: :( thanks for the reminder.

PR #2842, opened 2019-05-07
Add citus{0,1,2}-india to AWS

dannyroberts:
##### SUMMARY

@snopoke told me that citus on india is currently unpopulated, so I'm taking advantage of that replace it with a setup on AWS before there's any data and it becomes harder.


dannyroberts:
@NitigyaS do you know what causes

```
TASK [postgresql_base : Assert postgresql directory shows up in exactly one place] *************************************************************************************************************************
fatal: [10.203.10.74]: FAILED! => {"changed": false, "msg": "postgresl directory must either be /opt/data/postgresql or /var/lib/postgresql (and not both)"}
fatal: [10.203.10.190]: FAILED! => {"changed": false, "msg": "postgresl directory must either be /opt/data/postgresql or /var/lib/postgresql (and not both)"}
fatal: [10.203.10.77]: FAILED! => {"changed": false, "msg": "postgresl directory must either be /opt/data/postgresql or /var/lib/postgresql (and not both)"}
```

I am running just the normal first deploy command

```
cchq india deploy-stack --limit citus0:citus1:citus2 --skip-tags=users --skip-check
```

which is ending with that failure. I see postgresql is installed, and then /opt/data/postgresql is created, right before this check, so it makes sense that both `/var/lib/postgresql` and `/opt/data/postgresql` would be present.

NitigyaS:
@dannyroberts in the Postgresql playbook all the files from `/var/lib/postgresql` are moved to `/opt/data/postgresql` , but than since the home directory of `postgresql` user remains `/var/lib/postgresql` a few file are created there, and it can not be deleted, thus the above assertion fails. It is completely safe in above case to skip this task.

dannyroberts:
@NitigyaS can we make the check more specific so it doesn't fail unless there's really a problem? Do you have any suggestions for how I might do that?

src/commcare_cloud/ansible/roles/citusdb/meta/main.yml
```diff
@@ -3,7 +3,7 @@ dependencies:
   - role: monit
   - role: postgresql_base
     postgresql_version: '11'
-    postgresql_cluster_name: '11/main'
+    postgresql_cluster_name: 'main'
```
dannyroberts: I think this is the reason I was having the issue yesterday @NitigyaS. This was making it test paths like `/opt/data/postgresql/11/11/main/pg_log` instead of `/opt/data/postgresql/11/main/pg_log`

dannyroberts:
Okay @NitigyaS @snopoke I was eventually able to run
```
cchq india deploy-stack --limit citus0:citus1:citus2 --skip-tags=users --skip-check --branch=dmr/aws-india-citus
```
clean from start to finish.

To do that I first had to run

```
cchq india ansible-playbook deploy_citusdb.yml --limit citus0:citus1:citus2 --skip-check --branch=dmr/aws-india-citus --start-at-task='Add Citus apt key'
```
to make sure citus was installed, and then I could start over and run it clean from the beginning. Otherwise postgresql won't start, complaining it can't find the citus extension.

I didn't wade further into reorganizing it to run in the right order, but I wanted to flag that

dannyroberts:
Is there any command I need to run to propagate the inventory changes? I ran `update-config`, but there was no diff. I think other than that, this is good to merge. Once it's merged I'll shut off the other citus machines on softlayer and if that doesn't cause any problems after a few days delete them.

NitigyaS:
Since we have done any changes in `postgresql.yml` . CitusDB was not populated and thus no change in update-config

dannyroberts:
OK @NitigyaS so would you approve me merging this as is then?

PR #2844, opened 2019-05-07
Removing old kafka server from staging and adding systemd config for new kafka server

sanjay2916:


dannyroberts:
Will you remove it from terraform as well?

sanjay2916:
> Will you remove it from terraform as well?

once it is working fine, I will remove from terraform

PR #2845, opened 2019-05-07
add ppa for pango 1.38

calellowitz:
##### SUMMARY

This adds a ppa which includes backports the proper version of pango (1.38) to Ubuntu 14.04 so we can use it while we are in the process of upgrading our VMs. (only 1.36 is available in the main ubuntu repos and 1.38 is required for font selection with weasyprint)

##### ENVIRONMENTS AFFECTED

All
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Common Installs updated

@dannyroberts let me know if I did not handle the Ubuntu version part correctly


snopoke:
@calellowitz any reason why this was added to ALL vms and not just commcare vms?

PR #2848, opened 2019-05-07
Revert back to kafka0-staging

dannyroberts:
##### SUMMARY
@sanjay2916 I deployed this because we continued to see errors like https://sentry.io/organizations/dimagi/issues/1004425288/?environment=staging&project=136860&referrer=slack.

If you're able to get things working tomorrow we don't have to merge this PR, I just wanted to document what I did.

### DIFF
--- a/environments/staging/aws-resources.yml
+++ b/environments/staging/aws-resources.yml
@@ -17,7 +17,6 @@ pgproxy1-staging: 10.201.40.187
 pillow1-staging: 10.201.10.103
 proxy1-staging: 10.201.20.21
 proxy1-staging.public_ip: 107.20.83.77
-rabbit0-staging: 10.201.10.73
 rabbit1-staging: 10.201.10.129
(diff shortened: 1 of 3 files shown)


dannyroberts:
@emord said
>  fwiw I don't think we've ever migrate kafka data before. I believe we've always just wiped it and reset the pillow checkpoints for ICDS

I think the command for that is `KafakCheckpoint.objects.update(offset=0)`

sanjay2916:
> @emord said
> 
> > fwiw I don't think we've ever migrate kafka data before. I believe we've always just wiped it and reset the pillow checkpoints for ICDS
> 
> I think the command for that is `KafakCheckpoint.objects.update(offset=0)`

@Danny  Kafka data migrated to the new server and staging site is pointed to the new server.

do we need to remove the Kafka and zookeeper data and set to offset=0?

PR #2852, opened 2019-05-09
Redistribute queues across celery{4,5,6}-production

dannyroberts:
##### SUMMARY
so celery4 and celery6 are exact mirrors of each other

This is my stab at fixing https://dimagi-dev.atlassian.net/browse/PLAT-156

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
I just deployed this so it's live now

PR #2858, opened 2019-05-10
pgbouncer configuration fixes

snopoke:
The pgbouncer / postgresql config has to handle the following:
* monolith
* standby database where standby isn't in listed in postgresql.yml
* standby database that is listed in postgresql.yml
* primary and standby databases with the same name on multiple hosts
* chained standby databases

This proved to be too much to handle in the template so I decided to move the logic for which DB's are are which host into python and write the output into `.generated.yml`.


snopoke:
@emord thanks for the review - after discussion in devops meeting we decided to take a different approach. I'll be putting up a PR for that soon

PR #2859, opened 2019-05-11
don't store downtime locally

snopoke:
Query datadog for downtime record instead of storing it locally so that if the user ending the downtime is not the same as the user who started the downtime it will still get ended correctly on Datadog.

Also added an end time for the downtime to safeguard against it not being cancelled.

### DIFF
--- a/src/commcare_cloud/commands/ansible/downtime.py
+++ b/src/commcare_cloud/commands/ansible/downtime.py
@@ -31,6 +31,11 @@ class Downtime(CommandBase):
         Argument('-m', '--message', help="""
             Optional message to set on Datadog.
         """),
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/commands/ansible/downtime.py
```diff
@@ -31,6 +31,11 @@ class Downtime(CommandBase):
         Argument('-m', '--message', help="""
             Optional message to set on Datadog.
         """),
+        Argument('-d', '--duration', default=24, help="""
```
emord: May be worth noting the default is 24 hours if not set and that the units are in hours

PR #2602, opened 2019-01-29
demis-echis

demis-svenska:
I have modified some parameters on echis/public.yml, echis/app-process.yml and postgresql.yml files.
changes include:
          Memory-used by: elasticsearch, formplayer and jython
          Number of concurrent connections: postgresql
          celery and pillows: concurrency

PR #2863, opened 2019-05-13
Sk/pgbouncer hba

snopoke:
Revised version of https://github.com/dimagi/commcare-cloud/pull/2858 which does not auto-generate the HBA entries.

Still needs testing


src/commcare_cloud/ansible/roles/postgresql_base/templates/pg_hba.conf.j2
```diff
@@ -5,70 +5,22 @@ local   all             postgres                                peer
...
-{% endif %}
-
-{% if postgres_users.get('backup') %}
-host    all             {{ postgres_users.backup.username }}   {{ db_vlan_ip_range|default('0.0.0.0/0') }}    md5
-{% endif %}
-
-{% if postgres_users.get('tableau_readonly') %}
-host    all     {{ postgres_users.tableau_readonly.username }}   {{ tableau_server.split(':')[0] }}/32    md5
-{% endif %}
+{% for entry in postgres_hba_entries %}
```
dannyroberts: I think this should be `postgresql_hba_entries`

emord:
is this ready to be reviewed merged? Should someone take it over while you take your PTO?

If this is already deployed to TCL, and there are changes from master, it'd be nice to get this in to reduce noise in postgres deploys

snopoke:
I've made some recent changes - decided to remove pgbouncer HBA completely since I couldn't get it to work with with citus for some reason and I don't think it's essential.

I think this is good to go but I haven't done a final check on all envs yet so can't guarantee it. I was really hoping to get it out before I go but just haven't had the capacity.

snopoke:
Checked this against all our envs. There are a noop few diffs which I've rolled out on all envs except prod.

The only env I haven't checked is `echis`. The env config in master is quite out of date based on this PR: https://github.com/dimagi/commcare-cloud/pull/2602 which I believe is what they are using. Either way the changes should not adversely impact them.

@emord @dannyroberts

dannyroberts:
@sanjay2916 noticed had a question about pgbouncer_hba which led me to this PR as well as a couple remaining references in our code base:
- defining an unused variable: https://github.com/dimagi/commcare-cloud/blob/64208789c2f687876c6af679190d0fcb802943fe/src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml#L6
- in docs: https://github.com/dimagi/commcare-cloud/blame/b6749485b2dcea43282530676dc45a3c673fb7c0/docs/howto/add-celery-machines.md#L33

The first one seems save to just remove. I was wondering if you think the reference in the docs warrents being replaced with something else (like something similar but for pg_hba?) or if that step can just be removed altogether now

snopoke:
We've removed the pgbouncer HBA completely so we can remove that var and also remove the line about the instructions. There's no need to replace the instructions with anything.

Thanks @dannyroberts and @sanjay2916

dannyroberts:
https://github.com/dimagi/commcare-cloud/pull/3319

PR #2865, opened 2019-05-14
add citusdb to india config

snopoke:
The tag changes make it easier to update config / create db's etc without having to run all the installation tasks.
```
cchq india ap deploy_citusdb.yml --tags=configure
```

### DIFF
--- a/environments/india/postgresql.yml
+++ b/environments/india/postgresql.yml
@@ -1,6 +1,7 @@
 REPORTING_DATABASES:
   ucr: ucr
   icds-ucr: icds-ucr
+  icds-ucr-citus: icds-ucr-citus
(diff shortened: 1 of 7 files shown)


environments/india/postgresql.yml
```diff
@@ -32,6 +32,10 @@ dbs:
     - django_alias: icds-ucr
       name: icds-ucr
       django_migrate: True  # ICDS dashboard models are stored here
+    - django_alias: icds-ucr-citus
```
emord: Might have approved too soon. Should this get added to `REPORTING_DATABASES` so UCR engine_ids get set?

snopoke: b500712bd

PR #2866, opened 2019-05-15
Adding couchdb nodes in staging environment

sanjay2916:
Adding couchdb nodes in staging environment


environments/staging/terraform.yml
```diff
@@ -173,6 +173,28 @@ servers:
     volume_size: 400
     group: "couchdb2"
     os: trusty
+  - server_name: "couch3-staging"
+    server_instance_type: "t3.large"
```
snopoke: the current ones are `t2.large`. Is the change to `t3.large` intentional?

sanjay2916: yes Simon, t2.large instances are not EBS optimized now whereas t3.large are EBS optimized instances

PR #2867, opened 2019-05-15
swap pgshard8 and pgshard8standby

snopoke:
Since we have seen high CPU on pgshard8 we are going to try promote pgshard8standby to see if it has the same issues.


sravfeyn:
👍

PR #2860, opened 2019-05-13
pg config changes for UCR DBs from pgtune

snopoke:
* DB Version: 9.6
* OS Type: linux
* DB Type: dw
* Total Memory (RAM): 80 GB
* CPUs num: 32
* Connections num: 350
* Data Storage: ssd

max_connections = 350
shared_buffers = 20GB
effective_cache_size = 60GB
maintenance_work_mem = 2GB
checkpoint_completion_target = 0.9
wal_buffers = 16MB
default_statistics_target = 500
random_page_cost = 1.1
effective_io_concurrency = 200
work_mem = 1872kB
min_wal_size = 4GB
max_wal_size = 8GB
max_worker_processes = 32
max_parallel_workers_per_gather = 16

PR #2872, opened 2019-05-16
refactor pg inventory for easier diff view

snopoke:
Based on feedback here: https://github.com/dimagi/commcare-cloud/pull/2860#discussion_r283372222

I've done this in 2 ways:
1. [grouped by host / group](https://github.com/dimagi/commcare-cloud/blob/dfe3417964db41279f49df19ec1aeec3663dab31/environments/icds-cas/inventory/postgresql.csv)
    * this could also be sorted by var
2. [grouped by var](https://github.com/dimagi/commcare-cloud/blob/dfe3417964db41279f49df19ec1aeec3663dab31/environments/icds-cas/inventory/postgresql1.csv)

I think 2 is easier to read but interested to get other's feedback.

@dimagi/devops


sravfeyn:
2 is easier to read but I like #1 since we might usually be interested in updating bunch of config for a given group, than updating a var on multiple groups.

PR #2873, opened 2019-05-16
Update docs to say we use 18.04

esoergel:
##### SUMMARY
Update docs to say we use Ubuntu 18.04

##### ENVIRONMENTS AFFECTED
None

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
README

##### ADDITIONAL INFORMATION
I see a number of references to `ansible_distribution_version == '14.04'` - do those need to be updated with the OS upgrade?  Similarly, I see references to `os: trusty`, which may need updates.

Both of those are potentially substantive changes which I'm considering outside the scope of this PR.

### DIFF
--- a/docs/system/system-overview.md
+++ b/docs/system/system-overview.md
@@ -19,9 +19,9 @@ Would be nice to get a diagram in here.
 This section will cover OS, system, and user-level setup
 that is common among all VMs involved in a CommCare HQ production environment.
 
-### Ubuntu Linux OS (14.04)
+### Ubuntu Linux OS (18.04)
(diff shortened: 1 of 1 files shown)


dannyroberts:
> I see a number of references to ansible_distribution_version == '14.04' - do those need to be updated with the OS upgrade? Similarly, I see references to os: trusty, which may need updates.

We're still in the process of migrating from 14.04 to 18.04 on all the VMs in our own environments, but once we're all migrated, we'll be removing those as well.

PR #2874, opened 2019-05-16
Upgrade couchdb-cluster-admin to 0.5.0

dannyroberts:
This includes the upgrades I made to couchdb-cluster-admin in Feb and Mar during the AWS migration that made it more robust in a few more circumstances. Will be good to have in soon as we have a couple more migrations coming up (Ubuntu/india AWS)


dannyroberts:
@snopoke FYI. The last three commits should be reviewed individually, as the first of them copies a directory verbatim and then the next two show readable diffs for each of the two copies (the original and the copy).

Basically, the shard allocation strategy has (intentionally) changed in couchdb-cluster-admin 0.5.0 to strongly prefer keeping as more than half of copies of a shard in their original location for stronger consistency guarantees during migration, and I needed to update a test to reflect that. I decided to break `new_node` into `new_node` (`couch0,couch1:2 => couch0,couch1,couch2:3` with `couch2` getting a copy of all the shards) and `new_node_empty` (`couch0,couch1:2 => couch0,couch1,couch2:2` with `couch2` getting none of the shards) to capture both situations in the new strategy.

PR #2875, opened 2019-05-17
enable wal archiving on ICDS pgucr

snopoke:
This is to support file level backups by TCL


snopoke:
The backup tool will delete them though I have been thinking we should put a cron job in place as well.

PR #2878, opened 2019-05-20
add spare pgucrstandby5 as web24 for blob migration

sravfeyn:
@snopoke 
Planning to use this just to do a setup release and use for run blob migration. Not going to deploy proxy etc. So, no need to merge this.


sravfeyn:
Going to follow below steps.
```
cchq icds ansible-playbook deploy_shared_dir.yml --limit shared_dir_host
cchq icds deploy-stack --tags=pgbouncer_hba --limit postgresql
cchq icds deploy-stack --first-time --limit <new host>
# Also, run command to stop Django on the new machine
cchq <env> fab setup_release:keep_days=20
```
And then going to initiate migration commands.

snopoke:
watch out for pgbouncer hba changes, they are broken in master for the UCR databases. I think you should limit it to just `shard_dbs:main_dbs`

You should make sure that you delete any supervisor service files so that the cchq processes don't restart at any point.

PR #2879, opened 2019-05-20
set citus shard count per DB

snopoke:
##### SUMMARY
Adds support for setting PostgreSQL config parameters per DB

##### ENVIRONMENTS AFFECTED
* india

##### COMPONENT NAME
Postgresql

##### ADDITIONAL INFORMATION
This could be done at the server level but I felt that it is the kind of setting that is more appropriate at the DB level.

This makes use of two new Ansible features:
* postgresql_set: set config on a PG database server
* subelements: Jinja filter to replace `with_subelements` loop control

Note: changes to yaml list format are as a result of upgrading `pyyaml`


requirements.txt
```diff
@@ -2,58 +2,58 @@
 # This file is autogenerated by pip-compile
 # To update, run:
 #
-#    pip-compile --output-file requirements.txt setup.py requirements-python-lt-2.7.9-ssl-issue.in
+#    pip-compile --output-file=requirements.txt requirements-python-lt-2.7.9-ssl-issue.in setup.py
 #
 ansible-vault==1.1.1
-ansible==2.5.0
+ansible==2.8.0
```
calellowitz: why so many requirement upgrades?

src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
```diff
@@ -78,22 +78,17 @@
...
 
-- name: Create PostgreSQL databases
+- name: Update PostgreSQL database config
   become_user: "{{ db_is_remote and 'ansible' or 'postgres' }}"
   vars:
     ansible_ssh_pipelining: true
-  postgresql_db:
-    name: "{{ item.name }}"
-    state: present
+  postgresql_set:
```
calellowitz: Does this still create the dbs?

requirements.txt
snopoke: their first commit was just `make upgrade-requirements` without the ansible version change. The 2nd commit I changed the ansible version and ran it again.

src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
snopoke: hmm, not sure how that happened. Updated

snopoke:
tested this on `india`

snopoke:
@dannyroberts the ansible upgrade seems to have broken the path replacements in the ansible help text: adbaba7. I wasn't able to get the `ANSIBLE_LIBRARY` env var to work at all even though it's still listed in the docs. Would you be able to have a look?

dannyroberts:
Yeah sure

dannyroberts:
@snopoke is `ANSIBLE_LIBRARY` not working just in the sense that it's no longer modifying the output of `ansible -h`? It may just be that they decided that your environment variables shouldn't change the output of `ansible -h`, or that they changed the place in which that env var is loaded to affect the setting so that it's not called before generating the help text, creating a more or less innocuous but unintentional change. I think either way, that alone isn't anything to be worried about and it's fine for us to change our help text to match.

Is there another way in which `ANSIBLE_LIBRARY` isn't working? Like are our actual modules at `src/commcare_cloud/ansible/library` no longer getting used?

snopoke:
As far as I can tell it's just not being taken into account with the help text. If you're OK with that then no need to investigate further. For now I've remove the logic from the makefile which would replace the path.

PR #2881, opened 2019-05-21
Update firefighting docs for nfs hanging

dannyroberts:
##### SUMMARY
I made a few substantive changes (like stopping datadog), and the rest is just reformatting.

### DIFF
--- a/docs/firefighting/index.md
+++ b/docs/firefighting/index.md
@@ -266,7 +266,7 @@ It may happen, specially if the client crashes or has kernel oops, that a connec
     2. `umount` results in `umount.nfs: /mnt/shared_icds`: device is busy
 2. top all app processes (gunicorn, etc) and datadog
     1. `sudo supervisorctl stop all`
-    5. `sudo service datadog-agent stop`
+    2. `sudo service datadog-agent stop`
 3. Force umount 
     1. `sudo umount -f /mnt/shared_icds`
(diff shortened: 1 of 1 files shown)


docs/firefighting/index.md
```diff
@@ -257,25 +257,27 @@ $ mount /mnt/shared_data
...
-2. top all gunicorns and supervisor
-    a. `sudo supervirsorctl stop all`
-    b. jVerify “stuck” processes keep running: $ ps aux|grep gunicorn
-    c. $ sudo service supervisord stop
-    d. Verify “stuck” processes are gone: $ ps aux|grep gunicorn
+    1. `df` doesn’t work, it hangs. Same goes for `lsof`.
+    2. `umount` results in `umount.nfs: /mnt/shared_icds`: device is busy
+2. top all app processes (gunicorn, etc) and datadog
+    1. `sudo supervisorctl stop all`
+    5. `sudo service datadog-agent stop`
```
orangejenny: Curious why this is a 5? I thought you could write ordered lists with `1.` starting each item and then markdown takes care of the numbering - this item shows up as `ii` in the preview, not `v`.

dannyroberts: Oops yeah that was a mistake. I changed it to `2`. But yeah you're right it replaces any number with numbers in the right order (1, 2, 3, etc. / i, ii, iii, etc.).

PR #2882, opened 2019-05-21
Add jmx port to kafka service on bionic

dannyroberts:
##### SUMMARY
seems like systemd (which we now use on Ubuntu 18.04 Bionic) doesn't pull environment variables from `/etc/environment`, so to fix https://app.datadoghq.com/monitors/8131399?q=host%3Akafka1-india%2Cinstance%3Akafka_instance&live=5min you have to add it to the service definition.

affects:
- [x] india
- [x] staging


src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
```diff
@@ -9,8 +9,8 @@
 - name: Set JMX_PORT in /etc/environment
   lineinfile:
     dest: /etc/environment
-    line: 'JMX_PORT=9999'
-    state: present
+    line: 'JMX_PORT={{ KAFKA_JMX_PORT }}'
+    state: {{ 'present' if ansible_distribution_version == '14.04' else 'absent' }}
```
orangejenny: Confirming this is supposed to be 14.04 although the PR description references 18.04?

dannyroberts: Yeah this is saying for 14.04 do it the old way (put this line in `/etc/environment`) and for 18.04 don't, put it in the systemd service file instead

PR #2886, opened 2019-05-23
Dmr/test couchdb2.3.1

dannyroberts:
@sanjay2916 try using this as a starting point for setting up new ubuntu 18.04 couch nodes on couch 2.3.1. I was able to get it to run to the end of the setup, though I didn't then try to have the node join the cluster or migrate data yet.


dannyroberts:
@sanjay2916 was able to use this to set up the new staging machines, which are now on CouchDB 2.3.1. (Other envs are on CouchDB 2.1.1)

emord:
are we going to update the other envs? should we also update the couchdb docker for HQ?

dannyroberts:
Yeah we're going to update all envs, as part of the Ubuntu migration.

The question about docker is a really good one I hadn't considered. We use `klaemo/couchdb:2.0-dev` which has been deprecated (https://github.com/klaemo/docker-couchdb) and replaced. I'm not sure if it's a drop-in replacement or if it'll take a little trial and error to make it work.

PR #2893, opened 2019-05-28
update icds maintenance window

calellowitz:
##### SUMMARY

Updates the suggested maintenance window for ICDS to prevent conflicts with the aggregation. I also pushed the front up a bit because I think there is no reason to avoid deploying at the tail end of the india day and this gives a slightly wider window (although it won't actually prevent you from deploying so its more of a suggestion anyway)

@emord @snopoke @sravfeyn @Rohit25negi

### DIFF
--- a/environments/icds-cas/fab-settings.yml
+++ b/environments/icds-cas/fab-settings.yml
@@ -1,5 +1,5 @@
 acceptable_maintenance_window:
-  hour_start: 0
-  hour_end: 7
+  hour_start: 20
+  hour_end: 23
   timezone: Asia/Kolkata
 email_enabled: True


calellowitz:
FYI: @ctsims

environments/icds-cas/fab-settings.yml
```diff
@@ -1,5 +1,5 @@
 acceptable_maintenance_window:
-  hour_start: 0
-  hour_end: 7
+  hour_start: 20
+  hour_end: 23
```
snopoke: A deploy at 11pm won't be likely to end before 1130pm. Can we make this `22:30`?

snopoke:
Merged this despite comment since it's an improvement as is

PR #2894, opened 2019-05-28
Fix swiss backup

dannyroberts:
##### SUMMARY
The last backup from swiss was uploaded in July 2018. The snapshots were getting made on disk, but the script for uploading to their object storage had gotten broken. This PR fixes that.

##### ENVIRONMENTS AFFECTED
swiss

### DIFF
--- a/src/commcare_cloud/ansible/roles/backups/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/backups/tasks/main.yml
@@ -1,29 +1,16 @@
-- name: Create aws config directory
+- name: Set up aws credentials
   become: yes
-  file:
-    path: '~{{ blobdb_user }}/.aws'
-    owner: "{{ blobdb_user }}"
-    group: "{{ blobdb_user }}"
-    mode: 0700
-    state: directory
+  import_role:
+    name: backups
+    tasks_from: set_up_aws_credentials.yml
+  vars:
+    s3_user: "{{ blobdb_user }}"
+    s3_user_home_dir: "~{{ blobdb_user }}"
   when: blobdb_s3
   tags:
     - cron
     - backups
 
-- name: copy boto3 credentials file
-  become: yes
-  template:
-    src: "aws_credentials.j2"
-    dest: "~{{ blobdb_user }}/.aws/credentials"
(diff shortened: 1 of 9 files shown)


src/commcare_cloud/ansible/roles/couchdb/tasks/main.yml
```diff
@@ -52,7 +52,10 @@
   when: not couchdb_path.stat.exists and couchdb_compile|success
 
 - name: Add CouchDB user
-  user: name=couchdb createhome=no password=no state=present
+  user:
+    name: couchdb
+    home: '{{ couchdb_dir }}'
+    state: present
```
calellowitz: what does this change do?

src/commcare_cloud/ansible/roles/backups/tasks/main.yml
```diff
@@ -1,29 +1,16 @@
...
-    owner: "{{ blobdb_user }}"
-    group: "{{ blobdb_user }}"
-    mode: 0700
-    state: directory
+  import_role:
+    name: backups
+    tasks_from: set_up_aws_credentials.yml
+  vars:
+    s3_user: "{{ blobdb_user }}"
+    s3_user_home_dir: "~{{ blobdb_user }}"
```
calellowitz: what does `~{{ blobdb_user }}` resolve to?

src/commcare_cloud/ansible/roles/couchdb/tasks/main.yml
dannyroberts: This gives the couchdb user a home directory, which is necessary because our backup upload script relies on `~/.aws/credentials`. We were properly placing the credentials file at `{{ couchdb_dir }}/.aws/credentials`, but it wasn't getting picked up because that wasn't set as the couchdb user's home dir. (It didn't have a home dir.)

src/commcare_cloud/ansible/roles/backups/tasks/main.yml
dannyroberts: It resolves to `~root` which is normally /var/root` I think

dannyroberts:
Test passed and waiting for formal approval

PR #2895, opened 2019-05-28
Update blobdb migration docs

dannyroberts:
##### SUMMARY
While doing this recently I took the opportunity to expand on the parts I had to look up or ask @millerdev about.

### DIFF
--- a/docs/howto/migrate-s3-to-s3.md
+++ b/docs/howto/migrate-s3-to-s3.md
@@ -74,7 +74,7 @@ XFormInstance: checked 42 records
 
 Legend:
 - `Not referenced` is OK. It means that the blob that was said to be "missing"
-  isn't in the blobdb but also isn't referenced in blob metadata anymore, so it was likely deleted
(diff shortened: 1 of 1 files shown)


docs/howto/migrate-s3-to-s3.md
```diff
@@ -30,16 +30,80 @@ or from a Riak CS cluster to Amazon S3, for example.
...
+```
+
+Legend:
+- `Not referenced` is OK. It means that the blob that was said to be "missing"
+  isn't in the blobdb but also isn't referenced in blob metadata anymore, so it was likely deleted
+  while the migration was running.
+- `Not found` means that the missing blob is still referenced in blob metadata,
+  but not found in either the old backend or new backend.
+- `Found in (new|old) db` means that actually it is present in one of the backends,
+  unlike originally reported. These are generally 0.
```
millerdev:
```suggestion
  unlike originally reported. Re-run the `check_blob_logs` command with `--migrate` to migrate items "Found in old db".
```

I actually found that flag useful when I migrated staging to S3 because there was a period of time during the migration when it was failing in some way (couch was down?) that caused individual blob migrations to fail, and the situation was able to be recovered by this script.

PR #2898, opened 2019-05-29
Increase production celery resources for async_restore_queue

dannyroberts:
##### SUMMARY
Short term fix for https://docs.google.com/document/d/1U3h-IVdQLH_encgSM0Q42L4fJ1rRwomysfyeXgDMvSY/edit?ts=5ced7cdc#

This is already rolled out

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -224,7 +224,7 @@ servers:
     os: trusty
 
   - server_name: "celery0-production"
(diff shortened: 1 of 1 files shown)


dannyroberts:
I had considered increasing couchdb2 resources as well, but the cost of that seemed unlikely to be worth it when so far from what I can tell from graphs and such, memory on these machines is the main issue

PR #2899, opened 2019-05-29
use Python 3.7 instead of 3.6

nickpell:
Installs a Python PPA to support Python 3.7 and uses 3.7 in the virtualenv.

Steps to apply:

```commcare-cloud <env> ap deploy_common.yml --tags=common_installs```
```cchq <env> ansible-playbook deploy_commcarehq.yml --tags=py3 -e 'CCHQ_IS_FRESH_INSTALL=1'```

@dimagi/py3


esoergel:
What's the deal with `CCHQ_IS_FRESH_INSTALL=1`?  That seems like a non-sequitor - isn't it mostly pertinent to db migrations?

Does this need a changelog entry?

nickpell:
```CCHQ_IS_FRESH_INSTALL=1``` is required to run ```py3 install pip requirements```: 
https://github.com/dimagi/commcare-cloud/blob/1cc5602f1248e3241a1988c1de876012ecfd9fcd/src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml#L104

I'd prefer to make a single changelog entry containing all of the steps to migrate to Python 3 once the migration process is complete.

@esoergel

nickpell:
@esoergel What if we added a new flag to the check like ```CREATE_NEW_VIRTUALENV```?

esoergel:
@nickpell yeah, that'd be perfect

PR #2901, opened 2019-05-29
[prove-deploy] Fix couchdb install (2.1.1) on trusty

dannyroberts:
##### SUMMARY

Even though we are soon deprecating trusty, I think it's appropriate
to keep things working until we officially do. (I broke this in a PR I
merged today and the error showed up in the prove-deploy build on
master.)

This PR makes it install CouchDB 2.1.1 on Trusty and CouchDB 2.3.1 on Bionic. I didn't try to make 2.3.1 work on Trusty once I saw there were issues because we're about to deprecate support for Trusty anyway.

##### ENVIRONMENTS AFFECTED
No known environments were affected by the short-lived issue


src/commcare_cloud/ansible/roles/nginx/tasks/install.yml
```diff
@@ -1,13 +1,8 @@
...
-- name: Remove official nginx repository
-  become: yes
-  apt_repository: repo='ppa:nginx/stable' state=absent
-  when: ansible_distribution_version == '14.04'
-
 - name: Add nginx repository
   become: yes
   apt_repository: repo='ppa:ondrej/nginx' state=present
-  with_items: "{{ 'ppa:ondrej/nginx' if ansible_distribution_version == '14.04' else 'ppa:nginx/stable' }}"
+  with_items: 'ppa:nginx/stable'
```
emord: this is pre-existing, but this doesn't doesn't do anything. There's no evaluation of an item here, so I think `ondrej/nginx` is always being installed

dannyroberts: Oh wow, thank you @emord for pointing that out. Looks like I misread this task multiple times. The change I was trying to do was

```
apt_repository: repo='ppa:nginx/stable' state=present
```

The current commit doesn't provide anything of value so I'm just going to amend it. Thanks!

src/commcare_cloud/ansible/roles/nginx/tasks/install.yml
```diff
@@ -1,13 +1,7 @@
 ---
-- name: Remove official nginx repository
-  become: yes
-  apt_repository: repo='ppa:nginx/stable' state=absent
-  when: ansible_distribution_version == '14.04'
```
emord: Will this break nginx installs on trusty? wouldn't be great to fix couch only to break nginx?

fwiw I think i added this to support http/2 on nginx. I'm not sure if nginx's repository has been updated since or not.

dannyroberts: Given my last oversight, I could be confused about what I was seeing, but the motivation for this change was that there's currently an issue with the nginx install (that I noticed both when installing real machines and when running the prove-deploy build) where's what's installed is an old version that doesn't support the "zone" directive. My hope is that this fixes nginx config syntax errors like this one:
```
nginx: [emerg] unknown directive "zone" in /etc/nginx/sites-enabled/{env}_couchdb_25984
```

If it does, then that means that the version on `ppa:nginx/stable` is _later_ than the version on `ppa:ondreij/nginx`, which is what I thought I had discovered a couple weeks ago when installing on bionic. I guess we'll see

emord: Are you sure that that error is not from nginx/stable on 14.04 (which is what we were removing here)? nginx/stable is 1.12.0 for trusty (14.04) and nginx/stable is 1.14 for bionic and ondreij/stable is 1.16 for bionic, but that was updated 9 hours ago so likely 1.14 before

dannyroberts: I think the way I did this change now is right. According to http://nginx.org/en/docs/http/ngx_http_upstream_module.html#zone

> This directive appeared in version 1.9.0.

so when a few weeks ago I was looking into this, I concluded that it was using nginx < 1.9.0 and that was the cause of the above error. (I think I verified it in a different way as well, though I now forget exactly what I did.)

After the change in this PR, the error is now gone. That matches up with the versions you stated being installed by nginx/stable on trusty and bionic; both of those versions are >= 1.9, so it makes sense that the error would not be present in those versions. I'm not sure what version exactly `ondrej/nginx` was installing, but it seems to be <1.9 based on that error.

emord: If we just need nginx 1.14, then we could get rid of this repository totally https://packages.ubuntu.com/bionic/nginx

either way I don't have an issue with adding the official nginx ppa

dannyroberts: I see. You're saying on bionic we can probably get away with just using the bionic default. We must have added this in the first place because trusty's default nginx install was too old a version, right? I think for now we should just go with this, but I'll add a comment here that on bionic we may not need to add a separate repo at all.

PR #2902, opened 2019-05-31
Remove riak playbook

emord:
##### SUMMARY
This will remove riak setup from commcare-cloud

##### ENVIRONMENTS AFFECTED
None (managed by Dimagi)

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
riak

##### ADDITIONAL INFORMATION
This is not a complete PR, but I wanted to open it up for discussion. Now that ICDS is moving to a hosted object storage and our production environments are using s3, we will not be running riak in a production instance. 

Also we are currently in the process of upgrading all machines to Ubuntu 18.04, but we have no production servers on 18.04. Riak was never test on 18.04 (source: https://docs.riak.com/riak/kv/latest/setup/installing/debian-ubuntu/index.html), so its possible that there may be subtle bugs that stick around for quite a while. (like the issues with kafka and couch found during the migration). I'm pretty confused about riak's current development status. It appears that there is a "riak s2" that supports cloud storage, riakcs (what we have used) which has not been updated since Basho's demise, and riakkv which I believe has [community open source releases now](http://lists.basho.com/pipermail/riak-users_lists.basho.com/2019-May/039358.html)?

My ideal scenario would be:

0. Write change logs/give appropriate warnings that we won't be supporting riak setup directly in this repository
1. Wait for TCL object storage migration to finish
2. Delete those servers
3. Remove riak playbooks configuration
4. Update any documentation to say S3-compatible object storage instead of riak
5. Potentially point to self-hosted solutions and other ansible packages (such as https://github.com/basho-labs/ansible-riak though we probably shouldn't unless we can verify those work, with some migration path)

@ctsims and @dannyroberts I think you two would be the most able to answer whether or not this is a feasible change with any third party contracts that have been made. If its not possible we may need to keep around some of the old ubuntu 14.04 code or begin explicitly testing riak on 18.04 on some test environment (I'm not sure how involved it would need to be)?


millerdev:
+1 on the plan you outlined. It would be great to remove Riak CS from our list of supported services.

The amount of knowledge needed to run a production Riak CS cluster competently is not small. Now that Basho is no longer around to provide support for bugs (as we have found in the past) it feels like we are probably more on our own to come up with supported configurations. To be fair, I have not needed to interact with the developers currently maintaining Riak. They are probably great, although I do not know how much time they have to provide support.

dannyroberts:
Yeah this seems like the right option given the sorry state of the riakcs project, as you've outlined.

I think something to keep in mind is that we are committed to being able to run using an OSS version of  any service we depend on. It's not an issue to use proprietary backends (like S3) as long as we also support a non-proprietary version.

This leaves us in a slightly tricky spot. (I mean we're already in a tricky spot as you point out, since riakcs isn't tested on Ubuntu 18.04.) We support a filesystem backend, but plugging in an OSS scalable/HA object storage backend would require rolling your own automation to do so (or setting it up manually). I think the fact that such OSS exists (I'm thinking of min.io, for example) makes this kind of OK, but it's worth noting that until we have tried it out and have automation for setting it up, it's hard to say we actually support it.

Overall, I think _not_ ending support for riakcs right now would be disingenuous, so I'm in favor of the move. Whether or not we need to work to replace it with a different OSS component may depend on whether we foresee anyone outgrowing a filesystem backend anytime soon. Otherwise, that too is likely to fall out of testing unless we make it a priority.

ctsims:
One other option I wanted to mention here is that ICDS has transitioned to
a cloud object storage layer which I believe to be OpenStack Swift
<https://wiki.openstack.org/wiki/Swift>, which is still receiving active
releases.

It may be worth evaluating as an alternative if we're already dependent on
it elsewhere.

-Clayton

dannyroberts:
If we do pick a "winner" to go with I would choose carefully. I think in addition to swift, I think we should seriously consider https://min.io/. Minio has a lot going for it including ease of installation, flexibility, lightweightness, speed, simple scalability model, and "momentum" (https://trends.google.com/trends/explore?date=today%205-y&geo=US&q=minio,openstack%20swift). Not sure if the limits on bucket size that TCL is claiming on object storage are inherently theirs or swift's but that would also be a factor.

millerdev:
I agree, minio looks cool. Last time I evaluated it there were still some big issues and missing features related to production cluster management. Enough so that I wouldn't have thought of using it in production at the time. However, it does look like the project has a lot of momentum, so hopefully they've addressed those issues by now.

snopoke:
I don't have any real experience with openstack but it's not really designed to be used as a standalone service. The setup is quite complex and has a lot of different components.

emord:
So I looked around for an existing riak-cs ansible playbook, but could not find any that matched the version we use. The "official" one is for riak cs 1.4.5 https://github.com/basho-labs/ansible-riakcs and all up to date playbooks look to be for riak-kv and riak-ts. This makes https://github.com/dimagi/commcare-cloud/pull/2902/commits/b7f20e158f686bd978df2bff7fcca49c55a291a3 rather unsatisfying so I'm happy to hear any more documentation that we think we should add

emord:
TCL riak servers have been decommissioned and we're no longer maintaining any of these servers. Is this good to remove now? @snopoke @dannyroberts ?

docs/changelog/0023-remove-riak.md
```diff
@@ -0,0 +1,24 @@
+<!--THIS FILE IS AUTOGENERATED: DO NOT EDIT-->
```
dannyroberts: I think you (unfortunately) have to manually remove this file if this was the previously created one before a numbering change

emord:
will update changelog entry after https://github.com/dimagi/commcare-cloud/pull/3181 is merged so that we don't have a conflict

PR #2903, opened 2019-05-31
use Python 3.6.8 because of Ubuntu 14.04 compatibility

nickpell:
I was going to wait to see if we got off of 14.04 by the time I was ready to migrate Python, but after today's deploy mess I figured I'd switch us back to 3.6 now.

Python 3.7 fails to build the virtualenv in 14.04.   This has the nasty side-effect of leaving the virtualenv owned by ```root```, which causes ```clean_releases``` to fail.

https://github.com/dimagi/commcare-hq/pull/24298 will work in Python 3.6.8, which has a PPA for 14.04 and will be built from scratch in 18.04.  (I'm deliberating ignoring 16.04 for now because we only have one machine on that version and it's scheduled for update by the end of Q2 2019.)

Once this is merged I'll update staging to use 3.6.8.

Once we are off of 14.04, we can use the PPA and support Python 3.7.

@dimagi/py3


nickpell:
Will look into using this PPA for Python 3.6.8 on Ubuntu 18.04 instead of building from scratch:
https://launchpad.net/~jonathonf/+archive/ubuntu/python-3.6

millerdev:
> Python 3.7 fails to build the virtualenv in 14.04.

Do we know why? Is it something we can fix?

nickpell:
@millerdev Trying to install packages on 14.04 with the 3.7 deadsnakes PPA leads to errors like this:
https://stackoverflow.com/questions/41328451/ssl-module-in-python-is-not-available-when-installing-package-with-pip3?noredirect=1&lq=1

I'd rather confidently use 3.6 than use 3.7 if we have to hack it together to work.  Upgrading to 3.6->3.7 will be easy once all of our machines running Python are off of 14.04 (non-ICDS machines should be upgraded by the end of Q2, and I've talked to the ICDS team about prioritizing webworkers, celery, pillow machines etc).

nickpell:
bump

millerdev:
Sorry for the delay. I didn't realize I hadn't approved it.

PR #2904, opened 2019-06-04
include Python 3 virtualenv in all managed environments

nickpell:
Applied to:

- [x] production

- [x] india

- [x] swiss

- [x] pna

- [x] icds

### DIFF
--- a/environments/icds-cas/fab-settings.yml
+++ b/environments/icds-cas/fab-settings.yml
@@ -3,3 +3,4 @@ acceptable_maintenance_window:
   hour_end: 23
   timezone: Asia/Kolkata
 email_enabled: True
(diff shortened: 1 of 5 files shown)


nickpell:
@dimagi/team-commcare-hq Please run ```update-code``` in your commcare-cloud directory before deploying.  Failure to do so will cause subsequent deploys to block, and after we've migrated to Python 3 will cause downtime.

God forbid you don't update and find yourself in trouble, the Python 3 virtualenv can be rebuilt by running
```cchq <env> ansible-playbook deploy_commcarehq.yml --tags=py3 -e 'CREATE_NEW_VIRTUALENV=1'```

PR #2905, opened 2019-06-04
create virtualenvs when CREATE_NEW_VIRTUALENV is set

nickpell:
https://github.com/dimagi/commcare-cloud/pull/2899#issuecomment-497320405


esoergel:
:tiger:

PR #2906, opened 2019-06-04
include changelog instructions to upgrade to python 3

nickpell:


nickpell:
@dimagi/py3

dannyroberts:
@nickpell not sure if this came up in the meeting yesterday (sorry I couldn't make it) but I think it'd be worth merging just the changelog / docs part of this PR now, so we can start prompting people to update

millerdev:
@dannyroberts we did discuss that. It's one of the actions items outlined in [notes from the meeting](https://docs.google.com/document/d/1S-JEaB52HUuss0waKQGdJI-qTTWSqtHFNi_0udCK9wY/edit).

emord:
@millerdev @snopoke I think you guys were in the meeting were you discussed this. Why isn't `py3_run_deploy` mentioned in this changelog?

Also it appears that if anyone runs `update-supervisor-confs` they'll be using `py3_virtualenv_home` even if they haven't set any settings in fab-settings. I think that would break any py2 environments unless I'm missing something

snopoke:
@emord we didn't discuss the chanelog at all. If it's incorrect that's news to me and it should be updated ASAP. We should also put out an update on the forum (can just respond to the original posts).

PR #2907, opened 2019-06-04
configure file limit of PG

snopoke:
For Ubuntu 18.04 it doesn't work to change `/etc/security/limits.conf`


src/commcare_cloud/ansible/roles/postgresql_base/defaults/main.yml
```diff
@@ -20,6 +20,8 @@ postgresql_archive_path: "{{ postgresql_dir_path }}/wal_archive"
 postgresql_archive_command: "cp %p {{ postgresql_dir_path }}/wal_archive/%f"
 postgresql_enable_archive: no
 
+postgresql_systemd_config_dir: "/etc/systemd/system/postgresql@.service.d"
```
sravfeyn: Is this an OK setting to have on Ubuntu 14.04 as well?

snopoke: it won't have any effect on 14.04

PR #2910, opened 2019-06-04
fix apt deprecation warnings

snopoke:
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `name: "{{ item }}"`, please use `name:
['postgresql-{{ postgresql_version }}-plproxy', 'postgresql-contrib-{{ postgresql_version }}', 'postgresql-server-dev-{{ postgresql_version }}', 'python-docutils', 'libpq-dev', 'python-psycopg2', 'unzip']` and remove
the loop. This feature will be removed in version 2.11.

### DIFF
--- a/src/commcare_cloud/ansible/deploy_postgres.yml
+++ b/src/commcare_cloud/ansible/deploy_postgres.yml
@@ -28,12 +28,8 @@
     - name: Install python-psycopg2 to be able to set up remote postgresql
       become: yes
       block:
-        - apt: name={{ item }} state=latest
-          with_items:
-            - libpq-dev
-        - pip: name={{ item }} state=present
-          with_items:
-            - psycopg2
+        - apt: name=libpq-dev state=latest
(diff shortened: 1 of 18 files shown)


snopoke:
> oh man. i was just starting to do tihs 5698a0f

sorry for that. There's also this one if you feel so inclined:
```
[DEPRECATION WARNING]: evaluating postgresql_enable_archive as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the future. Also see CONDITIONAL_BARE_VARS configuration
 toggle.. This feature will be removed in version 2.12.
```

emord:
no problem i'd rather they get fixed by 2 people than never be fixed

@snopoke Yeah there are several of those. currently running through the entire deploy-stack for some collateral. was going to do one PR with those but feel free to pick up whichever ones you want to

PR #2911, opened 2019-06-04
Update pg wal archive cron

emord:
##### SUMMARY
Change pg wal archive to 12 hours

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
postgres

##### ADDITIONAL INFORMATION
@snopoke We found this was 2160 minutes on the ICDS server and the server was almost out of space. Is there a reason that deviates from the 60 minutes in the codebase? We removed all logs from greater than six hours ago (720 min) manually. The cron task still has 2160 minutes

### DIFF
--- a/src/commcare_cloud/ansible/roles/postgresql_base/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/postgresql_base/tasks/main.yml
@@ -88,7 +88,7 @@
   become: yes
   cron:
     name: "Clean PG WAL Archive"
-    job: "find {{ postgresql_archive_path }} -mmin +60 -delete"
+    job: "find {{ postgresql_archive_path }} -mmin +720 -delete"
     user: root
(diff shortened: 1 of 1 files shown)


snopoke:
The backup schedule hadn't been set. Originally I made this 60 mins until the backups were configured. Originally I changed it to `-mtime +2` but that was too long so after disk full on Sunday I set it to 36 hours assuming we were backing up every 24 hours.

Today I got them to change the backup to be every 6 hours so this should be at least 7 hours. I also think we'll need to increase the size of this disk again.

emord:
I messed up 720 min is 12 hours, not 6. So I think this is good to roll out? (note that neither @calellowitz nor I updated this cron task, we only ran find -delete manually)

snopoke:
I rolled this out

PR #2912, opened 2019-06-04
increase max locks per transaction

calellowitz:
##### SUMMARY

This increases `max_locks_per_transaction` for ucr machines on ICDS. It will hopefully address the 
```
OperationalError: out of shared memory
HINT:  You might need to increase max_locks_per_transaction.
```
errors we have seen in the aggregation logs and similar errors in `psql` shells. It is currently at 64 (the default). The internet did not have great advice on how high to set this, but a common suggestion seemed to be to double it if there were issues so I figured I would start there.

A little more info about the parameter itself is available [in the postgres docs](https://www.postgresql.org/docs/current/runtime-config-locks.html). Notably, this is a problem for long running transactions, especially those that modify a lot of tables, which describes our aggregation pretty well. It also has to be at least as high on all standbys as the master which is why I set it on all the machines, not just pgucr.

##### ENVIRONMENTS AFFECTED

ICDS

##### ISSUE TYPE

- Change Pull Request


emord:
do we think that this is due to the following?

> query of a parent table with many children

We have a lot of tables that have lots of children. perhaps we just went over our limits? Will be interesting to see if it affects citus

Rohit25negi:
I wonder how would this fix be maintained as we scale. As we scale, we might need to increase it further. Perhaps citus will help here because number of child tables are reduced.

snopoke:
Looks like this was already rolled out so I'm going to merge it

calellowitz:
@emord
I think definitely child tables are contributing to our higher than average lock utilization, but I am not sure it is an issue. We are fine from a memory perspective so increasing this limit should be harmless.

@Rohit25negi 
Yeah, it is definitely possible we will have to increase this further at some point, although citus removes a lot of the table inheritance that is causing these issues, so it may stop happening after the switch. I don't think it is a huge issue if we do have to raise it later, but I am hoping we won't

PR #2913, opened 2019-06-05
rename citus_cordinator to citus_master

snopoke:
`citus_cordinator` is misspelt and it's long
and harder to type than `citus_master`


sravfeyn:
👍

PR #2916, opened 2019-06-05
Various ansible cleanup

emord:
##### SUMMARY
Various deprecation warnings and cleanup I've found over past couple of days

##### ENVIRONMENTS AFFECTED
none

##### ISSUE TYPE
- Change Pull Request

### DIFF
--- a/src/commcare_cloud/ansible/deploy_postgres.yml
+++ b/src/commcare_cloud/ansible/deploy_postgres.yml
@@ -22,6 +22,7 @@
   roles:
     - {role: pgbouncer, tags: 'pgbouncer'}
 
+# Notes this should be filtered out if the environment does not have remote postgres
 - name: Remote PostgreSQL (e.g. Amazon RDS)
   hosts: django_manage
   tasks:
--- a/src/commcare_cloud/ansible/roles/java/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/java/tasks/main.yml
@@ -1,6 +1,7 @@
 ---
 - name: installing repo for Java 8+ in Ubuntu
   apt_repository: repo='ppa:openjdk-r/ppa'
+  when: ansible_distribution_version == '14.04'
 
 - name: Install Java
   become: yes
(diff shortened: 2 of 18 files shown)


src/commcare_cloud/ansible/deploy_postgres.yml
```diff
@@ -22,6 +22,7 @@
   roles:
     - {role: pgbouncer, tags: 'pgbouncer'}
 
+# Notes this should be filtered out if the environment does not have remote postgres
```
emord: right now this gets deployed to all servers even if we don't have anything in remote_postgresql @dannyroberts was that intentional?

src/commcare_cloud/ansible/roles/java/tasks/main.yml
```diff
@@ -1,6 +1,7 @@
 ---
 - name: installing repo for Java 8+ in Ubuntu
   apt_repository: repo='ppa:openjdk-r/ppa'
+  when: ansible_distribution_version == '14.04'
```
emord: openjdk-8 is already in bionic https://packages.ubuntu.com/bionic/openjdk-8-jdk so I think this should just be removed

src/commcare_cloud/ansible/deploy_postgres.yml
dannyroberts: No that's not intentional, good catch. In it's original version, this only had one task, which ran with an empty `with_items` if there was nothing in `remote_postgresql`. But then on May 20 I added the "Install python-psycopg2 to be able to set up remote postgresql" task to fix the install (something must have previously installed this on django_manage that doesn't anymore), not realizing that that adds an unnecessary install when there's nothing in `remote_postgresql`. If adding `when: groups.remote_postgresql is defined` or the the like to the block works, I support that

src/commcare_cloud/ansible/roles/java/tasks/main.yml
dannyroberts: You mean it should be removed once we're officially dropping trusty? I agree with that

PR #2917, opened 2019-06-05
remove rabbitmq requirement

emord:
##### SUMMARY
Removes rabbitmq as a hard requirement

##### ENVIRONMENTS AFFECTED
none

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
rabbitmq/couchdb

##### ADDITIONAL INFORMATION
As noted in https://dimagi-dev.atlassian.net/browse/SLTNS-192 the [erlang dependencies for couchdb2](https://github.com/dimagi/commcare-cloud/pull/2886/commits/24f83b6bca5aa08720c6d89c2a30ea97af41615c#diff-3402d5d869f77f376e1694a0fffa614aR8) are in conflict with the erlang versions needed to run rabbitmq on bionic, which means rabbitmq and couchdb cannot be run on the same server. 

My solution to this is to remove the requirement to have rabbitmq at all and instead use redis as the broker. This is our [default in commcare-hq](https://github.com/dimagi/commcare-hq/blob/master/settings.py#L512), though it does mean you can lose tasks if your tasks take too long and redis begins evicting keys. This should be fine on small, monolith deployments

### DIFF
--- a/src/commcare_cloud/ansible/roles/rabbitmq/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/rabbitmq/tasks/main.yml
@@ -9,7 +9,7 @@
 
 - name: Install rabbitmq
   become: true
-  apt: pkg=rabbitmq-server state=installed
+  apt: name=rabbitmq-server state=present
   notify:
   - restart rabbitmq
 
--- a/src/commcare_cloud/environment/main.py
+++ b/src/commcare_cloud/environment/main.py
@@ -374,7 +374,6 @@ class Environment(object):
             mapping[self.format_sshable_host(ansible_host, ansible_port)] = host.name
         return mapping
 
-
     @memoized_property
     def hostname_map(self):
         """Mapping of inventory hostname (IP) to assigned hostname"""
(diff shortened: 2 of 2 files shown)


emord:
@kaapstorm "too long" depends on the following variables:

* amount of memory given to redis
* number of tasks queued
* size of tasks queued (like if those tasks have full copies of `Application`)
* amount redis is being used by HQ

nickpell:
Looks good - getting off of RabbitMQ might unblock us from upgrading to the latest celery version too:
https://github.com/celery/celery/issues/4980

emord:
In case anyone wants more context on this PR:

I'm in no way suggesting we should get off rabbitmq. If we did, we'd probably want a separate redis instance for celery, or look into their "experimental" zookeeper support. I'm purely doing this to get around the issue of conflicting erlang versions when trying to install rabbitmq and couchdb on the same  server

Any Dimagi managed server should definitely still use rabbitmq if possible

nickpell:
Thanks for the clarification

PR #2919, opened 2019-06-05
Do not attempt to remove encrypted tmp

emord:
##### SUMMARY
Does not attempt to unmount /opt/tmp on running instances

##### ENVIRONMENTS AFFECTED
monoliths (pna/swiss)

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
/tmp

##### ADDITIONAL INFORMATION
I think if we were to deploy-stack on one of the monoliths this will fail, because commcare processes need to access /opt/tmp and you can't unmount a drive that's being used.

This was originally introduced in https://github.com/dimagi/commcare-cloud/pull/1918 so that we could cleanup database servers that had this folder deployed accidentally. I don't think it has any use now and causes issues on monoliths


snopoke:
I was thinking about these cleanup tasks that we have littered all over the place and I think in future a better pattern would be to create a single use playbook which can be run on relevant environments and removed once no longer needed.

emord:
@snopoke I agree with that, or just having a `cleanup` role in ansible for this sort of thing similar to what we have in HQ

PR #2920, opened 2019-06-05
fix npm issues in bionic

emord:
##### SUMMARY
Node and NPM can't be co-installed with couchdb on bionic

##### ENVIRONMENTS AFFECTED
none (future bionic monoliths)

##### ISSUE TYPE
- Change Pull Request

##### ADDITIONAL INFORMATION
couchdb2 requires libcurl4-openssl-dev which requires libssl-dev (happens to be version 1.1.0 in bioinic)

npm (through node-gyp and I think one other dependency) requires libssl1.0.0-dev which is incompatible with libssl-dev

installing npm forces libcurl4 to uninstall and I have no idea what effect that would have on couch. So as a resolution, I'm doing something similar to what we've been doing in trusty. Install from the official node source repository. When installing from that repository, it also includes npm so we don't need to install it separately package information: https://deb.nodesource.com/node_8.x/dists/bionic/main/binary-amd64/Packages

### DIFF
--- a/src/commcare_cloud/ansible/roles/common_installs/tasks/bionic_apt_installs.yml
+++ b/src/commcare_cloud/ansible/roles/common_installs/tasks/bionic_apt_installs.yml
@@ -3,7 +3,6 @@
   apt:
     name:
       - software-properties-common
-      - npm
       - python-minimal
       - python3-pip
       - ufw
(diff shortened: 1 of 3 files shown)


emord:
I'm not planning to proactively deploy this change to old bionic servers. There is currently a working node and npm setup on non-couch nodes,and this new repo is only required when we want couchdb2 to also include node. @dannyroberts let me know if htis is an issue, and I can schedule some time to add the repository to current servers (just india nad staging?)

dannyroberts:
Is the concern having different versions of node installed on different machines?

PR #2922, opened 2019-06-05
Don't create staticfiles

emord:
Opening this up to try to figure out what to do about this.

Right now this attempts to create a `/home/cchq/www/env/current/staticfiles`. This causes linking current to relelases/release_date to fail later on when commcare gets deployed https://github.com/dimagi/commcare-cloud/blob/2bade44f6382e0b6bfd4295822ceae68580c6b05/src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml#L58-L65

Removing this may cause nginx to fail to start if there's no staticfiles until there's a real deploy. At the moment I'm considering moving this into the commcarehq role after linking. I'm not sure if that would cause any other issues though. thoughts on taht approach?

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/tasks/install.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/tasks/install.yml
@@ -25,9 +25,19 @@
   become: yes
   user: name="{{ nginx_user }}" createhome=no password=no state=present groups="{{ cchq_user }}"
 
+- stat:
+    path: "{{ code_home }}"
+  register: code_home_folder
+
 - name: Create static files home
   become: yes
-  file: path="{{ nginx_static_home }}" owner="{{ cchq_user }}" group="{{ cchq_user }}" mode=0755 state=directory
+  file:
+    path: "{{ nginx_static_home }}"
+    owner: "{{ cchq_user }}"
+    group: "{{ cchq_user }}"
+    mode: 0755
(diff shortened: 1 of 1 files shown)


dannyroberts:
Hmm. `deploy_commcarehq.yml` runs before `deploy_proxy.yml` (https://github.com/dimagi/commcare-cloud/blob/77a48cbec6dd98144c9e81583356ac4de958dd02/src/commcare_cloud/ansible/deploy_stack.yml#L8-L9) so this must be coming from `deploy_db.yml` > `deploy_couchdb2.yml`?

emord:
yeah. that sounds right. I will look into that. maybe i'lljust make couch default to haproxy for monoliths to work around for now

src/commcare_cloud/ansible/roles/nginx/tasks/install.yml
```diff
@@ -25,9 +25,19 @@
   become: yes
   user: name="{{ nginx_user }}" createhome=no password=no state=present groups="{{ cchq_user }}"
 
+- stat:
+    path: "{{ code_home }}"
+  register: code_home_folder
```
emord: Not super happy with this solution because there's always a possibility (though unlikely) that we could move staticfiles out of code_home. With that said, I don't think that will actually happen, so I like this as a simple safeguard to not creating this folder too early

PR #2926, opened 2019-06-06
Fix redis monit

emord:
##### SUMMARY
Use systemctl to start/stop redis

##### ENVIRONMENTS AFFECTED
staging & india

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
redis

##### ADDITIONAL INFORMATION
the monit config for redis is currently broken as bionic uses systemd as a process manager and our redis library only installs /etc/init.d if systemd is not the process manager https://github.com/DavidWittman/ansible-redis/blob/cfe5d92a721b5b137bbbff20c57e038c5691c463/tasks/server.yml#L9-L22

monit status is currently green on staging and india because it uses the same pid file to check for status. if anyone tries monit start/stop on either of those environments it will fail

A follow up to this may be "do we need monit in a world where we have systemd?" or "how do we simplify our process managemnt?" @dannyroberts I think you'll have the most context on answering that question as part of plat-reliability

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml
@@ -103,23 +103,21 @@
 
 - name: Upgrade python packages
   pip:
-    name: "{{ item }}"
+    name:
+      - six
+      - pip
+      - setuptools
     state: latest
     virtualenv: "{{ virtualenv_home }}"
-  with_items:
-    - six
-    - pip
-    - setuptools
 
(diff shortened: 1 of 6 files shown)


emord:
bump

dannyroberts:
@emord I don't know off the top of my head how/whether systemd makes monit redundant, but if it does and it would mean we can simplify our stack, that's worth looking into. Creating https://dimagi-dev.atlassian.net/browse/PLAT-168 as a placeholder for following up

PR #2928, opened 2019-06-07
Add swap to celery1-india to avoid OOMs

dannyroberts:
##### SUMMARY
Celery is mostly powered just find on india, but experiences problems that it does not gracefully recover from when memory occasionally spikes and causes a process to be killed by the OOM killer. The last couple day's it's been causing the background_queue to get blocked until someone restarts it

##### ENVIRONMENTS AFFECTED
india

### DIFF
--- a/environments/india/inventory.ini
+++ b/environments/india/inventory.ini
@@ -15,7 +15,7 @@
 [airflow1]
 10.203.10.63 hostname=airflow1-india ufw_private_interface=ens5 ansible_python_interpreter=/usr/bin/python3
 [celery1]
-10.203.10.13 hostname=celery1-india ufw_private_interface=ens5 ansible_python_interpreter=/usr/bin/python3
+10.203.10.13 hostname=celery1-india ufw_private_interface=ens5 ansible_python_interpreter=/usr/bin/python3 swap_size=4G
 [couch2]
 10.203.40.187 hostname=couch2-india ufw_private_interface=eth0
 [couch3]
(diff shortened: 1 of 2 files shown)


dannyroberts:
This is rolled out now

millerdev:
I think I'd rather see us add more RAM than swap. If it's going down regularly then it probably doesn't have enough RAM, and adding swap will cause things to get slower during times of high RAM usage (and slowness is often hard to debug). Maybe we don't care if this cluster gets slow and would rather trade performance for operating cost 💰?

dannyroberts:
Yeah my take is that if we can hit reasonable time-to-start and time-to-run metrics by adding swap instead of RAM, I'd at least try that before adding RAM. I think for like 99% of the day we're fine on RAM, but the occasional spike will put us over and kill a process, which then causes other problems.

That said, https://app.datadoghq.com/dashboard/nbz-bie-nip/host-groups?screenId=nbz-bie-nip&screenName=host-groups&abstraction_level=1&aggregate_up=false&display_timeline=false&from_ts=1559579412852&is_auto=false&live=false&page=0&per_page=30&tile_size=m&to_ts=1560184212852&tpl_var_environment=india&tpl_var_host_group=celery&use_date_happened=true is showing that the swap has already filled up, and I see another OOM about 22h ago. So, not the results I was rooting for, and either better use of memory in dividing the queues or increasing RAM is going to be necessary.

I'm going to go ahead and merge this to reflect the situation, but I don't consider https://dimagi-dev.atlassian.net/browse/PLAT-167 resolved

PR #2929, opened 2019-06-07
Increase production redis_maxmemory to 16G (of 32G machine)

dannyroberts:
##### SUMMARY

I'm having trouble finding what the benefits of setting this at all are over setting to 0 (which means as much memory as the underlying OS allows), but it's currently set very low and that seems like a plausible explanation for occasional evictions and generally low % of unfreeable memory (https://app.datadoghq.com/dashboard/nbz-bie-nip/host-groups?screenId=nbz-bie-nip&screenName=host-groups&abstraction_level=1&aggregate_up=false&display_timeline=false&from_ts=1546365120000&is_auto=false&live=false&page=0&per_page=30&tile_size=m&to_ts=1559926320000&tpl_var_environment=production&tpl_var_host_group=redis&use_date_happened=true).

##### ENVIRONMENTS AFFECTED
production

### To roll out ###
It will require restarting the redis service, which may cause 500s while it blocks to restore the AOF into memory
- [x] `cchq production deploy-stack --limit redis --tags=swap`
- [ ] `cchq production ansible-playbook deploy_db.yml --limit redis --branch=dmr/production-redis-maxmemory --start-at-task='create redis config file'` <--- 500s

### DIFF
--- a/src/commcare_cloud/ansible/group_vars/redis.yml
+++ b/src/commcare_cloud/ansible/group_vars/redis.yml
@@ -6,9 +6,9 @@ redis_local_facts: false
 redis_logfile: "{{ redis_dir }}/redis.log"
 redis_service_name: redis-server
 redis_syslog_enabled: "yes"
-# Best practice: set redis_maxmemory to half of the RAM on redis
+# This implements the best practice of setting redis_maxmemory to half the RAM
 # Discussion: https://github.com/dimagi/commcare-cloud/pull/2929#issuecomment-499970466
-redis_maxmemory: 4gb
+redis_maxmemory: "{{ ansible_memory_mb.real.total // 2 }}m"
 redis_maxmemory_policy: volatile-lru
 redis_save: []
 redis_appendonly: "yes"
--- a/src/commcare_cloud/ansible/group_vars/redis.yml
+++ b/src/commcare_cloud/ansible/group_vars/redis.yml
@@ -6,6 +6,8 @@ redis_local_facts: false
 redis_logfile: "{{ redis_dir }}/redis.log"
 redis_service_name: redis-server
 redis_syslog_enabled: "yes"
+# Best practice: set redis_maxmemory to half of the RAM on redis
+# Discussion: https://github.com/dimagi/commcare-cloud/pull/2929#issuecomment-499970466
 redis_maxmemory: 4gb
 redis_maxmemory_policy: volatile-lru
 redis_save: []
--- a/environments/echis/public.yml
+++ b/environments/echis/public.yml
@@ -10,6 +10,7 @@ elastcsearch_backup_days: 3
 elasticsearch_version: 1.7.6
 elasticsearch_download_sha256: 78affc30353730ec245dad1f17de242a4ad12cf808eaa87dd878e1ca10ed77df.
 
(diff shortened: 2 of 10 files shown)


emord:
Its generally not safe to use more than 1/2 the memory available for redis. @calellowitz has tried similar ones before such as https://github.com/dimagi/commcare-cloud/pull/1865 and it caused OOM because when there is significant fragmentation in redis it rewrites valid keys to new parts of memory and does a switch over of the memory space its using. Its similar to couchdb's free disk space requirements. Theoretically you can get by using more than 1/2, but there's always a chance you lose everything.

Sorry for the lack of documentation, I was attempting to look through old PRs and could not find anything more appropriate than the previous one

calellowitz:
I can't find great docs on this, although a lot of it is https://redis.io/topics/persistence and on the linked pages about `BGREWRITEAOF`, but as emord says, redis rewriting the append only file can cause it to use up to double the memory it was using before. In the past this caused issues because the rewrite process was killed by the OOM killer, and the temporary files were not cleaned up. This process was repeated over and over until the disk filled. There is a bit more of a discussion around that in the PR and the retro that went out at that time.

calellowitz:
That all being said, doubling it to 16gb seems both safe and prudent at this point

dannyroberts:
Thank you guys so much for the knowledge and digging up these references. I'll make a note and link to this discussion in the code so the next person has the same context.

Another question: given the bad failure mode of ever hitting the OOM killer, what do you guys think of adding 1GB of swap as well? My logic is that since redis memory is fixed to half of the RAM anyway and there's little else running here, it's unlikely to ever need swap in normal conditions, but that if that we're using peak memory and then happen to run some other temp process (via ansible or something) that uses a bit of memory, swap would be there to prevent an OOM. Any thoughts on that? The thing we clearly don't want is for redis to ever end up routinely putting cache memory in swap.

emord:
that seems reasonable @calellowitz Also linked me to https://github.com/dimagi/commcare-cloud/pull/2789#discussion_r276728801

I think that a catch like https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml#L3-L5 may be useful?

src/commcare_cloud/ansible/group_vars/redis.yml
```diff
@@ -6,7 +6,9 @@ redis_local_facts: false
 redis_logfile: "{{ redis_dir }}/redis.log"
 redis_service_name: redis-server
 redis_syslog_enabled: "yes"
-redis_maxmemory: 4gb
+# This implements the best practice of setting redis_maxmemory to half the RAM
+# Discussion: https://github.com/dimagi/commcare-cloud/pull/2929#issuecomment-499970466
+redis_maxmemory: "{{ ansible_memory_mb.real.total // 2 }}m"
```
emord: Can you add 4gb to the environment files that didn't have this defined before? I'm mostly concerned about swiss/pna which have 32 GB of memory, but we don't want redis to use 16 GB

dannyroberts: Yes that's a very good point. Thanks

PR #2933, opened 2019-06-10
don't add citus to shared_preload_libraries to fix pg load errorbefor…

sravfeyn:
Because citus extension is installed way after, adding citus to `shared_preload_libraries` before that makes postgres run into startup error. This PR changes is so that citus library is added to the conf right just before citus installation

@snopoke


src/commcare_cloud/ansible/roles/citusdb/tasks/main.yml
```diff
@@ -20,6 +20,10 @@
    apt: 
     name: "{{citus_binary}}"
     state: present
+
+ - name: Add citus to postgres prelod libraries
+   shell: "pg_conftool 11 main set shared_preload_libraries 'citus, pg_stat_statements'"
```
snopoke: this should use `{{postgresql_version}}`

sravfeyn: Updated

PR #2936, opened 2019-06-10
Set datavol_device automatically on AWS envs

dannyroberts:
##### SUMMARY
Set `datavol_device` automatically on AWS envs based on `block_device` for the server in `terraform.yml`. This makes it so you just have to say in `terraform.yml` that a server should have a block device, and then it'll automatically be given the inventory variable it needs to make use of it.

##### ENVIRONMENTS AFFECTED
Just aws envs

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -217,17 +217,17 @@ es_master
 es_data
 
 [couch3]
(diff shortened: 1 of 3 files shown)


dannyroberts:
FYI @sanjay2916

PR #2939, opened 2019-06-10
Fix check_s3_backup.py script and change its check

dannyroberts:
it now checks for a recent daily and weekly backup, prints out whether
it found them, and exits with -1 if there isn't a recent daily backup or
there isn't a recent weekly backup

I'm not sure what the most sensical way to hook this check up to a datadog monitor is, but I figured I'd start by fixing the broken script that was clearly meant to do something like what I'm trying to do

### DIFF
--- a/src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
+++ b/src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 import argparse
 import boto3
+import botocore
 import datetime
 
 parser = argparse.ArgumentParser(description="S3 Backup Status.")
@@ -34,7 +35,13 @@ s3 = session.resource(service_name='s3',endpoint_url=endpoint,region_name='{{ aw
 s3 = session.resource(service_name='s3',endpoint_url=endpoint)
 {% endif %}
 
-bucket = s3.Bucket(BUCKET_NAME)
+def file_exists_in_s3(object_key):
+  try:
+    s3.Object(BUCKET_NAME, object_key).load()
+  except botocore.exceptions.ClientError:
+    return False
+  else:
+    return True
 
 today = datetime.datetime.today()
 
@@ -56,28 +63,26 @@ weekly_options = [
     for n in [0, 1, 2, 3, 4, 5, 6]
 ]
 
+# up to one match will be stored in these
 daily_matches = []
 weekly_matches = []
-# Shows Alert if backup is not uploaded since 2 days.
-for obj in bucket.objects.all():
-    if obj.key in daily_options:
-        daily_matches.append(obj)
-    if obj.key in weekly_options:
-        weekly_matches.append(obj)
+
+for options, matches in [(daily_options, daily_matches),
+                         (weekly_options, weekly_matches)]:
+    for filename in options:
+        if file_exists_in_s3(filename):
+            matches.append(filename)
+            break
 
 if daily_matches:
-    daily_matches.sort(key=lambda obj: obj.last_modified)
     match = daily_matches[-1]
-    print('[OK] Last daily {} backup is {}, uploaded @ {}'.format(
-        args.service, match.key, match.last_modified))
+    print('[OK] Last daily {} backup is {}'.format(args.service, match))
 else:
     print('[ALERT] No daily {} backup in last two days'.format(args.service))
 
 if weekly_matches:
-    weekly_matches.sort(key=lambda obj: obj.last_modified)
     match = weekly_matches[-1]
-    print('[OK] Last weekly {} backup is {}, uploaded @ {}'.format(
-        args.service, match.key, match.last_modified))
+    print('[OK] Last weekly {} backup is {}'.format(args.service, match))
 else:
     print('[ALERT] No weekly {} backup in last seven days'.format(args.service))
 
--- a/src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
+++ b/src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
@@ -39,8 +39,22 @@ bucket = s3.Bucket(BUCKET_NAME)
 today = datetime.datetime.today()
 
 backup_filename_format = '{prefix}_{backup_type}_{date}.tar.gz'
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
```diff
@@ -3,24 +3,67 @@ import argparse
...
+  'postgres': "postgres_{{ deploy_env }}",
+  'blobdb': "blobdb",
+}
+
+BUCKET_NAME = BUCKET_MAP[args.service]
+PREFIX = PREFIX_MAP[args.service]
+
+endpoint = {{ aws_endpoint|tojson }}
+
+if not endpoint.startswith("http"):
```
Rohit25negi: do you expect this script to work with http or only with https?

src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
```diff
@@ -3,24 +3,67 @@ import argparse
...
+    print('[ALERT] No daily {} backup in last two days'.format(args.service))
+
+if weekly_matches:
+    weekly_matches.sort(key=lambda obj: obj.last_modified)
+    match = weekly_matches[-1]
+    print('[OK] Last weekly {} backup is {}, uploaded @ {}'.format(args.service, match.key, match.last_modified))
+else:
+    print('[ALERT] No weekly {} backup in last seven days'.format(args.service))
+
+if not weekly_matches and not daily_matches:
```
Rohit25negi: According to the description of the PR. It should be `or` if you want to raise (-1) in case either of the one is not done. 
this will only raise -1 only when both of them  are not done.

src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
```diff
@@ -3,24 +3,67 @@ import argparse
...
-# Shows Alert if backup is not uploaded since 2 days. 
-for key in bucket.objects.all():
-    if today.day - key.last_modified.day > 2:
-        print "[ALERT]" , key.key , key.last_modified
-    else:
-        print "[OK]", key.key , key.last_modified
\ No newline at end of file
+backup_filename_format = '{prefix}_{backup_type}_{date}.tar.gz'
+daily_options = [backup_filename_format.format(prefix=PREFIX, backup_type='daily', date=(today - datetime.timedelta(days=n)).strftime('%Y_%m_%d')) for n in [0, 1]]
+weekly_options = [backup_filename_format.format(prefix=PREFIX, backup_type='weekly', date=(today - datetime.timedelta(days=n)).strftime('%Y_%m_%d')) for n in [0, 1, 2, 3, 4, 5, 6]]
```
millerdev: nit: the length of these lines makes it hard to see what's going on. Consider wrapping.

src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
```diff
@@ -3,24 +3,67 @@ import argparse
...
+
+daily_matches = []
+weekly_matches = []
+# Shows Alert if backup is not uploaded since 2 days.
+for obj in bucket.objects.all():
```
millerdev: Could this take a long time if the bucket contains a lot of items? What do you think of checking for existence of specific items in the bucket instead?

src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
dannyroberts: It should work with either. It's just that "https://..." also starts with "http"

src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
dannyroberts: You're totally right, thanks for catching that

src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
dannyroberts: Good point. For the use cases I know of, after a year of saving daily and weekly backups for 3 services, we'd be at around 1000 objects in the bucket, which seems like around the point where this might be worth worrying about. I'll take a pass at switching this to asking just for the ~9 objects it's looking for like you suggested

dannyroberts:
Ok I think all the comments are addressed. Thanks for the thoughtful review guys!

src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
```diff
@@ -34,7 +35,13 @@ s3 = session.resource(service_name='s3',endpoint_url=endpoint,region_name='{{ aw
 s3 = session.resource(service_name='s3',endpoint_url=endpoint)
 {% endif %}
 
-bucket = s3.Bucket(BUCKET_NAME)
+def file_exists_in_s3(object_key):
+  try:
```
millerdev: two-space indentation is inconsistent with other indentation in this file

src/commcare_cloud/ansible/roles/backups/templates/check_s3_backup.py.j2
```diff
@@ -34,7 +35,13 @@ s3 = session.resource(service_name='s3',endpoint_url=endpoint,region_name='{{ aw
...
+    s3.Object(BUCKET_NAME, object_key).load()
+  except botocore.exceptions.ClientError:
```
millerdev: nit: not sure how much it matters here, but `ClientError` does not necessarily mean "not found." Here's [how the blob db detects that condition
](https://github.com/dimagi/commcare-hq/blob/03499156b26de39121531b82e7ddbd5aa5405783/corehq/blobs/s3db.py#L184-L186) (still kinda messy)

```py
def is_not_found(err, not_found_codes=["NoSuchKey", "NoSuchBucket", "404"]):
    return (err.response["Error"]["Code"] in not_found_codes or
        err.response.get("Errors", {}).get("Error", {}).get("Code") in not_found_codes)
```

dannyroberts: I think in our case I'm okay with the mild sloppiness here. If we end up catching errors here that aren't because the file is missing, the worst that'll happen is that we get a false positive here. I might regret this later but I think for now I'll just leave this part as-is

PR #2940, opened 2019-06-10
Consolidate queues on india for lower memory usage

dannyroberts:
##### SUMMARY
A second attempt at addressing https://dimagi-dev.atlassian.net/browse/PLAT-167 memory issues.

Already rolled this out

##### ENVIRONMENTS AFFECTED
india


sravfeyn:
FYI @dannyroberts this caused the particular queues to be down. I didn't get chance to look into the root cause yet, but there's some more context in the interrupt slack channel. I ended up doing `update-supervisor-confs` with master, which fixed the issue. So, this change needs to be redeployed after addressing the issue.

dannyroberts:
This caused issues over night so i'm going to take a different approach

PR #2941, opened 2019-06-11
Uninstall existing Python 3.6 so it can be replaced with 3.6.8 from PPA

nickpell:
Not sure if there's a more direct way to install the right package, but this works.

@dimagi/py3

### DIFF
--- a/src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
@@ -30,6 +30,12 @@
   apt: update_cache=yes
   when: add_mosh_repo is changed or add_postgres_repo is changed or add_pango_repo is changed or add_python_ppa_trusty is changed or add_python_ppa_bionic is changed
 
+- name: Uninstall existing Python 3.6 so it can be replaced with 3.6.8 from PPA
+  apt:
+    name: python3.6
+    state: absent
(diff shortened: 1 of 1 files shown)


millerdev:
@nickpell Thinking about this more, is Python 3.6 that was uninstalled here the "system Python" on Ubuntu 18.04? If it is, it's probably not safe to uninstall it (as we found out). My experience from Mac OS is to never mess with the system Python in any way, but instead install Python(s) to be used by stuff I'm doing in a separate place that does not interfere with the default system installation. The [first article I found](https://websiteforstudents.com/installing-the-latest-python-3-7-on-ubuntu-16-04-18-04/) on the topic seems to corroborate this advice:

> Do not use the standard make install as it will overwrite the default system python3 binary…

...which suggests that messing with the system Python is a bad idea.

We may want to use [ppa:deadsnakes/ppa](https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) to install alternate versions of Python if the default system version does not meet our needs (although the security warning there is a little intimidating). Are there better ways to install non-default versions of Python that are safe to use in production environments?

nickpell:
@millerdev We're currently set up to use ppa:deadsnakes/ppa on trusty and ppa:jonathonf/python-3.6 on bionic and xenial.

The alternative I've found is to build Python from scratch.  Unfortunately that won't give us automatic security updates either.

PR #2942, opened 2019-06-11
Reduce concurrency on india celery queues to lower memory usage

dannyroberts:
##### SUMMARY
I'm not sure if this will work / still be enough processing power, but I wanted to try one last thing before throwing money at it

##### ENVIRONMENTS AFFECTED
india

### DIFF
--- a/environments/india/app-processes.yml
+++ b/environments/india/app-processes.yml
@@ -18,10 +18,10 @@ celery_processes:
     ucr_queue,ucr_indicator_queue,celery,export_download_queue,reminder_rule_queue,case_import_queue,case_rule_queue,icds_dashboard_reports_queue,async_restore_queue,celery_periodic:
       concurrency: 3
     saved_exports_queue:
-      concurrency: 2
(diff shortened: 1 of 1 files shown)


environments/india/app-processes.yml
```diff
@@ -18,10 +18,10 @@ celery_processes:
     ucr_queue,ucr_indicator_queue,celery,export_download_queue,reminder_rule_queue,case_import_queue,case_rule_queue,icds_dashboard_reports_queue,async_restore_queue,celery_periodic:
       concurrency: 3
     saved_exports_queue:
-      concurrency: 2
+      concurrency: 1
```
Rohit25negi: hey, what is the intent of reducing the concurrency of these three queues?

dannyroberts: The goal is to lower memory usage, to avoid hitting out-of-memory errors which randomly kill process and can get us into a stuck state where a queue is not getting processed but also not restarting.

PR #2943, opened 2019-06-12
add citusdb to django settings

snopoke:


environments/icds-cas/postgresql.yml
```diff
@@ -92,6 +93,13 @@ dbs:
       host: pgnondashboarducr0
       query_stats: True
       django_migrate: False
+    - django_alias: icds-ucr-citus
+        name: icds_ucr
+        django_migrate: True  # ICDS dashboard models are stored here
+        host: citusmaster0
+        pg_config:
+          - name: citus.shard_count
+            value: 64
```
snopoke: This set's how many shards will be created. Each shard is a table on a worker node. The default is 32. I think 64 should be enough but I'd like to get other's opinions.

snopoke:
Rolled out

PR #2946, opened 2019-06-12
Start supervisor on boot

proteusvacuum:
##### SUMMARY

When using systemd, on ubuntu 18.04 we need to enable supervisor to start on boot. 
I think this wasn't happening on 14.04 because upstart allows you to specify this directly: https://github.com/dimagi/commcare-cloud/blob/97b5ff49992e137131bca2dd66782f40574805fe/src/commcare_cloud/ansible/roles/supervisor/templates/supervisord.upstart.conf.j2#L3

### DIFF
--- a/src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
@@ -79,7 +79,7 @@
   failed_when: result.rc > 1
   changed_when: result.rc == 0
 
-- name: start supervisord
+- name: start and enable supervisord
   become: yes
-  service: name=supervisord enabled=yes
+  service: name=supervisord enabled=yes state=started
   when: which_supervisord.stdout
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
```diff
@@ -81,5 +81,5 @@
 
 - name: start supervisord
   become: yes
-  service: name=supervisord state=started
+  service: name=supervisord enabled=yes
```
dannyroberts: Does `enabled=yes` also imply `state=started`? I.e. in addition to making it start on boot, does `enabled=yes` make it start _now_? I tried to find where it said that in https://docs.ansible.com/ansible/latest/modules/service_module.html but maybe I just missed it.

proteusvacuum: ah, good point. I didn't test if that happened. updated.

PR #2947, opened 2019-06-12
Fix meaningless terraform diffs

dannyroberts:
##### SUMMARY

These two otherwise unrelated issues both were causing meaningless terraform diffs—diffs that showed up again even immediately after cleanly applying. This PR gets rid of those diffs, and other than removing unneeded resources/settings doesn't make any meaningful changes to our environments.

##### ENVIRONMENTS AFFECTED
production, india, staging

### DIFF
--- a/src/commcare_cloud/terraform/modules/network/main.tf
+++ b/src/commcare_cloud/terraform/modules/network/main.tf
@@ -79,6 +79,7 @@ resource "aws_vpn_connection_route" "vpn_connections" {
 }
 
 resource "aws_security_group" "vpn_connections" {
+  count = "${length(var.vpn_connections) == 0 ? 0 : 1}"
   name = "vpn-connections-sg-${var.env}"
   vpc_id = "${aws_vpc.main.id}"
   ingress {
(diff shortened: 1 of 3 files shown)


dannyroberts:
FYI @sanjay2916 on india this will introduce the following diffs:

```
  - module.network.aws_security_group.vpn_connections
```
and a diff like this for each server:
```
  ~ module.server__airflow1-india.aws_instance.server
      vpc_security_group_ids.#:          "3" => "2"
      vpc_security_group_ids.2859696134: "sg-00854e4d6441f81d1" => "sg-00854e4d6441f81d1"
      vpc_security_group_ids.710633306:  "sg-03e1d9374b9bd2b55" => "sg-03e1d9374b9bd2b55"
      vpc_security_group_ids.806092854:  "sg-03b1a640746791156" => ""
```
which removes the old vpn_connections security group (which, as per the first diff, no longer exists). It's okay to remove because we no longer maintain the VPN tunnel to softlayer, which is what that security group was for.

dannyroberts:
bump :)

PR #2953, opened 2019-06-14
add missing sikkim and new asm

mkangia:
##### SUMMARY
Adds two new locations on nginx for file sharing on ICDS

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request

##### ADDITIONAL INFORMATION
This is already deployed.
I had to only add for Assamese but i found that sikkim one was missing in commcare-cloud but was present on the server so i added that as well.
fyi @Rohit25negi i think sikkim was added by you last time, pinging to confirm that you didn't add it to commcare-cloud and so it should be present now


mkangia:
Thanks @emord

PR #2954, opened 2019-06-14
Retry until things come up

proteusvacuum:
##### SUMMARY

Adds retries to places where tasks were failing when certain services hadn't yet come up after the previous command.


proteusvacuum:
There are still a few issues to fix with the monolith deploy after this, but we're getting somewhere...

proteusvacuum:
No, it retries a maximum of 5 times before failing completely

dannyroberts:
This is a musing-level comment. Seeing how consistent the pattern is, I wonder if

```
until: not (result.failed and 'process not presently configured with monit' in result.msg)
```
would be more readable and less prone to typos than
```
until: not result.failed or 'process not presently configured with monit' not in result.msg
```
as a convention.

PR #2956, opened 2019-06-15
Add new production couchdb2 bionic machines

dannyroberts:
##### SUMMARY
Add three new machines for couchdb2 with Bionic installed. After recently discovering the 2.1.1 view compaction issue and manually working around it, the *total* disk usage on all couch nodes is now less than 5TB—down from around 28TB! These new nodes have disks adjusted down to the new expected usage. I'm also simplifying the setup to have 3 nodes. Because of the specific node sizes offered on AWS, the combined RAM and CPU of the 3 new nodes will be 12.5% greater than the combined RAM and CPU of the old nodes. The additional cost is minor ($372/mo) compared to what we're saving on disk ($3695/mo on disk + ~$2000/mo on snapshot usage = ~$5700/mo), and I think the potential for improved cluster performance—especially while we have ongoing couch-to-sql migrations—plus the simplified setup (fewer nodes to manage) makes it worth the cost.

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
@sanjay2916 FYI

dannyroberts:
OK this is all done. I've shut down the old nodes, and I will finally terminate them in a few days and make a second PR for that

PR #2963, opened 2019-06-18
Don't change nginx pid location to the same file

proteusvacuum:
The default location for this pid file is /run/nginx.pid
/var/run is symlinked to /run on Ubuntu.

When running the whole install playbook, we run into this bug since when nginx
is installed it has the pid file set as /run/nginx.pid;
https://trac.nginx.org/nginx/ticket/796

##### SUMMARY

If the `deploy-proxy` playbook is run on environments where nginx is already existing, reloading nginx might fail after this change. `service nginx stop` and `service nginx start` is required to be run. 

##### ENVIRONMENTS AFFECTED

All

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/src/commcare_cloud/ansible/roles/nginx/templates/nginx.conf.j2
@@ -1,7 +1,7 @@
 # {{ ansible_managed }}
 user {{ nginx_user }};
 worker_processes  {{ ansible_processor_count }};
-pid        /var/run/nginx.pid;
+pid /run/nginx.pid;
 
 worker_rlimit_nofile {{ nginx_worker_rlimit_nofile }};
 


proteusvacuum:
Open for review label because I'm in the process of testing this change.

proteusvacuum:
Removed the label. Tests worked.

emord:
I don't have a strong explanation for this, but I think this may have messed up some of our PID files. prod's couch nginx had this change deployed and on June 20/21 disk usage started increasing quickly. `/var/run/nginx.pid`/`/run/nginx.pid` were empty which caused logs to not be able to rotate https://github.com/dimagi/commcare-cloud/blob/755bd035d7977ff292303f6dccea380f9127f227/src/commcare_cloud/ansible/deploy_proxy.yml#L184 

prod's normal nginx proxy has not had this change and is still rotating and the logs.

proteusvacuum:
@emord this could indeed be the issue!
From my reading the script should:
- Rotate the logs (i.e. move it to a new file)
- Send `USR1` signal to the nginx pid
    - this creates a new log file, and closes the old one. I suppose if this step fails, the old file is still open by nginx and will keep writing to it

If you restart the proxy process, does the pid file exist?

emord:
> If you restart the proxy process, does the pid file exist?

@proteusvacuum I never actually restarted the process because restart/reload was broken because they also rely on the pid being on the file. Instead I added the pid to the file directly. so that those would work.

proteusvacuum:
Hmm, I think I mentioned that in the PR description, but there was probably a better way of signalling it. 
> If the deploy-proxy playbook is run on environments where nginx is already existing, reloading nginx might fail after this change. `service nginx stop` and `service nginx start` is required to be run.

emord:
Ah yes you did. Sorry for the poor reading comprehension.

What do you think of this following change https://github.com/dimagi/commcare-cloud/pull/3055 ?

Stopping and starting nginx, means that every environment would need to go through some downtime or there would have needed to be some effort to put in to mitigate the downtime. I think that that means, this change should have had a changelog associated with it and it should be on the PR author to schedule those events before its merged.

PR #2965, opened 2019-06-20
Get production couch to safety

dannyroberts:
##### SUMMARY
Working branch for P1 couchdb production maintenance

##### ENVIRONMENTS AFFECTED
production


environments/production/terraform.yml
```diff
@@ -243,6 +243,33 @@ servers:
     group: "couchdb2"
     os: bionic
 
+  - server_name: "couch13-production"
+    server_instance_type: c5.9xlarge
+    network_tier: "db-private"
+    az: "a"
+    volume_size: 30
+    block_device:
+      volume_size: 20
```
millerdev: Should these volume sizes be updated? The current volume size is 3000

dannyroberts: Oh yeah thank you. I'll also smooth over our terraform state so that it applies cleanly

PR #2966, opened 2019-06-20
Revert "Uninstall existing Python 3.6 so it can be replaced with 3.6.8 from PPA"

calellowitz:
Reverts dimagi/commcare-cloud#2941
@proteusvacuum @dannyroberts


millerdev:
@nickpell fyi

PR #2967, opened 2019-06-20
reinstall missing packages

calellowitz:
##### SUMMARY

Reinstalls packages uninstalled in the python3.6 removal. 
@dannyroberts @proteusvacuum


dannyroberts:
Is this the "long" list (all packages that are removed after a full install) or the "short" list (all the packages removed at that point in the playbook on first run)?

calellowitz:
This is the list i took from Farid's list under `From full stack deployed "monolith":` in the doc

proteusvacuum:
That is the "long list"

dannyroberts:
Run

```
cchq <env> ansible-playbook reinstall_missing.yml --skip-check
```
on

- [ ] production
- [ ] staging
- [ ] india
- [ ] icds

millerdev:
Reference the note I left on the original PR: https://github.com/dimagi/commcare-cloud/pull/2941#issuecomment-504407375

I think the correct solution is to revert to a state where the default system Python (and all its dependencies) are in the state that they were before the system Python 3 was uninstalled. If this means rebuilding all of the VMs, then that might be the best solution (painful though it may be). Otherwise I am not confident that the upstream security updates will be applied correctly (although I could be wrong about this). Secondly, I don't think we want to maintain a long list of dependencies like this since we are duplicating work done by the Ubuntu LTS packaging system (we don't explicitly need most of those packages except to have a working version of Ubuntu).

calellowitz:
@millerdev 
I think the three of us are all in agreement with that message. This list of packages was created by replicating the original uninstall and getting the full list of packages it removed. It is not meant to be a list of hq dependencies but a full revert of the uninstall. Similarly, we have removed the uninstall from our playbooks (PR was reverted yesterday), and this playbook is not part of deploy stack. It is meant to be a standalone utility to get us out of this mess, but was made a pull request both so code review could occur and so that if third parties had used our install scripts in the past week, we can provide them a solution to the issue.

This process was tested locally on virtualbox monolith deploys as well as staging VMs on AWS, and appears to restore full functionality. It at least allows the machines to reboot with full networking and function in their previous roles, and I have no reason to believe restoring the packages won't restore other functionality. Rebuilding all of these machines from scratch across the four affected environments, will be large undertaking, probably requiring notable downtime, and since we have no reason to believe it is necessary at this point, it seems like a large investment of time for something we do not expect to make a difference.

Whether or not we go down the full rebuild path, I believe this is still an important stopgap to have, just to keep our machines bootable in case they have to be rebooted at some point.

millerdev:
Great, glad to hear we're aligned.

> Rebuilding all of these machines from scratch across the four affected environments, will be large undertaking, probably requiring notable downtime, and since we have no reason to believe it is necessary at this point, it seems like a large investment of time for something we do not expect to make a difference.

I agree, it will be a large undertaking (not sure I agree about requiring downtime). I also agree that running machines in this state is the best thing to do in the short term.

However, I do think it will become increasingly risky in the longer term since we will have a more customized (less like stock Ubuntu) image on which we will be applying security updates, etc. and the likelihood of hitting issues related to that will only increase with time. This is going to be a long-term maintenance burden, and I am in favor of paying off that debt sooner rather than later. In the worst case it can cause issues like what we had yesterday with the couch cluster which ultimately required emergency down time—the worst kind, to be avoided at nearly any cost.

PR #2968, opened 2019-06-21
First deploy gets stuck on preindexing

proteusvacuum:


docs/setup/new_environment.md
```diff
@@ -230,6 +230,20 @@ $ commcare-cloud monolith deploy-stack --skip-check -e 'CCHQ_IS_FRESH_INSTALL=1'
 $ commcare-cloud monolith fab deploy
 ```
 
+The first time you run deploy (and whenever we have changes to our couch indices), it will kick off a "preindex" task. The deploy script will block until this task is complete. If it takes too long, the deploy script will pause, with a message:
```
emord: >  we have changes to our couch indices

I think `preindex_everything` also initializes the ES indicies when they don't exist.

dannyroberts:
That's a little surprising. Do we know what exactly it's indexing? There should be no data a this point if I'm understanding the context

proteusvacuum:
I think it was failing because kafka topics had not been created yet. We should probably add that to the monolith deploy script at some point.

PR #2977, opened 2019-06-25
Use default kafka log retention for icds

emord:
##### SUMMARY
Reverts https://github.com/dimagi/commcare-cloud/pull/2216 which was only intended to be for a short time

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
kafka

##### ADDITIONAL INFORMATION
This should result in roughly half disk space being used.

This does require brief downtime to change because it requires a kafka restart, though I am going to investigate using replicas so that it would not require dowtime

### DIFF
--- a/environments/icds-cas/public.yml
+++ b/environments/icds-cas/public.yml
@@ -32,7 +32,6 @@ redis_maxmemory_policy: allkeys-lru
 redis_auto_aof_rewrite_percentage: "50"
 
 kafka_log_dir: '{{ encrypted_root }}/kafka'
-kafka_log_retention: 1440 # 60 days
 
 KSPLICE_ACTIVE: no
 


environments/icds-cas/public.yml
```diff
@@ -32,7 +32,6 @@ redis_maxmemory_policy: allkeys-lru
 redis_auto_aof_rewrite_percentage: "50"
 
 kafka_log_dir: '{{ encrypted_root }}/kafka'
-kafka_log_retention: 1440 # 60 days
```
sravfeyn: What's the default value for this when it's not set? Is there a danger of not being able to reprocess few days old kafka changes?

emord: The default value is 28 days https://github.com/dimagi/commcare-cloud/pull/2216/files#diff-8d03e4e8b3ffdde0debe5a39f6edbea8R9

I moved it to 60 days as an emergency measure when we were having speed issues with the original object storage

emord:
Will roll this out during next downtime (maybe tomorrow?)

emord:
This has been rolled out

PR #2981, opened 2019-06-27
Add Virtualbox networking instructions

proteusvacuum:


proteusvacuum:
(Self merging but would love your feedback!)

sravfeyn:
Looks good, thanks!

Curious of the use-case for Virtualbox, is it for production use or just for testing (for which we have Vagrant setup though)?

proteusvacuum:
@sravfeyn, yes this is used for testing. Is our use of vagrant documented anywhere? Maybe I missed where we use that to set up monoliths, or maybe I should work on getting a vagrant monolith config set up instead of having these documents.

sravfeyn:
We have some [docs](https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/README.md#local-vagrant-setup) in the repo. Creating a monolith should simply involve creating a new development monolith env using the same Vagrantfile that creates few VMs right now.

sravfeyn:
Essentially most of the documentation that's done on Virtualbox here is automated in the Vagrantfile so one who wants to do a local setup can simply use `vagrant up` and not worry about configuring Virtualbox. It might be good to add a note in the docs here pointing to Vagrant based setup first and then try this if that doesn't work

proteusvacuum:
I'm going to be working on the vagrant scripts so that you can actually create a single monolith in one go. Right now it creates a bunch of unnecessary machines for a monolith.

PR #2983, opened 2019-06-27
Change note about online kafka migration

emord:
##### SUMMARY
Note that kafka migrations can occur online

##### ENVIRONMENTS AFFECTED
none

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
kafka

##### ADDITIONAL INFORMATION
I went through http://kafka.apache.org/082/documentation.html#basic_ops_cluster_expansion a few times to verify that this is the case and I've since rebalanced all of our topics (in batches) across all kafka servers while services were online and no errors came up

### DIFF
--- a/docs/services/kafka.md
+++ b/docs/services/kafka.md
@@ -55,8 +55,7 @@ $ ./kafka-topics.sh --alter --zookeeper=<zookeeper host>:2181 --topic <topci> --
 Pillowtop process configurations as described in the [CommCare docs](https://commcare-hq.readthedocs.io/pillows.html#parallel-processors).
 
 ### Move partitions
-**NOTE**: All processes accessing Kafka should be shutdown prior to following this process to avoid
- errors publishing or consuming Kafka messages.
+**NOTE**: This can be done while all services are online
 
 1. Create the list of topics to rebalance
 


emord:
merging because its just a doc update, but feel free to comment

docs/services/kafka.md
```diff
@@ -55,8 +55,7 @@ $ ./kafka-topics.sh --alter --zookeeper=<zookeeper host>:2181 --topic <topci> --
 Pillowtop process configurations as described in the [CommCare docs](https://commcare-hq.readthedocs.io/pillows.html#parallel-processors).
 
 ### Move partitions
-**NOTE**: All processes accessing Kafka should be shutdown prior to following this process to avoid
- errors publishing or consuming Kafka messages.
+**NOTE**: This can be done while all services are online
```
millerdev: 👍 

What changed that caused us to flip our position on this?

emord: I'm not sure why we had this initial position in the first place which is why I tagged Simon. The way kafka does this by default is quite safe. The (automated) process was:

1. Add partition to new broker as replica
2. Replicate the full partition data to the new broker
3. Elect the new broker as the leader
4. Remove the partition from the old broker

So its quite safe. I could possibly see some issues if your kafka cluster is totally maxed out on resources and you tried to do it at peak loads (later versions add some network throttling), but in general kafka is super light on resources and this is a safe process

PR #2985, opened 2019-06-28
Compact production CouchDB more aggressively

dannyroberts:
##### SUMMARY
Extend the time window during which compaction can happen from 1pm-5:30pm ET to 1pm-1am. In addition, compact at a lower fragmentation threshold. This will allow us to stay at an overall disk usage in steady state.

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -55,6 +55,10 @@ couchdb2:
   username: "{{ localsettings_private.COUCH_USERNAME }}"
   password: "{{ localsettings_private.COUCH_PASSWORD }}"
 
+couchdb_compaction_settings:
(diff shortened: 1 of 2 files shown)


dannyroberts:
Looks like this had the desired effect! https://app.datadoghq.com/dashboard/nbz-bie-nip/host-groups?abstraction_level=1&aggregate_up=false&display_timeline=false&from_ts=1561741120160&is_auto=false&live=false&page=0&per_page=30&tile_size=m&to_ts=1561913920160&tpl_var_environment=production&tpl_var_host_group=couchdb2&use_date_happened=true

PR #2986, opened 2019-07-01
Default GITHUB_APIKEY=None

emord:
Makes local setup a bit easier. @proteusvacuum @dannyroberts Does this seem like a reasonable default?


proteusvacuum:
100000% in agreement. This has been a very confusing step for local hosting partners.

PR #2989, opened 2019-07-02
Shard postgres optimization

emord:
##### SUMMARY
Changes work_mem and WAL sizes to be in line with pgtune

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
postgres


environments/icds-cas/inventory/postgresql.csv
```diff
@@ -73,7 +75,9 @@ shard_dbs,postgresql_maintenance_work_mem,2GB,S,,,,,
 shard_dbs,postgresql_max_connections,800,I,,,,,
 shard_dbs,postgresql_shared_buffers,12GB,S,,,,,
 shard_dbs,postgresql_wal_buffers,16MB,S,,,,,
-shard_dbs,postgresql_work_mem,41943kB,S,,,,,
+shard_dbs,postgresql_work_mem,3276kB,S,,,,,
```
emord: This is a big decrease from 42 MB to 3 MB. There does not appear to be any particular reason for it to have been so high in the first place. This is the original value from the beginning of this cluster. which hasn't been updated

PR #2991, opened 2019-07-03
Add ebsnvme mapping for AWS

dannyroberts:
##### SUMMARY
This is the solution AWS support pointed me to after some back and forth for the NVMe arbitrary-order mount issue. It establishes a rule that symlinks `/dev/sdf` (etc) to the proper nvme device names (the ones that are in unpredictable order), and then it uses the `/dev/sdf` alias throughout, thus establishing a stable handle for each volume.

##### ENVIRONMENTS AFFECTED
production, india (couch machines only)


environments/india/inventory.ini
```diff
@@ -17,11 +17,11 @@
 [celery1]
 10.203.10.13 hostname=celery1-india ufw_private_interface=ens5 ansible_python_interpreter=/usr/bin/python3 swap_size=4G
 [couch5]
-10.203.40.178 hostname=couch5-india ufw_private_interface=ens5 ansible_python_interpreter=/usr/bin/python3 datavol_device=/dev/nvme1n
+10.203.40.178 hostname=couch5-india ufw_private_interface=ens5 ansible_python_interpreter=/usr/bin/python3 datavol_device=/dev/sdf datavol_device1=/dev/sdf is_datavol_ebsnvme=yes
```
millerdev: Why is `datavol_device1` defined here? It also appears to be configured automatically in [src/commcare_cloud/ansible/roles/ecryptfs/defaults/main.yml](https://github.com/dimagi/commcare-cloud/pull/2991/files#diff-8a1586d90c304abac8ef7fc5ea06c793)

dannyroberts: I'm defining it here to override the automatically configured default. The default would have it be `/dev/sdf1`, but I want it to be `/dev/sdf`. I'm pretty sure `datavol_device1` is a bad name but I'm not sure what to call it—I thought of `datavol_partition`, but it doesn't actually have to be a disk partition, it can be any device.

dannyroberts:
I merged this even though that discussion wasn't all the way through—just wanted to make sure we're in a consistent state on master. Still open to (and frankly eager for) suggestions about the naming here

millerdev:
It's fine that you merged. I just wanted to flag that change (where `datavol_device1` was overridden to be the same as  `datavol_device`) in case it was a mistake.

I can't think of a better name earlier, and honestly I thought it was clear enough (it turned out to be what I had guessed it would be when I first saw it in the env config files).

src/commcare_cloud/ansible/ebsnvme/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
+- name: Install nvme-cli
+  become: yes
+  apt: name=nvme-cli state=present
+  notify: "Reload udev rules"
+
+- name: Download ebsnvme
+  become: yes
+  get_url:
+    url: "https://gist.githubusercontent.com/lbernail/d851e5b06eb32180a4b8ead2ce4f45db/raw/07772083fa1c2e567ed422eb401ac7783487d1c7/ebsnvme-id"
```
proteusvacuum: Would it be prudent to copy this somewhere we have control? Like into this repo? Linking to a gist seems a bit scary to me.

dannyroberts:
My bar is that I'd find simply referencing the gist not acceptable, but in
combination with the sha256 checksum, it's reasonably tamper proof / a
known quantity.

That's assuming security is the issue; was there another problem you saw
with it?

proteusvacuum:
If this person deleted the gist, would it break our script? I mostly just
don't want us to be in a situation where we are scrambling to set up a new
machine and this thing our scripts rely on has gone.
Similarly, if the gist has been changed and the sha256 is now different
(stopping the script), what are we supposed to do?

dannyroberts:
@proteusvacuum I think a change to the gist isn't a concern because the URL contains the specific version/commit hash, but I hadn't thought of what happens if they delete the gist. That makes sense as a reason to vendorize it, so I'll go with that approach. Thanks!

PR #2993, opened 2019-07-03
add CONN_MAX_AGE db setting

snopoke:
Though I don't expect this to have an impact on the current ICDS issues I think it's a good thing to implement anyway and is recommended by a number of PAAS providers like Heroku.

> Persistent connections avoid the overhead of re-establishing a connection to the database in each request. They’re controlled by the CONN_MAX_AGE parameter which defines the maximum lifetime of a connection. It can be set independently for each database.
> The default value is 0, preserving the historical behavior of closing the database connection at the end of each request. To enable persistent connections, set CONN_MAX_AGE to a positive number of seconds. For unlimited persistent connections, set it to None.
https://docs.djangoproject.com/en/2.2/ref/databases/#persistent-connections


snopoke:
rolled this out - it will take effect after deploy

PR #2995, opened 2019-07-03
Limit pgbouncer pool to postgres to 400 for shard databases

emord:
##### SUMMARY
Limit the connections from pgbouncer to postgres to 400

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
pgbouncer/postgres

##### ADDITIONAL INFORMATION
Discussed this with @dannyroberts today after we looked at metrics. On servers that experience the "noon issue" these regularly max out. On servers that do not, they never go above 300. The reasoning here is that the fewer concurrent queries will use less memory and disk i/o at one time and will churn through those queries while not destroying the OS page caching.


snopoke:
rolled out on ICDS

PR #2996, opened 2019-07-03
increase pgbouncer max connections on pgmain

calellowitz:
##### SUMMARY

During noon outages we have been saturating pgbouncer connections, during which db connections remain low. Increasing so we can better absorb the higher usage
@emord @sravfeyn @snopoke 

This continues a process begun last week (though apparently not in this repo). It is currently at 6k.


snopoke:
Looks like this already got rolled out

PR #2997, opened 2019-07-04
add rkumar key

mkangia:
Add Rohit Kumar ssh key and user

already ran `update-users` from this branch to set things up


mkangia:
Thanks @dannyroberts @sravfeyn

sravfeyn:
Just to make sure, did you run this only on staging? Because I had to do it on all other envs for him.

mkangia:
@sravfeyn thats correct. We did it only for onboarding session.
Sorry about that. I should have run it on all envs later. thanks for covering up here.

sravfeyn:
Np! Wanted to confirm if there was a bug or something, thanks.

PR #2998, opened 2019-07-05
assign more webworkers to mobile

snopoke:
##### SUMMARY
Convert half of the `hq_webworkers` to `mobile_webworkers`

##### ENVIRONMENTS AFFECTED
`icds`

##### ADDITIONAL INFORMATION
Based on metrics from the last day the `hq_webworkers` usage stays below 10% while the `mobile_webworkers` is up at 80-90%:

image


snopoke:
rolled this out

emord:
@calellowitz did tell me that he thought 4 was going to be more than sufficient for HQ, so this seems quite reasonable. from those metrics, could be even lower?

snopoke:
Yes, it looks like it could be even lower. If you decide to make further changes please also roll out the datadog config changes to webworkers to update the group tagging

PR #2999, opened 2019-07-05
Add datadog docs and sample dashboard JSON files

sravfeyn:
@calellowitz @mkangia 

eCHIS wants to setup datadog dashboards, so these docs should help them.


sravfeyn:
The JSON files are to be exported from account, and only we have access to that. The import is done by external CSPs who would not have our API keys, so they can only import. These files help them import without having to talk to us.

PR #3006, opened 2019-07-08
upgrade python packages before attempting to build the virtualenv

nickpell:
This is necessary for Python 3 in older Ubuntu versions.

@dimagi/py3


dannyroberts:
Actually the build failure is real—does `virtualenv_home` and/or `py3_virtualenv_home` contain the path to `current` or to a particular release in `releases`? If `current` then it may be creating a directory at `current`, when a later task is supposed to create a symlink to `current`

PR #3013, opened 2019-07-09
only run one instance of pillow retry queue

snopoke:
I noticed that this is running on every pillow machine.


snopoke:
rolled this out to ICDS

PR #3037, opened 2019-07-16
Limit based on mobile header for synclogs

emord:
##### SUMMARY
Allow first time syncs to pass through without rate limits

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
nginx

##### ADDITIONAL INFORMATION
based off my previous work here: https://github.com/dimagi/commcare-cloud/pull/953

As currently implemented it would not rate limit when `X-CommCareHQ-LastSyncToken` is not passed. If the header exists, it falls back to rate limiting based on `$server_name`

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/templates/site.j2
+++ b/src/commcare_cloud/ansible/roles/nginx/templates/site.j2
@@ -20,8 +20,10 @@ upstream {{ balancer.name }} {
 
 {% for var_map in item.server.get('mappings', []) %}
 map {{ var_map.variable }} {{ var_map.name }} {
-  ""     {{ var_map.value }};
-  default     "";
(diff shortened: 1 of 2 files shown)


emord:
This has been deployed to ICDS. The new behavior is:

If header `X-CommCareHQ-LastSyncToken` is provided, the requests are rate limited to 1 r/s with a burst of 10 requests.

If it is not provided, it is not explicitly rate limited.

Globally all restores are rate limited to 15 r/s with a burst of 50 requests.

In practice this means that new restores without a synclog could be rate limited to 14 r/s with a burst of 40 requests, if the rate limiting where existing syncs is saturated.

FYI @ctsims @snopoke @Rohit25negi 

buddy @kaapstorm

PR #3038, opened 2019-07-16
Better prioritization of non-incremental syncs

emord:
##### SUMMARY
Following up on https://github.com/dimagi/commcare-cloud/pull/3037. Spoke with @ctsims about this method

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
nginx

##### ADDITIONAL INFORMATION
With this PR the behavior will be:

Incremental syncs will be limited to 15 r/s with burst of 10 requests

Non-incremental syncs will be subject to 15 r/s, but have a burst of 100 requests.

To receive a 412 (if we lost the user's synclog) they must be able to get through the first rate limiting. If they get into the 412 they should be able to get into the burst of 90 requests in the non-incremental sync queue on a retry.

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
@@ -99,7 +99,7 @@ nginx_sites:
        proxy_read_timeout: 900s
        limit_reqs:
          - zone_name: "incremental_restore"
-           burst: 0
+           burst: 10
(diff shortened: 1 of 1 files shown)


emord:
Meant to follow up that this has been deployed.

@snopoke @Rohit25negi fyi for tomorrow.I think that you will want to change both restore and overall restore zones if you need to change this tomorrow/today

PR #3040, opened 2019-07-17
Separate management commands from celery processes

emord:
##### SUMMARY
This specifies management commands separately from the celery processes

##### ENVIRONMENTS AFFECTED
all
##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
celery

##### ADDITIONAL INFORMATION
Verified that this only produces the desired changes for ICDS (one instance of each mgmt command). Remaining tasks are:

- [x] Change other envs so that they are no-ops
- [x] Write change log so that non-Dimagi managed environments can be changed


emord:
Added the changelog for other users. @millerdev assuming you're the go to for reviewing this sort of change since you're on plat reliability and danny's on PTO? I did check prod/staging to make sure that no actual changes were made after this change

@snopoke I went ahead and deployed this to ICDS, so there should only be one instance of the mgmt commands for tomorrow

emord:
> Do you think it's worth standardising these so we don't need a separate ansible task for each?

I think that would be a reasonable change. I originally wrote it like this because they were already separate tasks.

PR #3046, opened 2019-07-19
Add python3 env for django-management commands

proteusvacuum:
##### SUMMARY

Add's an option to `django-manage` commands to run these in the python 3 environment.

### DIFF
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -192,6 +192,7 @@ class DjangoManage(CommandBase):
             E.g. '2018-04-13_18.16'.
             If none is specified, the `current` release will be used.
         """),
+        Argument('--py3', action='store_true', default=False, help="Use the python3 environment"),
     )
 
     def run(self, args, manage_args):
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -192,6 +192,7 @@ class DjangoManage(CommandBase):
             E.g. '2018-04-13_18.16'.
             If none is specified, the `current` release will be used.
         """),
+        Argument('--py3', action='store_true', default=False, help="Use the python3 environment"),
```
proteusvacuum: I wasn't sure if I should flip this condition so that it makes python 3 the default?
Or maybe we do that after the 30 day window?

orangejenny: What happens if someone does the wrong thing? Will it fail immediately/obviously?

An alternative could be adding a required parameter like `--python` that expects to be passed either 2 or 3. What do you think? It'd be a little annoying for the 30 days, but it would make the upgrade more obvious.

nickpell: @proteusvacuum there is already a setting in fab-settings.yml ```py3_run_deploy```

Should be able to do ```cchq <env> fab deploy --set py3_run_deploy=True``` or something like that.

millerdev: Is there a way to determine the default Python version for the env? Maybe look at what version of Python is being used for webworkers or celery? Seems to me this should use the default if no option is specified. The default should be able to be overridden with something like @orangejenny's suggestion of `--python=2` or `--python=3`.

PR #3049, opened 2019-07-22
reorganize playbooks

snopoke:
##### SUMMARY
Put primary playbooks in a sub-directory `playbooks` and use nested directories for additional playbooks.

* **ansible**
  * **playbooks**: main playbooks folder (deploy_stack and related)
    * **cleanup**: folder for playbooks used to do cleanup tasks (often once off)
    * **util**: folder for utility playbooks

The impetus for this is to cleanup the root ansible folder and also to have a place to put `cleanup` playbooks that doesn't add clutter.

@dimagi/devops FYI


dannyroberts:
Nice. Could we update .bash_completion to look in the new directories as well? https://github.com/dimagi/commcare-cloud/blob/a5d8c40adb36be63b68bdbcbb7e2f20aef7764ba/src/commcare_cloud/.bash_completion#L24

To do that in a DRY way, it might be simplest to add a `PLAYBOOK_PATHS` (or `playbooks`) property option to `manage-commcare-cloud get` and then use that in (or instead of) the `files=...` calculation linked above.

dannyroberts:
The build error
```
ERROR! the role 'ebsnvme' was not found in /home/travis/build/dimagi/commcare-cloud/src/commcare_cloud/ansible/playbooks/roles:/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/.ansible/roles:/home/travis/build/dimagi/commcare-cloud/src/commcare_cloud/ansible/playbooks
```

seems to be saying it's looking for roles defined relative to the playbook location, and thus not looking for them in the director where they are. Looks like previously we were getting this on the ANSIBLE_ROLES_PATH for free (presumably because that's where ansible looks by default, relative to the playbook), but now we're going to have to add it there explicitly, probably here: https://github.com/dimagi/commcare-cloud/blob/580090302796486efe5707a608ebd4e194b5b77c/src/commcare_cloud/environment/paths.py#L24. If the depth is consistent, we might be able to get away with just adding `:../../roles` (or something like that) to the current value.

PR #3051, opened 2019-07-22
add CASE_ES_DROP_FORM_FIELDS to settings

snopoke:
Ref https://github.com/dimagi/commcare-hq/pull/24914


snopoke:
rolled out to ICDS

sravfeyn:
Hey @snopoke Have we actively cleaned up these two fields for old docs on ES on ICDS? I was going through the documents and found most of them with non-empty values for these two fields.

snopoke:
no we haven't

sravfeyn:
Cleaning those fields could reduce the index size, right? Is it not worth?

sravfeyn:
I have tested on India and ICDS by estimating size of around 10k cases with and without those properties, and there is around 12-15% of size reduction with this.

PR #3054, opened 2019-07-23
Replacing celery0 with new celery server which has bionic OS

sanjay2916:
Replacing celery0 with new celery server which has bionic OS


orangejenny:
deployed from this branch successfully. merging.

PR #3058, opened 2019-07-24
improvements to postgres + citus roles / playbooks

snopoke:
* rename some vars for clarity
* improve how we initialise PG data directory
* fix deploy_stack to exclude citus from `deploy_postgres` play
* fix setup_pg_standby play for citus


src/commcare_cloud/ansible/roles/pg_backup/defaults/main.yml
```diff
@@ -1,9 +1,9 @@
 postgresql_version: '9.6'
 postgresql_port: 5432
 postgresql_backup_dir: "{{ encrypted_root }}/backups/postgresql"
-postgresql_dir_original_path: /var/lib/postgresql
+postgresql_home: /var/lib/postgresql
```
emord: I think that the user is actually `postgres` right? What do you think about either `postgres_home` or `postgresql_user_home`? I think I prefer the latter.

snopoke: the user is `postgres` but the home directory is `/var/lib/postgresql`. Var rename is fine with me.

snopoke: 0c28acf75

emord: thanks for that

PR #3060, opened 2019-07-24
don't expire user passwords

snopoke:


src/commcare_cloud/ansible/ansible.cfg
```diff
@@ -11,7 +11,7 @@ fact_caching_timeout = 21600
 timeout = 30
 
 inventory_plugins = ./plugins/inventory
-library = ./library
+library = ./library:~/.ansible/roles/lqueryvg.chage/library
```
snopoke: This is irritating. It works with `~/.ansible/roles` but then we hit this bug: https://github.com/ansible/ansible/issues/40561

snopoke:
> Is there a reason not to also change the system defaults in login.defs so that we don't have to proactively do this each time

No reason. We could do that as well.

snopoke:
Build failing because we have multiple different paths configured for ANSIBLE_ROLES_PATH. Trying to figure out which to standardize on.

https://github.com/dimagi/commcare-cloud/pull/1639/files#r306890668

snopoke:
On second thoughts, I'm not sure it's great to change the setting globally. Doing it per user means we can limit it to user's we create that we know have public key auth and strong passwords. For other admin users like from MSPs etc. I think we should let the normal expiration apply.

snopoke:
Closing in favour of https://github.com/dimagi/commcare-cloud/pull/3064

PR #3061, opened 2019-07-24
don't add standby nodes to master

snopoke:
Exclude hosts in `pg_standby` group when adding workers to the master.


snopoke:
Build failing because we have multiple different paths configured for `ANSIBLE_ROLES_PATH`. Trying to figure out which to standardize on.

snopoke:
@mkangia citus master doesn't have any knowledge about standby nodes so if we add them it will think it needs to distribute load to them as if they were normal worker nodes.

PR #3062, opened 2019-07-24
use 'include_role' to install postgres for citus

snopoke:
Reverts changes in #2933 

First commit is the main change.

Making `postgresql_base` a dependency means that the citus extension is configured
in the `postgresql.conf` file before the extension is installed which prevents PG from starting.

Adding the extension later means that when the role is re-run
the extension is removed from `postgresql.conf` during the `postgresql_base` tasks and then re-added later  by the `citusdb` role with a restart in between which causes errors.

Doing it with `include_role` installs the citus extension and then runs the tasks from the `postgresql_base` role.


sravfeyn:
Much better, thanks!

PR #3070, opened 2019-07-26
Apply dirty background bytes to ICDS shard dbs

emord:
##### SUMMARY
Applies `vm.dirty_background_bytes = 80 MB` to the shard databases so that it persists through reboot

##### ENVIRONMENTS AFFECTED
icds
##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
postgresql, shard_dbs

##### ADDITIONAL INFORMATION
Jesse had already applied this live on the shard dbs. This will write it to a sysctl file so that it persists through reboots.

The default is 10% of the machine's RAM which is very high. Its possible that we could set this on all database machines, but for now I am restricting to ICDS shard databases

@dimagi/scale-team FYI


src/commcare_cloud/ansible/roles/postgresql/tasks/main.yml
```diff
@@ -75,3 +75,12 @@
...
   shell: "{{postgres_install_dir}}/bin/psql -U postgres -p {{postgresql_port}} {{postgresql_dbs.form_processing.proxy.name}} -c \"GRANT USAGE on FOREIGN DATA WRAPPER plproxy to \\\"{{postgresql_dbs.form_processing.proxy.user}}\\\"\""
   when: postgresql_dbs.form_processing and postgresql_dbs.form_processing.proxy.host == inventory_hostname
+
+- name: Set vm.dirty_background_bytes
+  become: yes
+  sysctl:
+    name: vm.dirty_background_bytes
+    value: '{{ dirty_background_bytes }}'
+    state: present
+    sysctl_file: /etc/sysctl.d/90-commcare-cloud.conf
```
emord: We don't really have a custom sysctl_file. I decided to just have this here for now, but it could potentially be a variable later. Not sure if we would want it to be all of commcare-cloud or split by service

PR #3071, opened 2019-07-26
Add support for blocking IPs and ranges in nginx

emord:
##### SUMMARY
Add support for blocking IPs and ranges in nginx
##### ENVIRONMENTS AFFECTED
icds
##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
proxy
##### ADDITIONAL INFORMATION
https://nginx.org/en/docs/http/ngx_http_access_module.html#deny


emord:
@esoergel The way you would deploy this (and only this) is `cchq <env> ap deploy_proxy.yml --tags=nginx-sites` It looks like we don't quite have a place where this would fit at the moment. I'll think about this more

esoergel:
Thanks, that should suffice for documentation in the short-term as far as I'm concerned

PR #3074, opened 2019-07-26
Add couchdb restore script

proteusvacuum:
##### SUMMARY

Adds a restore script for monolith installations of couchdb.

### DIFF
--- a/src/commcare_cloud/ansible/roles/couchdb2/templates/restore_couchdb_backup.sh.j2
+++ b/src/commcare_cloud/ansible/roles/couchdb2/templates/restore_couchdb_backup.sh.j2
@@ -1,0 +1,48 @@
+#!/bin/bash
+BACKUP=$1
+TODAY=$(date +"%Y_%m_%d")
+
+read -p 'Are you sure you want to restore couchdb to a past state [y/N]: ' -r confirm
+if [ $confirm != 'y' ]; then
+    exit 1
+fi
+read -p "This will delete the current db and replace it with the contents of ${BACKUP}. Are you SURE you are sure [y/N]: " -r confirm2
+if [ $confirm2 != 'y' ]; then
+    exit 1
+fi
+
+# remove the trailing paths when extracting the backup file so we get only
+# the files we want
+if tar -Ptf "${BACKUP}" | grep -q "/opt/data/"
+then
+    remove_trailing_components=1
+fi
+
+echo 'extracting backup' &&
+    mkdir -p "{{ couch_backup_dir }}/extracted_backup_${TODAY}" &&
+    tar -xzf "${BACKUP}" -C "{{ couch_backup_dir }}/extracted_backup_${TODAY}" ${remove_trailing_components:+'--strip-components=3'} &&
+
+echo 'stopping couchdb. if this fails the script will exit' &&
+service couchdb stop || exit 1
+
+read -p 'continuing will overwrite the existing couchdb data directory. Are you sure? [y/N]: ' -r confirm3
+if [ $confirm3 != 'y' ]; then
+    exit 1
+fi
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/roles/couchdb2/templates/restore_couchdb_backup.sh.j2
```diff
@@ -0,0 +1,48 @@
...
+    rsync --append-verify -aH "{{ couch_backup_dir }}/extracted_backup_${TODAY}/" "{{ couch_data_dir }}" &&
+
+echo 'setting new ip addresses' &&
+for db_name in $(curl -s -XGET {{ inventory_hostname|ipaddr }}:{{ couchdb_node_port }}/_dbs/_all_docs -u {{ localsettings_private.COUCH_USERNAME }}:{{ localsettings_private.COUCH_PASSWORD }}  | grep -Po '"id":"(\w+\-?\w+)' | cut -c 7-) # the names of all the dbs currently known about by couch
+do
+    # replace all the old ip addresses in the _dbs database with the current host ip address so couch can find the shards
+    echo "setting ip address for: $db_name"
+    curl -s -XGET {{ inventory_hostname|ipaddr }}:{{ couchdb_node_port }}/_dbs/$db_name -u {{ localsettings_private.COUCH_USERNAME }}:{{ localsettings_private.COUCH_PASSWORD }} |
+        sed -r 's/(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}/{{ inventory_hostname|ipaddr }}/g' |
+        curl -s -XPUT {{ inventory_hostname|ipaddr }}:{{ couchdb_node_port }}/_dbs/$db_name -u {{ localsettings_private.COUCH_USERNAME }}:{{ localsettings_private.COUCH_PASSWORD }} -d @-
```
snopoke: This only works for a single node cluster right?

proteusvacuum: Correct. For multi-node clusters and more advanced usecases you can use `couchdb-cluster-admin`. 

For the purposes of "local hosting" a monolith `couchdb-cluster-admin` felt a bit too complicated and confusing to set up (especially when not actually migrating a cluster).

snopoke: In that case I think it's worth making that very clear in the docs as well as in the script itself. The script could even check the cluster membership to ensure that there is only one node.

src/commcare_cloud/ansible/roles/couchdb2/templates/restore_couchdb_backup.sh.j2
```diff
@@ -0,0 +1,52 @@
...
+
+read -p 'Are you sure you want to restore couchdb to a past state [y/N]: ' -r confirm
+if [ $confirm != 'y' ]; then
+    exit 1
+fi
+read -p 'This script will only work on single-node clusters. Damage may occur if this couchdb cluster has more than one node. Are you restoring to a single node? [y/N]: ' -r confirm_nodes
+if [ $confirm_nodes != 'y' ]; then
+    exit 1
+fi
+read -p "This will delete the current db and replace it with the contents of ${BACKUP}. Are you SURE you are sure [y/N]: " -r confirm2
```
emord: Not sure how best to word this, but `current db` is a little confusing since its talking about the entirety of the CouchDB instance and the namespaces in couch are also called `db`. 

I don't see an instance in this script that calls `commcare__users` or any of the dbs `db` though, which is good.

PR #3076, opened 2019-07-29
Replacing trusty celery servers with bionic celery servers on ICDS environment

sanjay2916:
Replacing trusty celery servers with bionic celery servers on ICDS environment


snopoke:
One issue with re-using the same hostnames is that DD uses hostname to group metrics. What is the rollout plan for these.

PR #3082, opened 2019-07-29
Change all default emails to example emails

emord:
##### SUMMARY
Change all default emails to example emails

##### ENVIRONMENTS AFFECTED
3rd party

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
localsettings


src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -103,25 +103,25 @@ datadog_patch_modules:
...
-internal_subscription_change_email: accounts+subchange+internal@dimagi.com
-billing_email: sales@dimagi.com
-invoicing_contact_email: billing-support@dimagi.com
-growth_email: growth@dimagi.com
-master_list_email: master-list@dimagi.com
-report_builder_add_on_email: sales@dimagi.com
-eula_change_email: eula-notifications@dimagi.com
-contact_email: info@dimagi.com
-soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
+root_email: commcarehq-ops+root@example.com
```
proteusvacuum: is there a way we can set this as an "email domain" variable, so people only have to change one thing when setting up a local instance?

emord: That may be possible, but I don't think that's super easy at the moment.

from https://docs.ansible.com/ansible/2.8/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable `group_vars` is one of the first places that variables are loaded and they are overwritten later on. So if we changed one of these to `commcarehq-ops+root@{configured_domain}` then `configured_domain` would be in public.yml (or some other var file) that is not yet loaded. We do some funky thing to generate an ansible compliant yml var file from our several var files, but I don't believe that any of them are loaded before `group_vars`. In the future we could setup `email_domain` in public.yml and generate these emails from that in our generated yaml.

PR #3083, opened 2019-07-30
Downtime commands to include airflow

sravfeyn:
@snopoke


sravfeyn:
Looks like it's alraedy done

PR #3084, opened 2019-07-30
set strict window for couch compaction on ICDS

snopoke:
##### SUMMARY
Prevent CouchDB compaction from continuing beyond allocated time period.

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
CouchDB

##### ADDITIONAL INFORMATION
https://docs.couchdb.org/en/stable/config/compaction.html#compaction-daemon-rules

### DIFF
--- a/environments/icds-cas/public.yml
+++ b/environments/icds-cas/public.yml
@@ -145,6 +145,9 @@ couchdb_cluster_settings:
   w: 2
   n: 2
 
+couchdb_compaction_settings:
+  _default: '[{db_fragmentation, "10%"}, {view_fragmentation, "10%"}, {from, "17:00"}, {to, "22:30"}]'
+
 couchdb2:
   username: "{{ localsettings_private.COUCH_USERNAME }}"
   password: "{{ localsettings_private.COUCH_PASSWORD }}"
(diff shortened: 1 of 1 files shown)


environments/icds-cas/public.yml
```diff
@@ -145,6 +145,9 @@ couchdb_cluster_settings:
   w: 2
   n: 2
 
+couchdb_compaction_settings:
+  _default: '[{db_fragmentation, "10%"}, {view_fragmentation, "10%"}, {from, "17:00"}, {to, "05:00"}, {strict_window, true}]'
```
mkangia: is 17:00 and 5:00 will respect UTC or IST?

snopoke: Machine time, so it should be UTC (22:00 to 10:00 IST).

snopoke:
@emord good question - currently our compaction mostly complete at about 05:30 which is 12.5 hours after starting. Ideally I'd like to have them end a bit earlier so that we're well out of peak but I'm not sure what the consequences of that will be in the long term. I was thinking of deploying this and seeing how it goes.

emord:
That seems fine. Though long term we need to get these on ubuntu 18.04 and couch 2.3.1 to fix the issue where its only compacting the first shard. Was found by @dannyroberts a while ago, ended up saving us a lot of money hosting couch for production.

snopoke:
Rollout ticket: https://dimagi-dev.atlassian.net/browse/IIO-369

PR #3086, opened 2019-07-30
fix check for user processes

snopoke:
This is simpler and also fixes the issue when the username gets
truncated in the output e.g. 'userna+'


rohit-dimagi:
This will fail when user is removed from system but still available in commcare repo.

snopoke:
yes, I realised that last week. Still thinking of a good fix

PR #3090, opened 2019-07-31
set postgresql_max_standby_delay for ICDS

snopoke:
This will allow queries on standby nodes to last up to 3 min before being cancelled.

### DIFF
--- a/environments/icds-cas/postgresql.yml
+++ b/environments/icds-cas/postgresql.yml
@@ -37,6 +37,7 @@ pgbouncer_override:
   pgbouncer_reserve_pool: 5
 
 postgres_override:
+  postgresql_max_standby_delay: 180s
   postgresql_max_stack_depth: 6MB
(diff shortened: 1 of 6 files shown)


environments/icds-cas/postgresql.yml
```diff
@@ -37,6 +37,7 @@ pgbouncer_override:
   pgbouncer_reserve_pool: 5
 
 postgres_override:
+  postgresql_max_standby_delay: 180s
```
rohit-dimagi: This should be int ?

snopoke: updated https://github.com/dimagi/commcare-cloud/pull/3090/commits/3f3989c3679b1bab925a0aabed9ab2873c10f953

snopoke:
Rollout ticket: https://dimagi-dev.atlassian.net/browse/IIO-370

PR #3091, opened 2019-07-31
set blockage_thresholds

snopoke:
Customise celery blockage thresholds for ICDS.

These get used to determine celery uptime. I'd like to configure these appropriately for ICDS so that we can get some useful uptime metrics.

### DIFF
--- a/environments/icds-cas/app-processes.yml
+++ b/environments/icds-cas/app-processes.yml
@@ -27,6 +27,7 @@ celery_processes:
        concurrency: 4
     submission_reprocessing_queue:
        concurrency: 2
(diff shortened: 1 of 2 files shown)


environments/icds-cas/app-processes.yml
```diff
@@ -1,6 +1,12 @@
 formplayer_memory: "8000m"
 gunicorn_workers_static_factor: 4
 
+celery_heartbeat_thresholds:
+  icds_aggregation_queue: 86400
```
calellowitz: This seems quite large. Maybe a couple hours? The agg should never be delayed by a full day.

PR #3092, opened 2019-07-31
Airflow config; remove quotes, escape special chars

sravfeyn:
This should fix airflow's ability to send email (which I tested using airflow's send_email method). It's not clear yet whether airflow is triggering that method at all on failure, which I am investigating.

@calellowitz @snopoke


sravfeyn:
Already deployed

PR #3094, opened 2019-08-01
add script to compact couch views and DBs

snopoke:
Temporary measure to compact couch databases and views until we can upgrade to >= 2.3.1

##### ENVIRONMENTS AFFECTED
ICDS


snopoke:
Rolled this out

PR #3097, opened 2019-08-01
Fewer couch nodes on ICDS

emord:
##### SUMMARY
We're finding that a lot of the errors in the couch logs are from the RPC between couch nodes. Reducing the number of couch nodes (thus having the data more local) should reduce need for RPC calls since all requests should be able to be served by any node.

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
couch


emord:
The data was migrated and this change was rolled out earlier today

snopoke:
This appears to have made things worse so I'm going to roll it back.

snopoke:
superseeded by https://github.com/dimagi/commcare-cloud/pull/3099

PR #3099, opened 2019-08-02
move couch proxy to last node

snopoke:
The proxy is pushing the CPU usage high. Moving it to a node that's not actively in use.


snopoke:
Rolled this out

PR #3101, opened 2019-08-02
Run s3 backup checks daily on swiss

dannyroberts:
##### SUMMARY
This we do daily backups on swiss but if they fail it's pretty silent. This adds a check that looks for the presence of a recent daily and weekly backup (just by matching filename) in s3, on a daily basis, and emails out if there's not.

##### ENVIRONMENTS AFFECTED
swiss

### DIFF
--- a/environments/swiss/public.yml
+++ b/environments/swiss/public.yml
@@ -77,7 +77,7 @@ report_builder_add_on_email: sales@dimagi.com
 eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
-check_s3_backups_email: commcarehq-ops+root@dimagi.com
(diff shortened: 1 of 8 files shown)


environments/swiss/public.yml
```diff
@@ -77,7 +77,7 @@ report_builder_add_on_email: sales@dimagi.com
 eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
-
+check_s3_backups_email: commcarehq-ops+root@dimagi.com,httu.admin@swisstph.ch
```
proteusvacuum: can you add this parameter to the backups documentation?

dannyroberts: Yes this is a great suggestion. Thanks

dannyroberts:
Took @proteusvacuum's suggestions

PR #3106, opened 2019-08-06
BlobDB docs

snopoke:
Add overview docs + docs for FS -> S3 migration


docs/services/blobdb.md
```diff
@@ -0,0 +1,37 @@
+# BlobDB
+{:.no_toc}
+
+* TOC
+{:toc}
+
+## Usage in CommCare
+
+The BlobDB is an internal term used to reference the service and classes responsble for storing binary data in CommCare.
```
emord: `BlobDB` is an internal term, but `blob` is not https://en.wikipedia.org/wiki/Binary_large_object. may be worth mentioning that it is a common acronym instead of something we just invented.

snopoke: :+1: 789412cf9eb48935680b5e366f0df5e71bf3dddb

PR #3111, opened 2019-08-06
a few small changes from things I noticed today

snopoke:


src/commcare_cloud/ansible/roles/thp/tasks/main.yml
```diff
@@ -26,4 +34,9 @@
     line: 'GRUB_CMDLINE_LINUX_DEFAULT="transparent_hugepage=never"'
     backrefs: yes
     state: present
+    register: grup_updated
   when: uses_grub.stat.exists
+
+- name: update grub
+  command: update-grub
```
emord: should we inform TCL of this?

snopoke: I guess we could :speaking_head:

snopoke:
This is not validated by TCL

PR #3113, opened 2019-08-07
increase maxconn for haproxy

snopoke:
I noticed that we're hitting the limit on ICDS so bumping this up.


snopoke:
Already rolled out

emord:
where did you find this error? I recall looking at some hqproxy logs, but may have not been looking at the correct ones. Could you update https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/commands/ansible/service.py#L374-L379 to include something on haproxy?

snopoke:
You can see it in the HAProxy stats either on DD or on `[IP]:1936`

I'll see what I can do about updating the service command

emord:
Thanks. I was mostly wondering because I had noted that we were hitting a max on network in/out on the couch node but didn't connect it to the max connections

PR #3115, opened 2019-08-07
set wal_log_hints = on

snopoke:
This will allow the use of pg_rewind which is useful in certain scenarios such as swapping a standby and master

https://www.postgresql.org/docs/10/app-pgrewind.html


snopoke:
I don't have a rollout plan as yet. There are also some other PG config changes waiting to be rolled out.

PR #3117, opened 2019-08-07
Clear filter for "OpenVPN Access Server 2.6.1"

dannyroberts:
##### SUMMARY
This previous filter recently started coming up empty. It doesn't matter at all for existing environments. For a new one, if I'm wrong and there are other AMIs for this owner, then that'll become apparent pretty quickly. (Also as a practical note, there are no plans to set up a new AWS environment anytime soon.)

##### ENVIRONMENTS AFFECTED
production, india, staging


dannyroberts:
@sanjay2916

PR #3127, opened 2019-08-11
Add three processes to ledger -> ES queue

emord:
##### SUMMARY
Add three processes to ledger -> ES queue

I think that this is all we would need to avoid falling behind again. i currently have 16 deployed to ICDS to catch up

##### ENVIRONMENTS AFFECTED
icds
##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
pillows

### DIFF
--- a/environments/icds-cas/app-processes.yml
+++ b/environments/icds-cas/app-processes.yml
@@ -509,7 +509,7 @@ pillows:
     KafkaDomainPillow:
       num_processes: 1
     LedgerToElasticsearchPillow:
-      num_processes: 1
+      num_processes: 4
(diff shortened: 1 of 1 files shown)


snopoke:
I've been wondering if we need this data in ES at all? I don't think ICDS uses any of the reports where the ES index is used.

emord:
@snopoke I looked around and I think that I could get rid of the whole index in code within a day. It's only used in a few places and seems pretty easy to change `LedgerAccessor` to do what we want with it. I think that hte reports that use this are:

* Stock Status by Product
* Aggregate Inventory
* exports

PR #3132, opened 2019-08-12
Make django-manage and sudoers use py3

dannyroberts:
##### SUMMARY
This changes the last accidental holdouts to use py3:
- django-manage
- the sudoers file (which includes the virtualenv python and pip as a whitelisted commands)

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/templates/sudoers.j2
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/templates/sudoers.j2
@@ -11,7 +11,7 @@ Runas_Alias HIPAA_ACTOR = {{ cchq_user }}
 Cmnd_Alias NGINX = /usr/sbin/nginx
 Cmnd_Alias PACKAGES = /usr/bin/apt-get
 Cmnd_Alias GIT = /usr/bin/git
-Cmnd_Alias PYTHON = /home/{{ cchq_user }}/www/{{ deploy_env }}/python_env/bin/python,/home/{{ cchq_user }}/www/{{ deploy_env }}/python_env/bin/pip
+Cmnd_Alias PYTHON = {{ py3_virtualenv_home }}/bin/python,{{ py3_virtualenv_home }}/bin/pip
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/roles/bootstrap-users/templates/sudoers.j2
```diff
@@ -11,7 +11,7 @@ Runas_Alias HIPAA_ACTOR = {{ cchq_user }}
 Cmnd_Alias NGINX = /usr/sbin/nginx
 Cmnd_Alias PACKAGES = /usr/bin/apt-get
 Cmnd_Alias GIT = /usr/bin/git
-Cmnd_Alias PYTHON = /home/{{ cchq_user }}/www/{{ deploy_env }}/python_env/bin/python,/home/{{ cchq_user }}/www/{{ deploy_env }}/python_env/bin/pip
+Cmnd_Alias PYTHON = {{ py3_virtualenv_home }}/bin/python,{{ py3_virtualenv_home }}/bin/pip
```
millerdev: Does this mean whatever this applied to is only possible with Python 3 now, and not Python 2? I'm not sure what implications that has, but seems like it could be bad for envs that are still on Python 2. Could that be a problem for anyone?

dannyroberts: I considered that and I think it's a good point to raise. I think there are basically two different reasons that made this seem okay to me:

- It follows recent precedent: I looked at other parts of the code that used to split based on the `PY3_RUN_*` variables, and those were similarly reduced to always use `py3_virtualenv_home` in this PR: https://github.com/dimagi/commcare-cloud/pull/2906. (This place never split on those, but maybe should have.)
- This hasn't seemed to cause any problems on any of the envs where we have for the last few weeks had the reverse setup, i.e. deploy etc. uses py3 virtualenv, but this file references the py2 virtualenv. It may just be that this line doesn't actually matter much for anything—I just wanted to be consistent about how we reference the virtualenv

PR #3142, opened 2019-08-15
 get submodule list before entering 'cd' context

snopoke:
I'm not sure why this started failing but I encountered it while setting up new VMs on ICDS. When executed inside the `with cd()` context the returned list is empty (despite actually executing the command to list the submodules).

I did not debug any further.

### DIFF
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -212,13 +212,14 @@ def _get_local_submodule_urls(path):
 
 
 def _get_remote_submodule_urls(path):
+    submodule_list = _get_submodule_list()
     with cd(env.code_current):
         remote_submodule_config = [
             GitConfig(
                 key='submodule.{}.url'.format(submodule),
                 value=sudo("git config submodule.{}.url".format(submodule))
             )
-            for submodule in _get_submodule_list()]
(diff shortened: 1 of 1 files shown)


snopoke:
I don't but I'm running into a few more issues related to permissions:

This check fails unless you pass `use_sudo=True`:
https://github.com/dimagi/commcare-cloud/blob/c4866a76e05bce6468c0c88d7d2f4de0f813d627/src/commcare_cloud/fab/operations/release.py#L278

Running `vitrualenv-close` is failing claiming a missing module `clonevirtualenv`:
https://github.com/dimagi/commcare-cloud/blob/c4866a76e05bce6468c0c88d7d2f4de0f813d627/src/commcare_cloud/fab/operations/release.py#L253

Running `virtualenv-clone` as root works fine so again I think this is a permissions issue. It's installed in the OS python 2.7 environment here:
https://github.com/dimagi/commcare-cloud/blob/24d5d0beaba4ebef8122ea2673e7b0aecb4dca08/src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml#L89

That's as far as I've got. Not sure why these are failing now. I've changed `django_manage` to `pillow21` in my environments on control and using `setup_limited_release` to test since it's way faster and easier to see errors.

PR #3143, opened 2019-08-15
set umask during bootrap and set for current session

snopoke:
##### SUMMARY
Prior to this the umask would only take effect after next login. It was also set after common installs were run making commands like `virtualenv-clone` fail.

### DIFF
--- a/src/commcare_cloud/ansible/roles/bootstrap-machine/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-machine/tasks/main.yml
@@ -4,3 +4,17 @@
       - htop
     state: present
   become: yes
+
+- name: Set proper umask
+  lineinfile:
+    dest: /etc/login.defs
+    state: present
+    regexp: "^UMASK"
+    line: "UMASK\t\t022"
+  tags:
+    - umask
+
+- name: Set proper umask for this session
+  shell: umask 022
+  tags:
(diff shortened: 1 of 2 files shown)


emord:
@snopoke I made a mistake in my pillow PR so I had to deploy two new machines, and this didn't actually work. The umask still caused the same issue :/

snopoke:
Did some digging, might be because `USERGROUPS_ENAB` is set to  "yes"
https://askubuntu.com/a/1066355

snopoke:
I've very hesitant to change this and I think it might be better to work around the more restrictive umask by setting umask specifically where needed e.g. with pip: https://docs.ansible.com/ansible/latest/modules/pip_module.html

emord:
@snopoke as noted in the email I sent, what do you think about restricting other environments umask in that case? I think it would be very beneficial to have the same starting point for such a core configuration

snopoke:
@emord I don't feel I know enough about these configs to know the ramifications of changes, particularly in cases like this where there are multiple settings involved that impact different parts of the system and interact with each other.

I agree that standardized environments is good and I'm ok with 'copying' a working configuration but that can also be an issue if we don't fully understand it and if we don't copy all the relevant settings.

In this specific case I'd like to compare how the other VMs are configured to get a better understanding of what's actually different.

PR #3144, opened 2019-08-15
logging decision

snopoke:
##### SUMMARY
Decision discussion regarding log processing / shipping framework to use.

@dimagi/devops @dimagi/scale-team I'd like to get input on this please

In the process of investigating I created near complete configs for logstash and syslog-ng:

* syslog-ng example config: https://gist.github.com/snopoke/cd69b0cb932649f4724252c6aae3e72f
* logstash example config: https://gist.github.com/snopoke/39ffbca325f0f4c31e5159a37b87531a

### DIFF
--- a/docs/decisions/0001-record-architecture-decisions.md
+++ b/docs/decisions/0001-record-architecture-decisions.md
@@ -1,0 +1,19 @@
+# 1. Record architecture decisions
+
+Date: 2018-07-04
+
+## Status
+
+Accepted
+
+## Context
+
+We need to record the architectural decisions made on this project.
+
+## Decision
+
+We will use Architecture Decision Records, as [described by Michael Nygard](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions).
+
+## Consequences
+
+See Michael Nygard's article, linked above. For a lightweight ADR toolset, see Nat Pryce's [adr-tools](https://github.com/npryce/adr-tools).
--- a/docs/decisions/0002-use-syslog-ng.md
+++ b/docs/decisions/0002-use-syslog-ng.md
@@ -27,8 +27,8 @@ executed by the agent on the logs. Version 6 of the agent does not support this 
 although v5 of the agent is still supported it requires Python 2.7 which has reached end of life and so
 will be deprecated soon.
 
-In order to facilitate the upgarde of the agent it is necessary to move the log processing
-functionality to a different system. The system that get's chosen should have the following characteristics:
+In order to facilitate the upgrade of the agent it is necessary to move the log processing
+functionality to a different system. The system that gets chosen should have the following characteristics:
 
 * low CPU and memory consumption
 * able to process the necessary volume of logs
--- a/docs/decisions/0002-use-syslog-ng.md
+++ b/docs/decisions/0002-use-syslog-ng.md
@@ -1,0 +1,101 @@
+# 2. Use syslog-ng as the logging framework of choice
+
+Date: 2019-08-15
+
+## Status
+
+Proposed
+
+## Context
+
+One component of monitoring CommCare is log collection and analysis. At the time of writing
+there is no centralized log collection or storage system in place. Individual process logs are
+stored on the machines where the process is running.
+
+There are a few logs that are converted into DataDog metrics and sent to DataDog:
+
+* nginx timing / request logs
+* nginx error logs
+* couchdb timing
+
+For these logs we do not send the raw log but rather convert the log lines into specific metrics
+to track request times, errors and apdex.
+
+This log processing is currently handled by the DataDog v5 in the form of
+[Python functions](https://github.com/dimagi/datadog-parsers) which are
+executed by the agent on the logs. Version 6 of the agent does not support this functionality and
+although v5 of the agent is still supported it requires Python 2.7 which has reached end of life and so
+will be deprecated soon.
+
+In order to facilitate the upgarde of the agent it is necessary to move the log processing
+functionality to a different system. The system that get's chosen should have the following characteristics:
+
+* low CPU and memory consumption
+* able to process the necessary volume of logs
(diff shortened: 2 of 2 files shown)


docs/decisions/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
+# 1. Record architecture decisions
```
emord: Should/could we move https://github.com/dimagi/commcare-cloud/tree/master/decisions in here or remove it so there aren't two places for decisions?

docs/decisions/0002-use-syslog-ng.md
```diff
@@ -0,0 +1,101 @@
...
+For these logs we do not send the raw log but rather convert the log lines into specific metrics
+to track request times, errors and apdex.
+
+This log processing is currently handled by the DataDog v5 in the form of
+[Python functions](https://github.com/dimagi/datadog-parsers) which are
+executed by the agent on the logs. Version 6 of the agent does not support this functionality and
+although v5 of the agent is still supported it requires Python 2.7 which has reached end of life and so
+will be deprecated soon.
+
+In order to facilitate the upgarde of the agent it is necessary to move the log processing
```
emord:
```suggestion
In order to facilitate the upgrade of the agent it is necessary to move the log processing
```

docs/decisions/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
...
+
+Accepted
+
+## Context
+
+We need to record the architectural decisions made on this project.
+
+## Decision
+
+We will use Architecture Decision Records, as [described by Michael Nygard](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions).
```
sravfeyn: Great idea, long due! Thanks for this.

sravfeyn:
Just to understand the process, this is a record of the decision made after an architecture discussion (via calls or a review doc), right? So any comments about the architecture itself are too late to be made here? It might be worth linking to architecture review doc where available to this decision record.

For this particular decision, is there a link to the architecture review doc?

docs/decisions/0002-use-syslog-ng.md
```diff
@@ -0,0 +1,101 @@
...
+functionality to a different system. The system that get's chosen should have the following characteristics:
```
czue:
```suggestion
functionality to a different system. The system that gets chosen should have the following characteristics:
```

docs/decisions/0001-record-architecture-decisions.md
snopoke: ha, I guess I missed that one

snopoke:
@sravfeyn there has been no discussion so far - this doc is to kick off that discussion - if you think a Gdoc is a better tool for such discussion I can move it there

snopoke:
@emord 
> What is the limitation of rsyslog and syslog-ng working together?

Great question - I do not believe there is a limitation except that the default syslog-ng install on ubuntu will disable rsyslog and replace it's functionality (for the normal system logging). That's not to say that we can't re-enable rsyslog and only use syslog-ng for our custom log shipping. We'd just need to remove the default syslog-ng config so that we don't get double logging for the system logs. I'll do some investigation to see how that would work since it does seem nicer not to mess with the default system logging.

> Does this mean that we will be replacing our logging handlers that write to /home/cchq/www/<env>/log with syslog handlers?

No, there is no need to change the way we are doing any of our current logging (unless we want to). This will only be used to replace the current log processing we do for couchdb and nginx.

millerdev:
>> What is the limitation of rsyslog and syslog-ng working together?
>
> Great question - I do not believe there is a limitation except that the default syslog-ng install on ubuntu will disable rsyslog and replace it's functionality (for the normal system logging). That's not to say that we can't re-enable rsyslog and only use syslog-ng for our custom log shipping.

Mostly speculating here, but I think this may be a non-standard setup, and possibly harder to support than allowing the default things to happen when syslog-ng is installed. My guess is it's not uncommon to use syslog-ng on Ubuntu machines, and there should be plenty of help online about how to address the various issues we may run into.

emord:
I didn't mean that we should keep rsyslog and not syslog-ng. If its well supported it seems fine to replace it with syslog-ng.

millerdev:
Update: I checked with my brother who is a sysadmin and has quite a bit of experience with both syslog-ng and rsyslog (they're primarily using rsyslog now). He had a few recommendations, some contrary to what I said before:

- there are ways to install syslog-ng without allowing it to become the default syslog utility... not sure if that’s possible on Ubuntu though. If it is, I would highly recommend installing separately.
- do not use the “system rsyslog” service to do custom syslog duties. Instead, run a separate rsyslog (or syslog-ng) service as a non-privileged user to handle custom logs. This takes some tweaking to allow it on privileged ports, but well worth the effort.
- I’m so familiar with rsyslog now that custom behavior is not an issue for me, although configuration docs for rsyslog do suck a lot. I recall having documentation struggles in syslog-ng as well though, but I do think rsyslog is worse, which is kinda shameful since it’s kinda become the de facto standard on most Linux distros.
- If syslog_service-to-syslog_service is your communication architecture, I would recommend rsyslog and using RELP as the message protocol.

snopoke:
@millerdev thanks for the insight from your brother. I'll do some testing with installing syslog without disabling rsyslog. From what I've read this can only be done by installing from source.

snopoke:
Quick update here. There is no way to install logstash from a package into Ubuntu without uninstalling rsyslog. I've tried installing from source but it's been a real trial. Not sure whether to try push forward on that or look at other options:
* logstash with [ruby filter](https://www.elastic.co/guide/en/logstash/current/plugins-filters-ruby.html)
* fluentd with [custom ruby plugin](https://docs.fluentd.org/plugin-development)

PR #3150, opened 2019-08-16
Revert "Revert "Add more case messaging pillows""

snopoke:
This reverts commit 79e9a535d8bd181bbab795fa37ab5941b99c0f59.

Reverted this to try and get through the pillow error queue faster but it turns out the slow part is getting the errors into the queue - not the processing side.


emord:
You mentioned that this probably isn't necessary. Do you still want to merge this?

snopoke:
On average we're getting about 450 case modifications per second and the processing time per change in 0.09s so we need at 40 processes to keep up. I'll adjust this to give us 60 processes.

PR #3151, opened 2019-08-16
register no_tags in pgbouncer role

snopoke:
##### SUMMARY
This was missing since pgbouncer was split out from the postgres role

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
pgbouncer

##### ADDITIONAL INFORMATION
Without this pgbouncer get's restarted on every run of the role.

### DIFF
--- a/src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
@@ -40,7 +40,7 @@
   tags:
     - pgbouncer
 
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
```diff
@@ -40,6 +40,11 @@
   tags:
     - pgbouncer
 
+# Allows to excecute task only when a tag is specified: https://serverfault.com/a/748864
```
emord:
```suggestion
# Allows to execute task only when a tag is specified: https://serverfault.com/a/748864
```

PR #3153, opened 2019-08-16
add umask for pip installs

snopoke:
##### SUMMARY
New Ubuntu VMs on Vmware have umask set to 0002 which makes pip installs unusable by non-privileged users. Setting the umask during install should fix that.

##### ISSUE TYPE
- Change Pull Request

### DIFF
--- a/src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
@@ -84,6 +84,7 @@
       - httplib2
       - pexpect
     executable: pip2
+    umask: 0022
   become: yes
 
 - name: Install virtualenv-clone
@@ -92,3 +93,4 @@
     name: virtualenv-clone
     version: 0.5.1
     executable: pip2
+    umask: 0022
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
```diff
@@ -92,3 +93,4 @@
     name: virtualenv-clone
     version: 0.5.1
     executable: pip2
```
emord: @dannyroberts not sure if you're tracking our `pip2` instances of where we will need to change to upgrade to python3

src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
```diff
@@ -84,6 +84,7 @@
       - httplib2
       - pexpect
     executable: pip2
+    umask: 0022
```
emord: I think that these two installs can be combined by adding:

`virtualenv-clone==0.5.1`

From the docs it seems like that should work (though there are no examples of mixing having versions and not having versions).

Alternatively, it'd probably be good to have versions for these libraries so we can ensure consistent dependencies

src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
dannyroberts: Oh interesting. These are basically the system dependencies we use rather than the application dependencies, and I see three places where we do that. Seems like they're fairly decoupled both from each other and from the commcare python 3 upgrade, but I agree we should try and remove the python 2 dependency here as well. Making myself a ticket for that: https://dimagi-dev.atlassian.net/browse/PLAT-200

PR #3155, opened 2019-08-16
Add -e 'CCHQ_IS_FRESH_INSTALL=1' to first time setup

dannyroberts:
##### SUMMARY
In response to https://forum.dimagi.com/t/python-2-deprecation-notice-30-days-to-upgrade/6699/9 I noticed our new_environment docs don't tell you to run it with `-e 'CCHQ_IS_FRESH_INSTALL=1'` on the first time setup, but I think that's just a mistake. A couple lines below it is included in the retry command.

##### ENVIRONMENTS AFFECTED
new setups

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/docs/setup/new_environment.md
+++ b/docs/setup/new_environment.md
@@ -193,7 +193,7 @@ You will be prompted for the ansible vault password that you entered in [Step 2]
 In this step, you will be prompted for an SSH password. This is the root user's password. After this step, the root user will not be able to log in via SSH.
 
 ``` bash
-$ commcare-cloud monolith deploy-stack --first-time
+$ commcare-cloud monolith deploy-stack --first-time -e 'CCHQ_IS_FRESH_INSTALL=1'
 ```
 
 ```


dannyroberts:
`--first-time` means the first time _for that machine_ (that's how you add a machine to an existing cluster as well) so you don't always want to use `CCHQ_IS_FRESH_INSTALL` when you use `--first-time`. That variable only affects very few things, mostly just causes it to skip certain data migrations. It's a very narrow set of things in which running it on `--first-install` would actually make things go wrong, but it's like totally conceivable that it'd eventually happen, and it also is just sort of semantically wrong.

emord:
that makes sense. i was over here running bootstrap-users and deploy-stack manually. I should probably re-read the those docs at some point

dannyroberts:
Yeah I'm fairly proud about `--first-time` as an alias for that :) Has probably saved me days'-worth of hours over the last 10 months.

PR #3156, opened 2019-08-16
Remove old india couch nodes

dannyroberts:
##### SUMMARY

They've already been stopped for months now, so we weren't incurring
charges or anything, but this is good cleanup we forgot to do.

##### ENVIRONMENTS AFFECTED
india (but doesn't really affect anything)


dannyroberts:
@sanjay2916 by the way in the future, you should feel free to make a PR to remove machines like this once a migration has gone well

sanjay2916:
> @sanjay2916 by the way in the future, you should feel free to make a PR to remove machines like this once a migration has gone well

sure Danny

PR #3162, opened 2019-08-19
Redis cluster migration docs

sravfeyn:
@snopoke

### DIFF
--- a/docs/howto/redis_cluster.md
+++ b/docs/howto/redis_cluster.md
@@ -33,9 +33,68 @@ Redis cluster can be expanded by adding more master nodes and resharding existin
   ```
 - Run below command  to add the node to the existing cluster. Run it on a redis host that is already part of the cluster 
    ```
-	/usr/local/src/redis-4.0.8/src/redis-trib.rb add-node <new_redis_host:redis_port> <existing_redis_host_ip>:<redis_port>
+  /usr/local/src/redis-4.0.8/src/redis-trib.rb add-node <new_redis_host:redis_port> <existing_redis_host_ip>:<redis_port>
   ```
 - Run below command in a tmux to reshard the data which should take few minutes to complete.
 ```
 ./redis-trib.rb rebalance --use-empty-masters <existing_redis_host_ip>:<redis_port>
 ```
+
+## How to migrate Redis cluster
+
+Below process can be used to migrate redis cluster from one set of nodes to another set of nodes if they are all on the same VLAN.
+
+The migration strategy involves adding new nodes to the existing cluster, moving redis slots (aka  shard data) from old to new nodes and then removing the old nodes from the cluster. All steps except removing the main redis node that's the interface for the application (`localsettings.REDIS_HOST`) can be executed without downtime.
+
+Please read the whole section thouroughly before doing the migration.
+
+
+- Setup the new redis nodes. This step can be done at any time before migration.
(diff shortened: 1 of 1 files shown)


docs/howto/redis_cluster.md
```diff
@@ -33,9 +33,68 @@ Redis cluster can be expanded by adding more master nodes and resharding existin
...
+    ```
+    Redis node ID for a host can be found via `redis-cli CLUSTER NODES`
+
+  - Delete the old node from cluster
+    ```
+    ./redis-trib.rb del-node <any_new_redis_host:redis_port> <old_redis_node_ID>
+     ```
+  - Make sure slots are redistributed as expected
+    ```
+    ./redis-trib.rb check <any_new_redis_host:redis_port>:6379
```
snopoke: should this be done before removing the old node?

docs/howto/redis_cluster.md
```diff
@@ -33,9 +33,68 @@ Redis cluster can be expanded by adding more master nodes and resharding existin
...
+    ```
+  Above three steps can be done without application downtime for all old redis nodes except the one corresponding to `localsettings.REDIS_HOST` as application processes all point to this node.
+
+- Before executing above steps on the final `localsettings.REDIS_HOST` node, stop all application services first.
```
snopoke: for this part is it not possible to do this:
1. Rebalance to remove all shards from old node
2. Update config to point HQ processes to new node
3. Restart all HQ processes
4. Remove old node

That would require much less downtime.

snopoke: If that is possible then it might make sense to change the overall process to:
1. Rebalance all nodes to move shards to new nodes
2. Update HQ config and restart processes
3. Remove all old nodes

sravfeyn: If I remember correctly, this was the first thing I tested. This is not possible because the shard rebalance command didn't work when I passed more than one node with zero weight or seemed to work and then there were errors where cluster went into a bad state. 

Also, the shard rebalance for the last node (which is the only one that requires downtime) doesn't take more than 10 minutes, so it's not a lot of downtime.

snopoke: You could still do each node individually and wait for it to complete, what I meant was doing the shard rebalance of the `REDIS_HOST` node while the system is still live so that all that's required is a restart of the HQ processes (which can be done without any downtime).

It should also be possible to change the `REDIS_HOST` before rebalancing the last node right?

sravfeyn: > It should also be possible to change the REDIS_HOST before rebalancing the last node right?

Yeah, that's true! Before removing the last one, we can just update the localsettings.REDIS_HOST to one of the new nodes that's already added. So, no downtime.

sravfeyn: I will update the docs, likewise.

sravfeyn: > There's no reason that this shouldn't work, but I haven't tested that (that Django talking to an empty redis node that's part of the cluster).

Tested locally, works well.

sravfeyn:
Thanks for the crucial input Simon. Updated the docs now.

docs/howto/redis_cluster.md
sravfeyn: It's mainly a debug command, so can be run anytime prior too.

PR #3170, opened 2019-08-20
optimize service operations

snopoke:
##### SUMMARY
Optimise the 'supervisor' based service commands to run more commands in parallel.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Service commands for:
* pillowtop
* celery

##### ADDITIONAL INFORMATION
When starting / stopping / restarting celery and pillowtop processes a separate ansible command is run for each host (usually) in serial. This can be a very slow process.

Since we generally have celery and pillow processes isolated to dedicated VMs it should be possible to operate on them all at once rather than per host.

This change should make running operations like this perform just a single ansible play:
```
cchq <env> service pillowtop restart
cchq <env> service celery stop
``` 

Note: this builds on https://github.com/dimagi/commcare-cloud/pull/3169

### DIFF
--- a/src/commcare_cloud/commands/ansible/helpers.py
+++ b/src/commcare_cloud/commands/ansible/helpers.py
@@ -74,20 +74,20 @@ def add_factory_auth_cmd(environment):
 
 
 def get_django_webworker_name(environment):
-    environment_environment = environment.meta_config.deploy_env
-    project = environment.fab_settings_config.project
-    return "{project}-{environment}-django".format(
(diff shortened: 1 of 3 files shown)


snopoke:
> I think we could remove a lot of code around this area if we used supervisor groups. Is there something that our code provides that couldn't be handled more cleanly with that?

:+1: that would be an improvement. I don't think there's anything that would prevent us from making use of it

PR #3174, opened 2019-08-20
Make fab deploy_formplayer the only way to deploy formplayer

dannyroberts:
##### SUMMARY
by removing it from the main deploy

Web Apps and App Preview break for a few minutes every time we deploy, which is daily on prod. We do fairly little active development on formplayer these days, so this will have little practical day-to-day consequences, and I think it's not terrible from a "philosophical" perspective either. So I propose we just don't deploy formplayer unless we explicitly want to, thereby removing 90% of the damage this flaw in formplayer architecture causes.

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -398,8 +398,6 @@ def kill_stale_celery_workers():
 def deploy_formplayer():
     execute(announce_formplayer_deploy_start)
     execute(formplayer.build_formplayer, True)
-    if getattr(env, 'NEEDS_FORMPLAYER_RESTART', False):
-        execute(supervisor.restart_formplayer)
 
 
 @task
@@ -801,8 +799,6 @@ def silent_services_restart(use_current_release=False):
     """
     execute(db.set_in_progress_flag, use_current_release)
     if not env.is_monolith:
-        if getattr(env, 'NEEDS_FORMPLAYER_RESTART', False):
-            execute(supervisor.restart_formplayer)
         execute(supervisor.restart_all_except_webworkers)
     execute(supervisor.restart_webworkers)
 
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -398,6 +398,8 @@ def kill_stale_celery_workers():
 def deploy_formplayer():
     execute(announce_formplayer_deploy_start)
     execute(formplayer.build_formplayer, True)
+    if getattr(env, 'NEEDS_FORMPLAYER_RESTART', False):
+        execute(supervisor.restart_formplayer)
 
 
 @task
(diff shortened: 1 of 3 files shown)


emord:
I don't think that `deploy_formplayer` actually restarts formplayer

https://github.com/dimagi/commcare-cloud/blob/64c76ecb1ad38c34bd570b9b43c59902ca5d2cd2/src/commcare_cloud/fab/fabfile.py#L398-L400

https://github.com/dimagi/commcare-cloud/blob/64c76ecb1ad38c34bd570b9b43c59902ca5d2cd2/src/commcare_cloud/fab/operations/formplayer.py#L56-L62

dannyroberts:
Wild. I'll update it to restart formplayer on deploy_formplayer

src/commcare_cloud/fab/fabfile.py
```diff
@@ -398,6 +398,8 @@ def kill_stale_celery_workers():
 def deploy_formplayer():
     execute(announce_formplayer_deploy_start)
     execute(formplayer.build_formplayer, True)
+    if getattr(env, 'NEEDS_FORMPLAYER_RESTART', False):
```
emord: Any reason to not get rid of this environment variable, and restart formplayer where this is being set? https://github.com/dimagi/commcare-cloud/blob/bde9f4c7b3a1b517fd95f3b012091beb77701941/src/commcare_cloud/fab/operations/formplayer.py#L58-L62

dannyroberts:
Great question, @snopoke. We're currently updating formplayer at a rate that's more "monthly" than it is "daily", largely tied to the commcare mobile release cycle it seems: https://github.com/dimagi/formplayer/pulls?q=is%3Apr+is%3Aclosed. So I think a good goal would be to deploy it monthly unless otherwise needed (like to get some rare priority bug fix).

_How_ we enforce, encourage, or nudge that is a good question. Something I've been thinking about doing is folding the `fab deploy` under a `commcare-cloud <env> deploy` command that also does a couple other things (you could imaging running `update-config` during that, for example), and also abstracts it away from `fab` the tool, letting us rewrite pieces of the deploy without necessarily using `fab`. So possibly the simplest approach here would just be to have an experience that's something like this:

```
$ commcare <env> deploy
--> You are about to deploy commcare
Deploy commcare? [Y/n]

--> We recommend deploying formplayer about once a month as well.
This causes about 1 minute of service interruption to Web Apps and App Preview,
but keeps these services up to date.
(You can run this later with `commcare <env> deploy formplayer`.)
Deploy formplayer? [y/N]
...
```

While it's a little clunky to push the decision back onto the maintainer like this without better information (like only asking if there's even new update to deploy) I think it's a good start that we can improve over time. What do you think?

dannyroberts:
Merging for today's deploy, but I want to make sure to finish the comms discussion

snopoke:
@dannyroberts I think moving to that kind of deploy workflow makes a lot of sense. I also think it shouldn't be too hard to set up a formplayer 'release' system that puts a new version + metadata in S3 which can be queried during deploy to determine if a new release is available.

PR #3176, opened 2019-08-20
New webworkers

emord:
##### SUMMARY
@snopoke I started this and changed the wcdadmin password, but realized i'm not sure how safe it is to change the shared_dir_host

##### ENVIRONMENTS AFFECTED
icds
##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
webworkers


environments/icds-cas/inventory/commcare.csv
```diff
@@ -79,3 +79,21 @@ web20,100.71.188.33,webworkers,mobile_webworkers,,,MUMGCCWCDPRDW020
 web21,100.71.188.68,webworkers,mobile_webworkers,,,MUMGCCWCDPRDW021
 web22,100.71.188.62,webworkers,mobile_webworkers,,,MUMGCCWCDPRDW022
 web23,100.71.188.74,webworkers,mobile_webworkers,,,MUMGCCWCDPRDW023
+web24,100.71.188.203,webworkers,celery,shared_dir_host,hq_webworkers,MUMGCCWCDPRDWV000
```
emord: Kind of annoyed these hostnames won't match with the given ones from TCL, but also think its better to not re-use the same hostname on a new host

snopoke:
re `shared_dir_host` there are basically no files in it so I think we should just deploy the new workers with the new `shared_dir_host` and leave the others as they are

emord:
These machines are deploy-stack'd. I didn't have a chance to verify their configuration and deploy the proxy, so they are not being hit by the proxy yet.

I also kept shared_dir_host on the old server because from #icds-dev it sounded like there were still files on there that GS may need and I did not want to break those workflows

snopoke:
@emord those files are not in active use and will remain there until we decommission web0.

snopoke:
@emord I've re-deployed the shared drive on the new workers so web24 is now `shared_dir_host` (but only for new workers).

I attempted to add them to the proxy but got this error:
`OfflineGenerationError: You have offline compression enabled but key "d6a1ca23f4935a901a243817bd75891c" is missing from offline manifest. You may need to run "python manage.py compress".`

I have not tried to fix it. I also noticed that we'll need to move the celery processes which are running on the old nodes and also that we only received 20 new VMs where we have 24 old ones.

snopoke:
> we only received 20 new VMs where we have 24 old ones

I checked the original manifest and there are 24 VMs listed - any reason that we've only configured 20 of them?

emord:
> any reason that we've only configured 20 of them?

I think this is probably just me screwing up. Sorry about that. I fixed the hostnames of the current servers and deployed the new ones

snopoke:
had the same offline compression issue. found a few problems:
1. `/home/cchq` was owned by root causing a number of deploy steps to fail
  - I haven't tracked this down yet
2. https://github.com/dimagi/commcare-cloud/pull/3190
3. https://github.com/dimagi/commcare-cloud/pull/3191

snopoke:
This is all live now

sravfeyn:
FYI, this caused share-dir issues last two days where folks were not able to download some files.  https://dimagi-dev.atlassian.net/browse/ICDS-824

snopoke:
Weird - those are only temporary files that should get re-generated.

PR #3179, opened 2019-08-20
Increase india celery

dannyroberts:
##### SUMMARY
End result:
- a processor for each queue runs on celery2 and celery3 (for redundancy)
- all auxiliary processes that need to run in just one place run on celery1
- each of the three machines is 1/2 the size of the original celery1 for a total of 50% more resources allocated to celery on india

##### ENVIRONMENTS AFFECTED
india


environments/india/app-processes.yml
```diff
@@ -10,24 +10,25 @@ celery_processes:
   celery1:
     beat: {}
     flower: {}
+  celery2,celery3:
```
mkangia: this is cool, i didn't know we could this. 👍

dannyroberts: Yeah, it's a new thing I added in https://github.com/dimagi/commcare-cloud/pull/3121 to make consistent setup across 2 or more machines really easy to maintain :)

mkangia: definitely easier and less lines :) thank you Danny

PR #3187, opened 2019-08-22
Add `deploy` command

dannyroberts:
##### SUMMARY
Easiest to review one commit at a time.

This makes `commcare-cloud <env> deploy` a top-level command that wraps current fab deploy and pulls the most common args and settings to up to the top level as well (`--resume` instead of `:resume=yes`, `--commcare-branch=foo` instead of `--set commcare_branch=foo`, etc.).

This also causes attempts to call `commcare-cloud <env> fab deploy` directly to print you a deprecation message and exit, to provide a reliable and discoverable onboarding path for the behavior change.

##### ENVIRONMENTS AFFECTED
general

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
deploy

##### ADDITIONAL INFORMATION

I tested it out on today's deploy. The beginning of it looked something like this:
```
$ cchq --control production deploy --branch=dmr/deploy-command
...
commcare_branch not specified, using 'master'. You can override this with '--commcare-branch=<branch>'
commcare-cloud production fab deploy_commcare --set code_branch=master --branch dmr/deploy-command
fab -f /home/droberts/commcare-cloud/src/commcare_cloud/fabfile.py production deploy_commcare --set code_branch=master --disable-known-hosts --system-known-hosts /home/droberts/commcare-cloud/environments/production/known_hosts
/home/droberts/.virtualenvs/ansible/local/lib/python2.7/site-packages/cryptography/hazmat/primitives/constant_time.py:26: CryptographyDeprecationWarning: Support for your Python version is deprecated. The next version of cryptography will remove support. Please upgrade to a 2.7.x release that supports hmac.compare_digest as soon as possible.
  utils.PersistentlyDeprecated2018,
[10.202.11.233] Executing task '_confirm_deploying_same_code'
You can add a config file to automate this step:
    $ cp /home/droberts/commcare-cloud/src/commcare_cloud/fab/config.example.py /home/droberts/commcare-cloud/src/commcare_cloud/fab/config.py
Then edit /home/droberts/commcare-cloud/src/commcare_cloud/fab/config.py
Github username (leave blank for no auth): dannyroberts
Github password:
[10.202.11.233] Executing task 'deploy_commcare'
Are you sure you want to preindex and deploy to production? [y/N] y
[10.202.11.233] Executing task '_setup_release'
...
```


src/commcare_cloud/commands/fab.py
```diff
@@ -62,6 +70,27 @@ def run(self, args, unknown_args):
...
+        print(cyan('The `commcare-cloud <env> fab deploy` command has been deprecated.'))
+        print()
+        sleep(3)
+        print(cyan('Instead, please use'))
+        print()
+        print(yellow('  commcare-cloud <env> deploy'))
+        sleep(4)
+        print()
+        print()
+        print(magenta('Thank you for using commcare-cloud.'))
```
snopoke: :heart_decoration:

src/commcare_cloud/cli_utils.py
```diff
@@ -8,12 +8,16 @@
...
-def ask(message, strict=False, quiet=False):
+def ask(message, strict=False, quiet=False, default=False):
     if quiet:
         return True
     yesno = 'YES/NO' if strict else 'y/N'
-    negatives = ('NO', 'N', 'n', 'no', '')
+    negatives = ('NO', 'N', 'n', 'no')
     affirmatives = ('YES',) if strict else ('y', 'Y', 'yes')
+    if default:
+        affirmatives += ('',)
```
millerdev:
```suggestion
        affirmatives += ('',)
        yesno = 'Y/n'
```

src/commcare_cloud/cli_utils.py
```diff
@@ -8,12 +8,16 @@
...
 
-def ask(message, strict=False, quiet=False):
+def ask(message, strict=False, quiet=False, default=False):
     if quiet:
         return True
     yesno = 'YES/NO' if strict else 'y/N'
-    negatives = ('NO', 'N', 'n', 'no', '')
+    negatives = ('NO', 'N', 'n', 'no')
     affirmatives = ('YES',) if strict else ('y', 'Y', 'yes')
+    if default:
```
millerdev: I think this whole `if`/`else` construct should be wrapped with `if not strict:`

src/commcare_cloud/commands/fab.py
```diff
@@ -62,6 +70,27 @@ def run(self, args, unknown_args):
...
+        print(magenta('Things have changed.'))
+        print()
+        sleep(2)
+        print(cyan('The `commcare-cloud <env> fab deploy` command has been deprecated.'))
+        print()
+        sleep(3)
+        print(cyan('Instead, please use'))
+        print()
+        print(yellow('  commcare-cloud <env> deploy'))
+        sleep(4)
```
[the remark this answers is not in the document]
snopoke: :rofl:

src/commcare_cloud/cli_utils.py
dannyroberts: Thanks that's a good point. Once I got into the spirit of this suggestion, I realized the interaction between the strict, quiet, and default args wasn't that well defined, so I gave each of them a job to carry out dutifully instead of each one making assumptions about the other arg (e.g. `quiet` and `strict` seemed to at least partially presuppose default=True).  It's true I think the `default` is unused at this point, but I was about to use it before I decided to stop where I was and PR it. (The use case I have in mind is to prompt on whether to deploy formplayer in the `deploy` command, but to default to "no".) Thanks for the feedback, let me know whether you're happy with my solution/approach.

src/commcare_cloud/commands/fab.py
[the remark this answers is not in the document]
dannyroberts: I'm mixed on this because I see what you're saying. I thought this worked well as a way to grab attention and have people read through the whole brief message the last time we went through one of these transitions (which was when we switched from running `fab <env> deploy` in the commcare-hq directory to using `commcare-cloud <env> deploy`). As not a very good reader myself, I imagined myself likely to absorb the message more quickly in the middle of a busy day if it's displayed to me like this versus just printed out into the screen at once (in which case I might run it again while flipping through tabs before realizing something was out of the ordinary and I should read it). So yeah I was going for engaging and slightly playful, but if instead it comes off as annoying and mildly patronizing or something like that then I'm happy to get rid of the pauses.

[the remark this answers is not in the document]
emord: I think we should be consistent about how we communicate these changes. That could mean a few things (some of these could be done together).

1. Write a changelog entry indicating that it will change soon
2. `fab deploy` tells you to use `deploy` and exits
3. `fab deploy` tells you before the command is run that it will be deprecated soon and you should use `deploy` in the future, but continues to deploy as normal
4. `fab deploy` works as it does now and tells you to run `deploy` in the future once deploy is complete

I think if I were trying to run this as a third party and not keeping up with the github PRs, then I'd expect a changelog entry and it'd be nice to have 2, 3 and/or 4 (probably either 3 with a prompt (`you know this is old right? y/n`) or 4 as the nicest experience). I don't think I'd expect something reminiscent of [sl](https://github.com/mtoyoda/sl)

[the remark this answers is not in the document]
millerdev: My $0.02: I also found this decision a bit strange, if kinda cute. The slow output over 10s would not indicate to me that I should read carefully and then do something different. I rarely look output of deploy commands until after they have completed, and my terminal clearly tells me when the script has exited so it would be easier for me to spot that something went wrong if it printed the error and exited immediately.

src/commcare_cloud/cli_utils.py
dannyroberts: I ended up changing my mind. Here's my reasoning pasted from the last commit ("Undo recent changes to ask")
> I think I was thinking about this the wrong way, and I decided I don't
actually want to touch this. I realized we rely on ask having a
particular meaning: it means "here's something that you should say yes to
if you want to do the normal thing". To present a consistent experience,
if we want the default to be no, we should just ask the question in the
negative: "Do you want to skip formplayer deploy? [Y/n]". This allows
us to always pick a predictable answer when using --quiet, which means "say
yes to everything". In the previous commit where I changed quiet to
always pick the default, that was obviously a problem because our
default is current No everywhere, but we want quiet to pick Yes for you.

dannyroberts: I ended up deciding against all this. I wrote up my rationale in the last commit (I was sloppy with my example and so the logic there doesn't quite make sense, but I think the overall sentiment was right).

dannyroberts:
@dimagi/team-commcare-hq FYI we're now using `commcare <env> deploy` in place of `commcare <env> fab deploy`.

- Changelog: https://dimagi.github.io/commcare-cloud/changelog/0029-add-deploy-command.html
- Docs: https://dimagi.github.io/commcare-cloud/commcare-cloud/commands/#deploy
- CLI help: `cchq production deploy -h`

If you forget and use `cchq <env> fab deploy` it will remind you and point you to these resources as well.

PR #3190, opened 2019-08-23
only install from a single requirements file

snopoke:
##### SUMMARY
Since we use pip-compile to generate the requirements files and that
'prod-requirements' includes 'requirements' there is no need to run pip install with both. It can cause
problems if the same dependency exists in both but with different versions.

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
deploy

### DIFF
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -288,7 +288,6 @@ def update_virtualenv(full_cluster=True):
                      user=env.sudo_user)
                 pip_install(cmd_prefix, timeout=60, quiet=True, proxy=env.http_proxy, requirements=[
                     posixpath.join(requirements, 'prod-requirements.txt'),
-                    posixpath.join(requirements, 'requirements.txt'),
                 ])
 
         _update_virtualenv(


snopoke:
FYI this caused pip install to fail since `requirements.txt` and `prod-requirements.txt` have different versions of `setuptools`

kaapstorm:
:unamused:  That's confusing. I see it's commented out in [the Python 2 requirements.txt](https://github.com/dimagi/commcare-hq/blob/4b247d34bf60578d2a4b687336f1cd9dd4b1085a/requirements/requirements.txt#L169) file. Is that an option?

snopoke:
Looks like that was done manually by @emord https://github.com/dimagi/commcare-hq/pull/25128

emord:
@snopoke @kaapstorm  That was not done manually by me. It was done by the previous version of our  `make requirements` @dannyroberts just changed the script and intentionally did not include python 2 https://github.com/dimagi/commcare-hq/pull/25142

dannyroberts:
👍 I remember postponing this very early on, but I think by now we've ironed things out to make sure prod-requirements.txt is always a superset of requirements.txt (with the same versions, etc) so I agree this seems fine now

PR #3192, opened 2019-08-23
check user exists before killing processes

snopoke:
##### SUMMARY
Check if a user exists before looking for processes to kill

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
users

##### ADDITIONAL INFORMATION
This will prevent errors from `ps` by not looking for processes belonging to users to don't exist

### DIFF
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/remove_users.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/remove_users.yml
@@ -1,7 +1,17 @@
 ---
 
+- name: Find users to remove
+  getent:
+    database: passwd
+    key: "{{ item }}"
+    fail_key: no
+  loop: "{{ dev_users.absent }}"
+
+- set_fact:
+    users_to_remove: "{{ user_check.results|selectattr('msg', 'undefined')|map(attribute='item')|list }}"
+
 - name: Get running processes
-  shell: "ps -o pid= -u \"{{ dev_users.absent|join(',') }}\""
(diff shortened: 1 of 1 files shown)


snopoke:
@kaapstorm the `user_check` var has the structure shown below so the expression iterates over the `results` list and selects elements where the `msg` attribute is undefined (since this attribute only exists if the user not found in the user db). It then extracts the `item` attribute from each element and returns it as a list.

```json
{
    "user_check": {
        "changed": false, 
        "msg": "All items completed", 
        "results": [
            {
                "ansible_facts": {
                    "getent_passwd": {
                        "username": null
                    }
                }, 
                "ansible_loop_var": "item", 
                "changed": false, 
                "failed": false, 
                "invocation": {
                   ...
                }, 
                "item": "username", 
                "msg": "One or more supplied key could not be found in the database."
            }, 
        ]
    }
}
```

kaapstorm:
:+1: Thank you for the explanation.

rohit-dimagi:
`FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'user_check' is undefined\n\nThe error appears to be in '/home/rohit/dimagi.com/commcare-cloud/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/remove_users.yml': line 10, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- set_fact:\n  ^ here\n"}
`

snopoke:
https://github.com/dimagi/commcare-cloud/pull/3197

PR #3204, opened 2019-08-29
Remove hotfix_deploy and force_update_static

emord:
##### SUMMARY
This removes hotfix_deploy and force_update_static.There's been a string of tickets that have caused JS issues recently such as https://dimagi-dev.atlassian.net/browse/ICDS-835.

I'm making this decision because I think its a better world to have one deploy that works than an extra one that breaks in some instances. (and at this point I think hotfix_deploy may only save ~5 minutes at most when there are js changes)

The particular issue may have been introduced by https://github.com/dimagi/commcare-hq/pull/25027 but this has always been a little sketchy.

##### ENVIRONMENTS AFFECTED
all
##### ISSUE TYPE
- Change Pull Request
##### COMPONENT NAME
deploy


orangejenny:
💯

sravfeyn:
hotfix_deploy definitely served a purpose though. Isn't there anything that could be done to fix the broken thing than to completely kill it?

If the issue is that the hotfix-deploy was being used to update static files, we could add some sort of check to make sure only python files are in the hotfix commits?

czue:
It might also be worth noting that yesterday `resource_versions` was incorrect on ICDS despite the last deploy being a full deploy. So this may not have solved the intended issue. (I didn't investigate beyond that)

orangejenny:
@czue If you're not investigating the underlying bug further, please assign the ticket to me.

czue:
I believe the ticket I was working on will be resolved by the compression change, but I just assigned [a new one](https://dimagi-dev.atlassian.net/browse/HI-865) to you with the information that I have.

orangejenny:
@czue Thanks.

emord:
heads up I'm quoting some of this out of order because I think it makes more sense in my explanation. This also got longer than I was expecting, but figured I should explain my full process rather than make decisions without the explanation.

> hotfix_deploy definitely served a purpose

From my understanding the purpose was quickly deploy python only changes by side stepping some processes in our deploy script. This was made a long time ago (furthest I got was 2015 https://github.com/dimagi/commcare-hq-deploy/commit/314e1e8c89da3ea495ba3c23f646046d61f7757b) and its been relatively unchanged, and the normal deploy process has been changed quite a bit to make it more reliable and faster.

>  this may not have solved the intended issue. 

At this time running `force_update_static` will result in some strange JS effects. In the initial case it was thought to only cause the top bar to not work on some pages, but we found a lot more issues when this happened on india. The linked jira ticket and retro should have more info on all the issues were found, but I doubt that it was comprehensive. I think no matter what removing force_Update_static (or fixing all issues we have with static files generation) is the correct way forward.

> It might also be worth noting that yesterday resource_versions was incorrect on ICDS despite the last deploy being a full deploy.(I didn't investigate beyond that)

I agree. I think there are definitely issues with our static management outside the scope of hotfix_deploy + force_update_static. and there is work that Jenny has been doing to figure those out. This also means that static file generation is a moving target and we need to keep two repositories and possibly multiple workflows in sync with (if we went the path of fixing hotfixes/force_update_static).

> Isn't there anything that could be done to fix the broken thing than to completely kill it?

anything can be fixed. I asked myself the following questions (and try to ask myself some similar types of questions for other bugs I find):

1. What does the hotfix workflow provide us? 
2. If hotfix_deploy didn't exist now, would I implement it? 
3. Would I ever recommend a third party to use hotfix_deploy?
4. Is fixing this workflow the best use of anyone's time?

For 1, As stated in my initial PR I think it may only be ~5 minutes per deploy at best which isn't a ton and a large time sink in deploys is that we restart the webworkers 1 by 1. This is especially painful on ICDS where we have > 20 webworkers.

For 2, my answer is no. If I wanted a faster deploy, I would likely spend my time implementing batching for our webworkers, or faster static file generation (likely trying to profile django's standard static file genration vs our custom one) or profiling other parts of the deploy.

For 3, nope. I think we have to consider commcare-cloud a third party tool and anything documented on the github page is fair for them to use. Providing them with a small clean interface that we can guarantee works should be important if we want to support them effectively.

For 4, I think if the answer to 3 was yes then it would be a good use of time to fix this because it would inevitably come up later. Also supporting it now, means that when we have to rewrite all our fab stuff for python 3 (and I say rewrite because they totally rewrote fabric to add py3 support https://www.fabfile.org/upgrading.html#upgrading and its not clear how migratable it is to fabric 2.x vs ansible) then its one more thing for that person to have to deal with.

Anyways, I am not the team and can't make these decisions unilaterally. If there is a strong opinion that we should add hotfix_deploy back to fabric then feel free to add it back. I personally think that its not worth it and we should invest that same time in making one process work.

mkangia:
That's a great PR, the communications on it and the decision making, solid. :)

czue:
Thanks for the very comprehensive note @emord - your conclusions make sense to me.

One question I had (maybe for @sravfeyn) is why `hotfix_deploy` is useful. Like, it might be a totally different thing than speed (e.g. it's safer to run in the middle of the day on CAS). So maybe there is room to rethink/redesign a new `hotfix_deploy` that better serves the needs we have today that could be put on a roadmap if there is a gap in the workflow created by this change.

Also very small point, but there is [a Python3 port of the Python2 Fabric API](https://pypi.org/project/Fabric3/) that is well maintained if we ever needed it.

snopoke:
Adding my quick thoughts. I definitely 'feel' like hotfix deploy is useful but I fully agree with Emord's decision points, in particular I think that the usefulness of hotfix is that it's faster (at least perceived to be) but if that's the goal then it makes more sense to optimize the really slow parts of the deploy rather than just leaving bits out.

I think this warrants a decision doc: https://github.com/dimagi/commcare-cloud/tree/master/decisions

sravfeyn:
Thanks for the writeup, Emord. Makes sense to me, if we can get a faster deploy out, we probably don't need hotfix.

> One question I had (maybe for @sravfeyn) is why hotfix_deploy is useful.

I think it's been an unstated process that during non-standard deploy schedule, if we want to make a quick bugfix, we do an hotfix deploy. This happens every interrupt cycle for me usually for P2/P1s. So, without a hotfix deploy anymore, are we supposed to do a full deploy for situations like this? (I think support might have some say regarding interrupt/p1/p2-fixes as well)

emord:
I believe that the safety in the `hotfix_deploy` process is that no one (maybe?) ever used a hotfix_deploy to deploy master, so that those deploys only contained a well understood subset of changes. Using `hotfix_deploy` meant you were deploying a code branch that was based off the latest deploy with a commit or two on top. This workflow has always been supported by the regular `deploy` script with the same argument to specify the code branch. And its documented in its new form (through commcare-cloud) https://dimagi.github.io/commcare-cloud/commcare-cloud/commands/#deploy

I think this may have been confused by using the generic term "[hotfix](https://en.wikipedia.org/wiki/Hotfix)" when that may or may not be deployed via a `hotfix_deploy`. I've certainly said "hotfix" when doing a normal deploy because it was a very specific change that we needed for a customer issue without using `hotfix_deploy`, so that may have been where confusion originated.  

> So, without a hotfix deploy anymore, are we supposed to do a full deploy for situations like this? (I think support might have some say regarding interrupt/p1/p2-fixes as well)

I think that any deploy during peak load should continue to use the process of creating a branch off of the latest deploy and using commcare-cloud's `--commcare-branch` flag. In other words during the day we should only deploy well understood changes to the current code that fix a very specific problem

dannyroberts:
@czue following up on your very small point
> Also very small point, but there is [a Python3 port of the Python2 Fabric API](https://pypi.org/project/Fabric3/) that is well maintained if we ever needed it.
Thank you I wasn't previously aware of that and that sounds like at the very least a good stop-gap to consider. Added a note on https://dimagi-dev.atlassian.net/browse/PLAT-181

PR #3205, opened 2019-08-29
Combine pillows for ICDS

emord:
##### SUMMARY
To be used in combination with https://github.com/dimagi/commcare-hq/blob/master/corehq/ex-submodules/pillowtop/management/commands/create_checkpoints_for_merged_pillows.py

I have intentionally not included form pillows here, because it would add unknownuserspillow processor which adds some couch/redis hits and I think we may want to remove`FormSubmissionMetadataTrackerPillow` as well if we move the app status over. This does remove ~200 processes (and db connections) as it is now so I think its still worthwhile.

My plan:

1. Deploy on Monday/Tuesday (don't want to change this over weekend)
2. Figure out any corrections on pillow/partition sizing (probably need half of these servers at most)
3. investigate how hard it will be to remove those form pillows
4. If goes well, make change to prod/swiss/pna
5. deprecate old pillows and write changelog

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

### DIFF
--- a/environments/icds-cas/app-processes.yml
+++ b/environments/icds-cas/app-processes.yml
@@ -518,13 +518,10 @@ pillows:
       num_processes: 1
     UserGroupsDbKafkaPillow:
       num_processes: 1
-    UserPillow:
+    user-pillow:
       num_processes: 1
     UpdateUserSyncHistoryPillow:
       num_processes: 16
-    kafka-ucr-main:
-      num_processes: 1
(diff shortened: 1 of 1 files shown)


emord:
for form metadata https://github.com/dimagi/commcare-hq/pull/25267

haven't been able to consider the impact of the unknown user pillow yet

PR #3209, opened 2019-08-31
remove old couch nodes

snopoke:


snopoke:
@rohit-dimagi sorry I missed you PR

PR #3214, opened 2019-09-04
change redis to use LFU

snopoke:
##### SUMMARY
LFU is relatively new eviction model that targets keys
which are used less frequently than others. This may
improve our cache usage. It also gives us access to the
`--hotkeys` command which lists keys that are most frequently
accessed.

https://redis.io/topics/lru-cache#the-new-lfu-mode

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
Redis


sravfeyn:
FWIW, our redis nodes haven't been evicting keys (probably because they have been able to manage memory), so not sure if this would impact anything directly now. But looks like a good change.

PR #3217, opened 2019-09-04
Add proxy1-production

dannyroberts:
##### SUMMARY
This machine has been mid-setup for a few weeks because I wasn't able to finish when I started. I think at this point it's best to add it to terraform and the inventory, but not add it to the proxy group. This will make running terraform for other reasons not ask to delete this machine, etc. until I'm done setting it up.

##### ENVIRONMENTS AFFECTED
production


environments/production/aws-resources.yml
```diff
@@ -7,6 +7,7 @@ couch12-production: 10.202.40.52
 couch13-production: 10.202.40.154
 couch14-production: 10.202.40.216
 couchproxy0-production: 10.202.40.162
+djangomanage0-production: 10.202.10.185
```
emord: Was this intentional to add?

dannyroberts: Good question. Looks like that machine must still be running despite having been replaced and not showing up in our inventory. I've checked previously with people that no one's using it, so I'm going to delete it and remove this line.

PR #3219, opened 2019-09-04
Fix terraform specs for production ES master nodes

dannyroberts:
### SUMMARY
They were previously not matching what was deployed

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -119,15 +119,17 @@ servers:
   - server_name: "esmaster3-production"
     server_instance_type: t3.large
     network_tier: "db-private"
-    az: "a"
+    az: "b"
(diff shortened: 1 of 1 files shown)


dannyroberts:
@sanjay2916 can you review this? I'm a little confused how they might have gotten this way

dannyroberts:
I think I'd decided 2 az's is more than enough redundancy. This really only protects from not being able to elect a master, and even that only partially. If there really is an AZ wide outage, our shards aren't zone aware anyway so it won't protect from that. Basically if we had it to do over, at this point I'd just put them all in the same AZ. I don't think I relayed that to anyone so when we had the chance, it looks like we just stuck with the same layout

PR #3223, opened 2019-09-04
Add proxy1-production to deploy

dannyroberts:
##### SUMMARY
Add proxy1 to the deploy in preparation for the flip. Will promote to the real proxy later by changing the elastic IP configuration in AWS

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
@emord our clients have a history of asking us to list all the IP addresses CommCare (mobile) hits, I believe so they can whitelist them with their ISP to get discounted mobile data for requests to that IP address or something like that. So things are easier on our side and on our clients if we keep the same public IP address unless there's a compelling reason to change it.

Make sense? Is there a downside to that that you can think of?

emord:
Yep. that makes sense. was mostly wondering if there were a technical reason to be doing this

PR #3224, opened 2019-09-05
update icds rate limiting

snopoke:
##### SUMMARY
add rate limit for multimedia downloads. apply rate limit for app downloads

##### ENVIRONMENTS AFFECTED
ICDS
##### ISSUE TYPE
- Change Pull Request
- Bugfix Pull Request

##### COMPONENT NAME
nginx

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
@@ -42,6 +42,10 @@ nginx_sites:
       zone: "$server_name"
       size: "10m"
       rate: "150r/s"
+    - name: "multimedia"
+      zone: "$server_name"
+      size: "10m"
+      rate: "50r/s"
    proxy_cache_path:
     - name: "app_cache"
       path: "{{ www_home }}/app_downloads/cache"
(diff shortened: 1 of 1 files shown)


snopoke:
already rolled out (though will different actual limits)

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -82,6 +90,10 @@ nginx_sites:
        proxy_cache_lock_timeout: 30s
        client_body_buffer_size: 5m
        proxy_ignore_headers: Set-Cookie
+       limit_reqs:
+         - zone_name: "app_downloads"
```
emord: was this unused before?

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -64,6 +68,10 @@ nginx_sites:
        proxy_next_upstream_tries: 1
        proxy_read_timeout: 900s
        proxy_cache: app_cache
+       limit_reqs:
+         - zone_name: "multimedia"
```
emord: From #server-fires it sounds like you also added something around limiting video files specifically? Or maybe that was just a suggestion of what we could do?

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
snopoke: yes

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
snopoke: Yes, we could split this out more. Another option would be to limit the bandwidth consumption here as well as just the requests per second:

https://nginx.org/en/docs/http/ngx_http_core_module.html#limit_rate

PR #3226, opened 2019-09-05
Add decision record about removal of hotfix_deploy

emord:
##### SUMMARY
Following up for from https://github.com/dimagi/commcare-cloud/pull/3204#issuecomment-528245936

@snopoke I wasn't sure on the 100% on what this should exactly cover, but I think I got the main points in here.

##### ENVIRONMENTS AFFECTED
none

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/decisions/0003-implement-one-command-to-deploy.md
+++ b/decisions/0003-implement-one-command-to-deploy.md
@@ -1,0 +1,40 @@
+# 3. Implement one command to deploy
+
+Date: 2019-09-05
+
+## Status
+
+Accepted
+
+## Context
+
+During some refactoring of our static file versioning code, there existed several issues which caused static files to be served incorrectly.
+These errors propagated in several different ways across the site, but the root issue was always that there was a bug around how JS files were being served.
+
+Full context can be found in the following PRs:
+
+- Initial refactoring: https://github.com/dimagi/commcare-hq/pull/24938
+- Failed attempts at fixing `force_update_static`:
+    - https://github.com/dimagi/commcare-cloud/pull/3148
+    - https://github.com/dimagi/commcare-cloud/pull/3186
+- Removal of `hotfix_deploy` and `force_update_static`: https://github.com/dimagi/commcare-cloud/pull/3204
+- Simplification of the deploy static file process: https://github.com/dimagi/commcare-hq/pull/25284
+
+## Decision
+
+We are removing the previous commands `hotfix_deploy` and `force_update_static`.
+Any future deploys must go through `commcare-cloud <env> deploy`
+
+## Consequences
+
+The `hotfix_deploy` command skipped over some unnecessary steps when deploying python only changes.
+This resulted in a somewhat faster deploy for urgent tasks.
+Since `hotfix_deploy` was originally implemented, many changes to our deploy process have been implemented that have made our standard deploy process faster and more stable than before.
+
+If we find that faster deploys are a necessity, we will invest in improving our standard deploy process.
+This may include, but not limited to:
+- Batching restarts of our webworkers at the end of the deploy
+- If no static files have changed since last deploy, skipping collect_static and other tasks
+- If no couch views have been changed, skipping syncing couch views
+- If no migrations have been introduced, skipping any migrations checks
+- If no dependency files have changed skip `pip install`, `bower update` etc


decisions/0003-implement-one-command-to-deploy.md
```diff
@@ -0,0 +1,40 @@
...
+The `hotfix_deploy` command skipped over some unnecessary steps when deploying python only changes.
+This resulted in a somewhat faster deploy for urgent tasks.
+Since `hotfix_deploy` was originally implemented, many changes to our deploy process have been implemented that have made our standard deploy process faster and more stable than before.
+
+If we find that faster deploys are a necessity, we will invest in improving our standard deploy process.
+This may include, but not limited to:
+- Batching restarts of our webworkers at the end of the deploy
+- If no static files have changed since last deploy, skipping collect_static and other tasks
+- If no couch views have been changed, skipping syncing couch views
+- If no migrations have been introduced, skipping any migrations checks
```
emord: this is already implemented in a way https://github.com/dimagi/commcare-hq-deploy/blob/f6537d991cc7d5e9fc99cf164ca260a1d74a653f/fab/operations/db.py#L106-L120

but could change it to a `git diff` style check if really necessary

decisions/0003-implement-one-command-to-deploy.md
```diff
@@ -0,0 +1,40 @@
...
+Any future deploys must go through `commcare-cloud <env> deploy`
+
+## Consequences
+
+The `hotfix_deploy` command skipped over some unnecessary steps when deploying python only changes.
+This resulted in a somewhat faster deploy for urgent tasks.
+Since `hotfix_deploy` was originally implemented, many changes to our deploy process have been implemented that have made our standard deploy process faster and more stable than before.
+
+If we find that faster deploys are a necessity, we will invest in improving our standard deploy process.
+This may include, but not limited to:
```
emord: I'm not actually suggesting we do all of these. Just wanted to provide some examples of potential avenues of work (batching restarts would be dope on ICDS though)

dannyroberts: That might not be _all_ that much work to do, but I want to bring up one detail that we have to get right (but which isn't all that hard either). On production (not sure about ICDS) we split `hq_webworkers` and `mobile_webworkers`, so we'd have to batch separately within those two pools to make sure that at least x% of hq_webworkers and x% of mobileworkers are up at any given time.

emord: I've been thinking about that recently and I was wondering about how much we care about no downtime deploys for mobile vs hq. Because we've historically invested in offline work for mobile, do we actually care if there's a 2 min downtime during a low period once a day? That's not acceptable for HQ because its much more visible and nothing would work offline, but for mobile i'm not 100% convinced. (with all that said no downtime deploys are always better)

This also would only apply if web apps is hitting hq workers and not mobile workers (and we didn't want to change that).

dannyroberts: Web Apps hits HQ workers and formplayer; formplayer hits mobile workers. So if mobile workers are down Web Apps are highly degraded and essentially down (i.e. you can't sync which happens every time you enter an app).

But a less extreme version would be to always to mobile workers in 2 batches and web workers in 3 or 4, showing a different preference for the importance of each in being able to meet 100% of request demand

snopoke:
Thanks @emord this is great.

decisions/0003-implement-one-command-to-deploy.md
millerdev: Web Apps requests are always routed to a single mobile worker per client session (I'm pretty sure about this, but please let me know if I'm not applying that correctly here). This means that each client will have a "highly degraded and essentially down" period regardless of whether they are all restarted in parallel or serially.

snopoke: I think there's some confusion here. Requests from the web to formplayer are routed to a formplayer instance based on client session (they don't go to webworkers at all).

Formplayer requests to HQ (submissions / restores) I think always go to the URL specified in the app profile (so would go via the public API like any other mobile request).

emord: FWIW my confusion was based on thinking about how on ICDS formplayer points to icds-cas.gov.in, but all mobile points to cas.commcarehq.org. https://github.com/dimagi/commcare-cloud/pull/975/ which is not generally applicable.

PR #3227, opened 2019-09-05
Configure ICDS combined pillows

emord:
##### SUMMARY
No effect until https://github.com/dimagi/commcare-hq/pull/25285 is deployed. The default is that these are True, so there is no need to deploy other environments.

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
localsettings/pillows


emord:
deployed

PR #3229, opened 2019-09-05
Switch to a new datadog monitors config arrangement

dannyroberts:
https://dimagi-dev.atlassian.net/browse/PLAT-199

I want to get feedback on this new layout for datadog environments. The main requirements were
- Allow more than one datadog "group" with different notification settings, etc.
- map monitor conf files to datadog monitor ids differently for each group

and then I also decided to put the files in a fixed place with a fixed naming scheme that the command can then rely on. The idea (which is not implemented here) is that you could the use `manage-commcare-cloud update-datadog-monitors icds` or `manage-commcare-cloud update-datadog-monitors saas`.

Also I decided to assign all existing monitors to the ICDS group and create new monitors for the SaaS group. I figure this way the ICDS monitors—which I mostly don't see but which are very important to the ICDS team—remain largely unchanged, while the SaaS ones—which I see all the time and will be able to tell if there's an issue with—get recreated from scratch.

Anyway, let me know what you think about this layout. Once we settle on something we all like, I can go ahead and implement the code changes that support the new config format and the new functionality


dannyroberts:
@snopoke Yeah I think there's something desirable about that, but I couldn't figure out how to do it in a nice way. There's the more stylistic awkwardness of the "production" environment also covering other envs, but at a more practical level, the code will need to look at _all_ the files (including `.ignore.yml`) to determine the set of monitor_ids it knows about in order to be able to tell the user about un-synced monitors. We could get rid of that feature, but I think knowing that everything's covered (or intentionally ignored) is a nice feature.

PR #3234, opened 2019-09-09
limit MM download bandwidth to 500k per request

snopoke:
##### SUMMARY
Limit the bandwidth usage per request for multimedia downloads. The limit will only kick in after the first 100k has been sent which mean it should only get applied to larger files like videos.

This works in combination with the rate limit to limit the bandwidth usage. 

At this rate a 12M file will take ~30s to download.

Setting the rate limit to 50r/s means our max bandwith usage for MM would be 24MiB/s.

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
nginx

We could also add `limit_rate_after 100k` to only limit larger files but that makes it much harder to know what our max usage will be.

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
@@ -72,6 +72,7 @@ nginx_sites:
          - zone_name: "multimedia"
            burst: 50
            nodelay: True
+       limit_rate: 500k
      - name: "/a/icds-test/apps/download"
        balancer: "{{ deploy_env }}_cas_commcare"
        proxy_redirect: "http://{{ CAS_SITE_HOST }} https://{{ CAS_SITE_HOST }}"


emord:
Is there any concern about how long the connections will need to stay open for this? Sounds like if all users are downloading some 12 M file then there could be 1500 connections just from that (50 initiated per second * time it takes to download that file).

> The limit will only kick in after the first 100k has been sent which mean it should only get applied to larger files like videos

> We could also add `limit_rate_after 100k` to only limit larger files but that makes it much harder to know what our max usage will be.

I'm confused by these two statements. Why would we add this line if its already being applied? Or what does adding that line give us?

snopoke:
> Is there any concern about how long the connections will need to stay open for this? Sounds like if all users are downloading some 12 M file then there could be 1500 connections just from that (50 initiated per second * time it takes to download that file).

The connections count does get high but it hasn't been a problem for so far.

>> The limit will only kick in after the first 100k has been sent which mean it should only get applied to larger files like videos

>> We could also add limit_rate_after 100k to only limit larger files but that makes it much harder to know what our max usage will be.

> I'm confused by these two statements. Why would we add this line if its already being applied? Or what does adding that line give us?

Sorry for the confusion. When I first created the PR I had the `limit_rate_after` in place but then removed it as I was concerned we'd go above our bandwidth limit.

PR #3236, opened 2019-09-09
exclude username from app download cache key

snopoke:
##### SUMMARY
App download caching has been broken for a long time (since `username` was added as a request param): https://github.com/dimagi/commcare-android/pull/1747

This changes the cache key to exclude username. 

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
nginx cache

##### ADDITIONAL INFORMATION
This takes a the approach or stripping out the username but leaving the rest of the URL intact. I felt this was safer than only using specific parts of the URI in the key (https://github.com/dimagi/commcare-cloud/commit/566ed45d8c6d9e3358e6fc8fb225965eef82f56a) since if the URI changes the cache may not work but at least we won't be serving incorrect resources.

I used https://github.com/nginxinc/NGINX-Demos/tree/master/nginx-regex-tester to test the regexes.

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
@@ -21,6 +21,12 @@ nginx_sites:
         # If the header is not provided, do not rate limit
         - key: '""'
           value: '""'
+    # Extract request path from request_uri (https://stackoverflow.com/a/45587134)
+    - name: "$request_uri_path"
+      variable: "$request_uri"
+      var_mappings:
+        - key: '"~^(?P<path>[^?]*)(\?.*)?$"'
+          value: '$path'
    limit_req_zones:
     - name: "incremental_restore" # Limits restores that are not new device or 412s
       zone: "$incremental_sync"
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -21,6 +21,19 @@ nginx_sites:
         # If the header is not provided, do not rate limit
         - key: '""'
           value: '""'
+    # Remove 'username' query param from request uri for use in cache key
+    # https://stackoverflow.com/a/45587134
+    # https://stackoverflow.com/a/35772678
+    - name: "$request_uri_no_username"
+      variable: "$request_uri"
+      default_value: "$request_uri"
+      var_mappings:
```
snopoke: I wasn't able to get a single regex to work for all combinations of param placement so just went with 2 versions + a default fall through

snopoke:
Note that if ICDS changes to use location based releases then this will need to be removed

mkangia:
> Note that if ICDS changes to use location based releases then this will need to be removed

I see the possibility of that feature flag being enabled while this change still active. ICDS won't use this for a while so if this is a temporary change and you would remove this soon then it's fine.
Otherwise why not handle this in the codebase [here](https://github.com/dimagi/commcare-hq/blob/2f7685481a28353ed619e46b2d560e54eb26e2ac/corehq/apps/app_manager/decorators.py#L105)?

snopoke:
> Otherwise why not handle this in the codebase here?

@mkangia handling this in the codebase puts all the load onto the webworkers and DBs. Yesterday when the caching wasn't working the site almost went down due to very high load on webworkers, redis and couch.

If the long term plan is to move to location based releases we'll need to find a way to cache the resources better than we do now.

PR #3240, opened 2019-09-09
Use Fabric3 (python3 compatible fork) instead of Fabric

dannyroberts:
##### SUMMARY

https://pypi.org/project/Fabric3/
Release version numbers mirror Fabric version numbers, followed by
.post1, .post2, etc. Currently pinning to the latest "post" for the
previously-pinned version, 1.10.2.

##### ENVIRONMENTS AFFECTED
all

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
deploy / fab

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,7 @@ dimagi-memoized==1.1.1
 dnspython==1.16.0
 docutils==0.14            # via botocore
 enum34==1.1.6             # via cryptography
-fabric==1.10.2
+fabric3==1.10.2.post3
 futures==3.2.0            # via s3transfer
 gevent==1.4.0             # via couchdb-cluster-admin
 greenlet==0.4.15          # via gevent
@@ -39,7 +39,7 @@ jsonobject==0.9.9
 markupsafe==1.1.1         # via jinja2
 ndg-httpsclient==0.5.1
 netaddr==0.7.19
-paramiko==2.4.2           # via fabric
+paramiko==2.4.2           # via fabric3
 passlib==1.7.1
 pyasn1==0.4.5             # via ndg-httpsclient, paramiko
 pycparser==2.19           # via cffi
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
         'datadog==0.2.0',
         'dimagi-memoized>=1.1.0',
         'dnspython',
-        'Fabric==1.10.2',
+        'Fabric3>=1.10.2,<1.11',
         # can remove once requests bumps its version requirement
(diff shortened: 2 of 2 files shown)


dannyroberts:
I did a staging deploy using the _latest_ version of Fabric3 without any problems, but I haven't tested the version pinned here (which matches the version we currently have Fabric pinned at). I'll give that a try as well

emord:
documenting this down because this project history is non-standard.

- `fabric` - The original project. now supports python 2 and 3. fabric 2.x is incompatible for the most part with fabric 1.x.
- `fabric2` - The exact same fabric project. versions begins at 2.x. created so people could have fabric 1.x and 2.x installed at the same time
- `fabric3` - fork of fabric 1.x to support python 3. will never be updated with new features. tells you to use `fabric` version 2.x

did you ever happen to tryout using `fabric` version 2.x? From the docs it appears that the original release did not have some of the same features as 1.x, but future releases were intended to. Do we know that our commands don't work?

> I did a staging deploy using the latest version of Fabric3 without any problems

Did you do this with a new python3 env, or your old python 2 one?

dannyroberts:
Nice, thanks for the added context there. I tested on python2 only. I'm not saying we're actually ready to shift to python3 for commcare-cloud, just that this is a potential unblocker for the fabric question. That said, I'm just believing the docs / purpose of the project that fabric3 works in python3; I haven't actually tested it.

dannyroberts:
> did you ever happen to tryout using fabric version 2.x? From the docs it appears that the original release did not have some of the same features as 1.x, but future releases were intended to. Do we know that our commands don't work?

I didn't know this bit, that they were planning on eventual backwards compatibility. I didn't test it and I don't know for sure that our commands won't work.

Do you think it'd be better to try that before merging this? I guess in the long term (if we don't remove fab altogether) we'd prefer to be on fabric>=2 than on fabric3 (but maybe only if it's really easy to do that)?

czue:
I don't believe current `fabric2` supports the Python2 API. I have successfully used `fabric3` on Python3 on other projects with no issues. I've also ported a project from Python2 fabric to Python3 which took a little while to wrap my head around, but wasn't so bad (though it was much simpler than ours).

Let me know if you ever think I can help on anything

dannyroberts:
Thanks @czue for that background. I've now used it for a few different things (deploy, etc.) without any issue, and the fact that you've used it before and never had an issue with it, plus their systematic version tracking of Fabric versions gives me pretty high confidence that this is a safe change. It's also super easy to revert if anyone ever has an issue with it since it's a clean swap-in replacement

PR #3241, opened 2019-09-09
Add interface for deploying formplayer

dannyroberts:
##### SUMMARY

Following up on https://github.com/dimagi/commcare-cloud/pull/3174. I realized we currently didn't have an "official" way to deploy formplayer ever since pulling out formplayer_deploy from normal deploy and shortly after switching to the deploy command.

Normal deploy includes a reminder to deploy formplayer occasionally with
the command to do so.

Normal deploy:
Screen Shot 2019-09-09 at 5 43 46 PM

Formplayer deploy:

Screen Shot 2019-09-09 at 5 59 44 PM

##### ENVIRONMENTS AFFECTED
all

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
Deploy (formplayer)

### DIFF
--- a/src/commcare_cloud/commands/deploy.py
+++ b/src/commcare_cloud/commands/deploy.py
@@ -1,3 +1,5 @@
+from clint.textui.colored import yellow, blue
+
 from commcare_cloud.alias import commcare_cloud
 from commcare_cloud.cli_utils import check_branch, ask
 from commcare_cloud.commands import shared_args
@@ -12,6 +14,9 @@ class Deploy(CommandBase):
     )
 
     arguments = (
+        Argument('component', nargs='?', choices=['commcare', 'formplayer'], default='commcare', help="""
+            The component to deploy.
+        """),
         Argument('--resume', action='store_true', help="""
             Rather than starting a new deploy, start where you left off the last one.
         """),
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/commands/deploy.py
```diff
@@ -28,14 +33,39 @@ class Deploy(CommandBase):
...
+                                                        quiet=args.quiet)
+        if args.component == 'commcare':
+            print(blue("You are about to deploy commcare"))
+            print(blue("branch: {}".format(commcare_branch)))
+            if ask('Deploy commcare?', quiet=args.quiet):
+                print(yellow("Formplayer will not be deployed right now,"))
+                print(yellow("but we recommend deploying formplayer about once a month as well."))
+                print(yellow("It causes about 1 minute of service interruption to Web Apps and App Preview,"))
+                print(yellow("but keeps these services up to date."))
+                print(yellow("You can do so by running `commcare-cloud <env> deploy formplayer`"))
```
emord: how long do you plan to keep this warning (info?) for?

Also do we have a standard of what colors in the terminal means? It might e good to have a page on the meaning behind each one so that we stay consistent

dannyroberts: Having quick color guide (that includes something like 3-5 colors) would be useful for consistency. So far my personal color guide has been something like
- never use explicit `black` or `white` (since you don't know the user's background color)
- use `cyan` for printing out which other shell commands are being run in the background (like the underlying `ansible`, `ansible-playbook`, `fab`, `terraform`, etc. bash command being run)
- default/no text color for "Do you want to X [y/N]?" prompts
Things we've been less consistent about are when to use `yellow`, `blue`, `magenta`, and `red`. I might propose the following guidelines for them
- `yellow` for PSAs/notices, and for highlighting the part of a longer message that is most likely to be useful as a tl;dr or refresher.
- `blue` for summary statements that are part of a longer scrolling output (like "Your private release is located here:" during deploy)
- `magenta` for warnings likely to be problematic
- `red` for errors

Thoughts?

dannyroberts:
> how long do you plan to keep this warning (info?) for?

I currently don't have a better plan than keeping until we're better able to detect whether a formplayer deploy is necessary on a specific deploy, at which case I'd change it to print nothing most times and a more specific message about a new formplayer version being available other times.

Does that seem reasonable?

The other thing I still have to do is move formplayer to a more neutral location (one that's not under the commcare-hq release/code root) so we no longer have to restart formplayer on every deploy (despite not having updated it) to avoid java errors from changed paths/symlinks.

src/commcare_cloud/commands/deploy.py
emord: On second thought, I think that this shouldn't need docs. What if we had a `messages.py` where `warning = yellow` or whatever and we didn't use colors, but instead used `warning(text)`, `notice(text)` and so on. That would likely be more maintanable and more obvious to others without needing to go somewhere else for docs.

Also for conformity with other tools, we probably want to use ansible's color selections: https://docs.ansible.com/ansible/latest/reference_appendices/config.html?highlight=colors#color-changed. For example `purple` / `magenta` is used for deprecation warnings. (alternatively we could specify whatever we want in `ansible.cfg`)

emord:
> Does that seem reasonable?

That seems reasonable. I'm mostly curious what our "contract" to our users is. There was some initial disagreement on how long to support python 2 & 3 and I was wondering what these decisions look like going forward.

dannyroberts:
@emord I think we're still trying to figure out the answer to that, and in the meantime just trying to make all of our changes either (1) not affect anyone else (2) come with with a reasonable onramp to the behavior change or (3) come with a changelog. (Sometimes we've done both 2 and 3.)

src/commcare_cloud/commands/deploy.py
dannyroberts: OK followed up here: https://github.com/dimagi/commcare-cloud/pull/3245.

dannyroberts: As for
> Also for conformity with other tools, we probably want to use ansible's color selections: https://docs.ansible.com/ansible/latest/reference_appendices/config.html?highlight=colors#color-changed. For example purple / magenta is used for deprecation warnings. (alternatively we could specify whatever we want in ansible.cfg)

I think that's a good idea, and I think just adding the semantic aliases for the colors we already use is a good start towards that. In that PR I also copied in the ansible color scheme info as a comment for consideration when choosing colors. Since it wasn't super clear cut how our categories relate to theirs in every case, I didn't try to reconcile them in the same PR, but that could be a follow up

PR #3244, opened 2019-09-11
Double UCR form pillows on ICDS

emord:
##### SUMMARY
This doubles the kafka-ucr-static-forms pillows as a temporary measure. I'm doing this because of the issue described https://github.com/dimagi/commcare-hq/pull/25334. There are roughly 16 topics that have gotten over the hump of duplicate forms and doing this will ensure those stay up to date. Once the remaining topics get past that hump, they should also be able to get up to date much faster.

I do not intend to merge this as it should be reverted once 

##### ENVIRONMENTS AFFECTED
icds
##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
pillows

### DIFF
--- a/environments/icds-cas/app-processes.yml
+++ b/environments/icds-cas/app-processes.yml
@@ -258,26 +258,26 @@ pillows:
   'pillow23':
     kafka-ucr-static-forms:
       start_process: 0
-      num_processes: 12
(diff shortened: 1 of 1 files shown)


emord:
we just caught up on these pillows

PR #3245, opened 2019-09-11
Semantic CLI color aliases

dannyroberts:
##### SUMMARY
As @emord suggested in https://github.com/dimagi/commcare-cloud/pull/3241, this changes our codebase to use semantic color aliases.

In a few cases, I took the liberty of changing colors to match the semantic values and thus the new convention.


emord:
nice. thanks for doing that

PR #3249, opened 2019-09-11
Add check for old commcare-cloud install

dannyroberts:
##### SUMMARY
One difficulty with keeping things in sync across developers and environments is that it's fairly easy not to notice you're on an old version of commcare-cloud, and sometimes that can cause issues that happen silently (like something not getting deployed the right way).

This is an unpolished spike at just always warning the user if their install is older than X (here 10) days. Before pushing this burdensome workflow out or taking more time to improve it, I thought I'd just put the idea out there to get feedback about whether something like this would be a positive addition.

### DIFF
--- a/src/commcare_cloud/cli_utils.py
+++ b/src/commcare_cloud/cli_utils.py
@@ -4,9 +4,12 @@ import re
 import subprocess
 
 import sys
+import time
+from datetime import timedelta
+
 from clint.textui import puts
 
-from commcare_cloud.colors import color_error, color_code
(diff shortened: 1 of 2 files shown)


emord:
whenever we log into a control machine, it tries to install all dependencies right? Is there a reason why we couldn't add a `git fetch` and see if the current branch contains `origin/master`? 

There could be some exception handling if the network doesn't exist I suppose, but maybe that warrants its own warning? I would imagine that most code deploys are going to take more network than `git fetch` on commcare-cloud

PR #3253, opened 2019-09-12
Set up formplayer in ansible

dannyroberts:
##### SUMMARY

This rewrites formplayer deploy using ansible (in a place that also gets run during first time setup), followed by a reread / update / restart command. It also changes the place to which formplayer gets deployed so it's no longer inside the commcare-hq repo.

Still todo in this PR (or a followup?):
- remove the old formplayer deploy
- stop restarting formplayer during normal deploy
- port over the feature for clearing out old formplayer releases
- (future) do a better job figuring out if a formplayer update even exists before going through w/ it

##### ENVIRONMENTS AFFECTED
all (but mostly under the hood / invisible)

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
formplayer deploy


src/commcare_cloud/ansible/roles/formplayer/defaults/main.yml
```diff
@@ -3,3 +3,4 @@ formplayer_sentry_dsn: ''
 # (you can't use s for seconds, but m, h, d work for minute, hour, day)
 # formplayer_archive_time_spec: '10m'
 formplayer_purge_time_spec: '2d'
+formplayer_release_dir: "{{ formplayer_build_dir }}/formplayer__{{ new_release_name }}"
```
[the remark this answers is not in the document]
millerdev: that's fine, it's not a big deal

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -67,18 +67,6 @@
...
-    src: "{{ formplayer_release_dir }}/libs/formplayer.jar"
-    dest: "{{ formplayer_build_dir }}/formplayer.jar"
-    mode: 0544
-    owner: "{{ cchq_user }}"
-    group: "{{ cchq_user }}"
-  tags:
-    - formplayer_deploy
-
-
 - name: Copy Spring application.properties
```
millerdev: Looks like Spring and Sentry properties as well as the formplayer jar are copied or linked directly into `formplayer_build_dir`. I think this means that if any of these deploy steps fails, then the build dir will potentially be in an inconsistent state (wrong properties for the linked jar, for example) and a subsequent restart of formplayer may have a bad outcome. This is the issue I was getting at in https://github.com/dimagi/commcare-cloud/pull/3253#discussion_r323863437

I was expecting the solution would be to copy all of these files into a new _release dir_, and then link that _release dir_ to the _build dir_ (maybe better named _current dir_?) as the final step. Does that sound like a good solution, or is there something I'm overlooking?

dannyroberts: Thank you for persistence, because you're right, I totally missed the point you were trying to make earlier, specifically the part about making sure we update the config files in a separate release as well.

I think the overall risk given our current process is low; or rather I don't think this introduces risks that aren't already in our current process. Currently, `update-config` will update the sentry and spring property files for formplayer (along with localsettings.py for commcare-hq code), without restarting formplayer. (I'm not sure if the live formplayer process then picks up these changes dynamically, but my guess is that it doesn't.) Then separately the formplayer.jar during the formplayer deploy and then restarted. (Also formplayer is currently getting restarted without the jar getting updated on every deploy as a workaround for an issue.) All that said, I do think that you're right that we can do better than that, and that with all the very-related changes in this PR, this PR might be the right time to get this right.

Something I do want to do is keep nice diffs during the update process. Probably the easiest way to achieve that would be to start by copying the files from the `current` directory, and then update them with the templated code. That seems easy enough and shouldn't add more than a couple seconds to the `formplayer deploy`.

I'll investigate this and unless it gets messy for some reason I haven't thought of, I'll aim to include the changes I come up with in this PR.

Thanks again for the suggestion and persistence!

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -51,10 +51,34 @@
...
   with_items:
     - "{{ formplayer_build_dir }}"
+    - "{{ formplayer_release_dir }}/libs"
     - "{{ formplayer_data_dir }}"
 
+- name: Download formplayer release
+  get_url:
+    url: "{{ formplayer_build_url }}"
+    dest: "{{ formplayer_release_dir }}/libs/formplayer.jar"
+    mode: 0544
```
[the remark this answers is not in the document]
dannyroberts: Hmm, I just realized it doesn't make much sense for this to have the `x` permission for anyone at all—to my knowledge `.jar` files are always run by an executable (`java`) and aren't ever called as executables themselves.

If that sounds right, I'm going to change this to 644.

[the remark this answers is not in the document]
millerdev: Yes... I should have noticed that too.

dannyroberts:
OK @millerdev can you do another pass at review for the new changes? I tested this on staging and the new formplayer deploy doesn't fail and seems to have all the desired properties

src/commcare_cloud/ansible/roles/formplayer/defaults/main.yml
```diff
@@ -3,4 +3,5 @@ formplayer_sentry_dsn: ''
 # (you can't use s for seconds, but m, h, d work for minute, hour, day)
 # formplayer_archive_time_spec: '10m'
 formplayer_purge_time_spec: '2d'
-formplayer_release_dir: "{{ formplayer_build_dir }}/formplayer__{{ new_release_name }}"
+formplayer_release_dir: >
+  {{ formplayer_build_dir }}/{{ "releases/" + new_release_name if new_release_name|default(None) else "current" }}
```
millerdev: Under what conditions will this evaluate to `{{ formplayer_build_dir }}/current`? Like when will `new_release_name` be blank or undefined? Is that expected to happen frequently?

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -54,24 +54,44 @@
...
     mode: 0744
     owner: "{{ cchq_user }}"
     group: "{{ cchq_user }}"
   tags:
     - formplayer_deploy
 
+- name: Check if the "current" directory exists
+  stat:
+    path: "{{ formplayer_build_dir }}/current"
+  register: current_stat
```
millerdev: should this have `changed_when: false`?

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -54,24 +54,44 @@
...
+  copy:
+    remote_src: yes
+    src: "{{ formplayer_build_dir }}/current/{{ item }}"
+    dest: "{{ formplayer_release_dir }}/{{ item }}"
+  with_items:
+    - application.properties
+    - sentry.properties
+  tags:
+    - formplayer_deploy
+  when: new_release_name|default(None) and current_stat.stat.exists
```
millerdev: I don't understand why we're copying these into the "current" directory, which is normally symlink? Will this `when` condition evaluate to true when `current` is a symlink? It seems like this will defeat the purpose of the atomic commit.

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
dannyroberts: Oh, does `stat` fail if the path isn't there? I assumed it just set `.stat.exists` to false.

src/commcare_cloud/ansible/roles/formplayer/defaults/main.yml
dannyroberts: `new_release_name` will be null when this is being invoked from the `update-config` command: https://github.com/dimagi/commcare-cloud/pull/3253/files#diff-384a623aa676400249bff4d290514e36R224 (or whenever else we choose to have this file apply in-place to the live `current` directory instead of making a new release).

The question confirms my suspicion as I was writing this last night that maybe the this of `new_release_name` is not very clean. My overall purpose is to have this file be able to be run in two modes:
- the most common mode, in which a clean release is made and then current is symlinked to the release
- the in-place mode where files are updated directly in the current release (to maintain the current behavior of `update-config`).

Would love to hear your thoughts on how I can do that in a cleaner way

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
dannyroberts: My intention (which I think this does) is that this task copy files _from_ the `current` directory (if the `current` directory exists and the `formplayer_release_dir` isn't already set to the path of the `current` directory in which case we'd be copying a file to itself) to the new release directory.

This isn't strictly necessary since these files get overwritten by the following two tasks, which recreate the file contents from a template. It does however have the benefit of having the template tasks print out a diff from the existing files instead of just printing the full output with a `+` before each line.

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
millerdev: No, but ansible will detect it as a change (and note it as such in the final report) when in fact it was just a check to set a local variable.

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
millerdev: > This isn't strictly necessary since these files get overwritten by the following two tasks ...

Ah, yes! I saw that they were being overwritten below, and that was adding to the confusion.

> It does however have the benefit of having the template tasks print out a diff from the existing files...

That makes a lot of sense—meaningful diffs are very nice. What do you think of adding a comment to explain that the reason for this task is so the following tasks produce meaningful diffs?

src/commcare_cloud/ansible/roles/formplayer/defaults/main.yml
millerdev: > maybe the this of `new_release_name` is not very clean

I think you missed a word in there.

> Would love to hear your thoughts on how I can do that in a cleaner way

It's not clear to me if `update-config` copies the formplayer jar into the "current" dir? If yes, that seems like a bit more than a config update. Do you agree? Maybe that's part of the problem? If not...

Maybe add another variable. Something like
```yaml
formplayer_release_dir: "{{ formplayer_build_dir }}/{{ "releases/" + new_release_name }}"
formplayer_config_dir: "'{{ formplayer_build_dir }}/{{ "releases/" + new_release_name if new_release_name else "current" }}'"
```

Config files are written to the config dir, which sometimes points to "current". Release files (the jar) are always written to the release dir (and this does not happen on `update-config`).

I don't know if that's much clearer. There are a fair amount of moving parts here (different things happen in different run modes), and that complexity makes it hard to reason about.

dannyroberts: Yeah `update-config` definitely shouldn't and I'm pretty sure doesn't download/update/touch  `formplayer.jar`. It runs this with `--tags=localsettings` so only tasks with that tag (which is just the two that update the config files `application.properties` and `sentry.properties`) run.

OK I like where you're suggestion is going. I think I'll also stop inlining the "/current" postfix everywhere and make a variable for that as well. Hopefully what I end up with is a bit more readable

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
dannyroberts: Oh! `changed_when`, I was obviously thinking `failed_when`.

Still though I'm not seeing it report this as a change

```
TASK [formplayer : Check if the "current" directory exists] ************************************************************************************************************************************************
ok: [10.201.10.234]
```

That's what I would expect (`stat` just pulls info about the path, so I wouldn't expect it to report a change), but it sounded like you were reporting that `stat` sometimes has weird behavior?

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
dannyroberts: Okay yeah a comment sounds great

dannyroberts:
(I had pushed a bunch of changes to this PR, but then realized it might be nice to review the new changes separately, and also to get these changes in sooner. So then I force pushed to only include a single commit that adds a comment.)

PR #3257, opened 2019-09-13
fix update known hosts

snopoke:
##### SUMMARY
Update local known hosts did not correctly remove the host from the host file before adding it causing duplicates in the file.

##### ENVIRONMENTS AFFECTED
All

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
update-local-known-hosts


dannyroberts:
```
======================================================================
ERROR: test_environments.test_all('icds-cas', {})
----------------------------------------------------------------------
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/parameterized/parameterized.py", line 387, in <lambda>
    nose_func = wraps(func)(lambda *args: func(*args[:-1], **args[-1]))
  File "/home/travis/build/dimagi/commcare-cloud/src/commcare_cloud/manage_commcare_cloud/tests/test_environments.py", line 14, in test_all
    environment.check()
  File "/home/travis/build/dimagi/commcare-cloud/src/commcare_cloud/environment/main.py", line 45, in check
    self.check_known_hosts()
  File "/home/travis/build/dimagi/commcare-cloud/src/commcare_cloud/environment/main.py", line 81, in check_known_hosts
    "{} ({})".format(sshable, hostname) for sshable, hostname in missing_hosts
EnvironmentException: The following hosts are missing from known_hosts:
100.71.188.133 (celery5)
```

snopoke:
@dannyroberts I found the issue was that the ansible playbook runs the local tasks to update the known_hosts file in parallel. Changing this to run serially fixes the issue but makes the command quite slow.

Considering that all of these tasks run locally it didn't seem necessary to use ansible for this at all and we could optimize this in python. The other benefit of doing that is we can remove old entries from the known_hosts files.

Let me know what you think.

dannyroberts:
Nice, I like that solution. Does that mean we can get rid of the entire `add-ssh-keys.yml` file as well?

snopoke:
> Does that mean we can get rid of the entire add-ssh-keys.yml file as well?

Yes d0c1420cc

PR #3269, opened 2019-09-18
Stop restarting formplayer during normal deploy

dannyroberts:
##### SUMMARY
I think we're finally ready to stop restarting formplayer during normal deploy. This PR will continuing restarting it as long as it detects that formplayer is running from the old release location (in which case the restart is necessary). Any env that has had `cchq <env> deploy formplayer` run on it since 16 hours ago or so will have formplayer running from the new location, and will not have formplayer restarted on normal deploys after this PR is merged.

I've tested the old/new formplayer location detection logic on staging (new) and india (old), and it worked as intended.

##### ENVIRONMENTS AFFECTED
all

##### ISSUE TYPE
- Change/Bugfix Pull Request

##### COMPONENT NAME
formplayer

### DIFF
--- a/src/commcare_cloud/fab/operations/formplayer.py
+++ b/src/commcare_cloud/fab/operations/formplayer.py
@@ -132,3 +132,11 @@ def _get_old_formplayer_builds(build_dir):
         old_builds = sorted(_get_builds(previous_build_paths.split('\n')), reverse=True)
         old_builds.remove(current_build)
         return old_builds
+
+
+def formplayer_is_running_from_old_release_location():
+    ps_formplayer = sudo('ps aux | grep formplaye[r]')
+    if env.code_current in ps_formplayer:
+        return True
+    else:
+        return False
--- a/src/commcare_cloud/fab/operations/supervisor.py
+++ b/src/commcare_cloud/fab/operations/supervisor.py
@@ -107,6 +107,15 @@ def restart_formplayer():
     _services_restart()
 
 
+@roles(ROLES_FORMPLAYER)
+def restart_formplayer_if_it_is_running_from_old_release_location():
+    from commcare_cloud.fab.operations.formplayer import \
+        formplayer_is_running_from_old_release_location
+
+    if formplayer_is_running_from_old_release_location():
+        _services_restart()
+
+
 def _services_restart():
     """Stop and restart all supervisord services"""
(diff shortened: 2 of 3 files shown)


dannyroberts:
India, Staging, and Production are all on the new formplayer layout, so once we get this merged in, I think we'll finally be done with the daily formplayer downtime!

src/commcare_cloud/fab/operations/formplayer.py
```diff
@@ -132,3 +132,11 @@ def _get_old_formplayer_builds(build_dir):
         old_builds = sorted(_get_builds(previous_build_paths.split('\n')), reverse=True)
         old_builds.remove(current_build)
         return old_builds
+
+
+def formplayer_is_running_from_old_release_location():
+    ps_formplayer = sudo('ps aux | grep formplaye[r]')
+    if env.code_current in ps_formplayer:
```
millerdev: nit
```suggestion
    return env.code_current in ps_formplayer
```
and drop remaining lines in function

src/commcare_cloud/fab/operations/supervisor.py
```diff
@@ -107,6 +107,15 @@ def restart_formplayer():
     _services_restart()
 
 
+@roles(ROLES_FORMPLAYER)
+def restart_formplayer_if_it_is_running_from_old_release_location():
+    from commcare_cloud.fab.operations.formplayer import \
+        formplayer_is_running_from_old_release_location
+
+    if formplayer_is_running_from_old_release_location():
+        _services_restart()
```
millerdev: This appears to be starting/stopping all services. Does "all services" mean "only formplayer" in this context?

dannyroberts: This maintains the same blemish the old code has which is it treats machines as either "monolith" (in which case you don't need to restart formplayer separately from "all"), or _just_ formplayer (in which case it restarts "all" as a way to restart just formplayer).

In any case, this PR makes this _not_ happen as long as you've run `deploy formplayer` since the formplayer location change from the last couple days.

src/commcare_cloud/fab/operations/formplayer.py
dannyroberts: https://github.com/dimagi/commcare-cloud/pull/3279

PR #3271, opened 2019-09-18
Fix issue where new formplayer deploy leaves it unable to find config files

dannyroberts:
##### SUMMARY
This wasn't caught because I'd done a few iterations on staging and by chance there were config files in the place it was looking. This updates it to look for the config files in the correct place.

##### ENVIRONMENTS AFFECTED
all

##### ISSUE TYPE
- Bugfix Pull Request


dannyroberts:
Because right now if someone ran `cchq production update-supervisor-confs` we'd be in a bad state, I'm going to merge this change before it's approved to get master back into a good state.

It occurred to me to instead revert the change that caused the issue in the first place and wait to merge it back in with this change until it was reviewed, but that came with its own potential risks, given that I have only tested the formplayer directory migration in the forward direction and not the backwards direction, and this small change is definitely the fix to the issue.

dannyroberts:
I'm sending out a retro with what I did, and I'm open to discussing on that thread whether the way I handled it afterwards (merging this in rather than reverting the larger change) was right

PR #3274, opened 2019-09-19
Update citus max connections

calellowitz:
##### SUMMARY

We have been facing a number of errors recently with citus connection issues, especially on the workers. I suspect it is because the master has more available connections than workers, and when it reaches out to the workers, they are already full of connections causing the query to fail. The workers do not have connection poolers in front of them to absorb the extra connections. This raises there connection limits above the pgbouncer pool number, and lowers that number a bit to provide some breathing room for things like manual connections.


calellowitz:
@snopoke 
I think that would be ideal, but I was not sure how possible that was. I know the various machines communicate with each other directly and I didn't know if putting pgbouncer in front would mess with that.

PR #3280, opened 2019-09-25
Add PHI-CAS env

mkangia:
##### SUMMARY
Adds a new environment phi-cas

##### ENVIRONMENTS AFFECTED
phi-cas

##### ISSUE TYPE
- Change Pull Request

##### ADDITIONAL INFORMATION
This is a monolith setup. It's suppose to just serve as a small instance of HQ that can be used to test an API integration. The outcome is suppose to be a very basic Application on this instance that can be installed on the phone and then at the backend when a new case is submitted by phone, the API integration is triggered to populate the case with a case property, called phi_id for now depending on the response. There would be some overhead maintaining the record of requests made. The implementation of that is yet to be finalised, the two options as of now are repeaters or case update rules.
We should not be really worried about taking backups or any sort of notifications, so i am happy to switch off any alert, notifications for this environment.
made the changes from a combination of swiss(from monolith perspective) and icds-cas(from CAS perspective)


environments/phi-cas/public.yml
```diff
@@ -0,0 +1,118 @@
+DATADOG_ENABLED: True
+DATADOG_INTEGRATIONS_ENABLED: True
```
snopoke: does it need datadog?

environments/phi-cas/public.yml
```diff
@@ -0,0 +1,118 @@
...
+elasticsearch_download_sha256: 78affc30353730ec245dad1f17de242a4ad12cf808eaa87dd878e1ca10ed77df.
+
+#elasticsearch_node_name: '??'
+elastcsearch_backup_days: 2
+
+redis_maxmemory: 4gb
+
+ssh_allow_password: False
+
+backup_blobdb: True
```
snopoke: should disable all backups

mkangia:
> Why do we need a separate env to test this on?

we have been asked to keep it separate from current TCL/ICDS-CAS setup.

environments/phi-cas/public.yml
mkangia: thought it would be useful to make sure server is performing as expected by visiting DD. Will it create noise? I can remove the slack channel mentioned.

environments/phi-cas/public.yml
mkangia: Done.

environments/phi-cas/public.yml
snopoke: ok fine

snopoke:
@mkangia let's keep this in a branch for now since I expect there's going to be some iteration on it once you start rolling it out.

mkangia:
> let's keep this in a branch for now

@snopoke sure simon. I wanted to make sure its good enough to actually even run once and then iterate on where it breaks. Once things are done, only then i would think about merging it or not even merging it since this is a throw away setup. So is it good run post i remove the slack channel?

sravfeyn:
This is probably a bit late suggestion (I just got the idea now), but if the only reason that a new env is being setup on this machine is that because its IP is whitelisted, can't we use a test domain on icds-cas.gov.in for this testing (the whitelisted IP would be still be same, since it's not VM specific but the client public IP of all our servers)

mkangia:
@sravfeyn 
i thought that consideration was already discarded, Simon even raised that earlier as a question, https://github.com/dimagi/commcare-cloud/pull/3280#issuecomment-534985876

Will test the API access for other servers just to be sure.

sravfeyn:
Ok, thanks, sounds fine if the burden of new env is weighed against other options.

mkangia:
thanks @sravfeyn 
i checked with Shayoni and looks like we need a separate server to keep things isolated from ICDS-CAS irrespective of the IP consideration so going to continue with it

mkangia:
@snopoke this is ready now.

mkangia:
However there is one more change i made [here](https://github.com/dimagi/commcare-cloud/blob/e0e133bfa4263d168659b494975525d6cda545e9/src/commcare_cloud/ansible/roles/nginx/templates/default.j2#L5) for PHI that i have not committed
```
 server {
   listen 127.0.0.1:80;
-  listen [::1]:80;

   location /nginx_stub_status {
```

the server does not support IPv6. I have deployed like this and things are working fine so if this is not serving any purpose, should this be removed so that people don't face issues when deploying on such servers?

environments/phi-cas/public.yml
```diff
@@ -0,0 +1,118 @@
...
+redis_maxmemory: 4gb
+
+ssh_allow_password: False
+
+backup_blobdb: False
+backup_postgres: False
+backup_es_s3: False
+backup_couch: False
+postgres_s3: False
+blobdb_s3: True
```
snopoke: should this be True?

environments/phi-cas/public.yml
```diff
@@ -0,0 +1,118 @@
...
+master_list_email: master-list@dimagi.com
+sales_email: sales@dimagi.com
+privacy_email: privacy@dimagi.com
+feedback_email: hq-feedback@dimagi.com
+eula_change_email: eula-notifications@dimagi.com
+contact_email: info@dimagi.com
+soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
+check_s3_backups_email: commcarehq-ops+root@dimagi.com
+
+TWO_FACTOR_GATEWAY_ENABLED: True
```
snopoke: Probably not True

snopoke:
Re-ipv6 issue - not sure why that failed for you - pretty sure all our other sites have it like that. I don't think remove it since it's working elsewhere.

mkangia:
Thanks @snopoke, updated.

mkangia:
> not sure why that failed for you

i looked at it briefly with Cal. He pointed that the server network does not support IPv6. When i contacted TCL they said that support for IPv6 is to be requested and is not added by default. Possibly something that has changed in templates that TCL is going to provide now but is made available on servers provisioned elsewhere. I am fine to leave that change in my local code for only PHI but good to note if fails for any new proxy machines being setup on TCL in future.

snopoke:
@mkangia :+1: thanks for the info

PR #3290, opened 2019-09-30
move async queue time to latter on ICDS

sravfeyn:
@emord 

Re. https://dimagi-dev.atlassian.net/browse/IIO-461


sravfeyn:
Actually should we move it out further into IST night time?

emord:
Deployed this and restarted the periodic queue.

> Actually should we move it out further into IST night time?

We can see how this turns out. I think technically we shouldn't even need this, except when we have a backlog of processing from change feeds like we have now

PR #3299, opened 2019-10-02
Adding AutoVacuum Params from 2ndquadrant report

rohit-dimagi:
##### SUMMARY

* Tuning of autovaccum and autoanalyze Parameters Based on 2ndquadrant report

##### ENVIRONMENTS AFFECTED

ICDS
##### ISSUE TYPE

- Change Pull Request


rohit-dimagi:
Some reads to understand the parameters. 
https://blog.gojekengineering.com/postgres-autovacuum-tuning-394bb99fe2c0
https://www.percona.com/blog/2018/08/10/tuning-autovacuum-in-postgresql-and-autovacuum-internals/
https://www.keithf4.com/per-table-autovacuum-tuning/

I would probably implement it first on pgsynclog0 db  which gets most table updates.
Also adding some Datadog metrics to analyze it ,if we have got the values right or not and tune it more if needed.
https://www.datadoghq.com/blog/postgresql-vacuum-monitoring/

src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
```diff
@@ -130,7 +130,13 @@ track_activity_query_size = 2048
 # AUTOVACUUM PARAMETERS
 
 autovacuum = on
-
+{% if not is_pg_standby and  inventory_hostname in groups['synclog_dbs'] or inventory_hostname in groups['shard_dbs'] %}
```
snopoke: why do we want to restrict this to `not is_pg_standby`?

Also if we want to customize it based on VM type we should use variables that get changed in the inventory files.

rohit-dimagi: > why do we want to restrict this to `not is_pg_standby`?
>
Vacuum and autovacuum replicate like any other write operation.Running vacuum or autovacuum on the slave doesn't do anything and is not necessary.

> Also if we want to customize it based on VM type we should use variables that get changed in the inventory files. l
> 
Like adding another Host Level Vars?

snopoke: > Vacuum and autovacuum replicate like any other write operation.Running vacuum or autovacuum on the slave doesn't do anything and is not necessary.

Sure but if we failover to a standby then we'd want this right? I don't think the autovacuum process runs when the DB is in recovery mode.

> Like adding another Host Level Vars?

Yes, or group level

rohit-dimagi: Updated

rohit-dimagi:
Updated PR with new changes.

rohit-dimagi:
ran the updates on pgsynclog machines

PR #3304, opened 2019-10-03
Double UCR form pillows on ICDS

emord:
##### SUMMARY
This is likely just a short term change for the next day or two until we get caught up on the backlog.

I've already deployed these new pillows but they are currently stopped until I'm able to finish adding a few indexes to our UCRs (upcoming HQ PR)

##### ENVIRONMENTS AFFECTED
icds
##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
pillowtop

### DIFF
--- a/environments/icds-cas/app-processes.yml
+++ b/environments/icds-cas/app-processes.yml
@@ -256,26 +256,26 @@ pillows:
   'pillow23':
     xform-pillow:
       start_process: 0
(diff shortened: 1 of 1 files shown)


emord:
As mentioned in description, this was short term. I've removed the extra pillows and master is consistent with the current state of the system

PR #3307, opened 2019-10-04
Whitelist IPs nginx

sravfeyn:
@snopoke 
The first commit is how it already exists, the second commit is the new change to actual list. 

Also, sidenote, there's ton of other manual changes to icds_cas_commcare that are not in ansible.


sravfeyn:
Updated.

PR #3310, opened 2019-10-05
cleanup after maintenance

snopoke:
##### SUMMARY
- citus cluster switched to new nodes on vmware
- commcarehq_ucr DB moved to pgmain0
- pgucr removed from localsettings

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request


snopoke:
Note for future reference until I find a better place to put it:

When moving a worker to a new node (or in this case failing over all to new nodes) you have to update the citus controller:

```sql
update pg_dist_node set nodename = '{{new_node_ip}}' where nodename = '{{old_node_ip}}';
update pg_dist_shard_placement set nodename = '{{new_node_ip}}' where nodename = '{{old_node_ip}}';
```

PR #3312, opened 2019-10-07
Require terraform < 0.11.15

dannyroberts:
##### SUMMARY

Ṫerraform refuses to work with shared state files created by a future version.
Today I ran into this when my terraform refused to operate because the shared
state file was created with terraform 0.11.15. My first attempt to solve
was to install 0.11.15, but this proved difficult, as my current method
of installing terraform on my Mac, Homebrew, did not recognize this version.
On terraform's own versions page https://releases.hashicorp.com/terraform/
0.11.15 is listed as "0.11.15-oci", which is nonstandard version naming
and the terraform github repo does not list 0.11.15 as a release at all.

It gives me no pleasure to force those with 0.11.15 to downgrade,
but it appears to be our best option here, given that Mac/Homebrew
is a common way to handle installs on our team, and the strange
nature of this 0.11.15 terraform release.

We will eventually want to support upgrading, but when we do, the effort will go into upgrading to 0.12.x which (despite the misleading version number) is a substantial and occasionally backwards incompatible upgrade.

### DIFF
--- a/src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
+++ b/src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
@@ -5,7 +5,7 @@ terraform {
     region  = {{ state_bucket_region|tojson }}
     encrypt = true
   }
-  required_version = "~> 0.11.0"
+  required_version = "~> 0.11.0, < 0.11.15"
 }
 
 provider "aws" {


dannyroberts:
FYI @dimagi/team-commcare-hq @sanjay2916 if you have terraform installed check its version with `terraform -v`. If it's `0.11.15`, please note that you'll have to downgrade to `0.11.14`. If you have any questions issues in the process, please reach out to me. Apologies for any inconvenience

emord:
> upgrading to 0.12.x which (despite the misleading version number) is a substantial and occasionally backwards incompatible upgrade.

FWIW that's acceptable when following semver https://semver.org/#spec-item-4 

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

PR #3314, opened 2019-10-08
Add gauge for number of locks not granted

emord:
##### SUMMARY
This adds a gauge for PG databases for number of locks that are blocking queries. I've used this quite a bit when changing the dashboard agg and also understanding our recent pillow issues.

I think it would be useful to have as a metric to better understand how this fluctuates throughout the day. I'm not sure how useful it will be to non-ICDS environments, but it doesn't seem like it would hurt

##### ENVIRONMENTS AFFECTED
all

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
postgres


emord:
I deployed this to all our postgres servers. I also added a line graph "Locks not granted (by host)" to our Postgres Overview dashboard.

PR #3315, opened 2019-10-08
Remove cloudant checks

emord:
##### SUMMARY
Noticed this while implementing a different datadog check. I think that this can be removed along with https://github.com/dimagi/datadog-checks/tree/98c6197fa23435cf5cc303d612c7ddf4e57e0658/cloudant. Will let @dannyroberts verify.

##### ENVIRONMENTS AFFECTED
production

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
datadog

##### ADDITIONAL INFORMATION
(only for production)

```bash
git checkout 87715d8f1e492154c0ea427995741793d7a4ccef  # may need to create a temporary branch?
cchq production aps --tags=datadog_integrations
```


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -577,8 +577,7 @@ INDICATOR_NAMESPACES = {
         ('mvp_indicators', "MVP"),
     ),
 }
-# should be 'ok' if couchdb version < 1.1.0, or cloudant, else 'update_after'
-COUCH_STALE_QUERY = '{{ localsettings.COUCH_STALE_QUERY | default('update_after') }}'
+COUCH_STALE_QUERY = 'update_after'
```
dannyroberts: I think this is also the default: https://github.com/dimagi/commcare-hq/blob/5aca7bd9d95b0805fec61417df6ffd65c618f252/settings.py#L735 FWIW. What do you think about just removing this line entirely?

emord: I'm not opposed to it. The main reason I didn't was that it would introduce a change to all environments localsettings which is a pain to deploy and introduces a bit of overhead. As it is now the PR only needs to be deployed to one environment.

emord:
@dannyroberts I missed your thumbs up on my comment. I believe this only effects production (just a minor cleanup for the general repo), so I'll let you merge it as SaaS lead. Let me know if there's anything else I should do.

dannyroberts:
I'm all set to roll this out in the morning, but will hold off since I'm about to leave

dannyroberts:
(done)

PR #3326, opened 2019-10-16
Use all current RIs

dannyroberts:
##### SUMMARY
Because of having shifted around some instance class types in the last few months, we're currently slightly underutilizing the Reserved Instances of the t3 and m5 instance types.

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
Rolled out

PR #3327, opened 2019-10-17
increase case-pillow processes

sravfeyn:
https://dimagi-dev.atlassian.net/browse/IIO-515

@emord @calellowitz

### DIFF
--- a/environments/icds-cas/app-processes.yml
+++ b/environments/icds-cas/app-processes.yml
@@ -277,32 +277,32 @@ pillows:
   'pillow27':
     case-pillow:
       start_process: 0
-      num_processes: 24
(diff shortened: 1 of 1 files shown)


sravfeyn:
Ha, ok. I was following [this](https://commcare-hq.readthedocs.io/pillows.html#parallel-processors) per Emord's suggestion on the ticket. 

If this config is not updated, the why do we need `update_supervisor_confs`?

emord:
> If this config is not updated, then why do we need update_supervisor_confs?

We don't need to update the supervisor configuration at this time. I think adding the partitions was just to ensure that we are able to do this step in the future

PR #3328, opened 2019-10-17
Adding new RabbitMQ server in celery broker URL on ICDS env

sanjay2916:
Adding new RabbitMQ server in celery broker URL on ICDS env

### DIFF
--- a/src/commcare_cloud/ansible/group_vars/all.yml
+++ b/src/commcare_cloud/ansible/group_vars/all.yml
@@ -124,7 +124,6 @@ check_s3_backups_email: null
 AMAZON_S3_ACCESS_KEY: "{{ localsettings_private.AMAZON_S3_ACCESS_KEY | default(None) }}"
 AMAZON_S3_SECRET_KEY: "{{ localsettings_private.AMAZON_S3_SECRET_KEY | default(None) }}"
 AMQP_HOST: "{{ groups.rabbitmq.0 }}"
-AMQP_HOST1: "{{ groups.rabbitmq.1 }}"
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -124,6 +124,7 @@ check_s3_backups_email: null
 AMAZON_S3_ACCESS_KEY: "{{ localsettings_private.AMAZON_S3_ACCESS_KEY | default(None) }}"
 AMAZON_S3_SECRET_KEY: "{{ localsettings_private.AMAZON_S3_SECRET_KEY | default(None) }}"
 AMQP_HOST: "{{ groups.rabbitmq.0 }}"
+AMQP_HOST1: "{{ groups.rabbitmq.1 }}"
```
snopoke: won't this cause an error if there isn't a 2nd host in the group?

snopoke: Since this is currently ICDS specific I don't think it needs to go here, you could just put it in `icds-cas/public.yml`

sanjay2916: > Since this is currently ICDS specific I don't think it needs to go here, you could just put it in `icds-cas/public.yml`

I have updated in public.yml

sanjay2916: > won't this cause an error if there isn't a 2nd host in the group?

in case if we do not have the second host in inventory group it throwing an error while updating config

"AnsibleUndefinedVariable: list object has no element 1"}

PR #3332, opened 2019-10-17
Downsize es{0,1}-india

dannyroberts:
##### SUMMARY
It's underutilizing RAM and CPU

Roll out one at a time: 

```bash
node=es0  # or es1
cchq india terraform apply --skip-secrets -target module.server__${node}-india.aws_instance.server
cchq india deploy-stack --tags=es_conf ${node}  # to change the RAM settings
cchq india after-reboot ${node}
```

##### ENVIRONMENTS AFFECTED
india


emord:
> It's underutilizing RAM

Is this true even if you consider that databases use the OS's page caching system to take advantage of "free" RAM?

emord:
Also india already has some memory settings hard set https://github.com/dimagi/commcare-cloud/blob/55fa0571aac410ed2f666542edf58b07b0c1ce86/environments/india/public.yml#L8

What is the following going to do?

> cchq india deploy-stack --tags=es_conf ${node}  # to change the RAM settings

dannyroberts:
Thanks for pointing out the hardcoded ES memory. I think I just want to remove that and let it go back to it's default of 1/2 the total (up to ~30g). Do you know of any issues with that? You're totally right this wouldn't have worked well, so thanks for catching that!

Re
> Is this true even if you consider that databases use the OS's page caching system to take advantage of "free" RAM?

If you include caching it sometimes goes up to 100% but hovers a little lower than that most of the time. My guess is that ES doesn't get a ton of usage on india and would do fine even with less RAM (even though it's happy to use all most of it if given the chance).

For context this is going from 2 x 16G RAM to 2 x 8G RAM.

dannyroberts:
Rolled this out

PR #3335, opened 2019-10-18
Change staging RDS instance class, size

dannyroberts:
##### SUMMARY
and engine version (to accommodate changing instance class from t2 to t3).

I already rolled this out through the AWS console, so this change is to just to bring our config into alignment

##### ENVIRONMENTS AFFECTED
staging

### DIFF
--- a/environments/staging/terraform.yml
+++ b/environments/staging/terraform.yml
@@ -173,9 +173,10 @@ proxy_servers:
 
 rds_instances:
   - identifier: "pg0-staging"
-    instance_type: "db.t2.medium"
+    instance_type: "db.t3.small"
     storage: 160
     multi_az: yes
+    engine_version: 9.6.9
 
 elasticache:
   create: no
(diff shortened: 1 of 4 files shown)


environments/staging/terraform.yml
```diff
@@ -173,9 +173,10 @@ proxy_servers:
 
 rds_instances:
   - identifier: "pg0-staging"
-    instance_type: "db.t2.medium"
+    instance_type: "db.t3.small"
     storage: 160
     multi_az: yes
+    engine_version: 9.6.9
```
emord: does `engine_version` == postgres version?

dannyroberts: Yeah, I adopted terraform's naming convention (which is maybe also AWS's?). It would be simple to change this to postgres version if you think that'd be better

emord: I don't have a preference on it especially since its a convention. It was fairly easy to realize what it should mean (I suppose its nice to have a generic name for "X_version" if we also run other managed services)

PR #3338, opened 2019-10-18
Disable groovy script execution on icds

rohit-dimagi:
##### SUMMARY

* Disable groovy script execution on icds

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Change Pull Request


rohit-dimagi:
@sanjay2916  Can you also apply this one too?

snopoke:
@dannyroberts FYI we are disabling groovy scripting in ES on ICDS for security reasons. It is currently used by they "Form Completion Time" report and 2 sets of custom reports: `pact` and `m4change` (none of which are used on ICDS).

The data for "Form Completion Time" could trivially be computed in the xform pillow prior to sending to ES in case SAAS is interested in disabling it as well.

PR #3341, opened 2019-10-19
Change kafka log retention to 21 days

emord:
##### SUMMARY
@ctsims changed this to 21 days temporarily during firefighting. (personally I think we should go even lower, but can open that for discussion)

I believe that this can be done without downtime by applying to one kafka node at a time since we have replication

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
kafka


emord:
Deployed during the downtime today. I did do this one node at a time and there did not appear to be any downtime (aside from the other services).

PR #3347, opened 2019-10-22
Terminating trusty pgproxy1 server on prod env

sanjay2916:
Terminating pgproxy1 on prod env


sanjay2916:
pgproxy1 removed from #3348

PR #3349, opened 2019-10-23
Postgres Backup configuration on standby

rohit-dimagi:
##### SUMMARY

* Postgres wal archiving destination changed to standby servers.

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Change Pull Request


src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
```diff
@@ -46,7 +46,11 @@ effective_io_concurrency = {{ postgresql_effective_io_concurrency }}
 # WRITE AHEAD LOG
 {% if postgresql_enable_archive and not is_pg_standby %}
 archive_mode = on
+{% if hot_standby_server %}
+archive_command = 'scp %p {{hot_standby_server}}:{{ postgresql_archive_path }}/%f'
```
snopoke: this is going to require no-password auth between the two servers. Are you going to set that up separately?

rohit-dimagi: It's already part of the setup process 
https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/deploy_postgres.yml#L60

environments/icds-cas/inventory/citusdb.csv
```diff
@@ -60,3 +60,29 @@ citusdb_worker,postgresql_random_page_cost,1.1,F,,,,,,
 citusdb_worker,postgresql_shared_buffers,20GB,S,,,,,,
 citusdb_worker,postgresql_wal_buffers,16MB,S,,,,,,
 citusdb_worker,postgresql_work_mem,1872kB,S,,,,,,
+citusmaster1,postgresql_enable_archive,True,B,,,,,,
```
snopoke: can you not just do this for the whole `citusdb` group?

src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
```diff
@@ -46,7 +46,11 @@ effective_io_concurrency = {{ postgresql_effective_io_concurrency }}
 # WRITE AHEAD LOG
 {% if postgresql_enable_archive and not is_pg_standby %}
 archive_mode = on
+{% if hot_standby_server %}
+archive_command = 'scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null %p {{hot_standby_server}}:{{ postgresql_archive_path }}/%f'
```
snopoke: why not just override `postgresql_archive_command` like you did with `postgresql_archive_path` instead of create a new branch here?

src/commcare_cloud/ansible/roles/ecryptfs/tasks/wal.yml
```diff
@@ -0,0 +1,66 @@
+---
+- name: Check for already mounted encrypted wal drive
```
snopoke: instead of copying these can't we put them in a separate tasks file that we can include and use a var to tell it which drive to use:

```
include: encrypt_dir.yml
vars:
    path: "{{ ecrypted_wal_archive }}"
```
(not sure if that's the right was to pass vars)

src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
rohit-dimagi: can't override in `<env>public.yml` or in playbook (it will throw undefined var hot_standby_server) . so keeping it here.

rohit-dimagi:
Reorganize some vars declaration for better control over it. it requires splitting citusdb group to split into two mutually exclusive group for standby and master servers like we have for `postgresql.csv`

PR #3353, opened 2019-10-23
Set STATIC_DATA_SOURCE_PROVIDERS to empty for ICDS

emord:
##### SUMMARY
Companion to https://github.com/dimagi/commcare-hq/pull/25683

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
localsettings

### DIFF
--- a/environments/icds-cas/public.yml
+++ b/environments/icds-cas/public.yml
@@ -229,6 +229,7 @@ localsettings:
   REMINDERS_QUEUE_ENABLED: True
   SMS_GATEWAY_URL:
   SMS_QUEUE_ENABLED: True
+  STATIC_DATA_SOURCE_PROVIDERS: []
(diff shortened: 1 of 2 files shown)


emord:
Thanks for asking. I definitely should have included that in the description.

ICDS does use the call center feature. The data source provider filters out [domains that do not have use fixtures](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/callcenter/data_source.py#L25) (side note: if that filter was moved to the ES query, maybe it would have performed better?)

`icds-cas` on the main server has that setting set to false https://www.icds-cas.gov.in/hq/admin/raw_doc/?id=642b7fde7232c0c75b2bf56b560c0139

PR #3355, opened 2019-10-25
statsd on on airflow machines

sravfeyn:
@kaapstorm


sravfeyn:
> I don't have context on this, but it looks OK. 🤷‍♂ I guess Airflow can use data that is aggregated using StatsD?

Yes, Airflow as a client sends metric data to a StatsD server, in our case datadog has a DogstatsD server that listens to this data and sends it to Datadog

PR #3365, opened 2019-10-29
Replacing Kafka and control servers with Bionic servers

sanjay2916:
Replacing Kafka and control servers with Bionic servers


dannyroberts:
@dimagi/team-commcare-hq Heads up, team! We now have a new production control machine, so next time you log in the commcare-cloud setup process will take longer than usual, as it'll have to do a fresh install of dependencies

PR #3369, opened 2019-11-04
Add confluence environment

dannyroberts:
##### SUMMARY
I know this isn't a commcare environment, but this is the simplest I can think of for the saas team to keep confluence under config management. Using this to add datadog and swap


dannyroberts:
> Did you add documentation for this anywhere? The confluence entry (with instructions for rebooting) in keepass comes to mind as a possible place to mention it.

That's a great idea, thanks. I'll add a note about it to that keepass entry.

PR #3378, opened 2019-11-07
updated the limit

Rohit25negi:
Updating the max connection limit of workers until the pgbouncer is set up there.


emord:
This caused issues when deployed. I've asked for a retro in https://dimagi-dev.atlassian.net/browse/IIO-566

PR #3386, opened 2019-11-13
Updating firefighting doc to handle WAL deletion

rohit-dimagi:
##### SUMMARY

* Updating firefighting doc to handle WAL deletion

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/docs/firefighting/index.md
+++ b/docs/firefighting/index.md
@@ -543,6 +543,32 @@ GROUP BY main_table
 ORDER BY n_tup_ins DESC;
 ```
 
+
+### Deleting old WAL logs
+At all the times, PostgreSQL maintains a write-ahead log (WAL) in the pg_xlog/ subdirectory for (version >=10 , pg_wal folder) of the cluster’s data directory. The log records for every change made to the database’s data files. These log messages exists primarily for crash-safety purposes.
+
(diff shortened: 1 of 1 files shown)


docs/firefighting/index.md
```diff
@@ -543,6 +543,32 @@ GROUP BY main_table
 ORDER BY n_tup_ins DESC;
 ```
 
+
+### Deleting old WAL logs
+At all the times, PostgreSQL maintains a write-ahead log (WAL) in the pg_xlog/ subdirectory for (version >=10 , pg_wal folder) of the cluster’s data directory. The log records for every change made to the database’s data files. These log messages exists primarily for crash-safety purposes.
```
emord: I think that we only have postgres >= 10  for the Citus clusters. I believe all others are still 9.6 right? Its a bit confusing that this sentence mentions `pg_xlog` first with `pg_wal` in parens, but the example commands below only use pg 11 and `pg_wal` (which will only be useful for ICDS)

rohit-dimagi: i have tried to make it more general over here. https://github.com/dimagi/commcare-cloud/pull/3395

PR #3390, opened 2019-11-13
Release batch app status pillow on ICDS

emord:
##### SUMMARY
Will release the batched app status pillow on ICDS. And sets the time to update couch documents between 1900 - 2200 UTC. (0030 - 0330 IST)

This is as scheduled in rollout section of https://github.com/dimagi/commcare-hq/pull/25726

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
pillows/couchdb

##### ADDITIONAL INFORMATION
Live bug: Currently all system forms will create a `PillowError` record as noted here https://github.com/dimagi/commcare-hq/pull/25923. That's a pretty small number of forms and we don't generally care about them, so I believe its ok to run with the errors until the next deploy.

First glance: Anecdotally, I noticed lower load on couch while rolling this out (I did 1 pilllow -> 1 server -> rest of servers), but there could have been other factors involved here so I am not sure how well it will translate to the middle of the day.

Factors to look for a rollback:
If pgmain0 has too high of load it may be writing too much to the table. I didn't notice any issues tonight as I was monitoring, but its something to watch.

If there is an increase in pillow errors that are not related to the one mentioned above (error_type: django.db.utils.IntegrityError) that appears to be related to these changes

To rollback:

while on master branch
```
cchq icds update-config
cchq icds service pillowtop restart --only xform-pillow
cchq icds service pillowtop restart --only UpdateUserSyncHistoryPillow
```


emord:
This didn't quite get through all the changes before EOD and I manually revoked the celery tasks to ensure it didn't affect performance during peak loads.

I've expanded the time window and made the tasks more frequent

PR #3394, opened 2019-11-13
Make django-manage use manage.py's return code as its own

dannyroberts:
##### SUMMARY
Thanks @millerdev for noticing! Couldn't tell from the history if it ever worked, but anyway this makes it do what I always intended it to and pass on the status code.

##### ISSUE TYPE
- Bugfix Pull Request


orangejenny:
thank you, I also ran into this yesterday

PR #3396, opened 2019-11-14
log using syslog

snopoke:
##### SUMMARY
Logging to a single file from Django doesn't really work well as described here: https://www.packetmischief.ca/2017/10/25/3-ways-to-fail-at-logging-with-flask/

This PR makes use of the Syslog handler to send logs from Django to syslog. It includes changes to route those logs to the desired log file as well as do log rotation.

##### ENVIRONMENTS AFFECTED
ALL

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
logging

##### ADDITIONAL INFORMATION
Global changes:
* make logrotate run hourly on ALL servers
* add `syslog` user to the `cchq` group and change the mode on the CCHQ log folders to allow syslog to write to it


snopoke:
The config for this could be simplified quite dramatically if we do some pre-processing in python:
```
commcare_loggers:
    kafka_producer_audit:
        level: DEBUG
        rotate:
            - hourly
            - rotate 5
```

PR #3380, opened 2019-11-08
Fixing wal archiving cleanup cron Job

rohit-dimagi:
##### SUMMARY

* Fixing wal archiving cleanup cron Job

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Bugfix Pull Request

PR #3405, opened 2019-11-18
Set archive_command to /bin/true for citusdb

emord:
##### SUMMARY
@rohit-dimagi I saw that the archive_command is `/bin/true` on all citus workers. Is this expected? I couldn't find a PR that included this change

##### ENVIRONMENTS AFFECTED
icds

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
citusdb

##### ADDITIONAL INFORMATION
A deploy currently looks like this:

```
--- before: /etc/postgresql/11/main/postgresql.conf
+++ after: /home/jemord/.ansible/tmp/ansible-local-581vMquJM/tmpHuNyIp/postgresql.conf.j2
@@ -35,7 +35,7 @@
 
 # WRITE AHEAD LOG
 archive_mode = on
-archive_command = '/bin/true'
+archive_command = 'cp %p /opt/data/postgresql/wal_archive/%f'
 archive_timeout = 300
 
 wal_level = hot_standby
```

### DIFF
--- a/environments/icds-cas/inventory/citusdb.csv
+++ b/environments/icds-cas/inventory/citusdb.csv
@@ -29,6 +29,7 @@ citusworker23standby,100.71.184.246,citusdb,citusdb_worker,pg_standby,MUMGCCWCDP
 group,var,val,type,,,,,,
 citusdb,ansible_python_interpreter,/usr/bin/python3,S,,,,,,
 citusdb,pgbouncer_max_connections,3500,I,,,,,,
+citusdb,postgresql_archive_command,/bin/true,S,,,,,,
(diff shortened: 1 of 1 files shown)


snopoke:
The change was made manually when we had the issue with archive logs failing. It was supposed to be temporary. @rohit-dimagi can you update it on the other DBs that got changed as well.

rohit-dimagi:
citusworker14,citusmaster1 still has old command on it's config. it's probably easier to remove archiving on all machine since we are not using it anyway and have all db machines state consistent. 
We can apply it during downtime we are planning today https://github.com/dimagi/commcare-cloud/pull/3380

PR #3408, opened 2019-11-19
Only tag deploy commits for Dimagi-controlled environments

esoergel:
##### SUMMARY
Currently, we attempt to tag the commits for all deploys.  This requires credentials with write access to the commcare-hq repo.  If the user hasn't copied over fab/config.yml, they'll be nudged to do so and prompted for a username and password, though they can leave this blank.  To silence the prompt, the user needs to copy that config and set their API key to use auth, or leave it as `None` to not tag their deploys.  In this latter case, they still get a warning that release tags are not being created.

None of this makes sense for non-Dimagi users, since they'll never be able to create tags.  This PR is my proposal for addressing that.  I'm introducing a flag to enable tagging deploy commits for specific environments - those controlled by Dimagi.  For these environments, I've attempted to preserve the existing behavior.  For other environments, we won't even look for the credentials in the first place, and we won't warn the user about them missing.

Notably, I did not enable these flags for echis or pna, since those envs are primarily outside our control.  I _did_ enable it for swiss, since Dimagi deploys it regularly.  LMK if you have feedback on those decisions.

A possible extension of this work would be to require credentials to deploy the envs that request deploy tags.  That would be nice, since it'd make the record a bit more reliable.  I'm considering that out of scope for this change, but would be open to submitting a subsequent PR to do that, or letting someone else run with it if they like the idea.

@dannyroberts FYI

##### ENVIRONMENTS AFFECTED
All, sorta

##### ISSUE TYPE

- Change Pull Request

 

##### COMPONENT NAME
Deploy tags

### DIFF
--- a/src/commcare_cloud/fab/utils.py
+++ b/src/commcare_cloud/fab/utils.py
@@ -163,6 +163,8 @@ class DeployMetadata(object):
 
 @memoized
 def _get_github_credentials():
+    if not env.tag_deploy_commits:
+        return (None, None)
     try:
         from .config import GITHUB_APIKEY
     except ImportError:
@@ -181,7 +183,7 @@ def _get_github_credentials():
 @memoized
 def _get_github():
     login_or_token, password = _get_github_credentials()
-    if not login_or_token:
+    if env.tag_deploy_commits and not login_or_token:
         print(magenta(
             "Warning: Creation of release tags is disabled. "
             "Provide Github auth details to enable release tags."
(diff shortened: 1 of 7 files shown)


src/commcare_cloud/fab/utils.py
```diff
@@ -177,30 +171,27 @@ def _show_no_auth_message():
...
 
 @memoized
 def _github_auth_provided():
-    try:
-        from .config import GITHUB_APIKEY
-    except ImportError:
-        return True  # user is prompted for auth
-    else:
-        return GITHUB_APIKEY is not None
+    return bool(_get_github_credentials()[0])
```
esoergel: I wanted to use the `Github` object itself here, but couldn't find anything handy describing whether credentials were provided without reaching into its internals.

PR #3415, opened 2019-11-21
Update redis conf to deployed settings

sravfeyn:
These changed were deployed a while ago but are not sticky since the config file wasn't updated.

### DIFF
--- a/environments/icds-cas/public.yml
+++ b/environments/icds-cas/public.yml
@@ -27,7 +27,8 @@ swift_url: https://mowcdmum.ipstorage.tatacommunications.com/auth/v1.0/
 
 nofile_limit: 65536
 
-redis_maxmemory: 6gb
+redis_maxmemory: 8gb
+redis_maxclients: 14000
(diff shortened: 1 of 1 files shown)


environments/icds-cas/public.yml
```diff
@@ -27,7 +27,8 @@ swift_url: https://mowcdmum.ipstorage.tatacommunications.com/auth/v1.0/
 
 nofile_limit: 65536
 
-redis_maxmemory: 6gb
+redis_maxmemory: 8gb
```
snopoke: I'm concerned that this is going to bite us in the future. See https://github.com/dimagi/commcare-cloud/pull/2929#issuecomment-499970466

sravfeyn: I think we have 16GB machines so, 8 should be fine even with that, since it's half? Also, I will look a bit more to see if we can avoid the rewrite to happen using peak usage so that we can get away with limitation.

PR #3423, opened 2019-11-25
fix pgbouncer config on citusworkers

snopoke:
##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Bugfix Pull Request


src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
```diff
@@ -13,6 +13,13 @@
 {{ db.name }} = host={{ "localhost" if db.host == db.pgbouncer_host else db.host }}
 {% endfor %}
 
+{% if current_host in groups.citusdb_worker %}
+{% set citus_master = (groups.citusdb_master | difference(groups.pg_standby))[0] %}
```
emord: @snopoke how did you deploy this? I'm getting an error every time I try to that looks like `AnsibleUndefinedVariable: Unable to look up a name or access an attribute in template string ....Make sure your variable name does not contain invalid characters like '-': argument of type 'AnsibleUndefined' is not iterable`

The result of however you deployed it was that the `databases` section on all the workers was left blank meaning the master couldn't connect to the workers and any distributed queries failed. I verified that the fix was to add `icds-ucr-citus = host=localhost` to this section, which I believe is what this code was intended to do.

snopoke: I deployed this with `cchq icds ap deploy_citusdb.yml --tags pgbouncer --limit citus_workers` which was basically all hosts with the var defined.

Fixed here: https://github.com/dimagi/commcare-cloud/pull/3429

snopoke: I think I fixed a different issue. I see the error you describe when running on `india`

snopoke: Update that PR with a fix

PR #3424, opened 2019-11-25
Revert "Adding new celery + webworkers"

emord:
Reverts dimagi/commcare-cloud#3411

These were not fully deployed. They did not have a virtualenv created which caused `virtualenv-clone` to fail during deploy


snopoke:
@rohit-dimagi did you do a full stack deploy?

PR #3440, opened 2019-12-03
Rk/web+celery

rohit-dimagi:
##### SUMMARY

* Adding New celery + webworkers

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Change Pull Request


src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml
```diff
@@ -11,6 +11,13 @@
 - import_tasks: www_log_dir.yml
   tags: logging
 
+- name: set correct permission on /home/cchq
+  become: yes
+  file:
+    path: "{{ cchq_home }}"
+    owner: cchq
+    group: cchq
+
```
rohit-dimagi: @snopoke  The ownership was set to root on these nodes.  I don't how it was working on past.

snopoke: How were they set up? I think this could be possible if the user had not been created prior to running this role. 

I think a better thing to add would be an assertion at the top of the file that the `{{ cchq_user}}` exists

rohit-dimagi: First i ran this 
`cchq icds deploy-stack --first-time --limit <node list> --branch= <> -u wcdadmin`
and then this .
`cchq icds deploy-stack  --limit <node list> --branch= <> `

snopoke: I can't find anything that would cause this - fine to leave this in for now until we can figure it out

PR #3444, opened 2019-12-05
Move toggles from HQ code to localsettings config

sravfeyn:
Relevant HQ PR https://github.com/dimagi/commcare-hq/pull/26120

I am migrating toggle states as defined in HQ to respective env's public files. I am doing this based on the best on which domain might be relevant on which envs. Below is the summary of changes

- all ICDS domain toggle states to ICDS and india env.
- all non ICDS domain toggle states to prod env
- eCHIS toggle state to eCHIS env and prod (But I think it's only relevant to prod as it is a testing domain)


environments/echis/public.yml
```diff
@@ -81,3 +81,7 @@ localsettings:
   ENABLE_PRELOGIN_SITE: False
   ENTERPRISE_MODE: True
   RESTRICT_DOMAIN_CREATION: True
+  STATIC_TOGGLE_STATES:
+    mobile_ucr_linked_domain:
+      always_enabled:
+        - 'fmoh-echis-staging'
```
sravfeyn: Hey Ethan, just FYI, we are moving toggle state declaration from HQ code to env config. This is the only eCHIS toggle I found, I think this domain exists only on Prod but not eCHIS env, but if it is, then they need to do an update-config with this change

@esoergel

sravfeyn:
This is deployed on ICDS/India/Prod now

PR #3445, opened 2019-12-05
remove unused variables from localsettings

czue:
Think these have been unnecessary since MVP was removed.

Just happened to see them in a slack chat.


biyeun:
YESSS REMOVE THEM ⚡️

PR #3451, opened 2019-12-05
Revert "Add more RAM to django_manage VM"

millerdev:
Undo #3368


millerdev:
Applied a few minutes ago.

PR #3452, opened 2019-12-05
Route by new formplayer_session cookie instead of sessionid cookie

dannyroberts:
#### SUMMARY
Currently when a user logs out and logs back in, they are routed
randomly to one of the formplayer machines. This change gives the caller
the freedom to route based on whatever they like, and we will use that
in commcare-hq to route based on user_id instead of django sessionid.

The caller can now route by setting either the X-FORMPLAYER-SESSION
header or the formplayer_session cookie. A blemish of this
implementation is that if they sometimes but not always send _both_ the
header and the cookie, that will break sticky routing.

Note that this is not safe to roll out until https://github.com/dimagi/commcare-hq/pull/26127 is rolled out. If that happened the behavior would either be to break sticky routing entirely or to always route to the same machine, I'm not sure.

##### ENVIRONMENTS AFFECTED
production for sure, but any env with multiple formplayer machines

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
Formplayer (Web Apps / App Preview)

##### ADDITIONAL INFORMATION
On prod the relevant part of the diff would be:

```diff
 upstream production_formplayer {
   zone production_formplayer 64k;

-    hash $cookie_sessionid$http_x_formplayer_session consistent;
+    hash $cookie_formplayer_session$http_x_formplayer_session consistent;
```

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -11,9 +11,11 @@ nginx_sites:
     - name: "{{ deploy_env }}_formplayer"
       hosts: formplayer
       port: "{{ formplayer_port }}"
-      # hash off sessionid cookie or X-FORMPLAYER-SESSION header
-      # but only one of the two will be present in any given request
-      method: "hash $cookie_sessionid$http_x_formplayer_session consistent"
+      # hash off formplayer_session cookie or X-FORMPLAYER-SESSION header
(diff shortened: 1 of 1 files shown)


dannyroberts:
FWIW The new cookie is named `formplayer_session` (with an underscore, not a dash). (The _/- thing affects $http_x_formplayer_session / X-FORMPLAYER-SESSION header, but that's already been rolled out and tested, no change there.)

dannyroberts:
I rolled this out and tested that it works as intended.

PR #3453, opened 2019-12-05
Fix direct fab calls without `--set code_branch=...`

dannyroberts:
##### SUMMARY
When I moved deploy into its own commcare-cloud command, I neglected
to retain the behavior of the many other fab calls we use regularly
but less frequently (setup_limited_release, etc.), and they would fail
unless you added e.g. `--set code_branch=master` to them.
This brings back earlier behavior of choosing the default branch
specified in fab_settings.yml

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -168,9 +168,15 @@ def load_env():
 def _setup_env(env_name):
     env.env_name = env_name
     load_env()
+    _confirm_branch(env.default_branch)
     _confirm_environment_time(env_name)
     execute(env_common)
     execute(_confirm_deploying_same_code)
+
+
+def _confirm_branch(default_branch):
+    if not getattr(env, 'code_branch', None):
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/fab/fabfile.py
```diff
@@ -168,11 +168,18 @@ def load_env():
 def _setup_env(env_name):
     env.env_name = env_name
     load_env()
+    _confirm_branch(env.default_branch)
     _confirm_environment_time(env_name)
     execute(env_common)
     execute(_confirm_deploying_same_code)
 
 
+def _confirm_branch(default_branch):
```
millerdev: nit: the name of this function says `confirm`, but it simply sets the attribute and prints a message. `_confirm_environment_time` prompts for a response (consistent with the conventional meaning of _confirm_).
```suggestion
def _set_code_branch(default_branch):
```

millerdev:
Thanks for the quick fix!

PR #3457, opened 2019-12-09
run django checks after updating config

snopoke:
##### SUMMARY
CommCare HQ has a number of custom Django checks which help to validate settings and configuration. This adds an option to running these after updating localsettings.

##### ISSUE TYPE
- Change Pull Request

### DIFF
--- a/src/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/src/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -222,8 +222,23 @@ class UpdateConfig(CommandBase):
 
     def run(self, args, unknown_args):
         unknown_args += ('-e', '{"_should_update_formplayer_in_place": true}')
-        return commcare_cloud(args.env_name, 'ansible-playbook', 'deploy_localsettings.yml',
+        rc = commcare_cloud(args.env_name, 'ansible-playbook', 'deploy_localsettings.yml',
                               tags='localsettings', branch=args.branch, *unknown_args)
+        if rc == 0 and ask("Would you like to run Django checks to validate the settings?"):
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -222,8 +222,23 @@ class UpdateConfig(CommandBase):
 
     def run(self, args, unknown_args):
         unknown_args += ('-e', '{"_should_update_formplayer_in_place": true}')
-        return commcare_cloud(args.env_name, 'ansible-playbook', 'deploy_localsettings.yml',
+        rc = commcare_cloud(args.env_name, 'ansible-playbook', 'deploy_localsettings.yml',
                               tags='localsettings', branch=args.branch, *unknown_args)
+        if rc == 0 and ask("Would you like to run Django checks to validate the settings?"):
```
emord: how long do these take? would it be worthwhile to just always run the checks?

src/commcare_cloud/commands/inventory_lookup/getinventory.py
```diff
@@ -29,6 +31,10 @@ def get_server_address(environment, group, exit=sys.exit):
         username += "@"
     else:
         username = ""
+
+    if re.match(r'(\d+\.?){4}', group):
```
emord: Is this for an ip address? may be worth a comment for this (or some constant with reasonable name)

src/commcare_cloud/commands/ansible/ansible_playbook.py
snopoke: `check_services` is the longest - that can take up to 30s.

PR #3459, opened 2019-12-10
set pillows to read from standbys

snopoke:
##### SUMMARY
Add env var to pillowtop processes to allow them to read from plproxy shard standbys if they are configured and available.

Not making this optional since it will only have effect if the proxy standbys are configured in localsettings.

##### ENVIRONMENTS AFFECTED
ALL (though only practically ICDS)

##### ISSUE TYPE
- Change Pull Request


snopoke:
rolled out to ICDS

PR #3460, opened 2019-12-10
set ownership and mode for supervisor files

snopoke:


rohit-dimagi:
@snopoke  This will get permission denied error because of `become: cchq`

PR #3468, opened 2019-12-12
Increase case-sql processes

sravfeyn:
We added kafka partitions a while ago https://dimagi-dev.atlassian.net/browse/IIO-515

@snopoke


snopoke:
Do these VMs have enough RAM / CPU to double the number of processes?

Also I think we need to increase form processors as well. There should be some nodes that aren't currently in use.

sravfeyn:
(icds) ansible@kafka5:/opt/kafka/bin$ ./kafka-topics.sh --describe --zookeeper=localhost:2181 --topic form-sql
Topic:form-sql	PartitionCount:96	ReplicationFactor:2	Configs:retention.ms=1814400000

PR #3470, opened 2019-12-12
fix permission issue

rohit-dimagi:


snopoke:
I think @sravfeyn already fixed these

PR #3473, opened 2019-12-13
Upgrading Elasticsearch to 2.4.6 version

sanjay2916:
Upgrading Elasticsearch to 2.4.6 version


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -428,6 +428,7 @@ ELASTICSEARCH_HOSTS = [
     {% endfor -%}
 ]
 ELASTICSEARCH_PORT = {{ ELASTICSEARCH_PORT }}
+ELASTICSEARCH_MAJOR_VERSION = 2
```
dannyroberts: Can you write this in a way that will work for environments both on ES 1 and ES 2? Currently if I merge this it will make all environments think they're running against ES 2.

PR #3480, opened 2019-12-18
send citus reads to standbys

snopoke:
##### SUMMARY
The original approach of using connection params to tell Citus to read from standbys only for certain connections as done in https://github.com/dimagi/commcare-hq/pull/26143 does not work with pgbouncer since it 'eats' the `options` start param (https://github.com/dimagi/commcare-cloud/pull/3425). In hind sight this is obvious since pgbouncer is multiplexing connections at the transaction level.

As an alternative we can make the Citus master standby node always route to the standbys and send read request to it instead of to the primary node.

With https://github.com/dimagi/commcare-hq/pull/26253 and https://github.com/dimagi/commcare-hq/pull/26254

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
CitusDB

##### ADDITIONAL INFORMATION
I also tested out using pgpool II instead of pgbouncer since it maintains the start parameters but the problem with this approach is the large number of DB connections that are required (~1500). This poses a problem as the scale increases.


environments/icds-cas/postgresql.yml
```diff
@@ -53,6 +57,10 @@ postgres_override:
     - contype: host
       users: TCLBKPADMIN
       netmask: 100.71.188.129/32
+    - contype: host
+      databases: replication
+      users: TCLBKPADMIN
+      netmask: 100.71.188.129/32
```
snopoke: This was missing when I rolled out the PG changes @rohit-dimagi is this required or just left over from some testing?

snopoke:
Deployed to ICDS

PR #3483, opened 2019-12-19
Migrating proxy servers to vmware

rohit-dimagi:
##### SUMMARY

* Migrating proxy servers to vmware

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Change Pull Request


src/commcare_cloud/ansible/roles/nginx/tasks/install.yml
```diff
@@ -28,6 +28,13 @@
     line: 'ULIMIT="-n 65536"'
   ignore_errors: "{{ ansible_check_mode }}"
 
+- name: Enable lpv6 on loopback interface
```
snopoke: why is this needed?

calellowitz:
Restarting services on proxy3, specific nginx, is failing, which is blocking deploy. It fails with `nginx: [error] invalid PID number "" in "/run/nginx.pid"` I have reverted this PR to allow deploy to proceed, but nginx is broken on that machine. I am not sure what the correct state is (whether we should be using that proxy or not), but this will need to be fixed before this PR can be merged again. 
Ticket relevant to this change appears to be https://dimagi-dev.atlassian.net/browse/IIO-651 but there is no information there either.

rohit-dimagi:
I would except deploy to be successful if `deploy-stack` is completed without any problem. i see nginx is running fine on that machine. Do you have any more context on the error ?

snopoke:
@rohit-dimagi I think you should be able to get more context on the error from the machine itself

rohit-dimagi:
Checked on server. didn't find anything unusual there.

rohit-dimagi:
Discussed with @calellowitz the task which failed was not reproduced so i'm going to merge this as is.

PR #3485, opened 2019-12-19
run tmux as logged in user instead of cchq

snopoke:
##### SUMMARY
By running tmux as the logged in user it solves a few problems:
* makes it easy to see who's tmux is running by looking at the process owner
* prevents joining another users tmux session when running a command


snopoke:
@dannyroberts bump

PR #3510, opened 2020-01-09
Add jenkins env

dannyroberts:
##### SUMMARY
This gives us a simple way to keep the users on the jenkins machine up to date. Also adding Shivam's public key.


environments/jenkins/inventory.ini
```diff
@@ -0,0 +1,24 @@
+[jenkins_server]
+jenkins.dimagi.com hostname=jenkins-server swap_size=1G
+
+[jenkins_mobile_agent_2]
+162.242.218.23 hostname=jenkins-mobile-agent-2
```
shubham1g5: Does defining a `swap_size` here adds a swap partition to the machine ? If so, even though not in scope of this work, can we add a similar `swap_size=1G` here.

dannyroberts: Yup, defining swap_size here and running a particular command will set the swap (though as a side note I think current it's not good at _changing_ the swap size to a new value once set). Happy to add swap to the machine. I'll do that in another PR

dannyroberts: https://github.com/dimagi/commcare-cloud/pull/3514

PR #3515, opened 2020-01-13
Setting master-slave replication and sentinel

rohit-dimagi:
##### SUMMARY

* Setting master-slave replication and sentinel with keepalived

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Change Pull Request


src/commcare_cloud/ansible/deploy_db.yml
```diff
@@ -20,23 +20,14 @@
 - import_playbook: deploy_couchdb2.yml
 
 - name: Redis
-  hosts: redis
-  become: true
-  roles:
-    - {role: redis, tags: 'redis'}
+  import_playbook: deploy_redis.yml
```
snopoke: :heart:

src/commcare_cloud/ansible/roles/keepalived/tasks/main.yml
```diff
@@ -15,6 +15,17 @@
   tags: keepalived
   notify: restart keepalived
 
+- name: Create redis master check script
```
snopoke: shouldn't this also be in the redis role?

rohit-dimagi:
Generic keepalived role and now all the relevant tasks are moved to couchdb and redis deploy role.

src/commcare_cloud/ansible/group_vars/redis.yml
```diff
@@ -13,7 +13,3 @@ redis_maxmemory_policy: volatile-lru
 redis_save: []
 redis_appendonly: "yes"
 redis_timeout: 300  # 5 minutes
-cluster_config:
```
snopoke: see below comment - I don't think this should be removed in this PR

src/commcare_cloud/ansible/roles/keepalived/tasks/main.yml
rohit-dimagi: Moved.

src/commcare_cloud/ansible/roles/redis/tasks/setup_cluster.yml
```diff
@@ -3,13 +3,5 @@
   command: 'gem install redis -v 3.3.3' # not the latest, because https://github.com/antirez/redis/issues/4272
   become: yes
 
-- name: Add cluster config
```
snopoke: I don't think this should be removed. Rather we should change the condition in `redis/tasks/main.yml`:
```
- include: setup_cluster.yml
  when: is_redis_cluster|bool and inventory_hostname in groups['redis_cluster_master']
```

src/commcare_cloud/ansible/roles/keepalived/tasks/main.yml
rohit-dimagi: Can't notify from playbook. where should i keep this ?

snopoke: Why does it need to notify for changes to this script?

rohit-dimagi: It's Needed for first run 
```
Jan 22 05:06:59 MUMGCCWCDPRDRDV05 Keepalived_vrrp[121144]: Script /etc/keepalived/redis_is_master cannot be accessed - No such file or directory
Jan 22 05:06:59 MUMGCCWCDPRDRDV05 Keepalived_vrrp[121144]: Disabling track script chk_service since not found/accessible
```
if it doesn't find the file at start time. it will disable tracking of service.  So we need to restart at the time of creating this file.

src/commcare_cloud/ansible/deploy_redis.yml
```diff
@@ -0,0 +1,63 @@
+- name: Redis
+  hosts: redis
+  become: true
+  roles:
+    - {role: redis, tags: 'redis'}
+
+- name: configure the master redis server
```
sravfeyn: This redis/master/sentinel setup is to be run for only ICDS env, right? Are we rolling this out to all envs?

snopoke: only for ICDS yes. This will only run on hosts in the `redis_master` group which no other envs have. Is there a better way you know of to segregate this for only ICDS?

sravfeyn: I just got confused with the playbook name, that's fine then. I would probably create a separate `setup_redis_ha.yml` and call that from here when the env is configured for Redis HA (similar to the [setup_cluster](https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/redis/tasks/main.yml#L4).)

snopoke: The example you give is basically exactly the same. It includes the tasks if there are hosts in the `redis_cluster_master` group

src/commcare_cloud/ansible/roles/keepalived/tasks/main.yml
rohit-dimagi: we have to keep this task in one of the role either the keepalived or redis .

snopoke: OK

PR #3518, opened 2020-01-13
Es upgrade on India environment

sanjay2916:
Es upgrade on India environment to 2.4.6 version


dannyroberts:
Good work today on the migration! I want to merge this, but I can't do so with the last commit there, so I'll close this PR and open a similar one with the commits cleaned up and the last commit removed

PR #3519, opened 2020-01-13
[Amazon SES] remove modifiers to commcarehq-noreply and use separate emails

biyeun:
##### SUMMARY
It looks like Amazon SES does, in fact, strip the `+` modifiers from email addresses. I verified these new emails with SES and now have SNS notifications forwarding to `commcarehq-bounces@dimagi.com`. From there I should be able to process the more reliably-formatted SNS notifications *per environment* with the raw emails as a fallback.

##### ENVIRONMENTS AFFECTED
production, india, swiss, staging

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
`localsettings` `DEFAULT_FROM_EMAIL`

### DIFF
--- a/environments/india/public.yml
+++ b/environments/india/public.yml
@@ -60,7 +60,7 @@ s3_signature_version: 's3v4'
 root_email: commcarehq-ops+root@dimagi.com
 server_email: commcarehq-noreply@dimagi.com
 server_admin_email: commcarehq-ops+admins@dimagi.com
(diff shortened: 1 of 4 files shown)


biyeun:
ok, SES didn't in fact remove the modifier (I didn't restart services after updating the config). however, the separate emails allow a better way to direct the SNS notifications to a specific email so that I can directly fetch only the relevant messages in the `commcarehq-bounces` inbox

PR #3537, opened 2020-01-17
increase plproxy connection limits

snopoke:
##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request (will require short downtime to roll out)


snopoke:
@rohit-dimagi can you handle rolling this out please. It's going to require a DB restart so we should schedule a small window of downtime (unless we can do it with just a reload)

rohit-dimagi:
We can do with reload but looks complicated. https://www.postgresql.org/docs/current/sql-altersystem.html
i think rolling out during off hour should be fine.

snopoke:
@rohit-dimagi I meant just doing a process reload instead of a restart `systemctl reload postgresql`

rohit-dimagi:
Yeah the changed parameter requires a restart. The shared doc contains how to do it with reload.

PR #3546, opened 2020-01-22
Sk/get alias

snopoke:
##### SUMMARY
make `get_machine_alias` more strict so that it only considers a group that has a single host (not child groups) as a valid alias.

##### ENVIRONMENTS AFFECTED
NA

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
For the test case added in this PR the diff in aliases returned by the function is as follows:
```diff
- 'celery'
+ 'demo_server2'
- 'control'
+ 'demo_server0'
- 'zookeeper'
+ 'demo_server1'
```


src/commcare_cloud/environment/schemas/app_processes.py
```diff
@@ -194,8 +194,7 @@ def get_machine_alias(environment, host):
     :param host: the inventory host (expressed as i.e. an ip address) to look up
     :return: the canonical group name for that host
     """
-    for group, hosts in environment.groups.items():
-        if len(hosts) == 1 and hosts[0] == host:
-            return group
-    else:
-        return host
+    for name, group in environment.inventory_manager.groups.items():
```
dannyroberts: Is the difference that value `group` values that come out of `environment.inventory_manager.groups` have `group.hosts` and `group.children` (or something) so it's organized more directly the way the inventory file is, rather than inlining child groups' hosts?

That's the point of this PR right? That previously a host called `celery0` might have its alias returned as just `celery` because `celery0` is the only host in the `celery` group, but this change makes it return `celery0` as the _cononical_ group name (and not just the name of some group that it's in)?

snopoke: Yes, that's exactly it. `group.hosts()` will resolve all the descendants to get a list of actual hosts which looses the hierarchical information about how groups are made up. Looking at the `Group` object directly allows us to get access to that information without resolving all the subgroups so that we can see what hosts are directly declared in a group.

This is still dependent on our inventory convention and would fail for an inventory like this:

```
[celery0]
10.0.0.1

[celery]
10.0.0.1
```

PR #3548, opened 2020-01-22
Add explicit RATE_LIMIT_SUBMISSIONS setting for SaaS rate limiting

dannyroberts:
##### SUMMARY
This lets me decouple it from the `ENTERPRISE_MODE` setting, so that I can test SaaS rate limiting on staging even when someone's testing an `ENTERPRISE_MODE` change on it.

##### ENVIRONMENTS AFFECTED
production, india, staging


snopoke:
With regards to documentation, we should start adding some docs for settings (presumably in the HQ repo). I'm not sure what that looks like but at least for new settings I think it would make sense.

PR #3553, opened 2020-01-28
Fix ES CORS to enable Elasticsearch Head

dannyroberts:
##### SUMMARY
allow-origin "null" means allowing requests from file:// urls.
So if you run Elasticsearch Head locally from file://.../index.html
then this will re-enable it.

I haven't fully validated this, but my guess is that the upgrade to ES 2
gave cors a stricter default setting.

##### ENVIRONMENTS AFFECTED
all

##### COMPONENT NAME
Elasticsearch

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
@@ -21,6 +21,7 @@ transport.tcp.port: {{ elasticsearch_tcp_port }}
 http.port: {{ elasticsearch_http_port }}
 
 http.cors.enabled: true
+http.cors.allow-origin: "null"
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -21,6 +21,7 @@ transport.tcp.port: {{ elasticsearch_tcp_port }}
 http.port: {{ elasticsearch_http_port }}
 
 http.cors.enabled: true
+http.cors.allow-origin: "null"
```
snopoke: Is this a valid setting for v1.7?

dannyroberts: Appears to be: https://www.elastic.co/guide/en/elasticsearch/reference/1.7/modules-http.html

PR #3576, opened 2020-02-04
fix places that refer to pillow schema

snopoke:


src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -201,12 +201,11 @@ def run(self, args, unknown_args):
...
     pillows = environment.app_processes_config.pillows
-    by_process = defaultdict(lambda: {'num_processes': 0, 'total_processes': None})
+    total_processes = defaultdict(int)
     for host, processes in pillows.items():
         for name, options in processes.items():
-            config = by_process[name]
-            config['num_processes'] += options.get('num_processes', 1)
-    return by_process
+            total_processes[name] += options.num_processes
+    return total_processes
```
dannyroberts: I'm confused about this part. Doesn't this change in function signature affect all of the callers of this function?

snopoke: yes, it does, this was a bug that I fixed at the same time - sorry should have called that out.

PR #3577, opened 2020-02-04
Adding logstash and filebeat on india

rohit-dimagi:
##### SUMMARY

* Adding logstash and filebeat on india

##### ENVIRONMENTS AFFECTED

* INDIA
##### ISSUE TYPE

- Change Pull Request

### DIFF
--- a/environments/india/inventory.ini
+++ b/environments/india/inventory.ini
@@ -88,6 +88,12 @@ hot_standby_master=10.203.10.190
 [proxy:children]
 proxy1
 
+[filebeat:children]
+proxy1
+
+[logproxy:children]
+proxy1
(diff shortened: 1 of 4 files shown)


environments/india/inventory.ini
```diff
@@ -88,6 +88,12 @@ hot_standby_master=10.203.10.190
 [proxy:children]
 proxy1
 
+[filebeat:children]
```
snopoke: Instead of creating a new inventory group for this what about just using the existing groups since we know where we want filebeat to run already?

rohit-dimagi: Make sense.

PR #3585, opened 2020-02-05
set ES default query timeout to 5 mins

sravfeyn:
@snopoke


sravfeyn:
Rolling this out.

sravfeyn:
Rolled out

PR #3592, opened 2020-02-11
Added Boston office IP to rate limiting whitelist

orangejenny:


orangejenny:
Is merging and then restarting nginx sufficient to roll this out?

orangejenny:
@snopoke thank you!

PR #3594, opened 2020-02-13
limit domain ES index to a single shard

snopoke:
##### SUMMARY
Index is very small so no need to sharding

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request


sravfeyn:
This is a good change. Thought, shards can't be shrinked without reindexing, I think?

snopoke:
The reindex was already done.

PR #3595, opened 2020-02-13
update DD log parsing for nginx

snopoke:
##### SUMMARY
* Parser changed to use a single function to parse all nginx timing data.
  * https://github.com/dimagi/datadog-parsers/pull/37
* Use different log paths for different sites

##### ENVIRONMENTS AFFECTED
ICDS

##### ISSUE TYPE
- Change Pull Request

##### COMPONENT NAME
Datadog


snopoke:
rolled out to ICDS

snopoke:
@dimagi/devops FYI there's a change here that can be applied to SaaS if you want. We're also working on moving this log parsing to logstash so we can upgrade DD agent so also fine to wait for that before rolling out changes.

PR #3597, opened 2020-02-13
Add restart always to elasticsearch

rohit-dimagi:
##### SUMMARY

* Add restart always to elasticsearch

##### ENVIRONMENTS AFFECTED

* ALL
##### ISSUE TYPE

- Change Pull Request


snopoke:
What's the impact of this.

rohit-dimagi:
Configures whether the service shall be restarted when the service process exits, is killed, or a timeout is reached. When the death of the process is a result of systemd operation (e.g. service stop or restart), the service will not be restarted.

PR #3599, opened 2020-02-13
Removed ils-gateway sms queue

orangejenny:
ILS Gateway ended a while ago.


orangejenny:
@dannyroberts Does this need any kind of rolllout, and if so what is it?

dannyroberts:
Just production, right? It'll need `cchq production update-supervisor-confs`, but I'll just run that right now

orangejenny:
Thanks @dannyroberts

PR #3643, opened 2020-02-20
change proxy ip in icds env

rohit-dimagi:
##### SUMMARY

* change proxy ip in `public.yml`

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Change Pull Request

PR #3644, opened 2020-02-20
ICDS Proxy server IP update 

sanjay2916:
ICDS Proxy server IP update


sanjay2916:
ref:  Pull Request #3643 ·

PR #3649, opened 2020-02-21
Rotate production-s3-key and use for kissmetrics/backups as well

dannyroberts:
##### SUMMARY
This also makes it use a different key from the one india users.

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
Gonna self-merge to not keep production locked

PR #3655, opened 2020-02-24
Update firefighting docs for couch disk full situation

dannyroberts:
##### SUMMARY
Some takeaways from @millerdev's firefighting this morning is that some key elements of what to do in this situation were missing from the docs, so this is an attempt to fill that in.

### DIFF
--- a/docs/firefighting/index.md
+++ b/docs/firefighting/index.md
@@ -121,7 +121,7 @@ $ cchq <env> service commcare stop
 
 ## Couch node is down
 
-If a couch node is down, the couch disk might be full. In that case, see "Couch node data disk is full" below. Otherwise, it could mean that the node is slow to respond, erroring frequently, or that the couch process or VM itself in a stopped state.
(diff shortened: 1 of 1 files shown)


docs/firefighting/index.md
```diff
@@ -121,9 +121,9 @@ $ cchq <env> service commcare stop
 
 ## Couch node is down
 
-If a couch node is down, it could mean that it is either very slow at responding to requests or it has stopped running.
+If a couch node is down, the couch disk might be full. In that case, see "Couch node data disk is full" below. Otherwise, it could mean that the node is slow to respond, erroring frequently, or that the couch process or VM itself in a stopped state.
```
millerdev:
```suggestion
If a couch node is down, the couch disk might be full. In that case, see [Couch node data disk is full](#couch-node-data-disk-is-full) below. Otherwise, it could mean that the node is slow to respond, erroring frequently, or that the couch process or VM itself in a stopped state.
```

PR #3656, opened 2020-02-25
Add ecrypted option on block_device in terraform.yml

dannyroberts:
##### SUMMARY
This will let us eventually use native AWS EBS encryption
instead of ecryptfs for our large data volumes.

Down the road this will enable zero-intervention startup, which is a prerequisite for any kind of automatic scale up and down, whether it's stopping a couple webworkers during trough periods, or actual autoscaling. It also may reduce the cost of our couch snapshot backups.

I spiked this and it appears to work, so I wanted to just get it in the the caveat so I could come back to it when I have time


dannyroberts:
Here's their high level docs on the encrypted EBS volumes: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html.

They have a key management system where the key is stored separately (with its own access control managed by them) from the data it encrypts.

From a HIPAA perspective it's compliant with the at-rest encryption standard, and from a threat model perspective, anyone (an AWS employee, say) who could exploit having access to the key would also be able to exploit having access to our live machines. Those are the broad strokes that went into thinking about whether using encrypted EBS is acceptable/preferable.

millerdev:
Thanks for the explanation, it makes sense.

PR #3668, opened 2020-02-27
sync_token mapping for proxy server

rohit-dimagi:
##### SUMMARY

* sync_token mapping for proxy server

##### ENVIRONMENTS AFFECTED

* ICDS
##### ISSUE TYPE

- Change Pull Request


src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -61,7 +61,38 @@ nginx_sites:
...
+          value: "$limit_bucket"
+        - key: "default"
+          value: '""'
+    - name: "$synced_old_bucket"
+      variable: "$restore_limit"
+      var_mapping:
+        - key: "1"
+          value: "$limit_bucket"
+        - key: "default"
+          value: '""'
```
snopoke: I think you can reduce this to:

```suggestion
    - name: "$synced_today_bucket"
      variable: "$days_since_last_sync"
      var_mapping:
        - key: "0"
          value: "$limit_bucket"
        - key: "default"
          value: '""'
    - name: "$synced_old_bucket"
      variable: "$days_since_last_sync"
      var_mapping:
        - key: "~[1-9]\d*"  # match number > 0
          value: "$limit_bucket"
        - key: "default"
          value: '""'        
```

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -61,7 +61,38 @@ nginx_sites:
...
+
    limit_req_zones:
+    - name: "synced_today_bucket"
+      zone: "$restore_limit"
+      size: "10m"
+      rate: "40r/s"
+    - name: "synced_old_bucket"
+      zone: "$restore_limit"
+      rate: "60r/s"  
     - name: "incremental_restore" # Limits restores that are not new device or 412s
```
snopoke: this can be removed completely since both `synced_today_bucket` and `synced_old_bucket` are 'incremental' syncs

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -160,6 +191,12 @@ nginx_sites:
        proxy_next_upstream_tries: 1
        proxy_read_timeout: 900s
        limit_reqs:
+         - zone_name: "synced_old_bucket"
+           burst: 10
+           nodelay: True
```
snopoke: we can remove the burst and nodelay - they don't serve any purpose for these global limits (only useful for per IP / per user)

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
rohit-dimagi: Removed

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
rohit-dimagi: updated.

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -61,11 +61,29 @@ nginx_sites:
         # If the header is not provided, do not rate limit
```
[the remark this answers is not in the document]
rohit-dimagi: Removed.

rohit-dimagi:
> A few more fixes. Also from my understanding with the way we have defined the mappings, the different limits will be applied independently (except for the `overall_restore` limit). This is because we assign and empty string to the key if it doesn't match which should mean that nginx doesn't apply the limit to that request:
> 
> > Requests with an empty key value are not accounted.(https://nginx.org/en/docs/http/ngx_http_limit_req_module.html)
> 
> SO, the limits for the different zones should only apply to that zone so having 40 & 60 will mean that we get 100r/s (assuming both limits get hit). In that case the limits should be lower. My suggestion would be:
> 
>     1. never synced: 10
> 
>     2. old syncs: 10
> 
>     3. synced today: 10
> 
>     4. overall (safety net): 30
> 
> 
> By implementing a new metric to track which rate limit zones are being hit we can adjust these going forward: https://dimagi-dev.atlassian.net/browse/IIO-755

Adjusted the rate value accordingly

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -54,23 +54,44 @@ nginx_sites:
...
-      variable: "$http_x_commcarehq_lastsynctoken"
-      # If header is provided, rate limit on $limit_bucket
-      default_value: "$global_limit"
-      var_mappings:
-        # If the header is not provided, do not rate limit
-        - key: '""'
+    - name: "$synced_never_bucket"
+      variable: "$days_since_last_sync"
+      var_mapping:
+        - key: ""
```
snopoke:
```suggestion
        - key: '""'
```

PR #3671, opened 2020-02-27
only set setting on citusdb_master

snopoke:


src/commcare_cloud/ansible/roles/citusdb/tasks/main.yml
```diff
@@ -32,7 +32,7 @@
      postgresql_shared_preload_libraries:
        - 'citus'
        - 'pg_stat_statements'
-     postgresql_custom_settings: "{{ [\"citus.use_secondary_nodes = 'always'\"] if is_pg_standby else []}}"
+     postgresql_custom_settings: "{{ [\"citus.use_secondary_nodes = 'always'\"] if is_pg_standby and 'citusdb_master' in groups else []}}"
```
rohit-dimagi: i don't why but this evaluation condition is not working as it should be.

snopoke: is there an error?

rohit-dimagi: No error . it evaluates to true even for the citusworkerstandby node.

snopoke:
```suggestion
     postgresql_custom_settings: "{{ [\"citus.use_secondary_nodes = 'always'\"] if is_pg_standby and inventory_hostname in groups['citusdb_master']|default([]) else []}}"
```

PR #3676, opened 2020-02-27
Support deploying specific commit

orangejenny:
As discussed with @dannyroberts a few weeks back, this is to make the couch to sql model migration easier for third parties. It lets us add a specific commit to [this kind of message](https://github.com/dimagi/commcare-hq/blob/f13091d1d39f4a27f91c23bb09d32a791ea2ce0a/corehq/apps/cleanup/management/commands/populate_sql_model_from_couch_model.py#L74-L77) with instructions for deploying that commit, running the management command with `cchq`, and then deploying master.


orangejenny:
@millerdev There are probably situations where it's unsafe to run old code, although I don't think this PR adds any risk, since it's already possible to deploy an old branch. I do like the thought of having the user-facing instructions tell you to set up a private release rather than do a full deploy.

dannyroberts:
@millerdev I think the intended use is for third party hosters who haven't updated their code in a while. Then we might expect the following timeline:

- They deploy to their instance on January 1
- We make a blocking change on Jan 15, "commit X"
- They try to deploy to their instance on Feb 1
- They see a message that says "to upgrade you first need to deploy from commit X, and then you can deploy the latest change"
- They deploy commit X
- They deploy the latest change

Since they'd have to have deployed commit X to be on any commit later than it, it's unlikely that anything we'd ask them to do would result in them going back in time. At least that's the idea. That said, having a check that the new commit contains the current commit in its history when deploying isn't a terrible idea.

src/commcare_cloud/commands/deploy.py
```diff
@@ -26,7 +26,10 @@ class Deploy(CommandBase):
             Skip the steps involved in recording and announcing the fact of the deploy.
         """),
         Argument('--commcare-branch', help="""
-            The name of the commcare-hq git branch to deploy. 
+            The name of the commcare-hq git branch to deploy. Mutually exclusive with commcare-hash.
+        """, default=None),
+        Argument('--commcare-hash', help="""
+            The hash of a commcare-hq commit to deploy. Mutually exclusive with commcare-branch.
```
dannyroberts: Thoughts on `--commcare-hash` versus `--commcare-commit`? Both `branch` and `commit` are deeply intertwined with our git repo, so it seems preferably to use the `git` terminology for both?

millerdev: +1 for `--commcare-commit` or if this supports any name for a given commit (branch, tag, commit) then `--commcare-rev`

orangejenny: Ha, I went back and forth over naming. My read of [this article](https://www.atlassian.com/git/tutorials/refs-and-the-reflog?section=special-refs) was that both branches and hashes are ways to describe a commit, so it'd be odd to have both "branch" and "commit" args that do different things, and that "hash" is the best way to refer to the SHA. Is that not the case?

This doesn't support tags - [this section](https://github.com/dimagi/commcare-cloud/pull/3676/files#diff-ad8be748d06708f97faa6e0587d23784R142-R147) - would need changes, and I'd rather not do that unless/until there's a need.

millerdev: I agree and it was in the back of my mind that it is strange to have both `--commcare-branch` as well as this new option. Can you elaborate on how they are different? Is it ever the case that one would use both at the same time?

> This doesn't support tags - [this section](https://github.com/dimagi/commcare-cloud/pull/3676/files#diff-ad8be748d06708f97faa6e0587d23784R142-R147) would need changes...

I think `git rev-parse` (which takes a branch name, tag name, commit hash, etc. and returns a commit hash) might work nicely there. [`vellum-to-hq` does something similar](https://github.com/dimagi/commcare-hq/blob/5b79c82821bd9f306f461884771db5d5306dcdfc/scripts/vellum-to-hq#L201).

orangejenny: Thank you, having `gitutils` to reference makes this simpler.

orangejenny:
@millerdev @dannyroberts Updated to support deploying from a tag, renamed the arg to `commcare-rev`, and added a changelog entry.

orangejenny:
@dannyroberts Any guidance on tests? They're failing with:
```
rm: cannot remove ‘docs/changelog/0033-update-deploy-args.md’: No such file or directory
make: *** [clean] Error 1
The command ".travis/tests.sh" exited with 2.
```
I did run `make` as described in the changelog readme. Is that supposed to be done in a separate PR?

dannyroberts:
Nice, awesome work!

src/commcare_cloud/fab/utils.py
```diff
@@ -140,11 +141,7 @@ def deploy_ref(self):
             return env.code_branch
 
         # turn whatever `code_branch` is into a commit hash
-        try:
-            branch = self.repo.get_branch(self._code_branch)
-            return branch.commit.sha
-        except GithubException:
-            return self.repo.get_commit(sha=self._code_branch).sha
+        return sh.git.bake(_tty_out=False)("rev-parse", self._code_branch).strip()
```
millerdev: I am sorry for leading you astray. I don't think this will work since this will most likely operate on the local repository (`commcare-cloud`), not `dimagi/commcare-hq` referenced by `self.repo`. I will see if I can come up with a better option...

orangejenny: That is a good point, this isn't acting in the right repo. Curiously, this does run without error, although I only tested enough to see it didn't throw an error, I didn't do a full deploy.

pygithub allows you to list a repo's tags; that's the approach I was considering when I said I didn't want to bother adding support for tags.

millerdev: I think this will work (tested locally with branch name, commit hash, and tag name)
```suggestion
        return self.repo.get_commit(self._code_branch).sha
```

PR #3708, opened 2020-03-12
Mmthiam add users pna

mmthiam:
##### ISSUE TYPE

- Change Pull Request
- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- Downtime Pull Request

 

##### ADDITIONAL INFORMATION

```paste below

```

* Before
```paste below

```

* After
```paste below

```


esoergel:
It looks like there are some issues with this branch:

> This branch is 5454 commits ahead, 9282 commits behind dimagi:master

Has PNA been maintaining a separate fork of commcare-cloud?  If not, the easiest thing to do would be to make a new branch off of dimagi:master, and cherry-pick the relevant commits on top of that, then open a new PR.

If there _is_ a separate fork with more changes, there are two possible routes to merge the two

**The fork only contains meaningful changes in `environment/pna/`:**
Make a new commit on top of `dimagi:master` which overwrites everything in `environment/pna/` to the versions currently in the fork.  I'd strongly recommend reviewing the diff closely to ensure the changes look accurate.

**The fork has substantive changes outside of `environnment/pna/`:**
This is much harder.  I'd start by attempting a rebase on top of master, or perhaps by cherry-picking or replicating the relevant changes.  Either way, the git history will need to be cleaned up for it to be mergeable.

### Keeping the fork separate
If there are changes necessary for PNA's environment which for whatever reason can't be PR'd into the main commcare-cloud repository, then the only option is to maintain a separate fork.  This would mean collaboration on that fork would have to happen in a different repository, which regularly merges in changes from `dimagi:master`

### Keeping the environment configuration separate
Many environments are stored in the main commcare-cloud repository, which makes it easier to keep them in sync with changes to commcare-cloud itself.  However, environments not managed by Dimagi, this approach likely doesn't make sense, since Dimagi approval is needed for edits.  Rather, the environment directory is stored in isolation, likely as another git repository.  Here is a [sample environment](https://github.com/dimagi/sample-environment) managed in that way.  That is how the [new environment instructions](https://dimagi.github.io/commcare-cloud/setup/new_environment.html) setup these configuration files.

Of course, if this route is taken, you will still need to monitor the [changelog](https://dimagi.github.io/commcare-cloud/changelog/) for breaking changes.  Should any modifications to commcare-cloud be required, those would also need to be PR'd in indvidually.

PR #3712, opened 2020-03-13
Fix production postgres password

dannyroberts:
##### SUMMARY
It was accidentally changed in 81387fe074707657b9fba0c22af207d9df9bdb26. @snopoke FYI. Nothing bad ended up happening, but after this commit `cchq production terraform plan` showed an unapplied diff for this password. (It was for the root user which isn't the one we use in the application, so locallsettings.py wasn't affected.)

To see what I mean run

```
curl https://gist.githubusercontent.com/dannyroberts/0c85bad47423400dfe5de50f9cc16942/raw/9cd3fe90bea30ddc7a729687b50558092c7abf59/git_log_vault.sh > git_log_vault.sh
read -s ANSIBLE_VAULT_PASS  # enter production vault password
ANSIBLE_VAULT_PASS="${ANSIBLE_VAULT_PASS}" bash git_log_vault.sh environments/production/vault.yml
```

That's a bash script I put together to find the commit that basically acts like `git log -p` but on a vault file.

##### ENVIRONMENTS AFFECTED
production


snopoke:
@dannyroberts thanks for the fix, looks like it must have been a paste error. Also nice script. I had to change it slightly to make it work: `--vault-password-file=src/commcare_cloud/ansible/echo_vault_password.sh`.

Seems like something useful to add to the repo

PR #3720, opened 2020-03-17
Making ecryptfs as optional in Ansible playbooks

sanjay2916:
Making ecryptfs as optional in Ansible playbooks

### DIFF
--- a/src/commcare_cloud/ansible/roles/ecryptfs/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/ecryptfs/tasks/main.yml
@@ -1,125 +1,121 @@
 ---
 
-- name: Check for valid partition table on block storage device
-  become: yes
-  shell: fdisk -l 2>&1 | grep 'Disk {{ datavol_device }} doesn.t contain a valid partition table'
-  register: partition_check
-  when: datavol_device is defined and ansible_hostname not in groups.get('lvm', [])
-  changed_when: false
-  failed_when: false
-  check_mode: no
+- block:
+    - name: Check for valid partition table on block storage device
+      become: yes
+      shell: fdisk -l 2>&1 | grep 'Disk {{ datavol_device }} doesn.t contain a valid partition table'
+      register: partition_check
+      when: datavol_device is defined and ansible_hostname not in groups.get('lvm', [])
+      changed_when: false
+      failed_when: false
+      check_mode: no
 
-- name: Create partition on block storage volume
-  become: yes
-  shell: echo -e "o\nn\np\n1\n\n\nw" | fdisk {{ datavol_device }}
-  # http://superuser.com/a/739586/86694
-  # o - clear in memory partition table
-  # n - new partition
-  # p - primary partition
-  # 1 - partition number 1
-  #   - default - start at beginning of disk
-  #   - default - use entire disk
-  # w - write the partition table
-  when: datavol_device is defined and partition_check.rc == 0 and ansible_hostname not in groups.get('lvm', [])
+    - name: Create partition on block storage volume
+      become: yes
+      shell: echo -e "o\nn\np\n1\n\n\nw" | fdisk {{ datavol_device }}
+      # http://superuser.com/a/739586/86694
+      # o - clear in memory partition table
+      # n - new partition
+      # p - primary partition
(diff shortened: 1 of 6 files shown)


src/commcare_cloud/ansible/roles/ecryptfs/tasks/main.yml
```diff
@@ -25,7 +25,7 @@
 - name: Format block storage volume
   become: yes
   filesystem: fstype=ext4 dev="{{ datavol_device1 }}"
-  when: datavol_device is defined
+  when: datavol_device is defined and root_encryption_mode != 'aws'
```
dannyroberts: can you use the positive condition instead :`and root_encryption_mode == 'ecryptfs'`? That way seems clearer to me.

Also instead of adding the same condition to every task, could you wrap all of the tasks in the file in a single block that has that condition? Here's an example of somewhere where we use that pattern: https://github.com/dimagi/commcare-cloud/blob/982a3cf7e8283122edd004739bed280ef5a5bfba/src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml#L52-L100

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -183,11 +183,19 @@ def context(self):
...
+                        ('datavol_device', '/dev/sdf'),
+                        ('datavol_device1', '/dev/sdf'),
+                        ('is_datavol_ebsnvme', 'yes'),
+                    ])
+                else :
+                    inventory_vars.extend([
+                        ('datavol_device', '/dev/sdf'),
+                        ('datavol_device1', '/dev/sdf'),
+                        ('is_datavol_ebsnvme', 'yes'),
+                    ])
```
dannyroberts:
```suggestion
                inventory_vars.extend([
                    ('datavol_device', '/dev/sdf'),
                    ('datavol_device1', '/dev/sdf'),
                    ('is_datavol_ebsnvme', 'yes'),
                ])
                if server.block_device.encrypted:
                    inventory_vars.append(
                        ('root_encryption_mode', 'aws'),
                    )
```

dannyroberts:
Hey Sanjay, looks pretty good! I left a few change requests. Once you have those in, could you
- manually test the code against staging and make sure that in addition to working for the new aws setup, it doesn't break things for the ecryptfs setup?
- make sure the test includes restarting a staging machine that uses ecryptfs, running after-reboot on it, and making sure that it does the right thing (decrypts /opt/data)?
- comment on this PR saying what you did to test and what the outcomes were
?

src/commcare_cloud/ansible/roles/ecryptfs/tasks/main.yml
```diff
@@ -1,119 +1,121 @@
...
 
-- name: Create blob DB symlink (cchq looks for it in the shared drive)
-  become: yes
-  when: not shared_drive_enabled
-  file: src="{{ encrypted_root }}/blobdb" dest="{{ shared_data_dir }}/blobdb" state=link
-  tags:
-    - ecryptfs-blobdb
+    - name: Create blob DB symlink (cchq looks for it in the shared drive)
+      become: yes
+      when: not shared_drive_enabled and root_encryption_mode != 'aws'
```
dannyroberts: It looks like this was left accidentally?

sanjay2916: yes Danny and removed

src/commcare_cloud/ansible/roles/ecryptfs/tasks/main.yml
```diff
@@ -1,119 +1,121 @@
...
-  shell: fdisk -l 2>&1 | grep 'Disk {{ datavol_device }} doesn.t contain a valid partition table'
-  register: partition_check
-  when: datavol_device is defined and ansible_hostname not in groups.get('lvm', [])
-  changed_when: false
-  failed_when: false
-  check_mode: no
+- block:
+    - name: Check for valid partition table on block storage device
+      become: yes
+      shell: fdisk -l 2>&1 | grep 'Disk {{ datavol_device }} doesn.t contain a valid partition tablef'
```
dannyroberts: The `f` looks like a stray keystroke?

PR #3722, opened 2020-03-19
disable materializing for citus workers

calellowitz:
##### SUMMARY

The child health monthly step of the aggregation seems to have very different performance characteristics based on whether or not the postgres query planner chooses to partially materialize the query. This penalizes materialization in the planner costing algorithm, although it will still use it if required for correctness. I suspect the best option is to try to improve postgres stats so it will just choose the correct plan, but given turnover in these tables (we delete and recreate a substantial part of it in each agg), that might be hard to do.

##### ENVIRONMENTS AFFECTED

ICDS
India

##### ISSUE TYPE

- Change Pull Request

 

##### COMPONENT NAME

Citus


calellowitz:
Deployed, so merging to PR to prevent accidental rollback

PR #3729, opened 2020-03-24
Simplify pull request template

esoergel:
I took a look at the 50 most recent PRs.  None included anything in the "ADDITIONAL INFORMATION" section.  Three included a "COMPONENT NAME", but that name was also in the PR title.  Where "ISSUE TYPE" was filled out, it was only ever "Change Pull Request" (not especially useful information).  "ENVIRONMENTS AFFECTED" was usually filled out, and that does seem useful to have.

Removing the rest of this since it's an annoyance everytime I make a new PR.

-------

##### ISSUE TYPE

- Change Pull Request
- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- Downtime Pull Request

 

##### ADDITIONAL INFORMATION

```paste below

```

* Before
```paste below

```

* After
```paste below

```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,6 +1,4 @@
 <!--- Provide a link to the ticket or document which prompted this change -->
 
-##### SUMMARY
-
 ##### ENVIRONMENTS AFFECTED
 <!--- list which environments are affected by this change or None if this doesn't change any environment files -->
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,40 +1,6 @@
-##### SUMMARY
-<!--- Describe the change below, including rationale and design decisions -->
+<!--- Provide a link to the ticket or document which prompted this change -->
```
calellowitz: Nit: I think if summary is not the first line in this template, people are like to put their description above it. I am neutral as to whether we even need that heading, but if we have it I think it should be first, and the comments/instructions about what to put there should be beneath it

esoergel: :+1: I just removed it.
(was copying from what commcare-hq does, but I agree that the heading doesn't really add anything)

PR #3735, opened 2020-03-25
queue async indicators all day

calellowitz:
We normally only queue indicators at night but given the lower form submissions as AWCs, we will run them all day to help clear the backlog.

##### ENVIRONMENTS AFFECTED

ICDS


mkangia:
let them run!

PR #3736, opened 2020-03-25
Add --tee arg to django-manage command

dannyroberts:
This gives a way to redirect output to a file on the remote machine, which comes in handy for workflows where the output of one command is the input of another, or simply for saving the output more durably especially in conjunction with `--tmux`

### DIFF
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -193,6 +193,9 @@ class DjangoManage(CommandBase):
             E.g. '2018-04-13_18.16'.
             If none is specified, the `current` release will be used.
         """),
+        Argument('--tee', dest='tee_file', help="""
+            Tee output to the screen and to this file on the remote machine
+        """)
     )
 
     def run(self, args, manage_args):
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -209,14 +212,24 @@ def run(self, args, manage_args):
             code_dir = '/home/{cchq_user}/www/{deploy_env}/current'.format(
                 cchq_user=cchq_user, deploy_env=deploy_env)
 
+        if args.tee_file:
+            tee_file_cmd = ' | tee {}'.format(shlex_quote(args.tee_file))
```
millerdev: Is there any concern that specifying the wrong (possibly relative) path could cause something important to be overwritten?

dannyroberts: That's a great point. I can think of two possibilities for addressing that. One would be to have the command include a check that that file doesn't already exist, the other would be to have all file paths be relative to a directory that's dedicated for this (like under `/opt/data/scratch` or something) and disallow paths whose absolute rendering does not fall under it

millerdev: Both of those seem reasonable, I think I like the first one better if it can be done reliably. One issue I can think of is that the code will likely want to validate the path on the machine (developer workstation) where the command is being initiated, but it really needs to perform the check on the remote machine (djangomanageN), so doing either of those checks seems non-trivial.

dannyroberts: OK what do you think about this approach f58e7a1? It runs a separate ssh command to first check if the file exists, then either issues an error message and exits or continues on with the target command. I've tested it both with and without `--tmux` and the behavior is what you'd expect. For commands that don't use `--tee` there's no check so no overhead, and for commands that do use `--tee` the overhead of a separate ssh connection for the check is tiny compared to even the startup overhead of `./manage.py`, so seems reasonable from that perspective too

PR #3748, opened 2020-04-01
Logging format change for Nginx

rohit-dimagi:
* https://dimagi-dev.atlassian.net/browse/IIO-807
##### ENVIRONMENTS AFFECTED

ALL

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/templates/nginx.conf.j2
+++ b/src/commcare_cloud/ansible/roles/nginx/templates/nginx.conf.j2
@@ -18,7 +18,7 @@ http {
         access_log /var/log/{{ nginx_access_log_name }};
         error_log /var/log/{{ nginx_error_log_name }} warn;
 
-        log_format timing '[$time_local] $upstream_cache_status $request $status $request_time $http_referer';
+        log_format timing '[$time_local] $upstream_cache_status $request $status $request_time $request_length $bytes_sent $http_referer';
         log_format rt_cache '$remote_addr - $remote_user - $upstream_cache_status - [$time_local] '
                     '"$request" $status $request_length $body_bytes_sent '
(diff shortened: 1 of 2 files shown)


rohit-dimagi:
I have changed format for rt_cache to align it more with timing ones and added the byte sent and request length parameters.

rohit-dimagi:
Since it will break logstash so i'll start with the cache hit and when it's completed I'll rollout for both.

snopoke:
> I have changed format for rt_cache to align it more with timing ones and added the byte sent and request length parameters.

Any reason for this? They are now identical so there's no need to have both. For now I think better to just change the `timing` one since that's only used for metrics. If we want to change the other one we'll need to get wider buy in.

src/commcare_cloud/ansible/roles/nginx/templates/nginx.conf.j2
```diff
@@ -18,10 +18,8 @@ http {
         access_log /var/log/{{ nginx_access_log_name }};
         error_log /var/log/{{ nginx_error_log_name }} warn;
 
-        log_format timing '[$time_local] $upstream_cache_status $request $status $request_time $http_referer';
-        log_format rt_cache '$remote_addr - $remote_user - $upstream_cache_status - [$time_local] '
-                    '"$request" $status $request_length $body_bytes_sent '
-                    '"$http_referer" "$http_user_agent"';
+        log_format timing '[$time_local] $upstream_cache_status $request $status $request_time $request_length $body_bytes_sent $http_referer';
```
snopoke, line 21: prefer `$bytes_sent` over `$body_bytes_sent` since it includes headers etc:
```suggestion
        log_format timing '[$time_local] $upstream_cache_status $request $status $request_time $request_length $bytes_sent $http_referer';
```

PR #3750, opened 2020-04-02
fix cpu count for nginx worker process

rohit-dimagi:
* https://dimagi-dev.atlassian.net/browse/IIO-808
##### ENVIRONMENTS AFFECTED

* ALL


rohit-dimagi:
Rolled out on icds , i'll restart the service during downtime.

PR #3751, opened 2020-04-02
commcare-cloud upgrade to python 3

AmitPhulera:
ansible_user = ec2-user

https://dimagi-dev.atlassian.net/browse/SAAS-135

This branch contains the changes that will eventually make commcare-cloud compatible with python 3

### To test it in your system
```
# make virtual env with python 3
mkvirtualenv cchq-py3 --python $(which python3) 
workon cchq-py3

# Clone couchdb-cluster-admin and install the branch that has support for python 3

mkdir cchq-test && cd cchq-test
git clone https://github.com/dimagi/couchdb-cluster-admin/
cd couchdb-cluster-admin
git checkout ap/upgrade-py3
pip install -e .

cd ..

# Clone and install commcare cloud
git clone https://github.com/dimagi/commcare-cloud/
cd commcare-cloud/
git checkout ap/upgrade-to-python3
pip-sync requirements.txt
pip install -e .

# for running unit tests
pip install -e .[test]
nosetests
```


environments/icds-cas/migrations/nic_to_tcl/couchdb-files-post.py
```diff
@@ -33,12 +35,12 @@ def main():
               rev = res.json()['_rev']
               url = node_url.format(old_node)
               res = requests.delete('{}?rev={}'.format(url, rev), auth=auth)
-              print('DELETE node {}'.format(old_node), res.status_code)
+              print(('DELETE node {}'.format(old_node), res.status_code))
```
snopoke, line 38: double brackets in a few places

scripts/es_snapshot.py
```diff
@@ -249,8 +252,8 @@ def run(self):
         for version in live_snapshots:
             print(f'    {version}')
 
-        confirm = input(
-            f"Are you sure you want to delete all files that don't belong to a live ES snapshot? [y/n]")
+        confirm = eval(input(
```
snopoke, line 255: why would we want to `eval` this? That seems very dangerous

dannyroberts: @AmitPhulera I think (confusingly) this file is already a py3-only file that is meant to be copied directly to a server any run on that server with its own `python3` executable, something that we get a hint at from the first line in the file
```
#!/usr/bin/env python3
```

So I agree with Simon. I think all of the changes in this file should be reverted so that the PR doesn't touch this file at all. That should also be true about any file whose first line matches `'^\#\!.*python3'`, which you can quickly find with `grep -R '^\#\!.*python3' .`

environments/icds-cas/migrations/nic_to_tcl/couchdb-files-post.py
AmitPhulera: Fixed.

scripts/es_snapshot.py
AmitPhulera: There were only two such files
- `es_snapshot.py`
- `pg_repack.py`

Reverted both of them back.

dannyroberts:
Looks like the pip install is failing on python3 causing problems later on in the tests
```
ERROR: Could not find a version that satisfies the requirement futures==3.3.0 (from -r /tmp/tmpec3tugur (line 23)) (from versions: 0.2.python3, 0.1, 0.2, 1.0, 2.0, 2.1, 2.1.1, 2.1.2, 2.1.3, 2.1.4, 2.1.5, 2.1.6, 2.2.0, 3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.0.4, 3.0.5, 3.1.0, 3.1.1)
```
I forget, maybe we've seen this error already

PR #3756, opened 2020-04-06
Added dynamic configuration for postgresql parameters

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/IIO-805

##### ENVIRONMENTS AFFECTED
ICDS


src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
```diff
@@ -21,7 +21,9 @@ ssl = {% if postgresql_ssl_enabled %}true{% else %}false{% endif %}
 
 # RESOURCE USAGE (except WAL)
 
-shared_buffers = {{ postgresql_shared_buffers }}
+{% if postgresql_shared_buffers %}
```
snopoke, line 24: this if statement is confusing (and the other one for `postgresql_effective_cache_size`). I don't think it should still be there

shyamkumarlchauhan: Removed the if conditions

PR #3771, opened 2020-04-15
Fix ES monit

dannyroberts:
Took me hours to figure this out. I was testing something out on staging ES, and as a side effect deployed monit. Apparently it was checking for a file that didn't exist and just restarting ES in a loop, because fixing the path to the correct path for the log file fixed it.

##### ENVIRONMENTS AFFECTED
Fixed would-be issue in all environments but ICDS


biyeun:
@dannyroberts yikes! nice catch

PR #3778, opened 2020-04-17
Add prod formplayer and hq_webworkers machines

dannyroberts:
Routine scale-up based on some stats I'm seeing from usage and expected usage.

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
Everything's set up and the machines were part of the last deploy, but no traffic is being routed to them yet.

I'll pick an off time on the weekend to run `cchq production ansible-playbook deploy_proxy.yml --tags=nginx_sites --branch=dmr/up-prod-formplayer-and-hq_webworkers` to add those in

dannyroberts:
Rolled this out just now. I also went through a somewhat elaborate process of syncing all (both) current machines' /opt/data/formplayer files to all machines (including the other current one), so that all machines had all db files to start with. I ran the syncing in a loop and then stopped formplayer and let the syncing continue to finalize getting everything into sync while I ran the above command to have nginx start routing traffic to new machines, and then killed the syncing and started formplayer back up after that was done.

Importantly, the syncing used the `rsync --update` flag
```
 -u, --update                skip files that are newer on the receiver
```
so that when both machines had files with conflicting names (because one of them had been down at some point and users had been routed to the other one) the newer one always won out.

Here are the commands I used
[rsyncing-formplayer-db-files.txt](https://github.com/dimagi/commcare-cloud/files/4496906/rsyncing-formplayer-db-files.txt)

This of course highlights the fact that the way we store dbs locally on formplayer machines and rely heavily on sticky routing to prevent very unintuitive behavior (especially on smsforms) makes our setup very brittle

PR #3786, opened 2020-04-28
Kafka upgrade

satyaakam:
https://dimagi-dev.atlassian.net/browse/IIO-752?atlOrigin=eyJpIjoiZGViYTQ4NDBhZmI4NGRiMmEyZWZiMDM4YmMxMTcxZDgiLCJwIjoiaiJ9
##### ENVIRONMENTS AFFECTED

Staging


src/commcare_cloud/ansible/roles/kafka/files/log4j_new.properties
```diff
@@ -0,0 +1,74 @@
+log4j.rootLogger=INFO, stdout, kafkaAppender
+
+log4j.appender.stdout=org.apache.log4j.ConsoleAppender
+log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=[%d] %p %m (%c)%n
+
+log4j.appender.kafkaAppender=org.apache.log4j.DailyRollingFileAppender
+log4j.appender.kafkaAppender.DatePattern='.'yyyy-MM-dd-HH
+log4j.appender.kafkaAppender.File=${kafka.logs.dir}/server.log
```
snopoke, line 9: are these meant to be resolved by ansible? If so it needs to be moved to the `templates` folder and the syntax is `{{var_name}}`

src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
```diff
@@ -55,6 +55,12 @@
 
 - name: Copies kafka log4j.properties
   copy: src=log4j.properties dest=/opt/kafka/config owner=root group=root mode=644
+  when: kafka_version  == '0.8.2.2'
+  tags: log4j
+
+- name: Copies kafka log4j_new.properties
+  copy: src=log4j_new.properties dest=/opt/kafka/config/log4j.properties owner=root group=root mode=644
```
snopoke, line 62: this will need to use the `template` module to resolve the vars

src/commcare_cloud/ansible/roles/kafka/files/log4j_new.properties
satyaakam: This variable is of kafka

src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
satyaakam: no we are not using template for the log4j.properties file

src/commcare_cloud/ansible/roles/kafka/files/log4j_new.properties
```diff
@@ -0,0 +1,74 @@
+log4j.rootLogger=INFO, stdout, kafkaAppender
```
snopoke, line 1: nit: consider renaming this file to `log4j_2.4.1.properties`

src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
```diff
@@ -117,3 +117,9 @@ zookeeper.connect={% for host in groups['zookeeper'] %}{{ host }}:{{ zookeeper_c
 
 # Timeout in ms for connecting to zookeeper
 zookeeper.connection.timeout.ms=1000000
+{% if kafka_version == "2.4.1" %}
+inter.broker.protocol.version=2.4
+log.message.format.version=2.4
+listeners=PLAINTEXT://:9092
+advertised_listener=PLAINTEXT://{{inventory_hostname}}:9092
+{% endif %}
```
snopoke, line 125: can we do a similar thing for `advertised.host.name={{ inventory_hostname }}` on line [34](https://github.com/dimagi/commcare-cloud/pull/3786/files#diff-c4593a9a7d190dc9851e42057de19cdeR34) (only check for the older version).

Also, are there any other legacy settings that should be removed?

src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
```diff
@@ -117,3 +117,9 @@ zookeeper.connect={% for host in groups['zookeeper'] %}{{ host }}:{{ zookeeper_c
 
 # Timeout in ms for connecting to zookeeper
 zookeeper.connection.timeout.ms=1000000
+{% if kafka_version == "2.4.1" %}
+inter.broker.protocol.version=2.4
+log.message.format.version=2.4
```
snopoke, line 122: I think these need to be configurable using a variable since initially they would be set to `0.8.2.2` and only changed to 2.4 after the upgrade is complete.

satyaakam: i don't know how we are going to check the health of the system with the old value and how to trigger adding the new value

src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
```diff
@@ -53,6 +53,23 @@
...
+    register: version_2.4_infile
+- name: Ensure old version
+  lineinfile:
+    path: /etc/kafka/config/server.properties
+    regexp: '^inter.broker.protocol.versioni=0.8.2.2'
+    state: absent
+    check_mode: yes
+    changed_when: false
+    register: version_0.8.2.2_infile
+
```
satyaakam, line 72: i tried checking the values in template , it throws errors the logic i am trying is  
version_2.4_infile.found will retrun true if it matches and version_0.8.2.2_infile.found will retrun true if it's matches

src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
```diff
@@ -117,3 +120,9 @@ zookeeper.connect={% for host in groups['zookeeper'] %}{{ host }}:{{ zookeeper_c
 
 # Timeout in ms for connecting to zookeeper
 zookeeper.connection.timeout.ms=1000000
+{% if kafka_version == "2.4.1" %}
+inter.broker.protocol.version=0.8.2.2
```
snopoke, line 124: There are 2 options here.
### 1. make the config completely configurable:

```jinja
{% for config, value in kafka_extra_config.items() %}
{{config}}={{value}}
{% endfor %}
```
Then in the config file you can have:
```yaml
kafka_extra_config:
    inter.broker.protocol.version: 0.8.2.2
    log.message.format.version: 0.8.2.2
    etc
```

### 2. The other option is to have a separate var for each:
```jinja
inter.broker.protocol.version={{kafka_inter_broker_protocol_version}}
log.message.format.version={{kafka_log_message_format_version}}
```

src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
```diff
@@ -53,8 +53,32 @@
...
+    
 - name: Copies kafka log4j.properties
   copy: src=log4j.properties dest=/opt/kafka/config owner=root group=root mode=644
+  when: kafka_version  == '0.8.2.2'
+  tags: log4j
+
+- name: Copies kafka log4j_2.4.1.properties
+  copy: src=log4j_2.4.1.properties dest=/opt/kafka/config/log4j.properties owner=root group=root mode=644
+  when: kafka_version  == '2.4.1'
   tags: log4j
```
snopoke, line 82: If we rename the old file then we can have a single task which uses the kafka version to select the correct filename
```suggestion
- name: Copies kafka log4j.properties
  copy: 
    src: "log4j_{kafak_version}.properties" 
    dest: /opt/kafka/config
    owner: root
    group: root
    mode: 644
  tags: log4j
```

src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
```diff
@@ -117,3 +120,9 @@ zookeeper.connect={% for host in groups['zookeeper'] %}{{ host }}:{{ zookeeper_c
...
+log.message.format.version=0.8.2.2
```
snopoke, line 125: what's the plan for changing these to `2.4` once the migration is complete? I still this this needs some work. I'll try take a look today.

satyaakam: last we looked at it @rohit-dimagi said he could not find a anyway to do it in Ansible , he suggested the second step to be a manual one

docs/services/kafka/upgrade.md
```diff
@@ -0,0 +1,50 @@
+# Upgrading Kafka
```
snopoke, line 1: Some things to check here:
1. Will running the `deploy_kafka.yml` playbook restart Kafka when it updates the config?
  - My read of the playbook is that it won't but I'm not 100% sure.
  - For the upgrade process we need to restart Kafka after the configs and binaries have changed

2. Not sure if step 5 & 6 need to be separate or if they can be combined.

src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
```diff
@@ -117,3 +123,6 @@ zookeeper.connect={% for host in groups['zookeeper'] %}{{ host }}:{{ zookeeper_c
 
 # Timeout in ms for connecting to zookeeper
 zookeeper.connection.timeout.ms=1000000
+
+inter.broker.protocol.version={{kafka_inter_broker_protocol_version}}
+log.message.format.version={{kafka_log_message_format_version}}
```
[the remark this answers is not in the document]
snopoke: it will start at `0.8.2.2` until you update the params in public.yml. If you look at the upgrade docs you'll see that the params are only changed after the upgarde.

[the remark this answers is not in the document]
rohit-dimagi: I see. thanks

docs/services/kafka/upgrade.md
rohit-dimagi: it doesn't do restart , we should add it that seems like right thing to do ?

rohit-dimagi:
@snopoke  can we merge it ? i'm planning to do upgrade on icds staging.

docs/services/kafka/upgrade.md
snopoke: Yes, I think we should add in handlers to restart the services

PR #3791, opened 2020-04-28
Wipe Postgres DBs, Kafka topics

kaapstorm:
Context: [SC-514](https://dimagi-dev.atlassian.net/browse/SC-514)

Two short playbooks to:
* Wipe Postgres databases created by `set_up_dbs.yml`
* Wipe Kafka topics created by the `create_kafka_topics` Django management command

Both commands require 

    wipe_environment_enabled: True

to be set in in public.yml. This is a precaution against running these playbooks in the wrong environment.

##### ENVIRONMENTS AFFECTED
Only environments that have `wipe_environment_enabled: True` in public.yml. The `wipe_postgres.yml` playbook does not work with remote DBs like Amazon RDS.

### DIFF
--- a/src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
+++ b/src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
@@ -19,6 +19,13 @@
 # The id of the broker. This must be set to a unique integer for each broker.
 broker.id={{ kafka_broker_id }}
 
+{% if allow_topic_deletion|default(False) %}
+################################## Bad Ideas ##################################
+
+# Allow topics to be deleted
+delete.topic.enable=true
+
+{% endif %}
 ############################# Socket Server Settings #############################
 
 # The port the socket server listens on
--- a/src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
+++ b/src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
@@ -19,7 +19,7 @@
 # The id of the broker. This must be set to a unique integer for each broker.
 broker.id={{ kafka_broker_id }}
 
-{% if allow_topic_deletion|default(False) %}
+{% if wipe_environment_enabled|default(False) %}
 ################################## Bad Ideas ##################################
 
 # Allow topics to be deleted
--- a/src/commcare_cloud/ansible/wipe_kafka.yml
+++ b/src/commcare_cloud/ansible/wipe_kafka.yml
@@ -8,9 +8,11 @@
         that: "{{ wipe_environment_enabled|default(False) }} == True"
         fail_msg: 'This playbook will delete all data. To continue, set
           "wipe_environment_enabled: True" in public.yml. Take care to unset
-          "wipe_environment_enabled" when the environment setup is complete.'
+          "wipe_environment_enabled" and run the `deploy_kafka.yml` playbook
+          again when the environment setup is complete.'
 
     - name: Import kafka role for `kafka_conf_dir` value
+      # ... and to update server.properties
       import_role:
         name: kafka
 
@@ -39,8 +41,6 @@
       template:
         src: roles/kafka/templates/server.properties.j2
         dest: "{{ kafka_conf_dir }}/server.properties"
-      vars:
-        allow_topic_deletion: True
 
     - name: Restart Kafka
       # `kafka-server` service does not support reload
@@ -54,15 +54,3 @@
         --delete
         --zookeeper localhost:{{ zookeeper_client_port }}
         --topic {{ item }}"
-
-    - name: Prevent Kafka server from deleting topics
-      template:
-        src: roles/kafka/templates/server.properties.j2
-        dest: "{{ kafka_conf_dir }}/server.properties"
-      vars:
-        allow_topic_deletion: False
-
-    - name: Restart Kafka
-      service:
-        name: kafka-server
-        state: restarted
--- a/src/commcare_cloud/ansible/wipe_kafka.yml
+++ b/src/commcare_cloud/ansible/wipe_kafka.yml
@@ -20,10 +20,8 @@
           "wipe_environment_enabled" and run the `deploy_kafka.yml` playbook
           again when the environment setup is complete.'
 
-    - name: Import kafka role for `kafka_conf_dir` value
-      # ... and to update server.properties
-      import_role:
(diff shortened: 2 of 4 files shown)


src/commcare_cloud/ansible/wipe_postgres.yml
```diff
@@ -0,0 +1,31 @@
...
+      vars:
+        ansible_ssh_pipelining: true
+      with_items: "{{ postgresql_dbs.all }}"
+      # Only delete the databases we created (standby DB is not created, and not deleted)
+      when: item.create and ((item.host == postgresql_host) or is_monolith|bool)
+      postgresql_db:
+        name: "{{ item.name }}"
+        state: absent
+        port: "{{ postgresql_port }}"
+        login_host: 127.0.0.1
```
esoergel, line 29: I'm guessing this should be variable, something like
https://github.com/dimagi/commcare-cloud/blob/04725527f60167847eb88f3c990e8b51e42c3d38/src/commcare_cloud/environment/schemas/postgresql.py#L162-L166

kaapstorm: I thought so too, and I had to iterate on this a lot before I managed to get it to work. `environment.groups['postgresql'][0]` is going to be the external IP address (the address of `eth0` or `enp0s1` or whatever). But using that address doesn't work, because it's not the address in `pg_hba.conf`. The address you need is hardcoded here:
https://github.com/dimagi/commcare-cloud/blob/587843fa1c8f3cd190b59183190f870f4af872c5/src/commcare_cloud/ansible/roles/postgresql_base/templates/pg_hba.conf.j2#L33

It wasn't intuitive to me, so I think I'll add a comment to help the next person who finds this.

esoergel: sounds good - just wanted to check that it wasn't an oversight

kaapstorm: :+1: ... well, I'm sure there is something here I am missing. When these DBs are created, these values are omitted:
https://github.com/dimagi/commcare-cloud/blob/bcc391b57986a92a9f768cf5bc800dac2c980228/src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml#L94-L96
I think this is because the command is using sockets, but I could not get that to work. I gave up, and went with `127.0.0.1` instead.

src/commcare_cloud/ansible/wipe_kafka.yml
```diff
@@ -0,0 +1,56 @@
...
+  become: true
+  hosts: zookeeper  # Doesn't need to run on all Kafka hosts
+  tasks:
+    - name: Check wipe_environment_enabled has been set to True
+      assert:
+        that: "{{ wipe_environment_enabled|default(False) }} == True"
+        fail_msg: 'This playbook will delete all data. To continue, set
+          "wipe_environment_enabled: True" in public.yml. Take care to unset
+          "wipe_environment_enabled" and run the `deploy_kafka.yml` playbook
+          again when the environment setup is complete.'
```
snopoke, line 12: as an extra precaution you could add a prompt to confirm: https://docs.ansible.com/ansible/latest/user_guide/playbooks_prompts.html

src/commcare_cloud/ansible/wipe_kafka.yml
```diff
@@ -0,0 +1,56 @@
...
+      service:
+        name: kafka-server
+        state: restarted
+
+    - name: Delete topics
+      loop: "{{ topics_result.out.split() }}"
+      command: "/opt/kafka/bin/kafka-topics.sh
+        --delete
+        --zookeeper localhost:{{ zookeeper_client_port }}
+        --topic {{ item }}"
```
snopoke, line 56: should you also update the server properties again to remove the `delete.topic.enable` setting?

src/commcare_cloud/ansible/wipe_kafka.yml
```diff
@@ -0,0 +1,56 @@
...
+      #    {
+      #        "changed": false,
+      #        "failed": false,
+      #        "out": "case\ncase-sql\ncommcare-user\ndomain\nform\nform-sql\n"
+      #               "group\nledger\nmeta\nsms\nweb-user\napp\nlocation\n"
+      #               "synclog-sql\n",
+      #        # ...
+      #    }
+
+    - name: Allow Kafka server to delete topics
```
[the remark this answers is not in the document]
kaapstorm: Yes! Thanks for catching that. I thought I'd dropped this task already.

src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
```diff
@@ -19,6 +19,13 @@
 # The id of the broker. This must be set to a unique integer for each broker.
 broker.id={{ kafka_broker_id }}
 
+{% if wipe_environment_enabled|default(False) %}
+################################## Bad Ideas ##################################
+
+# Allow topics to be deleted
+delete.topic.enable=true
```
snopoke, line 26: I was thinking about this and instead of putting this in the config you could use the `lineinfile` module to add / remove it which would be much easier than running the whole kafka role

src/commcare_cloud/ansible/wipe_kafka.yml
kaapstorm: If you do that, when zookeeper gets around to fulfilling your request to delete the topics, it sees it's not allowed to, and does nothing ... so the topics never get deleted. (This had me stumped for a while.) It seems you have to leave the server properties as-is until the topics are deleted (which, in my test environment, was done by the time I could check as soon as the playbook had finished).

src/commcare_cloud/ansible/roles/kafka/templates/server.properties.j2
kaapstorm: Thank you! I'll do that.

kaapstorm:
Labelling "do not merge" because the `./manage.py migrate` command does not create all of the tables.

kaapstorm:
The problem was that `./manage.py migrate` is the wrong command to rebuild the databases, because it rebuilds the default database. `./manage.py migrate_multi` is the right command. It is the equivalent of

    $ for db in default p1 p2 proxy synclogs ; do ./manage.py migrate --database $db ; done

PR #3792, opened 2020-04-29
Create env file for prometheus

rohit-dimagi:
* https://dimagi-dev.atlassian.net/browse/IIO-725
##### ENVIRONMENTS AFFECTED

ALL

### DIFF
--- a/src/commcare_cloud/environment/main.py
+++ b/src/commcare_cloud/environment/main.py
@@ -15,7 +15,6 @@ from ansible.vars.manager import VariableManager
 from ansible_vault import Vault
 from memoized import memoized, memoized_property
 from six.moves import shlex_quote
-from jsonobject import JsonObject
 
 from commcare_cloud.environment.constants import constants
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/environment/main.py
```diff
@@ -223,6 +224,16 @@ def proxy_config(self):
         proxy_config.check()
         return proxy_config
 
+    @memoized_property
+    def prometheus_config(self):
+        try:
+            with open(self.paths.prometheus_yml) as f:
+                prometheus_json = yaml.safe_load(f)
+        except IOError:
+            prometheus_json = {}
```
snopoke, line 233: If you make any of the fields required then you need can't call `PrometheusConfig.wrap({})` (empty dict). Instead you could just return an empty JsonObject (assuming the only call to it is `.to_json()`
```suggestion
            return JsonObject()
```
This will need to be imported `from jsonobject import JsonObject`

PR #3797, opened 2020-04-30
Volume backup for Elasticsearch on India and Production Enviornment

sanjay2916:
Volume backup for Elasticsearch on India and Production environment

### DIFF
--- a/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
+++ b/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
@@ -74,6 +74,7 @@ resource "aws_dlm_lifecycle_policy" "data_volume_backups" {
 
     target_tags = {
       GroupDetail = "couchdb2:data"
+      Group = "elasticsearch"
     }
   }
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
```diff
@@ -74,6 +74,7 @@ resource "aws_dlm_lifecycle_policy" "data_volume_backups" {
 
     target_tags = {
       GroupDetail = "couchdb2:data"
+      Group = "elasticsearch"
```
dannyroberts, line 77: Just checking, does terraform / AWS interpret this as "if `GroupDetail = "couchdb2:data"` AND `Group = "elasticsearch"`" or "if `GroupDetail = "couchdb2:data"` OR `Group = "elasticsearch"`"

dannyroberts: confirmed that the condition here matches volumes that have either tag

PR #3803, opened 2020-05-04
Sc/regenerate ssh host keys

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/IIO-825

##### ENVIRONMENTS AFFECTED
icds-staging and ICDS


rohit-dimagi:
i think proper place to have this should be in here. https://github.com/dimagi/commcare-cloud/tree/master/scripts/tcl

shyamkumarlchauhan:
> i think proper place to have this should be in here. https://github.com/dimagi/commcare-cloud/tree/master/scripts/tcl

agree for new VMs but for existing VMs to make the changes in the environment , shouldn't this be a fair choice ?

shyamkumarlchauhan:
> > i think proper place to have this should be in here. https://github.com/dimagi/commcare-cloud/tree/master/scripts/tcl
> 
> agree for new VMs but for existing VMs to make the changes in the environment , shouldn't this be a fair choice ?

Just tried to test it if this can be applied for the new VMs using reset.yml until the ansible user is actually deployed on to the servers . seems like this is not directly achievable until ansible user is deployed 

"msg": "unlinking failed: [Errno 13] Permission denied: '/etc/ssh/ssh_host_rsa_key' ", "owner": "root", "path": "/etc/ssh/ssh_host_rsa_key""

PR #3807, opened 2020-05-05
A more extreme approach to wiping persistent data

kaapstorm:
I have stopped trying to resolve the problems with my previous attempt at wiping persistent data from an environment. (cf. PR #3791)

Instead of just dropping databases, etc., this branch takes a more extreme approach. It uses the inverse of Ansible tasks to install persistent storage. It uninstalls the software and deletes the data directories. My hope is that this will make it simpler to rebuild an environment after it has been wiped. Considering I don't know exactly why my previous fails, I'm not sure this will avoid the same problem. But, you know, 

> *If at first you don't succeed, use a bigger hammer.*

A potential problem with this approach is that it is more destructive, and will remove non-CommCare data, if a host is being used for more than just CommCare.

All suggestions welcome.

### DIFF
--- a/src/commcare_cloud/ansible/purge_elasticsearch.yml
+++ b/src/commcare_cloud/ansible/purge_elasticsearch.yml
@@ -1,0 +1,39 @@
+---
+- name: Uninstall Elasticsearch and remove data directory
+  hosts: elasticsearch
+  become: true
+  vars_prompt:
+    - name: confirm_purge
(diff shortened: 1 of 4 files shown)


kaapstorm:
Yeah, I think this is simpler and faster than a full reprovision. ... But now that I've found where I was going wrong with rebuilding, I'm going to close this PR, and rather go with the previous PR, #3791 --

PR #3809, opened 2020-05-06
Pull out SaaS envs

dannyroberts:
https://dimagi-dev.atlassian.net/browse/CDTW-1

This moves SaaS environment config over to its own repo: https://github.com/dimagi/saas-commcare-environments. (If you're unable to access the link and have ever needed to deploy to staging, india, production, or swiss, ping me and I'll add you.)

This is a proof of concept; more work is needed before we'd be ready to go:
- people would have to check this repo out locally and change their `COMMCARE_CLOUD_ENVIRONMENTS` (in `~/.commcare-cloud/load_config.sh` if set up with `manage-commcare-cloud`), so we need documentation and/or transition tools for that
- we'd have to configure the control machine for these envs to check out this repo and set `COMMCARE_CLOUD_ENVIRONMENTS` to it, and have the update-code scripts update it as well

### Why just SaaS?
- If this went well we'd encourage the other divisions to adopt something similar.
- Each division's team is going to have their own needs, and I think we should be moving to a place where each can independently manage their configurations without overlap
- For example, to transition from the current system to the one I'm proposing, I can drive that on SaaS and work with the other SaaS devs to tailor it to our individual and team needs, but I can't drive the change on ICDS or Solutions
- To emphasize the separateness of ownership and to limit the ease of copying passwords across divisions
- To restrict access to only those who need it to do their jobs (but restricting it no further than that!)

##### ENVIRONMENTS AFFECTED
production, india, staging, swiss


dannyroberts:
Oh and anyone feel free to leave 👍 👎 emoji reactions on the description of this PR to help me gauge reaction even if you have no comment. For a limited time only I will consider 👎emoji reaction with no explanation unambiguously not rude 😄

orangejenny:
I'm not going to read this in detail right now, but I'm in favor of this general approach.

mkangia:
Few questions @dannyroberts 
1. You wrote
> manage their configurations without overlap

What was the overlap as of now?
2. Currently also the files for an environment in their own folders, how will this make things better?
3. I don't imagine anyone not needing access to sass-environments, at least at Dimagi, can you give an example where this would be useful?

millerdev:
I'm overall 👍 on this, but as with most security-minded measures, it will introduce some inconvenience. The first thing that comes to mind is that it will make it harder to quickly see if a variable is being used by any environment. On the other hand, that may be a good thing since it will push us to adopt more formal deprecation workflows rather than assuming that since none of the envs in this repo are using a variable then no one is using it anywhere (which is not necessarily true).

TLDR: this may make refactoring harder and/or motivate us to implement new processes.

dannyroberts:
Hey @mkangia good questions. Overall the goal is for SaaS to have an environment config repo over which we're solely responsible and not accountable to other divisions for, so we can move quickly without worrying about how it affects other divisions. I'd encourage other divisions to take that approach as well. We'd all then still be accountable to other divisions for changes we make to this shared codebase.

> What was the overlap as of now?
I think it's hard to see now, because we've so far tried to fit everything together in one place, but an example is how SaaS currently manages aws-specific automation. Right now it's in a kind of limbo state where it's kind of roughly generic, but going forward we need to be able to manage AWS-specific configuration and automation in a way where we aren't accountable to other divisions or third parties for our changes. How exactly we go through that is an open question (e.g. write our own commcare-cloud "plugins"/commands in this repo that aren't part of the main repo?) but we'll need a place to manage our own configurations that are compatible with main-repo commcare-cloud but may include other things as well, such as getting into the details about how our AWS accounts are configured, etc. It gives us clear boundary between a shared code asset and a code asset that belongs entirely to SaaS, where we also have full control over access (like making the repo private) and where we're free to make changes as we please without considering the effect on other divisions at Dimagi or third parties.

> 2. Currently also the files for an environment in their own folders, how will this make things better?

This will let us do things like have global saas configuration that's not part of a specific environment (like specifying our AWS account structure as configuration, maybe our datadog configurations) and put other supporting scripts here in ways that would be difficult to shoehorn into the current shared environments directory and/or which don't make sense to make part of main-repo commcare-cloud.

> 3. I don't imagine anyone not needing access to sass-environments, at least at Dimagi, can you give an example where this would be useful?
Yeah it may be that the best thing for SaaS is to give all developers at Dimagi access to the repo, but I think that's mostly up to SaaS to make the right decisions around. I think there's no _a priori_ reason that SaaS as a division is required to give write access to everyone in every division forever regardless of team size, etc, and that applies in the reverse as well. (There's no reason ICDS needs to give me write access to ICDS environment configs, for example; it's up to the ICDS division to decide that to meet whatever goals they might have.)

biyeun:
Also In favor of this. 👍 I want to second @millerdev's question about not being able to easily tell if a variable is being used by a different division. Perhaps we should form some better namespacing / consistency guidelines?

PR #3817, opened 2020-05-13
Increase prod formplayer_memory to ~50% of available memory

dannyroberts:
Formplayer machines are now 32G RAM, so setting formplayer max memory to 15G (just under 50%). The reason for not going closer to 100% is because formplayer does appear to benefit from using RAM for a fair amount of os caching.

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
I'll be rolling this out late tonight if it doesn't happen before then

PR #3819, opened 2020-05-14
AWS SSO CLI credentials

dannyroberts:
We're switching over to using AWS SSO for access management across our AWS accounts. There's going to be a bit of pain here, but I'll be careful to create accounts for everyone that's used console or CLI access in the last 90 days to minimize short term risk; on top of that, reach out to me if you need access that hasn't been provided. I'll also need to update our related docs on confluence. Overall, the new setup is much simpler for securely managing access to all of the accounts in our AWS Organization.

##### ENVIRONMENTS AFFECTED
production, india, staging

### DIFF
--- a/src/commcare_cloud/commands/terraform/aws.py
+++ b/src/commcare_cloud/commands/terraform/aws.py
@@ -294,13 +294,17 @@ class AwsSignIn(CommandBase):
 
 AWS_CREDENTIALS_PATH = os.path.expanduser('~/.aws/credentials')
 AWS_CONFIG_PATH = os.path.expanduser('~/.aws/config')
+AWS_DOT_DIR = os.path.expanduser('~/.aws/')
 AWS_SSO_CACHE_DIR = os.path.expanduser('~/.aws/sso/cache/')
 AWS_CLI_CACHE_DIR = os.path.expanduser('~/.aws/cli/cache/')
 
 
 @memoized
 def aws_sign_in(environment, duration_minutes=DEFAULT_SIGN_IN_DURATION_MINUTES, force_new=False):
+    _ensure_all_dirs(AWS_DOT_DIR)
     if environment.aws_config.credential_style == 'sso':
+        for path in (AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
+            _ensure_all_dirs(path)
         return _aws_sign_in_with_sso(environment)
     else:
         return _aws_sign_in_with_iam(environment.terraform_config.aws_profile, duration_minutes=duration_minutes,
@@ -549,3 +553,8 @@ def _iter_files_in_dir(directory):
         filepath = os.path.join(directory, filename)
         if os.path.isfile(filepath):
             yield filepath
+
+
+def _ensure_all_dirs(path, mode=0700):
+    if not os.path.exists(path):
+        os.makedirs(path, mode=mode)
--- a/src/commcare_cloud/commands/terraform/aws.py
+++ b/src/commcare_cloud/commands/terraform/aws.py
@@ -355,17 +355,18 @@ def _sync_sso_to_v1_credentials(aws_session_profile):
                 data = json.load(f)
             except JSONDecodeError:
                 continue
-            else:
-                if data.get('ProviderType') != 'sso':
-                    continue
-                credentials = data['Credentials']
-                comparison = _get_caller_identity({
-                    'AWS_ACCESS_KEY_ID': credentials.get('AccessKeyId'),
-                    'AWS_SECRET_ACCESS_KEY': credentials.get('SecretAccessKey'),
-                    'AWS_SESSION_TOKEN': credentials.get('SessionToken'),
-                })
-                if comparison == caller_identity:
-                    break
+
+            if data.get('ProviderType') != 'sso':
+                continue
+
+            credentials = data['Credentials']
+            comparison = _get_caller_identity({
+                'AWS_ACCESS_KEY_ID': credentials.get('AccessKeyId'),
+                'AWS_SECRET_ACCESS_KEY': credentials.get('SecretAccessKey'),
+                'AWS_SESSION_TOKEN': credentials.get('SessionToken'),
+            })
+            if comparison == caller_identity:
+                break
     else:
         raise ValueError((
             "Unable to find cached credentials immediately after SSO. "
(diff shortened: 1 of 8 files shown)


src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -287,13 +289,108 @@ class AwsSignIn(CommandBase):
...
+                    if data.get('ProviderType') != 'sso':
+                        continue
+                    credentials = data['Credentials']
+                    comparison = _get_caller_identity({
+                        'AWS_ACCESS_KEY_ID': credentials.get('AccessKeyId'),
+                        'AWS_SECRET_ACCESS_KEY': credentials.get('SecretAccessKey'),
+                        'AWS_SESSION_TOKEN': credentials.get('SessionToken'),
+                    })
+                    if comparison == caller_identity:
+                        break
```
millerdev, line 357: Consider moving this out to a helper function since it's a rather big "insane workaround."

Also, the `else` (`raise ValueError...`) clause below will not be executed if this `for`-loop exits without finding credentials. Is that a bug?

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -287,13 +289,108 @@ class AwsSignIn(CommandBase):
...
+            raise ValueError((
+                "SSO profile mis-configured and cannot fix automatically. "
+                "Edit [profile {}] in ~/.aws/config manually;"
+                "to start over, remove it from the file manually and try again.").format(aws_session_profile))
+        # Until this is built in to aws, we need this insane workaround to get backwards compatibility
+        # with the credential format that terraform uses
+        cli_cache_dir = os.path.expanduser('~/.aws/cli/cache/')
+
+        for filename in os.listdir(cli_cache_dir):
+            with open(os.path.join(cli_cache_dir, filename)) as f:
```
millerdev, line 342: This assumes that every item in `cli_cache_dir` can be opened as a file, but I think this will fail if there are directories in there. What do you think of using `os.path.isfile()` to check before opening?

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -287,13 +289,108 @@ class AwsSignIn(CommandBase):
...
+                                       silent=True)
+    except subprocess.CalledProcessError as e:
+        # 255: Not logged in / missing credentials
+        # 254: "The security token included in the request is expired"
+        if e.returncode not in (255, 254):
+            raise
+        # this means we're not signed in
+        return None
+    else:
+        return caller_identity
```
millerdev, line 389: nit
```suggestion
    return caller_identity
```

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -371,7 +468,61 @@ def _write_credentials_to_aws_credentials(
...
+
+    if section not in config.sections():
+        return False
+
+    try:
+        config.get(section, 'sso_start_url')
+    except configparser.NoOptionError:
+        return False
+    else:
+        return True
```
millerdev, line 485: nit
```suggestion
    return True
```

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -371,7 +468,61 @@ def _write_credentials_to_aws_credentials(
...
+    for filename in os.listdir(aws_sso_cache_path):
+        try:
+            with open(os.path.join(aws_sso_cache_path, filename), 'r') as f:
+                contents = json.load(f)
+        except JSONDecodeError:
+            continue
+        else:
+            if 'startUrl' in contents and 'expiresAt' in contents:
+                expiration = datetime.strptime(contents['expiresAt'], "%Y-%m-%dT%H:%M:%SUTC")
+                return datetime.utcnow() < expiration
```
millerdev, line 521: nit
```suggestion
        if 'startUrl' in contents and 'expiresAt' in contents:
            expiration = datetime.strptime(contents['expiresAt'], "%Y-%m-%dT%H:%M:%SUTC")
            return datetime.utcnow() < expiration
```

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: Yeah good point I'll definitely move this out to a helper function. Have to admit I did this a bit hastily so after applying your feedback I'll also do another quick pass myself at tidying up / tying up loose ends

dannyroberts: I don't think I understand the comment about the `else:` (`raise ValueError(...)`) below. The way I'm intending to use it is as part of a `for/else`, where the `else` is called if the end of the `for` loop is reached without hitting a `break`.

I found a way to contrive a test of this (stuck a pause in after the first call to `_get_caller_identity` and then snuck in and manually deleted all the files in `~/.aws/cli/cache/`), and it raised this error the way I expected it to

millerdev: +1 for the test. Yay!

Oh, I misunderstood the code. I thought the `else` was the other branch of `if not _has_valid_v1_session_credentials(aws_session_profile):`. Now I see it's with the `for`-loop. All good.

dannyroberts: It was just a manual test, so I think I don't deserve the "yay" 😬, but it was just enough to make sure that my understanding of the code logic was on the right track

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: Thanks yeah I do think it's clearer that way

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: 0a0a119

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: 9c4d325

dannyroberts:
OK I've been using this locally for the past week and it's now gotten to the point where it works every time. I've also tested that it works with a totally clean ~/.aws directory.

I've set up SSO accounts for people now. Follow the steps in the email you received now, and then the next time you need to use it from the command line, it'll pop open your browser and make you prove you're you, and then use that to generate local keys for you in the background that will last for a set amount of time (I think 1h), and which you can use for any of the three envs.

dannyroberts:
@dimagi/team-commcare-hq Here are the new instructions for using your Single Sign On users: https://confluence.dimagi.com/display/internal/Using+AWS+credentials+with+commcare-cloud

PR #3820, opened 2020-05-15
updated nginx configuration

shyamkumarlchauhan:
Updated nginx configuration

##### ENVIRONMENTS AFFECTED
Icds


dannyroberts:
This looks reasonable. Can you explain what motivated the change and what the effect is?

dannyroberts:
noticed that this affects the couchdb2_proxy as well which is probably unintentional? Not sure what the expected effect is and whether it's an issue. (I noticed that it shows up as an unapplied diff when I run `cchq production ansible-playbook deploy_couchdb2.yml --tags=proxy --limit=couchdb2_proxy`.)

PR #3822, opened 2020-05-17
Set up Global Accelerator => Application Load Balancer (for WAF)

dannyroberts:
This encodes the POC I did on staging to insert a Web Application Firewall between our machines and the internet, so traffic is filtered before it even gets to our nginx proxy.

This applies cleanly on staging, and I've also rolled it out on production. (Have not yet done anything on india, but likely will there too.)

##### ENVIRONMENTS AFFECTED
production, staging, india


dannyroberts:
(@ctsims FYI)

ctsims:
Woah, super fast turnaround on all of this. Thanks for pushing it forward.

PR #3837, opened 2020-05-20
Set prod ssl ciphers to the equivalent of ELBSecurityPolicy-FS-1-2-Res-2019-08

dannyroberts:
in https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html\#describe-ssl-policies

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
Worked on this with @ctsims

PR #3838, opened 2020-05-20
Make the aws ssl policy configurable

dannyroberts:
And set staging to most restrictive ssl policy, `ELBSecurityPolicy-FS-1-2-Res-2019-08` (https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies)

##### ENVIRONMENTS AFFECTED
staging


dannyroberts:
worked on this with @ctsims

PR #3839, opened 2020-05-21
Updating the npm install error - resolution steps

prem-fissionhq:
Recently, we ran into an issue with "npm_install". Phantomjs package was unable to download. therefore I had to download it manually and set the path.  Later, issue got resolved.


dannyroberts:
Thanks!

PR #3840, opened 2020-05-21
Making EBS encryption default value as True

sanjay2916:
Making EBS encryption default value as True


src/commcare_cloud/environment/schemas/terraform.py
```diff
@@ -53,7 +53,7 @@ class ServerConfig(jsonobject.JsonObject):
     network_tier = jsonobject.StringProperty(choices=['app-private', 'public', 'db-private'])
     az = jsonobject.StringProperty()
     volume_size = jsonobject.IntegerProperty(default=20)
-    volume_encrypted = jsonobject.BooleanProperty(default=False, required=True)
+    volume_encrypted = jsonobject.BooleanProperty(default=True, required=False)
```
dannyroberts, line 56: I would keep this at `required=True` which means "don't allow `null`/`None` as a value".

dannyroberts, line 56:
```suggestion
    volume_encrypted = jsonobject.BooleanProperty(default=True, required=True)
```

PR #3841, opened 2020-05-21
Remove IAM users for credential_style = sso

dannyroberts:
Already rolled out

##### ENVIRONMENTS AFFECTED
production, india, staging (only affects how devs access AWS Console/API)


dannyroberts:
Yeah I'm not really in touch with what ICDS is doing with AWS and didn't want to go deleting things they might be using, so instead I made "IAM credential setup" vs. "SSO credential setup" a n environment-level option, specified in `environments/*/aws.yml` as `credential_style: iam | sso`.

We'll may end up deleting it altogether, but didn't want to block SaaS switching with that convo

PR #3842, opened 2020-05-21
Making Default encryption value True for Block devices 

sanjay2916:
Making Default encryption value True for Block devices

### DIFF
--- a/environments/india/terraform.yml
+++ b/environments/india/terraform.yml
@@ -133,6 +133,7 @@ servers:
     volume_encrypted: no
     block_device:
       volume_size: 270
+      volume_encrypted: no
     group: couchdb2
     os: bionic
   - server_name: "couch6-india"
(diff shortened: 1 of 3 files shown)


environments/india/terraform.yml
```diff
@@ -133,6 +133,7 @@ servers:
     volume_encrypted: no
     block_device:
       volume_size: 270
+      volume_encrypted: no
```
dannyroberts, line 136: These should say `encrypted: yes`
https://github.com/dimagi/commcare-cloud/blob/95cd9999cd5d8180e71c47c4011c29ea4b33e3b6/src/commcare_cloud/environment/schemas/terraform.py#L66-L68
Confusing naming though I'll give you that!

sanjay2916: @dannyroberts  these block devices for CouchDB servers are not EBS encrypted and  values we required "encrypted: no" but both yes and no values are giving the same result with no terraform diff

dannyroberts: Strange. I tried to reproduce that locally, but I saw the behavior I'd expect instead of what you described

PR #3847, opened 2020-05-24
Roll out 24h compaction on all dbs on production

dannyroberts:
Our CPU usage is never particularly high throughout the day, so I believe we can afford to do this.

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
(already rolled out)

PR #3849, opened 2020-05-25
Let x forwarded for pass through

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-10914
##### ENVIRONMENTS AFFECTED
production, staging, india

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -29,8 +29,6 @@ nginx_sites:
    proxy_max_temp_file_size: 5120m
    proxy_set_headers:
    - "Host $host"
-   - "X-Forwarded-For $remote_addr"
-   - "X-Forwarded-Protocol  $scheme"
    access_log: "{{ log_home }}/{{ deploy_env }}-timing.log timing"
    locations:
     - name: /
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -29,8 +29,6 @@ nginx_sites:
    proxy_max_temp_file_size: 5120m
    proxy_set_headers:
    - "Host $host"
-   - "X-Forwarded-For $remote_addr"
```
ctsims, line 32: My initial guess was that this was here for consistency in logging. It might be worth making sure Auditcare works fine on india and icds after tihs change

dannyroberts: OK I took some time to double check that HQ respects the `, `-separated list of IPs convention sometimes used in the `X-Forwarded-For` header, and it appears [we do](https://github.com/dimagi/commcare-hq/pull/27621).

So, I think what I'll do here is add this back as
```
- "X-Forwarded-For $proxy_add_x_forwarded_for"
- "X-Forwarded-Protocol  $scheme"
```
where [$proxy_add_x_forwarded_for](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#var_proxy_add_x_forwarded_for) is set to `$remote_addr` or, if x-forwarded-for exists already, `$http_x_forwarded_for, $remote_addr`.

For setups with nothing in front of the nginx proxy, this will result in a change to the nginx file, but no change in the behavior. For setups with another proxy in front of the nginx proxy (like production), this will result in `X-Forwarded-For` being set to "{client_ip}, {alb_ip}", which HQ should know how to handle

PR #3851, opened 2020-05-26
requesting valid root_email address and adding SSL.md Document link

prem-fissionhq:
Changes:
1. Requesting user to set valid root_email address. This is to validate the SSL certificates which we are provision using Letsencrypt without any issues.
2. Adding SSL.md instructions link in the new_environment.md so that the user can obtain Valid SSL certificates for the monolith env.

Please let me know your suggestions. Thanks.

### DIFF
--- a/docs/setup/new_environment.md
+++ b/docs/setup/new_environment.md
@@ -311,11 +311,29 @@ $ commcare-cloud monolith after-reboot all
 ```
 
 
-### In Order to set up valid SSL certificates 
-If you wish to set up valid SSL certificate please follow instructions provide in this link (../services/nginx/ssl.md)
+## Step 7: Setting set up valid SSL certificates 
+If you wish to set up valid SSL certificate please follow below instructions 
+
+    ### 1. Request a letsencrypt cert
+
+Run the playbook to request a letsencrypt cert:
+```bash
+cchq <env> ansible-playbook letsencrypt_cert.yml --skip-check
+```
+
+### 2. Update settings to take advantage of new certs
+
+In `proxy.yml`:
+- set `fake_ssl_cert` to `no`
+
+and deploy proxy again.
+
+```bash
+cchq <env> ansible-playbook deploy_proxy.yml
+```
 
 
-## Step 7: Getting started with CommCareHQ
+## Step 8: Getting started with CommCareHQ
 
 ### Make a user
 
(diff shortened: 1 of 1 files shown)


docs/setup/new_environment.md
```diff
@@ -83,7 +83,9 @@ This document will walk you through the process of setting up a new monolith ser
      - `ALLOWED_HOSTS`
      - `server_email`
      - `default_from_email`
-
+3. Add valid `root_email` address in public.yml file
+    - `public.yml`
+      - `root_email`
```
esoergel, line 88:
```suggestion
      - `root_email`
```
I don't think this variable needs special mention - it's just like the others being edited in `public.yml`

docs/setup/new_environment.md
```diff
@@ -310,6 +312,11 @@ If you ever reboot this machine, make sure to follow the [after reboot procedure
 $ commcare-cloud monolith after-reboot all
 ```
 
+
+### In Order to set up valid SSL certificates 
+If you wish to set up valid SSL certificate please follow instructions provide in this link (../services/nginx/ssl.md)
```
esoergel, line 317: Setting up SSL isn't optional, so I'd like to integrate it fully into this new environment process.  Instead of linking to another documentation page, can you paste in only the relevant commands here?  With the change in https://github.com/dimagi/sample-environment/pull/8, step 1 should be taken care of automatically with the initial server deploy, so I think all that's necessary is setting fake_ssl_cert to no and running the commands from step 2 and 3.

prem-fissionhq: Sure, Ethan. Will update that.

prem-fissionhq: @esoergel updated the document. Please review.

docs/setup/new_environment.md
```diff
@@ -310,7 +310,30 @@ If you ever reboot this machine, make sure to follow the [after reboot procedure
 $ commcare-cloud monolith after-reboot all
 ```
 
-## Step 7: Getting started with CommCareHQ
+
+## Step 7: Setting set up valid SSL certificates 
+If you wish to set up valid SSL certificate please follow below instructions 
+
+    ### 1. Request a letsencrypt cert
```
esoergel, line 317:
```suggestion
### 1. Request a letsencrypt cert
```

PR #3853, opened 2020-05-27
Fix --control by adding back -A ssh flag

dannyroberts:
when the server arg is e.g. control:0 instead of control.
Introduced by https://github.com/dimagi/commcare-cloud/commit/608612f4553f9d53543a1dd09dd2ece9c8b1df97

Made this change last week and forgot to PR


orangejenny:
thank you for fixing!

PR #3854, opened 2020-05-27
Add idle timeout to ALB of 10 minutes

dannyroberts:
##### ENVIRONMENTS AFFECTED
production/staging/(eventually india)


dannyroberts:
@millerdev thanks. I looked it up and it appears to be set to 20m, so I'm doing that here as well

PR #3856, opened 2020-05-28
List of Commcare application ports information

prem-fissionhq:


esoergel:
Hey Prem, could you move that table into the ["Prerequisites" section of the new environment doc](https://github.com/dimagi/commcare-cloud/blob/master/docs/setup/new_environment.md#prerequisites)?  Here's an example of a table in markdown:
https://github.com/dimagi/commcare-cloud/blob/master/docs/howto/transfer-domain.md#1-switch-mobile-devices-to-a-proxy-url

Then could you add those IDs and descriptions to the script and instruct the developer to run that script in that same section?

prem-fissionhq:
Hello Ethan,
As per your suggestions, I have Converted CSV to markdown table and included the link in the new_environment.md file. Also, updated the script to provide a port description for the ports which are not accessible. 
Please review and let me know if any changes are required. Thanks

PR #3863, opened 2020-06-02
Adding new formplayer on staging env for AWS EFS shared volume testing

sanjay2916:
Adding new formplayer on staging env for AWS EFS shared volume testing


environments/staging/terraform.yml
```diff
@@ -107,6 +107,18 @@ servers:
       encrypted: true
     group: formplayer
     os: bionic
+    
+  - server_name: "formplayer2-staging"
+    server_instance_type: "t3.large"
+    network_tier: "app-private"
+    az: "a"
+    volume_size: 50
+    volume_encrypted: no
```
dannyroberts, line 116: It's fine since this is staging, but in the future we want to always use `volume_encrypted: yes` even on the boot volume.

sanjay2916: sure Danny

PR #3873, opened 2020-06-10
Adding senty role and playbook

rohit-dimagi:
* https://dimagi-dev.atlassian.net/browse/IIO-868
##### ENVIRONMENTS AFFECTED

* ICDS


rohit-dimagi:
> Surely there are good external roles for setting up sentry?

There are roles for it most of then are for docker setup. ex- https://github.com/getsentry/onpremise/
and rest of them are for version 9. sentry have added a lot of component and dependencies in version 10 and couldn't find any external role for that.

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
```diff
@@ -243,3 +243,6 @@ nginx_sites:
        balancer: "{{ deploy_env }}_cas_formplayer/"
        proxy_next_upstream_tries: 1
        proxy_read_timeout: 900s
+     - name: /sentry
+       proxy_pass: "http://{{ groups['sentry'][0] }}:8080"
+       proxy_read_timeout: 900s
```
rohit-dimagi, line 248: This seems to be going to django and throwing 404

environments/icds-cas/postgresql.yml
```diff
@@ -153,6 +153,9 @@ dbs:
           value: 64
     - name: airflow
       django_migrate: False
+    - name: sentrydb
+      django_migrate: False
```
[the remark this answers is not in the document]
rohit-dimagi: In test schema it's set as required property.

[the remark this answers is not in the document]
snopoke: hmm, ok

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
snopoke: I played around with this a bit but because Sentry uses redirects it won't work without making sentry aware that it's running with a prefix and I'm not sure that's possible. So I think a subdomain is likely the best approach. We should then do the same for graphana etc.

rohit-dimagi: I got this gist while looking around and tried but it gets stuck with redirect loop . https://gist.github.com/beeftornado/945d5180ffd402f5cd8d

snopoke: hmm, seems like it should be possible:
*nginx*
```
location /sentry {
		proxy_pass http://0.0.0.0:9000;
		proxy_redirect	off;

		proxy_set_header	Host	$host;
		proxy_set_header	SCRIPT_NAME	/sentry;
		proxy_set_header	X-Real-IP	$remote_addr;
		proxy_set_header	X-Forwarded-For	$proxy_add_x_forwarded_for;
		proxy_set_header	X-Forwarded-Proto	$scheme;
	}
```

*sentry config*
```
SENTRY_URL_PREFIX = 'http://<TOP_LEVEL_DOMAIN>.com/sentry'  # No trailing slash!
FORCE_SCRIPT_NAME = '/sentry'

SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
USE_X_FORWARDED_HOST = True

SENTRY_WEB_OPTIONS = {
    ....
    'secure_scheme_headers': {'X-FORWARDED-PROTO': 'https'},
}
```

rohit-dimagi: Yeah i tried this but couldn't figured out how to fix redirect loop.

snopoke: I played around with this for quite a while but couldn't fix the redirect loop

environments/icds-cas/proxy.yml
```diff
@@ -11,5 +11,8 @@ special_sites:
   - cas_ssl
   - cas_sentry
 
+special_sites:
```
snopoke, line 14: this is the same key as above so would override it

rohit-dimagi: changed var

src/commcare_cloud/ansible/roles/nginx/vars/cas_ssl.yml
rohit-dimagi: Going with subdomain route.

PR #3880, opened 2020-06-12
Feature/hourly backups

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

- Using existing backup scripts to create backups even on an hourly basis. By default for every 3 hours, backups are taken and this can also +/- frequency using a parameter `nadir_frequency_hour`. 
-  Added Hours to the date variable (%Y_%m_%d_%H).
scripts are modified for PostgreSQL, CouchDB, and BlobDB. 

Please verify and let me know your suggestions.


esoergel:
Can you rebase this branch to remove the unrelated commits?

It looks like this will keep the hourly backups for just as long as it keeps the daily backups.  This will dramatically increase the disk usage.  They should not be kept as long.

Can you implement this in such a way that it doesn't affect existing environments?  We should not increase the backup frequency for existing environments without their intervention.  We also should not change the format of the backup file name, as that could in theory disrupt existing backup systems (for instance, duplicating files matching a certain pattern).  I'm not sure how plausible that latter scenario is, but I don't want to chance it.

prem-fissionhq:
Sure @esoergel. I will work on it. Thanks

prem-fissionhq:
@esoergel updated the hourly scripts with the changes. I have separated the hourly script for CouchDB/Postgresql/blobdb because adding conditions in the existing script was not working as expected in case of daily and weekly file removal. I'm working on improving the find command in the existing script. will keep you posted. 

Please review my changes and let me know your suggestions.

prem-fissionhq:
@esoergel As per your suggestions, I made the improvements to the scripts and according to my knowledge, the changes will not affect the existing backup patterns. 
- The hourly backup's persistence can also be handled by <component>_backup_hours Parameter.  Please review and let me know your suggestions. Thanks

src/commcare_cloud/ansible/roles/pg_backup/tasks/backup_plain_clean.yml
```diff
@@ -1,3 +1,15 @@
+- name: Remove old Hourly Cron job
+  become: true
+  cron:
+    name: "Backup postgres hourly"
+    job: "/usr/local/sbin/create_postgres_dump.sh hourly {{ postgresql_backup_hours * 60 }}"
+    minute: 0
+    hour: "{{ nadir_hour_frequency|default(3) }}"
```
esoergel, line 7: should these `hour` and `minute` values line up with those in `pg_backup/tasks/backup_plain.yml `?

src/commcare_cloud/ansible/roles/pg_backup/templates/plain/create_postgres_dump.sh.j2
```diff
@@ -46,12 +57,16 @@ if [[ $? -ne 0 ]] ; then
     echo "Postgres Backup : Tar Failed"
     exit 2
 else
+    if [[ postgres_backup_hourly && "$BACKUP_TYPE" == "hourly" ]]; then
+        find {{ postgresql_backup_dir }} -mmin "+${HOURS_TO_RETAIN_BACKUPS}" -name "postgres_${ENV}_${BACKUP_TYPE}_*" ! -path "*weekly*" ! -path "*daily*" -delete;
+        find {{ postgresql_backup_dir }} -mmin "+${HOURS_TO_RETAIN_BACKUPS}" -name "postgres_${BACKUP_TYPE}_*" ! -path "*weekly*" ! -path "*daily*" -delete;
```
esoergel, line 62: This filename format hasn't existed for hourly backups, so I don't think this line is necessary

src/commcare_cloud/ansible/roles/pg_backup/templates/plain/create_postgres_dump.sh.j2
```diff
@@ -46,12 +57,16 @@ if [[ $? -ne 0 ]] ; then
...
+    fi
```
esoergel, line 63: shouldn't the block below this be in an `else` statement?

prem-fissionhq:
This Branch ran into some merge issues, will be raising another PR with my changes. 
Sorry for the inconvenience. Thanks

PR #3897, opened 2020-06-17
Add es23- and es24-production

dannyroberts:
I embarrassingly miscalculated the RAM when I made https://dimagi-dev.atlassian.net/browse/SAAS-10977 to "expand" the cluster. Adding two more for the end state of c5.9xlarge x 5 (instead of x 3)

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
@sanjay2916

PR #3900, opened 2020-06-18
run prometheus metrics server on all commcare hosts

snopoke:


src/commcare_cloud/ansible/deploy_commcarehq.yml
```diff
@@ -80,7 +80,7 @@
   tags: services
 
 - name: Prometheus Supervisor Config
-  hosts: webworkers,pillowtop
+  hosts: commcarehq
```
snopoke, line 83: the alternate to this is to update the when but that seems overly complex and I don't think there's harm in running the process on all hosts.

```
when: prometheus_monitoring_enabled|default(False) and (inventory_hostname in groups.webworkers or inventory_hostname in groups.pillowtop or app_processes_config.management_commands.get(inventory_hostname, {}))
```

rohit-dimagi: okay thanks.

PR #3901, opened 2020-06-18
Fix vendorized ebsnvme-id

dannyroberts:
This was another thing I had to do to get the setup of a new machine with a secondary volume on AWS working. This code hasn't changed in a while, and I'm originally responsible for it, but still I have no idea how it ever worked. I don't think `file` could have ever been copying it, and I don't think `ebsnvme-id` could have ever referred to `_vendor/ebsnvme-id`

##### ENVIRONMENTS AFFECTED
production/india/staging

### DIFF
--- a/src/commcare_cloud/ansible/roles/ebsnvme/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/ebsnvme/tasks/main.yml
@@ -5,8 +5,8 @@
 
 - name: Copy ebsnvme-id
   become: yes
-  file:
-    src: "ebsnvme-id"
+  copy:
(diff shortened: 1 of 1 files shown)


dannyroberts:
@sanjay2916 FYI

dannyroberts:
Confirmed with Sanjay that he's run into this before, so it seems like "this never worked" is the winning theory. I'm guessing I built a version of this, used it a bunch, and then followed up on PR feedback to vendorize it instead of pulling from a pinned gist commit (good advice), at which point I must have misunderstood whatever I ran to test it and introduced the issue.

PR #3907, opened 2020-06-19
remove citus from django settings on india

snopoke:
This change should be tested first before rolling out.

##### ENVIRONMENTS AFFECTED
india


rohit-dimagi:
Maybe we should also remove standby settings from inventory. ?

environments/india/postgresql.yml
```diff
@@ -53,9 +53,10 @@ dbs:
         shards: [820, 1023]
         pgbouncer_host: pgbouncer0
   custom:
-    - django_alias: icds-ucr-citus
-      name: icds-ucr-citus
-      django_migrate: True  # ICDS dashboard models are stored here
+    # remove citus from django
+    # django_alias: icds-ucr-citus
+    # django_migrate: True  # ICDS dashboard models are stored here
```
dannyroberts, line 58:
```
BadValueError: Property django_migrate is required.
```
looks like it needs to to change this to
```suggestion
    django_migrate: False
```

snopoke:
@rohit-dimagi the inventory should always represent what's actually there. If we are switching the standbys to primaries then we can make that change but in a separate PR

environments/india/postgresql.yml
```diff
@@ -53,9 +53,10 @@ dbs:
         shards: [820, 1023]
         pgbouncer_host: pgbouncer0
   custom:
-    - django_alias: icds-ucr-citus
-      name: icds-ucr-citus
-      django_migrate: True  # ICDS dashboard models are stored here
+    # remove citus from django
+    # django_alias: icds-ucr-citus
+    django_migrate: False
```
rohit-dimagi, line 58: missing indent.

snopoke:
```suggestion
     django_migrate: False
```

dannyroberts:
Came across this just now. @snopoke do you know if the plan's to close this, to merge it (because we just forgot), or to wait until a particular thing that's blocking it?

snopoke:
@dannyroberts it should be safe to roll this once this PR is merged: https://github.com/dimagi/commcare-hq/pull/28532

snopoke:
@dannyroberts actually we want to keep this a bit longer since we may need to use this cluster for testing the citus upgrade.

snopoke:
Rolled out to `india` on 2020/10/20

PR #3912, opened 2020-06-22
Changing disk IO scheduler

rohit-dimagi:
* https://dimagi-dev.atlassian.net/browse/IIO-698
##### ENVIRONMENTS AFFECTED

* ALL


src/commcare_cloud/ansible/roles/thp/tasks/disk_scheduler.yml
```diff
@@ -0,0 +1,24 @@
+- stat:
```
dannyroberts, line 1: This file looks really similar to src/commcare_cloud/ansible/roles/kernel_tune/tasks/disk_scheduler.yml. Are they both supposed to be in this PR?

rohit-dimagi: That was one stale file hanging around. it's removed now.

src/commcare_cloud/ansible/roles/kernel_tune/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- include: thp.yml
+  tags: thp
+- include: disk_scheduler.yml
+  tags: disk_scheduler
```
dannyroberts, line 5: Nit: newline at the end of the file (maybe configure your editor to do this for you: [pycharm example](https://stackoverflow.com/questions/36043061/automatically-add-newline-on-save-in-pycharm))

dannyroberts:
How are you planning to roll this out? Will we need to run `cchq <env> deploy-stack --tags=kernel_tune` on all machines, and is that a safe in-place operation or do we need to do something more involved like stop all processes first?

rohit-dimagi:
> How are you planning to roll this out? Will we need to run `cchq <env> deploy-stack --tags=kernel_tune` on all machines, and is that a safe in-place operation or do we need to do something more involved like stop all processes first?

Yeah that's the idea.
This doesn't require any process to be stopped. 
I'm testing reboot on one of tcl machine to see it doesn't throw any error during reboot. i see grub config is same for aws machines so if it worked on tcl it should work on aws too.

rohit-dimagi:
Tested reboot. with following kernel setting.
`GRUB_CMDLINE_LINUX_DEFAULT="transparent_hugepage=never elevator=noop"`
 i think it's ready to go.

src/commcare_cloud/ansible/roles/kernel_tune/tasks/disk_scheduler.yml
```diff
@@ -0,0 +1,34 @@
...
+      raw_disk_list: "{{ hostvars[inventory_hostname].ansible_devices.keys() | map('regex_search', 'sd.*') | select('string') | list }}"
+
+- name: set disk scheduler to noop for every raw device
+  command: "echo noop > /sys/block/{{ item }}/queue/scheduler"
+  with_items: "{{ raw_disk_list }}"
+
+- name: Add elevator=noop to default grub config if it's empty
+  lineinfile:
+    dest: /etc/default/grub
+    regexp: '^GRUB_CMDLINE_LINUX_DEFAULT=""$'
```
rohit-dimagi, line 15: i have noticed that in aws env this value is actually not empty so the transparent hugepage also hasn't been applied.

rohit-dimagi: @dannyroberts  there is a slight change required to make it applicable for AWS Env, I'll leave that decision on you.

PR #3914, opened 2020-06-22
add formplayer logging config

knguyen142:
https://trello.com/c/prhWvkky/37-hipaa-logging-improvements

##### ENVIRONMENTS AFFECTED

All with Formplayer

## Changes
Adds logging config similar to [this](https://github.com/dimagi/formplayer/pull/736) to formplayer so it can log all requests from the `RequestResponseFilter` logger to a separate file. 

- Exclusive logging: logs from `RequestResponseFilter` only go to that file, not the root stdout logger (`additivity="false"` see [here](http://logback.qos.ch/manual/configuration.html#overrridingCumulativity)).
- Logrotation: Uses the [FixedWindowRollingPolicy](http://logback.qos.ch/manual/appenders.html#FixedWindowRollingPolicy) to only store one file  overflow (`request_response.1.log`) and both `request_response.log` and `request_response.1.log` have a size limit of 4GB. The log tailer will only read from the main `request_response.log`.

Ran `check_mode` and the output looks correct, but I don't know enough about it to know for sure.


src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
```diff
@@ -7,7 +7,7 @@ commcarehq.environment={{ env_monitoring_id }}
 commcarehq.formplayerAuthKey={{ FORMPLAYER_INTERNAL_AUTH_KEY }}
 server.port={{ formplayer_port }}
 user.suffix=commcarehq.org
-logging.level.org.springframework.web.filter.CommonsRequestLoggingFilter=INFO
+logging.config={{ _formplayer_target_dir }}/logback-spring.xml
```
dannyroberts, line 10: I think you want `{{ formplayer_release_dir }}` here instead. `_formplayer_target_dir` [sometimes is a path that includes a symlink to the current release and sometimes is an absolute path](https://github.com/dimagi/commcare-cloud/blob/70c9666692762dc8ccc23e1e91918f936d97607d/src/commcare_cloud/ansible/roles/formplayer/vars/main.yml#L13-L15) (with no symlinks). Here we always want the absolute path.

src/commcare_cloud/ansible/roles/formplayer/templates/logback-spring.xml.j2
```diff
@@ -0,0 +1,42 @@
+<configuration>
+
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
```
dannyroberts, line 5: I'm not super familiar with the date formating syntax used here, but does this include the year-month-day implicitly, or only the time component?

src/commcare_cloud/ansible/roles/formplayer/templates/logback-spring.xml.j2
```diff
@@ -0,0 +1,42 @@
...
+        </encoder>
+    </appender>
+
+    <appender name="requests" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <file>request_response.log</file>
+        <!-- Keep a fixed number of logs. -->
+        <rollingPolicy class="ch.qos.logback.core.rolling.FixedWindowRollingPolicy">
+            <fileNamePattern>request_response.%i.log</fileNamePattern>
+            <minIndex>1</minIndex>
+            <maxIndex>1</maxIndex>
```
dannyroberts, line 15: Does this mean keep the current log file and one previous one?

src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
knguyen142: Hmm, why wouldn't we want the symlinked current directory? If the `formplayer` process is instantiated with the current release [scope](https://github.com/dimagi/commcare-cloud/blob/261e56ccb8ba55ae89cf8d0e68ae589639ebffbb/src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_formplayer_spring.conf.j2#L4) shouldn't all references be scoped to the current release?

src/commcare_cloud/ansible/roles/formplayer/templates/logback-spring.xml.j2
knguyen142: Ahh my mistake, fixed. Sample
```
2020-06-22 15:13:32 [main] INFO  application.Application - No active profile set, falling back to default profiles: default

```

src/commcare_cloud/ansible/roles/formplayer/templates/logback-spring.xml.j2
knguyen142: Yep, this is what I see on my local

Screen Shot 2020-06-22 at 3 15 27 PM

Tested with a small `1MB maxFileSize` and saw the rotation. I originally had a window of 3, but figured 3 files of n max size was equivalent to 1 of n*3 for our purposes, since we're going to tail it anyway. They also have some time based rolling policies that log to a file based on the time (daily, weekly, etc.) but when testing this it seemed more prone to uncapped file sizes.

src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
dannyroberts: I guess there's two things here. `_formplayer_target_dir` will sometimes be the one with the symlink (`formplayer_current_dir`=`{{ formplayer_build_dir }}/current`) and sometimes the real path (`formplayer_release_dir`=`{{ formplayer_build_dir }}/releases/{{ new_release_name }}`), so we should replace it with one or the other of those so that it's consistent between runs. (Why does it differ between runs? If you're doing a new deploy (`cchq <env> deploy formplayer`) then we first create a new release and this gets run in that context, so `_formplayer_target_dir` will be the new release; if you're running `cchq <env> update-config` then we want to affect the current directory in place, so it'll be the `current/` directory.)

Now, on this line, the question is do we want all releases to point to the `current/` directory's logging config, or to their own logging config? To your point, since formplayer is only ever run from the current directory, in practice it's a moot point. However since we do have to decide between the two, I'd say a release should point to its own logging config rather than current's logging config—it just makes more sense. I.e. if it ever _were_ to matter (like if formplayer were run directly from a release) running its own logging config would be the correct behavior.

knguyen142: Ahh that makes sense, updated.

knguyen142:
When running in check mode I see
```
```
 {
-    "group": 0,
-    "owner": 0,
+    "group": 1027,
+    "owner": 1026,
     "path": "/home/cchq/www/staging/formplayer_build/current/sentry.properties"
 }
```
Any idea of how this could have changed. If everything looks good, I think we might be good to deploy to staging? We can also dial down the `maxFileSize` to something like 10MB to see the rotation happening?

**Comment by dannyroberts** [comment]
@knguyen142 looks like that particular diff was something unique to staging, probably due to another change someone else had been playing around with. I committed that diff and it didn't show up on second run, so it does appear to have represented a real difference in user/group ownership.

The plan to change `maxFileSize` to something small and test rotation sounds good.

**Comment by knguyen142** [comment]
Sounds good, `maxFileSize` changed to `1MB`, which will eventually get hit in stage because there seems to be a health check pinging
`2020-06-23 14:01:59.453  INFO 21769 --- [nio-8181-exec-7] a.RequestResponseLoggingFilter           : {"date":"2020-06-23T14:01:59.453Z"..."requestUrl":"http://staging.commcarehq.org//serverup"}`

We might want to have a filter list in the logging logic to not pollute the log space? Regardless, I think this is ready to be deployed to staging, let's coordinate offline.

File: src/commcare_cloud/ansible/roles/formplayer/vars/main.yml [hunk 4]
```diff
@@ -1,5 +1,6 @@
 formplayer_build_dir: "{{ www_home }}/formplayer_build"
 formplayer_data_dir: "{{ encrypted_root }}/formplayer"
+formplayer_logging_dir: "{{ formplayer_data_dir }}/log"
```
dannyroberts, line 3: I made https://github.com/dimagi/commcare-cloud/compare/dmr/efs+kev/formplayer-requests-logging-config to combine this branch with https://github.com/dimagi/commcare-cloud/pull/3917. The one extra commit I added to make them play nicely was https://github.com/dimagi/commcare-cloud/commit/8b98e7b9665b0562877b8087862d693277d14345, which changes this line to
```
formplayer_logging_dir: "{{ encrypted_root }}/formplayer/log"
```
since the data dir points to the shared network drive and we want logging on /opt/data

knguyen142: Awesome, thanks @dannyroberts. I added this [commit](https://github.com/dimagi/commcare-cloud/pull/3914/commits/45bf4cbe1603dd33889f84181590fc9781c95df8) which deals with the flag that we talked about. Do you want me to commit it to the new branch and make a PR from that?

src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
```diff
@@ -51,3 +51,7 @@ redis.Password={{ localsettings.REDIS_PASSWORD }}
 endpoints.enabled=true
 management.address: 0.0.0.0
 management.port: 8081
+
+{% if formplayer_sensitive_data_logging %}
+sensitiveData.enableLogging=true
+{% endif %}
```
dannyroberts, line 57: Nit: should be a newline at the end of the file. I recommend configuring your editor to do this for you. (For pycharm you can check out this SO answer: https://stackoverflow.com/a/37260222)

dannyroberts:
Since my conflicting branch got merge, I just merged the conflict resolution branch (dmr/efs+kev/formplayer-requests-logging-config) into this PR, which just adds the one commit https://github.com/dimagi/commcare-cloud/pull/3914/commits/8b98e7b9665b0562877b8087862d693277d14345

knguyen142:
Added new null logger when logging disabled. Tested on staging with a formplayer that logs to the requestresponse class.

Cases:
1. When `formplayer_sensitive_data_logging` = `True`
Logs from the requestresponse class go to a separate log file, not to stdout.
2. When `formplayer_sensitive_data_logging` = `False`
Logs from the requestresponse do not go to separate file, and do not go to stdout.

PR #3916, opened 2020-06-23
scripts to automate to generate the login and sudo records

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/IIO-889

Tool to capture the login and sudo records for audit purpose.

##### ENVIRONMENTS AFFECTED
ICDS
ICDS-STAGING


src/commcare_cloud/ansible/roles/devops_scripts/tasks/main.yml
```diff
@@ -14,3 +15,60 @@
     owner: ansible
     group: ansible
     mode: "u=rwx,g=rwx,o=r"
+
+- name: Copy sudo-report.sh script to /usr/local/sbin
+  template:
```
snopoke, line 20: this could use the `copy` module since it's not actually a template

shyamkumarlchauhan: used copy module.

PR #3917, opened 2020-06-23
Set up formplayer EFS in ansible

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-10856 seems to be working fine on staging, but the configuration drift is annoying. Setting it up in ansible to match what we have now. This doesn't affect other AWS envs yet (other than to create an empty shared drive that won't actually mount anywhere.

##### ENVIRONMENTS AFFECTED
staging


dannyroberts:
This also makes the choice to store that at `/opt/formplayer_data/` instead of `/opt/data` to (1) show it's a different kind of thing and (2) allow us to manage the switchover carefully with minimal downtine once we get there. Also we want to store the hipaa somewhere under /opt/data/, but we don't want to store them on the shared formplayer data directory, and this keeps those separate.

src/commcare_cloud/ansible/deploy_formplayer.yml
```diff
@@ -1,6 +1,32 @@
...
+        shared_drive_enabled: yes
+        shared_mount_dir: /opt/formplayer_data
+        shared_mount_address: "{{ formplayer_efs_dns }}"
+    - import_tasks: roles/shared_dir/tasks/setup_client.yml
+      when: 'formplayer_efs_dns|default(False)'
+      vars:
+        shared_drive_enabled: yes
+        shared_mount_dir: /opt/formplayer_data
+        shared_mount_address: "{{ formplayer_efs_dns }}"
+        shared_mount_opts: "nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport"
```
millerdev, line 19: I take it from this line that these systems will mount the shared drive with NFS. Is that correct?

I'm assuming you are aware of the limitations of using SQLite on NFS on some level. There are many warnings about that, and I think we do a best effort with formplayer to not have muliple processes accessing the same database, but if that ever happens it could result in corruption. This may also be impacted by the locking mode formplayer uses. Does it use [WAL mode](https://sqlite.org/wal.html)?

- https://sqlite.org/lockingv3.html#how_to_corrupt "Your best defense is to not use SQLite for files on a network filesystem."
- https://sqlite.org/faq.html#q5

However, looks like we may be using a version of NFS in which locking is implemented properly, so maybe this is not a major concern? https://stackoverflow.com/a/52378465/10840

dannyroberts: SQLite + NFS + formplayer custom locking and whether the three will work together nicely as we have them is definitely the main area of concern, so thanks for these links. I'm also planning on going over this in detail with Clayton again to make sure formplayer locking and how it writes to sqlite will at least "logically" give us solid protection against concurrent writes and db corruption. In addition to that, I think we'd also stick with sticky-routing-by-default as a hedge: convince ourselves that full multi-machine concurrency will work perfectly, and then at the same time only rely on that when one machine is down/being restarted.

I'll be communicative about our reasoning here if and when we are close to deciding to roll this out in production. For now it'll live in staging, and this PR is to support that setup that currently only exists in staging so that we don't have to manage the drift manually.

millerdev: Thanks for the follow-up.

dannyroberts:
merging now to support staging formplayer setup. Will not affect production or india other than by creating an empty and unused EFS resource in terraform

PR #3925, opened 2020-06-26
Add config for NO_WWW_SITE_HOST in terraform ga_alb_waf

dannyroberts:
##### ENVIRONMENTS AFFECTED
production


dannyroberts:
Yeah sorry. It's preparing so I can point other DNS entires to the new IP address that are currently pointing to the old IP address still. This was always the intention when we switched to the new IP address (which routes through global accelerator, the application load balancer, and importantly the the web application firewall), but it took a while to finalize. I realized today I also need to add to this anything in `ALTERNATE_HOSTS`

PR #3930, opened 2020-06-29
Allow --commcare-rev with --set on deploy

dannyroberts:
Previously if you use --set, it will effectively overwrite the value
of --commcare-rev (since that also uses --set on fab downstream)

This change also removes some of the awkwardness of using --set
where you have to then explictly pass in the component arg or it
treats the value of --set as the component arg (and fails).
Now there is none of that.


src/commcare_cloud/commands/deploy.py
```diff
@@ -31,11 +31,14 @@ class Deploy(CommandBase):
...
         """, default=None),
+        Argument('--set', dest='fab_settings', help="""
+            fab settings in k1=v1,k2=v2 format to be passed down to fab 
+        """, default=None),
         shared_args.QUIET_ARG,
         shared_args.BRANCH_ARG,
     )
 
-    def run(self, args, unknown_args):
+    def run(self, args, unknown_args, color_warn=None):
```
orangejenny, line 41: How is `color_warn` supplied? It looks like lines 57-61 will fail hard if it's `None`.

src/commcare_cloud/commands/deploy.py
```diff
@@ -50,15 +53,23 @@ def run(self, args, unknown_args):
                 if deploy_component != 'both':
                     _warn_no_formplayer()
                 self.deploy_commcare(environment, commcare_rev, args, unknown_args)
-
         if deploy_component in ['formplayer', 'both']:
+            if deploy_component != 'both':
```
orangejenny, line 57: Nitpick: this would be clearer if it were `if deploy_component == 'formplayer'`

dannyroberts: I was following the convention three/four lines above 🤷

src/commcare_cloud/commands/deploy.py
dannyroberts: Thanks this was totally me not understanding what my IDE was doing for me, fixed now

src/commcare_cloud/commands/deploy.py
orangejenny: fair enough

PR #3934, opened 2020-07-01
ansible support for checking out additional git repositories

snopoke:
Part of work required to move ICDS code to a separate (private) git repo:
https://docs.google.com/document/d/1z21WtL33XpLdmn5GnC58Tkb5ra8tm20UYymAIrg9wgg/edit


src/commcare_cloud/environment/schemas/meta.py
```diff
@@ -1,6 +1,18 @@
...
 
+class GitRepository(jsonobject.JsonObject):
+    url = jsonobject.StringProperty(required=True)
+    dest = jsonobject.StringProperty(required=True)  # relative to the code_source/external directory
+    version = jsonobject.StringProperty(default="master")
+    requirements_path = jsonobject.StringProperty(default="requirements.txt")
+
+    def to_generated_variables(self):
+        vars = self.to_json()
+        vars["dest"] = "extra/" + self.dest
```
dannyroberts, line 12: Just curious so I can follow along. Will `extra` be a "sources root" or be a python package itself?

I.e. if we have

```
commcare-hq/
  ...
  extras/
    ...
    my_custom_repo/
      ...
      file.py
```
will we import things from file.py like `from extras.my_custom_repo.file import *` or `from my_custom_repo.file import *` or something else? Assuming these are meant to share a runtime with commcare-hq itself, will that get bootstrapped by adding the path to the django apps to `LOCAL_INSTALLED_APPS` in localsettings?

snopoke:
@dannyroberts this will work in the same way it does now for submodules: https://github.com/dimagi/commcare-hq/pull/28007

In `manage.py` we will add each subfolder of `extra` to the python path. So for example:
```
commcare-hq/
  ...
  extras/
    ...
    my_custom_repo/
      ...
      my_package/
      __init__.py
      file.py
```

We'd import this as `from my_package import file`.

PR #3944, opened 2020-07-06
Switch sentry to onprem for icds

rohit-dimagi:
https://dimagi-dev.atlassian.net/browse/IIO-868
##### ENVIRONMENTS AFFECTED

* ICDS


rohit-dimagi:
Using private ip in DSN instead of public domain name to reduce traffic.

rohit-dimagi:
> how sure are we that we'll be able to handle the load?

we have got 8x16 machine. i think we would be fine.

PR #3951, opened 2020-07-07
Migration alerts before deployment

AliRizvi1:
https://github.com/dimagi/commcare-hq/issues/27838

##### ENVIRONMENTS AFFECTED
Deployers will now be able to see a list of pull requests since the last deploy, in order to facilitate more visibility between code authors and deployers.

### DIFF
--- a/src/commcare_cloud/fab/utils.py
+++ b/src/commcare_cloud/fab/utils.py
@@ -31,10 +31,6 @@ from .const import (
 from six.moves import input
 
 LABELS_TO_EXPAND = [
-    # "product/all-users-all-environments",
-    # "product/prod-india-all-users",
-    # "product/feature-flag",
-    # "product/all-users",
     "reindex/migration",
 ]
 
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/fab/utils.py
```diff
@@ -273,3 +280,76 @@ def generate_bower_command(command, production=True, config=None):
...
+
+def _get_pr_labels(pr_number):
+    url = 'https://api.github.com/repos/dimagi/commcare-hq/issues/{}'.format(pr_number)
+    json_response = requests.get(url).json()
+    if not json_response.get('number'):
+        return []
+    assert pr_number == json_response['number'], (pr_number, json_response['number'])
+
+    return [
+        {'name': label['name'], 'color': label['color']}
```
esoergel, line 330: it looks like you only use `name`, so I'd just return a list of names

src/commcare_cloud/fab/utils.py
```diff
@@ -273,3 +280,76 @@ def generate_bower_command(command, production=True, config=None):
...
+        print("{0}. ".format(i), end="")
+        print ("{title} {url} | ".format(**pr), end="")
+        j = 0
+        num_labels = len(pr['labels'])
+        for label in pr['labels']:
+            j += 1
+            if (j == num_labels):
+                print("{name}".format(**label))
+            else:
+                print ("{name} ".format(**label), end="")
```
esoergel, line 354: an easier way to do this is:
```python
print(" ".join(label['name'] for label in pr['labels']))
```

AliRizvi1: Nice, the way I was doing it before seemed rather non-pythonic. Thanks for the tip!

esoergel:
One last thought on this PR - we follow PEP8 for code formatting, which dictates two blank lines between each top-level function definition.  I also see one other (minor) code formatting error.  Can you try setting up your editor to automatically run a python linter?  That will flag these sorts of things.  I'd recommend pylint if that's straightforward to set up, though its default configuration can be a bit overeager.

PR #3954, opened 2020-07-07
add changelog for moving to yarn

biyeun:
https://dimagi-dev.atlassian.net/browse/SAAS-10881

##### ENVIRONMENTS AFFECTED
all environments on deploy


changelog/0035-new-js-package-manager.yml
```diff
@@ -0,0 +1,27 @@
+title: Switch to new package manager
+key: new-js-package-manager
+date: 2020-07-20
+optional_per_env: no
+min_commcare_version:
```
millerdev, line 5: Should we set this so it is clear when this needs to be done? (the version, a git commit hash, may not exist yet)

PR #3960, opened 2020-07-09
fix default val for grafana group

rohit-dimagi:
The variables file is evaluated even for envs that aren't using the feature.
##### ENVIRONMENTS AFFECTED

* ICDS


rohit-dimagi:
@snopoke think we have to split this file, i just noticed it's not env specific file.

PR #3970, opened 2020-07-10
Integrate migration alerts into deploy

esoergel:
Implements https://github.com/dimagi/commcare-hq/issues/27838, using @AliRizvi1 's work in https://github.com/dimagi/commcare-cloud/pull/3951

```
[10.201.11.173] Executing task 'deploy_commcare'
List of PRs since last deploy:
1. Bump statici18n to version supporting Django 2.2 https://github.com/dimagi/commcare-hq/pull/28084 | product/invisible
2. [ICDS Dashboard] Pull the program summary data on the same level https://github.com/dimagi/commcare-hq/pull/28075 | product/custom
3. Fix sorting for FCT count column https://github.com/dimagi/commcare-hq/pull/28087 | product/all-users-all-environments
Are you sure you want to preindex and deploy to staging? [y/N] y
```

Since this significantly increases API requests to github, all users are now prompted to set up a github API key, though they can still abstain.

It also drops the use of git tags for diff URLs, instead, it always inspects the code root for the currently deployed commit.  We did this already as a fallback, and I don't know that there was ever a good reason to not always do that.

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -30,40 +30,28 @@ from __future__ import print_function
 import datetime
 import functools
 import os
+import pipes
 import posixpath
+from distutils.util import strtobool
 from getpass import getpass
 
-import pipes
 import pytz
-from distutils.util import strtobool
-
-from fabric import utils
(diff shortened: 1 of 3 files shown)


esoergel:
bump

PR #3973, opened 2020-07-13
List required ports for new installation

esoergel:
Making a new PR for https://github.com/dimagi/commcare-cloud/pull/3856
Looks like github got confused about the branches


prem-fissionhq:
Thanks Ethan.

kaapstorm:
(I've never seen netcat being used to test listening ports. That's cool.)

docs/setup/new_environment.md
```diff
@@ -10,7 +10,12 @@ This document will walk you through the process of setting up a new monolith ser
     * If you are using VirtualBox, you can follow the instructions on [Configuring VirtualBox for testing CommCareHQ](../howto/configure-virtualbox.md).
     * A domain name which directs to your server.
 * Access to the monolith via SSH with a user who has sudo access. If you installed the base Ubuntu 18.04 image yourself, this should be the default.
-
+* [This table](../Commcare_Ports_information.md) lists the ports that must be accessible in your network configuration. Please enable these ports in the Iptables/Firewall/Security groups.
+  * Run this command to verify that these ports are accessible:
+      ```bash
+      $ for PORT in 22 443 80 9010 8181 9092 2181 6379 5432 6432 5672 9200 9300 5984 4369 5555 25 465 587; do nc -zv localhost $PORT; done
```
sravfeyn, line 16: @esoergel 

I am a bit confused by this `Run this command to verify that these ports are accessible:`. Is this supposed to be done once CommCareHQ is installed to check that processes installed/running on these ports are accessible or is the command supposed to be run before the installation, in which case, I would expect `nc -zv $PORT` to fail since there wouldn't be a process listening on this port.

esoergel: This is supposed to be run before install to be sure that the server's network config is correct.  I don't recall what the failure conditions look like, but all we really need to do is determine whether they need their hosting provider to change something for us to use that port.  Do you know if this is discernible from the output?

sravfeyn: The output could say `refused` or `succeeded`. Per my understanding, `refused` doesn't necessarily mean that the ports are blocked by the hosting provider. Even if the ports are not blocked this will fail if there aren't any processes listening on these ports (which is true in most cases when nothing is installed). 

Also, most likely if ports are blocked, it's likely that is blocked on the internal IP and not on localhost (i.e. port access through localhost might be enabled but not through the IP).

PR #3981, opened 2020-07-15
added kill operation to stop elasticsearch in monit

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/IIO-915

added kill operation to stop elasticsearch in monit .

##### ENVIRONMENTS AFFECTED
ICDS
ICDS-Staging
Production
India
Staging

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
@@ -2,5 +2,5 @@ check file elasticsearch with path {{ elasticsearch_data_dir }}/logs/{{ elastics
    group database
    group elasticsearch
    start program = "/bin/systemctl start {{ elasticsearch_service_name }}"
-   stop program = "/bin/kill -9 $(ps -ef|grep -i {{ elasticsearch_service_name }}|grep -v grep |awk '{print $2}');/bin/systemctl stop {{ elasticsearch_service_name }}"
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
```diff
@@ -2,5 +2,5 @@ check file elasticsearch with path {{ elasticsearch_data_dir }}/logs/{{ elastics
    group database
    group elasticsearch
    start program = "/bin/systemctl start {{ elasticsearch_service_name }}"
-   stop program = "/bin/systemctl stop {{ elasticsearch_service_name }}"
+   stop program = "/bin/kill -9 $(ps -ef|grep -i {{ elasticsearch_service_name }}|grep -v grep |awk '{print $2}');/bin/systemctl stop {{ elasticsearch_service_name }}"
```
rohit-dimagi, line 5: why don't just use `pkill -f elasticsearch`? seems much cleaner.

sravfeyn: I remember when trying this manually it didn't work, but may be I was doing something wrong.

shyamkumarlchauhan: I did try pkill -f elasticsearch  and it works.... putting a sleep for 5 seconds after that to clear the killed process from memory.

PR #3982, opened 2020-07-15
update kafka version in localsettings

snopoke:
This change was missed when upgrading kafka. All envs are now on 2.4.1. 

This change to localsettings should not have any impact on Kafka usage at present. Excerpt from the docs:

```
api_version (tuple) –
Specify which Kafka API version to use. If set to None, the client will attempt to infer the broker version by probing various APIs. Different versions enable different functionality.

Examples

(0, 9) enables full group coordination features with automatic partition assignment and rebalancing,
(0, 8, 2) enables kafka-storage offset commits with manual partition assignment only,
(0, 8, 1) enables zookeeper-storage offset commits with manual partition assignment only,
(0, 8, 0) enables basic functionality but requires manual partition assignment and offset management.
```

#### ENVIRONMENTS AFFECTED
ALL


snopoke:
rolled out to icds

PR #3994, opened 2020-07-20
Refactor for logshipping (part 1)

dannyroberts:
##### ENVIRONMENTS AFFECTED
production, india, staging


dannyroberts:
I kind of have to do this in parts because testing it (the way we have it right now) means rolling things out on staging, so having this in a long-running branch would create significant drift

src/commcare_cloud/commands/terraform/migrations/0005_move_s3_log_bucket_to_logshipping.py
```diff
@@ -0,0 +1,3 @@
+def get_new_resource_address(environment, old_resource_address):
+    if old_resource_address == 'module.ga_alb_waf.aws_s3_bucket.front_end_alb_logs':
+        return 'module.logshipping.aws_s3_bucket.log_bucket'
```
millerdev, line 3: It seems less than ideal for this to implicitly return `None`, also that just seems like a weird return value given the name of the function.
```suggestion
        return 'module.logshipping.aws_s3_bucket.log_bucket'
    return None
```
However, returning `None` implies that the returned value needs to be checked whenever this function is called. This seems to be asking for some refactoring unless this is tying into some larger pattern in terraform?

dannyroberts: this file ties into a state-migrations pattern I set up when first working on terraform. Terraform's internal resource address for a resource is tightly coupled to where that resource is written down in the terraform code, so when you refactor terraform files, you often have to systematically rename resources.

the `terraform-migrate-state` commcare-cloud command goes through these files and runs any not-yet-run state migrations, by calling `get_new_resource_address` on each resource address and, if the return value is not `None`, performing a `terraform state mv <old> <new>` to rename the resource.

Does that clear up what you were asking? It's a thing that I custom-built, but it's super annoying to do this kind of thing across multiple environments without a custom-built thing like this as far as I could tell at the time from looking at how people handle this.

millerdev: That does make things clearer. Thanks for the explanation.

PR #4003, opened 2020-07-22
Add support for defining N identical machines in terraform

dannyroberts:
Easiest with ?w=1

Lets you add something like

`terraform.yml`
```
  - server_name: "formplayer_a{i}-production"
    server_instance_type: r5.2xlarge
    network_tier: "app-private"
    az: "a"
    volume_size: 600
    group: "formplayer"
    os: bionic
    count: 32
```

`inventory.ini.j2`
```
{{ __formplayer_a__ }}

[formplayer:children]
formplayer_a
```

And it'll create 32 identical machines and put them in the inventory for you under individual groups (`[formplayer_a000]` as well as a group that contains all of them `[formplayer_a]`)

##### ENVIRONMENTS AFFECTED
None yet, but about to use on production

### DIFF
--- a/src/commcare_cloud/environment/schemas/terraform.py
+++ b/src/commcare_cloud/environment/schemas/terraform.py
@@ -30,6 +30,12 @@ class TerraformConfig(jsonobject.JsonObject):
             data['aws_profile'] = data.get('account_alias')
         return super(TerraformConfig, cls).wrap(data)
 
+    def to_generated_json(self):
+        obj = self.to_json()
+        obj['servers'] = [server.to_generated_json() for server in self.servers]
+        obj['proxy_servers'] = [server.to_generated_json() for server in self.proxy_servers]
+        return obj
+
 
 class VpnConnectionConfig(jsonobject.JsonObject):
     _allow_dynamic_properties = False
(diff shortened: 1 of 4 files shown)


src/commcare_cloud/environment/schemas/terraform.py
```diff
@@ -58,6 +64,39 @@ class ServerConfig(jsonobject.JsonObject):
...
+        if self.count is None:
+            return [host_name]
+        else:
+            return [host_name.format(i='{:03d}'.format(i)) for i in range(self.count)]
+
+    def get_host_group_name(self):
+        if self.count is None:
+            raise ValueError("Can only call get_host_group_name() on a server with count")
+        else:
+            return self.server_name.split('-', 1)[0][:-3]
```
ctsims, line 94: the second slice fails a sniff test for me of whether it's obvious to introspect what this code is doing without context from outside of the file. I know it trims off the last 3 characters of the string, but it's not quite obvious to me why (or what will happen to strings that don't meet this pattern).

It would be helpful to either have an example pattern available in-code to see what the code is intended to pull out, or to maybe make the code more verbose in order to more unambiguously communicate what that pattern is intended to catch.

PR #4004, opened 2020-07-22
Add 32 new formplayer machines to replace old ones

dannyroberts:
Builds on https://github.com/dimagi/commcare-cloud/pull/4003.

##### ENVIRONMENTS AFFECTED
Production


src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -187,35 +187,44 @@ def context(self):
 
         servers = self.environment.terraform_config.servers + self.environment.terraform_config.proxy_servers
         for server in servers:
```
ctsims, line 189: nitpick: 'server' here is a misnomer, since this refers both to a server template and to a list of ids (names) associated with individual hosts that match that template. Maybe I'm behind the times but to me 'server' means an individual host, so someone in the future might find this confusing and make a mistake v. calling it `server_profile` or `server_spec`

PR #4005, opened 2020-07-22
Don't display too many PRs on deploy

esoergel:
Paranoid followup from https://github.com/dimagi/commcare-cloud/pull/3970
I was thinking of ways this might go bad, so I figured I'd put a couple guardrails on it.   Bonus colors for pizzazz too.


biyeun:
colors!!! 🌈

PR #4017, opened 2020-07-24
Add COVID custom case actions

proteusvacuum:
https://docs.google.com/document/d/1LLrD_jgtdP6VmoxvjqGEEPTvsDGv9WyhNzQcHrDkD8c/edit?ts=5f0ce3e9#

Required for https://github.com/dimagi/commcare-hq/pull/28191

##### ENVIRONMENTS AFFECTED

Production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -111,6 +111,10 @@ localsettings:
   ANALYTICS_DEBUG: False
   ANALYTICS_LOG_LEVEL: "warning"
   AUDIT_ALL_VIEWS: True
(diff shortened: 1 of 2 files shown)


proteusvacuum:
@orangejenny do we have test COVID data on staging? If so, then yes that's a great idea.

orangejenny:
@proteusvacuum Not quite yet, but it's in the works. It won't be ready to test this feature, but might be useful if we need to re-test in the future.

proteusvacuum:
@orangejenny got it. Added to staging in the other PR.

PR #4020, opened 2020-07-24
Do not pin pickle version

millerdev:
Old pickles will not be affected since the pickle version is embedded in the pickle.

##### ENVIRONMENTS AFFECTED

all


kaapstorm:
I didn't know we did this. But I did know that I broke some stuff locally when testing Python 3.8 and switching back to 3.6.

Python 3.6:
```python
>>> pickle.compatible_formats                                                                                                   
['1.0', '1.1', '1.2', '1.3', '2.0', '3.0', '4.0']
```

Python 3.8:
```python
>>> pickle.compatible_formats                                                                                                   
['1.0', '1.1', '1.2', '1.3', '2.0', '3.0', '4.0', '5.0']
```

I might use `'PICKLE_VERSION': 4` locally.

millerdev:
Oh, that's a good thing to keep in mind. We may need to re-enable something like this when we upgrade Python next.

PR #4023, opened 2020-07-27
Add support for ES7 in commcare-cloud

rohit-dimagi:
https://dimagi-dev.atlassian.net/browse/IIO-916
Overview of changes:
https://www.elastic.co/guide/en/elasticsearch/reference/5.6/breaking_50_settings_changes.html
* It uses log4j2.properties file for logging
* It needs to have a jvm.options file in the config folder
* It uses environment var for conf_dir setup instead of option parameter on es service
* it creates certain index at startup so we have to allow creation of those index instead of false param.
* Can't modify index level setting in elasiticsearch.yml
* plugin command is now elasticsearch-plugin
* script mode setting is now defaults to true
* Drop support for V1.7.x

##### ENVIRONMENTS AFFECTED

* ALL


src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -10,7 +10,10 @@ node.name: "{{ elasticsearch_node_name }}"
 {% if elasticsearch_node_zone is defined %}
```
sravfeyn, line 10: We need to make sure that we use different cluster name for ES7 hosts from existing ES2 hosts, so that ES2 and ES7 don't try to join each other (just to be safe)

sravfeyn:
Thanks for the PR, Rohit.

1. Did you test this locally?
2. Wondering if you looked at breaking changes in settings in ES5 and ES7 and took them into consideration?
    https://www.elastic.co/guide/en/elasticsearch/reference/5.6/breaking_50_settings_changes.html      
    https://www.elastic.co/guide/en/elasticsearch/reference/7.6/breaking-changes-7.0.html#breaking_70_settings_changes

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
sravfeyn: Also, we need to add ES7 datadog tags for ES7 hosts.

rohit-dimagi:
>     1. Did you test this locally?
Yes, i've tried to test it as much i can. 
>     2. Wondering if you looked at breaking changes in settings in ES5 and ES7 and took them into consideration?
>        https://www.elastic.co/guide/en/elasticsearch/reference/5.6/breaking_50_settings_changes.html
>        https://www.elastic.co/guide/en/elasticsearch/reference/7.6/breaking-changes-7.0.html#breaking_70_settings_changes
I have looked at the changes which we are using in our config but not all.

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
rohit-dimagi: > Also, we need to add ES7 datadog tags for ES7 hosts.

Added

rohit-dimagi: > We need to make sure that we use different cluster name for ES7 hosts from existing ES2 hosts, so that ES2 and ES7 don't try to join each other (just to be safe)

I can't think of how to make it work in role itself. i guess it's upto the user to be cautious.

sravfeyn: The cluster name can be set in the es config file. I thought we already have  var name for that, don't we? If not, we could add one and make sure to suffix es7 for ES7 hosts?

rohit-dimagi: Added an assert statement.

environments/staging/inventory.ini
```diff
@@ -125,12 +125,12 @@ redis2
...
-# es5
-# ES v7 hosts
-es6
-es7
+# old ES v7.9.1 hosts
+es4
+es5
+# ES v7.6.2 hosts
+# es6
+# es7
```
millerdev, line 133: ES v7.9.1 is not the "old" version. Can you explain what's going on here?

sravfeyn: Will revert this, was planning to upgrade staging, but it's not necessary.

sravfeyn:
@millerdev @snopoke This is ready to go.

src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml
```diff
@@ -1,18 +1,45 @@
+- name: "Check if Cluster Name is updated for Elasticsearch V7"
+  assert:
+    that:
+      - "'es7_' in elasticsearch_cluster_name"
+    fail_msg: "Elasticsearch V7 cluster name should start with es7_ prefix"
+  when: elasticsearch_version  == "7.9.1"
```
snopoke, line 6: what about checking for > v7?
```suggestion
  when: elasticsearch_version is version('7.0.0', '>=')
```

src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml
```diff
@@ -1,18 +1,45 @@
...
+  
+- name: Download Elasticsearch 2.4.x
   become: yes
   get_url:
     url: "https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-{{ elasticsearch_version }}.tar.gz"
     dest: "/home/{{ cchq_user }}/downloads/elasticsearch-{{ elasticsearch_version }}.tar.gz"
     sha256sum: "{{ elasticsearch_download_sha256 }}"
+  when: elasticsearch_version == "2.4.6"
+
+- name: Download Elasticsearch 7.9.1
```
snopoke, line 36: how about moving these downloads to the version specific task files (misc_v2 & misc_v7) (maybe rename them to es_v2 and es_v7

src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
```diff
@@ -0,0 +1,38 @@
...
+  replace:
+    path: "{{ elasticsearch_home }}/bin/elasticsearch.in.sh"
+    regexp: '^JAVA_OPTS="\$JAVA_OPTS -XX:\+HeapDumpOnOutOfMemoryError"'
+    replace: '#JAVA_OPTS="$JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"'
+
+- name: Install elasticsearch python client
+  become: yes
+  pip:
+    name: elasticsearch>=2.0.0,<3.0.0
+  when: elasticsearch_version == '2.4.6'
```
snopoke, line 23: seems unnecessary since the whole file is version scoped

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -81,14 +97,35 @@ cloud:
...
 {% if es_local_repo and not backup_es_s3%}
 path.repo: ["{{ elasticsearch_data_dir }}/snapshot"]
 {% endif %}
 
 
+
+{% if elasticsearch_version == '7.9.1' %}
+cluster.initial_master_nodes:
+{% for item in groups.elasticsearch %}
+{% if hostvars[item].elasticsearch_master is defined %}
```
snopoke, line 117: I think you want this to also be True correct:
```suggestion
{% if hostvars[item].elasticsearch_master|default(False) %}
```

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -98,3 +135,16 @@ node.master: {{ elasticsearch_master|default('false') }}
 {% endif %}
 
 node.data: {{ elasticsearch_data|default('true') }}
+{% endif %}
+
+{% if elasticsearch_version == '7.9.1'%}
+{% if elasticsearch_data|default('true') and elasticsearch_master|default('false') %}
```
snopoke, line 141: use bools not strings
```suggestion
{% if elasticsearch_data|default(True) and elasticsearch_master|default(False) %}
```

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -98,3 +135,16 @@ node.master: {{ elasticsearch_master|default('false') }}
...
+node.roles: [ master data ]
+{% endif %}
+{% if elasticsearch_master|default('false') %}
```
snopoke, line 144:
```suggestion
{% if elasticsearch_master|default(False') %}
```

millerdev (approved):

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -98,3 +135,16 @@ node.master: {{ elasticsearch_master|default('false') }}
...
+node.roles: [ master ]
+{% endif %}
+{% if elasticsearch_data|default('true') %}
```
snopoke, line 147:
```suggestion
{% if elasticsearch_data|default(False) %}
```

snopoke (changes requested):
Few comments. In general I think you should use the [version](https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#comparing-versions) test to compare versions instead of checking for exact matches which seems brittle to version updates.

src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml
sravfeyn: The following tasks after this are common to both versions, moving it to version specific files require duplicating the common tasks in both files.

src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
sravfeyn: Sure

sravfeyn:
Thanks for the comments @snopoke. I have updated now.

src/commcare_cloud/ansible/roles/datadog/templates/elastic.yaml.j2
```diff
@@ -7,3 +7,6 @@ instances:
   - url: http://{{ inventory_hostname }}:9200
     tags:
       - environment:{{ env_monitoring_id }}
+{% if elasticsearch_version is version('7.0.0', '>=') %}
+      - es_version: {{ elasticsearch_version }}
```
snopoke, line 11: any reason not to always include this tag?

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -81,20 +97,56 @@ cloud:
...
+{% endfor %}
+{% endif %}
+
+
+{% if elasticsearch_version == '2.4.6'%}
 # see also discovery.zen.minimum_master_nodes above for how a master gets elected
 {# only include master eligibility if it is set on at least one node in the cluster #}
 {% if groups.elasticsearch|map('extract', hostvars, 'elasticsearch_master')|select()|list %}
-node.master: {{ elasticsearch_master|default('false') }}
+node.master: {{ elasticsearch_master|default(False) }}
```
snopoke, line 132: I'm not sure if this needs to be `true/false` or if `True/False` is also OK. The ES docs use lowercase which I think is why this was a string before.

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -81,20 +97,56 @@ cloud:
...
 {% else %}
 # node.master: true
 {% endif %}
 
-node.data: {{ elasticsearch_data|default('true') }}
+node.data: {{ elasticsearch_data|default(True) }}
```
snopoke, line 137: same as above re lowercase

src/commcare_cloud/ansible/roles/datadog/templates/elastic.yaml.j2
sravfeyn: I guess we could, but this was mainly to identify the metrics just during the upgrade process.

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
sravfeyn: Ah, right. Updated to what it was

snopoke (approved):

PR #4029, opened 2020-07-28
Revert "ansible support for checking out additional git repositories"

calellowitz:
Reverts dimagi/commcare-cloud#3934

This broke production deploy. It looks like the release directory no longer pulls the newest code. It failed with `fatal: reference is not a tree: <commit_hash>` doing `git checkout <commit_hash>` in the commcare-hq code, where <commit_hash> matches the commit of the deploy tag


snopoke:
strange, I tested this by creating a private release and it worked find. I'll take a look tomorrow

PR #4031, opened 2020-07-28
Revert "Revert "ansible support for checking out additional git repositories""

snopoke:
Reverts dimagi/commcare-cloud#4029


snopoke:
There was one bug which I've fixed. I was able to deploy ICDS staging with not issues.

PR #4044, opened 2020-07-31
Add cron to clean kafka system logs

rohit-dimagi:
I looked at log4j.properties to achieve this but it's really kinda looked confusing. so i chose cron.
##### ENVIRONMENTS AFFECTED

* ALL


rohit-dimagi:
> Who not use logrotate?

seems fine to use logrotate too. switched.

PR #4055, opened 2020-08-04
New environment instructions

esoergel:
A couple minor tweaks to the new env instructions - see commit messages for details


esoergel:
ignoring build for docs only change

PR #4058, opened 2020-08-04
Use shared dir for staticfiles/CACHE

dannyroberts:
New functionality will be under an option set in fab-settings.yml, and will allow deploying to multiple proxies.

- [x] test that staging deploy works as is
- [x] test that staging deploy works with two proxies (see: https://github.com/dimagi/commcare-cloud/pull/4059)

##### ENVIRONMENTS AFFECTED
all that want to take advantage of it (and have a shared drive)


src/commcare_cloud/fab/operations/staticfiles.py
```diff
@@ -97,18 +97,57 @@ def collectstatic(use_current_release=False):
...
         sudo('{}/bin/python manage.py purge_compressed_files'.format(venv))
-    update_manifest(save=True, use_current_release=use_current_release)
+
+    push_manifest(use_current_release=use_current_release)
+
+
+def push_manifest(use_current_release=False):
+    if env.use_shared_dir_for_staticfiles:
+        with cd(env.code_root if not use_current_release else env.code_current):
+            git_hash = sudo('git rev-parse HEAD').strip()
```
calellowitz, line 114: could this (and the stuff below), be pulled into a function, or even work by passing an argument to the other function. it feels weird to mix fab operations and raw shell commands like this

src/commcare_cloud/fab/operations/staticfiles.py
```diff
@@ -97,18 +97,61 @@ def collectstatic(use_current_release=False):
...
         sudo('{}/bin/python manage.py purge_compressed_files'.format(venv))
-    update_manifest(save=True, use_current_release=use_current_release)
+
+    push_manifest(use_current_release=use_current_release)
+
+
+def push_manifest(use_current_release=False):
+    if env.use_shared_dir_for_staticfiles:
+        with cd(env.code_root if not use_current_release else env.code_current):
+            git_hash = _get_git_hash()
```
snopoke, line 114: what's the reasoning to use the git hash? why not use the release name? What happens if you deploy twice with the same git hash?

dannyroberts: Yeah I thought about that and couldn't figure out if one was clearly better than the other so I went with what the pre-existing process (the update_manifest management command) did and [used the git hash](https://github.com/dimagi/commcare-hq/blob/1b4a2a6c08ca3b50a4e6e86e2c57220a64f32b5c/corehq/apps/hqwebapp/management/commands/update_manifest.py#L61-L68).

I think if you deploy twice from the same hash the second time it'll overwrite the first time, but I think that's okay.

PR #4059, opened 2020-08-05
Test multi-proxy setup on staging

dannyroberts:
I did this to test https://github.com/dimagi/commcare-cloud/pull/4058. The big surprise here, which I do in this PR, is that we can switch on staging, production, and india from using the letsencrypt cert on nginx to using any old self-signed cert (here we use the built in snake-oil cert), because with the ALB, the only traffic going to nginx is coming from the ALB on the internal network, with no direct outside connections.

From https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html
> you can use self-signed certificates or certificates that have expired.
> Because the load balancer is in a virtual private cloud (VPC),
> traffic between the load balancer and the targets is authenticated
> at the packet level, so it is not at risk of man-in-the-middle attacks
> or spoofing even if the certificates on the targets are not valid.

Anyway, I rolled this out and it worked like a charm

##### ENVIRONMENTS AFFECTED
staging


snopoke:
> we can switch on staging, production, and india from using the letsencrypt cert on nginx to using any old self-signed cert

that's neat

PR #4065, opened 2020-08-05
Add ability to straddle two rabbitmq instances 

dannyroberts:
This adds a new option to migrate rabbitmq to a new machine rather than upgrading in place. It's not as clean as I'd hoped: specifically while `CELERY_BROKER_READ_URL` can take a list of endpoints, it actually just uses the first one unless it fails; i.e. further endpoints in that list are failover endpoints, not equal members in the list. What that means is basically that we need some celery nodes to have these "straddling" settings and others to have the future setup (so that they're reading from _and_ writing to the new rabbitmq instance) until the old rabbitmq instance is fully drained.

The alternative to this is to set up a cluster and learn to do the upgrade one node at a time when that's possible. We may want to have both options in our tool belt.

##### ENVIRONMENTS AFFECTED
available to all


millerdev:
:leg: :rabbit: :rabbit2: :leg:

PR #4068, opened 2020-08-06
create commcare server role in tf

knguyen142:
https://trello.com/c/prhWvkky/37-hipaa-logging-improvements

We need to create a new `CommCareServerRole` that gives servers write access to the request response firehose stream.

Changes:
- create a `server/iam` module directory
  - `server/iam` module creates commcare server role, attaches the aws managed cloudwatch and managed instance roles, creates a role policy for the request response stream writes, and creates the instance profile to be used by `aws_instance`
- Passes in the newly created instance profile to aws instances

NOTE: `iam_instance_profile` is only used during instance creation according to the [docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#iam_instance_profile), so adding this attribute does not replace the iam role on instances. I had to manually change the role for formplayer.

Progress and Testing:

All resources were applied on staging, and formplayer1 was manually moved over to the new iam role while formplayer2 was on the old to compare. It looks like formplayer1 is still reporting cloudwatch monitoring metrics.

The new role looks like:

Screen Shot 2020-08-06 at 3 07 39 PM

### DIFF
--- a/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
+++ b/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
@@ -79,6 +79,11 @@ resource "aws_dlm_lifecycle_policy" "data_volume_backups" {
   }
 }
 
+module "server_iam_role" {
+  source = "./modules/server/iam"
+  formplayer_request_response_logs_firehose_stream_arn = "${module.logshipping.formplayer_request_response_logs_firehose_stream_arn}"
+}
+
 {% for server_spec in servers + proxy_servers %}
 {%- for server_name in server_spec.get_all_server_names %}
 module "server__{{ server_name }}" {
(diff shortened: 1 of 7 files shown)


knguyen142:
@dannyroberts Yeah but it makes it easier to roll out since we can do it one instance at a time.

knguyen142:
Also @dannyroberts I'm not able to see the role on other environments like Prod

Screen Shot 2020-08-07 at 2 19 01 PM

Can you verify that Prod and India are only `CloudWatchAgentServerPolicy` and `AmazonSSMManagedInstanceCore`?

dannyroberts:
Yeah that's right. They are actually currently all _just_ `AmazonSSMManagedInstanceCore`, but adding `CloudWatchAgentServerPolicy` is a necessary step for taking the other general log shipping proof of concept to production, so this looks good.

PR #4069, opened 2020-08-07
update airflow to support py3

rohit-dimagi:
* https://dimagi-dev.atlassian.net/browse/IIO-929
##### ENVIRONMENTS AFFECTED

* ICDS , ICDS-STG


rohit-dimagi:
FYI If the version is changed in the task or there are different `virtualenv_python` attributes in the same task, then only the first one works and is never updated. workaround is to remove old version manually. 
Issue reported - https://github.com/ansible/ansible/issues/59350

PR #4071, opened 2020-08-07
Bug/new_environment.md document

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

Monolith


prem-fissionhq:
@esoergel 
1. The vault passwords generating section is placed in the wrong location. It was placed before creating a monolith environment, so the script will fail. I have corrected it. 
2. the line "You may be prompted for the Ansible vault password that you entered earlier." is not necessary since it observed that the update_known_hosts command haven't prompted for vault password anytime.

Please let me know if you have any suggestions/queries. Thanks

PR #4074, opened 2020-08-08
Resize proxy2-production

dannyroberts:
We recently added a second proxy node for the CPU.
Reducing the memory so that combined it is what this alone was before,
and so the two machines are the same.

(already applied)
##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -442,7 +442,7 @@ servers:
 
 proxy_servers:
   - server_name: "proxy2-production"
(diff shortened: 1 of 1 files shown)


biyeun:
@dannyroberts i approved...switched tasks...came back to chrome...then accidentally merged this one in thinking I was on the dimagi.com tab. this is what I get for working later in the day 😫  but I think it's ok regardless? anyway just FYI.

dannyroberts:
Yup, no worries this is good

PR #4084, opened 2020-08-12
update pub key

mkangia:
Updating my own pub key to move to new computer


satyaakam:
Why create a new one just copy the same key across devices

mkangia:
@satyaakam I had been using the same one, I thought it would be also a good practice to reset them as I set things up again.

PR #4088, opened 2020-08-13
rotate sentry keys

snopoke:
https://dimagi-dev.atlassian.net/browse/SAAS-11232

SENTRY_DSN
SENTRY_API_KEY
FORMPLAYER_SENTRY_DSN

### Environments impacted
staging, production, india


snopoke:
Note that for the Sentry API keys I've created internal integrations for each env and used their API keys instead of using keys tied to a user.

PR #4092, opened 2020-08-13
Rotate production s3 key

dannyroberts:
No specific reason to think this has been compromised, but it's a good time to rotate all the keys in the account

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
became part of https://github.com/dimagi/commcare-cloud/pull/4094

PR #4102, opened 2020-08-14
scale sentry workers

rohit-dimagi:
sometimes sentry stops processing altogether because of too many requests, this has happened a few times. if this fix doesn't work we have to apply quota limit instead.
##### ENVIRONMENTS AFFECTED

ICDS


snopoke:
I think we should apply a quota limit as well

rohit-dimagi:
Yeah this didn't worked. i have put a 5000 per minute limit on icds project.

PR #4114, opened 2020-08-18
Prefactor: Move ansible-vault code into a "secrets backend"

dannyroberts:
This is a set of changes that should hopefully make it easier to introduce an alternative secrets management method to the "locally stored ansible-vault" method we currently use.

This PR:
- makes our usage of secrets from the vault file a bit more uniform
- compiles an explicit list of known secrets that can be in the vault file
- replaces the big block of secrets assignments in all.yml with equivalent assignments generated from the explicit list of known secrets
- Pulls the assignment generation as well as all other code that supports the use of ansible-vault as a secrets management method into a new class `AnsibleVaultSecretsBackend`, which is accessed through an instance available at `environment.secrets_backend`

There will likely be more refactoring to do around the edges once we actually try to write a second implementation, but this is the main reorganization that will make that possible. The alternative implementation I am thinking of will use AWS Secrets Manager as the storage backend, but I can imagine ICDS someday wanting to move to another system as well.

The motivation for this change is to put some separation between developers' day-to-day usage and the secrets themselves. Having the secrets stored encrypted on all developers' laptops—but easily decrypt-able by developers with the key is just asking for a leak, as well as other tempting but bad practices like using a copy of an environments vault file as the starting point for a new environment.

##### ENVIRONMENTS AFFECTED
all (but it's an in place refactor that shouldn't affect anything


src/commcare_cloud/environment/secrets/backends/ansible_vault/tests/expected-generated-variables.yml
```diff
@@ -0,0 +1,64 @@
+AMAZON_S3_ACCESS_KEY: "{{ localsettings_private.AMAZON_S3_ACCESS_KEY | default(None) }}"
```
snopoke, line 1: Do you think in the future we'd get rid of this mapping file and just have the secrets named correctly to start with?

src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
```diff
@@ -0,0 +1,147 @@
...
+from ansible.parsing.vault import AnsibleVaultError
+from ansible_vault import Vault
+from memoized import memoized
+from six.moves import shlex_quote
+
+from commcare_cloud.environment.paths import ANSIBLE_DIR
+from commcare_cloud.environment.secrets.secrets_schema import get_known_secret_specs
+
+
+class AnsibleVaultSecretsBackend(object):
```
snopoke, line 17: A lot of this class is very specific to ansible vault storage. Is the long term plan to make this a base class with just the public interface or else a composite class and override / delegate the functionality?

src/commcare_cloud/environment/secrets/backends/ansible_vault/tests/expected-generated-variables.yml
dannyroberts: Yeah I was thinking that, although there's some value in setting the default to None. I didn't really decide on an approach here for that, though I do think the split between what's in `secrets` and what's in `localsettings_private` is sort of arbitrary and we'd do well to just get rid of that namespace distinction eventually

src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
dannyroberts: Yeah the plan is to make a base class with a public interface based on this. There will be a few more renamings of public functions here that are too vault-specific, but the overall structure is mostly right (or at least it's a good start)

src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
```diff
@@ -0,0 +1,147 @@
...
+    def __init__(self, name, vault_file_path, record_to_datadog=False):
+        self.name = name
+        self.vault_file_path = vault_file_path
+        self.record_to_datadog = record_to_datadog
+        self.should_send_vault_loaded_event = True
+
+    def prompt_user_input(self):
```
esoergel, line 24: nit: something like `prompt_for_password` might be more obvious

src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
```diff
@@ -0,0 +1,147 @@
...
+        return {
+            'ANSIBLE_VAULT_PASSWORD': self._get_ansible_vault_password_and_record(),
+        }
+
+    @staticmethod
+    def get_generated_variables():
+        secret_specs = get_known_secret_specs()
+        secret_specs_by_name = {secret_spec.name: secret_spec for secret_spec in secret_specs}
+        generated_variables = {}
+        for secret_spec in secret_specs:
```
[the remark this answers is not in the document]
dannyroberts: Hmm that's interesting. My first instinct was that this is going to be implemented differently depending on the secrets backend; this function is a core thing that'll be different between secrets backends because it's about telling ansible how to fetch each of the secrets and assign it to a variable. But on second thought, I think all the expression building (e.g. `| default(None)`) after the basic fetch syntax (e.g. `localsettings_private.SECRET_KEY`) might be pretty fixed regardless of the implementation. I'll think about what the common denominator is when I'm putting together the first other backend.

src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
dannyroberts: The reason I chose something generic here is that depending on the backend you might prompt for a password here (as the ansible-vault backend does) or for no input at all (if you're fetching secrets from a service where authorization is given based on being on an internal network or making the request from a specific machine) or to just make sure the user knows what they're doing "Do you want to access secrets for staging? If so please type 'staging':" or whatever else happens to be necessary for a particular secrets backend. The point of the function is just to give the calling code a way to force any potential user interaction to happen up front, whatever it may be.

esoergel: huh, okay, thanks for the explanation

src/commcare_cloud/environment/secrets/secrets.yml
```diff
@@ -0,0 +1,168 @@
+- name: AIRFLOW_FLASK_KEY
```
millerdev, line 1: Is it necessary to add all new secret names to this file to have them be accessible by commcare-cloud, or do reasonable defaults kick in for new secrets that are not listed in here? If new secrets must be added to this file, is that obvious (in the error message, for example) if this step is missed/forgotten?

In case it wasn't obvious from the question, reasonable defaults would be nice.

dannyroberts: That's really good feedback and something that slipped my mind to address. The way it works right now is somewhat analogous to how it works before: previously you could add any variable you wanted to vault.yml and it would be available in ansible contexts, but if you wanted to check for the value and assign it a default if it's missing, then you had to add a line to `all.yml`; now a line in this file plays the role of a line in `all.yml`.

Now, once we're using a different secrets backend, for a secret to be accessible from either backend it will need to go in this file. That's because another backend might fetch secrets in ansible (the one I'm considering will use https://docs.ansible.com/ansible/latest/plugins/lookup/aws_secret.html for example) and so there are no secrets passed in directly as ansible variables using `-e` as is done with vault.yml currently.

I considered maybe having the ansible-vault backend print out info messages saying

> your vault file contains an unknown secret "MY_CUSTOM_SECRET". Consider removing it if it's not needed or adding it to `src/commcare_cloud/environment/secrets/secrets.yml` if it is

or doing some kind of error checking like that, but I didn't want to do something too intrusive/annoying in the first pass. For now I'll quickly document this and then keep a look out for convenient places in the workflow to link people to the documentation.

dannyroberts: This is hopefully a good start https://github.com/dimagi/commcare-cloud/pull/4122, though docs are one thing and getting people to actually know to **read** them is another. I still intend to sneak a link to that in somewhere that'll be visible at the moment that it'll be useful to someone (like when a new undocumented secret is detected or something)

PR #4116, opened 2020-08-18
set ownership and mode on release directory

snopoke:
Currently the deploy depends on the ability to read files (or at
least check for their existence) inside the releases folder without
using 'sudo'.

Depending on the system defaults files may be created without the
necessary permissions and ownership.

##### ENVIRONMENTS AFFECTED
all

@shyamkumarlchauhan this should fix the icds-sandbox deploy (I'm testing it now)


snopoke:
The alternate to this is to make sure all commands are run as the `cchq` user

snopoke:
It does seem that most tasks already use `sudo` - I might take a stab at getting this to work without having to make things world readable

PR #4124, opened 2020-08-20
Changes during pgsynclog0-production fire

dannyroberts:
See commit messages for notes on how they're applied.

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -8,7 +8,7 @@ postgres_override:
 
 pgbouncer_override:
   pgbouncer_default_pool: 290
-  pgbouncer_max_connections: 3000
+  pgbouncer_max_connections: 10000
   pgbouncer_pool_mode: transaction
   pgbouncer_pool_timeout: 1
   pgbouncer_reserve_pool: 5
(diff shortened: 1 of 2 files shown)


environments/production/postgresql.yml
```diff
@@ -8,7 +8,7 @@ postgres_override:
 
 pgbouncer_override:
   pgbouncer_default_pool: 290
-  pgbouncer_max_connections: 3000
+  pgbouncer_max_connections: 10000
```
biyeun, line 11: just curious, what was your reasoning for tripling the pgbouncer_max_connections vs the 4x improvement for `instance_type`?

dannyroberts: There's no super data-driven reason for it. I tend to raise limits exponentially (i.e. by a percentage rather than a fixed amount), and tend to follow round number near-exponential patterns like (1, 2, 5, 10, 20, 50, 100, 200, 500, 1000...) or (1, 3, 10, 30, 100, 300, 10000). So I saw 3,000 and just went to the next number in the 1-3-10 pattern. Didn't give it much more thought than that to be honest

PR #4127, opened 2020-08-21
Add command in firefighting docs for killing formplayer sockets to solve lock issue

dannyroberts:
Adding something that helped recently to firefighting docs


ctsims:
This does largely match what we did although I think it might be useful to clarify that your command doesn't actually terminate **all** live socket connections, it kills a heuristically identified set of connections that happened to match up to what we saw were the stuck ones. I wouldn't want someone to think they are swinging a really broad hammer if they aren't.

dannyroberts:
Good point. Updated!

PR #4128, opened 2020-08-21
Make secrets backend configurable in meta.yml

dannyroberts:
It's on top of this change that I'll start testing a new backend based on [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/) and the [builtin `aws_secrets` ansible `lookup` plugin](https://docs.ansible.com/ansible/latest/plugins/lookup/aws_secret.html).

##### ENVIRONMENTS AFFECTED
none


src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
```diff
@@ -12,53 +12,62 @@
...
+    @classmethod
+    def from_environment(cls, environment):
+        try:
+            datadog_enabled = environment.public_vars.get('DATADOG_ENABLED')
+        except IOError:
+            # some test envs don't have public.yml
+            datadog_enabled = False
+
+        try:
+            commcare_cloud_use_vault = environment.public_vars.get('commcare_cloud_use_vault', True)
```
snopoke, line 37: with the ability to support different backends you could extract this out into a different class instead of customizing the vault backend:

```python
class EnvFileSecretsBackend(AbstractSecretsBackend)
    pass

class AnsibleVaultSecretsBackend(EnvFileSecretsBackend):
   pass
```

Just a thought

snopoke (approved):

src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
dannyroberts: That's not a bad idea. I'll hold off doing anything on that for now as I get this new backend working, but the next time we're changing something around that might be a good time to pull it out. At the very least it'd give us a place to write some doc strings about how you should never use it except for dummy environments like local or test environments

PR #4131, opened 2020-08-24
Change production compaction settings

dannyroberts:
We have a routine issue where ever couple weeks couch disk usage goes about 80% and I can resolve it by manually running compaction. Compaction is automatically running continuously which makes that really confusing. Last week I tested a hypothesis that the threshold, set at 6% fragmentation, was too low, and thus too many view and data files were being targeted for compaction at once, resulting in the compaction process not picking the most useful files to start with and not being able to keep up with the queue of things presumably needing compaction. To test this, I set couch14's threshold to 20% instead of 6%, telling it to focus on files with a higher bang for the buck. (If I set that too high, it'll wait too long to compact, so I still believe there's a sweet spot.)

The results were pretty conclusive, with couch14 immediately showing better disk usage than the other two machines and better able to maintain a flat level of disk usage over time once it reached equilibrium:
Screen Shot 2020-08-24 at 10 18 43 AM

This PR applies that change to all 3 production couch machines.

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -79,7 +79,7 @@ couchdb2:
   password: "{{ COUCH_PASSWORD }}"
 
 couchdb_compaction_settings:
-  _default: '[{db_fragmentation, "6%"}, {view_fragmentation, "6%"}, {parallel_view_compaction, true}]'
+  _default: '[{db_fragmentation, "20%"}, {view_fragmentation, "20%"}, {parallel_view_compaction, true}]'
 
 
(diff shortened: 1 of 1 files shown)


dannyroberts:
Unfortunately I think we can only graph the number of active tasks, not the number of db files over the fragmentation threshold _waiting_ to be compacted. (I double checked the metrics we collect in datadog.)

I'm not sure where a good place to put that is, but your summary's right

PR #4147, opened 2020-09-01
Add a production formplayer machine with a new configuration

dannyroberts:
I'm going to try this out with one machine. Everything I'm seeing in our usage and stats points to this being a reasonable configuration, but since you never know until a new configuration actually gets production traffic, I'm going to start cautiously with a single instance.

If there are no issues here, I will increase the count here and start removing formplayer_a machines to eventually replace all existing machines a few phases.

##### ENVIRONMENTS AFFECTED
production


environments/production/app-processes.yml
```diff
@@ -3,7 +3,7 @@ django_command_prefix: ''
 celery_command_prefix: ''
 gunicorn_workers_static_factor: 0
 gunicorn_workers_factor: 2
-formplayer_memory: "30g"
+formplayer_memory: "15g"
```
knguyen142, line 6: Just to clarify, this setting applies to both `formplayer_a` and `formplayer_b`, but you're going to just target the runs to the b's?

dannyroberts: That's right, yeah I should have said that. That said, assuming there are no issues with formplayer_b's setup, it would probably be safe to roll this out to other machines as well

PR #4148, opened 2020-09-02
Making the airflow repo private

ShubhamBansal1997:
Since we are more moving icds code private so the same, pipes repo (ariflow) is also turned private which in turns needs change in the deploy process to pull the private pipes repo.

##### ENVIRONMENTS AFFECTED
 - icds-staging
 - icds

### DIFF
--- a/src/commcare_cloud/ansible/roles/airflow/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/airflow/tasks/main.yml
@@ -10,6 +10,17 @@
     - "{{ airflow_home }}"
     - "{{ airflow_code_dir }}"
 
+- name: add deploy keys
+  become: yes
+  copy:
+    src: "{{ env_files_dir }}/{{ item.key }}"
+    dest: "{{ cchq_home }}/.ssh/{{ item.value }}"
+    owner: "{{ cchq_user }}"
+    group: "{{ cchq_user }}"
+    mode: 0600
+  loop: "{{ deploy_keys|dict2items }}"
+  when: deploy_keys is defined
+
 - name: Pull full Airflow source
   git:
     repo: "{{ airflow_repository }}"
@@ -18,6 +29,8 @@
     recursive: yes
     accept_hostkey: yes
     update: yes
+    deploy_key: deploy_key.pem
+    is_private: true
   become: yes
   become_user: "{{ cchq_user }}"
 
--- a/src/commcare_cloud/ansible/roles/airflow/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/airflow/defaults/main.yml
@@ -1,6 +1,6 @@
 airflow_code_dir: "{{ airflow_home }}/pipes"
 airflow_dag_dir: "{{ airflow_home }}/dags"
 airflow_version: "master"
-airflow_repository: https://github.com/dimagi/pipes.git
+airflow_repository: git@github.com:dimagi/pipes.git
 airflow_dag_concurrency: 16
 airflow_parallelism: 32


snopoke (commented):
You'll also need to update the fabric task [update_airflow](https://github.com/dimagi/commcare-cloud/blob/811e87f9b015c2c7db800c23b27cac3c018fe7ae/src/commcare_cloud/fab/operations/airflow.py#L9) (or maybe that should be removed if it isn't used.

You can see the approach that was taken with the other fabric tasks here: https://github.com/dimagi/commcare-cloud/blob/c1c5dc9f56fe8b5f137fc9742134644c5d5d710f/src/commcare_cloud/fab/operations/release.py#L46-L49

ShubhamBansal1997:
> You'll also need to update the fabric task [update_airflow](https://github.com/dimagi/commcare-cloud/blob/811e87f9b015c2c7db800c23b27cac3c018fe7ae/src/commcare_cloud/fab/operations/airflow.py#L9) (or maybe that should be removed if it isn't used.
> 
> You can see the approach that was taken with the other fabric tasks here:
> 
> https://github.com/dimagi/commcare-cloud/blob/c1c5dc9f56fe8b5f137fc9742134644c5d5d710f/src/commcare_cloud/fab/operations/release.py#L46-L49

Updated, please review again

src/commcare_cloud/ansible/roles/airflow/tasks/main.yml
```diff
@@ -10,6 +10,17 @@
     - "{{ airflow_home }}"
     - "{{ airflow_code_dir }}"
 
+- name: add deploy keys
```
snopoke, line 13: I don't think this is necessary since the keys are already added by the commcare role (which includes the airflow machines)

ShubhamBansal1997: added a new key

snopoke: this task is still not necessary - it's a duplicate of the one in commcarehq role which will already add the keys to the airflow machine.

I suppose if you haven't run the commcare role then the key won't exist so perhaps ok to duplicate - maybe we should pull it out into a separate file so we can reuse it (can do in new pr)

src/commcare_cloud/fab/operations/airflow.py
```diff
@@ -1,13 +1,20 @@
...
-from fabric.context_managers import cd
+from fabric.context_managers import cd, shell_env
 
 from ..const import ROLES_AIRFLOW
 
+
 @roles(ROLES_AIRFLOW)
 def update_airflow():
-    with cd(env.airflow_code_root):
+    deploy_key = 'deploy_key_airflow.pem'
```
snopoke, line 13: you need to update `environmetns/<env>/meta.yml` to list the new deploy key:

```diff
 deploy_keys:
   deploy_key.vault: deploy_key.pem
+  deploy_key_airflow.vault: deploy_key_airflow.pem
```

ShubhamBansal1997: added

src/commcare_cloud/ansible/roles/airflow/tasks/main.yml
```diff
@@ -18,6 +29,8 @@
     recursive: yes
     accept_hostkey: yes
     update: yes
+    deploy_key: deploy_key_airflow.pem
+    is_private: true
```
snopoke, line 33:
```suggestion
    key_file: deploy_key_airflow.pem
```

snopoke (changes requested):
One small change, sorry I didn't notice before.

src/commcare_cloud/ansible/roles/airflow/tasks/main.yml
ShubhamBansal1997: updated

snopoke (approved):

ShubhamBansal1997:
Tested on icds-staging, merging in

PR #4151, opened 2020-09-02
Add command to update a single user's public key

millerdev:
This saves a lot of time when a user's public key changes. The `update-users` command, by comparison, takes much much longer to run.

Usage: `cchq <env> update-user-key <username>`


dannyroberts:
I think you just need to run `make` and commit the result to fix the tests

PR #4155, opened 2020-09-02
Add aws secrets backend and move staging to it

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-11241

This one's kind of a big one but I tried to keep the commit history fairly clean. I've been nursing this PR for a while now and I'm sure there are some rough edges but I think I want to put it out there and get it on the path to getting merged.

This PR lets environments using AWS put `secrets_backend: aws-secrets` in their `meta.yml` file and switch to using AWS Secrets Manager instead of ansible-vault. The new `cchq <env> migrate-secrets <from_backend>` command lets you prime your new secrets backend (the one specified in `meta.yml`) with the values from the old one (`<from_backend>`).

This PR does not yet remove staging vault to leave a window for other PRs in the process of changing the staging vault file. I'll subsequently check to make sure they're the same, reconcile if they're not, and delete the vault file.

## migration/transition steps for staging

The migration/transition steps for staging are:
- [ ] Merge this PR
- [ ] Directly after that on an updated master, runs `cchq staging migrate-secrets ansible-vault` one final time
- [ ] Delete `environments/staging/vault.yml`, commit that, and make a PR for it

I (Danny) will make sure we do those

That's kind of it. There shouldn't be anything extra other devs have to do or know to continue interacting with staging normally. This PR contains all the changes needed to make staging seamlessly pull its secrets into memory form ̄WS Secrets Manager.

## Changes to dev workflows for aws-secrets (vs ansible-vault)

The only difference for you as a dev will be that
- you won't be asked for a vault password
- instead you will be asked to do the CLI-triggered AWS single sign on workflow (that some other commands already use) now and then
- you can no longer edit secrets using `ansible-vault view|edit environments/<env>/vault.yml`. Instead use `cchq <env> secrets view|edit <SECRET>`

@sanjay2916 FYI

##### ENVIRONMENTS AFFECTED
staging


knguyen142:
Can you outline the migration/transition steps?

dannyroberts:
@knguyen142 Updated the description

src/commcare_cloud/ansible/plugins/lookup/cached_aws_secret.py
```diff
@@ -0,0 +1,32 @@
+from ansible.plugins.lookup import aws_secret
+
+__metaclass__ = type
+
+DOCUMENTATION = aws_secret.DOCUMENTATION
+RETURN = aws_secret.RETURN
+
+
+class LookupModule(aws_secret.LookupModule):
+    _secrets_cache = {}
```
millerdev: You're adding caching here while the implementation of `AwsSecretsBackend.get_secret` constructs a new session and client on every call.

https://github.com/dimagi/commcare-cloud/blob/f5c96e856c5b0485f9309cb0a173f24781e30310/src/commcare_cloud/environment/secrets/backends/aws_secrets/main.py#L43-L46

Is it possible that that overhead is why it was slow enough that you needed to add caching? For example, is it possible that the client already does caching and this new cache you've added is not needed?

src/commcare_cloud/ansible/plugins/lookup/cached_aws_secret.py
```diff
@@ -1,27 +1,34 @@
...
                 print('Fetching', terms)
                 try:
                     value = super(LookupModule, self).run(terms, variables, **kwargs)
                 except Exception as e:
                     print('Caching error: ', e)
                     value = e
-                self._secrets_cache[terms] = value
-
-            value = self._secrets_cache[terms]
+                self.set_cache(term, value, inventory_dir=variables['inventory_dir'])
```
millerdev: Why save the exception in the cache?

src/commcare_cloud/ansible/plugins/lookup/cached_aws_secret.py
```diff
@@ -30,3 +37,33 @@ def run(self, terms, variables, **kwargs):
...
         else:
             return value
+
+    def get_cache(self, term, inventory_dir):
+        try:
+            with open(self._secrets_cache_filename(term, inventory_dir=inventory_dir), 'rb') as f:
+                contents = f.read()
+                if contents == b'FETCHING':
+                    time.sleep(1)
+                    return self.get_cache(term, inventory_dir=inventory_dir)
```
millerdev: The logic here is convoluted. It seems like it should be possible to refactor to remove the sleep and recursive call, which should clear it up quite a bit.

millerdev (commented):
It looks like the only tests are for happy-path secrets lookup, but there is a fair amount of error conditions handled in the code. Would it be possible to test more of those error conditions?

src/commcare_cloud/environment/secrets/backends/aws_secrets/main.py
```diff
@@ -21,5 +23,13 @@ def get_extra_ansible_env_vars(self):
         return {'AWS_REGION': self.environment.terraform_config.region}
 
     def get_secret(self, var):
-        # Will implement this shortly
+        from commcare_cloud.commands.terraform.aws import aws_sign_in
+        return (
+            boto3.session.Session(profile_name=aws_sign_in(self.environment)).client(
+                'secretsmanager', region_name=self.environment.terraform_config.region
+            )
+            .get_secret_value(SecretId='commcare-{}/{}'.format(self.environment.name, var))['SecretString']
```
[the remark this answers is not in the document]
dannyroberts: Oops, thanks. I think I'll opt to just use `secret_name_prefix` throughout

src/commcare_cloud/commands/secrets.py
```diff
@@ -36,3 +38,29 @@ def _secrets_edit(self, environment, secret_name):
...
+    )
+
+    arguments = (
+        Argument(dest='from_backend'),
+    )
+
+    def run(self, args, unknown_args):
+        environment = get_environment(args.env_name)
+        from_backend = all_secrets_backends_by_name[args.from_backend].from_environment(environment)
+        to_backend = environment.secrets_backend
```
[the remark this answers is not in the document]
dannyroberts: Yeah sure I'll add that

src/commcare_cloud/environment/secrets/backends/aws_secrets/main.py
```diff
@@ -25,14 +27,15 @@ def get_extra_ansible_env_vars(self):
 
     def get_secret(self, var):
         from commcare_cloud.commands.terraform.aws import aws_sign_in
-        return (
+        return json.loads(
             boto3.session.Session(profile_name=aws_sign_in(self.environment)).client(
                 'secretsmanager', region_name=self.environment.terraform_config.region
             )
```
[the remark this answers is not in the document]
dannyroberts: 🤦 thanks

src/commcare_cloud/ansible/plugins/lookup/cached_aws_secret.py
dannyroberts: I think no, but let me try to explain.

This `cached_aws_secret` plugin is called from ansible, while `AwsSecretsBackend.get_secret` is called from commcare-cloud. In commcare-cloud, we have full control of the code paths and it'll be easy to add caching later if it's needed, but typically `AwsSecretsBackend.get_secret` is called zero to one times within commcare-cloud, and maybe up to a handful in steady state normal operations; ansible on the other hand makes tons of repeat calls to the plugin: once per secret per task per machine. So if there is one task (e.g. copy localsettings) that needs 20 secrets on 12 machines, followed by another task (e.g copy applications.properties for formplayer) that needs a subset of those same secrets, say 10 secrets on 40 machines, then it'll make a total of 20 x 12 + 10 x 40 = 640 calls for secrets. With the caching I do here, we can instead make just 20 calls.

src/commcare_cloud/ansible/plugins/lookup/cached_aws_secret.py
dannyroberts: The most common exception is that the secret doesn't exist. I want to cache that so that next time the same secret is asked for in the same ansible run, we already know it doesn't exist and don't need to make another API request

src/commcare_cloud/ansible/plugins/lookup/cached_aws_secret.py
dannyroberts: It's probably possible to remove the recursive call, but I don't think it's possible to remove the sleep (maybe reduce it, but not remove it). What's going on here is that during the execution of a task, ansible makes N forks, where N is the number of machines the task will run on. Each of these forks then more or less simultaneously calls this plugin for secrets, usually in the same order. Instead of N forks collectively making N calls for secret X, we want whichever gets there first to claim that responsibility by writing 'FETCHING' to the cache file, and for all the other forks to wait until a value other than 'FETCHING' appears in the cache file. This results in one request per secret rather than N requests per secret.

dannyroberts: I took a stab at refactoring this for better readability

src/commcare_cloud/ansible/plugins/lookup/cchq_aws_secret.py
```diff
@@ -0,0 +1,91 @@
...
+                    contents = f.read()
+                    if contents == LOOKUP_IN_PROGRESS_ON_ANOTHER_FORK:
+                        # If another fork is in the progress of looking this up
+                        # wait a short while and then try again from the top of the while True loop
+                        time.sleep(.100)
+                        continue
+                    return cPickle.loads(Fernet(self._encryption_key()).decrypt(contents))
+            except (IOError, InvalidToken, cPickle.UnpicklingError):
+                with open(self._secrets_cache_filename(term, inventory_dir=inventory_dir), 'wb') as f:
+                    f.write(LOOKUP_IN_PROGRESS_ON_ANOTHER_FORK)
```
millerdev, line 72: There's a race condition here where two processes both try to read the cache, but it's empty, so they both start writing. Is there any chance that could cause a bad outcome?

src/commcare_cloud/ansible/plugins/lookup/cchq_aws_secret.py
```diff
@@ -0,0 +1,91 @@
...
+    to encrypt values and write them to disk in files <env_dir>/.generated/.
+    After the run is over, the encryption key is lost to time and the files are useless
+    and indecipherable even by the person who ran it.
+    """
+    def run(self, terms, variables, **kwargs):
+        terms = tuple(terms)
+        term, = terms
+        value = self.get_cache(term, inventory_dir=variables['inventory_dir'])
+        try:
+            if value is Ellipsis:
```
millerdev, line 36: Nit: this is an abuse of `Ellipsis`, but you might as well use the syntax as it was intended.
```suggestion
            if value is ...:
```

src/commcare_cloud/ansible/plugins/lookup/cchq_aws_secret.py
```diff
@@ -0,0 +1,91 @@
...
+                    if contents == LOOKUP_IN_PROGRESS_ON_ANOTHER_FORK:
+                        # If another fork is in the progress of looking this up
+                        # wait a short while and then try again from the top of the while True loop
+                        time.sleep(.100)
+                        continue
+                    return cPickle.loads(Fernet(self._encryption_key()).decrypt(contents))
+            except (IOError, InvalidToken, cPickle.UnpicklingError):
+                with open(self._secrets_cache_filename(term, inventory_dir=inventory_dir), 'wb') as f:
+                    f.write(LOOKUP_IN_PROGRESS_ON_ANOTHER_FORK)
+                return Ellipsis
```
millerdev, line 73:
```suggestion
                return ...
```

src/commcare_cloud/ansible/plugins/lookup/cchq_aws_secret.py
```diff
@@ -0,0 +1,91 @@
...
+    Adds commcare-cloud specific caching to aws_secret lookup plugin for big speedup
+    Also assumes the value is json and returns the string json-decoded unlike aws_secret.
+
+    To cache even between forks of the same process/run, we use a one-time process-wide key
+    to encrypt values and write them to disk in files <env_dir>/.generated/.
+    After the run is over, the encryption key is lost to time and the files are useless
+    and indecipherable even by the person who ran it.
+    """
+    def run(self, terms, variables, **kwargs):
+        terms = tuple(terms)
```
millerdev, line 32: nit: curious why this line is necessary?

src/commcare_cloud/ansible/plugins/lookup/cchq_aws_secret.py
```diff
@@ -0,0 +1,91 @@
...
+        value = self.get_cache(term, inventory_dir=variables['inventory_dir'])
+        try:
+            if value is Ellipsis:
+                logging.debug('Fetching secret {}'.format(terms))
+                try:
+                    value = super(LookupModule, self).run(terms, variables, **kwargs)
+                except Exception as e:
+                    logging.debug('Caching error fetching secret: {}'.format(e))
+                    value = e
+                self.set_cache(term, value, inventory_dir=variables['inventory_dir'])
```
millerdev, line 43: It's not clear what `term` and `value` are. Is there one term per secret, or something else? If one per secret, does that mean there is a separate cache file per secret?

src/commcare_cloud/ansible/plugins/lookup/cchq_aws_secret.py
```diff
@@ -0,0 +1,91 @@
...
+        except Exception as e:
+            logging.warn('There was an error fetching or caching the secret: {}'.format(e))
+            raise
+
+        if isinstance(value, Exception):
+            raise value
+        else:
+            return [json.loads(item) for item in value]
```
millerdev, line 51: Where is the corresponding `json.dumps` for this `json.loads`?

millerdev (approved):
I'm still a little skeptical of the secrets caching. It seems like a problem that everyone would have, so why do we need a special solution?

Other than the "wait forever" comment, which can be addressed in a separate PR if you want, I think most of my comments are nitpicks or just trying to understand the code it works. I don't have a good reason to hold this up other than hesitance over what seems like an overly complicated solution so I'm approving (a bit reluctantly).

src/commcare_cloud/ansible/plugins/lookup/cchq_aws_secret.py
dannyroberts: In the case that two processes wrote to the same file at the same time, leaving it in a corrupted state, other threads would try to read it, fail to decrypt and unpickle the corrupted binary, and just opt to fetch it fresh themselves. That's obviously less ideal than the caching working, but my estimation is that it's fairly low and when it does happen there's no real catastrophic effect.

dannyroberts (commented):
> It seems like a problem that everyone would have, so why do we need a special solution?

Yeah I had the same reaction. But I dug around for any sign of the ability to turn on caching both across forks of the same tasks and across tasks, and neither was built in to the plugin. So either not that many people use this built-in ansible lookup plugin, or it's intended more for fetching a couple passwords once rather than the way we're using it; hard to say. But for sure I took no pleasure in whipping together this caching solution; it's just that without it something that should take 15-20s would take minutes. I tried seeing if there was some concurrency-safe shared memory construct we could use to cache values in memory between the forks, and everything I saw was both somewhat complicated and needed to be set up during the forking itself, something that's outside of our control here. So encrypting with a temporary shared key into files stored locally was the best thing I could come up with that worked, unfortunately

src/commcare_cloud/ansible/plugins/lookup/cchq_aws_secret.py
millerdev: I see. Would it be possible to use `aws_secret.LookupModule` here rather than `AwsSecretsBackend` (which does the `loads` in `_get_secret`)? That would make the symmetry more obvious.

dannyroberts: We offlined. The issue is that here https://github.com/dimagi/commcare-cloud/pull/4155/files#diff-2898f3b64f32105af2b0b4f8e1c92617R33 where the lookup expression is inserted to a variable that gets passed into ansible, our ability to conditionally decode json and handle the error that would come up is pretty limited, and I found moving it into here was the only way to get the desired behavior on good and bad inputs. It's possible that if we simplified the different default values secrets variables can take (like by making them all default to `None` rather than some to `None` and some to `''`) then we could move this back out and make this lookup plugin compatible with `aws_secret`.

In the meantime I will just better document the incompatibility through comments.

millerdev (approved):

dannyroberts:
@dimagi/team-commcare-hq @sanjay2916  This PR changes staging to use AWS Secrets manager instead of ansible vault. Here's how this affects you:
- I'm handling migrating existing secrets on master from ansible to AWS Secrets manager. If you have an open PR that modifies staging/vault.yml (I don't think there are any) you will have to apply that change again to the new backend
- To change staging secrets, use `cchq staging secrets edit MY_SECRET` when you previously would have edited the vault file and made a PR. (You can double check your work by running `cchq staging secrets  view MY_SECRET` afterwards.)
- If you have any issues running any commands on staging with failures fetching secrets, please let me know right away. In the meantime, try running it on the control machine which has access to the secrets through EC2 IAM roles instead of through personal access keys
- Where you were previously prompted for a vault password, you will no longer be. If you used that as a "gate" just be aware you won't be able to anymore. If this becomes an issue, we can make the choice to add back a gate there, but I decided to start with the simpleminded approach that fewer prompts is better

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -343,6 +347,18 @@ def aws_sign_in(environment, duration_minutes=DEFAULT_SIGN_IN_DURATION_MINUTES,
                                      force_new=force_new)
 
 
+def is_ec2_instance_in_account(account_id):
+    try:
+        aws_instance_identity_doc = requests.get(
+            'http://169.254.169.254/latest/dynamic/instance-identity/document',
```
snopoke, line 353: how come this IP is hard coded? Is it a global AWS endpoint?

src/commcare_cloud/ansible/setup_pg_logical_replication.yml
```diff
@@ -144,7 +144,7 @@
       postgres
       postgresql_query:
         db: '{{ item.target_db_name }}'
-        query: "GRANT USAGE ON SCHEMA pglogical TO {{ secrets.POSTGRES_USERS.commcare }};"
+        query: "GRANT USAGE ON SCHEMA pglogical TO {{ postgres_users.commcare }};"
```
snopoke, line 147: @dannyroberts I assume this should also be applied to other places e.g. inside env config files for ICDS

dannyroberts: Yup. I didn't make a big stink about it because for environments using the ansible-vault backend, `secrets.*` syntax should continue to work as before, so if that's strictly in config files, I'd maybe encourage switching syntax but it's up to you

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: It's an AWS EC2 instance thing described here https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html. But basically that IP address is available from EC2 instances but returns information specific to the EC2 instance you are requesting it from. To quote that link
> The IP address 169.254.169.254 is a link-local address and is valid only from the instance.

Sorry for the very late reply!

PR #4159, opened 2020-09-04
Explicitly grant privs on all dbs

dannyroberts:
**Note** I opened this prematurely and then realized it didn't actually work to fix the issue on staging. Will come back to it

Previously I believe we relied on the the user being granted ownership over the db and then inheriting table privileges by default when that user created all the tables. On staging we switched users and the new user had no way to get privileges even though they became listed as the db owner. This PR
- Explicitly grants the new user all table privileges (in vault.yml)
- Has the code actually correctly apply those privileges to all dbs

The privs config (since you can't see it here) is:

```
      privs:
        - privs: ALL
          objs: ALL_IN_SCHEMA
          type: table
        - privs: ALL
          objs: ALL_IN_SCHEMA
          type: sequence
```

##### ENVIRONMENTS AFFECTED
staging mostly, but all


dannyroberts:
@sanjay2916

PR #4161, opened 2020-09-04
Fix docs link (pillwotop => pillowtop)

dannyroberts:
Here's an easy one!


knguyen142 (approved):
Where are the tests?! 😜

PR #4162, opened 2020-09-07
update kafka log config

snopoke:
DailyRollingFileAppender has been observed to exhibit synchronization issues and data loss.

In addition to that it makes it very hard to use in conjunction with logrotate. 

This PR updates the Kafka logging configuration to use a vanilla `RollingFileAppender` which will also take care of cleaning up old logs.

##### ENVIRONMENTS AFFECTED
all


snopoke:
Already rolled out to `icds-staging`

rohit-dimagi (approved):

rohit-dimagi:
Nice

PR #4169, opened 2020-09-10
Use suggested max file descriptors for ES

dannyroberts:
Found the following in ES log output:

> max file descriptors [32000] for elasticsearch process likely too low, consider increasing to at least [65536]

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
@@ -34,7 +34,7 @@ RestartSec=10
 StandardError=inherit
 
 # Specifies the maximum file descriptor number that can be opened by this process
-LimitNOFILE=32000
+LimitNOFILE=65536
 
 # Disable timeout logic and wait until process is stopped
 TimeoutStopSec=0


calellowitz (approved):

sravfeyn (approved):

snopoke (approved):

dannyroberts:
Ugh okay this is dangerous because [when we change this file we also automatically restart elasticsearch](https://github.com/dimagi/commcare-cloud/blob/2d29eb2e890404882dd906861327732702e7c2ce/src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml#L73-L89), and we do so cavalierly (by running `systemctl restart elasticsearch` on all nodes at once) rather than carefully (as in the es_rolling_restart playbook).

Any smart ideas for how to do this safely? I could replace the restart with a reload which is likely safe, but may also interfere with first-time setup. So I could restart it if `CCHQ_IS_FRESH_INSTALL` and reload it otherwise. Thoughts?

snopoke:
I've often had the same problem but haven't found a good general solution yet. I would likely roll this out manually by making the change in the using `cchq <env> run-module -b lineinfile ....` and then do the restart either manually or using the rolling restart playbook.

The other options is to allow skipping the handlers using some condition: https://stackoverflow.com/a/57677835/632517

PR #4173, opened 2020-09-11
Refuse to run terraform destroy

dannyroberts:
##### ENVIRONMENTS AFFECTED
saas envs

### DIFF
--- a/src/commcare_cloud/commands/terraform/terraform.py
+++ b/src/commcare_cloud/commands/terraform/terraform.py
@@ -46,6 +46,11 @@ class Terraform(CommandBase):
     )
 
     def run(self, args, unknown_args):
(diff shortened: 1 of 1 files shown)


dannyroberts:
@prem-fissionhq

knguyen142:
What is the reasoning for not allowing a destroy?

dannyroberts:
Just that it's a very destructive operation. If we ever ran it on production it would be a total disaster. It's an easy thing to change later and add a more flexible mechanism for preventing only unintended usage, but for now there's no time I can imagine needing to run this on staging, production, or india

knguyen142 (approved):

prem-fissionhq:
Thanks Danny for implementing it so quickly 🙂

PR #4184, opened 2020-09-15
Remove defunct prove-deploy build

dannyroberts:
The best little idea that never came to be. This build has sadly never passed, and we aren't able to get any value out of it. Rather than hold on, I think it's finally time to let it go. Any solution to the same problem (and I hear there's some work outside of SaaS on it) will likely end up being done differently anyway.

:(

##### ENVIRONMENTS AFFECTED
none


millerdev (approved):
:disappointed:

PR #4185, opened 2020-09-16
Test, refactor, and improve ssh args handling

dannyroberts:
Refactor the two places that pass in ssh options (primarily `UserKnownHostsFile`) to share code, testing and tidying up the behavior in the process. This is just a little cleanup, a little prefactoring before I add more behavior here.

Best reviewed commit by commit

##### ENVIRONMENTS AFFECTED
not env-specific


millerdev (approved):
Nice!

snopoke (commented):
@dannyroberts I'd be interested to hear more about your planned changes to auth and how that would impact deploy / commcare-cloud. We've been getting some questions about how to use LDAP or some other centralized auth system.

dannyroberts:
We are planning on using [AWS Systems Manager Session Manager](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager.html) (yes that's what the feature is called 🤦; sometimes also called "AWS SSM") which lets you use AWS access keys (which we get through AWS SSO's CLI integration that let's you get temporary access keys for use from the command line) to get a shell on a particular EC2 instance. Specifically here we'll be using AWS SSM's ssh-compatibility feature, which will let us use it as a drop in replacement both for the `ssh` command and for ansible's ssh connection. It's a little more complicated than simply adding more `-o` options to ssh, but that's part of it. Once we are on that will largely or entirely replace the need for a VPN for SaaS AWS envs. That's because it will look and feel the same, at the network level it's doing a very different thing: it's saying "Hey AWS, here's proof I'm allowed to ask for this, now give me a shell to EC2 instance i-xxxxx", as opposed to replying on a VPN to virtually connect your local network to the environment's private network ("VPC" in AWS speak).

There may be a couple more refactors before we're all the way there, and I'll make sure to ping you on all of them @snopoke in case it helps for what you're thinking about.

snopoke:
Thanks @dannyroberts that's helpful.

PR #4189, opened 2020-09-17
Fix X-Forwarded-For with set_real_ip_from

dannyroberts:
This adds a variable, trusted_proxies, to proxy.yml, and then uses it to configure nginx to set the `X-Forwarded-For` to the real client IP address as opposed to either (1) the full chain including whatever the user passed in or (2) just the upstream proxy's IP address.

https://dimagi-dev.atlassian.net/browse/SAAS-11335

##### ENVIRONMENTS AFFECTED
production, staging, india

### DIFF
--- a/environments/india/proxy.yml
+++ b/environments/india/proxy.yml
@@ -5,7 +5,5 @@ J2ME_SITE_HOST: 'j2me-india.commcarehq.org'
 tableau_server: 'tableau2.internal.commcarehq.org'
 TABLEAU_HOST: 'icds.commcarehq.org'
 primary_ssl_env: "india"
-
-# Nginx only accepts on cert, so cat the site cert and chain:
-# $ cat india.commcarehq.org.crt intermediate.crt > india.commcarehq.org.combined.crt
-letsencrypt_cchq_ssl: True
+trusted_proxies:
+  - 10.203.0.0/16
--- a/src/commcare_cloud/ansible/roles/nginx/templates/site.j2
+++ b/src/commcare_cloud/ansible/roles/nginx/templates/site.j2
@@ -48,6 +48,10 @@ proxy_cache_path {{ cache.path }} levels=1:2 keys_zone={{ cache.name }}:10m
 
 server {
 {% if trusted_proxies %}
+  # These lines change the value of $remote_addr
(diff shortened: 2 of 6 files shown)


ctsims (approved):
This matches my review of the docs

environments/india/proxy.yml
```diff
@@ -5,7 +5,5 @@ J2ME_SITE_HOST: 'j2me-india.commcarehq.org'
 tableau_server: 'tableau2.internal.commcarehq.org'
 TABLEAU_HOST: 'icds.commcarehq.org'
 primary_ssl_env: "india"
-
-# Nginx only accepts on cert, so cat the site cert and chain:
-# $ cat india.commcarehq.org.crt intermediate.crt > india.commcarehq.org.combined.crt
-letsencrypt_cchq_ssl: True
```
ctsims, line 11: Why did these get removed in this PR, by the way? I presume this was intentional?

src/commcare_cloud/ansible/roles/nginx/templates/site.j2
```diff
@@ -47,6 +47,13 @@ proxy_cache_path {{ cache.path }} levels=1:2 keys_zone={{ cache.name }}:10m
 {% endfor %}
 
 server {
+{% if trusted_proxies %}
+  real_ip_header X-Forwarded-For;
```
ctsims, line 51: One note: The PR description doesn't specifically say that this sets the forwarded for header, just that it adds the proxies, so it might be confusing for someone to run across.

environments/india/proxy.yml
dannyroberts: Yeah it was intentional. I guess kind of sloppy to do in the same commit, but this is all cruft ever since we changed how proxy certs are handled a while ago.

src/commcare_cloud/ansible/roles/nginx/templates/site.j2
dannyroberts: OK I added some comments here and I'll update the PR description

environments/india/proxy.yml
ctsims: No worries, just wanted to make sure I didn't ignore it in case it wasn't intentional

PR #4191, opened 2020-09-17
always force a TTY from ssh commands

snopoke:
I don't think there is any harm in doing this and it makes many more commands useful.


millerdev (approved):
I think this is fine. https://serverfault.com/a/593419/76543 has a good explanation of when and why to (not) use `ssh -t`

Test failures are related to the change.

snopoke:
@millerdev thanks for the link - considering the way that these commands are used I would expect it to be safe.

millerdev:
Agreed, it seems like a reasonable default. We may want to provide an option to turn it off in case someone were running a `commcare-cloud` command with cron or some other non-interactive process runner.

PR #4194, opened 2020-09-17
Fix tests and eliminate deprecation warning

millerdev:
Tests were failing because of `COMMCARE_CLOUD_DEFAULT_USERNAME` in my environment.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,24 @@
+from __future__ import absolute_import
+from mock import patch
+
+from commcare_cloud.commands.terraform.aws import StringIsGuess
+
+
+def setup_package():
+    global patches
+    patches = [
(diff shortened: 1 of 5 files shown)


tests/__init__.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import absolute_import
+from mock import patch
+
+from commcare_cloud.commands.terraform.aws import StringIsGuess
+
+
+def setup_package():
+    global patches
+    patches = [
+        patch.dict("os.environ", COMMCARE_CLOUD_DEFAULT_USERNAME=""),
```
dannyroberts, line 10: This is awesome, thank you. I've definitely hit this before too

dannyroberts (approved):

PR #4195, opened 2020-09-18
update separate envs repo if it exists

snopoke:
@mkangia this will make it easier to update both repos at the same time for ICDS

### DIFF
--- a/control/update_code.sh
+++ b/control/update_code.sh
@@ -32,9 +32,19 @@ function update_repo() {
     git submodule update --init --recursive
 }
 
-for repo in "commcare-cloud"
+repos="$HOME/commcare-cloud"
+
+# if environments points to a separate Git repo, update that as well
+if [ -n "$COMMCARE_CLOUD_ENVIRONMENTS" ] && \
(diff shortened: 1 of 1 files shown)


mkangia:
thanks @snopoke 

I ran into a similar double update challenge when I had to update requirements. Is that handled as well?

snopoke:
If you are using the icds-envs separately from commcare-cloud then all the requirements are installed by commcare-cloud. The other requirements are only necessary if you are using the icds envs standalone.

mkangia:
okay, thanks @snopoke, that means if one is using a separate repo of commcare-cloud, they would need to get into that directory to update requirements, similar to the challenge we had with code update that this PR resolves.
I am planning to just reset my control environment to be in a single directory as suggested in icds-commcare-environments since having two is less smooth.
Thanks for taking time for fixing this though. I leave it to you to decide to keep this change or discard it now.

snopoke:
@mkangia I still want to keep this as it's handy to have separate repos in cases where you are making changes to commcare-cloud (until we have a better workflow for handling that)

mkangia:
thanks @snopoke . Not well versed in shell so I'd defer to @dannyroberts for review on this.

dannyroberts (approved):

PR #4200, opened 2020-09-18
Logship /var/log/{syslog,auth.log,audit/audit.log}

dannyroberts:
##### ENVIRONMENTS AFFECTED
staging, production

### DIFF
--- a/src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
@@ -16,7 +16,7 @@
     register: cloudwatch_logs_conf
 
   - name: Reload cloudwatch agent if changed
-    shell: /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a append-config -m ec2 -s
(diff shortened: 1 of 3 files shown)


millerdev (approved):

knguyen142 (approved):

src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
```diff
@@ -16,7 +16,7 @@
     register: cloudwatch_logs_conf
 
   - name: Reload cloudwatch agent if changed
-    shell: /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a append-config -m ec2 -s
+    shell: /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s
```
knguyen142, line 19: Was `append` causing issues?  For some reason I had tried `fetch` but must've had different parameters because it wasn't working.

dannyroberts: Yeah append wasn't reloading the config when the config changed. `fetch` replaces the current configuration files with just the one you specify and then reloads the configuration, which since we have just one configuration file is always what we want when `cloudwatch_logs_conf` changed

Issue #29, opened 2015-01-26
getting crashing issues when running preindex commands


czue:
```
Traceback (most recent call last):
  File "/home/cchq/www/india/python_env/local/lib/python2.7/site-packages/gevent/greenlet.py", line 327, in run
    result = self._run(*self.args, **self.kwargs)
  File "/home/cchq/www/india/code_root_preindex/corehq/preindex/management/commands/sync_prepare_couchdb_multi.py", line 31, in do_sync
    sync_docs.sync(app, verbosity=2, temp='tmp')
  File "./submodules/dimagi-utils-src/dimagi/utils/couch/sync_docs.py", line 77, in sync
    temp=temp,
  File "./submodules/dimagi-utils-src/dimagi/utils/couch/sync_docs.py", line 22, in sync_design_docs
    push(design_dir, db, force=True, docid=docid)
  File "/home/cchq/www/india/python_env/local/lib/python2.7/site-packages/couchdbkit/designer/fs.py", line 393, in push
    doc = document(path, create=False, docid=docid)
  File "/home/cchq/www/india/python_env/local/lib/python2.7/site-packages/couchdbkit/designer/fs.py", line 385, in document
    return FSDoc(path, create=create, docid=docid, is_ddoc=is_ddoc)
  File "/home/cchq/www/india/python_env/local/lib/python2.7/site-packages/couchdbkit/designer/fs.py", line 44, in __init__
    ignorefile = os.path.join(path, '.couchappignore')
  File "/home/cchq/www/india/python_env/lib/python2.7/posixpath.py", line 68, in join
    elif path == '' or path.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

dannyroberts:
What was the command? And was this run directly or through fab?

czue:
it was something like `./manage.py preindex_everything 8 czue` run directly on the machine.

czue:
(it didn't fail persay, just saw these traces a bunch in the output)

PR #4204, opened 2020-09-22
adding new es node: 29

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

Production

I'm trying to add new elastic search node #29 to the production cluster, but the secondary volume is not getting mounted when ran the deploy_stack command and due to which elastic search has been not installed. I'm still debugging on the issue.


prem-fissionhq:
@dannyroberts I missed adding the "es29" node name under the es_data section, because of which I ran into that issue. We identified it and fixed it. Elasticsearch is up and running on es29 and currently, the data syncing is in process.

dannyroberts (approved):

PR #4206, opened 2020-09-22
adding cron job entries to kill celery tasks after 36h

prem-fissionhq:
Celery tasks
- soft kill > 36h
- kill immediately > 46h. 

Increased previous release folders retention period to 48 hours by default.

### DIFF
--- a/src/commcare_cloud/ansible/deploy_commcarehq.yml
+++ b/src/commcare_cloud/ansible/deploy_commcarehq.yml
@@ -139,4 +139,3 @@
   tasks:
     - include_tasks: roles/commcarehq/tasks/celery_cron.yml
   tags: services
-  
--- a/src/commcare_cloud/ansible/deploy_commcarehq.yml
+++ b/src/commcare_cloud/ansible/deploy_commcarehq.yml
@@ -133,3 +133,10 @@
       with_items: "{{ contents.stdout_lines }}"
       when: item not in supervisor_files|sum(start=[])
   tags: services
+
+- name: Celery tasks cleanup Cron job
+  hosts: celery
+  tasks:
+    - include_tasks: roles/commcarehq/tasks/celery_cron.yml
+  tags: services
+  
--- a/src/commcare_cloud/ansible/roles/commcarehq/tasks/celery_cron.yml
+++ b/src/commcare_cloud/ansible/roles/commcarehq/tasks/celery_cron.yml
@@ -1,0 +1,27 @@
+# To remove celery tasks that are running for more than 36 hours.
+#
+- name: Celery cron jobs for soft kill tasks running from last 36h
+  become: true
+  cron:
+    name: "Sending SIGTERM to celery processes running morethan 36h"
+    job: "kill -15 $(ps -eo comm,pid,etimes | awk '/celeryd:/ {if ($4 > 129599) { print $3 }}')"
+    minute: 15
+    hour: "*"
+    weekday: "0,1,2,3,4,5,6"
+    user: root
+    cron_file: celery_sigterm
+  tags:
+    - cron
+
+- name: Cron job to Kill Celery tasks running from last 46h
+  become: true
+  cron:
+    name: "Sending SIGTERM to celery processes running morethan 46h"
+    job: "kill -9 $(ps -eo comm,pid,etimes | awk '/celeryd:/ {if ($4 > 165599) { print $3 }}')"
+    minute: 0
+    hour: "*"
+    weekday: "0,1,2,3,4,5,6"
+    user: root
+    cron_file: celery_sigkill
+  tags:
+    - cron
--- a/src/commcare_cloud/ansible/roles/commcarehq/tasks/celery_cron.yml
+++ b/src/commcare_cloud/ansible/roles/commcarehq/tasks/celery_cron.yml
@@ -4,7 +4,7 @@
   become: true
   cron:
     name: "Sending SIGTERM to celery processes running morethan 36h"
-    job: "kill -15 $(ps -eo comm,pid,etimes | awk '/celeryd:/ {if ($4 > 129599) { print $3 }}')"
+    job: "kill -15 $(ps -eo comm,pid,etimes | awk '/celeryd:/ {if ($4 > 36*60*60) { print $3 }}')"
     minute: 15
     hour: "*"
     weekday: "0,1,2,3,4,5,6"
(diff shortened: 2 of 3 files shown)


prem-fissionhq:
@dannyroberts please review and let me know if you have any suggestions or improvements required on this. 
I tested the cron job entries on staging celery3 server and it worked fine. Please let me know if you see any issues. thanks.

src/commcare_cloud/ansible/deploy_commcarehq.yml
```diff
@@ -133,3 +133,10 @@
       with_items: "{{ contents.stdout_lines }}"
       when: item not in supervisor_files|sum(start=[])
   tags: services
+
+- name: Celery tasks cleanup Cron job
+  hosts: celery
+  tasks:
+    - include_tasks: roles/commcarehq/tasks/celery_cron.yml
+  tags: services
+
```
dannyroberts, line 142: Nitpick: looks like there's trailing whitespace on this line (two spaces?)

dannyroberts (commented):
This looks great. I had a couple small comments but happy to see the test worked and that the solution has a fairly small footprint!

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery_cron.yml
```diff
@@ -0,0 +1,27 @@
+# To remove celery tasks that are running for more than 36 hours.
+#
+- name: Celery cron jobs for soft kill tasks running from last 36h
+  become: true
+  cron:
+    name: "Sending SIGTERM to celery processes running morethan 36h"
+    job: "kill -15 $(ps -eo comm,pid,etimes | awk '/celeryd:/ {if ($4 > 129599) { print $3 }}')"
```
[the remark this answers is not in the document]
prem-fissionhq: @dannyroberts sure will make that change.

src/commcare_cloud/ansible/deploy_commcarehq.yml
prem-fissionhq: @dannyroberts sure will make that change.

src/commcare_cloud/ansible/deploy_commcarehq.yml
```diff
@@ -133,3 +133,9 @@
       with_items: "{{ contents.stdout_lines }}"
       when: item not in supervisor_files|sum(start=[])
   tags: services
+
+- name: Celery tasks cleanup Cron job
+  hosts: celery
+  tasks:
+    - include_tasks: roles/commcarehq/tasks/celery_cron.yml
+  tags: services
```
dannyroberts, line 141: I missed this in the original comment, but if you can try to set your text editor to end files with a newline.

dannyroberts (approved):

src/commcare_cloud/fab/fabfile.py
```diff
@@ -398,7 +398,7 @@ def setup_release(keep_days=0):
     _setup_release(parse_int_or_exit(keep_days), full_cluster=True)
 
 
-def _setup_release(keep_days=0, full_cluster=True):
+def _setup_release(keep_days=2, full_cluster=True):
```
snopoke, line 401: was this change intentional?

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery_cron.yml
```diff
@@ -0,0 +1,27 @@
...
+    user: root
+    cron_file: celery_sigterm
+  tags:
+    - cron
+
+- name: Cron job to Kill Celery tasks running from last 46h
+  become: true
+  cron:
+    name: "Sending SIGTERM to celery processes running morethan 46h"
+    job: "kill -9 $(ps -eo comm,pid,etimes | awk '/celeryd:/ {if ($4 > 46*60*60) { print $3 }}')"
```
snopoke, line 20: can we make these time limits configurable / support disabling these cron jobs?

src/commcare_cloud/fab/fabfile.py
dannyroberts: Yup, this was an important part of the change. The idea was to make sure a celery process cannot outlive the code that underlies it. A bit more motivation in https://dimagi-dev.atlassian.net/browse/SAAS-11324 and the p1 ticket it links to.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery_cron.yml
dannyroberts: @prem-fissionhq I think that's a good suggestion from Simon. Could you move `46` to a variable called `celery_hours_before_hard_kill` and `36` to a variable called `celery_hours_before_soft_kill`? A value of `None` should cause the cron job to be removed (I think something like `state: "{{ 'present' if celery_hours_before_soft_kill else 'absent' }}` should do it.)

@snopoke I do think this is good default behavior, but I also could see ICDS not wanting to guinea pig it along with us, which would be totally reasonable. @prem-fissionhq let's set the default to None and then use these values (36, 46) for SaaS envs by setting them in the public.yml files. (I know that was a lot; feel free to offline with me)

prem-fissionhq: Sure @dannyroberts. I will make changes accordingly.

snopoke: Just following up to make sure this change is coming

snopoke:
I've noticed that in most cases the cron job is failing because no processes are matching the criteria. It would be good to fix that so we don't keep getting the notification emails (it would be better to only notify if there were some processes killed). I think it would make sense to create a small script that the cron job can call which will only run `kill` if there are any processes to kill and will also notify.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery_cron.yml
prem-fissionhq: @snopoke yes, working on it.

prem-fissionhq: Updated the files which suggested changes. Creating new PR with the changes.

prem-fissionhq:
@snopoke handled the failure case with the separate script in the new PR. thanks

prem-fissionhq:
New PR: https://github.com/dimagi/commcare-cloud/pull/4245
Please review @snopoke @dannyroberts

PR #4209, opened 2020-09-22
Shlex quote ssh options

dannyroberts:
Review commit-by-commit. The third commit is the only behavior change. Typical of escaping, we never noticed this because we so far haven't used ssh options that require escaping (due to presence of ` `, `'`, `"`, etc), and so by accident the escaped and unescaped versions are the same. This fixes it to work for ssh option values that contain those characters.

### DIFF
--- a/src/commcare_cloud/commands/ansible/helpers.py
+++ b/src/commcare_cloud/commands/ansible/helpers.py
@@ -13,6 +13,7 @@ from commcare_cloud.colors import color_error, color_success
 from commcare_cloud.environment.paths import ANSIBLE_DIR, ANSIBLE_ROLES_PATH, ANSIBLE_COLLECTIONS_PATHS
 from six.moves import shlex_quote
 from six.moves import range
+from six.moves import map
 
 DEPRECATED_ANSIBLE_ARGS = []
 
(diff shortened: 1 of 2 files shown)


dannyroberts:
Nope, I got the logic here wrong. Will replace with another PR

dannyroberts:
Hey nice @millerdev :)
```
ERROR: Unexpected diff after modernize
diff --git a/src/commcare_cloud/commands/ansible/helpers.py b/src/commcare_cloud/commands/ansible/helpers.py
index 88b6162..6bc1c03 100644
--- a/src/commcare_cloud/commands/ansible/helpers.py
+++ b/src/commcare_cloud/commands/ansible/helpers.py
@@ -13,6 +13,7 @@ from commcare_cloud.colors import color_error, color_success
 from commcare_cloud.environment.paths import ANSIBLE_DIR, ANSIBLE_ROLES_PATH, ANSIBLE_COLLECTIONS_PATHS
 from six.moves import shlex_quote
 from six.moves import range
+from six.moves import map
```
for anyone who hasn't seen it https://github.com/dimagi/commcare-cloud/pull/4203 makes the build start to point out what you have to do to be python3 compatible with any new changes (by failing the build and printing out the diff).

snopoke (approved):

PR #4210, opened 2020-09-22
Get tests passing on Python 3

millerdev:
There is a lot more modernization work that still needs to be done, but this branch is a minimal change set to get tests running on Python 3. Hopefully this will be possible without tackling much of the unicode/bytes work, which is still pending.


dannyroberts:
In case it saves you some time, I can reproduce the latest travis test failure locally with `manage-commcare-cloud make-docs > docs/commcare-cloud/commands/index.md` That's called by `make`; `make` caches, so try `make clean; make` if you want to reproduce it that way and test the rest of that. There is also, separately, `make requirements` which runs during tests to make sure your requirements.txt is consistent with your `setup.py`. (Python2 and python3 might need separate requirements files.)

millerdev:
@dannyroberts Yes, thanks. I'm also able to reproduce locally. Some of the errors I was hitting earlier were more specific to differences between the travis build environment and mine and were harder to reproduce.

millerdev:
@dannyroberts @gherceg tests passed. :white_check_mark: This is ready for review.

src/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -15,7 +15,7 @@
 
 PostgresqlOverride = get_defaults_jsonobject(
     'postgresql_base',
-    allow_dump_from_pgstandby=jsonobject.BooleanProperty,
+    allow_dump_from_pgstandby=jsonobject.BooleanProperty(),
```
dannyroberts, line 18: Interesting. Your change looks right to me

dannyroberts (approved):
This is super awesome!

gherceg (approved):

PR #4213, opened 2020-09-24
update staging es machines

sravfeyn:
I have upgraded ES to 7.9.1 using the es4 and es5 machines. I have already rolled this out. 

We can now delete the es6 and es7 machines. @dannyroberts @sanjay2916

### DIFF
--- a/environments/staging/inventory.ini.j2
+++ b/environments/staging/inventory.ini.j2
@@ -105,12 +105,12 @@ redis2
 
 [elasticsearch:children]
 # Amazon EC2
-# old ES v2 hosts
-# es4
-# es5
-# ES v7 hosts
-es6
-es7
+# ES v7.9.1 hosts
+es4
+es5
+# ES v7.6.2 hosts
+# es6
+# es7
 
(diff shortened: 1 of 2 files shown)


environments/staging/inventory.ini.j2
```diff
@@ -105,12 +105,12 @@ redis2
...
 # Amazon EC2
-# old ES v2 hosts
-# es4
-# es5
-# ES v7 hosts
-es6
-es7
+# ES v7.9.1 hosts
+es4
+es5
```
millerdev, line 110: Did you reuse the old ES2 VMs, or are `es4` and `es5` new VMs? It's probably too late, but it's not a good idea to reuse host names because that can make it difficult to track the history any place where those names are referenced. Better to retire old names and make new ones for new things.

sravfeyn: Hey, I reused the same machines, haven't created new ones, so hostnames shouldn't be an issue

millerdev (approved):

PR #4216, opened 2020-09-24
Fix classic division on Python 2

gherceg:
Ran `python-modernize -wnf libmodernize.fixes.fix_classic_division` and ensured that places that relied on float division (`/`) were not updated to floor division (`//`).


gherceg:
Jumped the gun

PR #4220, opened 2020-09-28
Replace classic division with floor division where applicable

gherceg:
Step 3 of [this document](https://docs.google.com/document/d/17QQu5BAAwle5KHTMawTlsf4o78ZRXElIKlrBdHjrU8M/edit#).

### DIFF
--- a/src/commcare_cloud/ansible/roles/pg_repack/files/pg_repack.py
+++ b/src/commcare_cloud/ansible/roles/pg_repack/files/pg_repack.py
@@ -3,6 +3,7 @@
 # Script run by cron to execute pg_repack under certain conditions
 # Must be run as the ``postgres`` user.
 #
+from __future__ import division
 import argparse
 import multiprocessing
 import subprocess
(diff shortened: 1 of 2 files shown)


scripts/es_snapshot.py
```diff
@@ -348,8 +349,8 @@ def run(self):
                 suffix = f" ({total - success} of {total} shards failed)"
             print(f"    {snapshot} [{state}]{suffix}")
             if self.detail:
-                start = datetime.fromtimestamp(snap_info['start_time']/1000)
-                end = datetime.fromtimestamp(snap_info['end_time']/1000)
+                start = datetime.fromtimestamp(snap_info['start_time']//1000)
+                end = datetime.fromtimestamp(snap_info['end_time']//1000)
```
millerdev, line 353: Do we know for sure that `snap_info['start_time']` and `snap_info['end_time']` are always ints (not floats)?

millerdev (approved):
Approved unless the answer to my question is negative.

scripts/es_snapshot.py
gherceg: I think you are right and it is better to use float division here. @snopoke can you confirm?

snopoke: I don't know, I'd have to check.

This script was developed for PY3 only so it shouldn't need any change.

gherceg: Good point. I'll revert changes made to this file.

millerdev (approved):

PR #4222, opened 2020-09-28
Skip test_modernize.sh on Python 3

millerdev:
Test is inexplicably failing on Python 3 in https://github.com/dimagi/commcare-cloud/pull/4220

### DIFF
--- a/tests/test_modernize.sh
+++ b/tests/test_modernize.sh
@@ -6,6 +6,13 @@
 #
 # undo-modernizations.diff may need to be updated if new lines are added
 # that would be, but should not be modernized.
+
+PYVER=$(python --version 2>&1)
+[[ $PYVER =~ Python\ 2\. ]] || {
+    echo "Skipping modernize test on Python $PYVER"
+    exit 0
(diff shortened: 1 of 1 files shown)


tests/test_modernize.sh
```diff
@@ -6,6 +6,13 @@
 #
 # undo-modernizations.diff may need to be updated if new lines are added
 # that would be, but should not be modernized.
+
+PYVER=$(python --version 2>&1)
+[[ $PYVER =~ Python\ 2\. ]] || {
+    echo "Skipping modernize test on Python $PYVER"
```
gherceg, line 12: `$PYVER` already has "Python" in it, so does the echo string add a redundant Python?

gherceg (approved):

tests/test_modernize.sh
millerdev: Thanks. Fixed in https://github.com/dimagi/commcare-cloud/pull/4222/commits/b69718700d00aa9d301a34530306b7761889326d

PR #4229, opened 2020-10-01
add health check interval to redis config

biyeun:
https://dimagi-dev.atlassian.net/browse/SAAS-11355 (and possibly the P2 that came up)

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -424,6 +424,9 @@ redis_cache = {
 {% endif %}
     'OPTIONS': {
         'PARSER_CLASS': 'redis.connection.HiredisParser',
+        'REDIS_CLIENT_KWARGS': {
+            'health_check_interval': 30,
+        },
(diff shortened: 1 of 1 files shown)


biyeun:
ok, looks like staging was fine with the change 👍

dannyroberts (approved):
Sounds you've put it up on staging and run check_services to make sure there are no errors? If so and clicking around on staging works then I think I'm good with the change

millerdev (approved):
I don't remember the issue we faced last time we rolled this out, but this seems fine as long as that scenario has been tested and is no longer a concern.

dannyroberts:
We luckily _didn't_ roll it out last time, but when I was testing before rolling it out I found that when I ran `cchq staging update-config --limit django_manage` the `check_services` it runs after that failed, so I rolled back the config and raised the issue, which prompted the redis upgrade to 3.3.0

biyeun:
@dannyroberts `check_services` runs at the end of `update-config` yeah? if so I didn't see any errors then. and also staging looks stable now that cal undid whatever he was working on

dannyroberts:
Yup it does. This sounds good to roll out to me

PR #4236, opened 2020-10-02
Move kinesis-agent checkpoint file to a location whose permissions we control

dannyroberts:
##### ENVIRONMENTS AFFECTED
production, india, staging


knguyen142 (approved):
We should probably bring that file out to a var?

knguyen142:
Should we create a ticket to alert on this in the future? I'm thinking the easiest way is to set up an alert on the kinesis stream.

dannyroberts:
Yes definitely, I'll offline with you about that

PR #4242, opened 2020-10-05
remove removed setting

mkangia:
https://github.com/dimagi/commcare-hq/commit/0e2683ef5015b5fef7b9c8879f265c0a325d08b2#diff-bdf3ecebd8379ca98cc89e545fc90899

##### ENVIRONMENTS AFFECTED
None


snopoke (approved):

mkangia:
I have not run this on any (non icds) environment because i believe the setting is actually not available on any. So merging it.

PR #4246, opened 2020-10-06
tracing staging

calellowitz:
This adds django and java tracing from datadog to staging, as well as turns on some postgres settings that should allow greater insight into performance.

##### ENVIRONMENTS AFFECTED

Staging

### DIFF
--- a/environments/staging/app-processes.yml
+++ b/environments/staging/app-processes.yml
@@ -1,3 +1,4 @@
+formplayer_command_args: '-javaagent:/home/cchq/dd-java-agent.jar '
 management_commands:
   celery3:
     run_sms_queue:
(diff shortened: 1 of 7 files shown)


millerdev (approved):

environments/staging/app-processes.yml
```diff
@@ -1,3 +1,4 @@
+formplayer_command_args: '-javaagent:/home/cchq/dd-java-agent.jar '
```
dannyroberts, line 1: nit: the space at the end is unintuitive. It seems like it's not actually needed either, looking at where `{{ app_proceses_config.formplayer_command_args }}` is used below

PR #4248, opened 2020-10-07
Add couch db compaction to firefighting docs

dannyroberts:
This is something I find myself doing regularly which according to these docs also means I should add more disk ;)

### DIFF
--- a/docs/firefighting/index.md
+++ b/docs/firefighting/index.md
@@ -139,6 +139,25 @@ Monitors are setup to ping the proxy instead of couch instance directly, so the 
     b. It could also have ballooned (ICDS) which is out of our control
 5. If it's unresponsive and it's out of our control to fix it at the time, go to the proxy machine and comment out the fault node from the nginx config. This will stop sending requests to that server, but it will continue to replicate. When the slowness is over you can uncomment this line and begin proxying reads to it again
 
+## Couch node data disk is high
+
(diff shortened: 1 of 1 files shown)


docs/firefighting/index.md
```diff
@@ -139,6 +139,25 @@ Monitors are setup to ping the proxy instead of couch instance directly, so the
...
+
+```
+cchq ${env} ssh django_manage
+```
+
+and then post to the _compact endpoints of the larger dbs, e.g.:
+
+```
+curl -X POST http://${couch_proxy}:25984/commcarehq__auditcare/_compact -v -u ${couch_username} -H 'Content-Type: application/json' -d'{}'
+curl -X POST http://${couch_proxy}:25984/commcarehq__receiverwrapper/_compact -v -u ${couch_username} -H 'Content-Type: application/json' -d'{}'
```
gherceg, line 157: Not sure if it's worth adding here, but someone like me might wonder how to know what the ${couch_proxy} and ${couch_username} might be/where to find them.

gherceg (approved):
What's there looks good. Not sure if this is the place to address my comment, or if there is a better place to put information like that.

docs/firefighting/index.md
dannyroberts: Sounds good, I added a couple lines below

PR #4249, opened 2020-10-08
Increase production postgres connection limit

dannyroberts:
- raise postgres connection limit 300 => 500
- increase size of pgmain0

I was tempted to also change these settings to match [pgtune](https://pgtune.leopard.in.ua/#/) recommendations, but I chose not to hold off since changing any of these could conceivably have a _negative_ effect on performance. (Before we do, we should at least understand what the current/default values are in RDS/postgres.)
```
      shared_buffers: 16GB
      effective_cache_size: 48GB
      maintenance_work_mem: 2GB
      default_statistics_target: 100
      random_page_cost: 1.1
      effective_io_concurrency: 200
      work_mem: 20971kB
      min_wal_size: 1GB
      max_wal_size: 4GB
      max_worker_processes: 16
      max_parallel_workers_per_gather: 4
```

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -1,13 +1,13 @@
 postgres_override:
   allow_dump_from_pgstandby: yes
-  postgresql_max_connections: 300
+  postgresql_max_connections: 500
   postgresql_version: '9.6'
   postgresql_checkpoint_completion_target: '0.7'
   postgresql_wal_buffers: 16MB
   postgresql_max_wal_size: 2GB
 
 pgbouncer_override:
-  pgbouncer_default_pool: 290
+  pgbouncer_default_pool: 490
   pgbouncer_max_connections: 10000
   pgbouncer_pool_mode: transaction
   pgbouncer_pool_timeout: 1
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -490,7 +490,7 @@ proxy_servers:
 
 rds_instances:
   - identifier: "pgmain0-production"
-    instance_type: "db.t3.2xlarge"
+    instance_type: "db.m5.4xlarge"
     storage: 500
     max_storage: 5000
     multi_az: true


environments/production/postgresql.yml
```diff
@@ -1,13 +1,13 @@
...
-  postgresql_max_connections: 300
+  postgresql_max_connections: 500
   postgresql_version: '9.6'
   postgresql_checkpoint_completion_target: '0.7'
   postgresql_wal_buffers: 16MB
   postgresql_max_wal_size: 2GB
 
 pgbouncer_override:
-  pgbouncer_default_pool: 290
+  pgbouncer_default_pool: 490
```
dannyroberts, line 10: Changing it here annoyingly changes it for _all_ production postgres dbs. I know how to override `postgresql_max_connections` for just a single instance, but I don't think we have a way to override the pgbouncer pool size for just a set of dbs?

calellowitz: I think this is configurable in the inventory file on a per machine basis (per inventory group really). That would also prevent us from raising this beyond the postgres limit for any DBs

dannyroberts: Cal and I offlined a bit, but basically our code is set up to do this well if each postgresql machine also hosts pgbouncer—you just set `pgbouncer_default_pool` and `postgresql_max_connections` as inventory variables. The way we host in AWS envs, with a single pgbouncer machine pointing to multiple rds dbs, means that we would have to specify different values for different dbs in the same pgbouncer.ini config file. Since this isn't something we've tested, and we aren't even sure what the edge cases are there (you set `pool_size` per db, but if you have multiple dbs on a single host, do the pools sizes effectively get added together or is the largest one for a given host chosen?), so we opted to just raise the defaults and apply to all instances. No other db instances ever reach 300 connections currently, so raising that limit is in the short term largely theoretical anyway.

dannyroberts:
Weird. I'm seeing that instead of 500, `max_connections` is set to `LEAST({DBInstanceClassMemory/9531392},5000)` (the default value, a calculation based off RAM). That's fine for us since all our instances that's still more than 500. (I think for pgmain0 it's nearly 7,000.)

dannyroberts:
To roll out I ran

```
cchq production terraform apply
```
then manually modified the instance type in the RDS console to match and chose the option to apply immediately.
Concurrently for all the other dbs whose size wasn't changing but params were, I just rebooted them from the RDS console with the Failover option.

PR #4250, opened 2020-10-08
Have terraform ignore manual changes to the IP deny list

dannyroberts:
This PR is basically acknowledging that an IP deny list is a more dynamic thing than we want to hard code in a setting in a git repo

##### ENVIRONMENTS AFFECTED
saas aws envs

### DIFF
--- a/src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf
+++ b/src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf
@@ -52,6 +52,9 @@ resource "aws_wafv2_ip_set" "permanent_block" {
   ip_address_version = "IPV4"
   description        = "Manually Added IPs for denial"
   addresses          = ["195.54.160.21/32", "64.39.99.208/32"]
+  lifecycle {
+    ignore_changes = ["addresses"]
+  }
 
   tags = {
   }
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf
```diff
@@ -52,6 +52,9 @@ resource "aws_wafv2_ip_set" "permanent_block" {
   ip_address_version = "IPV4"
   description        = "Manually Added IPs for denial"
   addresses          = ["195.54.160.21/32", "64.39.99.208/32"]
+  lifecycle {
```
gherceg, line 55: Just so I understand, we still hard code/reference addresses listed here, but will ignore changes made to addresses?

gherceg (approved):

src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf
dannyroberts: Yeah sorry this would be less dissonant if I also replaced the addresses value above with an empty list. I'll do that

dannyroberts:
@ctsims FYI this PR should better align with how in practice we're doing this

ctsims:
@dannyroberts Thanks for the heads up. Is the implication that I should continue to think it's fine to update these manually in the AWS console / CLI?

dannyroberts:
@ctsims yup exactly

src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf
```diff
@@ -51,7 +51,10 @@ resource "aws_wafv2_ip_set" "permanent_block" {
   scope              = "REGIONAL"
   ip_address_version = "IPV4"
   description        = "Manually Added IPs for denial"
-  addresses          = ["195.54.160.21/32", "64.39.99.208/32"]
+  addresses          = []
```
millerdev, line 54: Should/could we add a comment here to note that these are managed in the AWS console?

PR #4256, opened 2020-10-08
prep for citus upgrade

snopoke:
* rename some vars to make it clearer what they are for
* enforce only 1 citusdb cluster (1 master node)
* specify the version of the citusdb extension to install. This will have the effect of upgrading the extension when a new version is specified


millerdev (commented):
test failures are related

snopoke:
Build passing now

shyamkumarlchauhan (approved):

PR #4258, opened 2020-10-08
lower health check interval to 15 sec

biyeun:
https://dimagi-dev.atlassian.net/browse/SAAS-11352

##### ENVIRONMENTS AFFECTED
all environments


millerdev (approved):
I glanced at the Jira ticket, but wasn't able to determine the theory behind this change. (why are we changing it?) An explanation might be useful for future research projects.

biyeun:
@millerdev Danny and I suspected that redis might be killing connections before 30 seconds is up. This is an attempt to see if this change makes any difference in that front.

PR #4261, opened 2020-10-10
adding task to remove particular user pubkey from ansible user

prem-fissionhq:
Adding task to remove particular user pubkey from _ansible_ user. 

command: 
cchq <env> django-manage update-users --tags=remove_pubkey -e '{"user":["username"]}' 

This is one of the approaches we can use to remove a particular a user from the ansible authorized_keys file. Please let me know your suggestions.

### DIFF
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -64,21 +64,22 @@
   tags:
     - new_ansible_sudo_pass
 
-- name: Authorize current devs to login as ansible
-  become: yes
-  authorized_key: user=ansible state=present key="{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}"
+- name: remove key using username
+  lineinfile:
+    dest: /home/ansible/.ssh/authorized_keys
+    state: absent
+    regexp: "username={{ item }}"
+  when: reset_user | default(False)
   with_items: '{{ dev_users.present }}'
   tags:
     - pubkey
 
-# task is remove old pubkey of particular devs from ansible and update with new pubkey
-- name: UnAuthorize current devs to login as ansible
+- name: Authorize current devs to login as ansible
   become: yes
-  authorized_key: user=ansible state=absent key="{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}"
-  with_items: '{{ user }}'
+  authorized_key: user=ansible state=present key="{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}" key_options=environment="username={{ item }}"
+  with_items: '{{ dev_users.present }}'
   tags:
-    - remove_pubkey
-  when: user is defined
+    - pubkey
 
 - import_tasks: remove_users.yml
   when: dev_users.absent
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -77,8 +77,7 @@
 
 - name: capturing username
   pause:
-    prompt: "Please enter your dimagi username"
-    seconds: 30
+    prompt: "Please enter your Dimagi username"
   register: username_input
 
 - fail: msg="The given username is not defined or empty or not a valid dev_user"
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -72,8 +72,8 @@
   register: ssh_pubkeys_results_list
 
 - name: iterate over ssh_pubkeys_list and join into a string
-    set_fact:
-      ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
+  set_fact:
+    ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
 
 - name: Authorize current devs to login as ansible
   become: yes
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -64,19 +64,20 @@
   tags:
     - new_ansible_sudo_pass
 
-- name: remove key using username
-  lineinfile:
-    dest: /home/ansible/.ssh/authorized_keys
-    state: absent
-    regexp: "username={{ item }}"
-  when: reset_user | default(False)
-  with_items: '{{ dev_users.present }}'
-  tags:
-    - pubkey
+- name: lookup ssh pubkeys from authorized_keys_dir and create ssh_pubkeys_list
+  set_fact:
+    ssh_pubkeys_list: "{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}"
+  with_items:
+    "{{ dev_users.present }}"
+  register: ssh_pubkeys_results_list
+
+- name: iterate over ssh_pubkeys_list and join into a string
+    set_fact:
+      ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
 
 - name: Authorize current devs to login as ansible
   become: yes
-  authorized_key: user=ansible state=present key="{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}" key_options=environment="username={{ item }}"
+  authorized_key: user=ansible state=present key="{{ ssh_pubkeys_string }}" exclusive=yes
   with_items: '{{ dev_users.present }}'
   tags:
     - pubkey
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -70,26 +70,42 @@
   with_items:
     "{{ dev_users.present }}"
   register: ssh_pubkeys_results_list
+  tags:
+    - pubkey
 
 - name: iterate over ssh_pubkeys_list and join into a string
   set_fact:
     ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
+  tags:
+    - pubkey
 
 - name: capturing username
   pause:
     prompt: "Please enter your Dimagi username"
   register: username_input
+  tags:
+    - capture-userinfo
+    - pubkey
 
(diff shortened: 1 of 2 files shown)


millerdev (commented):
I'm assuming this is part of a proposed solution to [SAAS-11294](https://dimagi-dev.atlassian.net/browse/SAAS-11294). Is this new command intended to be run before a user's key is changed?

If not, can you explain how this will be used in the process of changing a user's public key.

If yes, it seems easy to forget this step in the process of changing a user's key. Can we make it happen automatically when either `update-users` or `update-user-key` is run.

prem-fissionhq:
@millerdev yes, correct the ticket https://dimagi-dev.atlassian.net/browse/SAAS-11294 is where I'm updating my comments as well.
`Is this new command intended to be run before a user's key is changed?` 
it is an existing command with new tags `--tags=remove_pubkey `, to remove the existing key before we ran with a new key. 

`If yes, it seems easy to forget this step in the process of changing a user's key. Can we make it happen automatically when either update-users or update-user-key is run?`  

No. I think its better to handle this case separately. The current state of playbooks...
The `update-users (tags=users)` and `update-user-key (tags=pubkey)` command using `authorized_keys` module which will get all users info from the "dev_users.present" and verifies the content of each *.pub with the `/home/<username>/.ssh/authorized_keys` file, if _string not_ matched then it will add else it will ignore. 
Problem is 
1. if we control variable `state: {present|absent}` with the args, in case of `absent` it will delete all the users under `dev_users.present`. 
2. the update-user-key replaces new key **only inside** the respective user authorized_keys file. Will not replace for user `ansible`, only adds new key into the file by ignoring old key info ( you already aware of this issue).
3. The option we have now is to move the user to in `dev_users.absent` section. But this will delete the entire user and his key from all places (not feasible). 
So, as per my knowledge the easy way is to before updating 
    a) First remove the old key from <env> using a `some command` before adding the new key under authorized_keys directory. 
    b) add new key using update-user-key <username> command.  

Please let me know if you have any questions and correct me if im msising anything. Thanks

millerdev:
I added comments and an alternative implementation outline in [the ticket](https://dimagi-dev.atlassian.net/browse/SAAS-11294).

prem-fissionhq:
@millerdev sure, I will check and get back to you if I have any questions. Thanks.

prem-fissionhq:
@millerdev @dannyroberts 
I made a small update to handle the retried/reset user keys case. could you please review this? let me know your thoughts. 

Using the same `update-user-key` method we can do the following case.  
1. Appending an environment string with username={dev_username} to the SSH_key string stored in authorized_keys using authorized_keys.key_options variable. ( also helps us in identifying the dev users keys )
```
environment="username=devops" ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC1FbDcqccI8QfYqIvenbsusbWAG9QrrSblQeq/L7HGX7ziH7EehgjixQ9OcgmKvUrp8d ---------------------------- Eo/64IkpQ== devops@abv.com

environment="username=prem" ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQCmxq9jb07921uWPJVMfoV1qP4+femBqHxyvQX8McuTziYFlWi+RrIKNVSKPZChSTU----------------------------  DR/uUMln9== prem4662@gmail.com
```
2. Added new task to remove user key from ansible's authorized_keys file using regex.
`regrex: "username=prem"`
3. new parameter `reset_user=True` to remove old key (default: False)
`commcare-cloud <env> update-user-key <username> -e "reset_user=True"`

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
```diff
@@ -64,21 +64,22 @@
...
 
-- name: Authorize current devs to login as ansible
-  become: yes
-  authorized_key: user=ansible state=present key="{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}"
+- name: remove key using username
+  lineinfile:
+    dest: /home/ansible/.ssh/authorized_keys
+    state: absent
+    regexp: "username={{ item }}"
+  when: reset_user | default(False)
```
millerdev, line 72: If this option is not specified (e.g., because it is forgotten or the person running the command does not know to use it), then the retired key will not be removed. This is not ideal since we are using `update-user-key` more frequently than `update-users` these days because of performance, and therefore we are more likely to accidentally retain retired keys for a long time. Would it break anything if we change it to `default(True)`? I assume it will cause more churn and impact performance?

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
```diff
@@ -64,21 +64,22 @@
...
   tags:
     - pubkey
 
-# task is remove old pubkey of particular devs from ansible and update with new pubkey
-- name: UnAuthorize current devs to login as ansible
+- name: Authorize current devs to login as ansible
   become: yes
-  authorized_key: user=ansible state=absent key="{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}"
-  with_items: '{{ user }}'
+  authorized_key: user=ansible state=present key="{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}" key_options=environment="username={{ item }}"
```
millerdev, line 79: Will existing keys be updated automatically the first time this is run? If not, what is the plan to update existing authorized keys?

prem-fissionhq:
@dannyroberts the reason for not using the `exclusive=yes` parameter for the task `name: Authorize current devs to login as ansible` is we are running this task in a **loop** with the all items under` {{ dev_users.present }}`. And since it is in loop everytime the `exclusive=yes` is clearing the data inside the file and at the end we are seeing only one user's ssh_key in `/home/ansible/.ssh/authorized_keys` file. 

example: tested with 3 users.
Before: `exclusive=no` (default: no)

image

After: `exclusive=yes`

image


Thanks.

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
prem-fissionhq: Yes Daniel, using the default(True) will remove all user's the keys present in `{{ dev_users.present }}` first from the authorized_key file which are matching the with regex. 
I totally agree with you that the user might either forget or not might not be aware of using that option.

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
prem-fissionhq: we need to run the `update-users` in order to make this change available on all machines. Still thinking about the alternative procedure to make this option available. Please let me know your suggestions too. thanks

prem-fissionhq:
@shyamkumarlchauhan as discussed, I'm capturing the username and verifying their key in the new list of authorized_keys, if not present then we will not proceed further and make changes to existing/previous ansible user authorized_keys file. please review. thanks

prem-fissionhq:
I have made necessary changes to the deploy-stack command to skip prompt-user tasks during the first-time setup. Please review and let me know if I'm missing anything. Thanks.

dannyroberts (approved):
This now looks good to me. Can you talk a bit more about how you have tested this / what else you plan to do to test this? (Just want to be very certain this won't cause issues!)

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
```diff
@@ -64,9 +64,52 @@
...
+- name: iterate over ssh_pubkeys_list and join into a string
+  set_fact:
+    ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
+  tags:
+    - pubkey
+
+- name: capturing username
+  pause:
+    prompt: "Please enter your Dimagi username"
+  register: username_input
```
millerdev, line 85: Who is "your" referring to in this prompt? I'm confused about how this helps with removing retired keys from the authorized keys file since (a) there may be multiple retired keys and (b) the person running ansible will rarely, if ever, be the person who's key is being retired.

Will this prompt be presented every time any command that includes this role (other than `deploy_stack`) is run (ex: during `cchq ... update-users` or `cchq ... update-user-key`)? Will it be necessary to enter a username even when no keys have been retired?

millerdev (changes requested):
It's been a while since I looked at this, so my memory is a little foggy. However, I'm skeptical that this meets the objectives set out in [SAAS-11294 comment](https://dimagi-dev.atlassian.net/browse/SAAS-11294?focusedCommentId=92589). Is this intended to be run on all environments each time a user's key is retired?

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
prem-fissionhq: @millerdev "your" refers to the person who is running the `update-users` or `update-user-key` command. We would like to ensure that the "entered" user's key is not removed from the list of ssh-keys while updating the authorized_keys file. That user should be valid dev_users.present. 
yes, It will prompt for both update-users and update-user-key commands. 

@dannyroberts Tested in the following ways
1. deploy-stack --first-time
2. deploy-stack --skip-check
3. update-users - with valid dev_user name
4. update-users - without valid dev_user name
5. update-user-key ( found a bug today while testing again for screenshots, explanation below).
    - with/without valid dev_user name. 
    
Bug: Problem with `exclusive: yes` for `[bootstrap-users: Authorize current devs to login as ansible]` task. 
since update-users and update-user-key both runs `tags: pubkey` tasks. the `exclusive: yes` will overwrite the ansible authorized_keys file and adds only one key into it when the `update-user-key <username>` command is executed.
Sorry for missing this. I was more concentrating on the user_prompt testing. 
@millerdev thank you. Now, I understood your comment (ticket) on why you mentioned that we may have to rewrite or modify the logic for both commands. 

@dannyroberts I have a go through the tasks again. So, I will get back to you on this. 
Thanks.

dannyroberts: @millerdev yeah to put it another way, one of the risks we identified is that someone might somehow inadvertently remove their own key. To mitigate that risk, we use this task to ask the user running the command to input their own ssh username and then check to make sure that that username shows up in the exclusive list of keys we will be overwriting the authorized_keys file with. That makes it really hard for someone to inadvertently remove their own key.

We believe everything would work fine without this check as long as no one made any mistakes, but having this check helps prevent such a mistake.

dannyroberts:
I talked this over with Prem. The huge disconnect here we realized is that the PR title and PR description are totally out of date. The changes here I think are correct but they just don't do what's described in the PR text. He's going to close this PR and reopen it with a cleaned up commit history and a more accurate description. What happened was he opened this PR back in October and then through a number of conversations the approach we decided to take to solve the underlying problem changed substantially. Rather than targeted commands to add, replace, remove keys for specific users, we are now going back to a declarative approach that relies on `update-users` only but which (1) is much faster and (2) fully rewrites the authorized keys to match the full correct list (rather than separate commands to add and remove each key). Anyhow, apologies for the confusion, and we'll get this cleaned up in a new PR

prem-fissionhq:
Thanks, Danny for the quick call. really helped a lot. 
Closing this PR to create a new PR with a new description.

PR #4262, opened 2020-10-12
get correct version from DB to fix postgresql_ext bug

snopoke:
There are two issues that make this necessary.

1. The version of the citus extension does not match the version of the package (the package does not include the 'patch' version). This works OK for installing the extension the first time but after that the `postgresql_ext` ansible module will complain that `9.4 < 9.4.1`. 

2. The `postgresql_ext` ansible module does not correctly sort available versions. To work around the 1st point you can set the version to `latest` but a bug in the Ansible module where it's not sorting the versions means that the version it tries to install is somewhat random: https://github.com/ansible-collections/community.general/pull/1078

This change queries the DB for the latest available version that is >= the specified version and uses that as the input to the `postgresql_ext` module.

### DIFF
--- a/src/commcare_cloud/ansible/roles/citusdb/tasks/setup_cordinator.yml
+++ b/src/commcare_cloud/ansible/roles/citusdb/tasks/setup_cordinator.yml
@@ -1,19 +1,26 @@
 ---
- - name: Add Citus Extension on Coordinator Database
-   become: yes
-   become_user: postgres
-   postgresql_ext:
-     name: citus
-     db: "{{ item.name }}"
-     version: "{{ citus_version }}"
-   with_items: "{{ postgresql_dbs.all }}"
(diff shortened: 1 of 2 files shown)


millerdev (approved):

millerdev:
> This change queries the DB for the latest available version that is >= the specified version

Does this mean that if a version `9.10.0` were published, that this would fail because `'9.4' >= '9.10.0'`?

snopoke:
> Does this mean that if a version 9.10.0 were published, that this would fail because '9.4' >= '9.10.0'?

No, PostgreSQL only knows about versions that are installed on the system. This is exactly what should happen during an upgrade where you change the version in commcare-cloud and run the playbook which will:

1. Install the new version of the package (which uninstalls the previous version)
2. Restart PostgreSQL
3. Update the extension in the DB

PR #4267, opened 2020-10-13
Use pip-sync to install requirements on deploy

millerdev:
`pip-sync` should produce a more consistent virtualenv state. We recently had an issue (https://github.com/dimagi/commcare-hq/pull/28563) where a no longer required package was still present in some virtualenvs, causing deploy failures. `pip-sync` should prevent that from ever happening because it uninstalls packages that are present in the virtualenv but not required.

See also https://github.com/dimagi/commcare-hq/pull/28641

Test deployments with `cchq <env> --control deploy --branch=dm/pip-sync`:
- [x] staging
- [x] india


dannyroberts (approved):

millerdev:
This is still not safe to merge. It will most likely require a new change log entry to be added with steps to run `update-supervisor-confs` to remove errand-boy process configs.

snopoke (commented):
For ICDS this have the effect of uninstalling and then reinstalling any requirements in the ICDS requirements file (currently none). 

Just thinking that it would make more sense to pass all the necessary requirements files to pip-sync in the first place.

millerdev:
Rebased/moved errand-boy removal to it's own PR: https://github.com/dimagi/commcare-cloud/pull/4268

@snopoke your suggestion is implemented in https://github.com/dimagi/commcare-cloud/pull/4267/commits/d23dcb37fee386dcef00ebe5a6170a393b0705f6

snopoke (approved):

PR #4270, opened 2020-10-14
Work around pip2 SSL error

millerdev:
The error started [causing tests to fail](https://travis-ci.org/github/dimagi/commcare-cloud/builds/735737545) today. Not sure why.

##### ENVIRONMENTS AFFECTED
maybe only travis tests or maybe all?


snopoke (approved):
Fix seems OK but I'm surprised we are using python2 for supervisord. That is something that we should change.

PR #4276, opened 2020-10-16
Make control/init.sh less chatty

millerdev:
##### ENVIRONMENTS AFFECTED
all, but only affects developers running deploy

### DIFF
--- a/src/commcare_cloud/manage_commcare_cloud/install.py
+++ b/src/commcare_cloud/manage_commcare_cloud/install.py
@@ -1,5 +1,4 @@
-from __future__ import absolute_import
-from __future__ import unicode_literals
+from __future__ import absolute_import, print_function, unicode_literals
 import os
 import subprocess
 from six.moves import shlex_quote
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/manage_commcare_cloud/install.py
```diff
@@ -25,20 +24,24 @@ def run(self, args, unknown_args):
...
             print_command(cmd)
-            p = subprocess.Popen(cmd, stdin=subprocess.PIPE, shell=True, env=env)
-            p.communicate()
-
-        puts(color_notice("To finish first-time installation, run `manage-commcare-cloud configure`".format()))
-        return p.returncode
+            try:
+                subprocess.check_output(cmd, shell=True, env=env)
+            except subprocess.CalledProcessError as err:
+                print("process exited with error: %s" % err.returncode)
```
millerdev, line 43: Warnings and errors are already printed to stderr, so they will be visible. Could also `print(err.output)`, although that will likely mean that the error is buried way above in the output. We can change it if this turns out to be a bad decision.

dannyroberts (approved):

snopoke:
This is so much better! :heart:

PR #4279, opened 2020-10-19
Add Django SecurityMiddleware to staging

gherceg:
[Resolve security warnings](https://dimagi-dev.atlassian.net/browse/SAAS-11207)

This is only a change to staging in order to play around with some values related to strict transport security. 
##### ENVIRONMENTS AFFECTED

Staging

### DIFF
--- a/environments/staging/public.yml
+++ b/environments/staging/public.yml
@@ -134,7 +134,7 @@ localsettings:
     - 'django.middleware.security.SecurityMiddleware'
   SECURE_HSTS_SECONDS: 300
   SILENCED_SYSTEM_CHECKS:
-    - 'security.W008'
+    - 'security.W008' # the ALB and nginx already redirect HTTP to HTTPS so safe to silence this warning
   PILLOWTOP_MACHINE_ID: staging-hqdb0-ubuntu
   ALLOW_PHONE_AS_DEFAULT_TWO_FACTOR_DEVICE: True
   RATE_LIMIT_SUBMISSIONS: yes
--- a/environments/staging/public.yml
+++ b/environments/staging/public.yml
@@ -130,6 +130,11 @@ localsettings:
   J2ME_ADDRESS: "{{ J2ME_SITE_HOST }}"
   LOCAL_APPS:
     - {name: 'ddtrace.contrib.django', host: webworkers}
+  LOCAL_MIDDLEWARE:
+    - 'django.middleware.security.SecurityMiddleware'
+  SECURE_HSTS_SECONDS: 300
+  SILENCED_SYSTEM_CHECKS:
+    - 'security.W008'
   PILLOWTOP_MACHINE_ID: staging-hqdb0-ubuntu
   ALLOW_PHONE_AS_DEFAULT_TWO_FACTOR_DEVICE: True
   RATE_LIMIT_SUBMISSIONS: yes
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -832,13 +832,13 @@ GREENHOUSE_API_KEY = '{{ GREENHOUSE_API_KEY }}'
 LOCAL_LOGGING_CONFIG = json.loads('{{ localsettings.get('LOCAL_LOGGING_CONFIG', {})|to_json }}')
 LOCAL_MIDDLEWARE = {{ localsettings.get('LOCAL_MIDDLEWARE', [])|to_json }}
 
+{% if 'SECURE_HSTS_SECONDS' in localsettings %}
 SECURE_CONTENT_TYPE_NOSNIFF = {{ localsettings.get('SECURE_CONTENT_TYPE_NOSNIFF', True) }}
 SECURE_BROWSER_XSS_FILTER = {{ localsettings.get('SECURE_BROWSER_XSS_FILTER', True) }}
 SECURE_HSTS_SECONDS = {{ localsettings.get('SECURE_HSTS_SECONDS', 0) }}
 SECURE_HSTS_INCLUDE_SUBDOMAINS = {{ localsettings.get('SECURE_HSTS_INCLUDE_SUBDOMAINS', True) }}
-
 SILENCED_SYSTEM_CHECKS = {{ localsettings.get('SILENCED_SYSTEM_CHECKS', [])|to_json }}
-
+{% end if %}
 
 BYPASS_SESSIONS_FOR_MOBILE = {{localsettings.get('BYPASS_SESSIONS_FOR_MOBILE', False)}}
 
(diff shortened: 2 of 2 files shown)


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -832,6 +832,14 @@ GREENHOUSE_API_KEY = '{{ GREENHOUSE_API_KEY }}'
 LOCAL_LOGGING_CONFIG = json.loads('{{ localsettings.get('LOCAL_LOGGING_CONFIG', {})|to_json }}')
 LOCAL_MIDDLEWARE = {{ localsettings.get('LOCAL_MIDDLEWARE', [])|to_json }}
 
+{% if 'SECURE_HSTS_SECONDS' in localsettings %}
+SECURE_CONTENT_TYPE_NOSNIFF = {{ localsettings.get('SECURE_CONTENT_TYPE_NOSNIFF', True) }}
+SECURE_BROWSER_XSS_FILTER = {{ localsettings.get('SECURE_BROWSER_XSS_FILTER', True) }}
+SECURE_HSTS_SECONDS = {{ localsettings.get('SECURE_HSTS_SECONDS', 0) }}
```
snopoke, line 838: This should be set in nginx directly: https://github.com/dimagi/commcare-cloud/blob/master/environments/production/proxy.yml#L10

src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -832,6 +832,14 @@ GREENHOUSE_API_KEY = '{{ GREENHOUSE_API_KEY }}'
...
+SECURE_HSTS_INCLUDE_SUBDOMAINS = {{ localsettings.get('SECURE_HSTS_INCLUDE_SUBDOMAINS', True) }}
```
snopoke, line 839: This should probably also be done in nginx directly

src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -832,6 +832,14 @@ GREENHOUSE_API_KEY = '{{ GREENHOUSE_API_KEY }}'
 LOCAL_LOGGING_CONFIG = json.loads('{{ localsettings.get('LOCAL_LOGGING_CONFIG', {})|to_json }}')
 LOCAL_MIDDLEWARE = {{ localsettings.get('LOCAL_MIDDLEWARE', [])|to_json }}
 
+{% if 'SECURE_HSTS_SECONDS' in localsettings %}
+SECURE_CONTENT_TYPE_NOSNIFF = {{ localsettings.get('SECURE_CONTENT_TYPE_NOSNIFF', True) }}
+SECURE_BROWSER_XSS_FILTER = {{ localsettings.get('SECURE_BROWSER_XSS_FILTER', True) }}
```
snopoke, line 837: Already set in nginx: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml#L39

src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -832,6 +832,14 @@ GREENHOUSE_API_KEY = '{{ GREENHOUSE_API_KEY }}'
 LOCAL_LOGGING_CONFIG = json.loads('{{ localsettings.get('LOCAL_LOGGING_CONFIG', {})|to_json }}')
 LOCAL_MIDDLEWARE = {{ localsettings.get('LOCAL_MIDDLEWARE', [])|to_json }}
 
+{% if 'SECURE_HSTS_SECONDS' in localsettings %}
+SECURE_CONTENT_TYPE_NOSNIFF = {{ localsettings.get('SECURE_CONTENT_TYPE_NOSNIFF', True) }}
```
snopoke, line 836: Already set in nginx: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml#L40

gherceg:
@snopoke thanks for the feedback. The intention is to shift from setting these values in nginx over to django settings with the added benefit of receiving warnings when our middleware security practices need to be updated. If there is a reason we should keep the values in nginx I'm curious to hear it. 

@dannyroberts may have some thoughts on this as well.

dannyroberts:
Yeah for a little more context, @gherceg is pretty accurately describing the outcome of a conversation he and I had about how we want handle the fact that Django warns about this—and thus provides a dead simple active compliance process for us to follow—but we already do it in nginx which has wider applicability (i.e. it applies to formplayer too). My take was basically, "hey if it's easy let's do both" so we get the benefit of ongoing recommendations from Django as the recommendations evolve.

snopoke:
Thanks for the explanations. That all makes sense.

environments/staging/public.yml
```diff
@@ -130,6 +130,11 @@ localsettings:
   J2ME_ADDRESS: "{{ J2ME_SITE_HOST }}"
   LOCAL_APPS:
     - {name: 'ddtrace.contrib.django', host: webworkers}
+  LOCAL_MIDDLEWARE:
+    - 'django.middleware.security.SecurityMiddleware'
+  SECURE_HSTS_SECONDS: 300
+  SILENCED_SYSTEM_CHECKS:
+    - 'security.W008'
```
dannyroberts, line 137: From https://docs.djangoproject.com/en/3.1/ref/checks/:
> security.W008: Your SECURE_SSL_REDIRECT setting is not set to True. Unless your site should be available over both SSL and non-SSL connections, you may want to either set this setting to True or configure a load balancer or reverse-proxy server to redirect all connections to HTTPS.

Could you remind me why we want `SECURE_SSL_REDIRECT` to be `False`? Is it that we're just not ready to take that step yet (in which case maybe we let that warning stand), or that it doesn't apply to us / we don't intend to ever set it.

In case it's the problem described below, I'm pasting it, copied from https://docs.djangoproject.com/en/3.1/ref/settings/:

> If turning this to True causes infinite redirects, it probably means your site is running behind a proxy and can’t tell which requests are secure and which are not. Your proxy likely sets a header to indicate secure requests; you can correct the problem by finding out what that header is and configuring the SECURE_PROXY_SSL_HEADER setting accordingly.

But you know what, yeah maybe this doesn't matter at all since, the proxy we have in front of it already turns away HTTP requests and redirects them to HTTPS. Maybe just leave a comment to that effect.

dannyroberts (approved):

environments/staging/public.yml
gherceg: >But you know what, yeah maybe this doesn't matter at all since, the proxy we have in front of it already turns away HTTP requests and redirects them to HTTPS. Maybe just leave a comment to that effect.

Followed this line of thinking and added a comment to explain why we can safely silence this warning.

PR #4299, opened 2020-10-26
add matt's pubkey and username to the dimagi users list

biyeun:
so that matt can deploy :)

##### ENVIRONMENTS AFFECTED
all


environments/_users/dimagi.yml
```diff
@@ -23,6 +23,7 @@ dev_users:
     - sbansal
     - sherbers
     - spokhriyal
+    - matthewriley
```
calellowitz, line 26: two nits:
all other users are first initial last name
all other users are alphabetized

the second is probably more important than the first but it is nice to keep the username convention

calellowitz (commented):
Nothing blocking but a few tiny things

environments/_users/dimagi.yml
biyeun: @calellowitz i figured the username thing would come up... `mattewriley` is his local username and it would make ssh so much easier for him

biyeun: changed the order tho, that's fair

biyeun: also, not everyone's username is exactly first initial last name 👀

calellowitz: you and sravan always breaking things 😛 , but yeah. its a one line ssh config change to make it handle that automatically, and i think conventions are nice. but its obviously not a big deal and still clear who that user is

calellowitz (approved):

PR #4302, opened 2020-10-27
create new update-user command for when update-users is slow AF

biyeun:
`update-users` takes a really long time to run on environments like prod. 

this introduces the `add-new-user` command which can be used in combination with `update-user-key`

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/src/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -337,6 +337,27 @@ class UpdateUsers(_AnsiblePlaybookAlias):
         return AnsiblePlaybook(self.parser).run(args, unknown_args)
 
 
+class AddNewUser(_AnsiblePlaybookAlias):
+    command = 'add-new-user'
+    help = ("Adds a single user to machines for a given username "
+            "and group (because update-users takes forever).")
+    arguments = _AnsiblePlaybookAlias.arguments + (
+        Argument("username", help="username of the user needing to be added"),
+        Argument("dev_group", help="name of the dev group the user is in"),
+    )
+
+    def run(self, args, unknown_args):
+        args.playbook = 'deploy_stack.yml'
+        unknown_args += (
+            '--tags=new_user',
+            '--extra-vars={{"dev_users": {{"present": [{}]}}, "dev_group": "{}"}}'.format(
+                args.username,
+                args.dev_group
+            ),
+        )
+        return AnsiblePlaybook(self.parser).run(args, unknown_args)
+
+
 class UpdateUserPublicKey(_AnsiblePlaybookAlias):
     command = 'update-user-key'
     help = "Update a single user's public key (because update-users takes forever)."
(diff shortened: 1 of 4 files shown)


calellowitz (changes requested):
I think this command is really confusing and only does a small part of what is necessary to set up a user. The new account will exist, but they wont be able to auth as either of the two users they should have access to (doesn't add their key), and even if they can get in, they won't be able to do much of what they should be allowed to because the sudoers file won't be updated.

I am not opposed to trying to make a command that limits the effect to one user in an effort to speed this up, but I think if it does, it should handle the complete process rather than only a small part and therefore leave the user in a bad state.

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -337,6 +337,27 @@ def run(self, args, unknown_args):
         return AnsiblePlaybook(self.parser).run(args, unknown_args)
 
 
+class AddNewUser(_AnsiblePlaybookAlias):
+    command = 'add-new-user'
```
knguyen142, line 341: nit, but can this be `update-user` or `add-user` to denote that it's running `update-users` on a select group. Semantically, `update-users` can also add new users, so removing `new` from it makes it simpler.

biyeun:
@calellowitz i think my goal here was to have this command work in tandem with `update-user-key` as far as the key updates go. `update-user-key` should be run after this. I was not aware of updating the sudoers file. I'll revisit this again tomorrow

millerdev:
> `update-user-key` should be run after this.

Rather than adding a whole new command, could we combine this with `update-user-key` and call it `update-user`?

biyeun:
@millerdev ok sounds good.

biyeun:
@calellowitz @knguyen142 @millerdev thanks for all your inputs! how does the current state of things look?

calellowitz (approved):
This looks much better! Thanks for incorporating all the feedback! Do you have a sense how long this takes to run?

biyeun:
@calellowitz on staging it took a minute

calellowitz:
> @calellowitz on staging it took a minute

Wow. That does sound nice.

calellowitz (changes requested):
I should have caught this earlier, but this totally breaks the `sudoers` file. Since that template fills in allowed users by iterating over `dev_users`, this command replaces the allowed users with just the single user you are updating (plus `ansible` and `cchq`).

Relevant line https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/bootstrap-users/templates/sudoers.j2#L5

Current state of that line on staging `User_Alias HIPAA_USERS = ansible, matthewriley, cchq`

millerdev (commented):
I like the combined/new command. However...

At risk of exposing my own hypocrisy (because I implemented `UpdateUserPublicKey`), I'm a little nervous about switching to a workflow where we rarely or never run `update-users` (because it takes forever). If someone forgets to run one of these commands on an environment when a given person leaves or whatever, that change would previously have been applied the next time `update-users` was run. A single-user workflow means we may never realize that something was forgotten.

If we are moving to a single-user workflow for performance reasons and as a side-effect losing some confidence that all machines can only be accessed by active users, maybe it also makes sense to introduce a regularly-scheduled user access audit?

PR #4317, opened 2020-11-03
Use kafka-topics.sh --list to get topics for kafka data wipe

esoergel:
Fetch the list of topics from kafka itself, rather than django.  On a multi-server environment, kafka is often on a machine that doesn't have a django environment setup, so you can't run management commands directly.

##### ENVIRONMENTS AFFECTED

This only affects environments where we're wiping all data in preparation for restore from backups or data dumps - it shouldn't affect any production environment.

### DIFF
--- a/docs/howto/wipe_persistent_data.md
+++ b/docs/howto/wipe_persistent_data.md
@@ -37,11 +37,10 @@ How To Wipe Persistent Data
 
        $ cchq monolith ap wipe_kafka.yml
 
-   You can check they have been removed by confirming that
-   "__consumer_offsets" is the only topic remaining:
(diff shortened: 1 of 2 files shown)


kaapstorm (approved):

esoergel:
@snopoke pinging you on this since I think we talked about this approach at some point?  I don't recall why we went another direction initially, but this does seem fine.  One notable difference is that this also deletes `__consumer_offsets`, which I think is desirable (we could easily exclude it in the ansible task if needed).  If other services were using the same kafka instance, it would delete those topics too, but I can't think of a scenario where that'd be a reasonable concern.

snopoke (approved):

PR #4318, opened 2020-11-03
Use bionic image on Travis since most (all?) envs are now on 18.04

millerdev:
This may be the reason why Ansible uses pip2 instead of pip3 on Travis (unlike our production machines).

##### ENVIRONMENTS AFFECTED
travis (tests only)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
   - "2.7"
   - "3.6"
 os: linux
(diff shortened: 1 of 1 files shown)


gherceg (approved):
Makes sense to me to upgrade, but probably better to have someone more knowledgeable give the official LGTM.

dannyroberts:
💯

PR #4321, opened 2020-11-04
Switch back to apt-managed supervisor

millerdev:
This should be safe since we have upgraded from Ubuntu 14.04. Using the apt-managed version means security patches will be applied by the system, which is better than having the version hard-coded in an Ansible role.

I will roll this out on all SaaS-managed environments when it has been approved and merged.

##### ENVIRONMENTS AFFECTED
all

Deployed to
- [x] staging
- [ ] production
- [ ] india
- [ ] swiss


millerdev:
Should this be accompanied by a change log entry?

dannyroberts (approved):
nicely done

PR #4323, opened 2020-11-05
Saas 11380/pg repack installation

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

@dannyroberts 
Based on my recent learnings and discussion with you I made following changes to pg_repack process.

1. Added alter queries to execute after the Extension creation.
2. Adding shell script to check and run pg_repacking process if not is already running. In order to avoid permission denied, multiple triggers on pg_repack process.
3. cron to run on every Friday of the week.

Please review and let me know your thoughts. Once approved, I will execute in staging environment.
thanks


dannyroberts (approved):
This looks good to me

prem-fissionhq:
@dannyroberts
As discussed, the next steps after we going through the responses from the pg_repack team. I made changes in the PostgreSQL.yaml to connect the RDS instance from the pg_bouncer machine. I tested the changes in staging environment and it worked fine for me. 
- I also tried to reproduce the issues we faced like`tables missing` in pg_bouncer machine when process killed in middle of execution. There was _no_ impact on the `commcarehq_synclogs` tables in pg_bouncer when the process killed while running directly on the RDS.  
Please review and let me know if you have any other suggestions. Thanks


cc: @snopoke

src/commcare_cloud/ansible/roles/pg_repack/tasks/main.yml
```diff
@@ -62,23 +80,24 @@
     group: "postgres"
     mode: 0755
 
+- name: Copy pg_repack execution script
+  become: yes
+  template:
+    src: "pg_repack_cron.sh.j2"
+    dest: "/usr/local/sbin/pg_repack_cron.sh"
```
snopoke, line 87: you can't use the same filename for each job otherwise the file will get overwritten each time (this is executed in a loop).

src/commcare_cloud/ansible/roles/pg_repack/tasks/main.yml
```diff
@@ -62,23 +80,24 @@
...
-      --host {{ item.host|default('localhost') }}
-      --port {{ item.port|default(5432) }}
-      -d {{ item.database }}
-      {% if item.tables|default(None) %} --tables {{ item.tables|join(' ') }}{% endif %}
-      --table-size-limit {{ item.table_size|default(10) }}
-      --dead-tup-ratio {{ item.ratio|default(0.05) }}
-      {% if item.username|default(None) %} -U {{ item.username }}{% endif %}
-      {% if item.password|default(None) %} -W {{ item.password }}{% endif %}
-      {% if item.skip_superuser_check|default(pg_repack_skip_superuser_check) %} -k{% endif %}"
+    job: "/usr/local/sbin/pg_repack_cron.sh"
```
snopoke, line 97: same comment as above, you need to either pass in the parameters here or have a separate script file for each pg_repack job

snopoke (changes requested):

src/commcare_cloud/ansible/roles/pg_repack/tasks/main.yml
prem-fissionhq: Noted Simon. thank you.

prem-fissionhq:
@shyamkumarlchauhan as discussed, I have updated the pg_repack configuration to the production env files (postgresql.yml and public.yml) please review and let me know if you've more suggestions. 
thanks

shyamkumarlchauhan (approved):

PR #4327, opened 2020-11-10
Enable commcare-cloud --control with python3

millerdev:
Set the environment variable `CCHQ_VIRTUALENV` to `cchq` (or anything other than `ansible`) when running `commcore-cloud --control` commands to cause the virtualenv with that name to be used on the control machine. The virtualenv with will be created with python3 if it does not exist. Example:

```sh
CCHQ_VIRTUALENV=cchq commcare-cloud <env> --control ...
```

Use of virtualenvwrapper by init.sh will be phased out with Python 2.

##### ENVIRONMENTS AFFECTED
None. This is currently for testing only. Running `commcare-cloud` with Python 3 is still in development.


gherceg (approved):
Offlined with Daniel to get more context. Looks good.

dannyroberts (approved):
Nicely done

PR #4332, opened 2020-11-12
Remove redundent tmpreaper task on formplayer /tmp directory

dannyroberts:
it is already handled by a more general one that is on all machines. This was causing harmless but spammy errors when two different processes tried to delete the same files.

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -14,8 +14,18 @@
     user: root
     cron_file: purge_formplayer_files
   with_items:
-    - {name: 'Purge temp files', dir: '/tmp', time_spec: '2d'}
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -14,11 +14,21 @@
     user: root
     cron_file: purge_formplayer_files
   with_items:
-    - {name: 'Purge temp files', dir: '/tmp', time_spec: '2d'}
     - {name: 'Purge sqlite db files', dir: '{{ formplayer_data_dir }}', time_spec: '{{ formplayer_purge_time_spec }}'}
   tags:
     - cron
 
+- name: Remove redundant purging cron jobs (cleanup)
```
knguyen142, line 21: Should this be committed to the code base to be run everytime? Should we just run this ad-hoc once?

millerdev: I think we want to commit it since there are environments outside of our control who may want to run it as well.

millerdev (approved):

PR #4347, opened 2020-11-20
Explicitly use io.open in commands directory

gherceg:
[Replace open with io.open](https://dimagi-dev.atlassian.net/browse/SAAS-11366)

In addition to addressing the areas in the `commands` directory that use `open`, I added a new helper in `python_migration_utils` aimed at replacing the existing methods since all address the same issue/have the same implementation. 
##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/src/commcare_cloud/commands/terraform/aws.py
+++ b/src/commcare_cloud/commands/terraform/aws.py
@@ -23,7 +23,7 @@ from commcare_cloud.cli_utils import print_command
 from commcare_cloud.colors import color_notice, color_success
 from commcare_cloud.commands.command_base import Argument, CommandBase
 from commcare_cloud.environment.main import get_environment
-from commcare_cloud.python_migration_utils import open_python_dependent
+from commcare_cloud.python_migration_utils import open_for_write
 
 
 def check_output(cmd_parts, env, silent=False):
@@ -145,11 +145,11 @@ class AwsFillInventory(CommandBase):
 
         if not args.cached:
             resources = get_aws_resources(environment)
-            with open_python_dependent(environment.paths.aws_resources_yml) as f:
+            with open_for_write(environment.paths.aws_resources_yml) as f:
                 f.write(yaml.safe_dump(resources, default_flow_style=False))
         else:
             with open(environment.paths.aws_resources_yml, 'r', encoding='utf-8') as f:
-                # PY2: yaml.safe_load will return bytes regardless of read mode
+                # PY2: yaml.safe_load will return bytes when the content is ASCII-only bytes
                 resources = yaml.safe_load(f.read())
 
         with open(environment.paths.inventory_ini_j2, 'r', encoding='utf-8') as f:
@@ -532,7 +532,7 @@ def _write_credentials_to_aws_credentials(
     config.set(aws_profile, 'aws_secret_access_key', aws_secret_access_key)
     config.set(aws_profile, 'aws_session_token', aws_session_token)
     config.set(aws_profile, 'expiration', expiration.strftime("%Y-%m-%dT%H:%M:%SZ"))
-    with open_python_dependent(AWS_CREDENTIALS_PATH) as f:
+    with open_for_write(AWS_CREDENTIALS_PATH) as f:
         config.write(f)
 
 
@@ -570,7 +570,7 @@ def _write_profile_for_sso(
     config.set(section, 'sso_role_name', 'PowerUserAccessPlus')
     config.set(section, 'region', region)
     config.set(section, 'output', 'json')
-    with open_python_dependent(AWS_CONFIG_PATH) as f:
+    with open_for_write(AWS_CONFIG_PATH) as f:
         config.write(f)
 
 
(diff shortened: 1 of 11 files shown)


src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -532,7 +532,7 @@ def _write_credentials_to_aws_credentials(
     config.set(aws_profile, 'aws_secret_access_key', aws_secret_access_key)
     config.set(aws_profile, 'aws_session_token', aws_session_token)
     config.set(aws_profile, 'expiration', expiration.strftime("%Y-%m-%dT%H:%M:%SZ"))
-    with open(AWS_CREDENTIALS_PATH, 'w') as f:
+    with open_python_dependent(AWS_CREDENTIALS_PATH) as f:
         config.write(f)
```
gherceg, line 536: This is the call that made me think we need the same helper used for json dumps and yaml dumps for scenarios like these. Here is the implementation of `config.write()` in a module that does not seem to import `unicode_literals`: 
```    
def write(self, fp):
        """Write an .ini-format representation of the configuration state."""
        if self._defaults:
            fp.write("[%s]\n" % DEFAULTSECT)
            for (key, value) in self._defaults.items():
                fp.write("%s = %s\n" % (key, str(value).replace('\n', '\n\t')))
            fp.write("\n")
        for section in self._sections:
            fp.write("[%s]\n" % section)
            for (key, value) in self._sections[section].items():
                if key == "__name__":
                    continue
                if (value is not None) or (self._optcre == self.OPTCRE):
                    key = " = ".join((key, str(value).replace('\n', '\n\t')))
                fp.write("%s\n" % (key))
            fp.write("\n")
```

src/commcare_cloud/python_migration_utils.py
```diff
@@ -5,6 +5,19 @@
 import six
 
 
+def open_python_dependent(path):
```
gherceg, line 8: open to other names

millerdev: I think the name should mention or imply "write" mode. Maybe `open_for_write`? The docstring can explain how it's behavior depends on the version of Python.

src/commcare_cloud/commands/terraform/terraform.py
```diff
@@ -87,7 +92,7 @@ def run(self, args, unknown_args):
                 else ''
             )
 
-            with open(os.path.join(run_dir, 'secrets.auto.tfvars'), 'w') as f:
+            with open_python_dependent(os.path.join(run_dir, 'secrets.auto.tfvars')) as f:
                 print('rds_password = {}'.format(json.dumps(rds_password)), file=f)
```
millerdev, line 96: This doesn't need to use `open_python_dependent`, does it? The template string is a literal, so text. Please tell me `environment.get_secret()` returns text (if not, we should make it so).

src/commcare_cloud/python_migration_utils.py
```diff
@@ -5,6 +5,19 @@
...
+        - json.dump and json.dumps attempt to write byte strings on PY2, and text strings on PY3
+        - same for yaml.dump and yaml.safe_dump
+        - passing the file stream into a library that writes byte strings in PY2
+    """
+    if six.PY2:
+        return open(path, "wb")
+    return open(path, "w", encoding="utf-8")
+
+
 def open_for_json_dump(path):
```
millerdev, line 21: Can we remove this or change the implementation to use `open_python_dependent`? Same with `open_for_yaml_dump`.

millerdev (commented):
Probably worth adding a note that PyYAML loads and returns `bytes` (`str`) in Python 2, even when the source file is opened with `encoding="utf-8"`, as noted in https://stackoverflow.com/questions/2890146/how-to-force-pyyaml-to-load-strings-as-unicode-objects . As we have previously discussed, to my knowledge there is no good/non-hacky solution to that problem, so we are proceeding without any special workaround. In practice, this means that strings will automatically be coerced from bytes to text by Python 2 when they are read from a YAML file and then later combined with other text (`unicode`) objects. Also note that PyYAML does return text when loading content that contains non-ASCII characters.

src/commcare_cloud/python_migration_utils.py
gherceg: Yeah and the method already lives in a module named `python_migration_utils` so I agree it doesn't need to be in the method name. `open_for_write` sounds good to me.

src/commcare_cloud/commands/terraform/terraform.py
gherceg: Ah I believe you are right. I saw `json.dumps` and didn't think enough about the rest of this.

Just to talk through it, (on Python 2) as long as `rds_password` contains ASCII characters only, `json.dumps` will return a bytes string which will then be converted to unicode due to the string literal formatting.

millerdev: > as long as `rds_password` contains ASCII characters only, `json.dumps` will return a bytes string

`json.dumps` will always return ASCII bytes on Python 2. Non-ASCII characters are escaped.

> which will then be converted to unicode due to the string literal formatting.

correct.

millerdev (approved):
:medal_military:

PR #4351, opened 2020-11-23
allow any callback options (remove 'choices')

snopoke:
This removes the restriction on which callback options can be used allowing the 'user' to choose any of the available options.


millerdev (approved):
Description of "Why?" would be nice.

PR #4352, opened 2020-11-23
rolling restart playbook for webworkers and formplayer

snopoke:
I used this today to roll out config changes without impacting usage. In the long run it could be made into a dedicated command with better error handling etc. but just putting it here for now.

### DIFF
--- a/src/commcare_cloud/ansible/hq_rolling_restart.yml
+++ b/src/commcare_cloud/ansible/hq_rolling_restart.yml
@@ -1,0 +1,38 @@
+---
+- name: Webworkers
+  hosts: webworkers
+  become: true
+  serial: "{{ batch_size|default(1) }}"
+  tags: webworkers
+  tasks:
+    - name: Restart webworker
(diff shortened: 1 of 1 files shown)


millerdev:
We already do rolling restart for webworkers, right? Are you proposing that we migrate to this and deprecate the old way?

https://github.com/dimagi/commcare-cloud/blob/ae878bb634a7cff98ecf4d9d4efee1ac0bd46e30/src/commcare_cloud/fab/operations/supervisor.py#L98-L99

snopoke:
The fab version does not check for the worker to be up and it also removes the worker from the LB beforehand which I don't think is necessary during off peak. The fab version also does not have the ability to do them in batches or more than 1.

millerdev (approved):

millerdev:
I wonder if there's a good place to document the differences between the two different options? One good place would be the `--help` output for this playbook, although I'm not sure it's possible to customize that.

snopoke:
@millerdev I've added some text at the top of the playbook that outline the usage and differences with the fabric tasks.

PR #4356, opened 2020-11-23
Explicitly use io.open in remaining files in src/commcare_cloud

gherceg:
[Replace open with io.open](https://dimagi-dev.atlassian.net/browse/SAAS-11366)

Continuation of the PRs to replace open calls in python 2/3 with io.open calls. 
##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/src/commcare_cloud/fab/utils.py
+++ b/src/commcare_cloud/fab/utils.py
@@ -1,7 +1,5 @@
-from __future__ import absolute_import
-from __future__ import print_function
+from __future__ import absolute_import, print_function, unicode_literals
 
-from __future__ import unicode_literals
 import datetime
 import os
 import pickle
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/fab/utils.py
```diff
@@ -202,9 +199,9 @@ def _get_env_filename():
 
 
 def cache_deploy_state(command_index):
-    with open(os.path.join(PROJECT_ROOT, _get_checkpoint_filename()), 'w') as f:
+    with open(os.path.join(PROJECT_ROOT, _get_checkpoint_filename()), 'wb') as f:
```
gherceg, line 202: Applies to all of the instances where pickle is used, but from what I read it seems that pickling is always dealing with a byte stream which I why I chose to explicitly have a byte stream on both 2 and 3.

millerdev (approved):

PR #4358, opened 2020-11-23
increase shard size and slow threshold

calellowitz:
##### ENVIRONMENTS AFFECTED

Production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -503,50 +503,50 @@ rds_instances:
     engine_version: 9.6.15
 
   - identifier: "pgshard1-production"
-    instance_type: "db.m5.2xlarge"
+    instance_type: "db.m5.4xlarge"
     storage: 500
     max_storage: 2500
     multi_az: true
     engine_version: 9.6.15
     params:
       shared_preload_libraries: pg_stat_statements
-      log_min_duration_statement: 1000
+      log_min_duration_statement: 5000
   - identifier: "pgshard2-production"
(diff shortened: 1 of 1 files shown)


ctsims (approved):

ctsims (approved):

knguyen142 (approved):
Looks good, but we should note before doubling db instance again that for https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html `db.m5.8xlarge`, `PostgreSQL 11.6 & higher, 10.11 & higher, 9.6.16 & higher, 9.5.20 & higher` is supported while we are on 9.6.15.

knguyen142:
Do you want to increase `max_storage` on these as well just in case?

calellowitz:
> Looks good, but we should note before doubling db instance again that for https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html `db.m5.8xlarge`, `PostgreSQL 11.6 & higher, 10.11 & higher, 9.6.16 & higher, 9.5.20 & higher` is supported while we are on 9.6.15.

Great catch. It never even occurred to me that certain instances were not supported. It shouldn't be too hard to do a minor version upgrade, if that is required. 9.5.15 -> 9.5.16 should be pretty painless.

As far as disk space, I think we are fine. I am not sure about log retention policies, but a significant percent of today's growth was just logs and those will be much smaller going forward. It's easy enough to make that decision at a later date as well, if it continues to grow.

PR #4360, opened 2020-11-24
remove unused / legacy fabric code

snopoke:
##### ENVIRONMENTS AFFECTED
None


src/commcare_cloud/fab/operations/release.py
```diff
@@ -460,17 +459,6 @@ def copy():
     return copy
 
 
-@parallel
-@roles(ROLES_FORMPLAYER)
-def copy_formplayer_properties():
```
snopoke, line 465: this was part of the lazy migration for moving formplayer to it's own release location and is no longer required

orangejenny (approved):

PR #4361, opened 2020-11-24
remove commcare-hq source from formplayer machines

snopoke:
Removing 'formplayer' from the `ROLES_ALL_SRC` role means that the commcare-hq code
will not get deployed to these nodes. In the past formplayer was deployed inside the HQ code deploy structure
but that is no longer the case. There is no need to have the HQ code on formplayer machines.

##### ENVIRONMENTS AFFECTED
ALL

### DIFF
--- a/src/commcare_cloud/fab/const.py
+++ b/src/commcare_cloud/fab/const.py
@@ -12,7 +12,6 @@ ROLES_ALL_SRC = [
     'django_app',
     'django_celery',
     'django_pillowtop',
-    'formplayer',
     'staticfiles',
     'airflow',
     'django_manage'
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -272,7 +272,7 @@ def env_common():
     env.resume = False
     env.offline = False
     env.full_deploy = False
-    env.supervisor_roles = ROLES_ALL_SRC
+    env.supervisor_roles = ROLES_ALL_SERVICES
 
 
 @task


millerdev (approved):
Do we need to do any testing for this? It looks like a pretty simple change, but would hate to have overlooked some implicit dependency. I suppose staging does not have stand-alone formplayer nodes, right?

snopoke:
@millerdev I will test this on staging which does have separate formplayer machines

snopoke:
Did a staging deploy from this branch and also removed the `www/staging/current` symlink and renamed the `www/staging/releases` folder. Formplayer has continued to work fine.

snopoke:
Will merge and do a deploy on Thursday when usage should be very low.

millerdev:
@snopoke: @gherceg and I recreated the `current` symlink on the staging formplayer machines because that change broke the deploy on staging. We're still looking into why it failed.

gherceg:
> We're still looking into why it failed.

I think we can safely say it failed because a deploy is only setup to clone a virtualenv from the _current_ release. https://github.com/dimagi/commcare-cloud/blob/60c07b9089e10e406a9e7d01779e220799ea6a9c/src/commcare_cloud/fab/operations/release.py#L304-L307

The only way we create a new virtualenv is through the `Upgrade python packages` task
https://github.com/dimagi/commcare-cloud/blob/9c2547de64f3e88145c364a28e39e246fac303a5/src/commcare_cloud/ansible/roles/commcarehq/tasks/main.yml#L92-L94

@millerdev and I agree that since this is a rare case the best path forward would be to add an `else` condition that raises an error to make the cause of the failure obvious.

snopoke:
@millerdev apologies for that, I thought I had restored the state to what it needed to be

esoergel:
If I understand this correctly, there are two scenarios:

1. Formplayer is on a machine with other services (eg, in a monolith or small cluster).  My read is this change won't affect those environments at all, is that correct?

2. Formplayer is on a standalone machine.  This change will result in future deploys not pushing HQ code to those machines.  I assume that means existing machines will continue to have vestigial release directories that will just stop being updated, is that right?  Is it worth cleaning those up, or fine to leave around?

snopoke:
> 1. Formplayer is on a machine with other services (eg, in a monolith or small cluster).  My read is this change won't affect those environments at all, is that correct?

That is correct.

> 2. Formplayer is on a standalone machine.  This change will result in future deploys not pushing HQ code to those machines.  I assume that means existing machines will continue to have vestigial release directories that will just stop being updated, is that right?  Is it worth cleaning those up, or fine to leave around?

Yes it would be worth cleaning that up but it can be done separately.

dannyroberts (approved):
Very approve!

PR #4364, opened 2020-11-24
Raise exception if /current release directory does not exist when deploying 

gherceg:
Discovered this issue when HQ code was removed from formplayer machines. The machines were unable to create a new virtual environment during deploy. This is ok behavior since an ansible command handles creating a virtualenv when first being setup, and we do not expect to remove the /current directory from any machine that needs it. This will make it easier to debug in the future if it ever comes up.
##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -303,9 +303,7 @@ def update_virtualenv(full_cluster=True):
             _clone_virtual_env(env.py3_virtualenv_current, env.py3_virtualenv_root)
         elif not exists(env.py3_virtualenv_current):
             raise EnvironmentException(
-                "{} cannot be found. This means there is no virtual environment available to clone from. "
-                "You will need to recreate the /current directory on this machine that symlinks to the most " 
-                "recent release in order to deploy successfully.".format(env.py3_virtualenv_current)
+                "Virtual environment not found: {}".format(env.py3_virtualenv_current)
             )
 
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/fab/operations/release.py
```diff
@@ -304,6 +301,12 @@ def update():
         # Optimization if we have current setup (i.e. not the first deploy)
         if exists(env.py3_virtualenv_current) and not exists(env.py3_virtualenv_root):
             _clone_virtual_env(env.py3_virtualenv_current, env.py3_virtualenv_root)
+        elif not exists(env.py3_virtualenv_current):
+            raise EnvironmentException(
+                "{} cannot be found. This means there is no virtual environment available to clone from. "
+                "You will need to recreate the /current directory on this machine that symlinks to the most " 
+                "recent release in order to deploy successfully.".format(env.py3_virtualenv_current)
```
gherceg, line 308: Will this print out the path as I intend it to?

millerdev, line 308: You mean does string literal concatenation happen before formatting? Yes
```py
>>> "x{}"  "z".format("y")
'xyz'
```

However, missing the `current` symlink is just one possible reason this condition might fail, so the error message seems a bit more specific than necessary.

```suggestion
                "Virtual environment not found: {}".format(env.py3_virtualenv_current)
```

millerdev (approved):

PR #4366, opened 2020-11-27
add new settings

mkangia:
For
https://github.com/dimagi/commcare-hq/pull/28785
https://github.com/dimagi/commcare-hq/pull/28796
https://github.com/dimagi/commcare-hq/pull/28797/files#diff-8440ff6f063677238da8fc25000acb8bbfb4345c3b9d97a0cf81414d15702265R704

##### ENVIRONMENTS AFFECTED
icds-cas, icds-staging, icds-sandbox with new settings
for india, the setting `ENABLE_DRACONIAN_SECURITY_FEATURES` that was added for icds project originally on India, was removed. This is because it is not needed anymore and was infact, a left over from when icds project was moved out of india environment.


environments/india/public.yml
```diff
@@ -117,6 +117,9 @@ localsettings:
   EMAIL_SMTP_PORT: 587
   EMAIL_USE_TLS: yes
   ENABLE_DRACONIAN_SECURITY_FEATURES: yes
+  ADD_CAPTCHA_FIELD_TO_FORMS: True
+  DISABLE_AUTOCOMPLETE_ON_SENSITIVE_FORMS: True
+  CUSTOM_PASSWORD_STRENGTH_MESSAGE: 'Password Requirements: 1 special character, 1 number, 1 capital letter, minimum length of 8 characters.'
```
snopoke, line 122: Why are we adding these to `india`? I don't think they should be present there.

mkangia: that is a good point Simon. And a parallel conversation is needed, to check on why we even have the draconian features available on India, I guess that configuration was not removed after moving icds-cas out of india.
I will be taking that up. These changes are currently made in the PR to just wait for that confirmation.

snopoke: Correct, they were needed prior to having a dedicated icds-staging env. They can (and should) be removed.

mkangia: Hey Simon, 
Updated. Please share review on recent changes.

snopoke (approved):

mkangia:
this is deployed on India, no more captchas!

PR #4368, opened 2020-12-01
Increase pgucr0-production RDS instance to make use of RIs

dannyroberts:
It hasn't been increased this year while the rest of the cluster expanded around it.
It's the only db.t3 we use in production and since the maintainance in the last couple weeks
we are no longer using all of our RIs. Based on those two things it seemed worth it to throw
more hardware at this instance since we are already committed to paying for that regardless.

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -496,7 +496,7 @@ rds_instances:
       log_min_duration_statement: 1000
 
   - identifier: "pgucr0-production"
-    instance_type: "db.t3.large"
+    instance_type: "db.t3.2xlarge"  # increased from db.t3.large due to unused db.t3 RIs
     storage: 500
     max_storage: 5000
     multi_az: true


dannyroberts:
@sanjay2916 once this gets merged can you apply it during the IT maintenance window? (1) stop pillows (2) apply the RDS change (3) start pillows again once it's done.

sanjay2916:
> @sanjay2916 once this gets merged can you apply it during the IT maintenance window? (1) stop pillows (2) apply the RDS change (3) start pillows again once it's done.

@dannyroberts Sure, I will apply those changes, once PR merged

gherceg (approved):
The change looks good. Need to do some research into reserved instances myself, but how do we know this specific change (adding 2x and not something larger) will take full advantage of our RIs?

dannyroberts:
We'll be able to tell once this is applied, but there's no penalty for being under, just for being over, and a quick back of the envelope yesterday convinced me this would keep us under. https://confluence.dimagi.com/display/internal/SaaS+AWS+Cost+Optimization has a bit more information about how we use RIs (and Compute Savings Plans).

calellowitz (approved):

dannyroberts:
Great, thanks @sanjay2916. Good to go!

PR #4372, opened 2020-12-03
added couchdb cluster playground

knguyen142:
https://dimagi-dev.atlassian.net/browse/SAAS-11533

##### ENVIRONMENTS AFFECTED

None, do not merge, mainly for review.

This sets up a playground to try to reproduce the couchdb inconsistency.

- Docker-compose.yml to set up 3 nodes
  - cluster values `{'q': 8, 'n': 3, 'w': 2, 'r': 2}`
- steps.py to create scenario (functions to run doc creation thread, bring up/tear down a cluster, bring up/tear down a node)
- Uses a modified https://github.com/dimagi/datadog-checks/pull/23 to check for inconsistency

tldr; most of the time was spent on creating this environment, I have not been able to reproduce it so far. 

Instructions to set up in `README.md`

Sample output:
```
inserted 1 total documents, 1 new documents since last reported
Stopping server_1 ... 
inserted 297 total documents, 296 new documents since last reported
inserted 656 total documents, 359 new documents since last reported
inserted 868 total documents, 212 new documents since last reported
inserted 1126 total documents, 258 new documents since last reported
inserted 1418 total documents, 292 new documents since last reported
inserted 1654 total documents, 236 new documents since last reported
inserted 1960 total documents, 306 new documents since last reported
inserted 2243 total documents, 283 new documents since last reported
Starting server-1 ... 
Stopping server_2 ... 
inserted 2508 total documents, 265 new documents since last reported
inserted 2670 total documents, 162 new documents since last reported
inserted 2919 total documents, 249 new documents since last reported
inserted 3095 total documents, 176 new documents since last reported
inserted 3259 total documents, 164 new documents since last reported
inserted 3455 total documents, 196 new documents since last reported
inserted 3652 total documents, 197 new documents since last reported
inserted 3901 total documents, 249 new documents since last reported
inserted 4162 total documents, 261 new documents since last reported
Starting server-2 ... 
insertion done: inserted 4330 total documents
{'http://localhost:5984': 4330, 'http://localhost:15984': 4330, 'http://localhost:25984': 4330}
```

Next step is to create three threads to create documents on all 3 nodes (instead of just node-0 right now). Any other suggestions welcomed.

Sources used to create environment:
- https://medium.com/@singinjon/cluster-couchdb-2-and-docker-on-localhost-f0490649d960 
- https://github.com/apache/couchdb-docker/issues/74#issuecomment-644592838
- https://docs.couchdb.org/en/latest/setup/cluster.html#the-cluster-setup-api


knguyen142:
This is just for review.

couchdb-cluster-playground/lib/couchdb_custom.py
```diff
@@ -0,0 +1,99 @@
...
+        return self._request(self.local_port, path, host, url)
+
+    def _request(self, port, path, host, url=None):
+        url = url or "http://{}:{}".format(host or self.host, port)
+        response = self.session.get(url=urljoin(url, path))
+        response.raise_for_status()
+        return response.json()
+
+
+class CouchDBCustom():
```
millerdev, line 31: Is this class used anywhere? Maybe other parts of this module are unused as well?

couchdb-cluster-playground/steps.py
```diff
@@ -0,0 +1,152 @@
...
+
+def create_docs(reporting_interval=4):
+    def create_docs_thead():
+        # thread that runs and inserts documents
+        total_count = 0
+        last_reported_time = 0
+        new_count = 0
+
+        while should_create_docs:
+            doc = {"mydoc": "test"}
```
millerdev, line 38: It may be useful to try creating larger documents. Since we're hypothesizing that there is a bug in Couch it's possible it is able to easily keep up with tiny documents, but would have a harder time with larger ones.

couchdb-cluster-playground/steps.py
```diff
@@ -0,0 +1,152 @@
...
+                print(f'inserted {total_count} total documents, {new_count} new documents since last reported')
+                new_count = 0
+                last_reported_time = cur_time
+        print(f'insertion done: inserted {total_count} total documents')
+
+    # create database and start thread
+    server = Server("http://admin:password@localhost:5984")
+    db = server.get_or_create_db("shards")
+    x = threading.Thread(target=create_docs_thead)
+    x.start()
```
millerdev, line 53: Would creating a bunch of threads here create more load on Couch? Something like
```suggestion
    for i in range(num_threads):
        x = threading.Thread(target=create_docs_thead)
        x.start()
```

couchdb-cluster-playground/steps.py
```diff
@@ -0,0 +1,152 @@
...
+    password = 'password'
+    with requests.Session() as session:
+        session.auth = (user, password)
+        context = couch_info.SessionContext(host, port, local_port, session)
+        shards = couch_info.get_cluster_shard_details(context, [node['url'] for node in NODES])
+
+        doc_counts = {shard['node']: shard['doc_count'] for shard in shards}
+        print(doc_counts)
+
+        return len(set(doc_counts.values())) == 1
```
millerdev, line 86: Would it make sense to also check this against the total number of documents created?

couchdb-cluster-playground/steps.py
```diff
@@ -0,0 +1,152 @@
...
+            if ((cur_time - last_reported_time) > reporting_interval):
+                print(f'inserted {total_count} total documents, {new_count} new documents since last reported')
+                new_count = 0
+                last_reported_time = cur_time
+        print(f'insertion done: inserted {total_count} total documents')
+
+    # create database and start thread
+    server = Server("http://admin:password@localhost:5984")
+    db = server.get_or_create_db("shards")
+    x = threading.Thread(target=create_docs_thead)
```
millerdev, line 52: There are limitations to Python's threading performance. A library like asyncio (Python 3 only) or gevent may yield better performance, and should work well since this task should be primarily IO bound.

millerdev (commented):
:+1: looking good

couchdb-cluster-playground/lib/couchdb_custom.py
knguyen142: Yes that's right. I mostly copied and pasted the other file and only made modifications where needed so it didn't diverge too much.

couchdb-cluster-playground/steps.py
knguyen142: I think that might help if we wanted to explore the concurrency of a server, but I wonder how concurrent it will be with the Python GIL. I can try this, I was going to have different threads hit all 3 servers anyway.

couchdb-cluster-playground/steps.py
knguyen142: Ah currently it doesn't actually check for whether the request succeeded in creating a doc (misnomer in total docs created). I think the shard doc count is the source of truth though? Also, right now multiple runs will go off of the same `shards` db so docs created != shard docs, I thought having a big shard count from previous runs would be helpful as well.

couchdb-cluster-playground/steps.py
knguyen142: Yeah that would be helpful if we were testing the concurrency of the servers, but I think it's the replication that's probably giving us trouble? I'll see if it's not too much trouble to switch to an event model, though, since it would be better.

couchdb-cluster-playground/steps.py
knguyen142: I thought about this too, but figured many tiny docs was harder. But yeah I can put in a random length string.

couchdb-cluster-playground/steps.py
millerdev: Regardless of whether we are testing concurrency or not, I think it may be helpful to flood Couch with requests to the point that replication has a hard time keeping up with incoming data. It seems like it may be hard to do that with a single thread.

knguyen142: That's a good idea, I'll look into using `ab` to hammer it.

knguyen142:
I modified it to shell out and use `ab` to control the concurrency, create separate threads to hammer each node individually, and set high traffic such that a server usually goes down unintentionally at the end. The check via `db_is_consistent` says that it's consistent. I can show you a demo on Mon to dig deeper, but it seems pretty resilient so far.

knguyen142:
I tried to reproduce it without success, this was hammering the three nodes individually with huge requests (in volume and doc size) and took 2 nodes down sequentially. I ran it multiple times and they all eventually recovered. My guess is it’s not something easy to reproduce on a couchdb cluster in a good state, and at some point we got into a bad state and it is triggered more easily.

Will close this unless anyone else has any other ideas.

millerdev:
@knguyen142 Thanks for looking into it. I agree with your conclusion.

PR #4373, opened 2020-12-04
Fix python 3 related deploy issues

gherceg:
These appear to be the last hurdles to successfully run a python 3 deploy. 

1) arguments from commcare-cloud commands were explicitly encoded, which led to issues on python 3 since conversions from bytes to text were no longer done automatically. It is unclear why the arguments needed to be encoded in the first place.

2) The zipp library depended on `importlib-metadata` in addition to `importlib-resources` so that dependency was added to requirements3.txt

3) A `TypeError` was raised when caching the deploy state via pickling. The pickling failed because of a change on Python 3 that no longer supported pickling of `RLock` objects. Fortunately, the object that contained this attribute was under our control (`DeployMetadata`), and was due to a `memoized` property. Since this can be rebuilt when unpickling the object with no issues, a slightly hacky solution was used to avoid pickling any `memoized` property by overriding `__getstate__` on `DeployMetadata` and removing any attribute with the `_<name>_cache` naming scheme.

##### ENVIRONMENTS AFFECTED

None


src/commcare_cloud/alias.py
```diff
@@ -24,9 +24,24 @@ def encode_string(value):
...
+        else:
+            argv.extend(['--{}'.format(key), value])
+    return argv
+
+
 def commcare_cloud(*args, **kwargs):
     from .cli_utils import print_command
     from .commcare_cloud import call_commcare_cloud
-    argv = _encode_args('commcare-cloud', *args, **kwargs)
+    argv = _generate_args('commcare-cloud', *args, **kwargs)
```
gherceg, line 45: @dannyroberts I made this change last week because it wasn't obvious to me why the arguments needed to be encoded for either 2 or 3 but maybe you can provide some additional context here?

dannyroberts: Interesting. Does this change work both in python 2 and 3? I think since `' '` in python 2 means `b' '` and `' '` in python 3 means `u' '`, the like least-change change would be to replace `' '.join(shlex_quote(arg) for arg in command)` with `b' '.join(shlex_quote(arg) for arg in command)`, but I'm honestly not sure whether the `sys.argv` and functions we use to run commands are expected to use list/tuples of `bytes` or of `unicode`. So I'm not sure whether the change you made here or the one I brought up is the more correct fix

millerdev: > I think since `' '` in python 2 means `b' '`

Except not in the module because of `from __future__ import unicode_literals`

I don't think we want to use bytes in Python 3, so `b' '.join(...)` seems wrong to me.

The issue @gherceg hit was due to printing, and I think will not be resolved by that change anyway.

https://github.com/dimagi/commcare-cloud/blob/d38f180f52f25219878cd170a10dce0211a76019/src/commcare_cloud/cli_utils.py#L96-L97

millerdev (approved):

src/commcare_cloud/alias.py
gherceg: Python docs on sys.argv - `Note On Unix, command line arguments are passed by bytes from OS. Python decodes them with filesystem encoding and “surrogateescape” error handler. When you need original bytes, you can get it by [os.fsencode(arg) for arg in sys.argv].`

This does work on python 2 (tested via staging deploy), as well as python 3.

millerdev (approved):

millerdev:
:tada:

PR #4380, opened 2020-12-09
Add pgbouncer machines

dannyroberts:
##### ENVIRONMENTS AFFECTED
production


environments/production/aws-resources.yml
```diff
@@ -55,342 +55,6 @@ esmaster4-production.instance_id: i-0be1b553bdc9f0f5c
 esmaster5-production: 10.202.40.226
 esmaster5-production.instance_id: i-01e1279564c30ece6
 formplayer-efs: fs-150b0596.efs.us-east-1.amazonaws.com
-formplayer_b000-production: 10.202.10.90
```
millerdev, line 58: These seem like an odd thing to have in this PR. Is it here because other things we are doing concurrently are conflicting with the changes in this PR?

millerdev (approved):

environments/production/aws-resources.yml
dannyroberts: Yeah, this got in the way as I was trying to make the changes more central to this PR. After rolling out https://github.com/dimagi/commcare-cloud/pull/4378 yesterday I stopped (in the AWS console) the formplayer_b machines. aws-resources.yml is generated from what's actually in AWS so if I didn't then also remove formplayer_b from terraform, `cchq production aws-fill-inventory` was failing, and so I rolled that change in with this one to get it out quickly. The cleaner thing to do would have been to put it into that first PR, or to make a separate followup PR yesterday removing formplayer_b.

PR #4386, opened 2020-12-10
Add monit to formplayer to restart if OOM occurs

gherceg:
[Monit Investigation](https://dimagi-dev.atlassian.net/browse/SAAS-11555)

This is a solution to a symptom of recent formplayer issues, which is frequent OOMs that do not bring the entire formplayer process down. Similar to elasticsearch, the method here is to search the formplayer logs for an OOM and restart using supervisor if/when that occurs. 

##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -144,8 +144,8 @@
   tags:
     - localsettings
 
-# This should be the last task in the file
-# (and certainly the last one tagged with `formplayer_deploy`)
+# This should be the last task in the formplayer release process
+# (and the last one tagged with `formplayer_deploy`)
 # so that if setting up a release fails, `current` never gets linked to it
 # giving us nearly atomic formplayer deploys
 - name: Link formplayer jar
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -174,13 +174,16 @@
 
 - meta: flush_handlers
 
-- monit:
-    name: formplayer
-    state: monitored
-  tags: monit
-  ignore_errors: "{{ ansible_check_mode }}"
-  register: result
-  until: not result.failed or 'process not presently configured with monit' not in result.msg
-  retries: 5
-  delay: 20
+# NOTE: this check currently fails because of a bug in community.general.monit
+# A PR has been made to fix this, and we can uncomment this once that is merged/released
+# https://github.com/ansible-collections/community.general/pull/1532
+#- community.general.monit:
+#    name: formplayer
+#    state: monitored
+#  tags: monit
+#  ignore_errors: "{{ ansible_check_mode }}"
+#  register: result
+#  until: not result.failed or 'process not presently configured with monit' not in result.msg
+#  retries: 5
+#  delay: 20
 
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -183,3 +183,4 @@
   until: not result.failed or 'process not presently configured with monit' not in result.msg
   retries: 5
   delay: 20
+
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -161,19 +161,17 @@
     - formplayer_deploy
   when: not _should_update_formplayer_in_place
 
-- block:
 
-  - name: Formplayer monit config
-    become: yes
-    template:
-      src: "monit.formplayer.j2"
-      dest: "/etc/monit/conf.d/formplayer"
-      group: root
-      owner: root
-      mode: 0640
-    notify: reload monit
-    tags: monit
-
+- name: Formplayer monit config
+  become: yes
+  template:
+    src: "monit.formplayer.j2"
+    dest: "/etc/monit/conf.d/formplayer"
+    group: root
+    owner: root
+    mode: 0640
+  notify: reload monit
+  tags: monit
   when: use_monit_for_formplayer == true
 
 - meta: flush_handlers
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -160,3 +160,26 @@
   tags:
     - formplayer_deploy
   when: not _should_update_formplayer_in_place
+
+- name: Formplayer monit config
+  become: yes
+  template:
+    src: "monit.formplayer.j2"
+    dest: "/etc/monit/conf.d/formplayer"
+    group: root
+    owner: root
+    mode: 0640
+  notify: reload monit
+  tags: monit
+
+- meta: flush_handlers
+
+- monit:
+    name: formplayer
+    state: monitored
+  tags: monit
+  ignore_errors: "{{ ansible_check_mode }}"
+  register: result
+  until: not result.failed or 'process not presently configured with monit' not in result.msg
+  retries: 5
+  delay: 20
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -183,3 +183,4 @@
   until: not result.failed or 'process not presently configured with monit' not in result.msg
   retries: 5
   delay: 20
+  
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -83,7 +83,7 @@
 
 - name: Formplayer request-response logging directory
   file:
-    path: "{{ formplayer_service_log_dir }}"
+    path: "{{ formplayer_access_log_dir }}"
     state: directory
     owner: "{{ cchq_user }}"
     group: "{{ cchq_user }}"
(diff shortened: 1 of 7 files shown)


src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -160,3 +160,26 @@
   tags:
     - formplayer_deploy
   when: not _should_update_formplayer_in_place
+
```
gherceg, line 163: I'm not confident in placing this at the end of the file due to the comment above the last task `Link formplayer jar`, but figure this is unrelated to a deploy?

src/commcare_cloud/ansible/roles/formplayer/templates/monit.formplayer.j2
```diff
@@ -0,0 +1,5 @@
+ check file formplayer with path {{ formplayer_logging_dir }}/formplayer-spring.log
+  group formplayer
+  start program = "supervisorctl start {{ project }}-{{ deploy_env }}-formsplayer-spring"
```
gherceg, line 3: Made an assumption that the formplayer role has access to the `project` and `deploy_env` variables

src/commcare_cloud/ansible/roles/formplayer/templates/monit.formplayer.j2
```diff
@@ -0,0 +1,5 @@
...
+  stop program = "supervisorctl stop {{ project }}-{{ deploy_env }}-formsplayer-spring"
+  if content = "java.lang.OutOfMemoryError" then restart
```
calellowitz, line 5: we might want to make this only for the gc overhead one, not all OOMs, because I think it recovers from individual thread OOMs relative gracefully.

dannyroberts: Also, and this really is the most annoying thing for me to ask, but is there any user input that they can put here that makes this literal string show up in the log? If so we're sort of giving people a button they can press to restart a process on our server. I don't want to block the PR on this, but if there's a more complete way to match against a line such that a user can't make that line show up in the file at will, then it'd be worth doing that. For example if we know that the line will always start with a like a timestamp in brackets then a space, then the specific error with no text in between, then you can make a regular expression match it including a ^ start of line match that would make it much harder to do that way. Really annoying comment, I know

dannyroberts (approved):

dannyroberts (commented):
.

dannyroberts:
Looks like this is still much needed in the short term—just want to make sure there are clear next steps! If not feel free to ping me on slack

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -182,4 +182,5 @@
   register: result
   until: not result.failed or 'process not presently configured with monit' not in result.msg
   retries: 5
-  delay: 20
\ No newline at end of file
+  delay: 20
+
```
millerdev, line 186: Does github also not like an extra newline, or is there other whitespace on that line?

src/commcare_cloud/ansible/roles/formplayer/templates/monit.formplayer.j2
```diff
@@ -2,4 +2,4 @@
   group formplayer
   start program = "supervisorctl start {{ project }}-{{ deploy_env }}-formsplayer-spring"
   stop program = "supervisorctl stop {{ project }}-{{ deploy_env }}-formsplayer-spring"
-  if content = "java.lang.OutOfMemoryError" then restart
\ No newline at end of file
+  if content = "java.lang.OutOfMemoryError: GC overhead limit exceeded" then restart
```
millerdev, line 5: I think Danny was saying this pattern should be anchored to the beginning of the line with a regex that positively matches a date pattern or something that is part of the logging output. Otherwise a user may be able to type the string "java.lang.OutOfMemoryError: GC overhead limit exceeded" in an input box and submit to trigger a restart.

gherceg: Yeah sorry for the lack of updates on the PR. I offlined with Danny about that on Friday and that is still in progress. This initial change addressed Cal's feedback.

gherceg:
Upon approval I will test this on staging. Would appreciate if someone can address the unresolved comments I made when creating the PR. Thanks.

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -160,3 +160,26 @@
...
+- monit:
+    name: formplayer
+    state: monitored
+  tags: monit
+  ignore_errors: "{{ ansible_check_mode }}"
+  register: result
+  until: not result.failed or 'process not presently configured with monit' not in result.msg
+  retries: 5
+  delay: 20
```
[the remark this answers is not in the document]
gherceg: Configured my IDE to add a newline on save but things still look strange. Will address that now.

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
gherceg: Looking into this now. Thought my IDE was configured appropriately.

src/commcare_cloud/ansible/roles/formplayer/templates/monit.formplayer.j2
gherceg: Fortunately in the logs, there is a newline prior to output containing the OOM so added this regex:
`"^Caused by: java.lang.OutOfMemoryError: GC overhead limit exceeded"`

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
dannyroberts: Hey @gherceg your logic seems sound. I think you should update the comment to be more specific/correct and say something more like
```
# This should be the last task in the formplayer release process
# (and the last one tagged with `formplayer_deploy`)
...
```

src/commcare_cloud/ansible/roles/formplayer/templates/monit.formplayer.j2
dannyroberts: That assumption seems good, but you'll also be able to tell for sure when you run this (even in check mode)

src/commcare_cloud/ansible/roles/formplayer/templates/monit.formplayer.j2
millerdev: This made me think of another possible issue. I assume end users are unable to enter a string with newlines that end up in the logs? If not, that seems like a problem here. However, if that is an issue I propose that we deal with it in a follow-up PR rather than dragging this one out longer.

millerdev (approved):
I'm a little shaky on monit, so not confident in my ability to verify correctness of config files, so take my approval with a grain of salt.

dannyroberts (approved):
Looks good. If this works on staging I think we can start by rolling it out on just the machines that have had issues recently, and when it proves to be useful and not harmful, we roll it out to all machines

dannyroberts:
Just an update, @gherceg and I worked through some issues on this PR and tested it on staging, and while it does work, the ansible tasks don’t apply cleanly for some reason (the
```
monit:
  name: formplayer
  state: monitored
```
task says “process not presently configured with monit” when clearly the process is configured with monit) and so we'll have to come back to this tomorrow. But the good news is that making that line show up in the log file does now make formplayer restart on staging, so the proof of concept is there and this PR is very close.

knguyen142:
Kind of late to this, but what is the rationale for running monit to start the supervisor service instead of just monit? Is this what we already do with ES?

Have we confirmed the strong correlation between the OOM log entry and formplayer getitng in a bad state? If not, as part of the rollout, can we have a `dry-run` phase where instead of monit restart the formplayer supervisor service, it alerts us? I'm worried that we might be restarting too often.

gherceg:
> My guess is it's to keep conformity with our infrastructure since everything is on supervisor?

This is pretty much the reason for using monit as the tool to perform a check on the service, and supervisor to actually handle starting/stopping the service. There doesn't seem to be anything inherently wrong with this setup, though if migrating to monit is desirable for any reason that would fall outside the scope of this PR.

> Have we confirmed the strong correlation between the OOM log entry and formplayer getitng in a bad state? 

Monit will only trigger a restart if an OOM occurs due to garbage collection running out of memory which the process appears to be unable to recover from based on the correlation we have seen with formplayer machines being unreachable after these specific OOMs occur. However we still only plan on rolling this out on 1 formplayer machine to start. Specifically, to a machine that has been exhibiting issues with GC memory and already causes interruptions to users. We should be able to tell from there if this causes any unwanted restarts.

@knguyen142 hope that answers your questions.

gherceg:
Had to comment out the monit task that confirms monit is running on the machine until [this PR](https://github.com/ansible-collections/community.general/pull/1532) is approved/merged/released. This task isn't essential to getting this to work so we can do without it for now.

gherceg:
The most recent commit disables this for all machines by default. To avoid making this PR too much of a mess, I will do that on a separate PR and link it back here.

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -161,16 +161,20 @@
...
+    template:
+      src: "monit.formplayer.j2"
+      dest: "/etc/monit/conf.d/formplayer"
+      group: root
+      owner: root
+      mode: 0640
+    notify: reload monit
+    tags: monit
+
+  when: use_monit_for_formplayer == true
```
dannyroberts, line 177: I might be missing the point, but it looks to me like you could forego the `- block` and indentation, and just add the `when: ...` straight to the task.

dannyroberts (approved):
Looks good, with the one suggestion

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
gherceg: Updated 👍  thanks

dannyroberts (approved):

gherceg:
I fixed a bug in my recent changes [here](https://github.com/dimagi/commcare-cloud/pull/4412)

PR #4387, opened 2020-12-10
Remove commcare-hq src from formplayer (part 2)

dannyroberts:
Followup to https://github.com/dimagi/commcare-cloud/pull/4361. This is both just nice cleanup and something that we found since that PR was merged blocks us from running `update-config`. Once a formplayer deploy has gone out, current/localsettings.py will no longer exist and a task that tries to update it fails.

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/ansible/deploy_commcarehq.yml
+++ b/src/commcare_cloud/ansible/deploy_commcarehq.yml
@@ -73,14 +73,14 @@
   tags: services
 
 - name: Management Command Supervisor Config
-  hosts: commcarehq:formplayer
+  hosts: commcarehq
   tasks:
     - include_vars: roles/commcarehq/vars/main.yml
     - include_tasks: roles/commcarehq/tasks/management_commands.yml
   tags: services
 
 - name: Prometheus Supervisor Config
-  hosts: commcarehq:formplayer
+  hosts: commcarehq
   tasks:
     - include_vars: roles/commcarehq/vars/main.yml
     - include_tasks: roles/commcarehq/tasks/prometheus.yml
@@ -88,7 +88,7 @@
   tags: services
 
 - name: Prometheus Django runner
-  hosts: commcarehq:formplayer
+  hosts: commcarehq
   tasks:
     - include_vars: roles/commcarehq/vars/main.yml
     - include_tasks: roles/commcarehq/tasks/prometheus_django_runner.yml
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/deploy_commcarehq.yml
```diff
@@ -73,22 +73,22 @@
   tags: services
 
 - name: Management Command Supervisor Config
-  hosts: commcarehq
+  hosts: commcarehq:formplayer
```
snopoke, line 76: If the HQ code isn't on formplayer machines then we can't run management command from them either.

```suggestion
  hosts: commcarehq
```

src/commcare_cloud/ansible/deploy_commcarehq.yml
```diff
@@ -73,22 +73,22 @@
...
   tasks:
     - include_vars: roles/commcarehq/vars/main.yml
     - include_tasks: roles/commcarehq/tasks/management_commands.yml
   tags: services
 
 - name: Prometheus Supervisor Config
-  hosts: commcarehq
+  hosts: commcarehq:formplayer
```
snopoke, line 83: This is only needed if there are HQ processes running on the machine.
```suggestion
  hosts: commcarehq
```

src/commcare_cloud/ansible/deploy_commcarehq.yml
```diff
@@ -73,22 +73,22 @@
...
   tasks:
     - include_vars: roles/commcarehq/vars/main.yml
     - include_tasks: roles/commcarehq/tasks/prometheus.yml
       when: prometheus_monitoring_enabled|default(False)
   tags: services
 
 - name: Prometheus Django runner
-  hosts: commcarehq
+  hosts: commcarehq:formplayer
```
snopoke, line 91: Same as above:
```suggestion
  hosts: commcarehq
```

snopoke (changes requested):

src/commcare_cloud/ansible/deploy_commcarehq.yml
millerdev: What management commands do we need to run on a formplayer machine?

snopoke: @millerdev none, hence why I suggested removing it.

millerdev: I'm sorry, I misinterpreted your comment. I thought you were saying we need to keep the commcarehq source on these machines so we can run management commands on them.

dannyroberts (commented):
Thanks @snopoke for your careful review there. That makes this PR even simpler!

millerdev (approved):

PR #4392, opened 2020-12-14
bump memory on staging control to get rid of OOMs when updating users

biyeun:
##### ENVIRONMENTS AFFECTED
staging


mjriley:
Approving, although I'm concerned that this process requires enough memory to oom. I think we should investigate this in the future.

PR #4405, opened 2020-12-18
Enforce python 3 usage

gherceg:
As part of the Python 3 migration on commcare-cloud, this change ensures users are aware they need to upgrade to Python 3. If they must use Python 2 they can still to do so, but this makes it slightly more difficult (must pass in `--force-commcare-cloud-to-use-python2`). We will merge this once we feel Python 3 is ready for use on all environments. 
##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/src/commcare_cloud/commcare_cloud.py
+++ b/src/commcare_cloud/commcare_cloud.py
@@ -171,12 +171,23 @@ def make_command_parser(available_envs, formatter_class=RawTextHelpFormatter,
 
 
 def call_commcare_cloud(input_argv=sys.argv):
+    # make sure user aware they are running on python 2 env
+    force_python_2 = "--please-let-me-use-python2"
+    if sys.version_info[0] == 2:
+        if force_python_2 not in input_argv:
+            print('Error: you must upgrade to Python 3. If you really have to, you can use Python 2 with this option'
+                  ' {}'.format(force_python_2))
+            exit(-1)
+
     put_virtualenv_bin_on_the_path()
     parser, subparsers, commands = make_command_parser(available_envs=get_available_envs())
     args, unknown_args = parser.parse_known_args(input_argv[1:])
 
     if args.control:
         run_on_control_instead(args, input_argv)
+
+    # remove force_python_2 arg before running command
+    unknown_args.remove(force_python_2)
     try:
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commcare_cloud.py
```diff
@@ -171,12 +171,25 @@ def make_command_parser(available_envs, formatter_class=RawTextHelpFormatter,
 
 
 def call_commcare_cloud(input_argv=sys.argv):
+    # make sure user aware they are running on python 2 env
+    force_python_2 = "--force-commcare-cloud-to-use-python2"
+    if not os.environ.get("TRAVIS_TEST") and sys.version_info[0] == 2:
+        if force_python_2 not in input_argv:
+            print('Error: you must upgrade to Python 3. Though not desirable, if you really have to '
+                  'you can use Python 2 with this option {}'.format(force_python_2))
```
millerdev, line 179: We should provide instructions for creating a Python 3 virtualenv in this error message. That will probably be more effective than simply providing an option to continue with Python 2.
```suggestion
            from textwrap import dedent
            print(dedent(f"""
                Error: you must upgrade to Python 3. Though not desirable, if you really have
                to you can use Python 2 with this option {force_python_2}

                Setup Python 3
                - Create a new Python-3-based virtualenv
                - pip install -r requirements3.txt
                - rm -rf src/commcare_cloud.egg-info
                - pip install -e .
                - manage-commcare-cloud install
            """))
```

millerdev (approved):
Looks great!

src/commcare_cloud/commcare_cloud.py
gherceg: Made the change but stuck with `"{}".format()` since this needs to work on Python 2

dannyroberts (approved):

gherceg:
@dimagi/team-commcare-hq We are moving to Python 3 on Commcare-Cloud. 

Changes in this PR will not prevent you from using Python 2, but will encourage you to move to Python 3. Instructions to upgrade can be found [here](https://docs.google.com/document/d/17QQu5BAAwle5KHTMawTlsf4o78ZRXElIKlrBdHjrU8M/edit#).

PR #4407, opened 2020-12-21
Fix terraform for configs without elasticache_cluster

dannyroberts:
A change made for https://dimagi-dev.atlassian.net/browse/SAAS-11511 wasn't backwards compatible, and this PR makes it backwards compatible.

##### ENVIRONMENTS AFFECTED
production, india (staging had a different config so wasn't hit by it)


dannyroberts:
@rameshganne @sanjay2916 FYI. The existing configuration was failing on production because elasticache_cluster `{{ elasticache_cluster.node_groups|tojson }}` errored when `elasticache_cluster` was none (the default).

gherceg (approved):

PR #4408, opened 2020-12-22
Formplayer deployment fixes

mjriley:
Following the process outlined [here](https://confluence.dimagi.com/pages/viewpage.action?spaceKey=internal&title=How+to+create+a+new+machine+in+an+AWS+environment) for a formplayer machine failed on 2 steps:
- an inability to find /dev/sd[x], because that device did not exist
- accessing a directory in formplayer which did not exist

The first seems to have been caused by the common mounting operations overriding the work of deploy_ebsnvme_mapping. The error encountered:
`
TASK [ecryptfs : Format block storage volume] *******************************************************************************************************************************
fatal: [10.201.10.118]: FAILED! => {“changed”: false, “msg”: “Device /dev/sdf not found.“}
`

The second is caused by referencing a folder that was expected to be created by another unrelated task.
The error encountered:
`
TASK [commcarehq : define formplayer spring services] **************************
failed: [10.201.10.238] (item={u'env_vars': {u'no_proxy': u'', u'https_proxy': u'', u'http_proxy': u'', u'TMPDIR': u'/opt/tmp'}}) => {"ansible_loop_var": "item", "changed": false, "checksum": "27597806eb5b832790fcf71a8d6d2e473cec1324", "item": {"env_vars": {"TMPDIR": "/opt/tmp", "http_proxy": "", "https_proxy": "", "no_proxy": ""}}, "msg": "Destination directory /home/cchq/www/staging/services does not exist"}
`

### DIFF
--- a/src/commcare_cloud/ansible/deploy_stack.yml
+++ b/src/commcare_cloud/ansible/deploy_stack.yml
@@ -2,8 +2,8 @@
 # deploy-stack.yml
 
 - import_playbook: update_apt_cache.yml
-- import_playbook: deploy_ebsnvme_mapping.yml
 - import_playbook: deploy_common.yml
+- import_playbook: deploy_ebsnvme_mapping.yml
 - import_playbook: deploy_lvm.yml
 - import_playbook: deploy_db.yml
 - import_playbook: deploy_commcarehq.yml
--- a/src/commcare_cloud/ansible/roles/commcarehq/tasks/formplayer.yml
+++ b/src/commcare_cloud/ansible/roles/commcarehq/tasks/formplayer.yml
@@ -9,6 +9,17 @@
   tags:
     - formplayer_deploy
 
+- name: create services home
+  become: yes
+  file:
+    path: "{{ service_home }}"
+    owner: "{{ cchq_user }}"
+    group: "{{ cchq_user }}"
+    mode: 0777
+    state: directory
+  tags:
+    - formplayer_deploy
+
 - name: define formplayer spring services
   become: yes
   template:


knguyen142:
These changes make sense to me, but I'm not too familiar with these ansible roles so will leave for @dannyroberts. Is it possible @dannyroberts, @mjriley and I forgot to run something to run into these issues? Because we regularly create new formplayer instances right?

Also, @mjriley it might be helpful if you could include the errors we ran into which correspond to each fix.

dannyroberts:
> The first seems to have been caused by the common mounting operations overriding the work of deploy_ebsnvme_mapping

Huh. That's fascinating. Any idea of what task in particular under deploy_common is undoing the ebsnvme mapping? If we could fix that instead of changing the order then that'd be better, but if there are no clues there I don't need to hold up this PR on that

dannyroberts (approved):

dannyroberts:
I also wanted to add some context but it's just FYI and doesn't really affect this PR. The device mapping issue you fixed here affects any machine that uses a separate data volume from the boot volume—that shows up in terraform.yml base `block_device` and in `inventory.ini` as `datavol_device` that are autopopulated by the templating code that works off of `inventory.ini.j2`. I'm actually not sure why on staging the formplayer machines are set up that way, probably to test something a while ago. Regardless, in production, while formplayer machines don't use that setup, elasticsearch and couchdb2 machines do, so this fix is valuable.

(The other fix about the services directory is definitely formplayer-specific, and caused by our recent recategorization for formplayer as "not commcarehq" so that we wouldn't be loading the commcare-hq source there unnecessarily.)

mjriley:
> Any idea of what task in particular under deploy_common is undoing the ebsnvme mapping? 

Not at the moment. I want to get this fix out to avoid other devs running into the same issues, but I'm making a note to investigate this later.

PR #4409, opened 2020-12-22
Prioritize BlockRules in AWS WAF frontend ACL

skodde:
The DimagiBlockRules rule in the AWS WAF frontend-waf ACL should come first
in order to block any IP address manually added to the related IP sets
regardless of paths explicitly permitted in other rules.

Fix: https://dimagi-dev.atlassian.net/browse/SAAS-11562

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf
+++ b/src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf
@@ -188,6 +188,21 @@ resource "aws_wafv2_web_acl" "front_end" {
 
   rule {
     priority = "0"
+    name = "DimagiBlockRules"
+    override_action { none {} }
+    statement {
+      rule_group_reference_statement {
+        arn = "${aws_wafv2_rule_group.dimagi_block_rules.arn}"
+      }
+    }
+    visibility_config {
+      cloudwatch_metrics_enabled = true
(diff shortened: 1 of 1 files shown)


mjriley (commented):
Changes make sense, but would it be better to use priorities with gaps in between, to allow similar changes to avoid modifying every existing priority? [AWS Docs](https://docs.aws.amazon.com/waf/latest/APIReference/API_Rule.html) mention: `The priorities don't need to be consecutive, but they must all be different`.
Going with 100, 200, etc would leave us plenty of room for additions.

dannyroberts (approved):
Looks good!

dannyroberts:
@mjriley that's an interesting idea that I hadn't thought of before. What I like about it is what you said, that it means diffs of PRs that just reorder can have a smaller diff that doesn't include renumbering everything. What I don't like about it is that there's then arbitrariness to those numbers that would cry out to be renormalized back to 100, 200, 300, etc which then defeats the purpose. If there were a simple way to pull these out into a list and then insert the numbering in automatically, that to me would be maybe the nicest because it takes the arbitrary numbering (which is conceptually redundant with the order items appear) out of our hands. I'll mull it over and especially if it's a frequent annoyance we can probably do something about it later

PR #4414, opened 2020-12-24
Include monit handlers to have access to 'reload monit'

gherceg:
Properly include the monit reload handler to give the formplayer playbook access to it/the ability to run it. Tested by making no-op changes on c023 and verified that monit reloaded as expected.
##### ENVIRONMENTS AFFECTED

None


dannyroberts (approved):
Oh nice find. Yeah this seems better than the way we talked about

PR #4432, opened 2021-01-06
internal domain: updating resolv.conf symbolic link for OS-distribution Ubuntu

rameshganne:
Here, I am updating the deploy_common playbook to update the symbolic link of /etc/resolv.conf file. 

Updated file: deploy_common.yml

**/run/systemd/resolve/resolv.conf** - this file contains the nameserver & domain search details. This update does not impact the public-domain resolutions. It can also resolve Route-53 private hosted-zones.

These changes can update the symbolic link for all environments { **Staging, Prod & India** }

### DIFF
--- a/src/commcare_cloud/ansible/deploy_common.yml
+++ b/src/commcare_cloud/ansible/deploy_common.yml
@@ -83,17 +83,6 @@
   hosts: all_commcarehq
   tasks:
 
-  - name: set running configuration
-    become: yes
-    lineinfile:
-      dest: /etc/resolv.conf
-      line: 'nameserver {{ item }}'
-    with_items: "{{ nameservers }}"
-    when: nameservers is defined
-    tags:
-      - dns
-      - after-reboot
-  
   - name: Checks existing resolv.conf
     stat:
       path: /run/systemd/resolve/resolv.conf
@@ -119,6 +108,17 @@
       dest: /etc/resolvconf/resolv.conf.d/base
       line: 'nameserver {{ item }}'
       create: yes
+    with_items: "{{ nameservers }}"
+    when: nameservers is defined
+    tags:
+      - dns
+      - after-reboot
+  
+  - name: set running configuration
+    become: yes
+    lineinfile:
+      dest: /etc/resolv.conf
+      line: 'nameserver {{ item }}'
     with_items: "{{ nameservers }}"
     when: nameservers is defined
     tags:
--- a/src/commcare_cloud/ansible/deploy_common.yml
+++ b/src/commcare_cloud/ansible/deploy_common.yml
@@ -93,7 +93,26 @@
     tags:
       - dns
       - after-reboot
+  
+  - name: Checks existing resolv.conf
+    stat:
+      path: /run/systemd/resolve/resolv.conf
+    register: resolvconf
+    tags:
+      - dns
+      - after-reboot
 
+  - name: Updating resolv.conf symbolic link 
+    when: ansible_distribution == "Ubuntu" and resolvconf.stat.exists
+    become: yes
+    file:
+      src: /run/systemd/resolve/resolv.conf
+      dest: /etc/resolv.conf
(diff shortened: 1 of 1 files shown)


shyamkumarlchauhan (approved):

src/commcare_cloud/ansible/deploy_common.yml
```diff
@@ -93,7 +93,26 @@
...
+  - name: Updating resolv.conf symbolic link 
+    when: ansible_distribution == "Ubuntu" and resolvconf.stat.exists
+    become: yes
+    file:
+      src: /run/systemd/resolve/resolv.conf
+      dest: /etc/resolv.conf
+      state: link
+    tags:
+      - dns
+      - after-reboot
```
dannyroberts, line 114: In the task above this, we make sure that the `nameservers` appear in `/etc/resolv.conf`, but then we overwrite /etc/resolve.conf in this task with a link to `/run/systemd/resolve/resolv.conf`. Where does `/run/systemd/resolve/resolv.conf` come from? Does the Ubuntu distribution for AWS include aws nameservers that will resolve private dns entries we specify in route53?

rameshganne: In this case, we are changing  the symbolic link of /etc/resolv.conf. Presently, This file is mapping to **_/run/systemd/resolve/stub-resolv.conf_** which is containing the 127.0.0.53 DNS stub as the only DNS server [nameserver] and maintained by _**systemd-resolved**_ system service. This stub DNS-nameserver is unable to resolve AWS Route-53 Private hosted-zone records.

To resolve the private hosted-zones also, we are changing the symbolic link to **_/run/systemd/resolve/resolv.conf_** which contains information about all known DNS nameservers [ AmazonProvidedDNS ] and maintained by _**systemd-resolved**_ system service. This known DNS nameserver is able to resolve AWS Route-53 Private hosted-zone records also.

dannyroberts: OK thanks that makes sense. the task above this edits /etc/resolv.conf to include whatever we put in the `nameservers` variable. I think on AWS environments we don't need to specify custom `nameservers` (india does right now, but you can remove that; I'm pretty sure it's not necessary anymore). But third party environments (this is open source software and we encourage third parties running commcare to use it as well) may depend on `nameservers` to use appropriate DNS nameservers for their area or to override whatever their ISP's default is. Can you think through how this would work in that case? (I think as it's written we'd add nameservers to `/run/systemd/resolve/stub-resolv.conf` and then use `/run/systemd/resolve/resolv.conf`, which defeats the purpose of adding nameservers in the first place, right?

dannyroberts (approved):

dannyroberts:
Let me know when it's ready to merge, once it's been tested on staging and india

dannyroberts:
@rameshganne @shyamkumarlchauhan ready to merge now?

rameshganne:
Yes @dannyroberts @shyamkumarlchauhan  . I had executed it on India & Staging environments.

snopoke:
@rameshganne @dannyroberts @shyamkumarlchauhan this broke the `after-reboot` command for some people (https://forum.dimagi.com/t/update-config-failing/8171/9). Can you advise what the resolution is if `/etc/resolve.conf` already exists?

src/commcare_cloud/ansible/deploy_common.yml
joxl: > Presently, This file is mapping to **/run/systemd/resolve/stub-resolv.conf**

Is this the system "default state", or is this a modified system state (i.e. overridden by some custom configuration)?

> we are changing the symbolic link to **/run/systemd/resolve/resolv.conf**

Why? What populates this file, and why are its contents different from `/run/systemd/resolve/stub-resolv.conf`?

Is there `systemd-resolved.service` documentation that outlines why/how to do this sort of thing?

PR #4433, opened 2021-01-06
Allow environment-specific requirements to be specified

biyeun:
https://dimagi-dev.atlassian.net/browse/SAAS-11606

##### ENVIRONMENTS AFFECTED
staging, release script / environment meta

This is my proposal for a solution to specify environment-specific requirements (namely the requirements for SSO) which require additional `apt` dependencies and is not necessary to be installed on all environments (SSO was made an optional app, see https://github.com/dimagi/commcare-hq/pull/28938)


millerdev (approved):
Fixed failing test by following procedure outlined [here](https://docs.google.com/document/d/17QQu5BAAwle5KHTMawTlsf4o78ZRXElIKlrBdHjrU8M/edit#bookmark=id.gv9c89h23noz).

biyeun:
@millerdev after chatting with @dannyroberts today, we came to the conclusion that it will be a bad idea from a maintenance and stability perspective to make `sso` an optional app, as it will be tightly woven into other parts of the codebase touching authorization as I work further on this feature, even though it's fairly separate now in its early stages.

It would be risky to have only tests with the `sso` app enabled and no test runners without it (as the default state of HQ) and not catch import errors and related errors that could occur. Additionally, it would add a great deal of complexity when actually integrating sso with the rest of hq.

Lastly, a separate requirements file might result in certain dependency versions being out of sync with those in requirements or prod-requirements, and that's why it's better to have a single requirements file per environment.

Basically, we decided it would be unwise to separate the sso requirements and app from the rest of hq and we will just bite the bullet of initially requiring a `deploy-stack` command to be run on all our environments (with a changelog announcement) and having the apt-get requirements as part of the requirements fr running HQ. We should focus on getting this PR merged in instead https://github.com/dimagi/commcare-cloud/pull/4426 (I need to add the changelog announcement here)

PR #4437, opened 2021-01-07
Updated proxy template files to support python 3

mjriley:
Some of the code within jinja templates referred to unsupported code from python 2. This specific PR just handled those templates still using `.iteritems()` rather than `.items()`.


gherceg (approved):
Thanks @mjriley!

PR #4440, opened 2021-01-08
Adding new/larger machine to replace pgbouncer5

millerdev:
##### ENVIRONMENTS AFFECTED
production


calellowitz (approved):

millerdev:
Had to use `--force-commcare-cloud-to-use-python2` to update local known hosts because of a bug with Python 3. That bug should be fixed in https://github.com/dimagi/commcare-cloud/pull/4441

PR #4442, opened 2021-01-08
Default to cchq (python 3 venv) in init script

gherceg:
When users ssh into a machine, the init script still defaults to python 2. This will use the `cchq` venv running python 3, and install it if it does not exist.
##### ENVIRONMENTS AFFECTED

None


millerdev (approved):

dannyroberts (approved):
awesome thanks

PR #4444, opened 2021-01-11
add config for second pgbouncer

calellowitz:
This is obviously not the ideal way to accomplish the change, but this is POC for how the config files would need to look. Ideally it would be configured in some sort of loop where we specify a number of processes per machine, but in an effort to have this ready in case we need to roll it out in an emergency I have not done so.

We should not merge this as is, but can use it for a deploy tonight to a single machine if necessary. If we decide this is a route we want to go down, we can clean up the code before merge.

##### ENVIRONMENTS AFFECTED

N/A

### DIFF
--- a/src/commcare_cloud/ansible/roles/pgbouncer/handlers/main.yml
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/handlers/main.yml
@@ -17,7 +17,7 @@
 
 - name: Restart pgbouncer2
   become: yes
-  action: service name=pgbouncer state=restarted
+  action: service name=pgbouncer2 state=restarted
 
 - name: Reload pgbouncer2
   become: yes
--- a/src/commcare_cloud/ansible/roles/pgbouncer/handlers/main.yml
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/handlers/main.yml
@@ -10,3 +10,15 @@
 - name: Reload pgbouncer
   become: yes
   action: service name=pgbouncer state=reloaded
+
+- name: Start pgbouncer2
+  become: yes
+  action: service name=pgbouncer2 state=started
+
+- name: Restart pgbouncer2
+  become: yes
+  action: service name=pgbouncer state=restarted
+
+- name: Reload pgbouncer2
+  become: yes
+  action: service name=pgbouncer2 state=reloaded
--- a/src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer2.ini.j2
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer2.ini.j2
@@ -20,8 +20,8 @@
 ;;; Administrative settings
 ;;;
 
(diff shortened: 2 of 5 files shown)


src/commcare_cloud/ansible/roles/pgbouncer/handlers/main.yml
```diff
@@ -10,3 +10,15 @@
...
   become: yes
   action: service name=pgbouncer state=reloaded
+
+- name: Start pgbouncer2
+  become: yes
+  action: service name=pgbouncer2 state=started
+
+- name: Restart pgbouncer2
+  become: yes
+  action: service name=pgbouncer state=restarted
```
dannyroberts, line 20: Should this say `Restart pgbouncer`? Or name=pgbouncer2? Or is there something else I'm missing

src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer2.ini.j2
```diff
@@ -0,0 +1,221 @@
+;;; {{ansible_managed}}
```
dannyroberts, line 1: how is this file different from `pgbouncer.ini.j2`? If it's not can they just use the same template?

src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer2.init.j2
```diff
@@ -0,0 +1,100 @@
+#!/bin/sh
```
dannyroberts, line 1: how is this file different from `pgbouncer.init.j2`? If it's not can they just use the same template?

src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer2.ini.j2
calellowitz: `logfile`, `pidfile`, `unix_socket_dir` as per `so_reuseport` instructions in https://www.pgbouncer.org/config.html

src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer2.init.j2
calellowitz: in this one the name and binary are different (it still calls `/usr/sbin/pgbouncer` but needs to use pgbouncer2 as the name for the config file and other variables). It is possible these could be combined to be shared if the binary variable `BIN` were hardcoded and the the `NAME` were passed in as a variable

src/commcare_cloud/ansible/roles/pgbouncer/handlers/main.yml
calellowitz: name should be `pgbouncer2` good catch

calellowitz:
Closed in favor of https://github.com/dimagi/commcare-cloud/pull/4454

PR #4446, opened 2021-01-11
Add changelog for python3 support

gherceg:
We should have added a changelog for this change earlier. Informs users that Python 3 is supported and preferred, along with instructions for upgrading, and how to still use Python 2 if necessary.
##### ENVIRONMENTS AFFECTED

None


millerdev (approved):
Looks like the build is failing with
```
    If you're getting a failure here with a change to the changelog docs
    make sure you're following the directions in
    https://github.com/dimagi/commcare-cloud/blob/master/changelog/README.md
    and not editing changelog markdown files directly.
```

millerdev (approved):

PR #4447, opened 2021-01-12
Support Network Load Balancer over multiple pgbouncer instances

dannyroberts:
This has been tested on staging and appears to work quite well, and seems to allow for traffic to switch between endpoints without a formplayer restart.

This includes
- the ansible changes required to specify a pgbouncer_endpoint that's separate from where pgbouncer is installed, and also specify multiple pgbouncers as for a given db. (The Network Load Balancer is the pgbouncer_endpoint and the hosts it points to are the `pgbouncer_hosts`.)
- All the terraform setup for creating the NLB and routing traffic from it to the target machines
- All the glue code that lets you reference this endpoint as a fake host group in inventory, similar to how we handle RDS instances. (They're endpoints in that you can make requests to them, but not hosts in that you can ssh into them and do stuff.)
- Changes that exemplify usage on staging, including adding a new pgbouncer machine, pgbouncer0-staging and a load balancer over it and pgproxy2-staging.

##### ENVIRONMENTS AFFECTED
staging


millerdev (approved):
Looks fine to me. Not sure I understand everything that's going on here.

shyamkumarlchauhan (approved):

PR #4288, opened 2020-10-21
adding new parameter dedicated_migration_process to run migrations on process#0

prem-fissionhq:
Added new variable dedicated_migration_process to switch run_migrations process_num on #0  (as example: added for SqlSMSPillow and xform-pillow). 

Added condition to add variable only when dedicated_migration_process is defined. 

please review and let me know your suggestions. Thanks.

PR #4449, opened 2021-01-12
adding new parameter dedicated_migration_process to run migrations on process#0

prem-fissionhq:
Added new variable dedicated_migration_process to pillows to run migrations and transactions on different process_num's. #0 (as an example: added for case-pillow and xform-pillow).

Added condition to add variable only when dedicated_migration_process is defined.

please review and let me know your suggestions. Thanks.

Old PR: https://github.com/dimagi/commcare-cloud/pull/4288
HQ PR: https://github.com/dimagi/commcare-hq/pull/28854

ENVIRONMENTS AFFECTED

staging
production.


AmitPhulera (approved):

shyamkumarlchauhan (approved):

dannyroberts (approved):

shyamkumarlchauhan (approved):

prem-fissionhq:
the change has been rollbacked as we have seen that "changes to process by kafka partition " metrics is consistently increasing after rollout of this change.
It was happening only with the partitions of case-pillow 15,31 and xform-pillow 15, 31. Whereas, the process was correctly getting split between 0 and other processes.  

image

Looking into this we have set the dedicated_migration_process parameter to False.

PR #4454, opened 2021-01-13
Multiple PgBouncer processes on a single machine

skodde:
PgBouncer by default runs only one single-threaded process.
We implement an initial set of changes to allow us to run multiple
PgBouncer processes on a single machine, which share the same TCP
socket by using SO_REUSEPORT via the so_reuseport configuration
option.
The number of processes can be defined as a variable in the inventory
for the needed hosts, and defaults to 1.

https://dimagi-dev.atlassian.net/browse/SAAS-11714

To roll out, run:
```
cchq <env> ansible-playbook --branch=mp/pgbouncer-multiprocess deploy_db.yml --tags=pgbouncer --limit=<pgbouncer-machine>
```

##### ENVIRONMENTS AFFECTED
All


millerdev (commented):
I assume after this change connection limits will be multiplied by the number of pgbouncer processes running on the machine, is that right? So for example, if we previously had a single pgbouncer with max connections set to 500, a new machine with two pgbouncers, each having max connections of 500, will effectively result in a max connections of 1000.

Will we be adjusting the max connections per pgbouncer process so the total max connections comes out to what it was before (per postgres database), or are we increasing the maximum as in the example above? If the latter, do we know what effect that will have on postgres?

src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
```diff
@@ -1,8 +1,10 @@
 pgbouncer_port: 6432
-pgbouncer_pid_file: "/var/run/postgresql/pgbouncer.pid"
+pgbouncer_pid_file: "/var/run/postgresql/pgbouncer-{{ item }}.pid"
```
calellowitz, line 2: Is `{{ item }}` defined when this file is evaluated? I thought that was local to the playbook being executed, and these are constant default values. I did not realize you could use these as templates to be passed in.

environments/staging/inventory.ini.j2
```diff
@@ -29,7 +29,7 @@ proxy
 web7
 web8
 
-{{ __pgproxy2__ }}
+{{ __pgproxy2__ }} pgbouncer_processes=2
```
calellowitz, line 32: I think there is a command you need to run to make this appear in the actual inventory, that compiles the file from this template.

src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
```diff
@@ -36,7 +36,10 @@ listen_port = {{ pgbouncer_port }}
 ;unix_socket_dir = /tmp
 ;unix_socket_mode = 0777
 ;unix_socket_group =
-unix_socket_dir = /var/run/postgresql
+unix_socket_dir = {{ pgbouncer_socket_dir }}
```
calellowitz, line 39: In my testing, I had to create this directory, does this playbook work if that directory does not already exist? For example if we set this to 3 processes?

src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
```diff
@@ -44,11 +35,31 @@
...
 
-- name: pgbouncer init script
-  template: src=pgbouncer.init.j2 dest=/etc/init.d/pgbouncer
+- name: Clean up possible previous instances
+  shell: |
+    systemctl stop pgbouncer-multiprocess@*
+    systemctl disable pgbouncer-multiprocess@*
+    rm -rf {{ pgbouncer_config_home }}/pgbouncer-*.ini
+    rm -rf {{ pgbouncer_config_home }}/userlist-*.txt
+    rm -rf /var/run/postgresql/pgbouncer-*
```
calellowitz, line 44: This looks like it will mean you cannot run this playbook while pgbouncer is running (since it will blow away the current config and socket directory). What happens if we want to increase the number of processes on a machine from 1 to 2 rather than start from scratch?

calellowitz (commented):
General idea look great. Left a few questions, and have the same question as Daniel about changing connection limits after this.

environments/staging/inventory.ini.j2
skodde: Yes, that's:

`cchq <env> aws-fill-inventory [--cached]`

src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
skodde: Yes, it's gonna be defined when used by the task.

src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
skodde: It does not at the moment, but we're also not using unix sockets at all (right?).
We can decide if we want to expand this part and deal with the directories, or just deactivate it explicitly and drop it.

src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
skodde: Right. I mean, you can, but of course it would be a hard transition. I'm only leaving the logs, because they could be of value.
This part can definitely be improved/expanded, let's define some requirements and I'll work on a better solution.

skodde:
> I assume after this change connection limits will be multiplied by the number of pgbouncer processes running on the machine, is that right? So for example, if we previously had a single pgbouncer with max connections set to 500, a new machine with two pgbouncers, each having max connections of 500, will effectively result in a max connections of 1000.
> 
> Will we be adjusting the max connections per pgbouncer process so the total max connections comes out to what it was before (per postgres database), or are we increasing the maximum as in the example above? If the latter, do we know what effect that will have on postgres?

We're gonna adjust the pgbouncer pool size based on how many processes are running so that the total will be the same as before.

src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
snopoke: @calellowitz ansible does lazy evaluation of template strings so as long as the var is defined when it's needed it should work

src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
```diff
@@ -1,8 +1,10 @@
...
 pgbouncer_config_home: /etc/pgbouncer
-pgbouncer_ini: "{{ pgbouncer_config_home }}/pgbouncer.ini"
-pgbouncer_users: "{{ pgbouncer_config_home }}/userlist.txt"
+pgbouncer_ini: "{{ pgbouncer_config_home }}/pgbouncer-{{ item }}.ini"
+pgbouncer_users: "{{ pgbouncer_config_home }}/userlist-{{ item }}.txt"
```
snopoke, line 5: is it necessary to have different files for each process?

src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
snopoke: Is the goal here to ensure that we don't have dangling config if we reduce the number of processes? If so we could delete all config from `pgbouncer_processes` up to some max number like 32 or 64? Or else we could use the same approach we use for supervisor configs: https://github.com/dimagi/commcare-cloud/blob/16f055deef97ca4795bd2fb9ee3cc08d0e7fedcb/src/commcare_cloud/ansible/deploy_commcarehq.yml#L97-L128

environments/staging/inventory.ini.j2
skodde: Synced in last commit.

src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
skodde: Turns out we may need them for some features, I've added the directory creation in the last commit.

src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
skodde: For the config file, yes, because we need to specify different log files, pid files, and socket directories for each process.
For the userlist, ah! Maybe not?

skodde: Fixed in the last commit, we now use a single userlist file.

src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
skodde: Yes, but also to be able to scale up or down without restarting everything from scratch every time. The new implementation should be much more clean and effective, let me know if you think we could still move some parts of it elsewhere.

src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
```diff
@@ -1,39 +1,55 @@
...
   tags:
     - pgbouncer
-    - configure
 
-- name: pgbouncer users
-  template: src=pgbouncer.users.j2 dest="{{ pgbouncer_users }}"
-  notify: Reload pgbouncer
+- name: pgbouncer quit if version is too old
+  fail:
+    msg: "pgbouncer version currently installed is too old"
```
snopoke, line 16: Add more context to error message: 
```suggestion
    msg: "pgbouncer version currently installed is too old: {{ ansible_facts.packages['pgbouncer'][0].version }} < {{ pgbouncer_min_version }}"  
```

src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
```diff
@@ -44,18 +60,90 @@
...
+    - configure
+
+- name: pgbouncer stop and disable excess processes
+  become: yes
+  systemd:
+    name: "pgbouncer-multiprocess@{{ item }}"
+    state: stopped
+    enabled: no
+  with_sequence: start="{{ pgbouncer_processes + 1 }}" end="{{ pgbouncer_current_processes | max }}"
+  when: pgbouncer_current_processes | max > pgbouncer_processes
```
snopoke, line 80: =this task and the next 2 could be put into a block with a single when which helps to show that they are all related. Alternately the could be extracted to a separate file and only included if necessary.

```
block:
  - name: pgbouncer stop and disable excess processes
  - name: pgbouncer remove conf for excess processes
  - name: pgbouncer remove unix socket directory for excess processes
when: pgbouncer_current_processes | max > pgbouncer_processes
```

src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
skodde: Added a second line.

src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
skodde: Added them in a block. There's only 3 so unless we need to call them from elsewhere I think it's ok to keep them in the main file.

dannyroberts (approved):

snopoke:
@dannyroberts @skodde It doesn't look like there was a change log or announcement for this. Can you please send one out. Some users' are having issues with old services: https://forum.dimagi.com/t/update-config-failing/8171/13

dannyroberts:
Thanks for raising, @snopoke.
https://dimagi-dev.atlassian.net/browse/SAAS-12182 to track

PR #4458, opened 2021-01-13
Fix Clean Releases

biyeun:
We had a P1 triggered on the date this PR was created which resulted in all but the current release folder being deleted. As a result, when a rollback was attempted there was no other release to roll back to

Two key points:
- added a safeguard check to make sure that if less than three valid releases are found, we do not delete previous releases ever
- when `use_sudo=True` on `files.contains`, we should always use `shell=True`

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -415,7 +415,7 @@ def clean_releases(keep=3):
         for index, release in enumerate(reversed(releases)):
             if release == current_release or release == os.path.basename(env.code_root):
                 valid_releases += 1
-            elif files.contains(RELEASE_RECORD, release, use_sudo=True):
+            elif files.contains(RELEASE_RECORD, release):
                 valid_releases += 1
                 if valid_releases > keep:
                     to_remove.append(release)
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -415,7 +415,7 @@ def clean_releases(keep=3):
         for index, release in enumerate(reversed(releases)):
             if release == current_release or release == os.path.basename(env.code_root):
                 valid_releases += 1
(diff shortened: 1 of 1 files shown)


knguyen142 (approved):

src/commcare_cloud/fab/operations/release.py
```diff
@@ -415,7 +415,7 @@ def clean_releases(keep=3):
         for index, release in enumerate(reversed(releases)):
             if release == current_release or release == os.path.basename(env.code_root):
                 valid_releases += 1
-            elif files.contains(RELEASE_RECORD, release, use_sudo=True):
+            elif files.contains(RELEASE_RECORD, release, use_sudo=True, shell=True):
```
dannyroberts, line 418: It sounds like you found something subtle and non-obvious. Can you elaborate here? Depending, perhaps a comment in the code would also be helpful

calellowitz: It was mostly trial and error. We determined this was always returning `False`, and so we looked into what arguments we could change to make it return correctly. Removing `use_sudo` had the same effect but it appears Simon added that for a reason. When comparing this command that failed, and the `exists` command which worked, we noticed this one passed `shell=False` by default and the other did not. The docs pointed us to [this description](https://docs.fabfile.org/en/1.11/api/contrib/files.html?highlight=sed#fabric.contrib.files.sed) of that argument which says

```
The shell argument will be eventually passed to run/sudo. It defaults to False in order to avoid problems with many nested levels of quotes and backslashes. However, setting it to True may help when using ~fabric.operations.cd to wrap explicit or implicit sudo calls. (cd by it’s nature is a shell built-in, not a standalone command, so it should be called within a shell.)
```

We do use `cd` in this code (a few lines up as a context manager) so that seemed to be it.

PR #4459, opened 2021-01-13
cache state after clean releases

calellowitz:
We currently never checkpoint the completion of clean releases so `resume` will always run it again even if it has finished. That caused issues today, and is generally unnecessary, so this should properly checkpoint it, by caching state on completion rather than before attempting tasks.

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -497,8 +497,9 @@ def deploy_checkpoint(command_index, command_name, fn, *args, **kwargs):
     if env.resume and command_index < env.checkpoint_index:
         print(blue("Skipping command: '{}'".format(command_name)))
         return
-    cache_deploy_state(command_index)
     fn(*args, **kwargs)
+    cache_deploy_state(command_index + 1)
+
 
 
 def announce_deploy_start():


src/commcare_cloud/fab/fabfile.py
```diff
@@ -497,8 +497,9 @@ def deploy_checkpoint(command_index, command_name, fn, *args, **kwargs):
     if env.resume and command_index < env.checkpoint_index:
         print(blue("Skipping command: '{}'".format(command_name)))
         return
-    cache_deploy_state(command_index)
     fn(*args, **kwargs)
+    cache_deploy_state(command_index + 1)
```
dannyroberts, line 501: Might this now cause a similar problem where if things fail before this, there's an issue resuming?

calellowitz: I don't think so.
Old version
```
cache(0)
run(0)
cache(1)
run(1)
....
run(7)
```

new version
```
run(0)
cache(1)
run(1)
...
run(7)
cache(7)
```

and `run(0)` happens whether or not `0` is cached. It might not let you resume when it fails during `step 0` but the effect of doing a new deploy and resuming from `step 0` are the same.

dannyroberts: Gotcha. OK

dannyroberts (approved):

PR #4461, opened 2021-01-14
Fix command printing

millerdev:
Resolve error in Django checks phase of `cchq ... update-config`. Given the location of the issue, I'm assuming could have affected other commands as well.
```
  File "src/commcare_cloud/cli_utils.py", line 96, in <genexpr>
    command = ' '.join(shlex_quote(arg) for arg in command)
  File "/usr/lib/python3.6/shlex.py", line 314, in quote
    if _find_unsafe(s) is None:
TypeError: expected string or bytes-like object
```

In this case the `arg` was a `ansible.inventory.host.Host` object, which must be converted to a string before it is passed to `shlex_quote`.

##### ENVIRONMENTS AFFECTED
all / none specifically

### DIFF
--- a/src/commcare_cloud/alias.py
+++ b/src/commcare_cloud/alias.py
@@ -2,10 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 
 def _generate_args(*args, **kwargs):
-    argv = []
-
-    for arg in args:
-        argv.append(arg)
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):

gherceg:
Thanks @millerdev!

PR #4465, opened 2021-01-15
add caching at the proxy for app resource downloads

snopoke:
Add location and mappings to nginx config to cache app resources on the proxy. This is something we had set up for ICDS which protects the Django workers from multiple requests for the same resource.

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -17,10 +17,28 @@ nginx_sites:
       # (If a request sends _both_, that will break sticky routing
       # unless it's done perfectly consistently. Please don't!)
       method: "hash $cookie_formplayer_session$http_x_formplayer_session consistent"
+   mappings:
+     # Remove 'username' query param from request uri for use in cache key
+     # https://stackoverflow.com/a/45587134
+     # https://stackoverflow.com/a/35772678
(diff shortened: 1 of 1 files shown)


dannyroberts (commented):
App downloads are still considered non-secret, right? My only comment here is that this would ignore any auth done at the django level for these routes in the future

calellowitz (approved):

snopoke:
> App downloads are still considered non-secret, right? 

That's correct.

> My only comment here is that this would ignore any auth done at the django level for these routes in the future.

This is true. I don't believe there any current plans to support auth for these resources. Also any QA plan for that would likely fail with this enabled.

I'll put this on staging for now.

snopoke:
Rolled out to staging. I can see the caching happening in the nginx logs.

snopoke:
My main concern with this was Formplayer and whether the caching would impact live preview. I have confirmed that Formplayer does not use this endpoint to download apps, it uses the `api/download_ccz` endpoint. I have further confirmed that app preview and webapps work as expected with updating to released apps as well as live preview updating to the the latest version of the app.

snopoke:
@dannyroberts any reason to hold off on rolling this out? We're hoping this will reduce the impact of a future Formplayer change that will force re-installation of all apps on formplayer machines.

dannyroberts (approved):
Nope, no reason to hold off. Looks good to me!

snopoke:
I'll deploy on Monday

PR #4467, opened 2021-01-15
remove incorrect actuator properties

snopoke:
I don't think we are using actuator for anything and anyway, these properties are not correctly named, they should be `management.server.*`

##### ENVIRONMENTS AFFECTED
None


snopoke:
Merging this since these properties currently have no effect

PR #4472, opened 2021-01-19
Use pgbouncer6-production for commcarehq_proxy

dannyroberts:
From issues on pgbouncer3

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -54,7 +54,10 @@ dbs:
   form_processing:
     proxy:
       host: pgbouncer3
-      pgbouncer_host: pgbouncer6
+      pgbouncer_endpoint: pgbouncer6
+      pgbouncer_hosts:
+        - pgbouncer3
+        - pgbouncer6
     partitions:
       p1:
         shards: [0, 204]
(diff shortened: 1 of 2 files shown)


millerdev (approved):

environments/production/postgresql.yml
```diff
@@ -54,7 +54,10 @@ dbs:
   form_processing:
     proxy:
       host: pgbouncer3
-      pgbouncer_host: pgbouncer3
+      pgbouncer_endpoint: pgbouncer6
+      pgbouncer_hosts:
+        - pgbouncer3
+        - pgbouncer6
```
millerdev, line 60: I don't understand what these options are doing. Are they documented anywhere?

dannyroberts:
We opened this during firefighting to roll out if we needed to stop the bleeding but in the end I think we're planning on finding something better than this. Next steps will be planned in https://dimagi-dev.atlassian.net/browse/SAAS-11763.

environments/production/postgresql.yml
dannyroberts: Great question. These were introduced in #4447, but I didn't document them. I just wrote up some quick docs for them in https://github.com/dimagi/commcare-cloud/pull/4474.

PR #4475, opened 2021-01-20
Changes that help when reconfiguring pgbouncer

dannyroberts:
These are just generic things that help when configuring pgbouncer especially the ones related to pl/proxy. The logic fix generalizes better to other situations, and the tag just makes it easier to update the pg_hba when needed to accommodate a new pgbouncer machine if necessary.

##### ENVIRONMENTS AFFECTED
None

### DIFF
--- a/src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
@@ -10,7 +10,7 @@
 {% set current_host = inventory_hostname %}
 {% if is_monolith %}{% set current_host = DEFAULT_POSTGRESQL_HOST %}{% endif %}
 {% for db in postgresql_dbs.by_pgbouncer_host.get(current_host, []) %}
-{{ db.name }} = host={{ "localhost" if [db.host] == db.pgbouncer_hosts else db.host }} port={{ postgresql_port }}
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
```diff
@@ -10,7 +10,7 @@
 {% set current_host = inventory_hostname %}
 {% if is_monolith %}{% set current_host = DEFAULT_POSTGRESQL_HOST %}{% endif %}
 {% for db in postgresql_dbs.by_pgbouncer_host.get(current_host, []) %}
-{{ db.name }} = host={{ "localhost" if [db.host] == db.pgbouncer_hosts else db.host }} port={{ postgresql_port }}
+{{ db.name }} = host={{ "localhost" if db.host == inventory_hostname else db.host }} port={{ postgresql_port }}
```
dannyroberts, line 13: @skodde sorry to do this again, but if we accept this change, can you make sure to port this change over to #4454 as well?

dannyroberts: Just confirming this change was accepted, so we will indeed need to have it ported to your PR @skodde

skodde: Done, thanks for the heads up!

PR #4477, opened 2021-01-21
Add ability to specify io1 iops on RDS

dannyroberts:
Setting the iops param implies io1; not setting or setting to 0 implies gp2. We'll need this to change an RDS instance's storage type from gp2 to io1.

I've tested that this PR doesn't change the output of terraform plan on existing environments. I haven't tested that setting iops = 9000 (or whatever) matches up to manually changing the storage type to io1 with 9000 iops because we don't have any such machines, but it makes sense that it would and if it doesn't then this PR is at least a step towards there and it should be pretty clear how to do the rest to get terraform output back in sync once that's done.

##### ENVIRONMENTS AFFECTED
None


shyamkumarlchauhan (approved):

skodde (commented):
I think it would be useful to explicitly add storage_type, even if implicitly set to io1 when iops is set. It would provide better readability and clarity for future changes.

dannyroberts:
This is still a fine change, but we no longer need it, since we realized that IOPS and throughput were actually _not_ the limiting factors. So I'm going to close this, and if we need this sometime in the future we can come back to it (and follow up on skodde's very reasonable comment).

PR #4479, opened 2021-01-21
Update change log with Python 2 EOL date

millerdev:
Python 2 support will end on 2021-03-04 (cleared with Solutions team).

I plan to post an update to [this message](https://forum.dimagi.com/t/commcare-cloud-release-notes-2021-01-11/7939) on the forum after this PR is approved.

##### ENVIRONMENTS AFFECTED
all / none specifically


src/commcare_cloud/commcare_cloud.py
```diff
@@ -175,19 +175,24 @@ def call_commcare_cloud(input_argv=sys.argv):
     force_python_2 = "--force-commcare-cloud-to-use-python2"
     if not os.environ.get("TRAVIS_TEST") and sys.version_info[0] == 2:
         if force_python_2 not in input_argv:
-            from textwrap import dedent
-            print(dedent("""
+            exit(dedent("""
                 Error: you must upgrade to Python 3. Though not desirable, if you really have
```
gherceg, line 179: I think we should also update this error output to reflect the official drop date.

src/commcare_cloud/commcare_cloud.py
```diff
@@ -175,19 +175,24 @@ def call_commcare_cloud(input_argv=sys.argv):
...
                 to you can use Python 2 with this option {}
```
gherceg, line 180: I think we should also update this error output to reflect the official drop date

gherceg (approved):
Left a comment to include official drop date in the error output. Besides that this looks good.

src/commcare_cloud/commcare_cloud.py
millerdev: Good idea https://github.com/dimagi/commcare-cloud/pull/4479/commits/144ad80224e6f36cf730f786e17cac5be49169eb

dannyroberts (approved):

proteusvacuum (changes requested):
After our discussion on Slack (and discovering an issue with python 3 support in a commcare-cloud dependency), I just want to make sure that the core workflows of `commcare-cloud` have been tested before we completely remove support for python 2. 

Currently the `andrewrothstein.couchdb-cluster` role does not support py 3, which immediately will prevent any new commcare installations by third parties, but this could also come back to hurt us (e.g. we are fighting a couchdb fire, and realize that this role is actually broken).

millerdev:
The known issues with `andrewrothstein.couchdb-cluster` should be fixed by https://github.com/dimagi/commcare-cloud/pull/4481

PR #4488, opened 2021-01-25
Increase couchproxy max worker connections on production

dannyroberts:
Shyam noticed there were errors in the log showing that we had hit this limit

##### ENVIRONMENTS AFFECTED
production


dannyroberts:
This was applied yesterday

millerdev (approved):

PR #4492, opened 2021-01-27
nginx health check endpoint

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-11731

deploy this PR 
Change the Target group end point to /healthz
change success . code from . 302 to 200
and then set  the interval . to 29s
Observe the status of health . check . 
set the interval back to 30s

##### ENVIRONMENTS AFFECTED
Staging [ It first need to be tested here]
India 
Production

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/templates/default.j2
+++ b/src/commcare_cloud/ansible/roles/nginx/templates/default.j2
@@ -10,12 +10,3 @@ server {
       deny all;
   }
 }
-
-server {
-  listen {{ ansible_default_ipv4.address }}:80;
-
-  location /nginx-health {
-     access_log off;
-     return 200 "healthy\n";
-  }
-}
--- a/src/commcare_cloud/ansible/roles/nginx/templates/default.j2
+++ b/src/commcare_cloud/ansible/roles/nginx/templates/default.j2
@@ -10,3 +10,12 @@ server {
       deny all;
   }
 }
+
+server {
+  listen {{ ansible_default_ipv4.address }}:80;
+
+  location /nginx-health {
+     access_log off;
+     return 200 "healthy\n";
+  }
+}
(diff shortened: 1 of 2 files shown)


dannyroberts (approved):

dannyroberts:
Nice simple approach. This looks better than what we have now. It occurred to me there's some failure mode where we mess up the configuration of an nginx node so it can't reach webworkers or formplayer machines; in that case we would want it to be "unhealthy" since the issue would be with the connection configuration and not the with the downstream server. So in summary our previous health check will too often mark proxy servers Unhealthy incorrectly, while this check may in rare cases mark proxy servers Healthy incorrectly. I still think this is the better situation since the current false positive happens much more often than the theoretical false negative would.

Also in case you hadn't found it already, this https://github.com/dimagi/commcare-cloud/blob/1833bd90845efc7cd30a0a5b1e2d8bc8a165482b/src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf#L441-L448 is where we configure the health check in code. (Fine to do it manually first and/or submit that in a separate PR.)

shyamkumarlchauhan:
> Nice simple approach. This looks better than what we have now. It occurred to me there's some failure mode where we mess up the configuration of an nginx node so it can't reach webworkers or formplayer machines; in that case we would want it to be "unhealthy" since the issue would be with the connection configuration and not the with the downstream server. So in summary our previous health check will too often mark proxy servers Unhealthy incorrectly, while this check may in rare cases mark proxy servers Healthy incorrectly. I still think this is the better situation since the current false positive happens much more often than the theoretical false negative would.
> 
> Also in case you hadn't found it already, this
> 
> https://github.com/dimagi/commcare-cloud/blob/1833bd90845efc7cd30a0a5b1e2d8bc8a165482b/src/commcare_cloud/terraform/modules/ga_alb_waf/main.tf#L441-L448
> 
> is where we configure the health check in code. (Fine to do it manually first and/or submit that in a separate PR.)

PR for . terraform

https://github.com/dimagi/commcare-cloud/pull/4494

PR #4496, opened 2021-01-28
PgBouncer multiprocess tmpfiles configuration

skodde:
Add tmpfiles configuration for pgbouncer-multiprocess socket directories,
which are otherwise deleted by systemd-tmpfiles.

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/tasks/main.yml
@@ -110,6 +110,12 @@
     - pgbouncer
     - configure
 
+- name: pgbouncer tmpfiles configuration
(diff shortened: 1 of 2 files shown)


dannyroberts:
I'm not so familiar with systemd-tmpfiles. Do these configurations tell it _not_ to delete our pgbouncer pid files? Maybe a couple more sentences in the description is all I need

dannyroberts (approved):
Offlined with Marco. For others who don't already know how this works, my understanding from talking to Marco is that systemd manages the /run/ directory and in the worst case you can expect everything under /run/ to be volatile (blown away on restart) unless there's an entry for it in a file under /etc/tmpfiles.d/.

This was discovered on the staging server. The previous pgbouncer configuration was had pid files in `/run/postgresql` which is already properly managed, but to support multiple pgbouncer processes we need multiple pgbouncer pid directories, and this changes makes sure _those_ directories also get automatically created on reboot.

snopoke:
Thanks for the extra detail!

PR #4501, opened 2021-02-01
[formplayer] set -Xms to the same value as -Xmx and specify -XX:-UseCompressedOops

biyeun:
https://dimagi-dev.atlassian.net/browse/SAAS-11801

There are two ideas for alleviating the stress on formplayer machines:
1) increase number of machines by 10% (not this ticket)
2) tweak the jvm settings to be a little more robust. setting `-Xms` to the same value as `-Xmx` means that the gc process will be less stressed out and disabling `UseCompressedOops` would in theory allow us to bump the memory to more than 32GB. 

##### ENVIRONMENTS AFFECTED
all environments


dannyroberts:
By itself disabling compressed oops would only make things worse I believe. Like you said, if you have more than about 32GB of java heap then you can't use compressed oops, but I think if we wanted to have a larger heap we would want to be very careful about how to make sure that configuration works. It could unintuitively have worse performance.

PR #4506, opened 2021-02-02
Kev/cloudwatch agent install

knguyen142:
https://dimagi-dev.atlassian.net/browse/SAAS-11699 and https://dimagi-dev.atlassian.net/browse/SAAS-11773

##### ENVIRONMENTS AFFECTED

All

Changes to the `cloudwatch_logs` role so that if the agent is not installed on the machine, ansible installs it. Installation according to https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/download-cloudwatch-agent-commandline.html

## Testing

Tested on staging by uninstalling the agent on rabbit3 and using ansible to install. 

First run on rabbit3 to install (successfully ran the deb install):

```
PLAY RECAP *****************************************************************************************************************************************************************************
10.201.10.175              : ok=9    changed=5    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0

✓ Apply completed with status code 0
```

Second run on rabbit3 after install (skipped the logic to install cwa deb because agent found):

```
PLAY RECAP *****************************************************************************************************************************************************************************
10.201.10.175              : ok=5    changed=4    unreachable=0    failed=0    skipped=5    rescued=0    ignored=0

✓ Apply completed with status code 0
```

Unrelated run on rabbit2, which already had cwa agent installed (similar to above, skipped because cwa agent found):

```
PLAY RECAP *****************************************************************************************************************************************************************************
10.201.10.239              : ok=4    changed=2    unreachable=0    failed=0    skipped=6    rescued=0    ignored=0

✓ Apply completed with status code 0
```


dannyroberts:
Nice. I guess `sudo yum install amazon-cloudwatch-agent` didn't work?

knguyen142:
> Nice. I guess `sudo yum install amazon-cloudwatch-agent` didn't work? I think if it did it might be preferable to use yum (and [ansible's yum module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/yum_module.html)), but if not this works!

I think yum is just for Amazon's own OS Amazon Linux 2, and we're running a debian OS?

The preface for the yum install: `The CloudWatch agent is available as a package in Amazon Linux 2. If you are using this operating system, you can install the package by entering the following command...`

knguyen142:
It appears that `yum` can be used with Ubuntu, but many people are saying this might lead to chaos. @skodde What are your thoughts on this? Should we stick with the debian install way or install yum on Ubuntu to try and install the cloudwatch agent?

skodde:
I think we should definitely not mix different package managers, and my understanding is that yum and related utils are present on Ubuntu not to install packages directly but only to deal with them as needed (e.g. working with repos, bootstrapping, extracting files, etc.).

A couple of comments on the code itself:
1) You're setting 'update_cache: true' which should trigger an 'apt-get update, but you're manually installing a .deb and this step is not needed.
2) You're verifying the signature of the package but I don't see where you import the related public key. Unless I'm missing this passage, we should import it for the verification to be useful, and we should probably do that directly from a file we save in our configuration, and not by downloading it at the same time as the .deb and signature, otherwise we'd be using the same potentially insecure channel to get all of the pieces of the verification chain, making it useless.

Ideally we could make this more simple and generic by using instead the standard ansible package module and setting up a ppa for our use containing this package (and potentially others we may need), with the advantage that it can then be part of the normal update cycles and that we'll have control on the version available on our systems.
It would also be easier to adapt by other users, as they'd only need to configure the proper repo and not changing the code.

knguyen142:
Thanks, @skodde. I made the changes you suggested, let me know how it looks.

skodde (approved):

src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
```diff
@@ -1,26 +1,18 @@
...
-      url: https://s3.amazonaws.com/amazoncloudwatch-agent/debian/amd64/latest/amazon-cloudwatch-agent.deb
-      dest: /opt/
-      mode: '0666'
-    become: true
-  - name: install cloudwatch deb file
-    apt:
-      deb: "/opt/amazon-cloudwatch-agent.deb"
-      state: present
-    become: true
+
```
knguyen142, line 5: @prem-fissionhq I removed your changes here from https://github.com/dimagi/commcare-cloud/pull/4507 because this is a more secure way of doing it. I also had issues with collectd hanging when applying this to india, are you sure collectd is needed?  I can add it back if it's needed.

prem-fissionhq: Thanks @knguyen142 for suggestions. Will cross check and get back to you.

prem-fissionhq: Not mandatory @knguyen142. We generally use `collectd` software to send the metrics to the CloudWatch agent that's why I have installed it. Please go ahead and remove it. thanks.

knguyen142: Ah will we be missing something if we don't have that? We can definitely re-add, I was getting some errors so I removed it, but that might be a one off.

dannyroberts: I think it's for people who use cloudwatch the way we use datadog. I think there's no need to install collectd right now is my understanding

prem-fissionhq: Yes, Correct Danny.

PR #4513, opened 2021-02-05
Add swap (2G) to pgproxy2-staging

skodde:
##### ENVIRONMENTS AFFECTED
Staging

### DIFF
--- a/environments/staging/inventory.ini
+++ b/environments/staging/inventory.ini
@@ -37,7 +37,7 @@ web7
 web8
 
 [pgproxy2]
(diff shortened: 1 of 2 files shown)


millerdev (approved):
Curious to hear about your thoughts on swap sometime. I typically prefer to configure machines without swap and instead ensure they always have enough RAM. Troubleshooting why things are slow (possibly due to thrashing) has been much harder a problem to diagnose than "it fell over because it ran out or RAM".

skodde:
In this particular instance it was just an immediate way of letting the machine find the resources it needed during very limited and sporadic peaks, so the impact on performance was acceptable, especially on staging. In general I'd agree that memory should be sized properly, but a small amount of swap could still be beneficial under various workloads, not necessarily because of the additional space, but because of the way memory management works (at least on not very recent kernels).
Let's discuss offline about this.

PR #4519, opened 2021-02-08
cloudwatch_logs remove config file check

knguyen142:
On a fresh run the config file would be created and its status is not changed, so the reload will not happen. I couldn't find a way to run if it changed or was created, but removing it should also work and we always reload.  
I think the better thing to do is to use handlers and notify a reload, but that would be a bigger refactor.

### DIFF
--- a/src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
@@ -22,13 +22,11 @@
       owner: cwagent
       mode: 0644
     become: yes
-    register: cloudwatch_logs_conf
 
-  - name: Reload cloudwatch agent if changed
+  - name: Reload cloudwatch agent
     shell: /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s
       -c file:/opt/aws/amazon-cloudwatch-agent/etc/cloudwatch_logs.json
     become: yes
-    when: cloudwatch_logs_conf.changed
 
   - name: Ensure cloudwatch agent enabled
     shell: /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a start
--- a/src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
@@ -25,8 +25,9 @@
     notify: Reload cloudwatch agent
 
   - name: Ensure cloudwatch agent enabled
-    shell: /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a start
-    become: yes
+    shell: /bin/true
+    notify: Start cloudwatch agent
+    when: '"\"status\": \"running\"" not in ctl_status.stdout'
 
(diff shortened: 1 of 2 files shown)


skodde (commented):
I think you can easily move "Reload cloudwatch agent" into a handler and notify it from "Create cloudwatch logs config", so that it's called when the configuration is created or modified, without worrying about registering and checking its output.

knguyen142:
I'm sure there's a better way of doing this, I'm just not an ansible developer so the basic things are not so basic for me :)

skodde:
I just pushed a commit with that change, let me know if it makes sense and if it passes your tests.

knguyen142:
Thanks @skodde. I ran it and got 

```
TASK [cloudwatch_logs : Create cloudwatch logs config] **********************************************************************************************************
--- before
+++ after
@@ -1,5 +1,5 @@
 {
-    "group": 0,
-    "owner": 0,
+    "group": 999,
+    "owner": 999,
     "path": "/opt/aws/amazon-cloudwatch-agent/etc/cloudwatch_logs.json"
 }

changed: [10.201.10.175]

TASK [cloudwatch_logs : Ensure cloudwatch agent enabled] ********************************************************************************************************
changed: [10.201.10.175]

RUNNING HANDLER [cloudwatch_logs : Reload cloudwatch agent] *****************************************************************************************************
changed: [10.201.10.175]
```

Does this mean that the handler was run after the `Ensure cloudwatch agent enabled` event though we put it before? In either case, I'm not sure if it matters.

skodde:
Correct, as handlers run at the end by default, but we can call it earlier if needed.
What would be the correct sequence in this case?

knguyen142:
I'm used to testing it with the reload then starting the service, but maybe the reload also does a restart and that doesn't matter.

skodde:
It seems that the reload is indeed a restart and it will start the service:

`-s: optionally restart after configuring the agent configuration`

However, it still makes sense to maintain the original order and to add an additional check on the actual status of the service through the variable you register at the beginning.

See my last commit and let me know what you think.

src/commcare_cloud/ansible/roles/cloudwatch_logs/tasks/main.yml
```diff
@@ -22,16 +22,12 @@
...
-    shell: /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s
-      -c file:/opt/aws/amazon-cloudwatch-agent/etc/cloudwatch_logs.json
-    become: yes
-    when: cloudwatch_logs_conf.changed
+    notify: Reload cloudwatch agent
 
   - name: Ensure cloudwatch agent enabled
-    shell: /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a start
-    become: yes
+    shell: /bin/true
```
knguyen142, line 28: Doesn't this mean it always restarts? did you mean to do a status check?

skodde: The check is on this line:

`when: '"\"status\": \"running\"" not in ctl_status.stdout'`

Calling /bin/true is just a way to always notify the handler when the test passes.

knguyen142: Ahh I see you're checking the `ctl_status` from before. That makes sense and LGTM.

skodde (approved):

PR #4523, opened 2021-02-10
Adding Elasticache cluster node type & other attributes - Prod env only

rameshganne:
As per datadog historical data [ last 13 months] related to Production env Redis instance, The maximum used memory is 53.44 GiB out of usable memory 128 GiB and average used memory 14.75 GiB & 20.75 GiB per 6 months. 

In this case, we can use cache.r5.2xlarge which is having 8 vCPU & 52.82 GiB memory. Presently, we are using an r5.4xlarge [ 16 vCPU & 128 GiB memory ] instance for Redis service.

We are going to use three Elasticache nodes for high availability & launching these in different availability zones within the region.

We can scale up the Elasticache cluster node type based on service metrics like Memory usage, Evictions, Swap Usage & Memory fragmentation ratio.

ref: SAAS-11834

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -458,6 +458,21 @@ elasticache:
   create: no
 
 elasticache_cluster:
-  create: no
+  create: yes
+  cache_node_type: "cache.r5.2xlarge"
+  cache_engine_version: "4.0.10"
+  cache_prameter_group: "default.redis4.0"
+  automatic_failover: true
+  transit_encryption: false
+  at_rest_encryption: true
+  auto_minor_version: false
+  cluster_size: 3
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):

dannyroberts:
@shyamkumarlchauhan I'll let you coordinate with @rameshganne on when to merge this, but just a reminder that I think that you can merge and apply this a few days before the maintenance window to just make sure the setup works and is configured the way you want it. This might be preferable because then you can have the new redis endpoint already ready to go by the time you schedule the maintenance to migrate the data over and point app configs over to the new endpoint.

shyamkumarlchauhan (approved):

PR #4529, opened 2021-02-12
Sc/couch compact script

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-11851
This need to be tested on staging for a week before we merge it.

It seems view_compaction does not work with script which is currently in place.
so updating the script so that view compaction takes place.

##### ENVIRONMENTS AFFECTED
Staging
India
Production


dannyroberts:
Is this script run automatically anywhere, or just for manual use? My understanding is that we use couch's automated compaction https://docs.couchdb.org/en/2.3.1/config/compaction.html#compactions and that these do cover both db and view compactions. It may be that we're not running it aggressively enough since a change about a month ago: https://github.com/dimagi/commcare-cloud/pull/4486

src/commcare_cloud/ansible/roles/couchdb2/templates/compact_couchdb.sh.j2
```diff
@@ -109,7 +109,7 @@ do
       fi
 
       DEcho "${DB} - Getting view list ...";
-      DESIGNLIST=`CouchReq "GET" "/${DB}/_all_docs?startkey=\"_design/\"&amp;endkey=\"_design0\"&amp;include_docs=true" | awk -F ":" '{print $3}' | awk -F \" '{print $2}' | grep "^_design/" | sed -e s:"_design/"::g`;
+      DESIGNLIST=`CouchReq "GET" "/${DB}/_design_docs|jshon -e rows -a -e id -u|awk -F '/' '{print $2}'`;
```
dannyroberts, line 112: Was the issue here that the line above simply didn't work?

shyamkumarlchauhan: Seems , this script has never been in place in any of SaaS environments ever . as default value has been always false. https://github.com/dimagi/commcare-cloud/blob/3a3005834e2c8909898c0ded7887580819726255/src/commcare_cloud/ansible/roles/couchdb2/defaults/main.yml#L13

shyamkumarlchauhan: and also looks like the existing line is not producing a valid output which this line can take as input for compaction . https://github.com/dimagi/commcare-cloud/blob/3a3005834e2c8909898c0ded7887580819726255/src/commcare_cloud/ansible/roles/couchdb2/templates/compact_couchdb.sh.j2#L117

dannyroberts:
I guess I'm still confused why we need this and can't just tune couch's internal auto-compaction process to run more aggressively. @shyamkumarlchauhan thoughts?

shyamkumarlchauhan:
> I guess I'm still confused why we need this and can't just tune couch's internal auto-compaction process to run more aggressively. @shyamkumarlchauhan thoughts?

@dannyroberts  As per Datadog graph [https://app.datadoghq.com/dashboard/hr7-uza-kw5/couchdb-dashboard?from_ts=1610965975162&fullscreen_end_ts=1613557992947&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1610965992947&fullscreen_widget=33&live=true&to_ts=1613557975162](url) compaction on views is not an internal thing , however the compaction of DB is and hence we need this script so that view compaction can also run. [ that is what has freed up the space] .

dannyroberts:
Very strange. I'm pretty sure I've seen both db compaction and view compaction running before in active tasks. So I don't think it's just simply that view compactions aren't happening at all; but it does look from the graph you linked like the compaction is not _nearly_ aggressive enough to keep up

PR #4534, opened 2021-02-17
celery queue for ush background tasks

snopoke:
Dedicated queue for running USH background tasks to avoid blocking the existing queues. Currently the only task that would use this is the USH formplayer restore priming.

@calellowitz can I have moved the user import tasks to use this queue as well: https://github.com/dimagi/commcare-hq/pull/29181
I think this is safe to do since the user import tasks shouldn't be running in parallel with the formplayer priming tasks.

I've left the `user_import_queue` config in place for now to make the changes easier to roll out / roll back.

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -57,6 +57,8 @@ celery_processes:
       concurrency: 9
     user_import_queue:
       concurrency: 5
+    ush_background_tasks:
+      concurrency: 8
   celery_a006:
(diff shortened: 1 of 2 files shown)


calellowitz (approved):
Assuming we are comfortable with the resource usage, looks good to me.

dannyroberts (approved):

PR #4536, opened 2021-02-17
Update staging properties to support detailed tagging

gherceg:
Only allow `form_name` tag for the `web-apps-test` domain. This is to test this feature.
##### ENVIRONMENTS AFFECTED

Staging


src/commcare_cloud/environment/secrets/secrets.yml
```diff
@@ -166,3 +166,5 @@
   legacy_namespace: localsettings_private
 - name: ansible_sudo_pass
   required: yes
+- name: DETAILED_TAGGING_DOMAINS
```
gherceg, line 169: @dannyroberts is the new naming convention all lower case (e.g., `ansible_sudo_pass`) or is that an exception to the rule?

commcare-cloud-bootstrap/environment/private.yml.j2
```diff
@@ -48,3 +48,4 @@ localsettings_private:
   SECRET_KEY: '{{ generate_uuid() }}'
 
 ansible_sudo_pass: '{{ generate_uuid() }}'
+DETAILED_TAGGING_DOMAINS: ''
```
gherceg, line 51: @dannyroberts I'm not sure if this is necessary to be able to reference the variable in other files, like in the `public.yml` file below.

src/commcare_cloud/environment/secrets/secrets.yml
calellowitz: I believe the convention is to match where it is used. So, lowercase for these formplayer settings since they are lowercase in the formplayer config, and uppercase for the python localsettings, since they are all uppercase python constants.

gherceg: Thanks @calellowitz

dannyroberts (approved):
Looks good. Is this supposed to be formatted as a comma separated list of items in a string? Might want to make that clearer with a comment or something, though I guess people can look up how it's used

gherceg:
>Is this supposed to be formatted as a comma separated list of items in a string? Might want to make that clearer with a comment or something, though I guess people can look up how it's used

I should have thought about this sooner. Ideally it would end up in a format similar to the `formplayer_detailed_tags` variable (a list). However given that this is stored different, it seems my only option is to store this as a comma separated list inside a string, and have formplayer parse this correctly when reading it. Does that sound right?

dannyroberts:
@gherceg It sounds like you're asking whether the secrets backend supports only storing strings or if it has support for arbitrary JSON. It has support for storing arbitrary JSON so if you want it to be a list of strings it can be a list of strings. You then probably want the default value to be `[]`. Double check me on all of that of course, but I'm pretty sure we have settings that take advantage of that already, though not sure if we have any that are a list as opposed to an "object"/dict

src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
```diff
@@ -65,3 +65,8 @@ server.forward-headers-strategy=NATIVE
 {% if formplayer_enable_cache|default(False) %}
 spring.cache.type=caffeine
 {% endif %}
+
+{% if formplayer_detailed_tagging_domains is defined %}
+detailed_tagging.domains="{{ formplayer_detailed_tagging_domains|join('", "') }}"
+detailed_tagging.tag_names="{{ formplayer_detailed_tags|join('", "') }}"
```
gherceg, line 71: @calellowitz here's what I settled on. I preemptively did the same for `detailed_tagging.domains`, but wouldn't be surprised if that ends up formatted differently.

gherceg: So when either list is empty it'll result in one entry of an empty string on the formplayer side. I'm trying to justify this by telling myself these _shouldn't_ be empty, but the best solution seems to be to include a check on the list's length before formatting this way, otherwise leave it empty.

calellowitz (approved):

PR #4537, opened 2021-02-17
Create hsts variable to be used in both public and proxy files

gherceg:
To better maintain specifying an hsts value in both the public.yml and proxy.yml files, it is ideal to have one variable where we will update the value when necessary. 

For context on why there are two places we set this value: originally this value was added on the nginx side in the proxy.yml file. The downside of this approach was that we missed out on django's recommended security practices, so rather than silencing the current warning about not specifying an hsts value, we decided to also specify this value on the django side.
##### ENVIRONMENTS AFFECTED

Staging


environments/staging/public.yml
```diff
@@ -137,7 +137,7 @@ localsettings:
     - {name: 'ddtrace.contrib.django', host: webworkers}
   LOCAL_MIDDLEWARE:
     - 'django.middleware.security.SecurityMiddleware'
-  SECURE_HSTS_SECONDS: 300
+  SECURE_HSTS_SECONDS: {{ hsts_value_in_seconds }}
```
dannyroberts, line 140: I think you need the quotes here and above for this to be valid yaml. Remember that this is a yaml file where some of the strings are formatted in a way that ansible interprets as variable interpolation, _not_ a jinja2 file that outputs yaml.

Were you having an issue with these values showing up as strings / with quotes around them where they get dumped into templates for the relevant config files?

gherceg:
Closing in favor of the alternative approach mentioned in the PR above.

PR #4542, opened 2021-02-18
Sc/change compaction settings

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-11851

##### ENVIRONMENTS AFFECTED
Production


environments/production/public.yml
```diff
@@ -84,7 +84,7 @@ couchdb2:
   password: "{{ COUCH_PASSWORD }}"
 
 couchdb_compaction_settings:
-  _default: '[{db_fragmentation, "40%"}, {view_fragmentation, "40%"}, {from, "22:00"}, {to, "03:00"}]'
+  _default: '[{db_fragmentation, "50%"}, {view_fragmentation, "50%"}, {from, "22:00"}, {to, "03:00"}]'
```
dannyroberts, line 87: I might suggest widening the window a little bit. It might be that it chooses to only do a certain number at a time and that 5 hours a day isn't enough to start enough of the compactions to keep up with the day's growth. 14:00-22:00 is the highest load time for couch, so if we want to stop initiating compactions, say, 6 hours before that window, we could still afford to do `{from, "22:00"}, {to, "03:00"}`.

Also I just wanted to make sure our understanding of the change you made here is the same. I read this as raising the threshold at which a view or db is eligible for compaction. Previously if a view/db was 40% bigger than it needed to be then it was eligible, but now it needs to be 50% bigger than it needs to be in order to be eligible. In other words this makes fewer views/dbs eligible than before. That may be fine; I'm not sure how it decides which views to prioritize, but if raising the threshold forces it to focus on the more fragmented views first then that's a win for sure.

dannyroberts (approved):

PR #4549, opened 2021-02-23
update config and logging for formplayer sentry upgrade

snopoke:
Changes to formplayer config for sentry upgrade.

This needs to be rolled out in conjunction with the formplayer changes: https://github.com/dimagi/formplayer/pull/886
The logging config changes particularly will error if the old version of formplayer is running.

Is it worth trying to split this up into two:
1. Deploy config changes
2. Deploy formpalyer
3. Deploy logging changes (with restart)

##### ENVIRONMENTS AFFECTED
all

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -120,7 +120,6 @@
     mode: 0644
   with_items:
     - filename: application.properties
-    - filename: sentry.properties
     - filename: logback-spring.xml
   tags:
     - formplayer_deploy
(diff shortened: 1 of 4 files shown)


src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -120,7 +120,6 @@
     mode: 0644
   with_items:
     - filename: application.properties
-    - filename: sentry.properties
```
dannyroberts, line 123: Looks like the idea is that by just _not copying_ `sentry.properties` to the new release, we're essentially "deleting" it, and that's why there's no cleanup task necessary here. Took me a second, but makes sense

dannyroberts (approved):
Is it going to be a challenge at all to coordinate this with the corresponding formplayer release at all? I imagine if we time it right we just expect third parties to always deploy the latest formplayer from the latest commcare cloud and it should just work unless someone tries to do something outside of that model

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
snopoke: Yes, that's correct.

snopoke:
> Is it going to be a challenge at all to coordinate this with the corresponding formplayer release at all? I imagine if we time it right we just expect third parties to always deploy the latest formplayer from the latest commcare cloud and it should just work unless someone tries to do something outside of that model

Thanks for reminding me of this. I had forgotten that we deploy the config along with the each release.

PR #4554, opened 2021-02-26
Prem/kinesis firehose migration

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-11804

##### ENVIRONMENTS AFFECTED

Staging Kinesis firehose stream updated with hive partitions for better query performance. 
Please review @dannyroberts @shyamkumarlchauhan thanks.


dannyroberts (approved):
Looks good. Probably shouldn't merge until it's applied to production

prem-fissionhq:
@dannyroberts added lambda function to add partitions every hour. Still working on adding cloud watch event to trigger this lambda.

src/commcare_cloud/terraform/modules/logshipping/check_file_lambda.py
```diff
@@ -0,0 +1,113 @@
...
+    DATABASE = 'default'
+    TABLE = 'formplayer_request_response_logs'
+
+    # S3 constant
+    S3_OUTPUT = 's3://dimagi-commcare-staging-logs'
+    S3_BUCKET = 'dimagi-commcare-staging-logs'
+    STREAM_PATH='formplayer-request-response-logs-partitioned-staging'
+    verifyKey = f"{STREAM_PATH}/year={year}/month={month}/day={day}/hour={hour}"
+
+    addPartitionQuery=f"ALTER TABLE {TABLE} ADD PARTITION (year={year}, month={month}, day={day}, hour={hour}) LOCATION 's3://{S3_BUCKET}/{STREAM_PATH}/year={year}/month={month}/day={day}/hour={hour}/';"
```
dannyroberts, line 95: This will only add the partition for the current hour, so if the lambda function for some reason doesn't fire one hour (perhaps an error of some sort) then the partitions for those hours will silently never get added, right? Can this run `MSCK REPAIR {TABLE}` instead? Is that too expensive or something?

prem-fissionhq: It is too expensive Danny. It will scan all files and adds partitions.

dannyroberts (approved):

prem-fissionhq:
@dannyroberts Please review my PR. thanks.

shyamkumarlchauhan (approved):

dannyroberts (approved):

PR #4556, opened 2021-02-26
Add detailed tagging domains to production

gherceg:
Initially tested on [this PR](https://github.com/dimagi/commcare-cloud/pull/4536) on staging, but also just tested a branch with an empty `detailed_tag_names` list on staging to ensure it behaves as expected which it does. 

Rolling this out empty so that we can first add domains, then update which tags are considered "detailed" after. This will allow for a smooth transition from a reporting metrics side since `form_name` is already deployed. 
##### ENVIRONMENTS AFFECTED

Production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -65,6 +65,8 @@ formplayer_purge_time_spec: '10d'
 formplayer_sensitive_data_logging: true
 formplayer_forward_ip_proxy: true
 formplayer_enable_cache: true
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):

gherceg:
I realized I did not add empty brackets which is how I tested it on staging, and is more clearly intentional. Assuming it still looks good to you @dannyroberts.

dannyroberts (approved):

PR #4563, opened 2021-03-03
Enable django security middleware on swiss and india

gherceg:
These changes have [been on staging](https://github.com/dimagi/commcare-cloud/pull/4279) for awhile now and should be safe to roll out. Will start with India and Swiss and then rollout to Production.
##### ENVIRONMENTS AFFECTED

India, Swiss

### DIFF
--- a/environments/india/public.yml
+++ b/environments/india/public.yml
@@ -132,6 +132,11 @@ localsettings:
   INVOICE_PREFIX: India-
   INVOICE_STARTING_NUMBER: 5000
   J2ME_ADDRESS: "{{ J2ME_SITE_HOST }}"
+  LOCAL_MIDDLEWARE:
+    - 'django.middleware.security.SecurityMiddleware'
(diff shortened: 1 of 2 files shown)


gherceg:
@mjriley just wanted to make sure this doesn't conflict with any work you've done.

mjriley (approved):
Approving in that these changes don't break anything I'm doing. Not familiar with these settings, but I'd imagine staging exposed any issues?

dannyroberts (approved):

gherceg:
@dannyroberts it looks like this change worked on india but not on swiss. Is there an extra step needed to have this change take effect on swiss besides a deploy?

dannyroberts:
I think it would require `cchq <env> update-config` followed by a service restart (or deploy) in order to take affect. Not sure if that's what you meant; does that help?

gherceg:
🤦 definitely knew that but forgot. Thanks!

PR #4565, opened 2021-03-03
formpalyer config changes

snopoke:
##### ENVIRONMENTS AFFECTED
ALL

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -136,8 +136,6 @@
   with_items:
     - template: application.properties.j2
       filename: application.properties
-    - template: sentry.properties.j2
-      filename: sentry.properties
     - template: logback-spring.xml.j2
       filename: logback-spring.xml
   tags:
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -140,6 +140,7 @@
       filename: logback-spring.xml
   tags:
     - localsettings
+    - formplayer_deploy
 
 # This should be the last task in the formplayer release process
 # (and the last one tagged with `formplayer_deploy`)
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -136,12 +136,11 @@
   with_items:
     - template: application.properties.j2
       filename: application.properties
-    - template: sentry.properties.j2
```
snopoke, line 139: this file was removed here: https://github.com/dimagi/commcare-cloud/pull/4549

src/commcare_cloud/ansible/roles/formplayer/templates/logback-spring.xml.j2
```diff
@@ -7,8 +7,6 @@
         </encoder>
     </appender>
 
-    <appender name="SENTRY" class="io.sentry.logback.SentryAppender" />
```
snopoke, line 10: this is causing duplicate sentry events

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -136,12 +136,11 @@
...
-      filename: sentry.properties
     - template: logback-spring.xml.j2
       filename: logback-spring.xml
   tags:
     - localsettings
+    - formplayer_deploy
```
snopoke, line 143: adding this to deploy so that any config changes also go out on deploy

src/commcare_cloud/ansible/roles/formplayer/templates/logback-spring.xml.j2
gherceg: Just curious what prompted adding this [here](https://github.com/dimagi/commcare-cloud/pull/4549)?

gherceg (approved):
Looks good to me. Just left one non-blocking question.

snopoke:
@gherceg I added that when going through the sentry docs. The issue is that we report errors to sentry directly and we also log the errors. So with this config in place we end up double reporting. Sentry is able to deduplicate on the client side but only if you include the actual exception in the log which we don't do in many places.

PR #4570, opened 2021-03-04
Added error when entering a malformed secret

orangejenny:
It took some time for me to figure out how to update a secret that's a list of values.

##### ENVIRONMENTS AFFECTED
None

### DIFF
--- a/src/commcare_cloud/commands/secrets.py
+++ b/src/commcare_cloud/commands/secrets.py
@@ -47,7 +47,11 @@ class Secrets(CommandBase):
         try:
             secret_value = json.loads(secret_value)
         except ValueError:
-            pass
+            puts(color_error(
+                'Could not understand value, please use valid json: {value}'
+                .format(value=secret_value)
+            ))
+            exit(-1)
         environment.secrets_backend.set_secret(secret_name, secret_value)
 
 
--- a/src/commcare_cloud/commands/secrets.py
+++ b/src/commcare_cloud/commands/secrets.py
@@ -47,11 +47,13 @@ class Secrets(CommandBase):
         try:
             secret_value = json.loads(secret_value)
         except ValueError:
-            puts(color_error(
-                'Could not understand value, please use valid json: {value}'
-                .format(value=secret_value)
-            ))
-            exit(-1)
+            old_value = environment.secrets_backend.get_secret(secret_name)
+            if isinstance(old_value, list) or isinstance(old_value, dict):
+                puts(color_error(
+                    'Could not understand value, please use valid json: {value}'
+                    .format(value=secret_value)
+                ))
+                exit(-1)
         environment.secrets_backend.set_secret(secret_name, secret_value)
 
 


gherceg (commented):
I don't think I have enough knowledge of all of the different secrets to say whether **all secrets** should be json compatible which this change implies. Maybe @dannyroberts has better insight?

src/commcare_cloud/commands/secrets.py
```diff
@@ -47,7 +47,11 @@ def _secrets_edit(self, environment, secret_name):
         try:
             secret_value = json.loads(secret_value)
         except ValueError:
-            pass
+            puts(color_error(
+                'Could not understand value, please use valid json: {value}'
+                .format(value=secret_value)
+            ))
+            exit(-1)
```
dannyroberts, line 54: Yeah this is a bit of a blemish, but I wanted to accept here either JSON if it parses that way or else a string. For string values we to be able to type in `****myspecialsecret****` and not `"****myspecialsecret****"`, but we also want a way to input JSON.

That said If we're fine just making people type in `"****myspecialsecret****"` for a string secret, then this change will do that. Perhaps if we go that route we could be kind to that person and add something like 'If this was a string, did you mean to put it in double quotes ("...")?'. I'd say most of these secrets are strings

dannyroberts (commented):
I'm on the fence. See my comment. I think as long as nothing relies on this programmatically (and I don't think anything does?) it should be safe to do, and just changes the input convention

src/commcare_cloud/commands/secrets.py
orangejenny: I see your point. When most secrets are strings, having to add quotes does seems pretty annoying. What if I grabb the current value, and only raise an error here if the current value is a list/dict **and** the new secret doesn't parse as json? It wouldn't be perfect - for new values, I'd just assume the value should be a string, but that could be incorrect - but it'd be better.

dannyroberts: Yeah I suppose that would work!

orangejenny: de9cf8e2a978ec610f44b1f0e18115aa0b2880b0

src/commcare_cloud/commands/secrets.py
```diff
@@ -47,7 +47,13 @@ def _secrets_edit(self, environment, secret_name):
         try:
             secret_value = json.loads(secret_value)
         except ValueError:
-            pass
+            old_value = environment.secrets_backend.get_secret(secret_name)
+            if isinstance(old_value, list) or isinstance(old_value, dict):
+                puts(color_error(
+                    'Could not understand value, please use valid json: {value}'
+                    .format(value=secret_value)
```
millerdev, line 54: It seems bad or at least inconsistent to print the secret value after accepting it using `getpass` (concealed input).

millerdev (commented):
This seems like the wrong level of abstraction to be guessing at type constraints (that issue existed before this PR).

Example: However unlikely, if someone entered a secret (intended to be a string) that started with `["` and ended with `"]` (with no other `"`) then it will be interpreted incorrectly, and presumably once that is set it will not be easy/intuitive to change that secret back from a list to a string without entering it as a JSON-encoded string, which would be quite confusing.

Maybe if the new secret or old value is valid JSON it should start with a prompt: "The [old/new] value is a valid JSON [dict/list/string]. Did you intend for this value to be a JSON structure?" If the answer is yes, then accept the input as is, if not, then encode as JSON (thereby escaping unintentional JSON formatting).

orangejenny:
I don't think I'm going to get to this anytime soon. Closing out.

PR #4572, opened 2021-03-05
Add append and remove operations to secrets command

gherceg:
There are lots of domains to add to the `detailed_tagging_domains` secret and this will make my life a bit easier. Most secrets are not lists so this may not be that useful long term, but is definitely useful now. 
##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/docs/commcare-cloud/commands/index.md
+++ b/docs/commcare-cloud/commands/index.md
@@ -640,12 +640,12 @@ Output as CSV
 View and edit secrets through the CLI
 
 ```
-commcare-cloud <env> secrets {view,edit} secret_name
+commcare-cloud <env> secrets {view,edit,list-append,list-remove} secret_name
 ```
 
 ##### Positional Arguments
 
-###### `{view,edit}`
+###### `{view,edit,list-append,list-remove}`
 
 ###### `secret_name`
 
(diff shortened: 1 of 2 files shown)


dannyroberts (approved):
Looks good, my only comment is that I don't think a changelog entry is required. The changelog is only for things that we need to report out to third parties that they need to take some actions for things to continue working. Since this is something that (1) only affects SaaS envs and (2) doesn't require any particular action, we don't need a changelog entry for it.

Otherwise, looks good!

dannyroberts:
I think I mentioned this on some other channel, but in case I didn't, long term I think feature flags are the conceptually right construct for a giant list of domains that should receive some kind of special treatment. To use feature flags instead, the challenge would be enabling HQ and formplayer to share the same feature flag data in a reasonable way and write a little client library in java for however we decide that data should be shared.

gherceg:
>I think I mentioned this on some other channel, but in case I didn't, long term I think feature flags are the conceptually right construct for a giant list of domains that should receive some kind of special treatment. To use feature flags instead, the challenge would be enabling HQ and formplayer to share the same feature flag data in a reasonable way and write a little client library in java for however we decide that data should be shared.

Yep I have a personal ticket created to look into this in the future. If it seems worth making a platform ticket on JIRA I'm happy to do so.

orangejenny (approved):
I just looked at the last commit, Danny had approved the previous ones.

PR #4576, opened 2021-03-09
Enable detailed tagging on india and swiss

gherceg:
I should have done this sooner, but this will limit the values for `form_name` and `module_name` to be "_other" except for domains specified, which right now is none. Since this feature is specifically for USH I don't expect it to be used on these environments, but it is there if needed and will reduce Datadog billing in the meantime. 
##### ENVIRONMENTS AFFECTED

Swiss, India


gherceg:
Old and forgotten. Will create a new one to achieve a similar result now that feature flags are available on formplayer.

PR #4580, opened 2021-03-11
Auditcare db and routing setting for staging

millerdev:
Prepare to write auditcare data to PostgreSQL instead of CouchDB. This mirrors configuration that will be used on prod: separate DB, custom routing. Unlike prod, it uses the same pg host as other dbs (prod will use a separate RDS instance).

https://dimagi-dev.atlassian.net/browse/SAAS-11381

##### ENVIRONMENTS AFFECTED
staging

### DIFF
--- a/src/commcare_cloud/environment/schemas/postgresql.py
+++ b/src/commcare_cloud/environment/schemas/postgresql.py
@@ -232,7 +232,11 @@ class PostgresqlConfig(jsonobject.JsonObject):
             self.dbs.main, self.dbs.synclogs,
         ] + (
             self.dbs.form_processing.get_db_list() if self.dbs.form_processing else []
-        ) + [self.dbs.ucr, self.dbs.formplayer] + self.dbs.custom + self.dbs.standby if _f]
+        ) + [
+            self.dbs.ucr,
+            self.dbs.formplayer,
+            self.dbs.auditcare,
+        ] + self.dbs.custom + self.dbs.standby if _f]
 
     def _check_reporting_databases(self):
         referenced_django_aliases = set()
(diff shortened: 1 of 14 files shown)


src/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -294,6 +298,7 @@ class SmartDBConfig(jsonobject.JsonObject):
     ucr = jsonobject.ObjectProperty(lambda: UcrDBOptions, required=True)
     synclogs = jsonobject.ObjectProperty(lambda: SynclogsDBOptions, required=False)
     form_processing = jsonobject.ObjectProperty(lambda: FormProcessingConfig, required=False)
+    auditcare = jsonobject.ObjectProperty(lambda: AuditcareDBOptions, required=False)
```
millerdev, line 301: This change and/or the one above it appears to have caused an "auditcare" database to be created on every environment, which was not my intention. I only want an auditcare database if specified explicitly in `postgresql.yml` for the environment. I thought that's what `required=False` would do here, but I guess not? Is there some other option for that?

tests/test_postgresql_config.py
```diff
@@ -42,3 +45,24 @@ def test_postgresql_config(env_name):
             lineterm=''
         ))
     ))
+
+
+def update_configs():
+    for env_name in TEST_ENVIRONMENTS:
+        env = Environment(DefaultPaths(env_name, environments_dir=TEST_ENVIRONMENTS_DIR))
+        if os.path.exists(env.paths.generated_yml):
+            print("updateing config:", env_name)
```
mjriley, line 54:
```suggestion
            print("updating config:", env_name)
```

src/commcare_cloud/environment/schemas/postgresql.py
snopoke: Looks like you have to explicitly set it to `null` to have it excluded: https://github.com/dimagi/commcare-cloud/blob/master/.travis/environments/travis/postgresql.yml#L5.

I think if you made the default `None` it would be excluded unless specified in the yaml:

```
auditcare = jsonobject.ObjectProperty(lambda: AuditcareDBOptions, required=False, default=None)
```

millerdev:
Cleaned up commit history with rebase, which included separating [formatting-only changes to generated postgres test configs](https://github.com/dimagi/commcare-cloud/pull/4580/commits/b72994f80753734ccad300de2a90c1f1248d6276) from the [introduction of the new "auditcare" db](https://github.com/dimagi/commcare-cloud/pull/4580/commits/20e6914c6c675bc99210e2b3123297730f217e82).

snopoke (approved):

millerdev:
Applied to staging with
```sh
cchq --control staging ap deploy_postgres.yml --branch=dm/auditcare-staging
```

PR #4584, opened 2021-03-15
drop superfluous /

snopoke:
This updates the formplayer URL mapping to match `/formplayer/` (extra `/` at the end). The result of this is that formplayer will no longer receive requests with an initial double `/` e.g `//new-form`.

This fix is required for the rollout of https://github.com/dimagi/formplayer/pull/931 which rejects the double slash.

##### ENVIRONMENTS AFFECTED
ALL (already deployed on staging)

This should also get a changelog which I will add tomorrow.


skodde (approved):

dannyroberts (approved):

src/commcare_cloud/manage_commcare_cloud/make_docs.py
```diff
@@ -70,7 +70,7 @@ def wrap_lines(text):
     def escape_html_outside_backticks(text):
         parts = text.split('`')
         for i in range(0, len(parts), 2):
-            parts[i] = cgi.escape(parts[i])
+            parts[i] = html.escape(parts[i], quote=False)
```
snopoke, line 73: cgi.escape is deprecated in 3.2 and removed in 3.8. The alternative is `html.escape` which has been included since 3.2.

snopoke:
Also posted on the forum: https://forum.dimagi.com/t/commcare-cloud-release-notes-2021-03-16/8088

PR #4586, opened 2021-03-15
Do not run service check on remote database

millerdev:
set_up_dbs.yml on remote db may have been broken by https://github.com/dimagi/commcare-cloud/commit/7b5a7f456ea16fde7c95428bc1aedd7fe2127cf9


dannyroberts (approved):
Nice find. Looks like this task was added without the "remote db" model in mind, and skipping it seems fine

PR #4601, opened 2021-03-23
Update password text to token for Github

gherceg:
The use of a password in this context has been deprecated and we need to use our token to login when deploying, not a password.

From what I've experienced, following the current text and putting a token in for the first field ("Github username or token") does not work.

##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/src/commcare_cloud/fab/utils.py
+++ b/src/commcare_cloud/fab/utils.py
@@ -153,15 +153,14 @@ class DeployMetadata(object):
         return DeployDiff(self.repo, self.last_commit_sha, self.deploy_ref)
 
 
-GITHUB_CREDENTIALS = None
+GITHUB_TOKEN = None
 
 
-def _get_github_credentials(message, force=False):
-    global GITHUB_CREDENTIALS
+def _get_github_token(message, force=False):
+    global GITHUB_TOKEN
 
-    if GITHUB_CREDENTIALS is not None:
-        if not force or GITHUB_CREDENTIALS[0]:
-            return GITHUB_CREDENTIALS
+    if GITHUB_TOKEN is not None and not force:
+        return GITHUB_TOKEN
 
     try:
         from .config import GITHUB_APIKEY
@@ -173,15 +172,13 @@ def _get_github_credentials(message, force=False):
             "    $ cp {project_root}/config.example.py {project_root}/config.py\n"
             "Then edit {project_root}/config.py"
         ).format(project_root=PROJECT_ROOT))
-        username = input('Github username (leave blank to skip): ') or None
-        token = getpass('Github token: ') if username else None
-        GITHUB_CREDENTIALS = (username, token)
+        GITHUB_TOKEN = getpass('Github token: ')
     else:
-        GITHUB_CREDENTIALS = (GITHUB_APIKEY, None)
-    return GITHUB_CREDENTIALS
+        GITHUB_TOKEN = GITHUB_APIKEY
+    return GITHUB_TOKEN
 
 
-def get_github_credentials(message=None, force=False):
+def get_github_token(message=None, force=False):
     if not message:
         message = "This deploy script uses the Github API to display a summary of changes to be deployed."
         if env.tag_deploy_commits:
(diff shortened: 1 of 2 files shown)


biyeun (approved):

joxl (approved):

src/commcare_cloud/fab/utils.py
```diff
@@ -173,9 +173,9 @@ def _get_github_credentials(message, force=False):
             "    $ cp {project_root}/config.example.py {project_root}/config.py\n"
             "Then edit {project_root}/config.py"
         ).format(project_root=PROJECT_ROOT))
-        username = input('Github username or token (leave blank to skip): ') or None
-        password = getpass('Github password: ') if username else None
-        GITHUB_CREDENTIALS = (username, password)
+        username = input('Github username (leave blank to skip): ') or None
+        token = getpass('Github token: ') if username else None
+        GITHUB_CREDENTIALS = (username, token)
```
millerdev, line 178: Username should not be needed. Does that sound right?
```suggestion
        token = input('Github token (leave blank to skip): ') or None
        GITHUB_CREDENTIALS = (token, None)
```

millerdev (commented):
This has annoyed me for a while. Thank you!

src/commcare_cloud/fab/utils.py
gherceg: I think you're right and meant to test this myself. I've always put in my username, probably unnecessarily so. Just confirming you typically skip that field and just put in the token?

millerdev: No, if you skip the field it will not prompt for the token, so you currently have to enter both (unnecessarily).

gherceg:
Turned into more of a refactor. Even though the Github class accepts a username and password, entering username/password combo does not work on our end. Rather than look into why it fails, it seems easy enough to say we only support tokens since they are more secure. @snopoke would like to get your input on this.

src/commcare_cloud/fab/utils.py
```diff
@@ -173,34 +172,32 @@ def _get_github_credentials(message, force=False):
...
-    login_or_token, password = get_github_credentials()
-    return Github(login_or_token=login_or_token, password=password)
+    token = get_github_token()
+    return Github(login_or_token=token, password=None)
 
 
 @memoized
 def _github_auth_provided():
-    return bool(get_github_credentials()[0])
+    return bool(get_github_token()[0])
```
millerdev, line 200: I think this has a bug
```suggestion
    return bool(get_github_token())
```

millerdev (changes requested):
Nice refactor

src/commcare_cloud/fab/utils.py
```diff
@@ -153,15 +153,14 @@ def diff(self):
...
 
-def _get_github_credentials(message, force=False):
-    global GITHUB_CREDENTIALS
+def _get_github_token(message, force=False):
+    global GITHUB_TOKEN
 
-    if GITHUB_CREDENTIALS is not None:
-        if not force or GITHUB_CREDENTIALS[0]:
-            return GITHUB_CREDENTIALS
+    if GITHUB_TOKEN is not None and not force:
```
snopoke, line 162: I think this logic is wrong. The intention is to allow blank tokens unless 'force=true' in which case a token is required. I've also dropped the `not None` since an empty token is equivalent to `None` as far as the Github API library is concerned.

```suggestion
    if GITHUB_TOKEN or not force:
```

src/commcare_cloud/fab/utils.py
snopoke: +1

snopoke (changes requested):

src/commcare_cloud/fab/utils.py
gherceg: Thanks for catching this.

src/commcare_cloud/fab/utils.py
gherceg: Thanks I didn't understand the intention originally but it makes sense now. Made the change and updated the parameter name to from `force` to `required`.

snopoke (approved):

millerdev (approved):

PR #4600, opened 2021-03-23
SQL auditcare configuration for prod and india

millerdev:
Preparation for rolling out [SQL auditcare](https://github.com/dimagi/commcare-hq/pull/29377).

##### ENVIRONMENTS AFFECTED
production and india

PR #4602, opened 2021-03-23
Add changelog for auditcare move to SQL

dannyroberts:
https://github.com/dimagi/commcare-hq/pull/29377. Changes recommended here are based on what we did here https://github.com/dimagi/commcare-cloud/pull/4600.

Rolling the change out a bit faster than imagined to prevent a problem we had today.

##### ENVIRONMENTS AFFECTED
None


changelog/0043-optionally-configure-auditcare-db.yml
```diff
@@ -0,0 +1,42 @@
+title: Optionally configure auditcare db
+key: optionally_configure_auditcare_db
+date: 2021-03-23
+optional_per_env: yes
+min_commcare_version:
```
millerdev, line 5: Should we add a revision here? Maybe https://github.com/dimagi/commcare-hq/pull/29377/commits/87ac2df454b4bcf0ebd623dbd208f014eba546b0 (the last commit of https://github.com/dimagi/commcare-hq/pull/29377)?

changelog/0043-optionally-configure-auditcare-db.yml
```diff
@@ -0,0 +1,42 @@
...
+  If you want to have it write to a different database instead, you can configure this using the
+  `LOCAL_CUSTOM_DB_ROUTING` variable and an entry in postgresql.yml.
+update_steps: |
+  There are no special steps if you are fine with the default.
+
+  To configure auditcare to go to its own PostgreSQL database instead, use the following:
+  ```
+  # public.yml
+  LOCAL_CUSTOM_DB_ROUTING:
+    auditcare: auditcare
```
millerdev, line 25:
```suggestion
  localsettings:
    LOCAL_CUSTOM_DB_ROUTING:
      auditcare: auditcare
```

changelog/0043-optionally-configure-auditcare-db.yml
```diff
@@ -0,0 +1,42 @@
...
+  ```
+
+  If you have more than one postgres machine and you want auditcare on one other than the default one,
+  update your postgresql.yml to point to the right machine. If you have only one postgres machine
+  or want to use your default postgres machine, you can skip this step.
```
millerdev, line 30: I think an `auditcare:` entry is required in postgresql.yml if you want to have auditcare write to a separate database, so you cannot "skip this step." If this step is skipped the auditcare database will not be created.

changelog/0043-optionally-configure-auditcare-db.yml
```diff
@@ -0,0 +1,42 @@
...
+
+  ```
+  # postgresql.yml
+  auditcare:
+    host: <machine>
```
millerdev, line 35:
```suggestion
  dbs:
    auditcare:
      host: <machine>
```

changelog/0043-optionally-configure-auditcare-db.yml
```diff
@@ -0,0 +1,42 @@
+title: Optionally configure auditcare db
+key: optionally_configure_auditcare_db
+date: 2021-03-23
+optional_per_env: yes
+min_commcare_version:
+max_commcare_version:
+context: |
+  Some enterprise deployments of CommCare use a backend feature called "auditcare" as part of
+  their audit logging strategy. Historically it has used CouchDB as its data backend,
+  but it is being switched to use PostgreSQL instead. If you care about this feature then
```
millerdev, line 10: What do you think of adding another sentence in here? Something like

_Auditcare is enabled by default, so it is active unless you went out of your way to disable it when you configured CommCare._

changelog/0043-optionally-configure-auditcare-db.yml
millerdev:
```suggestion
min_commcare_version: 87ac2df454b4bcf0ebd623dbd208f014eba546b0
```

changelog/0043-optionally-configure-auditcare-db.yml
dannyroberts: For other dbs you don't have to. If you don't mention a db, it'll default to pointing to `DEFAULT_POSTGRESQL_HOST` which defaults to the first `postgresql` host. For example you can see in https://github.com/dimagi/commcare-cloud/blob/e6b6b8f0c1cec257c520c3ac2fe11b4f749088f8/environments/swiss/postgresql.yml#L23-L26 that not much is specified under `dbs`.

changelog/0043-optionally-configure-auditcare-db.yml
dannyroberts: Like this?

```suggestion
  but it is being switched to use PostgreSQL instead.
  Auditcare is enabled and minimally configured by default,
  so it is active unless you went out of your way to disable it when you configured CommCare.
  If you care about this feature then
```

changelog/0043-optionally-configure-auditcare-db.yml
millerdev: Eh, I think this is wrong. Maybe we want to set `max_commcare_version` instead? Or maybe not at all. I'm not sure.

changelog/0043-optionally-configure-auditcare-db.yml
dannyroberts: Daniel and I offlined, and https://github.com/dimagi/commcare-cloud/blob/31e36f990a9714bd81c3579973b93473a776a16e/src/commcare_cloud/environment/schemas/postgresql.py#L301 is what I missing.

millerdev (approved):

proteusvacuum:
This looks great. Thanks!

PR #4606, opened 2021-03-24
updated Production resources files - NFS to EFS migration plan ref: SAAS-12000

rameshganne:
I am adding the required changes to mount the EFS file-system on NFS client hosts.

_Information about changes_:

**_src/commcare_cloud/ansible/deploy_shared_dir.yml_** - updating with play-book name (deploy_efs_mount_shared_dir.yml)

updating AWS production resource  file and required variables of EFS ansible role : **_environments/production/aws-resources.yml_** and **_environments/production/inventory.ini.j2_**

Environment : Production
Verified migration process in Staging & India envs.
review request: @dannyroberts

### DIFF
--- a/src/commcare_cloud/ansible/deploy_shared_dir.yml
+++ b/src/commcare_cloud/ansible/deploy_shared_dir.yml
@@ -1,2 +1,2 @@
 ---
-- import_playbook: roles/shared_dir.yml
+- import_playbook: deploy_efs_mount_shared_dir.yml
--- a/src/commcare_cloud/ansible/deploy_shared_dir.yml
+++ b/src/commcare_cloud/ansible/deploy_shared_dir.yml
@@ -5,7 +5,4 @@
     - efs_shared_mount_dir is defined
     - efs_shared_mount_options is defined
 - import_playbook: roles/shared_dir.yml
-  when: 
-    - efs_shared_mount_endpoint is not defined
-    - efs_shared_mount_dir is not defined
-    - efs_shared_mount_options is not defined
+  when: efs_shared_mount_endpoint is not defined
--- a/src/commcare_cloud/ansible/deploy_shared_dir.yml
+++ b/src/commcare_cloud/ansible/deploy_shared_dir.yml
@@ -1,2 +1,11 @@
 ---
 - import_playbook: deploy_efs_mount_shared_dir.yml
+  when:
+    - efs_shared_mount_endpoint is defined
+    - efs_shared_mount_dir is defined
+    - efs_shared_mount_options is defined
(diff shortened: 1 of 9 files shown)


dannyroberts (approved):

src/commcare_cloud/ansible/deploy_shared_dir.yml
```diff
@@ -1,2 +1,2 @@
 ---
-- import_playbook: roles/shared_dir.yml
+- import_playbook: deploy_efs_mount_shared_dir.yml
```
dannyroberts, line 2: I'm not sure I understand why we have to replace the old playbook with the new one.

Keep in mind that third party environments use commcare-cloud too as an open source tool for managing their commcare deployments, and they won't be using EFS. So whatever we do we need to support both NFS and EFS.

dannyroberts: A potential easy solution is to add

```suggestion
- import_playbook: roles/shared_dir.yml
  when: efs_shared_mount_dir is not defined
- import_playbook: deploy_efs_mount_shared_dir.yml
  when: efs_shared_mount_dir is defined
```

I could be wrong. If that makes sense to you, try running that against staging and making sure it does the right thing.

dannyroberts (changes requested):

rameshganne:
I have updated PR to support EFS/NFS service. Here is the information about changes:
- src/commcare_cloud/ansible/deploy_shared_dir.yml included conditions for EFS: 
`- efs_shared_mount_endpoint is defined - efs_shared_mount_dir is defined - efs_shared_mount_options is defined` allow to configure EFS mount points in client instances.
`efs_shared_mount_endpoint is not defined` - allow to configure NFS mount points in client instances.

- In the above EFS conditions, mentioned three variables are mandatory to configure EFS. 
- In above NFS conditions, If we are not mention` efs_shared_mount_endpoint` in public.yml [ example: environments/production/public.yml ] file, It configure NFS mount-points in instances by using role [ commcare-cloud/src/commcare_cloud/ansible/roles/shared_dir ].

- Updated variable name for mount options & removed `efs_mount_dir_data_sync` variable which is not need all time [one time use only] in India, Prod & Staging environments public.yml files [example: environments/india/public.yml]

- Updated play-book with mount options variable name - src/commcare_cloud/ansible/deploy_efs_mount_shared_dir.yml & src/commcare_cloud/ansible/deploy_nfs_efs_data_sync.yml

We can use NFS also by removing `efs_shared_mount_endpoint` from public.yml [ example: environments/production/public.yml ] file.

dannyroberts (approved):

PR #4609, opened 2021-03-25
Add additional certificate to certifi for celery

skodde:
Add the needed "DigiCert SHA2 Secure Server CA" intermediate
certificate to certifi, in order to fix the validation error
with smspro.mtn.ci until they can fix their configuration to
serve the complete certificate chain.

https://dimagi-dev.atlassian.net/browse/SAAS-11999

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
+++ b/src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
@@ -16,7 +16,6 @@
       - "{{supervisor_celery_workers}}"
 
 - name: Add celery_bash_runner files
-  become: yes
   template:
     src: "../templates/celery_bash_runner.sh.j2"
     dest: "{{ item.file_name }}"
@@ -29,7 +28,6 @@
       python_options: "-O"
 
 - name: define special celery services
-  become: yes
   template:
     src: "{{ item.template }}"
     dest: "{{ item.file_name }}"
@@ -46,7 +44,6 @@
       should_apply: "{{ app_processes_config.celery_processes.get(inventory_hostname).beat is defined }}"
 
 - name: define celery workers
-  become: yes
   template:
     src: "../templates/supervisor_celery_workers.conf.j2"
     dest: "{{ supervisor_celery_workers }}"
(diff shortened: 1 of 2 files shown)


gherceg (commented):
Mentioned this over slack, but it looks like we may need to use Python's SSL instead, as this [SO post](https://stackoverflow.com/questions/27068163/python-requests-not-handling-missing-intermediate-certificate-only-from-one-mach/66111417#66111417) outlines.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
```diff
@@ -16,7 +16,6 @@
       - "{{supervisor_celery_workers}}"
 
 - name: Add celery_bash_runner files
-  become: yes
```
gherceg, line 19: Just curious are these removals just cleanup because sudo isn't necessary?

gherceg (approved):
Approved. I tested this on staging since Marco made this change and it works as expected. Just left a question for my own understanding.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
```diff
@@ -61,3 +58,38 @@
         no_proxy: "{% if http_proxy_address is defined %}{{ groups['all'] | join(',') }},{{ app_processes_config.additional_no_proxy_hosts }}{% endif %}"
         PROMETHEUS_PUSHGATEWAY_HOST: "{% if prometheus_celery_pushgateway is defined %}{{ prometheus_celery_pushgateway }}{% endif %}"
         TMPDIR: '{{ encrypted_tmp }}'
+
+- name: add "DigiCert SHA2 Secure Server CA" certificate to certifi
+  blockinfile:
+    path: "{{ py3_virtualenv_home }}/lib/python3.6/site-packages/certifi/cacert.pem"
```
millerdev, line 64: For my own understanding, this task will add an extra CA to this file, but should not affect any other CAs (of which I'm guessing there are many) in that file?

At first I thought it was replacing the file, which seemed alarming since that would break SSL for many sites. Do I have that right?

millerdev: Should we add a task to a backlog somewhere to revisit and verify
- this is still needed
- the cert is still valid and has not been compromised

Unfortunately I'm not sure we have good procedural support to ensure that type of task actually gets done.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
```diff
@@ -61,3 +58,38 @@
...
+      YmFsUm9vdENBLmNybDA3oDWgM4YxaHR0cDovL2NybDQuZGlnaWNlcnQuY29tL0Rp
+      Z2lDZXJ0R2xvYmFsUm9vdENBLmNybDAwBgNVHSAEKTAnMAcGBWeBDAEBMAgGBmeB
+      DAECATAIBgZngQwBAgIwCAYGZ4EMAQIDMA0GCSqGSIb3DQEBCwUAA4IBAQB3MR8I
+      l9cSm2PSEWUIpvZlubj6kgPLoX7hyA2MPrQbkb4CCF6fWXF7Ef3gwOOPWdegUqHQ
+      S1TSSJZI73fpKQbLQxCgLzwWji3+HlU87MOY7hgNI+gH9bMtxKtXc1r2G1O6+x/6
+      vYzTUVEgR17vf5irF0LKhVyfIjc0RXbyQ14AniKDrN+v0ebHExfppGlkTIBn6rak
+      f4994VH6npdn6mkus5CkHBXIrMtPKex6XF2firjUDLuU7tC8y7WlHgjPxEEDDb0G
+      w6D0yDdVSvG/5XlCNatBmO/8EznDu1vr72N8gJzISUZwa6CCUD7QBLbKJcXBBVVf
+      8nwvV9GvlW+sbXlr
+      -----END CERTIFICATE-----
```
dannyroberts, line 94: Can you go through your process of verifying this certificate and/or what mitigations there are if it's an incorrect certificate and/or when it expires? Just want to make sure we think about that carefully and document our reasoning because trusting a malicious CA certificate gives its holder the ability to impersonate any website to us.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
skodde: All tasks had/needed it, so I just put `become: true` in src/commcare_cloud/ansible/deploy_commcarehq.yml when including it instead, as it was done for similar cases. It should be safe because it's not included anywhere else.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
skodde: > For my own understanding, this task will add an extra CA to this file, but should not affect any other CAs (of which I'm guessing there are many) in that file?

It adds an intermediate certificate signed by a CA root certificate we already trust (it's included by default), without affecting the others.

> At first I thought it was replacing the file, which seemed alarming since that would break SSL for many sites. Do I have that right?

Yes, it just adds that block, it does not replace the file, which would indeed break TLS connections to all sites using certificates signed by a different intermediate or root certificate.

> Should we add a task to a backlog somewhere to revisit and verify
> 
>     * this is still needed
> 
>     * the cert is still valid and has not been compromised
> 
> 
> Unfortunately I'm not sure we have good procedural support to ensure that type of task actually gets done.

I think that's a good idea. Ideally we want to keep this in place only until smspro.mtn.ci fixes their configuration.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
skodde: I downloaded the intermediate certificate here: https://cacerts.digicert.com/DigiCertSHA2SecureServerCA-2.crt.pem
and verified it is signed by the root certificate "DigiCert Global Root CA", which is included and trusted by default in certifi.

It expires on 22/Sep/2030, and we could monitor for revocation alerts until the configuration on the other end is fixed and we can safely revert this change.

dannyroberts: Okay great, thanks for explaining. That sounds like a safe procedure for being certain we have a trustworthy cert here: you pulled it from the known website over https, and you verified that the contents were signed by a cert CA that we already trust by default included in this widely used library.

For me and anyone else, how can one determine that this cert is signed by a cert included in `certifi`? Is there some python expression or command line command you can run on it that shows that?

dannyroberts (approved):

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
skodde: One easy way would be to install the certifi package and use the openssl verify command by specifying the file of trusted certificates provided by certifi:

`openssl verify -CAfile venv/lib/python3.6/site-packages/certifi/cacert.pem DigiCertSHA2SecureServerCA-2.crt`

PR #4612, opened 2021-03-26
set acceptable maintenance window for production

snopoke:
This will provide a warning during deploy if it is outside of the maintenance window.

##### ENVIRONMENTS AFFECTED
Production


millerdev (approved):

dannyroberts (approved):
Very nice, thanks

PR #4613, opened 2021-03-26
Remove more Python 2 code

millerdev:
Still todo: completely remove `six`

Review by commit.


control/init.sh
```diff
@@ -157,11 +121,7 @@ source ${COMMCARE_CLOUD_REPO}/src/commcare_cloud/.bash_completion
 YELLOW='\033[0;33m'
 BLUE='\033[0;34m'
 NC='\033[0m' # No Color
-if [ "$CCHQ_PYTHON" == 3 ]; then
-    SITE_PACKAGES=$(python -c 'import site; print(site.getsitepackages()[0])')
-else
-    SITE_PACKAGES=$VENV/lib/python2.7/site-packages
-fi
+SITE_PACKAGES=$(python -c 'import site; print(site.getsitepackages()[0])')
```
gherceg, line 123: Does this need to be `python3`? I see it wasn't before but just checking.

gherceg (approved):
This feels good 😀 one question but nothing blocking. Looks good.

control/init.sh
millerdev: It doesn't because the virtualenv is activated at this point, so `python` is the same as `python3`

Probably the same applies to the one in `realpath`, but I changed that one because it is defined before the virtualenv is activated.

dannyroberts (approved):

PR #4614, opened 2021-03-26
Display help message with no args

mjriley:
This should fix trying to execute commcare-cloud with no arguments. Previously, it would error out as the deploy command's documentation expected at least 1 argument.  Additionally, changed the behavior so it will display the help documentation with no arguments, rather than simply claiming 'not enough arguments'

### DIFF
--- a/src/commcare_cloud/commands/deploy.py
+++ b/src/commcare_cloud/commands/deploy.py
@@ -46,6 +46,10 @@ class Deploy(CommandBase):
     )
 
     def modify_parser(self):
+        if len(sys.argv) <= 1:
+            # No environment specified, so no need to add environment-specific repositories
+            return
+
         env_name = sys.argv[1]
         if env_name not in get_available_envs():
             return
--- a/src/commcare_cloud/commcare_cloud.py
+++ b/src/commcare_cloud/commcare_cloud.py
@@ -182,7 +182,13 @@ def call_commcare_cloud(input_argv=sys.argv):
 
     put_virtualenv_bin_on_the_path()
     parser, subparsers, commands = make_command_parser(available_envs=get_available_envs())
-    args, unknown_args = parser.parse_known_args(input_argv[1:])
+
+    raw_args = input_argv[1:]
+    if not raw_args:
+        parser.print_help()
+        return
+
+    args, unknown_args = parser.parse_known_args(raw_args)
 
     if args.control:
         run_on_control_instead(args, input_argv)


src/commcare_cloud/commands/deploy.py
```diff
@@ -46,6 +46,10 @@ class Deploy(CommandBase):
     )
 
     def modify_parser(self):
+        if len(sys.argv) <= 1:
+            # No environment specified, so no need to add environment-specific repositories
```
snopoke, line 50: surely an environment is required?

src/commcare_cloud/commcare_cloud.py
```diff
@@ -182,7 +182,13 @@ def call_commcare_cloud(input_argv=sys.argv):
 
     put_virtualenv_bin_on_the_path()
     parser, subparsers, commands = make_command_parser(available_envs=get_available_envs())
-    args, unknown_args = parser.parse_known_args(input_argv[1:])
+
+    raw_args = input_argv[1:]
+    if not raw_args:
+        parser.print_help()
+        return
```
snopoke, line 189: :+1:

src/commcare_cloud/commands/deploy.py
millerdev: The next line (`sys.argv[1]`) fails if an environment is not specified.

millerdev (approved):

src/commcare_cloud/commands/deploy.py
snopoke: @millerdev is that an answer to my question? With this change what will happen if you don't specify an env?

millerdev: I didn't really understand your first question. With this change you will see a help message rather than a traceback with `IndexError: list index out of range`

mjriley: I could have included more context -- the rationale was that in the process of mentoring, just to see if we had installed `commcare-cloud` correctly, I had the mentee run `cchq` with no args. When it printed out a traceback, it made both of us think that we had installed it incorrectly. Commcare-cloud should handle this situation gracefully.  This line was what was causing the traceback -- if the user specifies no arguments, it's not trying to run anything, it's just trying to figure out what commands are legal and then respond with a 'missing args' message.

snopoke: Thanks. It's just not at all obvious from this code change that returning `None` will result in the help text being shown.

snopoke (approved):

PR #4615, opened 2021-03-27
Prevent cleanup from deleting formplayer log files

dannyroberts:
In rare cases this could delete /opt/formplayer/logs when no logs were written for a while, which could cause issues.

I played around with these commands and I think they work, but we'd need to test it more before deploying to production.


prem-fissionhq:
@dannyroberts I have applied these changes in Staging and India Env. 
Before I have also tried these commands.

millerdev (approved):

PR #4616, opened 2021-03-30
Only ask for github token (not username)

gherceg:
Take 2. A premature merge of [this PR](https://github.com/dimagi/commcare-cloud/pull/4601) resulted in an issue and needed to be reverted. I've pulled in the commits from that PR into the first commit here, and the second commit fixes the issue originally seen (which was the credentials not being asked for).

@snopoke I know the logic for forcing a token was in question during the last PR, and I've changed it back to an `and` statement because I believe that is what we want. You mentioned "The intention is to allow blank tokens unless 'force=true' in which case a token is required" so if force is true, regardless of the value of `GITHUB_TOKEN`, we want to fetch the token from the config file or ask the user again.

### DIFF
--- a/src/commcare_cloud/environment/schemas/meta.py
+++ b/src/commcare_cloud/environment/schemas/meta.py
@@ -17,10 +17,10 @@ def get_github(repo):
     token = None
     if repo.is_private:
         message = "This environment references a private repository: {}".format(repo.url)
-        token = get_github_token(message, required=True)
+        token = get_github_token(message, force=False)
         if not token:
             raise EnvironmentException("Github credentials required")
-    return Github(login_or_token=token, password=None)
+    return Github(login_or_token=token)
 
 
 class GitRepository(jsonobject.JsonObject):
--- a/src/commcare_cloud/environment/schemas/meta.py
+++ b/src/commcare_cloud/environment/schemas/meta.py
@@ -8,20 +8,19 @@ from memoized import memoized_property, memoized
 
 from commcare_cloud.environment.exceptions import EnvironmentException
 from commcare_cloud.environment.secrets.backends import all_secrets_backends_by_name
-from commcare_cloud.environment.secrets.backends.abstract_backend import AbstractSecretsBackend
-from commcare_cloud.fab.utils import get_github_credentials
+from commcare_cloud.fab.utils import get_github_token
 import six
 
 
 @memoized
 def get_github(repo):
-    login_or_token, password = None, None
+    token = None
     if repo.is_private:
         message = "This environment references a private repository: {}".format(repo.url)
-        login_or_token, password = get_github_credentials(message, force=True)
-        if not login_or_token and not password:
+        token = get_github_token(message, required=True)
+        if not token:
             raise EnvironmentException("Github credentials required")
-    return Github(login_or_token=login_or_token, password=password)
+    return Github(login_or_token=token, password=None)
 
 
 class GitRepository(jsonobject.JsonObject):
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/fab/utils.py
```diff
@@ -153,15 +153,15 @@ def diff(self):
...
 
-def _get_github_credentials(message, force=False):
-    global GITHUB_CREDENTIALS
+def _get_github_token(message, force=False):
+    global GITHUB_TOKEN
 
-    if GITHUB_CREDENTIALS is not None:
-        if not force or GITHUB_CREDENTIALS[0]:
-            return GITHUB_CREDENTIALS
+    # return existing token if it exists and caller has not requested a fresh token fetch
```
snopoke, line 162: see main comment for details but a 'fresh token fetch' isn't required here - only a non-empty token.

src/commcare_cloud/fab/utils.py
```diff
@@ -153,15 +153,15 @@ def diff(self):
         return DeployDiff(self.repo, self.last_commit_sha, self.deploy_ref)
 
 
-GITHUB_CREDENTIALS = None
+GITHUB_TOKEN = None
 
 
-def _get_github_credentials(message, force=False):
-    global GITHUB_CREDENTIALS
+def _get_github_token(message, force=False):
```
snopoke, line 159: I think the rename to `required` was better.

snopoke (changes requested):
>@snopoke I know the logic for forcing a token was in question during the last PR, and I've changed it back to an and statement because I believe that is what we want. You mentioned "The intention is to allow blank tokens unless 'force=true' in which case a token is required" so if force is true, regardless of the value of GITHUB_TOKEN, we want to fetch the token from the config file or ask the user again.

I liked the name change from the last PR from `force` to `required`. If the user has already supplied a token either via user input or the config file then there is no need to ask again and we can just return the same token. So you should change it back to an `or`. There is another subtlety you need to account for which is that the user may enter a blank token (press enter at the prompt without entering anything).

```python
if GITHUB_TOKEN is not None:
    # this implies the user has given a token already (though the token may be an empty string)
   if not required or GITHUB_TOKEN:
        return GITHUB_TOKEN  # return the token as long as it's not 'empty' or is not required
```

src/commcare_cloud/environment/schemas/meta.py
```diff
@@ -8,20 +8,19 @@
...
 
 @memoized
 def get_github(repo):
-    login_or_token, password = None, None
+    token = None
     if repo.is_private:
         message = "This environment references a private repository: {}".format(repo.url)
-        login_or_token, password = get_github_credentials(message, force=True)
-        if not login_or_token and not password:
+        token = get_github_token(message, force=False)
```
gherceg, line 20: Just realized I switched this.

src/commcare_cloud/fab/utils.py
```diff
@@ -173,34 +173,32 @@ def _get_github_credentials(message, force=False):
...
+def get_github_token(message=None, required=False):
     if not message:
         message = "This deploy script uses the Github API to display a summary of changes to be deployed."
         if env.tag_deploy_commits:
             message += (
                 "\nYou're deploying an environment which uses release tags. "
                 "Provide Github auth details to enable release tags."
             )
-    return _get_github_credentials(message, force)
+    return _get_github_token(message, allow_empty)
```
snopoke, line 190:
```suggestion
    return _get_github_token(message, required)
```

snopoke (approved):

millerdev:
Thanks for sticking with this.

PR #4618, opened 2021-03-30
Trigger QA tests on successful staging deploy

millerdev:
The tests live at https://github.com/dimagi/dimagi-qa

FYI @kiran-tripathy 

The way this turned out anyone hosting HQ can setup their own automation repo on github (or theoretically some other service with a similar API) and have successful deploy events triggered there. To configure your environment to notify on successful deploy you need to set two things:

1. Set `deploy_event_url` in your environment's _fab-settings.yml_
2. Create a `deploy_event_token` entry in your secrets backend. This can be configured with `cchq <env> secrets edit deploy_event_token`.

##### ENVIRONMENTS AFFECTED
successful staging deploys will trigger this event


gherceg (approved):
Probably best to have Danny officially sign off, but code looks good to me. I think we also want to keep this [confluence page](https://confluence.dimagi.com/display/internal/SaaS+Env+Secrets+and+Accounts) up to date if you don't mind adding this new secret.

millerdev:
Thanks @gherceg. Updated the confluence page.

src/commcare_cloud/fab/operations/release.py
```diff
@@ -368,6 +371,30 @@ def record_successful_deploy():
...
 
+@roles(ROLES_DEPLOY)
+def publish_dimagi_qa_event(name):
+    if env.deploy_env != "production":
+        return
+    environment = get_environment(env.deploy_env)
+    github_token = environment.get_secret("dimagi_qa_github_token")
+    if not github_token:
+        print(red(f"skipping dimagi-qa {name} event: "
+            "dimagi_qa_github_token secret not set"))
```
dannyroberts, line 382: Is this message is appropriate to show (1) on staging and (2) on third party environments? Especially for third party environments it implies they're doing something wrong, but I think we expect only dimagi to be setting up this automated QA hook, right?

dannyroberts (approved):
Had one question about third parties, but overall this looks good

src/commcare_cloud/fab/operations/release.py
millerdev: This will only be shown if the check above passes (currently `if env.deploy_env != "production"`). I'm going to change that though to make the above check based on a variable set in the environment. Just need to figure out how to do that with fabric.

millerdev:
Tested on staging, it works! :tada: 

First github action triggered via staging deploy: https://github.com/dimagi/dimagi-qa/runs/2239966609?check_suite_focus=true

dannyroberts (approved):
Nice I like the more generic wording

src/commcare_cloud/fab/operations/release.py
```diff
@@ -368,6 +372,31 @@ def record_successful_deploy():
         })
 
 
+@roles(ROLES_DEPLOY)
```
snopoke, line 375: i don't think this is necessary since it's just being called as a python function.

millerdev: I was mostly blindly matching what other nearby/similar functions are doing. Does the same apply to `record_successful_deploy` and `record_successful_release` or are they somehow different?

snopoke: `record_successful_deploy` and `record_successful_release` are executed as tasks so they need the decorator to know which hosts to run on.

millerdev: :+1: I think I'm going be doing a bit of other related work for the QA team, so will fix it when I do that.

PR #4619, opened 2021-03-30
Lower repeat_record queue concurrency to 100

gherceg:
One of the changes made for [investigating check_repeaters impact on couch](https://dimagi-dev.atlassian.net/browse/SAAS-12011).

Celery is much faster at processing repeat records (`process_repeat_record` task) than queuing them (`check_repeaters` task). The thought here is to shift the bottleneck from couch  to the `process_repeat_record` task. By reducing the concurrency of the repeat_record queue, there's a chance repeat records are processed slowly enough to not overwhelm couch. The side effect would be an additional delay for when repeat records are delivered, which is a much smaller risk/blast radius than overloading couch. 

We will keep an eye on the impact of this PR once merged.


dannyroberts (approved):

kaapstorm (approved):
Curious to find out how this turns out ... and how we go about finding the right number of workers when we switch to SQL.

PR #4625, opened 2021-04-01
updated update-users command and deprecating update-user-key

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-11294

OLD PR: https://github.com/dimagi/commcare-cloud/pull/4261

##### ENVIRONMENTS AFFECTED

Update-user-key:
deprecating `cchq <env> update-user-key <username>` command. 

Update-users: 
`cchq <env> update-users`
enabled `exclusive:yes` attribute to the authorized_key module while adding ssh_keys into ansible user's authorized_key file. 
This is to remove all retried users ssh_keys and only active ssh_keys are having given access to ansible user. 

User_input section:
Adding user_input section where are asking to enter the valid `ssh_username` (ex: pkasha is my ssh_username) of the user running the update-user command. Using this we are making sure that the above given ssh_user's ssh_key is present in the new ssh_pubkey_list. If given ssh_username is not in dev_users.present/empty/invalid we are failing the task and not updating the ansible authorized_key file. 

I have tested user_input section in following ways:
1. deploy_stack --first-time (user_input is skipped).
2. deploy_stack --skip-check (user_input is prompted, user key is validated, all keys are inserted/added properly if there is any change in user's ssh_keys). 
3. tags=bootstrap_users/users. 
    - user_input is prompted. 
4. update-users command with new/updated keys
    - user_input is prompted. Validation is happening as expected. ansible's authorized_key file data is cleared and replaced
5. update-users command with no key updates or user additions. 
    - user_input is prompted. Validation is happening as expected. ansible's authorized_key file data is not cleared or replaced.   No changes detected.
6. tested with invalid ssh_username  ( means user who is not present under dev_users.present ).
image

@dannyroberts @millerdev Please review and let me know if any changes are necessary or if im missing anything.

### DIFF
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/defaults/main.yml
@@ -3,4 +3,4 @@ bash_PS1: |
   '\[\e]0;({{ env_monitoring_id }}) \u@{{ alt_hostname|default('\h') }}: \w\a\]\[\033[01;33m\]({{ env_monitoring_id }})\[\033[00m\] \[\033[01;32m\]\u@{{ alt_hostname|default('\h') }}\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
 cchq_uid: null
 cchq_gid: null
-ssh_user: ansible
+ssh_user: null
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/defaults/main.yml
@@ -3,3 +3,4 @@ bash_PS1: |
   '\[\e]0;({{ env_monitoring_id }}) \u@{{ alt_hostname|default('\h') }}: \w\a\]\[\033[01;33m\]({{ env_monitoring_id }})\[\033[00m\] \[\033[01;32m\]\u@{{ alt_hostname|default('\h') }}\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
 cchq_uid: null
 cchq_gid: null
+ssh_user: ansible
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -82,7 +82,7 @@
 - name: "Verifying ssh_user key in the ssh_pubkeys_string list"
   assert:
     that:
-      - '"{{ lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }}" in ssh_pubkeys_string'
+      - 'lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }} in ssh_pubkeys_string'
   when: ssh_user is not none
   tags:
     - validate_key
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -64,10 +64,51 @@
   tags:
     - new_ansible_sudo_pass
 
+- name: lookup ssh pubkeys from authorized_keys_dir and create ssh_pubkeys_list
+  set_fact:
+    ssh_pubkeys_list: "{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}"
+  with_items:
+    "{{ dev_users.present }}"
+  register: ssh_pubkeys_results_list
+  tags:
+    - pubkey
+
+- name: iterate over ssh_pubkeys_list and join into a string
+  set_fact:
+    ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
+  tags:
+    - pubkey
+
+- name: capturing username
+  pause:
+    prompt: "Please enter your Dimagi username using which you SSH into servers."
+  register: username_input
+  tags:
+    - capture-userinfo
+    - pubkey
+- debug:
+    msg: "Entered username: {{ username_input.user_input }}"
+  tags:
+    - capture-userinfo
+    - pubkey
+
+- fail: msg="The given username is not defined or empty or not a valid dev_user"
+  when: (username_input.user_input == "" ) or (username_input.user_input|length == 0) or (username_input.user_input not in "{{ dev_users.present }}")
+  tags:
+    - capture-userinfo
+    - pubkey
+
+- name: "Verifying {{ username_input.user_input }} user key in the ssh_pubkeys_string list"
+  assert:
+    that:
+      - '"{{ lookup("file", authorized_keys_dir ~ username_input.user_input ~ ".pub" ) }}" in ssh_pubkeys_string'
+  tags:
+    - capture-userinfo
+    - pubkey
+
 - name: Authorize current devs to login as ansible
   become: yes
-  authorized_key: user=ansible state=present key="{{ lookup('file', authorized_keys_dir ~ item ~ '.pub') }}"
-  with_items: '{{ dev_users.present }}'
+  authorized_key: user=ansible state=present key="{{ ssh_pubkeys_string }}" exclusive=yes
   tags:
     - pubkey
 
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -83,7 +83,7 @@
   assert:
     that:
       - 'lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub") in ssh_pubkeys_string'
-  when: ssh_user is not none
+  when: (ssh_user)
   tags:
     - validate_key
     - pubkey
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -79,31 +79,13 @@
   tags:
     - pubkey
 
-- name: capturing username
-  pause:
-    prompt: "Please enter your Dimagi username using which you SSH into servers."
-  register: username_input
-  tags:
-    - capture-userinfo
-    - pubkey
-- debug:
-    msg: "Entered username: {{ username_input.user_input }}"
-  tags:
-    - capture-userinfo
-    - pubkey
-
-- fail: msg="The given username is not defined or empty or not a valid dev_user"
-  when: (username_input.user_input == "" ) or (username_input.user_input|length == 0) or (username_input.user_input not in "{{ dev_users.present }}")
-  tags:
-    - capture-userinfo
-    - pubkey
-
-- name: "Verifying {{ username_input.user_input }} user key in the ssh_pubkeys_string list"
+- name: "Verifying ssh_user key in the ssh_pubkeys_string list"
   assert:
     that:
-      - '"{{ lookup("file", authorized_keys_dir ~ username_input.user_input ~ ".pub" ) }}" in ssh_pubkeys_string'
+      - '"{{ lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }}" in ssh_pubkeys_string'
+  when: "'ansible' not in ssh_user"
   tags:
-    - capture-userinfo
+    - validate_key
     - pubkey
 
 - name: Authorize current devs to login as ansible
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -82,7 +82,7 @@
 - name: "Verifying ssh_user key in the ssh_pubkeys_string list"
   assert:
     that:
-      - 'lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }} in ssh_pubkeys_string'
+      - 'lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub") in ssh_pubkeys_string'
   when: ssh_user is not none
   tags:
     - validate_key
--- a/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
@@ -83,7 +83,7 @@
   assert:
     that:
       - '"{{ lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }}" in ssh_pubkeys_string'
-  when: "'ansible' not in ssh_user"
+  when: ssh_user is not none
   tags:
     - validate_key
     - pubkey
--- a/src/commcare_cloud/cli_utils.py
+++ b/src/commcare_cloud/cli_utils.py
@@ -12,6 +12,10 @@ from commcare_cloud.colors import color_code, color_error
 
 from .environment.paths import ANSIBLE_DIR
 from .environment.main import get_environment
+from .commands.terraform.aws import (
+    get_default_username,
+    print_help_message_about_the_commcare_cloud_default_username_env_var,
+)
 
 
 def ask(message, strict=False, quiet=False):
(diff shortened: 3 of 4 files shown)


src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -345,12 +345,7 @@ class UpdateUserPublicKey(_AnsiblePlaybookAlias):
     )
 
     def run(self, args, unknown_args):
-        args.playbook = 'deploy_stack.yml'
-        unknown_args += (
-            '--tags=pubkey',
-            '--extra-vars={{"dev_users": {{"present": [{}]}}}}'.format(args.username),
-        )
-        return AnsiblePlaybook(self.parser).run(args, unknown_args)
+        return "update-user-key command is deprecated. Use update-users command."
```
dannyroberts, line 348: I think the return value is supposed to be the process return code (like 0 for success, 1 or some other number for error). I think you can have it print this message. Here's an example of a similar place we do this https://github.com/dimagi/commcare-cloud/blob/47c5f6ee16c850996c48f438dd42b8e760786d7b/src/commcare_cloud/commands/ansible/run_module.py#L175-L177 (this will print it as a yellow "notice").

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
```diff
@@ -64,10 +64,51 @@
...
+    prompt: "Please enter your Dimagi username using which you SSH into servers."
+  register: username_input
+  tags:
+    - capture-userinfo
+    - pubkey
+- debug:
+    msg: "Entered username: {{ username_input.user_input }}"
+  tags:
+    - capture-userinfo
+    - pubkey
```
millerdev, line 93: @dannyroberts I think this is related to https://github.com/dimagi/commcare-cloud/pull/4261#discussion_r605027191, is that right? To me this seems to violate the typical run behavior of ansible in a surprising and non-intuitive way. Except at the beginning when prompting for sudo password, for example, ansible does not usually stop in the middle of a run to prompt for input (and I'd be annoyed if it did). I think if we really need this information we should prompt for it in commcare-cloud before running ansible and/or provide a way for developers to put this information in a file similar to `./src/commcare_cloud/fab/config.py` to avoid the prompt.

I'm not convinced that the risk is even worth it though. We do a "check mode" run first by default to prevent mistakes like that. I don't think it's worth adding more inconvenience, not to mention code to maintain, to avoid a situation that is rarely a concern. If someone did accidentally remove their own key it would be easy enough to get another developer to help them add it again.

dannyroberts: Yeah it's related to that comment. @millerdev I agree with what you said about the awkwardness of it. In terms of risk, the outcome we want to avoid isn't so much someone removing just themselves (though that would be mildly annoying), it's more someone removing _everyone_ including themselves, or a large chunk of people including themselves, accidentally. We could still use SSM (on AWS) to get access, but with our current tools it would be a very manual/awkward process. It would require human error for sure, but I still think it's foreseeable enough and has bad enough an outcome that it's worth guarding against. I'm not married to this particular _way_ of guarding against it though.

Since we're already skipping the prompt for `deploy-stack --first-time`, and the only other place this likely to run is `update-users`, I agree with the feedback that we should prompt for this from the python code in `update-users` at the beginning rather than here in the middle of the ansible script.

prem-fissionhq: Thanks, @dannyroberts @millerdev for your suggestions. I will work on taking input from python and passing it to ansible.

src/commcare_cloud/commands/ansible/ansible_playbook.py
prem-fissionhq: Noted Danny. Will update in my next commit. thank you.

src/commcare_cloud/ansible/roles/bootstrap-users/defaults/main.yml
```diff
@@ -3,3 +3,4 @@ bash_PS1: |
   '\[\e]0;({{ env_monitoring_id }}) \u@{{ alt_hostname|default('\h') }}: \w\a\]\[\033[01;33m\]({{ env_monitoring_id }})\[\033[00m\] \[\033[01;32m\]\u@{{ alt_hostname|default('\h') }}\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
 cchq_uid: null
 cchq_gid: null
+ssh_user: ansible
```
prem-fissionhq, line 6: @dannyroberts Need some suggestions here. Could you please help? 
I ran out of ideas and mentioned default user as `ansible` and skipping task if it ssh_user value 'ansible'. I was facing error 'user (variable) is not defined'

dannyroberts: Can you explain some more? I don't think I'm following

src/commcare_cloud/cli_utils.py
```diff
@@ -33,6 +34,16 @@ def ask_option(message, options, acceptable_responses=None):
...
+def ask_string(message, args):
+    environment = get_environment(args.env_name)
+    if message not in environment.users_config.dev_users.present:
+        allowed_users = environment.users_config.dev_users.present
+        puts(color_error(
+            "Unauthorized user {}.\n\n"
+            "Please pass in one of the allowed ssh users:{}"
+            .format(message, '\n  - '.join([''] + allowed_users))))
+        return False
+    return True
```
millerdev, line 46: This should be named less generically. Also, rather than passing in `args`, should pass in `env_name` since this context does not need `args` for anything else, and indeed does not even know what it is. Finally, it should return a username rather than a boolean value.
```suggestion
def get_dev_username(env_name):
    environment = get_environment(env_name)
    username = default_username = get_default_username()
    while True:
        if not username or default_username.is_guess:
            username = input(f"Enter your SSH username ({default_username}): ")
            if not username:
                username = default_username
        if username in environment.users_config.dev_users.present:
            break
        allowed_users = environment.users_config.dev_users.present
        env_users = '\n  - '.join([''] + allowed_users)
        puts(color_error(
            f"Unauthorized user {username}.\n\n"
            f"Please pass in one of the allowed ssh users:{env_users}"
        ))
        username = ""
    if default_username.is_guess:
        print_help_message_about_the_commcare_cloud_default_username_env_var(username)
    return username
```

millerdev (changes requested):
How long does it take to run `update-users` on production with these changes?

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
```diff
@@ -64,10 +64,33 @@
...
+  set_fact:
+    ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
+  tags:
+    - pubkey
+
+- name: "Verifying ssh_user key in the ssh_pubkeys_string list"
+  assert:
+    that:
+      - '"{{ lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }}" in ssh_pubkeys_string'
+  when: "'ansible' not in ssh_user"
```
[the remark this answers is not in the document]
prem-fissionhq: when we execute the `cchq env deploy-stack --tags=users/bootstrap-users` command the ansible is throwing an error stating that the variable undefined. It is directly calling to playbooks so the value is undefined. 
- Therefore I had to provide the default value to the ssh_user variable, I tried with `None` and `empty` but I missed the `null` option.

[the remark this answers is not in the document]
prem-fissionhq: I believe that Null will work for my case.

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
```diff
@@ -64,10 +64,33 @@
...
+- name: iterate over ssh_pubkeys_list and join into a string
+  set_fact:
+    ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
+  tags:
+    - pubkey
+
+- name: "Verifying ssh_user key in the ssh_pubkeys_string list"
+  assert:
+    that:
+      - '"{{ lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }}" in ssh_pubkeys_string'
```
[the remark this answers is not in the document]
prem-fissionhq: It didn't break in my tests but I will update it.

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -332,8 +332,11 @@ class UpdateUsers(_AnsiblePlaybookAlias):
     """
 
     def run(self, args, unknown_args):
+        username = input('Enter ssh username: ')
+        if not ask_string(username, args):
+            username = ""
         args.playbook = 'deploy_stack.yml'
-        unknown_args += ('--tags=users',)
+        unknown_args += ('--tags=users', '-e ssh_user="'+username+'" ',)
```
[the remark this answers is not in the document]
prem-fissionhq: thanks, Daniel. I was not aware of this. I will update it and going forward I will use this way.

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -345,13 +348,10 @@ class UpdateUserPublicKey(_AnsiblePlaybookAlias):
...
-        unknown_args += (
-            '--tags=pubkey',
-            '--extra-vars={{"dev_users": {{"present": [{}]}}}}'.format(args.username),
-        )
-        return AnsiblePlaybook(self.parser).run(args, unknown_args)
-
+        puts(color_notice(
+                "We have deprecated the update-user-key command. Please use update-users command to modify user's access.\n"
+                "Where you will be asked for your ssh_username which you use to login into servers."))
+        return 0 # exit code
```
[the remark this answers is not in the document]
prem-fissionhq: Sure, I will update it accrodingly. thank you.

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -332,8 +332,11 @@ class UpdateUsers(_AnsiblePlaybookAlias):
     """
 
     def run(self, args, unknown_args):
+        username = input('Enter ssh username: ')
+        if not ask_string(username, args):
+            username = ""
```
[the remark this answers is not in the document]
prem-fissionhq: Will try this Daniel. thanks

src/commcare_cloud/cli_utils.py
prem-fissionhq: Got it :-)

src/commcare_cloud/cli_utils.py
```diff
@@ -11,6 +11,7 @@
 from commcare_cloud.colors import color_code, color_error
 
 from .environment.paths import ANSIBLE_DIR
+from .environment.main import get_environment
```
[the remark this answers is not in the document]
prem-fissionhq: this was giving us a circular import issue so moved import statements inside the method.

src/commcare_cloud/ansible/roles/bootstrap-users/defaults/main.yml
prem-fissionhq: this worked when I used `ssh_user is not none` in the playbook. Tested and pushed changes.

prem-fissionhq:
@millerdev As per your suggestions, I have updated the files and tested and it worked fine for me. Please review and let me know. 
@dannyroberts

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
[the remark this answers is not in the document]
dannyroberts: Can this be changed to simply

```
- 'lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) in ssh_pubkeys_string'
```
?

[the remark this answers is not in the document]
prem-fissionhq: Yes, Danny Updated.

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
```diff
@@ -83,7 +83,7 @@
   assert:
     that:
       - '"{{ lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }}" in ssh_pubkeys_string'
-  when: "'ansible' not in ssh_user"
+  when: ssh_user is not none
```
millerdev, line 86: Does this actually work? I would expect that "none" must be capitalized (`None`) to be valid Python code. If it did not throw an error then maybe Ansible is doing something funny here (expression evaluates to false due to `NameError` or something like that), possibly causing the task to always be skipped?

src/commcare_cloud/cli_utils.py
```diff
@@ -33,6 +34,30 @@ def ask_option(message, options, acceptable_responses=None):
         r = input('Please enter one of {} :'.format(', '.join(options)))
     return r
 
+def get_dev_username(env_name):
+    from .commands.terraform.aws import (
+    get_default_username,
+    print_help_message_about_the_commcare_cloud_default_username_env_var,)
```
millerdev, line 40: nit
```suggestion
    from .commands.terraform.aws import (
        get_default_username,
        print_help_message_about_the_commcare_cloud_default_username_env_var,
    )
```

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -314,7 +314,7 @@ def run(self, args, unknown_args):
         args.playbook = 'deploy_stack.yml'
         args.use_factory_auth = True
         public_vars = environment.public_vars
-        unknown_args += ('--tags=bootstrap-users',) + get_user_arg(public_vars, unknown_args, use_factory_auth=True)
+        unknown_args += ('--tags=bootstrap-users', '--skip-tags=validate_key', ) + get_user_arg(public_vars, unknown_args, use_factory_auth=True)
```
millerdev, line 317: nit
```suggestion
        unknown_args += ('--tags=bootstrap-users', '--skip-tags=validate_key') + get_user_arg(public_vars, unknown_args, use_factory_auth=True)
```

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
```diff
@@ -64,10 +64,33 @@
...
+- name: iterate over ssh_pubkeys_list and join into a string
+  set_fact:
+    ssh_pubkeys_string: "{{ ssh_pubkeys_results_list.results | map(attribute='ansible_facts.ssh_pubkeys_list') | list | join('\n') }}"
+  tags:
+    - pubkey
+
+- name: "Verifying ssh_user key in the ssh_pubkeys_string list"
+  assert:
+    that:
+      - 'lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) }} in ssh_pubkeys_string'
```
millerdev, line 85: This has a syntax error
```suggestion
      - 'lookup("file", authorized_keys_dir ~ ssh_user ~ ".pub" ) in ssh_pubkeys_string'
```

millerdev (changes requested):

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
prem-fissionhq: @millerdev we are seeing this error when used `None` and it's not skipping every time, with `update-users` it is working as expected. 

when tags=bootstrap-users/users
```
TASK [bootstrap-users : Verifying ssh_user key in the ssh_pubkeys_string list] *********************************************************************************************************************************
task path: /tmp/commcare-cloud/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml:82
fatal: [127.0.0.1]: FAILED! => {"msg": "The conditional check 'ssh_user is not None' failed. The error was: template error while templating string: no test named 'None'. String: {% if ssh_user is not None %} True {% else %} False {% endif %}\n\nThe error appears to be in '/tmp/commcare-cloud/src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml': line 82, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Verifying ssh_user key in the ssh_pubkeys_string list\"\n  ^ here\n"}
```

millerdev: Is there an error with `none` spelling? If not, what is the output of the task?

src/commcare_cloud/commands/ansible/ansible_playbook.py
prem-fissionhq: Sure. Corrected. Previously, `unknown_args += ('--tags=bootstrap-users',)` so I followed the same thinking that it might have to end like that.

src/commcare_cloud/ansible/roles/bootstrap-users/tasks/main.yml
millerdev: Ansible behavior here is confounding and, to add insult to injury, difficult to debug. After thrashing around for a while I think I found a solution that works
```suggestion
  when: (ssh_user)  # https://medium.com/opsops/wft-bool-filter-in-ansible-e7e2fd7a148f
```

prem-fissionhq: thanks Daniel. I have updated the file.

dannyroberts (approved):
I think this PR is good to go. I can think of other minor tweaks around the edges, like optimizing how we generate `ssh_pubkeys_string` so it doesn't have to happen separately and redundantly on every machine, but as it is this PR is already an improvement.

I'm not worried about the edge cases you raised @prem-fissionhq. When someone is running `deploy-stack --tags=users` or something like that directly, that's rare enough and the other issue is rare enough that them happening together is small enough of a risk that we don't need to spend too much energy on it.

millerdev (approved):
Thanks for your persistence on this @prem-fissionhq

prem-fissionhq:
Thanks, Danny and Daniel for your help and continuous support. I have learned a lot with this ticket.

millerdev:
:tada:

PR #4627, opened 2021-04-01
adding pgbouncer6, pgbouncer7 as targets for pgmain, pgucr and pgsynclog NLB

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

Production

Adding two instances as target machines for pgmain, pgsynclogs and pgucr NLB's. 
This way there is redundancy and we can swap in and out pgbouncer nodes without downtime for those dbs. 

We will be running cchq <env> terraform apply command for specific targets. 
@dannyroberts Please let me know if I have to run update-config as well. 

Thanks.

### DIFF
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -68,6 +68,7 @@ dbs:
     pgbouncer_endpoint: pgmain_nlb
     pgbouncer_hosts:
       - pgbouncer7
+      - pgbouncer6
   form_processing:
     proxy:
       host: pgbouncer3
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -43,6 +43,7 @@ dbs:
     pgbouncer_endpoint: pgmain_nlb
     pgbouncer_hosts:
       - pgbouncer7
+      - pgbouncer6
   formplayer:
     host: rds_pgformplayer0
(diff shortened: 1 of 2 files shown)


dannyroberts:
> We will be running cchq terraform apply command for specific targets. Please let me know if I have to run update-config as well.

You will have to run `cchq production ansible-playbook deploy_postgres.yml --tags=pgbouncer` first to make both machines be configured to serve traffic for all relevant dbs, and then you'll need to apply the terraform changes (with `-target`).

prem-fissionhq:
Sure, Danny. Thanks you.

environments/production/postgresql.yml
```diff
@@ -54,12 +55,14 @@ dbs:
...
     query_stats: True
   synclogs:
     host: rds_pgsynclog0
     pgbouncer_endpoint: pgsynclogs_nlb
     pgbouncer_hosts:
       - pgbouncer6
+      - pgbouncer7
   auditcare:
     host: rds_pgauditcare0
     pgbouncer_endpoint: pgmain_nlb
```
dannyroberts, line 68: I think you want to add `pgbouncer6` here as well, since the auditcare db also uses the pgmain_nlb (which is fine)

dannyroberts (approved):

PR #4633, opened 2021-04-05
Run QA tests on formplayer deploy

millerdev:
Move `publish_deploy_event` to a common place and call it on successful formplayer release.

The first commit is minor lint cleanup.

##### ENVIRONMENTS AFFECTED
Deploy events are currently only enabled on staging

### DIFF
--- a/src/commcare_cloud/commands/deploy.py
+++ b/src/commcare_cloud/commands/deploy.py
@@ -16,6 +16,7 @@ from commcare_cloud.commands.command_base import Argument, CommandBase
 from commcare_cloud.commands.terraform.aws import get_default_username
 from commcare_cloud.environment.main import get_environment
 from commcare_cloud.environment.paths import get_available_envs
+from commcare_cloud.events import publish_deploy_event
 
 
 class Deploy(CommandBase):
@@ -112,7 +113,7 @@ class Deploy(CommandBase):
             )
         rc = run_ansible_playbook_command()
         if rc != 0:
-            return rc
+            return
         rc = commcare_cloud(
             args.env_name, 'run-shell-command', 'formplayer',
             ('supervisorctl reread; '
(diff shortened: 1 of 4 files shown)


src/commcare_cloud/commands/deploy.py
```diff
@@ -112,7 +113,7 @@ def run_ansible_playbook_command():
             )
         rc = run_ansible_playbook_command()
         if rc != 0:
-            return rc
+            return
```
dannyroberts, line 116: My recollection is that these `CommandBase` subclasses are supposed to return the response code that we want the process to exit with. In this case the choice was made that if the ansible deploy exits with a non-zero status, then we want the cchq deploy command to exit with that same status. I'm not sure what behavior returning `None` here results in. Can you fill me in on your thinking behind this change, what prompted it, etc.?

millerdev: I think this function is only called in one place, and the return code is ignored

https://github.com/dimagi/commcare-cloud/blob/b20e3cec40b01d0c86f923d893d80b453debf0b2/src/commcare_cloud/commands/deploy.py#L91

Is there some magic happening somewhere that is calling this in a place I didn't notice?

dannyroberts:
Oh! I see what you're saying. In that case, I would have fixed it by having
the calling function you noted return the value it returns, but clearly
that behavior isn't relied on anywhere for this command, so what you did
here won't break anything.

dannyroberts (approved):

src/commcare_cloud/commands/deploy.py
millerdev: From @dannyroberts response
> I would have fixed it by having the calling function you noted return the value it returns

That call is in an `if` block, so the return value of the `run()` method would be inconsistent, sometimes returning a value and sometimes not depending on the type of deploy.

PR #4638, opened 2021-04-07
Updating efs_file_system terraform module to support Terraform version 12 & higher

rameshganne:
Updating efs_file_system terraform module to support Terraform version 12 & higher

- Verified on Terraform version 12.30
- First-class expressions: Updating expressions from string interpolation to directly as argument values
- Using "splat" operator `file_system_id = aws_efs_file_system.elastic_file_system[count.index].id`
- Updating variable type ex: `type        = bool`


rameshganne:

PR #4639, opened 2021-04-07
updating r53-private-zone-create-update & r53-record-create-update terraform module to support v12

rameshganne:
Updating r53-private-zone-create-update & r53-record-create-update terraform module to support v12.
ref: SAAS-12057


rameshganne:

PR #4640, opened 2021-04-07
updating elasticache-cluster terraform module to support v12 & including multi_az_options attribute

rameshganne:
Updating elasticache-cluster terraform module to support v12 & including multi_az_options attribute.
Ref: SAAS-12055


rameshganne:

PR #4647, opened 2021-04-12
Downsizing number of es nodes in 1a from Production 

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

Production

Removed es_a015, es_a016, es_a017, es_a018, es_a019,  es_a020 nodes from ES prod cluster. We initially added 11 nodes in 1a az and now we are in process of reducing nodes in 1a.


prem-fissionhq:
It is completed. We can close this ticket.

PR #4652, opened 2021-04-13
Added linked_domain_queue

orangejenny:
https://dimagi-dev.atlassian.net/browse/USH-912

As discussed on slack, adding a high-concurrency queue for linked project space tasks. This is primarily for the `push_models` task, which is used infrequently and by a small number of projects, but when it's used, it can spawn 50+ tasks at once and there's a user waiting for everything to complete, so it isn't really a "background" action.

@mkangia I removed `environments/phi-cas/app-processes.yml` but let me know if that's not okay.

##### ENVIRONMENTS AFFECTED
This touches all environments but is only really relevant for production and maybe india/echis, which use linked project spaces but I don't know if they're using the tasks, which only apply to pushing content and grabbing missing multimedia.


environments/production/app-processes.yml
```diff
@@ -35,6 +35,8 @@ celery_processes:
       concurrency: 134
     export_download_queue:
       concurrency: 5
+    linked_domain_queue:
+      concurrency: 20
```
calellowitz, line 39: This is per machine on 6 machines, so this is 120 processes. I suspect you want something like 3 or 4

orangejenny: Indeed, 3 sounds good. 4e29f1256

calellowitz (approved):

orangejenny:
Added back the phi-cas config because tests expect it. Not digging into that today.

mkangia:
Should be okay to remove that environment, phi-cas, entirely, raising a PR.

mkangia:
Added PR to remove PHI-CAS entirely, thank you for notifying @orangejenny

PR #4654, opened 2021-04-14
updating server & logshipping module to support terraform v12 ref: SAAS-12058

rameshganne:
Updating server & logshipping terraform modules to support v12.
Ref: SAAS-12058

**Server module:** added metadata_options & volume-type. It can support for EBS gp3 volume type. 
metadata_options - Customize the metadata options of the instance. It can refer to as Instance Metadata Service Version 2 (IMDSv2) &  IMDSv1.


rameshganne:
I had already included above changes on https://github.com/dimagi/commcare-cloud/pull/4849.

PR #4664, opened 2021-04-16
Add way to exclude couch host from couch proxy

dannyroberts:
e.g.

```
cchq <env> ap deploy_couchdb2.yml --limit couchdb2_proxy --tags=proxy -e couch_excluded_hosts=<couch_host>
```

Tooling created to make restarting couch nodes one at a time without downtime simpler during https://dimagi-dev.atlassian.net/browse/SAAS-12077.

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/templates/site.j2
+++ b/src/commcare_cloud/ansible/roles/nginx/templates/site.j2
@@ -13,7 +13,9 @@ upstream {{ balancer.name }} {
   {% endif %}
 
 {% for host in groups[balancer.hosts] | default(balancer.hosts) %}
+{% if not balancer.excluded_hosts or host not in groups[balancer.excluded_hosts] | default([]) %}
   server {{ host }}:{{ balancer.port }}{% for key, value in (balancer.params | default({})).items() %} {{ key }}={{ value }}{% endfor %};
+{% endif %}
 {% endfor %}
 }
 {% endfor %}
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/nginx/templates/site.j2
```diff
@@ -13,7 +13,9 @@ upstream {{ balancer.name }} {
   {% endif %}
 
 {% for host in groups[balancer.hosts] | default(balancer.hosts) %}
+{% if not balancer.excluded_hosts or host not in groups[balancer.excluded_hosts] | default([]) %}
```
calellowitz, line 16: I understand the intent of this line, but I am not totally sure I understand exactly what is going on. What does `groups[balancer.excluded_hosts]` do? What happens when you pass a list to `groups` like that? Why is it not `if not balancer.excluded_hosts or host not in groups[balancer.excluded_hosts]`?

Nit: i think it would be clearer if `groups[balancer.excluded_hosts] | default([])` was in parentheses. Since defaults typically apply to the full expression and its hard to see what it applies to in this case without rereading it a few times.

calellowitz (approved):
Approved because I understand the intent and it seems correct, but a lot of questions about the actual change, which confuses me a bit.

src/commcare_cloud/ansible/roles/nginx/templates/site.j2
dannyroberts: Hey @calellowitz `balancer.excluded_hosts` is actually, like `balancer.hosts` above it, not a list of hosts but the string name of a single inventory group that includes all the hosts to be excluded. For this variable in practice that will usually be a host group with a single host in it (i.e `couch12`), but it could include multiple hosts of we had a use case for that.

calellowitz: ah, that makes sense.

dannyroberts: I'll follow up separately about | default

PR #4665, opened 2021-04-16
Upgrade couchdb to 3.1.1

skodde:
https://dimagi-dev.atlassian.net/browse/SAAS-12102

Preliminary changes to upgrade couchdb to version 3.1.1

##### ENVIRONMENTS AFFECTED
All


dannyroberts (commented):
Is there a way to support both couch 2.3.1 and 3.1.1  with the same code, and use a `when:` condition on a public.yml variable called `couch_version` or something like that?

I think for the `couchdb2/meta/main.yml` dependencies that would be straightforward, but I'm not sure about the changes to `ansible/requirements.yml`.

I think if it's not possible (or requires a lot of extra testing and/or forking the dependencies somehow) then it's not worth it, but if it's straightforward then it's useful to know that the rollback plan is just changing a variable and applying again.

skodde:
> Is there a way to support both couch 2.3.1 and 3.1.1 with the same code, and use a `when:` condition on a public.yml variable called `couch_version` or something like that?
> 
> I think for the `couchdb2/meta/main.yml` dependencies that would be straightforward, but I'm not sure about the changes to `ansible/requirements.yml`.
> 
> I think if it's not possible (or requires a lot of extra testing and/or forking the dependencies somehow) then it's not worth it, but if it's straightforward then it's useful to know that the rollback plan is just changing a variable and applying again.

It should be implemented in the latest commit, let me know.

src/commcare_cloud/ansible/deploy_cloudwatch_logs.yml
```diff
@@ -49,7 +49,7 @@
             log_stream_name_suffix: "nginx_error"
             should_include: "{{ 'couchdb2_proxy' in group_names }}"
           # couchdb2
-          - file_path: "/usr/local/couchdb2/apache-couchdb-2.3.1/var/log/couchdb.stderr*"
+          - file_path: "/usr/local/couchdb2/couchdb/var/log/couchdb.stderr*"
```
dannyroberts, line 52: Is this a symlink to the other path?

dannyroberts (approved):

src/commcare_cloud/ansible/deploy_cloudwatch_logs.yml
skodde: Correct, set up by the included role.

millerdev (approved):
Looks good, but I'm not familiar with the particulars of upgrading couch, so my review may not be worth much.

proteusvacuum:
I haven't followed this work too closely, but it seems like there aren't any backwards incompatible changes here, right? 
Will there be a changelog entry and forum post for our local hosters once this gets merged?

dannyroberts:
@proteusvacuum I'll make sure a changelog is paired with this PR as followup. This PR itself doesn't require others to upgrade yet, but of course will eventually and the sooner we have that changelog published the sooner we can lock in the removing couch 2.x. https://dimagi-dev.atlassian.net/browse/SAAS-12183

dannyroberts (approved):

PR #4667, opened 2021-04-16
tests for timeago

snopoke:
Follow up from https://github.com/dimagi/commcare-cloud/pull/4623#discussion_r614296909


millerdev (approved):
Thanks

PR #4669, opened 2021-04-16
HQ deploy refactor part 1

snopoke:
[USH-860](https://dimagi-dev.atlassian.net/browse/USH-860)

Prep work for moving code from https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/hqadmin/management/commands/record_deploy_success.py to commcare-cloud (or reusing what's already there.

It's not a 1:1 refactor as there are some changes in ordering of confirmation messages. I've also removed the `preview_deploy` fab task. If we want that functionality I can add it back outside of fabric.


orangejenny:
@esoergel added `preview_deploy` pretty recently and might have thoughts on how useful it has been.

esoergel:
@orangejenny thanks for flagging.  I'm not particularly attached - I think I wanted it once or twice and it was easy enough to add, so I did, figuring it might be useful for people who deploy more regularly.  If no one's using it and it's in the way, fine to kill.

src/commcare_cloud/commands/deploy/commcare.py
```diff
@@ -1,28 +1,48 @@
...
-                   '--set', ','.join(fab_settings),
-                   branch=args.branch, *unknown_args)
+    announce_deploy_start(environment, "CommCare HQ")
+    start = datetime.utcnow()
+    rc = commcare_cloud(
+        environment.name, 'fab', 'deploy_commcare{}'.format(fab_func_args),
+        '--set', ','.join(fab_settings), branch=args.branch, *unknown_args
+    )
+    if rc != 0:
+        return rc
```
millerdev, line 37: Is this return code used? It seems odd that the function would sometimes return a code and sometimes return `None`.
```suggestion
        return
```

src/commcare_cloud/commands/deploy/commcare.py
```diff
@@ -32,19 +52,115 @@ def confirm_deploy(deploy_revs, diffs, args):
...
+            "Did you specify the correct branch using --set code_branch=<branch> ?"
+        )
+
+    message = (
+        "Whoa there bud! You're deploying {code_branch} which happens to be "
+        "the same code as was previously deployed to this environment.\n"
+        "{branch_specific_msg}\n"
+        "Is this intentional?"
+    ).format(code_branch=code_branch, branch_specific_msg=branch_specific_msg)
+    print(message)
```
millerdev, line 132: nit: we're on Python 3 now, so f-strings are fair game.
```suggestion
    print(
        f"Whoa there bud! You're deploying {code_branch} which happens to be "
        f"the same code as was previously deployed to this environment.\n"
        f"{branch_specific_msg}\n"
        f"Is this intentional?"
    )
```

src/commcare_cloud/environment/remote.py
```diff
@@ -0,0 +1,21 @@
...
+    def __init__(self, environment):
+        self.environment = environment
+        # the default 'cchq' is redundant with ansible/group_vars/all.yml
+        self.cchq_user = environment.public_vars.get('cchq_user', 'cchq')
+        self.deploy_env = environment.meta_config.deploy_env
+
+    @memoized_property
+    def code_current(self):
+        return '/home/{cchq_user}/www/{deploy_env}/current'.format(
+            cchq_user=self.cchq_user, deploy_env=self.deploy_env)
```
millerdev, line 17: nit
```suggestion
        return f'/home/{self.cchq_user}/www/{self.deploy_env}/current'
```

src/commcare_cloud/environment/remote.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def release(self, release_name):
+        return '/home/{cchq_user}/www/{deploy_env}/releases/{release}'.format(
+            cchq_user=self.cchq_user, deploy_env=self.deploy_env, release=release_name)
```
millerdev, line 21: nit
```suggestion
        return f'/home/{self.cchq_user}/www/{self.deploy_env}/releases/{self.release}'
```

src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -32,6 +32,12 @@ def url(self):
         """Human-readable diff URL"""
         return "{}/compare/{}...{}".format(self.repo.html_url, self.last_commit, self.deploy_commit)
 
+    @property
+    def deployed_commit_matches_latest_commit(self):
+        if self.last_commit and self.deploy_commit:
+            short, long = sorted([self.last_commit, self.deploy_commit], key=lambda x: len(x))
+            return self.last_commit == self.deploy_commit or long.startswith(short)
```
millerdev, line 39: Is it intentional that this function implicitly returns `None` when the `if` condition fails?
```suggestion
            return self.last_commit == self.deploy_commit or long.startswith(short)
        return False
```

millerdev (approved):
Nice refactor. A lot was moved or changed, and I'm not sure I caught every nuance. Do you have a plan for testing or are you confident enough you're just going to release without testing?

src/commcare_cloud/commands/deploy/commcare.py
snopoke: I think it should return the return code rather than None: https://github.com/dimagi/commcare-cloud/blob/6fc24a584ed2ff82bddd136be6a1526e4c8985cf/src/commcare_cloud/commcare_cloud.py#L206-L208

snopoke:
> Nice refactor. A lot was moved or changed, and I'm not sure I caught every nuance. Do you have a plan for testing or are you confident enough you're just going to release without testing?

I was planning on doing some manual testing prior to release.

snopoke:
Successfully deployed `india`

snopoke:
@millerdev I think this is ready to go now

millerdev (approved):
Thanks for testing it. In addition to India I assume you've also run this on staging?

I wonder how hard it would be to add more automated tests to facilitate better confidence in refactors like this?

snopoke:
> Thanks for testing it. In addition to India I assume you've also run this on staging?

Yes, also tested on staging

> I wonder how hard it would be to add more automated tests to facilitate better confidence in refactors like this?

I'm open to suggestions here. I did consider it but there's quite a lot of indirection and calling other commcare-cloud commands / subprocesses. And also a fair amount of user input checking.

PR #4677, opened 2021-04-21
refacor github api usage

snopoke:
Follow up from https://github.com/dimagi/commcare-cloud/pull/4623#discussion_r609069901

This moves the code out of the `fab` package and into
the base commcare_cloud package.

It also makes use of generic prompts instead of supporting custom
prompts. Furthermore it will only prompt for credentials
if they are required either for private repos or for environments
that perform release tagging.

Since this new utility also includes a permissions check I have removed the `tag_setup_release` which was
only introduced to validate push permissions on the repo for creating the release tag.


src/commcare_cloud/github.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def github_repo(repo_name, repo_is_private=False, require_write_permissions=False):
+    token = None
+    if repo_is_private or require_write_permissions:
+        token = get_github_credentials(repo_name, repo_is_private, require_write_permissions)
+
+    repo = Github(login_or_token=token).get_repo(repo_name)
+    if require_write_permissions and not repo.permissions.push:
+        raise Exception(f"Github token does not have write permissions for '{repo_name}'")
```
esoergel: Nice, I like the explicit permissions check

src/commcare_cloud/github.py
```diff
@@ -0,0 +1,60 @@
...
+        from .config import GITHUB_APIKEY
+    except ImportError:
+        try:
+            from .fab.config import GITHUB_APIKEY
+        except ImportError:
+            pass
+        else:
+            print(color_notice(f"[Deprecation Warning] Config file has moved."))
+            print(color_notice(f"New location is {PROJECT_ROOT}/config.py or else use the "
+                               f"'GITHUB_TOKEN' environment variable."))
```
esoergel: Can you add in a note about where the config file currently is or construct a `mv` command to fix it?  I imagine most people have forgotten where the config file lives.

src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -59,7 +58,7 @@ def get_diff_context(self):
             context["errors"].append("Versions are identical. No changes since last deploy.")
             return context
 
-        if not (github_auth_provided() and self.current_commit and self.deploy_commit):
+        if not (self.current_commit and self.deploy_commit):
```
esoergel, line 61: If auth isn't provided, github restricts you to [60 requests per hour](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting).  Since this fetches each PR individually, there's a high likelihood of hitting this limit.  I'm not sure what that failure mode looks like, but I think the code should either require you to provide auth or (preferably) skip the diff and warn you to add auth if you want to see it.

millerdev (commented):
Approved after @esoergel's concerns are addressed.

src/commcare_cloud/github.py
snopoke: 470b8043e

src/commcare_cloud/fab/deploy_diff.py
snopoke: 504a2aaae

snopoke:
updated

src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -63,9 +64,16 @@ def get_diff_context(self):
             return context
 
         context["compare_url"] = self.url
+
+        if not self.repo.permissions:
+            # using unauthenticated API calls, skip diff creation to avoid hitting rate limits
+            print(color_warning("Diff generation skipped. Supply a Github token to see deploy diffs."))
```
millerdev, line 73: Maybe out of scope for this PR, but would it be possible to provide a github link to the difference? Like the links in the _Deploy History_ section of the [HQ system admin page](https://www.commcarehq.org/hq/admin/system/).

millerdev (approved):

src/commcare_cloud/fab/deploy_diff.py
esoergel: That's the `compare_url` just above this, actually.

esoergel (approved):

src/commcare_cloud/fab/deploy_diff.py
millerdev: Oh, nice. Looks like it's already included.

PR #4685, opened 2021-04-28
Csmit/sc 1071 monolith e2e install test

Charl1996:
See ticket [here](https://dimagi-dev.atlassian.net/browse/SC-1071)

##### ENVIRONMENTS AFFECTED
No environment files changed, just public facing markdown file and ansible task files.


src/commcare_cloud/ansible/roles/couchdb2/defaults/main.yml
```diff
@@ -12,3 +12,4 @@ couchdb_backup_hours: 1
 couchdb_auth_cache_size: 200
 compact_couch_cron: False
 couch_backup_hourly: False
+couchdb_os_process_limit: 1000
```
Charl1996, line 15: This value is based on the production value in _environments/production/public.yml_

src/commcare_cloud/fab/operations/release.py
```diff
@@ -205,7 +205,6 @@ def create_code_dir(full_cluster=True):
     roles_to_use = _get_roles(full_cluster)
 
     @roles(roles_to_use)
-    @parallel
```
Charl1996, line 208: This was causing a fatal issue during first deployment. Removing the decorator and running it sequentially solved it.

docs/setup/new_environment.md
```diff
@@ -59,7 +59,18 @@ This document will walk you through the process of setting up a new monolith ser
...
 
     ``` bash
-    $ sudo apt update && sudo apt install python-pip sshpass
+    $ sudo apt update && sudo apt install python3-pip sshpass
+    ```
+   Now update your pip3 version; you might encounter installation issues otherwise.
+    ``` bash
+    $ sudo -H pip3 install --upgrade pip
+    ```
+   Make python3 your default python version.
```
proteusvacuum, line 68: Why do we need to do this? Doesn't everything run in a virtual env, where you can set the python version? I think changing the default python version can have weird knock-on effects.

docs/setup/new_environment.md
```diff
@@ -207,7 +222,7 @@ Even though we will be running all commands locally, we still need to add the us
 1. Copy the example fab config file:
 
     ``` bash
-    $ cp ~/commcare-cloud/src/commcare_cloud/fab/config.example.py ~/commcare-cloud/src/commcare_cloud/fab/config.py
+    $ cp ~/commcare-cloud/src/commcare_cloud/config.example.py ~/commcare-cloud/src/commcare_cloud/fab/config.py
```
proteusvacuum, line 225: It's weird that the example file wouldn't be in the same directory that it should end up in. I think instead of this, moving the example into `commcare_cloud/fab/` might be better.

docs/setup/new_environment.md
```diff
@@ -59,7 +59,18 @@ This document will walk you through the process of setting up a new monolith ser
 1. Install Required Packages
 
     ``` bash
-    $ sudo apt update && sudo apt install python-pip sshpass
+    $ sudo apt update && sudo apt install python3-pip sshpass
+    ```
+   Now update your pip3 version; you might encounter installation issues otherwise.
+    ``` bash
+    $ sudo -H pip3 install --upgrade pip
```
proteusvacuum, line 66: Curious - why is the `-H` flag necessary here?

proteusvacuum (approved):
A few questions, but looks good! Thanks for working through it :)

docs/setup/new_environment.md
Charl1996: Ok, I'll just move it back in then. I thought it was done so on purpose, but based on Ethan's comment it looks like an oops :)

src/commcare_cloud/fab/operations/release.py
Charl1996: Eish, ok. It might be worth looking into then.
If I remember correctly the issue was related to an ambiguous input or something which resulted from the parallel execution.
I think I'll spend a bit of time seeing how to get around this, even if it means using a different command like you mentioned.

docs/setup/new_environment.md
Charl1996: The -H flag was actually a suggestion by a warning output on the terminal after running it without. It seems like it has to do with pip wanting to modify the .cache dir in $HOME. Also I see people use it on forums, so I just kind of went with it.

docs/setup/new_environment.md
Charl1996: Running `source /commcare-cloud/control/init.sh` resulted in a package distribution error which seemed to be a compatibility issue I solved by using python 3 rather than 2. I then though it might be good to rather default the _python_ command to python3 rather than 2 for any future compatibility issues, so I just defaulted the system. This was, admittedly, a bold move, but I did not run into any issues so I thought I'd leave it at that.

Also, it may be a good idea since support for python 2 was discontinued in January this year.
(Feel free to let me know if you think this was a...less smart move. I can maybe see if there is a workaround then)

docs/setup/new_environment.md
kaapstorm: Interesting! Yeah, `sudo -H` sets the home directory to /root (if you're sudoing as root, and root's home directory is /root).

This would have been useful to know! I use `apt` to manage system-wide Python packages, and `pip` to manage packages for a virtualenv or my own user, so I haven't encountered this with pip. But with `npm` I have often found that root-owned cached files mess with user-owned cached files! Thank you.

docs/setup/new_environment.md
kaapstorm: We're installing on a fresh Ubuntu 18.04 (so 18.04.5 right now) right? Isn't the default already `python3`?

I think it's only going to be `python2` if you're installing Ubuntu 18.04.1 or upgrading from an older Ubuntu. It might be easier to tell users to download an up-to-date Ubuntu 18.04 release. (18.04.5 Server was released on 2020-08-13.)

... Although ... it might be worth confirming that Python 3 is the default regardless, in case they didn't install a recent 18.04.

src/commcare_cloud/fab/operations/release.py
```diff
@@ -229,7 +228,7 @@ def record_successful_release():
         files.append(RELEASE_RECORD, str(env.code_root), use_sudo=True)
 
 
-#TODO make this a nicer task
+# TODO make this a nicer task
```
kaapstorm, line 231: This TODO dates back to 2016, written by a dev who is no longer at Dimagi. I think we should either create a ticket and put it in the next sprint ... or ... preferably ... just delete the TODO cos it ain't gonna happen. :upside_down_face:

Charl1996: I guess if it dates back to 2016 it's not needed, so I'd rather remove the comment.

docs/setup/new_environment.md
Charl1996: That's not a bad point. I was assuming the default would not be python3. I'll maybe add a comment for the user to check it first, otherwise "follow these steps to default it to 3"

docs/setup/new_environment.md
esoergel: I'm confused about which one we're talking about, but the correct location for the file (and as Farid said, the example file) is in the root dir:
https://github.com/dimagi/commcare-cloud/blob/1aba598c3878d539eae70e3b51ebfe8f9a5a146f/src/commcare_cloud/github.py#L37-L42

esoergel: My understanding is that the destination file should also be in the root dir, ie:
```suggestion
    $ cp ~/commcare-cloud/src/commcare_cloud/config.example.py ~/commcare-cloud/src/commcare_cloud/config.py
```
Was this not your experience when going through the setup?

Charl1996: O, right! Sorry I missed you completely for a moment. Mmm...yes that would make 100% sense.

esoergel (approved):

src/commcare_cloud/fab/fabfile.py
```diff
@@ -314,6 +314,9 @@ def _setup_release(keep_days=2, full_cluster=True):
     :param keep_days: The number of days to keep this release before it will be purged
     :param full_cluster: If False, only setup on webworkers[0] where the command will be run
     """
+    if not env.password:
```
Charl1996, line 317: This just ensures the sudo password is populated for any subsequent sudo(...) commands. By explicitly prompting the user already here, no underlying "parallel ambiguity" w.r.t. sudo password inputs should exist.

src/commcare_cloud/ansible/roles/couchdb2/defaults/main.yml
dannyroberts: I think it probably doesn't hurt either way, but I agree there's no reason to change it in general. We're not even convinced it mattered in production, but it almost certainly doesn't matter on other environments, so I think we can continue to omit it.

@Charl1996 was the problem that something errored if this wasn't set? In that case I'd set it to the [couch default of 100](https://docs.couchdb.org/en/stable/config/query-servers.html#query_server_config/os_process_limit).

Charl1996: @dannyroberts yes it complained that the default value was not found. I'll set it to 100 then :+1:

src/commcare_cloud/fab/fabfile.py
biyeun: @Charl1996 what sudo password is it expecting here? not entirely comfortable giving any executable code the carte blanche to my _local_ machine's root password (which was apparently the input it was looking for)

biyeun: ah it looks like if i type anything then it still goes through. can this check also see if the command is being run in a monolith context? It really adds another annoying prompt in the workflow for regular SaaS deploys.

biyeun: Another thought is that it doesn't seem great to be storing ANY passwords in a variable, but rather using the ansible `sudo()` commands properly instead. I think I'm struggling to understand why this check is here at all.

snopoke: +1 - this should not be here. If the machine's sudoers file is set up correctly it should not be necessary

snopoke: https://github.com/dimagi/commcare-cloud/pull/4700

Charl1996: @biyeun I'm not _completely_ satisfied with this approach myself, but found that it works great in monolith context. I think your suggestion w.r.t. checking monolith context is great! I will tinker a bit to add it. 

**So, why the check at all?**
During the monolith install deploy step, the process would be stopped due to fatal error: 

> Fatal error: Needed to prompt for a connection or sudo password (host: <host>), but input would be ambiguous in parallel mode
 
This was due to the **@parallel** decorator decorating a function which simply creates a directory with **sudo()** (see function _create_code_dir_ in _src/commcare_cloud/fab/operations/release.py_). The real issue seems to be the fact that it wants to prompt a sudo password because it does not know it at this point. Apparently when using the **sudo()** command you can store the password in the fabric env so the command won't prompt again, that's why I reasoned to prompt the user first and get the password, so by the time the **sudo()** command executes it already knows the password.

> "ah, it looks like if I type anything then it still goes through" 

This is concerning. When I tested it with a wrong password it failed at the sudo command step creating the directory. It seems there's factors in play in a non-monolith setup which I don't know of. Correct me if I'm wrong, but it doesn't look like this new logic is affecting a non-monolith context apart from the actual prompting...?

Charl1996: @snopoke I had a look at the sudoers file and everything looks in order, so I didn't think it could be that?

**What I looked for**
I basically made sure my user is listed under

> User privilege specification

with 

> ALL=(ALL:ALL) ALL

I might be completely wrong on this as well, this is kind of fresh ground for me...should I be looking for something else?

snopoke: The deploy generally uses the `ansible` user unless otherwise specified. For deploy the ansible user needs sudo permissions to perform actions as the `cchq` user.

If you know that command that was requesting a password you could test it out on the machine:
```
cchq <env> ssh ansible@<machine name>

ansible@X~$ sudo -u cchq <command>
```

Not all commands are allowed so it may be that the command being attempted is not in the sudoers file correctly.

Since this isn't an issue on any of our existing envs I'm sure it has something to do with how the test env is being set.

PR #4687, opened 2021-04-28
Add option to deploy both commcare and formplayer

millerdev:
I thought this had previously been possible. Is this a good idea or will it not work? I think it's supposed to be the default behavior if no component is specified and the environment metadata has `always_deploy_formplayer: true`, but it doesn't look like that option is used by any of our environments.

Motivation: the deploy process now specifies that both HQ and formplayer should be deployed by on-call SaaS developer. Would be nice to be able to kick that off with a single command.

##### ENVIRONMENTS AFFECTED
n/a

### DIFF
--- a/docs/commcare-cloud/commands/index.md
+++ b/docs/commcare-cloud/commands/index.md
@@ -1061,15 +1061,15 @@ Deploy CommCare
 
 ```
 commcare-cloud <env> deploy [--resume] [--skip-record] [--commcare-rev COMMCARE_REV] [--set FAB_SETTINGS]
-                            [{commcare,formplayer,both}]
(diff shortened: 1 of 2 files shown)


dannyroberts:
Yeah `always_deploy_formplayer: true` was added to be used by third party environments I believe

dannyroberts (approved):
From looking at the code, I think this should work

snopoke (commented):
Since this hasn't worked before there's the opportunity to change the way it works if we don't think 'both' is a good option. Personally I think `all` is more standard than `both` but I don't have strong feelings about it.

snopoke (approved):

PR #4690, opened 2021-04-30
Fix API throttling settings and change default values

czue:
These have been wrong for ages.

I also updated the defaults to match [the HQ defaults](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/api/resources/meta.py#L46-L49), so that we don't accidentally change the current behavior.

##### ENVIRONMENTS AFFECTED

All / none.


snopoke (approved):
Not sure why we even have these here if they are the same as the defaults and also not configurable.

czue:
> Not sure why we even have these here if they are the same as the defaults and also not configurable.

Ha, good point. Should we just remove them? I guess one answer would be a form of documentation in case we (or anyone) wants to change them. But not sure that's actually that helpful.

snopoke:
> > Not sure why we even have these here if they are the same as the defaults and also not configurable.
> 
> Ha, good point. Should we just remove them? I guess one answer would be a form of documentation in case we (or anyone) wants to change them. But not sure that's actually that helpful.

Fine to leave it as is for now.

PR #4693, opened 2021-04-30
Only display merged PRs on master

gherceg:
Came across this recently because I had merged 2 PRs via Github into a parent branch that was not master. I went to deploy staging today and was alarmed to see those PRs (the first and the last one below) were about to be deployed, especially since the parent branch hasn't been merged yet. 
Screen Shot 2021-04-30 at 4 56 37 PM

I saw that this code searches for "Merge pull request" in the commit message and is why it picked up these 2 branches as well. This change makes sure the merge commit is actually on master, so now we will only display merged PRs on master.
Screen Shot 2021-04-30 at 4 56 42 PM

### DIFF
--- a/src/commcare_cloud/fab/deploy_diff.py
+++ b/src/commcare_cloud/fab/deploy_diff.py
@@ -102,13 +102,27 @@ class DeployDiff:
             **self.get_diff_context()
         )
 
-    def _get_pr_numbers(self):
+    def _get_commits_on_master_since_last_deploy(self):
+        last_deployed_commit = self.repo.get_commit(self.current_commit)
+        return self.repo.get_commits(sha='master', since=last_deployed_commit.commit.author.date)
+
+    def _get_merge_commits_since_last_deploy(self):
         comparison = self.repo.compare(self.current_commit, self.deploy_commit)
         return [
-            int(re.search(r'Merge pull request #(\d+)',
-                          repo_commit.commit.message).group(1))
+            repo_commit
             for repo_commit in comparison.commits
             if repo_commit.commit.message.startswith('Merge pull request')
+        ]
+
+    def _get_pr_numbers(self):
+        merge_commits_since_last_deploy = self._get_merge_commits_since_last_deploy()
+        master_commits_since_last_deploy = self._get_commits_on_master_since_last_deploy()
+        # only care about merge commits on master
+        return [
+            int(re.search(r'Merge pull request #(\d+)',
+                          merge_commit.commit.message).group(1))
+            for merge_commit in merge_commits_since_last_deploy
+            if merge_commit in master_commits_since_last_deploy
         ]
 
     def _get_pr_info(self, pr_number):


src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -102,15 +102,29 @@ def render_diff(self, template_name):
             **self.get_diff_context()
         )
 
-    def _get_pr_numbers(self):
+    def _get_commits_on_master_since_last_deploy(self):
+        last_deployed_commit = self.repo.get_commit(self.current_commit)
+        return self.repo.get_commits(sha='master', since=last_deployed_commit.commit.author.date)
+
+    def _get_merge_commits_since_last_deploy(self):
         comparison = self.repo.compare(self.current_commit, self.deploy_commit)
```
dannyroberts, line 110: if `current_commit` is the latest commit on `master` and `deploy_commit` is the last deployed commit (do I have that right?) then I don't think I get how this could include things that aren't in the `master` branch's history. Can you explain this to me a bit more?

dannyroberts (approved):
Wow, interesting find

src/commcare_cloud/fab/deploy_diff.py
gherceg: The only explanation I have is the Comparison object picks up all commits in between the two specified commits by date, regardless of branch. As far as I could tell from [the docs](https://pygithub.readthedocs.io/en/latest/github_objects/Repository.html#github.Repository.Repository.compare), there isn't a way to specify a branch when retrieving a comparison between two commits.

snopoke:
> Came across this recently because I had merged 2 PRs via Github into a parent branch that was not master. I went to deploy staging today and was alarmed to see those PRs (the first and the last one below) were about to be deployed, especially since the parent branch hasn't been merged yet.

@gherceg this doesn't make sense to me. You say you deployed staging but staging isn't even on master so it would make sense that you saw PRs that weren't yet merged. Also the staging branch is rebuilt often and force pushed so I don't think the diff from one staging commit to the next is likely to give coherent results. I think it would be better to just exclude diffs for staging.

It's a pity you didn't include the other deploy details, then we could look at the diff in the git repo and see if it was the github API or if that's actually what git gives you. My hunch is that git would have given you the same diff for the two commits.

I'm going to revert this and make the change to not generate diffs for staging.

snopoke:
https://github.com/dimagi/commcare-cloud/pull/4698

gherceg:
@snopoke I randomly stumbled upon this now and am sorry I never saw it back in May. Despite the removal of diffs from staging, I know you are planning to implement better diffing for staging deploys in the future so I figured it is worth continuing this conversation. 

My understanding of this PR was that it impacted all environments, not just staging. In order to determine the list of PRs that have been merged _into master_ since the last deploy, we search for _all_ merge commits since the commit of the last deploy. **The issue is that this search does not factor in which branch the merge commit is on**, meaning this has the potential to display PRs that have been merged into a non-master branch. 

For example, if I have a base branch `gh/base` and a child off of that branch `gh/base/child`, and I merge `gh/base/child` into `gh/base`, the PR associated with `gh/base/child` will show up under the list of PRs since the last deploy. This can be confusing when deploying because this PR is not actually merged into master, but this gives the impression that it has been.

As far as this behavior on staging specifically, you do mention that staging is force pushed often and therefore these may not yield reliable results, but as long as the commit of the last deploy is constant (which it should be?), shouldn't the diff between then and the latest build on staging accurately reflect merge commits since the last deploy? I could be lacking understanding here.

snopoke:
@gherceg
 >  we search for all merge commits since the commit of the last deploy

This is partially true, we search for "all merge commits that are in the list of commits returned by git when comparing the previous deploy commit with the current deploy commit".

This should not include any commits that are not included in the 'current deploy commit'. It would however include merge commits from other branches if the final results has been merged into master, so that is something we could try filter out somehow.

In your example the diff should not include those merges until the main `gh/base` branch is merged into master and then it will also show the merge of the child into the base branch.

If you have an example that contradicts this I'd be interested to see it.

> As far as this behavior on staging specifically, you do mention that staging is force pushed often and therefore these may not yield reliable results, but as long as the commit of the last deploy is constant (which it should be?), shouldn't the diff between then and the latest build on staging accurately reflect merge commits since the last deploy? I could be lacking understanding here.

Since the staging branch is rebuilt each time we deploy by starting with the current master and merging in all the staging branches the diff will always show merge commits for every branch (because the merge commits from the current rebuild will not be the same as the ones from the previous rebuild).

I don't have any plans to implement staging diffs since I don't know of a way to generate anything meaningful but if you can think of something I'd be interested to hear about it.

gherceg:
Ah I see. So in the `gh/base` example, if `gh/base` were included in the staging.yaml file, the merge commit for `gh/base/child` into `gh/base` would show up in the PRs list because the current deploy commit includes that merge. So this is only relevant to the staging workflow.

And I don't have any suggestions for staging diffs. Seems like we'd have to change the way we build the autostaging branch to get any meaningful diffs from it (merge in updates rather than entirely rebuild? But removing a branch that hasn't been merged yet doesn't seem trivial in that case).

PR #4701, opened 2021-05-05
setting up Rabbitmq cluster in multi-az in India env

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-12111
##### ENVIRONMENTS AFFECTED

India 

- Added new rabbitmq node in India env. Not yet joined with the cluster. 
- Replaced rabbitmq repo distribution and signing key with **packagecloud.io** as we observed that `dl.bintray.com` i.e **bintray** stopped their service from May 1st, 2021 and the existing links are giving 403 response. 

Bintray status. 
https://blog.rabbitmq.com/posts/2021/03/migrate-off-of-bintray/
https://status.bintray.com/

package cloud
https://www.rabbitmq.com/install-debian.html#supported-debian-distributions


dannyroberts (approved):

environments/india/public.yml
```diff
@@ -44,6 +44,9 @@ kafka_log_dir: "{{ encrypted_root }}/kafka"
 
 rabbitmq_version: 3.8.5-1
 
+#RabbitMQ Broker URL
+BROKER_URL: 'amqp://{{ AMQP_USER }}:{{ AMQP_PASSWORD }}@{{ AMQP_HOST }}:5672/{{ AMQP_NAME }};amqp://{{ AMQP_USER }}:{{ AMQP_PASSWORD }}@{{ groups.rabbitmq.1 }}:5672/{{ AMQP_NAME }}'
+
```
dannyroberts, line 49: Naively my first thought was you could probably also do this by just switching `kafka1` and `rabbit2` under `[rabbitmq:children]` in the inventory above. Does something else rely on that order that you need to stay the same?

dannyroberts (approved):

environments/india/public.yml
prem-fissionhq: Yes, Danny. The sequence should be the same. Because kafka1 is right now acting as Master node. Please let me know if we are missing anything. thanks

PR #4704, opened 2021-05-06
Deploy formplayer after commcare

millerdev:
Previously `Deploy.run` returned immediately after commcare (HQ) deploy, even if both commcare and formplayer were meant to be deployed.

### DIFF
--- a/src/commcare_cloud/commands/deploy/command.py
+++ b/src/commcare_cloud/commands/deploy/command.py
@@ -90,17 +90,22 @@ class Deploy(CommandBase):
             if environment.meta_config.always_deploy_formplayer:
                 deploy_component.append('formplayer')
 
+        rc = 0
         if 'commcare' in deploy_component:
             if 'formplayer' not in deploy_component:
                 _warn_no_formplayer()
-            return deploy_commcare(environment, args, unknown_args)
+            rc = deploy_commcare(environment, args, unknown_args)
         if 'formplayer' in deploy_component:
             if 'commcare' not in deploy_component:
                 if args.commcare_rev:
                     print(color_warning('--commcare-rev does not apply to a formplayer deploy and will be ignored'))
                 if args.fab_settings:
                     print(color_warning('--set does not apply to a formplayer deploy and will be ignored'))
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/deploy/command.py
```diff
@@ -90,17 +90,22 @@ def run(self, args, unknown_args):
...
-            return deploy_commcare(environment, args, unknown_args)
+            rc = deploy_commcare(environment, args, unknown_args)
         if 'formplayer' in deploy_component:
             if 'commcare' not in deploy_component:
                 if args.commcare_rev:
                     print(color_warning('--commcare-rev does not apply to a formplayer deploy and will be ignored'))
                 if args.fab_settings:
                     print(color_warning('--set does not apply to a formplayer deploy and will be ignored'))
-            return deploy_formplayer(environment, args)
+            if rc:
```
mjriley, line 104: Nit: I think this implicit check favors brevity over readability. Perhaps `rc` is guarantee enough that this is a shell-style return code, but I think it's implicit that 0 indicates success here, and it would be better to just spell it out (not everyone is familiar with the shell convention):

```
SUCCESS = 0
if rc == SUCCESS...
```

src/commcare_cloud/commands/deploy/command.py
```diff
@@ -90,17 +90,22 @@ def run(self, args, unknown_args):
...
+                print(color_error("Skipping formplayer because commcare failed"))
```
mjriley, line 105: Is it necessary to mention that we skipped formplayer when commcare fails? I don't like that this code is coupled to the fact that the commcare build needed to execute right before this -- if that order was switched, or if another component was added in between, then this check would fail. It seems simpler to just return immediately if any stage fails, but then we lose these kind of messages.

mjriley (approved):

src/commcare_cloud/commands/deploy/command.py
mjriley: It might be overkill when we are only supporting two components, but I guess the generic way to handle this would have been to track both the return code as well as the component name, so we could print them together.

PR #4706, opened 2021-05-07
White listed URL

Rohit25negi:
Ticket: https://dimagi-dev.atlassian.net/browse/INDIV-8
WAF was blocking some of the requests from this endpoint. As this endpoint allows XML kind of script to be taken from the user so it needs to be whitelisted.
relevant PR:  https://github.com/dimagi/commcare-hq/pull/29674


dannyroberts (approved):
Assuming that this change came from running `./manage.py list_waf_allow_patterns` from https://github.com/dimagi/commcare-hq/pull/29674, this looks good to me

Rohit25negi:
> Assuming that this change came from running `./manage.py list_waf_allow_patterns` from [dimagi/commcare-hq#29674](https://github.com/dimagi/commcare-hq/pull/29674), this looks good to me

yes, Also tested this on staging. Working fine.

PR #4710, opened 2021-05-07
Use newer cipher set ELBSecurityPolicy-FS-1-2-Res-2020-10 on staging

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-12091

Use newer cipher set `ELBSecurityPolicy-FS-1-2-Res-2020-10` on staging which removes CBC-based ciphers

##### ENVIRONMENTS AFFECTED
staging only


dannyroberts:
I'm going ahead and applying this on staging prior to approval and merge. (Which is okay because it's staging.)

Issue #2410, opened 2018-11-15
Have lower-touch way to permanently specify default user


dannyroberts:
Would be nice to have an environment variable to check for what your ssh username.(`COMMCARE_CLOUD_DEFAULT_USERNAME`?)

This would then be used in `ssh` command by default, as well as other commands that infer the username for ssh or VPN.

millerdev:
SSH already has a way to override usernames in its configuration, so it's not really a problem that needs to be solved for SSH. I have this in ~/.ssh/config:
```
Host *
    User dmiller
```
This is for places in commcare-cloud that use `getpass.getuser()` and provide no way to override that username.

PR #4715, opened 2021-05-12
get username in SSH commands

snopoke:
This is useful for checking the username and also allows users to have different usernames on their local machines from those configured in commcare-cloud

### DIFF
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -8,7 +8,7 @@ import sys
 from clint.textui import puts
 from six.moves import shlex_quote
 
-from commcare_cloud.cli_utils import print_command
+from commcare_cloud.cli_utils import print_command, get_dev_username
 from commcare_cloud.commands.command_base import Argument, CommandBase
 from commcare_cloud.environment.main import get_environment
 from ..ansible.helpers import get_default_ssh_options_as_cmd_parts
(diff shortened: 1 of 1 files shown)


Charl1996 (approved):
Looks plausible to me

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -67,6 +67,9 @@ def run(self, args, ssh_args):
         if ':' in address:
             address, port = address.split(':')
             ssh_args = ['-p', port] + ssh_args
+        if '@' not in address:
+            username = get_dev_username(args.env_name)
+            address = f'{username}@{address}'
```
dannyroberts, line 72: This means rather than using `ssh`'s built-in inference, cchq will use `COMMCARE_CLOUD_DEFAULT_USERNAME` to override that default if it's present, otherwise it will explicitly provide the same default (your current os username) that `ssh` would have come up with on its own?

Just trying to make sure my understanding of the intent/behavior is correct before I approve

dannyroberts: I think my main question is then what happens with anything dev users might currently have in their `~/.ssh/config`. I think I remember @millerdev having thoughts about this in a previous PR from some time ago, but I wasn't able to dig up where that was.

millerdev: Perhaps you were thinking of https://github.com/dimagi/commcare-cloud/issues/2410

snopoke: It is true that the ssh config can be used to overide the username provided you either:
1. know the specific addresses you want to override (or address patterns)
2. want to override it globally as suggested in #2410

I think that's fine to keep that, what about using using `ssh <host> -G | grep 'user '` to check the SSH username and provide a warning / error if it's not in the list of allowed hosts?

PR #4716, opened 2021-05-12
adding pgshard_nlb for form_processing partitions

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

Staging

Configured form_processing partitions to route traffic via NLB with pgbouncer0 as the target node. 
Updated shards configuration to use NLB URL instead of pgproxy2 ip.


dannyroberts (approved):
excellent

PR #4719, opened 2021-05-14
[SSO]  Add `ENFORCE_SSO_LOGIN` to `localsettings` and enable for staging.

biyeun:
Related to this PR https://github.com/dimagi/commcare-hq/pull/29726

##### ENVIRONMENTS AFFECTED
staging


millerdev (approved):
**Not meant to be addressed in this PR**: I wonder if we could introduce a bit more structure in the link between `public.yml` `localsettings` and how the values in it get mapped into the `localsettings.py.j2` template? The boilerplate in that template is annoying.

PR #4724, opened 2021-05-17
Implement Ansible Vault _set_secret function

skodde:
Note that the original YAML format is not preserved (comments, etc.).

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
+++ b/src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
@@ -134,6 +134,8 @@ class AnsibleVaultSecretsBackend(AbstractSecretsBackend):
         return context[var_name]
 
     def _set_secret(self, var, value):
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/environment/secrets/backends/ansible_vault/main.py
```diff
@@ -134,7 +134,12 @@ def _get_secret(self, var_name):
         return context[var_name]
 
     def _set_secret(self, var, value):
-        raise NotImplementedError
```
dannyroberts, line 137: Might be worth leaving a comment here to the same effect as what you wrote in the description, that no attempt was made to preserve yaml formatting or comments. Also that no edge-cases (creating a new file if it doesn't exist, etc.) are handled.

dannyroberts (approved):

dannyroberts (approved):

PR #4728, opened 2021-05-18
Add WAF routes to terraform const.py

Charl1996:
See route annotation PR [here](https://github.com/dimagi/commcare-hq/pull/29753).

##### ENVIRONMENTS AFFECTED
All

This PR contains changes which I did not make but were on CCHQ master when I executed `./manage.py list_waf_allow_patterns` (I pulled master before executing). I added the relevant reviewers for the other changes.

### DIFF
--- a/src/commcare_cloud/commands/terraform/constants.py
+++ b/src/commcare_cloud/commands/terraform/constants.py
@@ -6,6 +6,8 @@
 from __future__ import unicode_literals
 COMMCAREHQ_XML_POST_URLS_REGEX = r"""
 ^/a/([\w\.:-]+)/api/v([\d\.]+)/form/$
+^/a/([\w\.:-]+)/api/v0.6/case/$
+^/a/([\w\.:-]+)/api/v0.6/case/([\w-]+)/$
 ^/a/([\w\.:-]+)/apps/([\w-]+)/multimedia/uploaded/app_logo/([\w\-]+)/$
 ^/a/([\w\.:-]+)/apps/([\w-]+)/multimedia/uploaded/image/$
 ^/a/([\w\.:-]+)/apps/edit_form_attr/([\w-]+)/([\w-]+)/([\w-]+)/$
@@ -17,6 +19,7 @@ COMMCAREHQ_XML_POST_URLS_REGEX = r"""
 ^/a/([\w\.:-]+)/cloudcare/apps/report_formplayer_error
 ^/a/([\w\.:-]+)/data/export/custom/download_data_files/$
 ^/a/([\w\.:-]+)/data_dictionary/import$
+^/a/([\w\.:-]+)/dhis2/map/(\w+)/$
 ^/a/([\w\.:-]+)/fixtures/edit_lookup_tables/upload/$
 ^/a/([\w\.:-]+)/fixtures/fixapi/
 ^/a/([\w\.:-]+)/importer/excel/bulk_upload_api/$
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/terraform/constants.py
```diff
@@ -17,6 +19,7 @@
 ^/a/([\w\.:-]+)/cloudcare/apps/report_formplayer_error
 ^/a/([\w\.:-]+)/data/export/custom/download_data_files/$
 ^/a/([\w\.:-]+)/data_dictionary/import$
+^/a/([\w\.:-]+)/dhis2/map/(\w+)/$
```
Charl1996, line 22: My own change

src/commcare_cloud/commands/terraform/constants.py
```diff
@@ -6,6 +6,8 @@
 from __future__ import unicode_literals
 COMMCAREHQ_XML_POST_URLS_REGEX = r"""
 ^/a/([\w\.:-]+)/api/v([\d\.]+)/form/$
+^/a/([\w\.:-]+)/api/v0.6/case/$
+^/a/([\w\.:-]+)/api/v0.6/case/([\w-]+)/$
```
proteusvacuum, line 10: @esoergel do you know why these weren't included here before? Did you add these routes to the WAF manually?

src/commcare_cloud/commands/terraform/constants.py
```diff
@@ -28,6 +31,7 @@
 ^/a/([\w\.:-]+)/receiver/submission/?$
 ^/a/([\w\.:-]+)/settings/users/commcare/fields/$
 ^/a/([\w\.:-]+)/settings/users/commcare/upload/$
+^/a/([\w\.:-]+)/settings/users/web/upload/$
```
proteusvacuum, line 34: @orangejenny same question as above

src/commcare_cloud/commands/terraform/constants.py
esoergel: Ah, I did not add these to the WAF manually, but I did decorate the view with `@waf_allow('XSS_BODY')`.  Is this a build step that's supposed to be done whenever that decorator is applied?

Charl1996: @esoergel According to [this doc](https://confluence.dimagi.com/display/internal/How+traffic+gets+to+our+system%3A+Global+Accelerator%2C+ALB%2C+and+WAF) (see section _Allow-listing blocked traffic in the WAF_) this `constants.py` file has to be updated accordingly when using the decorator for a route view.

Charl1996: @esoergel  I'm curious, is your view that's being decorated working for whatever reason you decorated it for? If you did not update this file and for argument's sake your changes works fine, I'm wondering if updating this file is technically necessary then?

esoergel: I'm not sure in what circumstances the WAF would block access to an undecorated view, but that view hasn't been released, so it's only been used lightly, in testing.

src/commcare_cloud/commands/terraform/constants.py
kaapstorm: I approve this change. If you would like to merge this before any open questions about other changes are resolved, you could PR just this line, and leave other changes for later.

kaapstorm (approved):

Charl1996:
I extracted my changes to [this PR](https://github.com/dimagi/commcare-cloud/pull/4755). Keeping this one open though.

proteusvacuum:
What's blocking here? I believe if those routes are decorated they should be included here, right? @stephherbers and @esoergel I think we just need you to confirm that you just forgot to add these routes after decorating your views. 
Alternatively, please remove the decorator from your views if you don't think it's necessary.

src/commcare_cloud/commands/terraform/constants.py
orangejenny: Steph's offline, but I'm reasonably confident this should be here. This view was decorated with `waf_allow` in https://github.com/dimagi/commcare-hq/pull/28832/ and it's likely just an oversight that this file wasn't also updated.

esoergel:
I do indeed want this view decorated and did not know I needed to take any steps beyond applying the decorator.  @Charl1996 if you can apply my change with your own, that would be very helpful.

kaapstorm:
:+1: Sounds like this change is correct. Merging.

PR #4734, opened 2021-05-20
Set commits to None on github error

proteusvacuum:
This was causing an issue on eCHIS on deploy as github was not accessible, and the default setting was set to `True`


snopoke (approved):
Thanks

PR #4735, opened 2021-05-20
Fix terraform v0.11 error w/ archive provider

dannyroberts:
This was something @skodde and I noticed locally when running terraform. Removing just this one line fixes it. The error is

> Error: provider.archive: : invalid or unknown key: source


dannyroberts:
@rameshganne @prem-fissionhq FYI

skodde (approved):

PR #4737, opened 2021-05-21
adding control1-production into inventory

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

Production


environments/production/inventory.ini.j2
```diff
@@ -213,9 +213,11 @@ swap_size=8G
 [couchdb2_proxy:children]
 couchproxy3
 
+{{ __control1__ }}
 {{ __control2__ }}
 
 [control:children]
+control1
 control2
```
dannyroberts, line 221: I think you _don't_ want control1 to be in the `[control]` group, because we're going to be using it as a machine from which to run a script, not as the control machine (which we use for deploys etc and which certain commands use intentionally that way)

prem-fissionhq: @dannyroberts sure. thanks for correcting me.

dannyroberts (approved):

PR #4754, opened 2021-05-26
updated EFS-staging transition_to_ia & template file to add multi EFS ref:SAAS-12263

rameshganne:
In the process of v12 validation & testing, I am updating EFS attribute `transition_to_ia`  from staging resource file & removing `{% if not loop.first %}, {% endif %}`  from template file.

### DIFF
--- a/environments/staging/terraform.yml
+++ b/environments/staging/terraform.yml
@@ -429,7 +429,7 @@ r53_private_zone:
 efs_file_systems:
   - efs_name: "efs_staging"
     create: yes
-    transition_to_ia:  "AFTER_14_DAYS"
+    transition_to_ia:  "AFTER_7_DAYS"
     create_access: no
     create_mount: yes
     create_record: yes
--- a/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
+++ b/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
@@ -302,7 +302,7 @@ module "efs_file_system__{{ efs_file_system.efs_name }}" {
 {% endfor %}
 
 {%- for az in az_codes %}
-{%- for efs_file_system_mount in efs_file_systems %}{% if not loop.first %}, {% endif %} 
+{%- for efs_file_system_mount in efs_file_systems %} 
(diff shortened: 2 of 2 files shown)


environments/staging/terraform.yml
```diff
@@ -429,7 +429,7 @@ r53_private_zone:
 efs_file_systems:
   - efs_name: "efs_staging"
     create: yes
-    transition_to_ia:  "AFTER_14_DAYS"
+    transition_to_ia:  "AFTER_7_DAYS"
```
dannyroberts, line 432: Can you explain why this change was necessary for v12?

src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
```diff
@@ -313,7 +313,7 @@ module "efs_mount__{{ efs_file_system_mount.efs_name }}__{{ az }}" {
 {% endfor %}
 {% endfor %}
 
-{%- for efs_file_system_route53 in efs_file_systems %}{% if not loop.first %}, {% endif %}
+{%- for efs_file_system_route53 in efs_file_systems %}
```
dannyroberts, line 316: Can you explain why these changes (removing a comma) were necessary for v12? I'm not sure why they would have ever worked on v11 either 🤔

rameshganne: I had included a new EFS resource to validate module but I could not create it with existing code because adding comma(,) after EFS resource block in commcarehq.tf file. I got know,  `{% if not loop.first %}, {% endif %}` is not required to create multiple EFS resources. To resolve it, I am removing `{% if not loop.first %}, {% endif %}`. 

After changes, I had verified it by adding multiple EFS file-systems.

environments/staging/terraform.yml
rameshganne: This change does not relate to v12(sorry for using `In the process of v12 validation & testing` in the initial comment). In the general practice & to reduce EFS storage price, I am updating the attribute for moving the files which have not been accessed for the last 7 days to the IA storage class.

dannyroberts (approved):

PR #4762, opened 2021-05-28
Remove references to detailed tagging domains

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-12284

Now that feature flags are available in formplayer, detailed tagging domains do not need to be stored in secrets. This PR removes all references to detailed tagging domains, but leaves the detailed tags themselves as those are still applied at this level.


gherceg:
Removed reference from [confluence page](https://confluence.dimagi.com/display/saas/SaaS+Env+Secrets+and+Accounts) too.

snopoke (approved):

PR #4765, opened 2021-06-01
Add CaseSearch only pillow to prod

sravfeyn:
https://dimagi-dev.atlassian.net/browse/SAAS-12291?focusedCommentId=146855

Adding this temporarily to reprocess Case Search pillows as the index was down. The case-search pillow [checkpoint](url) was reset to case-pillow's [Friday checkpoint](https://www.commcarehq.org/admin/hqadmin/historicalpillowcheckpoint/58597/change/?_changelist_filters=q%3Dcase-pillow-hqcases_2016-03-04-case_search_2018-05-29-messaging-sync).

@snopoke @calellowitz

PR #4767, opened 2021-06-01
Revert "Add CaseSearch only pillow to prod"

snopoke:
Reverts dimagi/commcare-cloud#4765

Pillow is now caught up so removing this.


snopoke:
Already rolled out

PR #4774, opened 2021-06-02
fixing reroute example command which was not ended correctly 

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-12298
##### ENVIRONMENTS AFFECTED

The reroute example command was not correctly ended with brackets and quotes. Fixed them and added another example. 

@shyamkumarlchauhan @dannyroberts Thanks

### DIFF
--- a/docs/firefighting/index.md
+++ b/docs/firefighting/index.md
@@ -827,7 +827,7 @@ Currently on ICDS (maybe on prod/india) shard allocation is disabled. In case a 
   - Shard assignment can be checked via Elasticsearch [shards API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-shards.html) or the shards graph on Elasticsearch datadog dashboard
 - If any primaries are not allocated. Use rereoute API ([official docs](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/cluster-reroute.html))
   - Reroute according to existing shard allocation
-  - The rerouting of unassigned primary shards will cause data loss (w.r.t es_2.4.6).
+  - The rerouting of unassigned primary shards will cause data loss (w.r.t es_2.4.6). <br>
   :warning: The <b>allow_primary</b> parameter will force a new empty primary shard to be allocated without any data. If a node which has a copy of the original shard (including data) rejoins the cluster later on, that data will be deleted: the old shard copy will be replaced by the new live shard copy.
 
   - Example reroute command to allocate replica shard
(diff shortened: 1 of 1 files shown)


shyamkumarlchauhan (approved):

docs/firefighting/index.md
```diff
@@ -829,11 +829,23 @@ Currently on ICDS (maybe on prod/india) shard allocation is disabled. In case a
...
+    ```
+    curl -XPOST 'http://<es_url>:<port>/_cluster/reroute' -d '{
+    "commands" : [
+        {
+          "allocate" : {
+              "index" : "case_search_2018-05-29", "shard" : 0, "node" : "es-b013-production", "allow_primary": true
+             }
+          }
+        ]
+      }'
```
dannyroberts, line 848: This is actually an example of "allocate primary shard" and should go under the previous bullet point, right?

dannyroberts (approved):

docs/firefighting/index.md
```diff
@@ -827,14 +827,29 @@ Currently on ICDS (maybe on prod/india) shard allocation is disabled. In case a
   - Shard assignment can be checked via Elasticsearch [shards API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat-shards.html) or the shards graph on Elasticsearch datadog dashboard
 - If any primaries are not allocated. Use rereoute API ([official docs](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/cluster-reroute.html))
   - Reroute according to existing shard allocation
-  - Example reroute command to allocate primary shard
+  - Example reroute command to allocate primary shard <br>
+    :warning: The <b>allow_primary</b> parameter will force a new empty primary shard to be allocated without any data. If a node which has a copy of the original shard (including data) rejoins the cluster later on, that data will be deleted: the old shard copy will be replaced by the new live shard copy.
```
dannyroberts, line 831: I think because of this, we basically never want to use allow_primary = true. Can we keep the warning not to use allow_primary = true, but remove the examples? In a firefighting situation people are too likely to copy examples without reading the warning, so rather than saying "here's how you do this thing that will delete data, but don't do it", we keep the warning but don't show them how to do the thing we're telling them not to do

prem-fissionhq: Agreed Danny. Let's remove examples and add a link to official docs which helps us to get examples in case we need them.

dannyroberts (changes requested):
based on discoveries and discussion since original approval

prem-fissionhq:
@dannyroberts Could you please review my changes and let me know if any changes are necessary. thanks

dannyroberts (approved):

PR #4789, opened 2021-06-10
Fix CommCareSecretsAccess target prefix

skodde:
There may be discrepancies between the environment name we set in
terraform and the actual name of the environment, so let's use
account_alias instead as the prefix for the secrets.

##### ENVIRONMENTS AFFECTED
All


millerdev (commented):
The impact of this change is not clear to me. Do we need to migrate secrets after applying this change so we can continue to read them?

skodde:
> The impact of this change is not clear to me. Do we need to migrate secrets after applying this change so we can continue to read them?

For all of our current environments the resulting name will be the same, so no migration or change is needed.
However, we're setting up a new one where this is not the case (due to AWS length limits on some resource names), so we need to use a different (and more proper) variable.

millerdev (approved):

PR #4790, opened 2021-06-11
initial commit to create nlb in production and update pgbouncer_endpoint

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

Production

Routing formprocessing shards requests via pgshard_nlb. 
pgshard_nlb is not created yet in production.


environments/production/postgresql.yml
```diff
@@ -72,43 +72,43 @@ dbs:
   form_processing:
     proxy:
       host: pgbouncer3
-      pgbouncer_endpoint: pgbouncer3
+      pgbouncer_endpoint: pgshard_nlb
```
dannyroberts, line 75: I don't think we want to do this part yet, because only pgbouncer3 is capable of serving the pl/proxy database, which then routes to the shard databases. That was my understanding, did anything change there or was this just an oversight? I have not been following very closely.

environments/production/terraform.yml
```diff
@@ -407,6 +407,11 @@ pgbouncer_nlbs:
     targets:
       - pgbouncer6-production
       - pgbouncer7-production
+  - name: pgshard_nlb-production
+    targets:
+      - pgbouncer6-production
+      - pgbouncer7-production
```
dannyroberts, line 413: This seems okay and it's up to you to consider and weigh the pros and cons of each, but I want to point out that these two lines are an "additional change" in the sense that currently this gets routed to `pgbouncer9`. The most conservative approach I think would be to do this PR with just a single target here, `- pgbouncer9-production`, thus having it be the smallest change possible so you can focus surgically on one change at a time.

For an example as to why that is safer, note also that if we were to use pgbouncer6 and pgbouncer7, they would have to be in `pgbouncer_hosts` of the `p*` shard configurations in `environments/production/postgresql.yml`, which I don't see in this PR. (So if you choose that route, make sure to make that change as well.)

dannyroberts (changes requested):

environments/production/terraform.yml
prem-fissionhq: @dannyroberts please ignore these terraform.yml changes, these are incorrect as you noticed. It should be 3 and 9 only.

prem-fissionhq:
@dannyroberts correct my changes, could you please review this? 
thanks

dannyroberts (approved):

PR #4791, opened 2021-06-11
preparing plan to add new couchdb node in az-b production

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-11856
##### ENVIRONMENTS AFFECTED

Plan to add a new CouchDB node in az-b and we will be using snapshot-based copying of data onto this new node before running the migration process which is a manual process.

Thanks,


environments/production/terraform.yml
```diff
@@ -110,6 +110,17 @@ servers:
     group: "couchdb2_proxy"
     os: ubuntu_pro_bionic
 
+  - server_name: "couch11-production"
```
dannyroberts, line 113: Not a big deal, but curious why 11 and not 15 which would be the logical next number to pick.

environments/production/terraform.yml
```diff
@@ -110,6 +110,17 @@ servers:
...
+    server_instance_type: c5a.24xlarge
+    network_tier: "db-private"
+    az: "b"
+    volume_size: 60
+    volume_encrypted: no
```
dannyroberts, line 118: should be `volume_encrypted: yes` here too.

dannyroberts (commented):
I approve pending the `volume_encrypted` change

environments/production/terraform.yml
prem-fissionhq: @dannyroberts we just wanted it to be in serial. Instead of 12,13,15,16... if we want to have 11,12,13...

environments/production/terraform.yml
prem-fissionhq: Danny, I will discuss this with you once.

prem-fissionhq:
We have removed couch14 from the Proxy and placed it under maintenance. And added a new node couch11 into the cluster, updated proxy to accept active traffic.

dannyroberts (approved):

environments/production/terraform.yml
prem-fissionhq: @dannyroberts I'm sorry missed this notification. 
yes, the root volume also has to be encrypted but I didn't notice the volume_encrypted parameter is in two places.
Yesterday, I realized it and looked into ways to encrypt the root volume. Seems like the best possible solution is to - take a snapshot 
- create a new volume from the snapshot 
- stop the instance and attach the new volume as the root volume to the instance.
- Finally, import the changes to the state file.

dannyroberts: OK it's not urgent, so why don't you just make a ticket for this and we can do it after all other couch work is done

prem-fissionhq: Sure, Danny. I will create a ticket for this task.

PR #4796, opened 2021-06-16
Support RDS postgres major version higher than 9.6

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-11793
##### ENVIRONMENTS AFFECTED

Staging

adding local variables to get major and minor version values using engine_version we pass and apply them to postgres configuration parameters `family` and `major_engine_version` selection. 

These changes has been tested on Staging environment


shyamkumarlchauhan (approved):

src/commcare_cloud/terraform/modules/postgresql/main.tf
```diff
@@ -1,5 +1,16 @@
...
+
+  version_parts = regex("^(?P<major>[0-9]+)(?:\\.(?P<minor>[0-9]+))?", var.rds_instance["engine_version"])
+  version_parts_number = {
+    major = tonumber(local.version_parts["major"])
+  }
+
+  uses_shorthand_major_version = local.version_parts_number["major"] >= 10
+
+  computed_major_engine_version = local.uses_shorthand_major_version ? local.version_parts["major"] : "${local.version_parts["major"]}.${local.version_parts["minor"]}"
+  major_engine_version = ""
```
dannyroberts, line 12: I'm dropping this text from the postgres docs https://www.postgresql.org/support/versioning/ just to provide context for others, as it helped me:

> Starting with PostgreSQL 10, a major version is indicated by increasing the first part of the version, e.g. 10 to 11. Before PostgreSQL 10, a major version was indicated by increasing either the first or second part of the version number, e.g. 9.5 to 9.6.
>
> Minor releases are numbered by increasing the last part of the version number. Beginning with PostgreSQL 10, this is the second part of the version number, e.g. 10.0 to 10.1; for older versions this is the third part of the version number, e.g. 9.5.3 to 9.5.4.

src/commcare_cloud/terraform/modules/postgresql/main.tf
```diff
@@ -41,10 +52,10 @@ module "postgresql" {
   subnet_ids = "${var.subnet_ids}"
 
   # DB parameter group
-  family = "postgres9.6"
+  family = "postgres${local.computed_major_engine_version}"
 
   # DB option group
-  major_engine_version = "9.6"
+  major_engine_version = major_engine_version = local.major_engine_version == "" ? local.computed_major_engine_version : var.major_engine_version
```
dannyroberts, line 58: Should `major_engine_version = major_engine_version = ` be just `major_engine_version = `? I think this probably needs to be changed but let me know if I'm misunderstanding

dannyroberts (changes requested):

dannyroberts:
Approach looks good overall, just the somewhat minor comment

src/commcare_cloud/terraform/modules/postgresql/main.tf
prem-fissionhq: I just learned that `regex` has been introduced in 0.12.7 version, this change is not working with terraform11. Making changes to support both versions. 
https://github.com/hashicorp/terraform/blob/v0.12.7/CHANGELOG.md

dannyroberts:
In addition to the issue you found, another nit: This PR is currently called "RDS postgres major version upgrade from 9.6 to 12.5" which is way scarier than what it's actually doing. It's making it possible for an environment to use a version other than 9.6.x. Maybe "Support RDS postgres major version higher than 9.6" or something like that. Do I have that right?

prem-fissionhq:
> In addition to the issue you found, another nit: This PR is currently called "RDS postgres major version upgrade from 9.6 to 12.5" which is way scarier than what it's actually doing. It's making it possible for an environment to use a version other than 9.6.x. Maybe "Support RDS postgres major version higher than 9.6" or something like that. Do I have that right?

sure Danny. thanks for correcting. Will change.

prem-fissionhq:
@dannyroberts the `regex` expression is still not supported with the current terraform version. 
These changes are working as expected without any issues with versioning (10,11..etc). Could you please suggest the next steps on this? 
These changes have been applied to staging as well, without any issues.  

thanks.

dannyroberts (approved):

PR #4797, opened 2021-06-16
Upgrade datadog client library

millerdev:
Reviewed [change log](https://github.com/DataDog/datadogpy/blob/master/CHANGELOG.md) and verified that we are not using any APIs with backward-incompatible changes.

This was originally done as a precursor for another library upgrade, but then after I had verified the safety of this change I realized it was not strictly required for that other upgrade. Therefore I am spinning it off as a separate PR so that work doesn't go to waste.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -44,7 +44,7 @@ cryptography==3.3.2
     #   ansible
     #   commcare-cloud (setup.py)
     #   paramiko
-datadog==0.2.0
+datadog==0.41.0
     # via commcare-cloud (setup.py)
 decorator==4.4.2
     # via datadog
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ install_deps = [
     'clint',
     'couchdb-cluster-admin>=0.7.0',
     'cryptography>=3.2',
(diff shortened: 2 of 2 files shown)


dannyroberts:
I don't have much context here, but just wanted to make sure you checked that it was compatible with the agent version we're using (datadog agent v5 I believe). I'm guessing since you didn't see anything in the changelog about it that it is

dannyroberts (approved):
Approved assuming nothing comes up from the consideration I noted

millerdev:
@dannyroberts Correct, the change log did not mention agent versions at all. As an additional precaution, I checked the github issues and pull requests to see if people have reported problems related to agent versions not being supported. Found a few related to v6 and v7 support (all closed now), but nothing about v5 not being supported. I'm not sure how else to tell other than to try it and revert if it breaks. Is there a way to test it before merging?

snopoke (approved):
I don't think this impacts the agent at all (or even interacts with the agent unless you use the `dogstatsd` metrics exporter (which we don't). This get's used for creating events and calling other DD APIs.

PR #4804, opened 2021-06-23
remove reference to allow_primary in docs

snopoke:
`allow_primary` is a dangerous parameter to include:

> The allow_primary parameter will force a new empty primary shard to be allocated without any data. If a node which has a copy of the original shard (including data) rejoins the cluster later on, that data will be deleted: the old shard copy will be replaced by the new live shard copy.


dannyroberts:
Oops, just merged a similar change by Prem https://github.com/dimagi/commcare-cloud/pull/4774 that now conflicts. If there's something else you want to tweak on top of that let me know and I can review

sravfeyn (approved):

PR #4805, opened 2021-06-24
adding chrony package for better time syncronization

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-12195
##### ENVIRONMENTS AFFECTED

All

Applied on staging server as of now.

### DIFF
--- a/environments/india/public.yml
+++ b/environments/india/public.yml
@@ -240,3 +240,7 @@ efs_shared_mount_dir: "shared{{ '_' ~ deploy_env if deploy_env != 'production' e
 efs_shared_mount_endpoint: "shared-efs.india.commcare.local:/"
 efs_shared_mount_options: "_netdev,nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport"
 shared_dir_gid: 1500
+
+# for better time syncronization on aws servers
+# replace ntp with chrony.
+use_chrony: True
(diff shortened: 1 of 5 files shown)


dannyroberts:
Looks like this will mean both `chrony` and `ntp` are installed? Have you looked into what the docs say about having them both installed? Just want to make sure they're compatible and won't interfere with each other

prem-fissionhq:
@dannyroberts We verified that chrony is uninstalling existing NTP package if it already installed so we thought of supporting both pacakges. Please let us know your suggestions. I think we can remove ntp from installing if not neccessary.  thanks

dannyroberts:
OK I don't know enough of the details but I wanted to point out that `ntp` is installed in this same list (https://github.com/dimagi/commcare-cloud/pull/4805/files#diff-73416f28d12dc5cf4883b6343daca0dbefbf48502f82e43492db07f7c43becb2R59), which is at best redundant if `chrony` automatically uninstalls `ntp` (and at worst could interfere).

prem-fissionhq:
@dannyroberts I have discussed again on this with Shyam and will use a var to control this installation. However, we are sure (tested) that chrony will uninstall the NTP while installing.

shyamkumarlchauhan (approved):

shyamkumarlchauhan (approved):

prem-fissionhq:
@shyamkumarlchauhan can I use the same PR to apply changes on India or shall I apply after merging please let me know. thanks

shyamkumarlchauhan (approved):

prem-fissionhq:
@shyamkumarlchauhan 
removing a task that verifies timesync server address in conf file as it is not required. The module lineinline updates the file only if not exist. 
Creating new task to install ntp package only when use_chrony is false. this is to avoid installation and uninstallation of chront and ntp packages.

shyamkumarlchauhan (approved):

prem-fissionhq:
Applied on India Env.

dannyroberts (approved):

PR #4807, opened 2021-06-29
Cs/monolith install cchq updates

Charl1996:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-1476)

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/docs/setup/new_environment.md
+++ b/docs/setup/new_environment.md
@@ -341,11 +341,6 @@ Deploying CommcareHQ for the first time needs a few things enabled first.
     ```bash
     $ commcare-cloud monolith ansible-playbook letsencrypt_cert.yml --skip-check
     ```
-   Note: you might or might not need to create the `/etc/letsencrypt` directory yourself. If the command above fails due
-   to no such directory found, create one with the following command:
-    ``` bash
-    $ sudo mkdir /etc/letsencrypt
-    ```
 
 2. Update settings to take advantage of new certs:
     ```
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/deploy_logstash.yml
```diff
@@ -5,7 +5,7 @@
   vars:
     sansible_logstash_family: 7.x
     sansible_java_version: 8
-    sansible_java_apt_keyserver: 'http://keys.gnupg.net'
+    sansible_java_apt_keyserver: 'https://keyserver.ubuntu.com'
```
Charl1996, line 8: There has been a mention on the #devops slack channel regarding the host `keys.gnupg.net` being down for some while. I had to replace the keyserver host with this one for it to work. 

Why `keyserver.ubuntu.com`?
No real reason other than the fact that it's a different keyserver. We might go with `https://pgp.mit.edu/` also.

That said, I'm not sure if this is what we want to do, ie permanently changing the keyserver (until it's down, maybe changing it again, etc). I'm not ansible-smart, but is there a way we could try a fallback keyserver in case the one fails? 

Open to suggestions...

docs/setup/new_environment.md
```diff
@@ -341,11 +341,6 @@ Deploying CommcareHQ for the first time needs a few things enabled first.
     ```bash
     $ commcare-cloud monolith ansible-playbook letsencrypt_cert.yml --skip-check
     ```
-   Note: you might or might not need to create the `/etc/letsencrypt` directory yourself. If the command above fails due
-   to no such directory found, create one with the following command:
-    ``` bash
-    $ sudo mkdir /etc/letsencrypt
-    ```
```
esoergel, line 348: :star: Love seeing steps removed

esoergel (approved):
Would be nice to get a second opinion on the keyserver question, but this looks good to me

PR #4813, opened 2021-07-06
Extract celery queue

mkangia:
I believe due to multiple queues being handled by the same worker, it has been going down repeatedly, and the resolution has been to restart this worker.
This PR just extracts out the celery queue, out, which is the default queue and should be handling lots of tasks, and hence it would reduce the load on the worker.

##### ENVIRONMENTS AFFECTED

`staging`


calellowitz (approved):

mkangia:
Rolled out on staging

PR #4814, opened 2021-07-08
Make after-reboot idempotent (no restarts!)

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-11220
##### ENVIRONMENTS AFFECTED

All

Avoid restarting the Kafka service when already started in case of after-reboot.

### DIFF
--- a/src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
@@ -93,9 +93,21 @@
     daemon_reload: yes
   when: ansible_distribution_version == '18.04'
 
+# Allows to excecute task only when a tag is specified: https://serverfault.com/a/748864
+- shell: /bin/true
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
```diff
@@ -93,9 +93,21 @@
...
+  changed_when: false
+  register: no_tags
+
+- name: start Kafka service
+  service: name=kafka-server enabled=yes state=started
+  when: no_tags is not defined
+  register: kafka_service_started
+  tags: after-reboot
+
+#Restarting a service right after it's been started can cause an inconsistent state. adding condition.
```
dannyroberts, line 107:
```suggestion
# Restarting a service right after it's been started will cause an unnecessary/non-idempotent restart
```

dannyroberts (approved):

PR #4823, opened 2021-07-15
Logship /var/log/commands.log

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-12453

##### ENVIRONMENTS AFFECTED
staging india production

### DIFF
--- a/src/commcare_cloud/ansible/deploy_cloudwatch_logs.yml
+++ b/src/commcare_cloud/ansible/deploy_cloudwatch_logs.yml
@@ -23,6 +23,9 @@
           - file_path: "/var/log/apt/history.log"
             log_group_name_suffix: "var-log-apt-history"
             should_include: true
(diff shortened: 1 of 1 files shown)


dannyroberts:
Applied on staging with the desired result

joxl (approved):
I had assumed shell history was already being aggregated across systems. This change seems like a good time to engage a discussion about how to handle the scenario where a user accidentally enters sensitive data in a shell command (e.g. password, API token, PII, PHI, etc).  If the shell logs are not being aggregated, sanitizing the places such data was written on an isolated system is fairly straight-forward. With remote aggregation, sanitizing logs of sensitive data can become challenging. If we don't already have one, I think it's worth considering and planning what our policy is in the scenario where sensitive data is accidentally (or otherwise) entered as a shell command (or part of one).

dannyroberts:
Joel and I offlined to a different channel

PR #4824, opened 2021-07-15
Have our tooling enforce 24 char minimum length for openvpn password

dannyroberts:
Helpful check for what our team already does, namely use a password manager and randomly generated passwords of a reasonable length.

Discussion on https://dimagi-dev.atlassian.net/browse/SAAS-12455

##### ENVIRONMENTS AFFECTED
staging, india, production


mjriley (approved):
Looks good

PR #4826, opened 2021-07-20
Add ability to use aws ssm to connect through the control machine

dannyroberts:
Note: currently requires `export COMMCARE_CLOUD_USE_AWS_SSM=1` to work

## Using AWS SSM to log onto control machine

This is the "main" change. This adds an alternative to using the VPN: using SSM to log onto the control machine and running all further commands from the control machine. Eventually this will be the default, but in the short term to use this functionality you will need to add `export COMMCARE_CLOUD_USE_AWS_SSM=1` to your `~/.profile` or run it once per new terminal tab you open. After that, you can just use the `cchq` command normally: `cchq staging ssh control` and `cchq --control staging deploy`, etc. all work as they did before; the difference is to do anything that request connecting to a machine other than control, you first have to run `cchq staging ssh control`, and then run the command from there.

The first time you do this (or anytime it's not already installed) it'll ask you to install a necessary plugin from AWS:

Screen Shot 2021-07-23 at 5 01 15 PM

## Forward Port command
In addition, this PR adds a `forward-port` command that is designed to allow you to connect securely to browser-based management consoles like couch fauxton, celery flower, and elasticsearch head—all without connecting to the VPN.

Using this command will require some first-time setup. Here's a screenshot of the interactive setup experience

Screen Shot 2021-07-23 at 2 49 37 PM

You'll notice that it requires root access, but it doesn't run any of those commands itself, letting you examine what you're running before you run it.

Subsequent runs are then much smoother:

Screen Shot 2021-07-23 at 2 52 15 PM

My hope is that the setup will be fairly straightforward, and the ongoing experience afterwards will be very nice, nicer than what we currently have with the VPN.

##### ENVIRONMENTS AFFECTED
production, india, staging

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -132,6 +132,10 @@ if ! grep -q init-ansible ~/.profile 2>/dev/null; then
             echo '[ -t 1 ] && source ~/init-ansible' >> ~/.profile
             printf "${YELLOW}→ Added init script to ~/.profile\n"
         ;;
+        [Nn]* )
+            echo '# [ -t 1 ] && source ~/init-ansible' >> ~/.profile
+            printf "${YELLOW}→ Okay, I won't ask you again. Added a line you can uncomment in ~/.profile if you want it later.\n"
+        ;;
         * )
             printf "\n${BLUE}You can always set it up later by running this command:\n"
             printf "${BLUE}echo '[ -t 1 ] && source ~/init-ansible' >> ~/.profile${NC}\n"
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -17,7 +17,7 @@ from ..ansible.helpers import get_default_ssh_options_as_cmd_parts
 from ..terraform.aws import aws_sign_in, is_aws_env, is_ec2_instance_in_account
 from ...alias import commcare_cloud
 
-from ...colors import color_error
+from ...colors import color_error, color_notice
 from .getinventory import (get_monolith_address, get_server_address,
                            split_host_group)
 
@@ -306,7 +306,22 @@ class ForwardPort(CommandBase):
 
     def run(self, args, unknown_args):
         environment = get_environment(args.env_name)
+        nice_name = environment.terraform_config.account_alias
         remote_host_key, remote_port = self._SERVICES[args.service]
         loopback_address = f'127.0.{environment.terraform_config.vpc_begin_range}'
         remote_host = lookup_server_address(args.env_name, remote_host_key)
-        return commcare_cloud(args.env_name, 'ssh', 'control', '-NL', f'{loopback_address}:{remote_port}:{remote_host}:{remote_port}')
+        puts(color_notice('Make sure you have done the following:'))
+        puts(color_notice(f'  - Run `sudo ifconfig lo0 alias {loopback_address}`.'))
+        puts(color_notice(f'  - Edit `/etc/hosts` (with sudo) to include the line `{loopback_address} {nice_name}`.'))
+        puts(color_notice(f'  - Exit and retry this command after those changes.'))
+        puts()
+        puts(color_notice(f'{args.env_name} {args.service} should now be available to you at {nice_name}:{remote_port}'))
+        puts(color_notice(f'^C to stop port-forwarding and exit'))
+        puts()
+        try:
+            return commcare_cloud(args.env_name, 'ssh', 'control', '-NL', f'{loopback_address}:{remote_port}:{remote_host}:{remote_port}')
+        except KeyboardInterrupt:
+            puts()
+            puts('Connection closed.')
+            # ^C this is how we expect the user to terminate this command, so no need to print a stacktrace
+            return 0
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -299,9 +299,9 @@ class ForwardPort(CommandBase):
     Port forward to access a remote admin console
     """
     _SERVICES = {
-        'flower': ('celery[0]', 5555),
-        'couch': ('couchdb2_proxy[0]', 25984),
-        'elasticsearch': ('elasticsearch[0]', 9200),
+        'flower': ('celery[0]', 5555, '/'),
+        'couch': ('couchdb2_proxy[0]', 25984, '/_utils/'),
+        'elasticsearch': ('elasticsearch[0]', 9200, '/'),
     }
 
     arguments = (
@@ -313,7 +313,7 @@ class ForwardPort(CommandBase):
     def run(self, args, unknown_args):
         environment = get_environment(args.env_name)
         nice_name = environment.terraform_config.account_alias
-        remote_host_key, remote_port = self._SERVICES[args.service]
+        remote_host_key, remote_port, url_path = self._SERVICES[args.service]
         loopback_address = f'127.0.{environment.terraform_config.vpc_begin_range}'
         remote_host = lookup_server_address(args.env_name, remote_host_key)
         local_port = self.get_random_available_port()
@@ -333,7 +333,7 @@ class ForwardPort(CommandBase):
                 return -1
             puts()
 
-        puts(color_notice(f'You should now be able to reach {args.env_name} {args.service} at {color_link(f"http://{nice_name}:{local_port}/")}.'))
+        puts(color_notice(f'You should now be able to reach {args.env_name} {args.service} at {color_link(f"http://{nice_name}:{local_port}{url_path}")}.'))
         puts(f'Interrupt with ^C to stop port-forwarding and exit.')
         puts()
         try:
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -355,7 +355,7 @@ class ForwardPort(CommandBase):
     @staticmethod
     def is_etc_hosts_alias_set_up(loopback_address, nice_name):
 
-        grep_regex = rf"{loopback_address}\s+{nice_name}"
+        grep_regex = rf"^{loopback_address}\s+{nice_name}"
         try:
             subprocess.check_output(f'grep -E {shlex_quote(grep_regex)} /etc/hosts', shell=True)
         except subprocess.CalledProcessError:
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -17,7 +17,7 @@ from ..ansible.helpers import get_default_ssh_options_as_cmd_parts
 from ..terraform.aws import aws_sign_in, is_aws_env, is_ec2_instance_in_account
 from ...alias import commcare_cloud
 
-from ...colors import color_error, color_notice
+from ...colors import color_error, color_notice, color_link
 from .getinventory import (get_monolith_address, get_server_address,
                            split_host_group)
 
@@ -326,8 +326,8 @@ class ForwardPort(CommandBase):
                 return -1
             puts()
 
-        puts(color_notice(f'{args.env_name} {args.service} should now be available to you at {nice_name}:{remote_port}'))
-        puts(color_notice(f'^C to stop port-forwarding and exit'))
+        puts(color_notice(f'You should now be able to reach {args.env_name} {args.service} at {color_link(f"http://{nice_name}:{remote_port}/")}.'))
+        puts(f'Interrupt with ^C to stop port-forwarding and exit.')
         puts()
         try:
             return commcare_cloud(args.env_name, 'ssh', 'control', '-NL', f'{loopback_address}:{remote_port}:{remote_host}:{remote_port}')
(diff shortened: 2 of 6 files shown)


src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -269,3 +285,28 @@ def _get_ssh_args(remote_command):
...
+    arguments = (
+        Argument('service', choices=_SERVICES.keys(), help=f"""
+            The remote service to port forward. Must be one of {','.join(sorted(_SERVICES.keys()))}.
+        """),
+    )
+
+    def run(self, args, unknown_args):
+        environment = get_environment(args.env_name)
+        remote_host_key, remote_port = self._SERVICES[args.service]
+        loopback_address = f'127.0.{environment.terraform_config.vpc_begin_range}'
```
dannyroberts, line 310: So this will be for example 127.0.10.201 for staging

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -269,3 +285,28 @@ def _get_ssh_args(remote_command):
...
+        remote_host = lookup_server_address(args.env_name, remote_host_key)
+        return commcare_cloud(args.env_name, 'ssh', 'control', '-gNL', f'{loopback_address}:{remote_port}:{remote_host}:{remote_port}')
```
dannyroberts, line 312:
```suggestion
        return commcare_cloud(args.env_name, 'ssh', 'control', '-NL', f'{loopback_address}:{remote_port}:{remote_host}:{remote_port}')
```

control/init.sh
```diff
@@ -132,6 +132,10 @@ if ! grep -q init-ansible ~/.profile 2>/dev/null; then
             echo '[ -t 1 ] && source ~/init-ansible' >> ~/.profile
             printf "${YELLOW}→ Added init script to ~/.profile\n"
         ;;
+        [Nn]* )
+            echo '# [ -t 1 ] && source ~/init-ansible' >> ~/.profile
```
calellowitz, line 136: if someone explicitly says no, adding a line, even a commented one, seems like the wrong behavior.

calellowitz (commented):
I know I had a chance to comment on the email and didn't, but one thought as I go through this is if it is possible to have the address for the port forwarding be something more human readable. While the convention shown here is good to prevent someone from confusing local flower and production, it is unlikely most developers know the ip range for the diferent clusters off the top of their heads, making it easy to confuse staging and production.

We used to have internal dns entries (or public ones?) so that when you wanted to look at flower you went to `hqcelery1.www.commcarehq.org` or `hqcelery1.india.commcarehq.org` and that made it abundantly clear in the URL bar, which cluster you were viewing. If it is possible to do something like that here, even if it involved mucking about in `/etc/hosts` that would be much clearer than an opaque IP address in the url bar.

dannyroberts:
@calellowitz yeah this is my fault for writing such a long email, but the /etc/hosts suggestion is buried in there.
You just have to add a line like "127.0.10.201 commcare-staging" to your /etc/hosts for each env, and then you're able to go to e.g. commcare-staging:25984 in your browser. This does require using sudo, as does enabling these nonstandard loopback addresses, which is the thing I don't love about it. But yeah the ability to give it a human readable name was the thing I liked most about this approach

control/init.sh
dannyroberts: Yeah that's true. The thing is that without adding this commented out line, it'll all you every single time. I wanted to make it easy for people to say no once and then never be asked about it again. I had the control machine in mind but I guess some people install it locally using this script as well? I'm open to changing this.

calellowitz:
> @calellowitz yeah this is my fault for writing such a long email, but the /etc/hosts suggestion is buried in there.
> You just have to add a line like "127.0.10.201 commcare-staging" to your /etc/hosts for each env, and then you're able to go to e.g. commcare-staging:25984 in your browser. This does require using sudo, as does enabling these nonstandard loopback addresses, which is the thing I don't love about it. But yeah the ability to give it a human readable name was the thing I liked most about this approach

Nice. Is there a way to enforce that? It would be cool if it were automatic or only listening at that hostname, but the option is certainly better than not having it.

dannyroberts:
Interesting, I wasn't thinking of actually enforcing it, but rather documenting it (which admittedly I didn't do in this PR) and making it the most pleasant way to use the tool.

dannyroberts:
Added those instructions

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -306,7 +306,22 @@ class ForwardPort(CommandBase):
 
     def run(self, args, unknown_args):
         environment = get_environment(args.env_name)
+        nice_name = environment.terraform_config.account_alias
         remote_host_key, remote_port = self._SERVICES[args.service]
         loopback_address = f'127.0.{environment.terraform_config.vpc_begin_range}'
         remote_host = lookup_server_address(args.env_name, remote_host_key)
-        return commcare_cloud(args.env_name, 'ssh', 'control', '-NL', f'{loopback_address}:{remote_port}:{remote_host}:{remote_port}')
+        puts(color_notice('Make sure you have done the following:'))
+        puts(color_notice(f'  - Run `sudo ifconfig lo0 alias {loopback_address}`.'))
```
joxl, line 314: Would be nice to include an ansible playbook that devs can run on their workstation to handle this boilerplate in an idempotent way.  It could handle the `/etc/hosts` records and support multiple OSes (e.g. support `ifconfig` and `iproute2`) in an elegant way. Then this comment could be something simpler like:

```python
puts(color_notice('  - Run `ansible-playbook ./.../loopbacks.yml -i localhost --diff --check`.'))
puts(color_notice('    to verify your workstation configuration, or to enforce:'))
puts(color_notice('  - Run `ansible-playbook ./.../loopbacks.yml -i localhost --diff -K`.'))
```

dannyroberts: Hmm, I see what you're saying. I have a bit of a reaction to the idea of maintaining dev environments with ansible, because everyone does things a little differently and keeping those working right and not interfering with people's individual setups seems like it'll create a really leaky abstraction that's more frustrating than helpful.

Perhaps instead of trying to inline it here, I should link to documentation that has more of a "if you're on Mac do X, if you're on Linux do Y" approach. I don't know. What do you think about my concerns about the leakiness of it?

dannyroberts: And actually regardless @joxl can you tell me the equivalent command you have to run to get this to work for you locally on Linux? (If you have any trouble getting this set up, ping me on slack because that'll be useful info too.)

joxl: Yeah, I would not suggest that such a playbook get executed automatically, I would prefer it be up to the developer to do so. But I'm guessing 99% of devs will be doing something very similar (add some loopback aliases and respective host records), so it could be handy.

Linux commands are (Marco also provided this in his email feedback regarding `iproute2`):
```bash
# observe current state of 'lo' interface
$ ip addr show lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever

# add an alias for 'lo'
$ sudo ip addr add 127.0.49.201/8 dev lo

# observe change
$ ip addr show lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet 127.0.49.201/8 scope host secondary lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever

# remove it (if you wish)
$ sudo ip addr del 127.0.49.201/8 dev lo
```

dannyroberts: Thanks. I went ahead and made the script detect whether the user has run the setup steps, and then run them through any steps they still have to do interactively. I tried it my self a bunch of times with different setup missing and made sure the experience made sense in every combination.

I'll reach out on slack to have someone test the Linux experience.

dannyroberts:
Both I (Mac) and Marco (Linux) have tested the setup and all the commits from the last few hours came from us iterating offline to fix issues or make improvements. This is ready for review, and I invite you all to try it locally and tell me what you think

millerdev (commented):
I am -1 on the local network configuration requiring elevated privileges on the local machine. I think it would be safer to simply print out the full URL (containing an IP address) that the user should use to connect to the service. The local port could be a randomly chosen available port, or possibly a specific local port. The latter would require us to choose a standard set of local ports for each environment, which has some disadvantages that I think make it a less attractive option.

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -269,3 +291,74 @@ def _get_ssh_args(remote_command):
...
+        environment = get_environment(args.env_name)
+        nice_name = environment.terraform_config.account_alias
+        remote_host_key, remote_port = self._SERVICES[args.service]
+        loopback_address = f'127.0.{environment.terraform_config.vpc_begin_range}'
+        remote_host = lookup_server_address(args.env_name, remote_host_key)
+
+        while not self.is_loopback_address_set_up(loopback_address):
+            puts(color_error('To make this work you will need to run set up a special loopback address on your local machine:'))
+            puts(color_notice(f'  - Mac: Run `sudo ifconfig lo0 alias {loopback_address}`.'))
+            puts(color_notice(f'  - Linux: Run `sudo ip addr add {loopback_address}/8 dev lo`.'))
```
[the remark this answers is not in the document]
dannyroberts: Agreed this is really dangerous and makes the current implementation a non-starter. Specifically that someone could go to http://commcare-staging:5984/ and see local data when they expect to see remote data opens the door for exactly the type of confusion we want to avoid.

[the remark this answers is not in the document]
skodde: This is something we can easily fix by binding for example on port 0, obtaining a random available port.
I think it's still a good idea to maintain the current implementation in place, because we want to have the environment name spelled out in the link, clearly visible in the open window/tab/terminal. We just need to work around these potential conflicts, and the above should be an effective solution.

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -269,3 +291,74 @@ def _get_ssh_args(remote_command):
...
+            if not ask("Follow the instructions above or type n to exit. Ready to continue?"):
+                return -1
+            puts()
+
+        while not self.is_etc_hosts_alias_set_up(loopback_address, nice_name):
+            puts(color_error('Okay, now the last step is to set up a special alias in your /etc/hosts:'))
+            puts(color_notice(f'  - Edit /etc/hosts (e.g. `sudo vim /etc/hosts`) and add the line `{loopback_address} {nice_name}` to it.'))
```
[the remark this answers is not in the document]
skodde: I don't consider this bad practice, we clearly state the intent of the operation and the command itself, which has to be manually entered. This is not a 'curl | sh' type of situation.
It also concerns two very basic network configurations that the audience should be able to understand.

dannyroberts:
I think I've now incorporated the feedback, though there's some disagreement among reviewers. I also think that the benefits of the visual segregation of different environments in the url bar (commcare-production, commcare-india, etc) is important, and worth making people run these commands with `sudo` if that's the only way to achieve it. I had the same reaction to using `sudo` initially, but in the absence of an alternative to achieve the same result, I decided enforcing that visual distinction was more imporant.

dannyroberts:
It now looks like Screen Shot 2021-08-12 at 6 24 48 PM, where that is a random unused port

skodde (approved):

dannyroberts:
I'm going to go ahead and merge this in now, but I don't want to close debate. For now these features will be available to anyone who wants to preview them `export COMMCARE_CLOUD_USE_AWS_SSM=1`. Until there's a followup PR, conversation can continue on this PR.

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -316,6 +316,7 @@ def run(self, args, unknown_args):
         remote_host_key, remote_port = self._SERVICES[args.service]
         loopback_address = f'127.0.{environment.terraform_config.vpc_begin_range}'
         remote_host = lookup_server_address(args.env_name, remote_host_key)
+        local_port = self.get_random_available_port()
```
millerdev, line 319: Noting that this is vulnerable to a race condition where some other process claims the port before the SSH connection+forward is established. Obviously it is impossible to report to the user (before establishing the connection) which port will be chosen if port selection is deferred to the SSH process by passing 0 in the port-forward directive. Not sure if it is possible to do it another way to avoid that issue?

Unfortunately I think the failure mode will not cause the entire SSH connection to fail (only the port forward will fail with `Could not request local forwarding.`), so unless you read the terminal output closely it may not be obvious.

dannyroberts: Yeah this is a good point. I thought of this, but didn't think it was likely in practice, though I could very much be overlooking how frequently OS processes ask for random ports and the chance of them overlapping. The ideal solution would have been if there was some way to tell SSH to bind to a random port and tell us what port it chose, but I couldn't find any mechanism in SSH -L that supported that. (When I tried, it appeared not to support `0` as a port option meaning choose a random port, in addition to the problem you mentioned of then not being able to report it to the user, but maybe I was doing something wrong.)

PR #4827, opened 2021-07-20
Change the couch apt repo from bintray to jfrog

proteusvacuum:
This is the repo listed in the couchdb docs: https://docs.couchdb.org/en/stable/install/unix.html 
The previous bintray address has been sunset: https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

### DIFF
--- a/src/commcare_cloud/ansible/roles/couchdb2-preinstall/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/couchdb2-preinstall/tasks/main.yml
@@ -31,8 +31,8 @@
 
 - name: Add Couch apt repo (for libmozjs dependency)
   block:
-    - apt_key: url=https://couchdb.apache.org/repo/bintray-pubkey.asc state=present
-    - apt_repository: repo="deb https://apache.bintray.com/couchdb-deb bionic main" state=present
+    - apt_key: url=https://couchdb.apache.org/repo/keys.asc state=present
+    - apt_repository: repo="deb https://apache.jfrog.io/artifactory/couchdb-deb bionic main" state=present
   when: ansible_distribution_version == '18.04'
 
 - name: Add Erlang apt repo


dannyroberts:
Does this affect existing installations? I.e. will running this changed role against an existing installation of couch result in any changes like reinstallation or a restart?

proteusvacuum:
@dannyroberts That's a good question, I'm not sure! The same pattern was followed here: https://github.com/dimagi/commcare-cloud/pull/4701 do you know if that had adverse effects?

dannyroberts:
We offlined and agreed to do a test on staging first which I'll make sure happens before we merge

prem-fissionhq:
@dannyroberts @proteusvacuum I have run this on Staging env and it is not uninstalling the existing couch or restarting the service.

dannyroberts (approved):

src/commcare_cloud/ansible/roles/couchdb2-preinstall/tasks/main.yml
```diff
@@ -31,8 +31,8 @@
 
 - name: Add Couch apt repo (for libmozjs dependency)
   block:
-    - apt_key: url=https://couchdb.apache.org/repo/bintray-pubkey.asc state=present
-    - apt_repository: repo="deb https://apache.bintray.com/couchdb-deb bionic main" state=present
```
dannyroberts, line 35: @skodde rather than having the process be that we manually remove these, what do you think about adding these two lines back but changing `state=present` to `state=absent`? I think that would be the normal process for an update like this and would make it so we don't have to manually remove files from each machine

skodde: Fixed here: https://github.com/dimagi/commcare-cloud/pull/4878/commits/be095f19d2dc4180079da3d780022e8d028dc50e

PR #4835, opened 2021-07-23
Don't error on key_type-key pairs

kaapstorm:
A self-hosting partner has a `known_hosts` file that has multiple key_type-key pairs on the same line. (See **tests/test_ops_tool.py:23** `test_get_host_key_map()` for an example.)

This change is to support that. Current behavior is preserved.

##### ENVIRONMENTS AFFECTED

No Dimagi environments are affected.

### DIFF
--- a/src/commcare_cloud/commands/ansible/ops_tool.py
+++ b/src/commcare_cloud/commands/ansible/ops_tool.py
@@ -416,11 +416,14 @@ def _get_host_key_map(known_host_lines):
     for line in known_host_lines:
         if not line:
             continue
-        csv_hosts, key_type, key = line.split(' ')
+        # line is formatted "<hosts> <key_type> <key> <key_type> <key> ..."
+        csv_hosts, *key_types_keys_list = line.split(' ')
+        key_type_key_pairs = chunked(key_types_keys_list, 2)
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -437,3 +440,17 @@ def _get_lines(proc):
...
+    """
+    Collect data into fixed-length chunks
+
+    >>> list(chunked('ABCDEFG', 3, 'x'))
+    [('A', 'B', 'C'), ('D', 'E', 'F'), ('G', 'x', 'x')]
+
+    """
+    # Pulled shamelessly from itertools recipes
+    # https://docs.python.org/3/library/itertools.html#itertools-recipes
+    args = [iter(iterable)] * n
```
snopoke, line 455: so sneaky!

snopoke (approved):

src/commcare_cloud/commands/ansible/ops_tool.py
kaapstorm: Yeah! I would not have thought of that. :nerd_face:

PR #4840, opened 2021-08-02
Find and remove all but last 2 formplayer releases

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-11939

This adds a cleanup task to the `deploy_formplayer` playbook, which keeps all but the last 2 formplayer releases sorted by last modified time. 

I'm not quite sure how to properly organize/structure something like this. Is the cleaner approach to put these tasks in their own file that the deploy playbook can call? Definitely looking for feedback here. Thanks.

### DIFF
--- a/src/commcare_cloud/ansible/deploy_formplayer.yml
+++ b/src/commcare_cloud/ansible/deploy_formplayer.yml
@@ -25,3 +25,19 @@
       tags: kinesis-agent
   handlers:
     - include: roles/monit/handlers/main.yml
+
+- name: Cleanup formplayer releases
+  hosts: formplayer
+  tasks:
+    - name: List of releases
+      find:
+        paths="{{ formplayer_build_dir }}"
+        file_type=directory
+      register: releases
+
+    - name: Keep the last 2 releases
+      file:
+        path="{{ item.path }}"
+        state=absent
+      with_items:
+        - "{{ (releases.files | sort(attribute='ctime'))[:-2] }}"
--- a/src/commcare_cloud/ansible/group_vars/formplayer.yml
+++ b/src/commcare_cloud/ansible/group_vars/formplayer.yml
@@ -1,1 +1,2 @@
 datadog_integration_jmx: true
+formplayer_build_dir: '/home/cchq/www/staging/formplayer_build/releases'


src/commcare_cloud/ansible/deploy_formplayer.yml
```diff
@@ -25,3 +25,19 @@
...
+        paths="{{ formplayer_build_dir }}"
+        file_type=directory
+      register: releases
+
+    - name: Keep the last 2 releases
+      file:
+        path="{{ item.path }}"
+        state=absent
+      with_items:
+        - "{{ (releases.files | sort(attribute='ctime'))[:-2] }}"
```
dannyroberts, line 43: I'd suggest that we do two more things here:
- move the cleanup _before_ we create a new release
- resolve the 'current' directory's symlink and make sure that that directory is excluded from the list of directories to be removed.

This will result in us always knowing that the previous release is there to roll back to even if it's not chronologically the most recent release. That could happen if a directory is created for a release but there's an error and it exits before updating the symlink. The deploy process is designed that way, so that if there's an error during the deploy, it will effectively be a no-op not affecting the running environment at all. In such a case though you might have

- release 1
- release 2 <-- current
- release 3 (failed)
- release 4 (failed)
- release 5 (failed)

and you wouldn't want the next deploy to only save release 5. With my suggestion it would save release 5, release 4 (previous 2) and release 3 (current release), and then create the next release as well.

dannyroberts (approved):

src/commcare_cloud/ansible/deploy_formplayer.yml
snopoke: how would we know if a release directory was from a failed deploy or not? In HQ we keep a metadata file of all the successful releases. We could do something similar with formplayer.

src/commcare_cloud/ansible/roles/formplayer/vars/main.yml
```diff
@@ -10,6 +10,7 @@ formplayer_build_url: "{{
 }}"
 formplayer_current_dir: "{{ formplayer_build_dir }}/current"
 formplayer_release_dir: '{{ formplayer_build_dir }}/releases/{{ new_release_name }}'
+formplayer_releases_dir: '{{ formplayer_build_dir }}/releases'
```
gherceg, line 13: I was tempted to rename `formplayer_release_dir` to `formplayer_current_release_dir`, but if there is a suggestion for a better name for `formplayer_releases_dir` open to hearing it to avoid potential confusion in the future.

dannyroberts: You could try `formplayer_releases_base_dir`, which emphasizes that it's the parent directory to each release dir

gherceg:
Tested on staging. Pre-deploy `formplayer4` looked like this:
Screen Shot 2021-08-12 at 10 28 44 AM
With the current release being:
Screen Shot 2021-08-12 at 10 29 08 AM

Post-deploy (notice the previous `current` release is there):
Screen Shot 2021-08-12 at 10 34 58 AM
With the current release being:
Screen Shot 2021-08-12 at 10 35 13 AM

PR #4841, opened 2021-08-04
Renamed user_case to usercase

stephherbers:
[Ticket](https://dimagi-dev.atlassian.net/browse/SUPPORT-10242) and another installment of changing "user_case" to "usercase"

This now matches in hq [here](https://github.com/dimagi/commcare-hq/blob/master/custom/covid/rules/custom_criteria.py#L12) when before it was preventing this rule to run.


stephherbers:
@orangejenny I'll merge from now but definitely not opposed to a more preventative solution

proteusvacuum:
Oh dear, sorry about this! 
I agree with moving it back into HQ.

stephherbers:
Sounds good thanks for your input!

PR #4848, opened 2021-08-11
slack deploy notifications

snopoke:
Having been doing a fair number of deploys recently I've decided it would be better to automate the deploy notifications rather than rely on users to `yo` and also remember to update when the deploy is complete.

This PR adds integration with Slack to send deploy notifications. The initial message is posted to the selected channel (configured in `meta.yml` and upon completion the success / failure message is posted as a thread reply to the original message:

Screenshot from 2021-08-10 17-44-49

In future this could be updated include the full diff output which would be useful for non-staging deploys.

##### ENVIRONMENTS AFFECTED
Currently only `staging`

Can be reviewed by commit.


skodde (approved):

docs/monitoring/slack.md
```diff
@@ -0,0 +1,21 @@
...
+
+To configure commcare-cloud to send messages to Slack:
+
+1. Create a [Slack app](https://api.slack.com/authentication/basics) and copy the
+access token.
+2. Set the access token as a commcare-cloud secret:
+
+    Run the following command and paste in the token when prompted:
+    ```shell
+    commcare-cloud <env> secrets edit slack_token
```
dannyroberts, line 14: My intention when writing the `secrets_backend` changes was to keep a list of all possible secrets in https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/environment/secrets/secrets.yml. I forget if I made `commcare-cloud <env> secrets edit slack_token` actually fail if it's not available, but apparently I didn't?

dannyroberts: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/README.md#secrets-management is the closest thing I wrote to docs about this, I think.

src/commcare_cloud/commands/deploy/slack.py
```diff
@@ -0,0 +1,94 @@
...
+            "blocks": blocks
+        }
+        if thread_ts:
+            data["thread_ts"] = thread_ts
+        response = requests.post("https://slack.com/api/chat.postMessage", json=data, headers=headers)
+        try:
+            response.raise_for_status()
+        except RequestException as e:
+            raise SlackException(e)
+        return response.json()["ts"]
```
esoergel, line 65: what's a `ts`?  I assume a message identifier of some kind?  It'd be nice to use a more verbose term on our side

src/commcare_cloud/commands/deploy/slack.py
```diff
@@ -0,0 +1,94 @@
...
+    client.send_deploy_start_message(context)
+
+
+def notify_slack_deploy_end(environment, context, is_success):
+    try:
+        client = SlackClient(environment)
+    except SlackException:
+        return
+
+    client.send_deploy_end_message(context, is_success)
```
esoergel, line 24: What's the rationale behind putting this outside the `try/except`?  It's not that unlikely for this integration to break with a version change or something, so this will force us to fix it, but getting an exception during the deploy is pretty painful.  That used to happen with some regularity on record_deploy_success, and there was often confusion about whether the deploy needed to be redone or whether we needed to do something to make sure the environment was in a good state.

An alternative would be to loudly report the failure, but let the script continue.

esoergel (approved):

src/commcare_cloud/commands/deploy/utils.py
```diff
@@ -44,19 +74,21 @@ def announce_deploy_start(environment, service_name, commcare_rev=None):
...
     send_email(
         environment,
-        subject=f"{service_name} deploy to {environment.name} failed",
+        subject=f"{context.service_name} deploy to {environment.name} failed",
     )
 
 
-def announce_deploy_success(environment, service_name, diff_ouptut):
+def announce_deploy_success(environment, context):
+    notify_slack_deploy_end(environment, context, is_success=False)
```
millerdev, line 86:
```suggestion
    notify_slack_deploy_end(environment, context, is_success=True)
```

src/commcare_cloud/commands/deploy/slack.py
```diff
@@ -0,0 +1,94 @@
...
+        blocks = self._get_message_blocks("*Deploy Started* :rocket:", context)
+        thread_ts = self._post(blocks)
+        context.set_meta_value('slack_thread_ts', thread_ts)
+
+    def send_deploy_end_message(self, context, is_success):
+        thread_ts = context.get_meta_value('slack_thread_ts')
+        if is_success:
+            message = "*Deploy Success* :checkered_flag:"
+        else:
+            message = "*Deploy Success* :x:"
```
millerdev, line 48:
```suggestion
            message = "*Deploy Failed* :x:"
```

docs/monitoring/slack.md
snopoke: acb9cb1ad

src/commcare_cloud/commands/deploy/slack.py
snopoke: yes, it's a message identifier. I do give it a name where it's actually used.

src/commcare_cloud/commands/deploy/slack.py
snopoke: updated to catch errors and output a warning message

esoergel (approved):

PR #4850, opened 2021-08-11
Add new route to terraform constants.py

Charl1996:
[Here's](https://dimagi-dev.atlassian.net/browse/SC-1640) the ticket for the original feature.
Summary:
I added a route which we tell Telerivit to send the status of a SMS to (failed, delivered, etc.), but the route gives a 405 when I test it, although locally it works fine, so I suspect it might be AWS which blocks it for some reason.

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/commands/terraform/constants.py
+++ b/src/commcare_cloud/commands/terraform/constants.py
@@ -38,6 +38,7 @@ COMMCAREHQ_XML_POST_URLS_REGEX = r"""
 ^/jserror/$
 ^/log_email_event/([\w]+)/?$
 ^/telerivet/in/?$
+^/telerivet/status/([\w\-]+)/$
 """.strip().split()
 
 COMMCAREHQ_XML_QUERYSTRING_URLS_REGEX = """


dannyroberts (changes requested):
The code returned by the waf is 403. [405 means Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405), so I'd check if you're using a GET when you need to use a POST or vice versa.

Aside from that (if I'm off base and the solution really is adding this route to the waf) can you confirm you followed the process involving `./manage.py list_waf_allow_patterns` outlined in https://confluence.dimagi.com/display/saas/How+traffic+gets+to+our+system%3A+Global+Accelerator%2C+ALB%2C+and+WAF and link to the corresponding PR?

Charl1996:
@dannyroberts Sorry for replying only now - this issue has not been big priority. 

Regarding the 405 - yes, method not allowed, however, I do use the correct method. I have in the meantime solved the issue. Turns out I was missing a trailing `/` (I'm a bit surprised since usually it's redundant or I would have expected route not found; method error seems a bit misleading)

Charl1996:
I'm going to close this PR then since the issue is not related to AWS.

PR #4852, opened 2021-08-12
Ce/deploy success emails

calellowitz:
This fixes two issues noticed in the recent confusion around an unsuccessful production deploy.

The first commit makes it so that the success email does not send when deploy stops to allow preindex to complete, by raising the exception. I could have just changed the exit code for the command, but this makes it a lot more obvious that the deploy was not successful, hopefully limiting confusing, and bringing it into line with other types of errors.

The second commit changes the success email to always go to admins even if another group is specified, which I think is the correct behavior. For the past 4 months, very few developers have been getting these emails (only those on product@ and Farid), and I think generally it makes sense for these to always go to the same list as other envs.

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -402,7 +402,6 @@ def deploy_checkpoint(command_index, command_name, fn, *args, **kwargs):
 
 def _deploy_without_asking(skip_record):
     try:
-        raise PreindexNotFinished()
         for index, command in enumerate(ONLINE_DEPLOY_COMMANDS):
             deploy_checkpoint(index, command.__name__, execute_with_timing, command)
     except PreindexNotFinished:
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -402,6 +402,7 @@ def deploy_checkpoint(command_index, command_name, fn, *args, **kwargs):
 
 def _deploy_without_asking(skip_record):
     try:
+        raise PreindexNotFinished()
         for index, command in enumerate(ONLINE_DEPLOY_COMMANDS):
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/fab/fabfile.py
```diff
@@ -402,6 +402,7 @@ def deploy_checkpoint(command_index, command_name, fn, *args, **kwargs):
 
 def _deploy_without_asking(skip_record):
     try:
+        raise PreindexNotFinished()
```
dannyroberts, line 405: I think you put this here just to test it and then left it in?

src/commcare_cloud/fab/fabfile.py
```diff
@@ -411,6 +412,7 @@ def _deploy_without_asking(skip_record):
              "and wait for an email saying it's done. "
              "Thank you for using AWESOME DEPLOY.")
         )
+        raise
```
dannyroberts, line 415: Does this end up looking like a crash to the person running the deploy?

dannyroberts: Because if it does you could alternatively use `exit(-1)` or some such, after printing a note for the user?

src/commcare_cloud/fab/fabfile.py
calellowitz: Ugh yes

src/commcare_cloud/fab/fabfile.py
calellowitz: It doesn't look any more like a crash than other deploy errors (we raise those exceptions too in the next block), and as mentioned in the description I considered that but wanted it to be obvious that the deploy failed, since that is effectively the current situation, and it was missed today.

dannyroberts (approved):

PR #4864, opened 2021-08-23
tweaks to slack deploy integration

snopoke:
* Slight change to "Deploy Started" message (remove the emoji)
* Add a reaction to the message on success failure (:heavy_check_mark: or :x: )
* Update docs

### DIFF
--- a/src/commcare_cloud/commands/deploy/slack.py
+++ b/src/commcare_cloud/commands/deploy/slack.py
@@ -44,10 +44,13 @@ class SlackClient:
         if not self.channel:
             raise SlackException("Channel not configured")
 
+        # bot must have joined the channel in order to create reactions
+        self._post("https://slack.com/api/conversations.join", {"channel": self.channel})
+
     def send_deploy_start_message(self, context):
         blocks = self._get_message_blocks("*Deploy Started* :rocket:", context)
-        thread_ts = self._post_blocks(blocks)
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/commands/deploy/slack.py
```diff
@@ -56,6 +59,8 @@ def send_deploy_end_message(self, context, is_success):
         else:
             message = "*Deploy Failed* :x:"
         blocks = self._get_message_blocks(message, context)
+        emoji = "white_check_mark" if is_success else "x"
```
joxl, line 62: Kind of a nit but not really (I don't think it matters that much), the emoji "code names" are easy to miss sprinkled around in message strings. Since they're Slack-specific, it would be nice to store them in a more prominent place and insert them with string formatting.

joxl (approved):

src/commcare_cloud/commands/deploy/slack.py
joxl: Well, string formatting isn't applicable to the specific line I commented on here. I was referring to how they're hard-coded throughout.

millerdev: Maybe use named constants?

millerdev (approved):

millerdev (approved):

PR #4871, opened 2021-08-27
add notification text to slack messages

snopoke:
Add text to the slack messages that will get shown in notifications. This should fix the 'content can not be displayed' messages.

https://stackoverflow.com/a/56327506/632517


snopoke:
> > This should fix the 'content can not be displayed' messages
> 
> @snopoke is this an issue we have hit? I haven't noticed in slack if that is the case, or does it appear in deploy output? Also it is worth noting the slack message does include `Environment: <env> Service: <service> User: <user>` below the message, but I like the idea of having all of that text in the message.

If you have Slack notifications enabled (e.g. desktop notifications) then you'd see the 'content can not be displayed' message. Most notifications only support simple text (and maybe markdown) so the advanced 'blocks' display can't be rendered.

PR #4877, opened 2021-08-30
updating waf delivery stream configuration to write data in s3 in hive format

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-12491
##### ENVIRONMENTS AFFECTED

Staging, India, Production
- As of now, applied only on Staging env.


src/commcare_cloud/terraform/modules/ga_alb_waf/waflogs_add_partitions.py
```diff
@@ -0,0 +1,112 @@
...
+    day = utc_datetime.strftime("%d")
+    hour = utc_datetime.strftime('%H')
+
+    DATABASE = 'default'
+    TABLE = 'waf_logs'
+
+    # S3 constant
+    S3_OUTPUT = f"s3://dimagi-commcare-{environment}-logs"
+    S3_BUCKET = f"dimagi-commcare-{environment}-logs"
+    STREAM_PATH=f"frontend-waf-partitioned-{environment}"
```
dannyroberts, line 91: Are these the only lines that differ from the other similar lambda for formplayer request response logs? I wonder if there is a way to share code between the two. Is there a way to have a single lambda that can be called with different environment variables? If so then maybe these can be environment variables

prem-fissionhq: @dannyroberts sure. will look into it.

prem-fissionhq: @dannyroberts if we use this Partition projection feature from AWS, then we don't need a Lambda function to add new partitions. I'm looking into this. Would like to try this on Staging for waf logs, Please let me  know your thoughts.

https://docs.aws.amazon.com/athena/latest/ug/waf-logs.html#create-waf-table-partition-projection

prem-fissionhq:
@dannyroberts PR is ready for review. 
Before merging the PR we need to make sure that the new required partitioned folder is created in the existing folder and in the background we can initiate the data copy.

src/commcare_cloud/terraform/modules/ga_alb_waf/waflogs_add_partitions.py
dannyroberts: Yeah this sounds great, and it sounds like it's working out. I'll wait until you ping me again before reviewing further then.

dannyroberts: Because I would imagine can delete this lambda now right?

prem-fissionhq: Yes, Danny, we don't need that for waf logs partitioning. 
For Formplayer logs I will create a separate ticket to create the existing table and delete lambda function.

PR #4879, opened 2021-08-30
lower compaction ratio

calellowitz:
https://dimagi-dev.atlassian.net/browse/SAAS-12222

This makes couch compaction more aggressive (when the ratio of total to live docs is lower than the current setting).

Yes, couch calls their compaction process `smoosh`. No, I did not laugh either.

##### ENVIRONMENTS AFFECTED

All

### DIFF
--- a/src/commcare_cloud/ansible/roles/couchdb2/templates/extras.ini.j2
+++ b/src/commcare_cloud/ansible/roles/couchdb2/templates/extras.ini.j2
@@ -14,3 +14,10 @@ use_kill_all = {{ couchdb_rexi_use_kill_all }}
 os_process_timeout = 20000
 ; set limit to previous default of 4GB on 2.3.1
 max_document_size = 4000000000
+
+; compaction settings
+[smoosh.ratio_dbs]
+min_priority = 1.3
+
+[smoosh.ratio_views]
+min_priority = 1.3


skodde (approved):

src/commcare_cloud/ansible/roles/couchdb2/templates/extras.ini.j2
```diff
@@ -14,3 +14,10 @@ use_kill_all = {{ couchdb_rexi_use_kill_all }}
 os_process_timeout = 20000
 ; set limit to previous default of 4GB on 2.3.1
 max_document_size = 4000000000
+
+; compaction settings
+[smoosh.ratio_dbs]
+min_priority = 1.3
+
+[smoosh.ratio_views]
+min_priority = 1.3
```
dannyroberts, line 23: Just to make sure I'm reading this right, conceptually this means views and dbs are compacted once they become 30% than they "need to be", right?

dannyroberts (approved):

src/commcare_cloud/ansible/roles/couchdb2/templates/extras.ini.j2
calellowitz: Yeah, when the size of the db or view is 130% of the size of active data in that db (30% is deleted/updated/etc).

PR #4882, opened 2021-09-01
CouchDB skip restart when only changing config

skodde:
##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/ansible/roles/couchdb2/handlers/main.yml
+++ b/src/commcare_cloud/ansible/roles/couchdb2/handlers/main.yml
@@ -2,10 +2,20 @@
 
 - name: restart couchdb2
   service: name=couchdb2 state=restarted
+  when: "'couchdb_extras' not in ansible_run_tags"
 
 - name: start couchdb2
   service: name=couchdb2 state=started
 
 - name: stop couchdb2
   service: name=couchdb2 state=stopped
+
+- name: config changes
+  pause:
+    prompt: >-
+      The configuration changes won't be applied until CouchDB is restarted
+      (please follow your internal documentation to do that safely), or the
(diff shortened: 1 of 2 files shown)


calellowitz (approved):
It would be great if we could make just the restart conditional, rather than have two separate copy tasks, but this looks fine if that is not possible

src/commcare_cloud/ansible/roles/couchdb2/handlers/main.yml
```diff
@@ -9,3 +9,11 @@
 - name: stop couchdb2
   service: name=couchdb2 state=stopped
 
+- name: config changes
+  pause:
+    prompt: >-
+      The configuration changes won't be applied until CouchDB is restarted
+      (please follow your internal documentation to do that safely), or the
+      runtime configuration modified manually (e.g. via API calls) if
+      applicable. Press Enter to acknowledge and continue
```
dannyroberts, line 18: I think we don't want to happen on first time install, right? Any way to prevent that?

calellowitz (approved):

skodde:
> It would be great if we could make just the restart conditional, rather than have two separate copy tasks, but this looks fine if that is not possible

Addressed this in the last changes, let me know if that works.

src/commcare_cloud/ansible/roles/couchdb2/handlers/main.yml
skodde: This is only presented when run with the 'couchdb_extras' tag, which seems unlikely for a first time install, or are you thinking about a different scenario?

dannyroberts: Oh I see I hadn't digested what the line below meant. That makes sense, all good

PR #4883, opened 2021-09-02
remove formplayer_deploy tag from datadog task

snopoke:
@dannyroberts is there a reason we need to run this during formplayer deploy?


dannyroberts (approved):
Huh, weird. I can't just by looking at the history retrace my steps, so I think I agree, this must have been thrown in there by mistake somehow. Doesn't make sense to include on formplayer deploy

PR #4884, opened 2021-09-02
clean staticfiles before collecting

calellowitz:
https://dimagi-dev.atlassian.net/browse/SAAS-12555

I am not certain this would have prevented the issue in the ticket, and it is probably too late to know for sure, but it seems plausible. Specifically, I have confirmed that if copying a file is interrupted (with `cp`), a partially copied file is left, and I have confirmed that if a partial file exists in the staticfiles directory, it is not replaced with the full file when you run `collectstatic` again.

Beyond that, this seems pretty safe. I don't love that the directory is hardcoded, but we use it elsewhere, and because its hardcoded rather than interpolated there is no risk of it ending up null and trying to delete root.

### DIFF
--- a/src/commcare_cloud/fab/operations/staticfiles.py
+++ b/src/commcare_cloud/fab/operations/staticfiles.py
@@ -53,6 +53,8 @@ def collectstatic(use_current_release=False):
     """
     venv = env.py3_virtualenv_root if not use_current_release else env.py3_virtualenv_current
     with cd(env.code_root if not use_current_release else env.code_current):
+        if not use_current_release:
+            sudo('rm -rf staticfiles')
         sudo('{}/bin/python manage.py collectstatic --noinput -v 0'.format(venv))
         sudo('{}/bin/python manage.py fix_less_imports_collectstatic'.format(venv))
         sudo('{}/bin/python manage.py compilejsi18n'.format(venv))


dannyroberts (approved):

orangejenny (approved):

millerdev:
You probably know this, but this seems likely to have a negative impact on the performance of collectstatic, which will make the deploy process slower. It seems like it would be better to detect when a copy fails and remove that single file (or make the copy atomic by copying to a temp location and then moving when done) rather than removing and re-copying everything.

calellowitz:
> You probably know this, but this seems likely to have a negative impact on the performance of collectstatic, which will make the deploy process slower. It seems like it would be better to detect when a copy fails and remove that single file (or make the copy atomic by copying to a temp location and then moving when done) rather than removing and re-copying everything.

Fortunately, that is not the case. `staticfiles` does not exist in the newly created releases on all the webworkers, so the new line will noop. The only time it will end up actually deleting files is if a deploy failed partway through collecting static. In that case the resumed deploy may indeed be slightly slower but I think that is worth it for the certainty that it works correctly, and certainly feels safer than trying to change how these built in django commands work internally.

Interestingly, the one machine that does have a partially filled `staticfiles` directory is the proxy, which copies some of the compressed js over earlier in the process, but the `compress` operation actually runs faster on this branch than with the behavior from master (consistently `40s` vs `50s` on staging), I suspect because it's internal diff mechanism must not be very efficient and it is able to make optimizations when it knows the directory must be created from scratch.

PR #4900, opened 2021-09-16
Handle exclude_ucrs option

sravfeyn:
Pillows that handle UCRs support this option, so adding that support to config.
Depends on https://github.com/dimagi/commcare-cloud/pull/4900


sravfeyn:
@proteusvacuum

snopoke:
The way we've done this in the past is to define a local pillow in commcare-cloud using the `LOCAL_PILLOWTOPS` setting and then reference that in `app_processes.yml`:

**public.yml**
```
localsettings:
  LOCAL_PILLOWTOPS:
    custom:
      - name: 'case-pillow-custom',
        class: 'pillowtop.pillow.interface.ConstructedPillow',
        instance: 'corehq.pillows.case.get_case_pillow',
        params:
           exclude_ucrs:
             - dont_run_me
```

**app-processes.yml**
```
pillowtop:
    case-pillow-custom:
      num_processes: 1
```

sravfeyn:
True. For self hosters it's much easier to turn on a flag like this to use the disable-UCR feature than to add a custom pillow.

sravfeyn:
Bumping to check if this is g2g

sravfeyn:
@proteusvacuum

snopoke (approved):

PR #4901, opened 2021-09-16
Streamline control install

dannyroberts:
This PR is geared towards making the steps outlined in this section https://dimagi.github.io/commcare-cloud/setup/manual_installation.html#setup (which this PR also edits for clarity) cleaner to follow as the complete personal control machine environment setup for new users.

##### ENVIRONMENTS AFFECTED
None

### DIFF
--- a/control/check_install.sh
+++ b/control/check_install.sh
@@ -46,19 +46,16 @@ fi
 
 if [ ! -f "${FAB_CONFIG}" ]
 then
-    for OLD_FAB_CONFIG_DIR in "${COMMCARE_CLOUD_REPO}/src/commcare_cloud/fab"
-    do
-        if [ -d "${OLD_FAB_CONFIG_DIR}" ]
+    OLD_FAB_CONFIG_DIR="${COMMCARE_CLOUD_REPO}/src/commcare_cloud/fab"
+    if [ -d "${OLD_FAB_CONFIG_DIR}" ]
+    then
+        OLD_FAB_CONFIG="${OLD_FAB_CONFIG_DIR}/config.py"
+        if [ -f "${OLD_FAB_CONFIG}" ]
         then
-            OLD_FAB_CONFIG="${OLD_FAB_CONFIG_DIR}/config.py"
-            if [ -f "${OLD_FAB_CONFIG}" ]
-            then
-                cp "${OLD_FAB_CONFIG}" "${FAB_CONFIG}"
-                printf "${YELLOW}→ Copied $(realpath ${OLD_FAB_CONFIG}) to ${FAB_CONFIG}\n"
-                break
-            fi
+            cp "${OLD_FAB_CONFIG}" "${FAB_CONFIG}"
+            printf "${YELLOW}→ Copied $(realpath ${OLD_FAB_CONFIG}) to ${FAB_CONFIG}\n"
         fi
-    done
+    fi
 else
     printf "${GREEN}✓ ${FAB_CONFIG} exists\n"
 fi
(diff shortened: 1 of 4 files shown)


control/check_install.sh
```diff
@@ -45,13 +46,11 @@ fi
 
 if [ ! -f "${FAB_CONFIG}" ]
 then
-    for OLD_FAB_DIR in "${COMMCARE_CLOUD_REPO}/../commcare-hq/deployment/commcare-hq-deploy" \
-                       "${COMMCARE_CLOUD_REPO}/../commcare-hq-deploy" \
-                       "${COMMCARE_CLOUD_REPO}/fab"
+    for OLD_FAB_CONFIG_DIR in "${COMMCARE_CLOUD_REPO}/src/commcare_cloud/fab/config.py"
```
millerdev, line 49: This is a file path, not a directory, which I think will not work correctly below. Should this `for`-loop be eliminated?

src/commcare_cloud/config.example.py
```diff
@@ -1,4 +1,8 @@
 # To generate a Github apikey follow these instructions: https://github.com/blog/1509-personal-api-tokens
 # Minimally you will need a key with repo -> public_repo
 
+# On a control machine shared with others,
+# leave this as None to be prompted for a token each time instead,
+# to avoid storing private github token where even trusted others could access it.
```
millerdev, line 6: :+1: 

nit
```suggestion
# to avoid storing private github token where others could access it.
```

src/commcare_cloud/github.py
```diff
@@ -52,6 +52,10 @@ def get_github_credentials(repo_name, repo_is_private, require_write_permissions
             f"    $ cp {PROJECT_ROOT}/config.example.py {PROJECT_ROOT}/config.py\n"
             f"Then edit {PROJECT_ROOT}/config.py"
         )
+        print(color_notice(
+            "To generate a Github apikey follow these instructions: https://github.com/blog/1509-personal-api-tokens\n"
```
skodde, line 56:
```suggestion
            "To generate a GitHub access token, follow these instructions: https://github.com/blog/1509-personal-api-tokens\n"
```

control/check_install.sh
dannyroberts: Oops, sorry this is a copy-paste error. I'll fix this

dannyroberts: Okay updated and tested this directly. Thanks for catching

dannyroberts: Addressing the for-loop suggestion as well

dannyroberts: Addressed

control/check_install.sh
```diff
@@ -45,30 +46,24 @@ fi
...
-    for OLD_FAB_DIR in "${COMMCARE_CLOUD_REPO}/../commcare-hq/deployment/commcare-hq-deploy" \
-                       "${COMMCARE_CLOUD_REPO}/../commcare-hq-deploy" \
-                       "${COMMCARE_CLOUD_REPO}/fab"
-    do
-        if [ -d "${OLD_FAB_DIR}" ]
+    OLD_FAB_CONFIG_DIR="${COMMCARE_CLOUD_REPO}/src/commcare_cloud/fab"
+    if [ -d "${OLD_FAB_CONFIG_DIR}" ]
+    then
+        OLD_FAB_CONFIG="${OLD_FAB_CONFIG_DIR}/config.py"
+        if [ -f "${OLD_FAB_CONFIG}" ]
```
dannyroberts, line 53: @skodde said (offline):
> since you removed the for loop, you could also remove having the check separately on the dir and the config file, and keep only the latter

skodde (approved):

control/check_install.sh
dannyroberts: Updated :)

PR #4961, opened 2021-11-03
Allow Python to be upgraded

millerdev:
Change ansible and fabric infrastructure to allow the Python version to be parameterized per environment. This does not actually change the version of Python yet, it only lays the groundwork for future upgrades.

To ease the upgrade process in the future a new `python_env` symlink is being created that points to the active Python virtualenv for a given commcare-hq release. A new virtualenv can be created and the symlink updated to point to it with the following steps:
- Update `python_version` and possibly `use_deadsnakes` in `environments/<env>/public.yml`
- Run the following command
  ```bash
  cchq <env> update-supervisor-confs --tags=python,sudoers
  ```
  - `update-supervisor-confs` is used instead of `deploy-stack` so services can be restarted immediately. Answer `y` to the prompt to "update supervisor to use the new configurations."
  - The `sudoers` tag is only needed when upgrading from Python 3.6.

Request for comments: `cchq <env> update-users` takes a long time, ~and `cchq <env> bootstrap-users` will do the job (update the sudoers file, see cbd475ba5dd1c4318e264d45028380d9838a7674 commit message). However, `bootstrap-users` applies directly without a `--check` run first, which means it could do additional unexpected things without warning. Is there a better (faster) command than `update-users` that will run with a `--check` and updates sudoers?~ Resolution: `--tags=python,sudoers`

Tested on staging.

:blowfish: Review by commit.

##### ENVIRONMENTS AFFECTED
All: various commands will create a `python_env` symlink and `update-supervisor-confs` will cause supervisor configs to change to use the symlink after this is merged. The impact of the sudoers file change is not clear to me. ~It may be necessary to run `update-users` after this is merged so that some (management?) commands continue to work.~


dannyroberts (approved):
I think I missed it; what makes running `update-users` required here? Is it that that calls common installs or some other thing that could instead be called directly, or is it something integral to the actual thing within `update-users` that takes so much time (copying everyone's public keys and stuff like that)?

dannyroberts:
I see
> One use of `virtualenv_home` that is not addressed by this commit is in src/commcare_cloud/ansible/roles/bootstrap-users/templates/sudoers.j2. This means that `cchq <env> update-users` will be required when upgrading from Python 3.6.

Could you find the task that runs that and add a tag you can use to run just that task?

millerdev:
> Could you find the task that runs that and add a tag you can use to run just that task?

Brilliant! It appears there is already a tag just for that, which simplifies things quite a bit (and one less command to run). Before I looked into this I did not realize that `update-users` was `deploy_stack.yml --tags=users` command.

calellowitz (approved):

dannyroberts:
Seeing this note:
> The impact of the sudoers file change is not clear to me.
is there a chance that running `update-supervisor-confs` or some other command on india or production right now will cause an issue? I'm not totally following, but is there a way to test the part of this you're no sure of to make sure it's safe?

millerdev:
@dannyroberts There is a chance. I'm not sure how the `Cmnd_Alias PYTHON` line is used in practice. I think it allows some users to execute Python commands, but it is unclear when that is done during normal operations.

PR #4963, opened 2021-11-03
Add roles to docs

kaapstorm:
Adds operation roles to documentation. Content is originally from partner-only docs.

NOTE: This branches off **sr/docs** 

Jira: [SC-1819](https://dimagi-dev.atlassian.net/browse/SC-1819)

##### ENVIRONMENTS AFFECTED

None. This change is documentation only.


kaapstorm:
Easier to review [rendered by GitHub](https://github.com/dimagi/commcare-cloud/blob/8ce8c36a8e9e8e58b938b748827f2c23af2a79b2/hosting_docs/source/operations/8-roles.rst).

Charl1996:
@kaapstorm This is a neat doc!

I was just reviewing the ticket and saw bullet point nr. 2,
> Relevant information which is not already included in CommCare-cloud ReadTheDocs is added into ReadTheDocs (for example - go live checklist, prerequisites, architecture overview, etc)

This PR explains the Roles and Responsibilities so I don't think the content of my question should be included here, but I was just wondering what the expected outcome of that bullet point would look like? (it definitely sounds like something we would already have, but was just curious what your findings were)

sravfeyn:
This is great, thanks! 

I think it fits nicely here in the flow https://commcare-cloud.readthedocs.io/en/latest/installation/prereqs/4-requirements.html

proteusvacuum (approved):

PR #4967, opened 2021-11-05
It's "CommCare HQ"

kaapstorm:
Maybe it's because I didn't get enough sleep on Wednesday night. Maybe it was [that funny post about PHP](https://aloneonahill.com/blog/if-php-were-british/). Maybe it's The Zen Of Python,

> There should be one-- and preferably only one --obvious way to do it.

Maybe it's just Friday.

commcare_hq_01

commcare_hq_02

commcare_hq_03

It's two words.

##### ENVIRONMENTS AFFECTED
None. Documentation only.


proteusvacuum (approved):
👀

PR #4969, opened 2021-11-05
Make --control use -tt to pass through stdin

dannyroberts:
According to the ssh docs:
> force tty allocation, even if ssh has no local tty.

##### ENVIRONMENTS AFFECTED
None

### DIFF
--- a/src/commcare_cloud/commcare_cloud.py
+++ b/src/commcare_cloud/commcare_cloud.py
@@ -116,7 +116,7 @@ def run_on_control_instead(args, argv, force_latest_code):
             '&& {cchq} {cchq_args}'
         )
     cmd_parts = [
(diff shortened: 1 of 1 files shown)


dannyroberts:
@millerdev I think I originally made this as a way to recreate scp behavior using redirects through `commcare-cloud --control ...`. Do you know if we found a better workaround already? If not, maybe it's still worth merging.

millerdev (commented):
The "better" way to do things like `scp` probably involves commcare-cloud-managed SSH config files, which would allow us to run commands directly like
```sh
scp staging-celery:path/to/file ~/Downloads/
```
rather than wrapping every different command (like `rsync`, `scp`, etc.) with custom commcare-cloud code. @joxl has a ideas about how to do this.

PR #4971, opened 2021-11-06
Set security related headers in nginx only, remove from django

gherceg:
This was brought up as part of [this ticket](https://dimagi-dev.atlassian.net/browse/SAAS-12042). I didn't realize that duplicate headers were being returned, so we need to change our approach. 

In a perfect world, we could set these values in django to take advantage of up to date security warnings when services are checked, in addition to setting these values in nginx to take advantage of being on the perimeter of our system. Unfortunately it looks like we have to compromise, and it is more important to have these values set in nginx so they apply to all requests, not just requests coming back from django services. 

I removed the security related variables from the local settings template, and silenced the django warnings that resulted from not setting these security vars. I saw that Simon had already added the XSS and Content Type headers to `cchq_ssl.yml`, so I added the strict transport security header there as well.  I assume that is preferable to adding directly to the site.j2 file anyway. 

I've updated staging with this new setup by running the following:
1. `cchq --control staging update-config --branch=gh/only-set-hsts-in-nginx` to remove the django security headers
2. `cchq --control staging fab restart_services` to make new local settings changes stick
3. `cchq --control staging ansible-playbook deploy_proxy.yml --tags=nginx_sites --limit proxy --branch gh/only-set-hsts-in-nginx` to update nginx config file on proxy machines.

I checked the `staging.commcarehq.org` domain using https://httpstatus.io/ and the results look good.

Screen Shot 2021-11-07 at 9 43 03 AM

##### ENVIRONMENTS AFFECTED

Staging, Swiss, India, Production

### DIFF
--- a/environments/swiss/public.yml
+++ b/environments/swiss/public.yml
@@ -129,11 +129,6 @@ localsettings:
   #  PILLOWTOP_MACHINE_ID:
   LOCAL_MIDDLEWARE:
     - 'django.middleware.security.SecurityMiddleware'
-  SILENCED_SYSTEM_CHECKS:
-    - 'security.W004' # we handle Strict-Transport-Security in nginx (see .../roles/nginx/vars/cchq_ssl.yml)
-    - 'security.W006' # we handle X-Content-Type-Options in nginx (see .../roles/nginx/vars/cchq_ssl.yml)
-    - 'security.W007' # we handle X-XSS-Protection in nginx (see .../roles/nginx/vars/cchq_ssl.yml)
-    - 'security.W008' # the ALB and nginx already redirect HTTP to HTTPS so safe to silence this warning
   REDIS_DB: '0'
   REDIS_HOST: "localhost"
   REDIS_PORT: '6379'
--- a/environments/swiss/public.yml
+++ b/environments/swiss/public.yml
@@ -130,9 +130,9 @@ localsettings:
   LOCAL_MIDDLEWARE:
     - 'django.middleware.security.SecurityMiddleware'
   SILENCED_SYSTEM_CHECKS:
-    - 'security.W004' # we handle Strict-Transport-Security in nginx (see .../roles/nginx/templates/site.j2)
-    - 'security.W006' # we handle X-Content-Type-Options in nginx (see .../roles/nginx/templates/site.j2)
-    - 'security.W007' # we handle X-Xss-Protection in nginx (see .../roles/nginx/templates/site.j2)
+    - 'security.W004' # we handle Strict-Transport-Security in nginx (see .../roles/nginx/vars/cchq_ssl.yml)
+    - 'security.W006' # we handle X-Content-Type-Options in nginx (see .../roles/nginx/vars/cchq_ssl.yml)
+    - 'security.W007' # we handle X-XSS-Protection in nginx (see .../roles/nginx/vars/cchq_ssl.yml)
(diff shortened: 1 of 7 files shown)


skodde (approved):

gherceg:
Somewhat related, I noticed `cas_ssl.yml` as well. Could we remove that file?

snopoke:
> Somewhat related, I noticed `cas_ssl.yml` as well. Could we remove that file?

there are probably references to that somewhere that would also need to be removed

environments/swiss/public.yml
```diff
@@ -129,8 +129,10 @@ localsettings:
   #  PILLOWTOP_MACHINE_ID:
   LOCAL_MIDDLEWARE:
     - 'django.middleware.security.SecurityMiddleware'
-  SECURE_HSTS_SECONDS: "{{ nginx_hsts_max_age }}"
   SILENCED_SYSTEM_CHECKS:
```
snopoke, line 132: since these apply globally to all envs it might make more sense to set it directly in the localsetting template instead of having to copy paste in all env files.

gherceg: Yeah that is a good idea, and I just made that change. When you say _all envs_, that is not entirely true right? I don't have great context on the remaining envs outside of staging, swiss, india, and prod, but do we need to take those into consideration?

snopoke: All envs use nginx so they should be covered by the nginx changes (assuming they update their nginx config)

gherceg: But is there any chance that other envs don't enforce HTTPS at all endpoints?

snopoke: I think the approach should be that a default install of CommCare has this set up, either at the nginx level or at the Django level. If a 3rd party is doing something custom then it is up to them to ensure appropriate security. 

Besides that, is there even a way to enable these security features in Django through commcare-cloud?

gherceg: > I think the approach should be that a default install of CommCare has this set up, either at the nginx level or at the Django level. If a 3rd party is doing something custom then it is up to them to ensure appropriate security.

That makes sense. 

> Besides that, is there even a way to enable these security features in Django through commcare-cloud?

I don't fully understand the question. Once using django security middleware, [these settings](https://docs.djangoproject.com/en/3.2/topics/security/#ssl-https) can be used in a `settings.py` file. If your question is asking specifically from a 3rd party perspective, I'm not sure. I'm assuming they would fork commcare-cloud, and in that case have control of security so my initial point on non-https endpoints is moot. Thanks Simon.

gherceg:
I realized that the only need to silence these warnings is because we include django security middleware, which we only added in the first place to set strict transport security values. I can see an argument for keeping it in that as new warnings come up, we would be notified of them when services are checked, but curious what others think.

environments/swiss/public.yml
snopoke: >> Besides that, is there even a way to enable these security features in Django through commcare-cloud?

> I don't fully understand the question. Once using django security middleware, these settings can be used in a settings.py file. If your question is asking specifically from a 3rd party perspective, I'm not sure. I'm assuming they would fork commcare-cloud, and in that case have control of security

Yea, I was meaning without forking, is there a way to change things with just config in commcare-cloud.

gherceg:
Running update-config on all envs. Will rely on the pending deploy to handle updating nginx config. @skodde that should be sufficient right?

PR #4976, opened 2021-11-11
Add includeSubDomains to Strict-Transport-Security header

gherceg:
Followup to [the PR](https://github.com/dimagi/commcare-cloud/pull/4971) that moved all security related header setup to nginx. I was cautious about including `includeSubDomains` in the header, but in hindsight I'm not sure why since we were already doing this at the django level. Alas, here we are so this is a very small PR that I plan to roll out by running `cchq --control <env> ansible-playbook deploy_proxy.yml --tags=nginx_sites --limit proxy`
##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -55,7 +55,7 @@ nginx_sites:
    - "X-Forwarded-Protocol  $scheme"
    access_log: "{{ log_home }}/{{ deploy_env }}-timing.log timing"
    add_header:
-    - 'Strict-Transport-Security "max-age={{ nginx_hsts_max_age|default(0) }}"'
+    - 'Strict-Transport-Security "max-age={{ nginx_hsts_max_age|default(0) }}; includeSubDomains"'
     - 'X-XSS-Protection "1; mode=block"'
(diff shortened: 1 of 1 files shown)


snopoke (approved):

millerdev:
@gherceg Would it make sense to publish a change log entry so externally hosted environments know to run this maintenance command?

gherceg:
@millerdev thanks for the comment. I didn't think about self hosters but should have. I just made [this PR](https://github.com/dimagi/commcare-cloud/pull/5028) since it would be the last in this sequence of HSTS PRs. 

Since this change is on the nginx level, it implies that we are limiting self hosters to using HTTPS all the time which seems like a good thing, but want to double check that this is indeed what we want. @dannyroberts @skodde double checking with you guys on this, and if a change log entry should be published externally.

dannyroberts:
I think anyone who's using the standard commcare-cloud deployment is already forced to use HTTPS for commcare-cloud (HTTP redirects to HTTPS for example). But it's conceivable that they could have a subdomain of the domain that they use for CommCare that uses HTTP. We wouldn't _recommend_ it, as HTTP is subject to man-in-the-middle attacks as well well as not protecting sensitive data, if any; but someone could do it, and it's probably not our place to stop them? I think the easiest solution would be to just make this an option in public.yml and set it to `yes` for our envs.

PR #4981, opened 2021-11-16
use correct flway config property

snopoke:
This was not the correct property name. It is working with the current version of Spring because Spring falls back to the main datasource properties. From Spring v2.5 the fallback is removed: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.5-Release-Notes#flyway-and-liquibase-jdbc-urls

Docs for correct property: https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.data-migration.spring.flyway.user

See also [Spring Upgrade PR](https://github.com/dimagi/formplayer/pull/1023)


snopoke:
Note: this is already rolled out to staging

dannyroberts (approved):

PR #4988, opened 2021-11-19
Add a celery node

skodde:
Add a celery node by increasing the instance count from 7 to 8
and assign it to the main group.

https://dimagi-dev.atlassian.net/browse/SAAS-12815

##### ENVIRONMENTS AFFECTED
Production


skodde:
> Doesn't this need to update inventory and aws-resources files as well

Yes, I thought it would be best to get a preliminary approval before setting up the instance and then committing the subsequent changes.

calellowitz (approved):

calellowitz (approved):

dannyroberts (approved):

dannyroberts:
> Yes, I thought it would be best to get a preliminary approval before setting up the instance and then commit the subsequent changes.

@skodde @calellowitz this is exactly correct. Since part of the PR needs to be made after the the instance is created, and our policy is to have all code reviewed and approved before it is applied, it's correct to first get an approval to apply the change, then update the files as needed after the change is applied, followed by a final review (which in the standard situation is just a review of autogenerated files) and then merging.

PR #5013, opened 2021-12-07
Use default python_version from group_vars/all.yml

millerdev:
##### ENVIRONMENTS AFFECTED
All, since https://github.com/dimagi/commcare-cloud/pull/5006 has been merged.

### DIFF
--- a/src/commcare_cloud/environment/paths.py
+++ b/src/commcare_cloud/environment/paths.py
@@ -4,6 +4,7 @@ import os
 import sys
 from distutils.sysconfig import get_python_lib
 from io import open
+from pathlib import Path
 
 import yaml
 from memoized import memoized, memoized_property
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/environment/paths.py
```diff
@@ -49,6 +51,10 @@ def get_env_file_path(self, filename):
     def public_yml(self):
         return self.get_env_file_path('public.yml')
 
+    @lazy_immutable_property
+    def group_vars_all_yml(self):
+        return self.ansible_path / "group_vars/all.yml"
```
calellowitz, line 56: Is this syntax new? I assume `Path` is overriding `/` to concatenate the parts, but I have never seen this before.

calellowitz (approved):
Approved assuming the answer to my question is yes.

src/commcare_cloud/environment/paths.py
millerdev: `Path` was introduced in Python 3.4. It's really nice.

PR #5014, opened 2021-12-07
Better passwords

kaapstorm:
This PR swaps the uuid library out for the [secrets](https://docs.python.org/3/library/secrets.html) library (new in Python 3.6) which is intended for managing passwords, tokens, etc.

##### ENVIRONMENTS AFFECTED

None. This change only affects commcare-cloud-bootstrap, tools for bootstrapping new environments.

### DIFF
--- a/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
+++ b/commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
@@ -5,11 +5,11 @@ import json
 import os
 import random
 import re
+import secrets  # New in 3.6
 import shutil
 import string
 import subprocess
 import sys
-import uuid
 from io import open
 
 import jinja2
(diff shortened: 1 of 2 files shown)


commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
```diff
@@ -1,22 +1,18 @@
-from __future__ import absolute_import, print_function, unicode_literals
-
 import argparse
 import json
 import os
 import random
 import re
+import secrets  # New in 3.6
```
millerdev, line 6: Because I'm looking at the PR description I know this means "Python 3.6" but that may not be obvious in other contexts.
```suggestion
import secrets  # New in Python 3.6
```

millerdev (approved):

commcare-cloud-bootstrap/commcare_cloud_bootstrap.py
kaapstorm: Yeah. After I wrote the PR description I was wondering whether we really need this comment at all. :shrug: Keeping it.

PR #5015, opened 2021-12-08
Instructions for installing a new cluster

kaapstorm:
Part of self-hosting onboarding material.

Easier to review [rendered doc](https://github.com/dimagi/commcare-cloud/blob/nh/new_cluster/docs/setup/new_cluster.md).

:office: 

##### ENVIRONMENTS AFFECTED

None. Documentation only.


kaapstorm:
Incorporated lessons from last week. Dropping "do not merge".

docs/setup/new_cluster.md
```diff
@@ -0,0 +1,695 @@
...
+| webworker1 |     2 | 8192 MiB | 30 GiB                  |
+| webworker2 |     2 | 8192 MiB | 30 GiB                  |
+| db1        |     2 | 8192 MiB | 30 GiB + 60 GiB         |
+| db2        |     2 | 8192 MiB | 30 GiB + 60 GiB + 20 GiB|
+
+db1 has additional storage for databases, and db2 has one extra volume
+for databases, and another for a shared NFS volume.
+
+
+(Example) Creating a small cluster using libvirt with KVM
```
proteusvacuum, line 30: I'm not sure we should include this, as we don't support the VM level stuff on the machines (and in fact, we prefer you use some cloud hosting provider to do this for you). If you wanted to keep these docs around for us to reference, I'd stick them in confluence.

docs/setup/new_cluster.md
```diff
@@ -0,0 +1,695 @@
...
+
+   3. Restart the VM.
+
+            $ virsh start webworker1
+
+Configuration of your virtual machines is complete. Adapt the
+instructions above for your requirements and your virtualization stack.
+
+
+Prepare all VMs for automated deploy
```
proteusvacuum, line 250: From here downwards seems very similar to the monolith installation docs. Maybe we can reference that, and just point out what is different for a cluster?

docs/setup/new_cluster.md
kaapstorm: Confluence might be a good idea. I'd like to keep this around for reference, because setting this up myself was really useful for working through the process we are expecting third-party / on-prem hosting providers to follow. Some projects who want to self-host are doing so explicitly because they don't want to host in the cloud.

docs/setup/new_cluster.md
kaapstorm: Referencing the monolith instructions could be informative. But I'd like to push back against "just point out what is different". I don't think that is a more reader-friendly approach.

In fact, I would offer our HQ [DEV_SETUP](https://github.com/dimagi/commcare-hq/blob/master/DEV_SETUP.md) documentation as a good example of what _not_ to do.

Here is how Slack does it: First you [choose your OS](https://slack.com/intl/en-gb/help/categories/360000049043-Getting-started#download-slack). Then, if you choose Linux (for example) there are similar, but separate, [instructions for each distro](https://slack.com/intl/en-gb/help/articles/212924728-Download-Slack-for-Linux--beta-#ubuntu-1). You can imagine how unreadable those instructions would be if Slack had taken our approach:

> ...
> Step 4:
> 1. (Ubuntu) Double-click the file to view Slack in Ubuntu Software Centre.
> 2. (Fedora) Open the file in your package manager.
> 3. (RHEL) Open Terminal and run the following commands ...
>
> ...

DEV_SETUP should start off with by linking to the right section for your operating system, and then present only the instructions applicable to you. Sure, there will be a lot of duplication. But it would be a whole lot more readable.

Readability is what I hoped to achieve with the structure I chose here.

proteusvacuum: OK, but I think the only difference here is how you set up the inventory file, right? This seems like a lot of overhead to manage two of the same document when I think this could just be a subsection of "setting up a new installation". It would then be like the tabs in slack.

kaapstorm: Hmmm. That sounds fair.

I'll merge them into a "setting up a new installation" doc.

kaapstorm: @proteusvacuum [a20e336](https://github.com/dimagi/commcare-cloud/pull/5015/commits/a20e336e4354bc4e8ea64853c569c5b23d3e2769)

Probably easier to review the [merged doc as rendered](https://github.com/dimagi/commcare-cloud/blob/a20e336e4354bc4e8ea64853c569c5b23d3e2769/docs/setup/new_environment.md).

Charl1996:
My only comment would be the last question [here](https://github.com/dimagi/commcare-cloud/pull/5037/files#r788506730).

proteusvacuum (approved):
Love it, thanks!

PR #5019, opened 2021-12-08
Bump concurrency from 20 to 40 for the reminder_queue

gherceg:
The reminder_queue takes ~an hour to process. Last week, we bumped concurrency to 40 to handle a larger than usual backlog, but that change was reverted. Now that we are consistently seeing slower times, it makes sense to bump this value permanently.  
##### ENVIRONMENTS AFFECTED

Production

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -47,7 +47,7 @@ celery_processes:
       concurrency: 1
     reminder_queue:
       pooling: gevent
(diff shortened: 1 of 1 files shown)


skodde (approved):

gherceg:
It turns out the underlying issue was due to some workers being blocked, so this PR is no longer needed.

dannyroberts (approved):

PR #5037, opened 2021-12-15
Self-hosting training plan

kaapstorm:
Note that this PR is based off the **sr/docs** branch.

Easier to review [the rendered document](https://github.com/dimagi/commcare-cloud/blob/nh/training_plan/hosting_docs/source/training/0-index.rst).

Please let me know if you think there is anything missing, prerequisites I've skipped, useful links I should add, etc.

##### ENVIRONMENTS AFFECTED

None. Documentation only.

### DIFF
--- a/hosting_docs/source/training/0-index.rst
+++ b/hosting_docs/source/training/0-index.rst
@@ -51,9 +51,7 @@ CommCare HQ DevOps engineer.
 |                                                         |         |
 +---------------------------------------------------------+---------+
 | A detailed understanding of the installation process:   | 1 hour  |
-| See "Setting up a new CommCare HQ environment           |         |
-| `on a single machine`_" and "Setting up a new CommCare  |         |
-| HQ environment `on a small cluster`_".                  |         |
+| See `Setting up a new CommCare HQ environment`_.        |         |
 |                                                         |         |
 +---------------------------------------------------------+---------+
 | `Server management basics`_                             | 15 mins |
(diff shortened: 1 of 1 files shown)


hosting_docs/source/training/0-index.rst
```diff
@@ -0,0 +1,97 @@
...
+
++---------------------------------------------------------+---------+
+| A high-level understanding of the services that make up | 1 hour  |
+| CommCare HQ. See the `CommCare HQ Services Overview`_.  |         |
+|                                                         |         |
++---------------------------------------------------------+---------+
+| A detailed understanding of the installation process:   | 1 hour  |
+| See "Setting up a new CommCare HQ environment           |         |
+| `on a single machine`_" and "Setting up a new CommCare  |         |
+| HQ environment `on a small cluster`_".                  |         |
```
Charl1996, line 56: I get a 404 when accessing [this](https://dimagi.github.io/commcare-cloud/setup/new_cluster.html) link?

kaapstorm: Nice catch @Charl1996 . It is [this file](https://github.com/dimagi/commcare-cloud/blob/nh/new_cluster/docs/setup/new_cluster.md) added by (unmerged) PR #5015

Charl1996 (approved):

hosting_docs/source/training/0-index.rst
Charl1996: Out of curiosity...

Since [this](https://github.com/dimagi/commcare-cloud/blob/a20e336e4354bc4e8ea64853c569c5b23d3e2769/docs/setup/new_environment.md) PR touches both monolith and small cluster, does it make sense to use the same link for `on a single machine` and `on a small cluster`?

Also, would it not be..."better"...to link `on a single machine` to the already existing hosting docs [Single Server Production Install](https://commcare-cloud.readthedocs.io/en/latest/installation/monolith/2-install-prod.html)? (Especially because this PR's doc sits in hosting docs)

It seems with [this](https://github.com/dimagi/commcare-cloud/blob/a20e336e4354bc4e8ea64853c569c5b23d3e2769/docs/setup/new_environment.md) PR and [this](https://commcare-cloud.readthedocs.io/en/latest/installation/monolith/2-install-prod.html) docs we're saying the same things in 2 different documents...unless there's a good reason for segregating the different docs?

kaapstorm: Thank you for the heads-up that I should change this too!

Yeah, the fact that we say exactly the same thing in two places in two different ways is ridiculous. Think of the wasted hours that went into that. Now think of the wasted hours about to be spent fixing it.

kaapstorm: @Charl1996  f61bd231cae6eb6593eb923fa22e3fb43eafd182

kaapstorm: (Please excuse the kneejerk response about wasted hours. The new "2-install-prod" doc looks like it's closely based on (a copy of?) the older "new_environment" doc. So not a complete waste of time. Hopefully merging it with the newly merged "new_environment" doc won't be a lot of effort.)

PR #5039, opened 2021-12-17
Decrease malt_generation_queue worker concurrency

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-12878

Having all ~64 tasks run at the same time overwhelms Elasticsearch, which in turn impacts other areas of the system that depend on ES. By reducing concurrency, we are limiting the number of malt related tasks running at the same time by half, which will hopefully give ES some breathing room. 
##### ENVIRONMENTS AFFECTED

Production


calellowitz (approved):

environments/production/app-processes.yml
```diff
@@ -93,7 +93,7 @@ celery_processes:
       num_workers: 2
     malt_generation_queue:
       pooling: gevent
-      concurrency: 40
+      concurrency: 20
```
dannyroberts, line 96: 7 machines * 40 * 2 = 560 tasks running at a time.
7 machines * 20 * 2 = 280 tasks running at a time. That's still probably too much and maybe more than the total tasks? I might suggest
```
      concurrency: 3
      num_workers: 1  # this is the default so you can also just exclude the property
```
for a total concurrency of 21.

Although the problem may have to do more with the individual queries than how many are running at the same time, so if that's the case then even this lower concurrency won't help and the fix has to be made in the code.

dannyroberts: Cal pointed out to me that I'm wrong and that this is running on one single machine not 7. So the numbers are just 80 => 40. Thanks Cal

calellowitz (approved):

gherceg:
I reduced the concurrency more with the following reasoning. Originally we had a concurrency of 80, but since the malt task chunks domains up into 1k sized tasks, and we have ~64k domains in production, we only spawned off 64 tasks. If that overloaded ES, I doubt 40 simultaneous tasks would be much better. Going to try 20 out this weekend, and will make sure to purge the malt generation queue before adding these workers back.

gherceg:
Given that flower is not functioning properly, and I am fresh off of a back and fourth with USH about minimizing risk this weekend, I am going to walk back my previous statement about trying this out this weekend.

gherceg:
Updated to basically remove all concurrency, including only having one worker. This would put the run time for updating the MALT on all domains close to what it was before, but there is a performance improvement in the pipeline that might give us more confidence in bumping concurrency, albeit still cautiously.

skodde (approved):

gherceg:
Alright. Concurrency of 2. Final answer. It's nice to avoid a situation where the tasks are completely blocked by a longer running task, so 2 or more is ideal instead of 1. And 2 will certainly not overload ES especially with recent optimizations.

dannyroberts (approved):

PR #5043, opened 2021-12-17
Reject `cchq <env> service commcare restart`

dannyroberts:
This was prompted by a history of occasional operation errors resulting from unclear guidance (both on the team and from the tool itself) on which command to use to restart webworkers and/or services based on commcare-hq code. `cchq <env> service commcare restart` is a trap because (1) it does a non-rolling restart of webworkers, resulting in webworker downtime (2) it does a formplayer restart which always causes downtime. Similarly `cchq <env> service webworkers restart` has problem 1 as well.

This change makes those two commands (specifically/only with the restart action) to display an error message suggesting alternatives and then refuse to proceed (exiting with return code 1).

Screen Shot 2021-12-17 at 2 13 33 PM

##### ENVIRONMENTS AFFECTED
None

### DIFF
--- a/src/commcare_cloud/commands/ansible/service.py
+++ b/src/commcare_cloud/commands/ansible/service.py
@@ -467,7 +467,7 @@ class CommCare(SingleSupervisorService):
             puts(color_notice("To restart formplayer, which always causes downtime, use the 'formplayer' service command."))
             puts(color_error("Refusing to run commcare service command with action 'restart'."))
             return 1
-        super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
+        return super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
 
 
 class Webworker(SingleSupervisorService):
@@ -488,7 +488,7 @@ class Webworker(SingleSupervisorService):
             puts(color_code(f"  cchq {self.environment.name} fab restart_webworkers"))
             puts(color_error("Refusing to run webworkers service command with action 'restart'."))
             return 1
-        super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
+        return super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
 
 
 class Formplayer(SingleSupervisorService):
--- a/src/commcare_cloud/commands/ansible/service.py
+++ b/src/commcare_cloud/commands/ansible/service.py
@@ -480,6 +480,16 @@ class Webworker(SingleSupervisorService):
     def supervisor_process_name(self):
         return get_django_webworker_name(self.environment)
 
+    def run(self, action, host_pattern=None, process_pattern=None):
+        if action == 'restart':
+            puts(color_warning("The 'webworker' service command rejects the 'restart' action"))
+            puts(color_warning("in order to protect against accidental downtime."))
+            puts(color_notice("For a no-downtime restart of webworkers, please use"))
+            puts(color_code(f"  cchq {self.environment.name} fab restart_webworkers"))
+            puts(color_error("Refusing to run webworkers service command with action 'restart'."))
+            return 1
+        super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
+
 
 class Formplayer(SingleSupervisorService):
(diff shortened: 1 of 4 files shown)


calellowitz (commented):
Instead, does it make sense to just have `service commcare restart` call `fab restart_services` under the hood, like `deploy` does? Then we can keep consistent commands, but have them do the right thing. I don't think anyone expects this to restart formplayer so that should be fine.

dannyroberts:
I think for that to be consistent, we'd have to also remove formplayer from the list (otherwise `start`/`stop` would start/stop formplayer too, but `restart` wouldn't). The fab commands also don't accept a `--limit` or `--only` option, which we'd then also need to check for. This was the simplest thing I was sure would work. I think you could get what you're suggesting to work too, but it's a bigger change with more potential for an error in implementation that makes it do something surprising and I wanted to avoid that

calellowitz (approved):
Makes sense, thanks

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -457,6 +457,18 @@ class CommCare(SingleSupervisorService):
...
         return 'all'
 
+    def run(self, action, host_pattern=None, process_pattern=None):
+        if action == 'restart':
+            puts(color_warning("The 'commcare' service command rejects the 'restart' action"))
+            puts(color_warning("in order to protect against accidental downtime."))
+            puts(color_notice("For a no-downtime restart of commcare-hq services, please use one of the following"))
+            puts(color_code(f"  cchq {self.environment.name} fab restart_services  # for all processes that run commcare-hq code."))
+            puts(color_code(f"  cchq {self.environment.name} fab restart_webworkers  # for webworkers only"))
+            puts(color_notice("To restart formplayer, which always causes downtime, use the 'formplayer' service command."))
```
millerdev, line 467: It seems better to be explicit about the command rather than leaving room to guess what _'formplayer' service command_ means.
```suggestion
            puts(color_code(f"  cchq {self.environment.name} service formplayer restart  # for formplayer, causes downtime"))
```

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -457,6 +457,18 @@ class CommCare(SingleSupervisorService):
...
+            puts(color_error("Refusing to run commcare service command with action 'restart'."))
+            return 1
+        super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
```
millerdev, line 470:
```suggestion
        return super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
```

src/commcare_cloud/commands/ansible/service.py
```diff
@@ -468,6 +480,16 @@ class Webworker(SingleSupervisorService):
...
 
+    def run(self, action, host_pattern=None, process_pattern=None):
+        if action == 'restart':
+            puts(color_warning("The 'webworker' service command rejects the 'restart' action"))
+            puts(color_warning("in order to protect against accidental downtime."))
+            puts(color_notice("For a no-downtime restart of webworkers, please use"))
+            puts(color_code(f"  cchq {self.environment.name} fab restart_webworkers"))
+            puts(color_error("Refusing to run webworkers service command with action 'restart'."))
+            return 1
+        super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
```
millerdev, line 491:
```suggestion
        return super().run(action, host_pattern=host_pattern, process_pattern=process_pattern)
```

src/commcare_cloud/commands/ansible/service.py
dannyroberts: What I wanted to avoid is to make it too easy to copy-paste a command that one should avoid running because it causes downtime. So I intentionally inlined the ones above and then referenced this one verbally. It's a hedge against people not reading carefully.

dannyroberts:
After forgetting letting this sit for far too long, I've made the small updates and responded. Could I get another pass? Apologies for the wait!

millerdev (approved):

PR #5045, opened 2021-12-18
Temporarily remove malt_generation_queue altogether

dannyroberts:
I'm applying this to capture more persistently the current state that `malt_generation_queue`. @skodde @gherceg please review this retroactively, and give it approval but then we can just close it.

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -91,11 +91,6 @@ celery_processes:
       pooling: gevent
       concurrency: 40
       num_workers: 2
-    malt_generation_queue:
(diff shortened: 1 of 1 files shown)


calellowitz (approved):

dannyroberts:
Actually I'll just merge this now, and then @gherceg you can revert it when you want to bring it back up. Make sure to purge the queue or run the "revoke" command to revoke tasks as they come up or else it'll process the whole backlog as soon as it comes up and cause trouble we don't want

PR #5079, opened 2022-01-06
Ce/auditcare pool size

calellowitz:
https://dimagi-dev.atlassian.net/browse/SAAS-12885

This PR adds the ability to set pgbouncer pool size at the db level, and reduces the auditcare pool size. This will result in a total pool size of 150 (3 machinces * 2 processes * 25 connections), a sharp reduction from the 1000 we currently allow, but still well above our usage except during the outages.

##### ENVIRONMENTS AFFECTED

Production


dannyroberts (approved):

dannyroberts:
test updates look good ✅

PR #5080, opened 2022-01-06
Revert "increase shared buffers to 40%"

calellowitz:
Reverts dimagi/commcare-cloud#5074

This didn't work and may have negative effects, so reverting for now.


dannyroberts (approved):

dannyroberts:
@calellowitz the previous default (that this will take it back to) was 25% of memory, right?

calellowitz:
> @calellowitz the previous default (that this will take it back to) was 25% of memory, right?

Yep!

PR #5082, opened 2022-01-07
Add reCAPTCHA secret

minhaminha:
[Jira Ticket](https://dimagi-dev.atlassian.net/browse/SAAS-12670)

This is part 1 of the ticket (adding CAPTCHA). 

##### ENVIRONMENTS AFFECTED

Only staging will be affected in this PR.


environments/staging/public.yml
```diff
@@ -152,6 +152,7 @@ localsettings:
   RATE_LIMIT_SUBMISSIONS: yes
   SES_CONFIGURATION_SET: staging-email-events
   SNS_EMAIL_EVENT_SECRET: "{{ SNS_EMAIL_EVENT_SECRET }}"
+  RECAPTCHA_PUBLIC_KEY: '6LeaE-sdAAAAAJm75fPLW9j0TBuX5dMUEJ14rQzw'
```
dannyroberts, line 155: Even though it's not strictly secret (and is visible in browsers), we have a convention of storing both the public and private key of a key pair (or analogously both the username and the password for a set of credentials) as "secrets". Part of it is that they logically go together and they change together, so it's more convenient if when they change you can change them the same way. Part of it is that keys like this _look_ like secrets, which can get people used to putting things that _look_ like secrets in `public.yml`, which could down the road easily lead to a mistake where we put an actual secret there.

So my in short my suggestion is that you move this into `secrets.yml` like the other one.

src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -711,6 +711,14 @@ ENABLE_DRACONIAN_SECURITY_FEATURES = True
 ADD_CAPTCHA_FIELD_TO_FORMS = True
 {% endif %}
 
+{% if localsettings.RECAPTCHA_PUBLIC_KEY is defined %}
+RECAPTCHA_PUBLIC_KEY = {{ localsettings.RECAPTCHA_PUBLIC_KEY }}
```
dannyroberts, line 715: I think these two lines need to reference `RECAPTCHA_PUBLIC_KEY` now instead of `localsettings. RECAPTCHA_PUBLIC_KEY`

dannyroberts (approved):

PR #5091, opened 2022-01-11
use correct units

calellowitz:
Fixed the units to be correct

##### ENVIRONMENTS AFFECTED

Production


dannyroberts:
I think this isn't necessary. Because of our internal unit conversion, our code interprets `32MB` and `4096` as the same thing here, and 32MB is easier to read.

PR #5099, opened 2022-01-17
Lower connection limits for auditcare

calellowitz:
We have 3 pgbouncer machines, so this limits connections from 150 (3 * 50) to 48 (3 * 16), in an effort to make sure that at the limit, the machine is still functional.

##### ENVIRONMENTS AFFECTED

Production

### DIFF
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -78,7 +78,7 @@ dbs:
   auditcare:
     host: rds_pgauditcare0
     pgbouncer_endpoint: pgmain_nlb
-    pgbouncer_pool_size: 8
+    pgbouncer_pool_size: 16
     pgbouncer_hosts:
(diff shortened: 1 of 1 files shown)


millerdev (approved):

millerdev:
Not sure I'm following that math. 6*50 is 300, not 150.

calellowitz:
> Not sure I'm following that math. 6*50 is 300, not 150.

:man_facepalming: fixed

millerdev (approved):

calellowitz:
Updated to take into account the 30 connections from the reserve pool. This changes the math slightly, and now the pool size is capped at 24 (3 * 8), but 30 additional connections become available after 5 seconds of inactivity making the true cap 54.

millerdev (approved):

PR #5100, opened 2022-01-18
Single install script to install CommCareHQ on a monolith

sravfeyn:
The current monolith install [instructions](https://dimagi.github.io/commcare-cloud/setup/new_environment.html) are very long and involves a lot of manual steps. This script automates all the steps (WIP) in it so that all a user has to do is download the repo, fill a very small config file and run the script.

The script does below
1. installs system requirements
2. installs commcare-cloud
3. bootstraps an environments config directory using the installed ansible
4. runs cchq deploy-stack using ansible local connection
Screenshot 2022-01-18 at 4 51 05 PM

I am still deciding whether the script should also do the code deploy and SSL setup or ask the user to do it separately. Also it currently skips the port check as that step is not correct. See https://github.com/dimagi/commcare-cloud/pull/3973#discussion_r786473153

@proteusvacuum


oneclick_monolith_install/bootstrap-env-playbook.yml
```diff
@@ -0,0 +1,67 @@
...
+      when: config_is_not_valid
+    - name: Create environments directory
+      file: path={{env_path}} state=directory
+    - name: Create env's users directory
+      file: path={{envs_path}}/_users state=directory
+    - name: Create ssh-public-keys directory
+      file: path={{envs_path}}/_authorized_keys state=directory
+    - name: Copy templated env files
+      template:
+        src: "./sample-environment/{{item}}.j2"
```
sravfeyn, line 32: Note that this kind of env bootstrap exists via [commcare-cloud-bootstrap](https://github.com/dimagi/commcare-cloud/tree/master/commcare-cloud-bootstrap) but it uses very specific format and requires a lot more details from the user as it fits into a different workflow.

I have also observed that there is https://github.com/dimagi/sample-environment but that repo is not getting upto date when something is modified in the commcare-cloud repo so it has a tendency to get outdated.

oneclick_monolith_install/README.md
```diff
@@ -0,0 +1,5 @@
+## About One Click Monolith Install
```
proteusvacuum, line 1: What if we just call this the "Monolith Install", and the manual installation becomes the secondary way you would do something more custom.

oneclick_monolith_install/cchq-install.sh
```diff
@@ -0,0 +1,73 @@
...
+# install comcare-cloud
+DIR="${BASH_SOURCE%/*}"
+if [[ ! -d "$DIR" ]]; then DIR="$PWD"; fi
+source "$DIR/../control/init.sh"
+touch /var/log/ansible.log && chmod 666 /var/log/ansible.log
+
+
+printf "\nStep 3: Initializing environments directory for storing your CommCareHQ instance's configuration \n"
+ansible-playbook --connection=local --extra-vars "@$config_file_path" "$DIR/bootstrap-env-playbook.yml"
+printf "\n Encrypting your environment's passwords file using ansible-vault. Please store this password safely as it will be required latter \n"
```
proteusvacuum, line 49:
```suggestion
printf "\n Encrypting your environment's passwords file using ansible-vault. Please store this password safely as it will be required later \n"
```

oneclick_monolith_install/cchq-install.sh
```diff
@@ -0,0 +1,73 @@
...
+printf "\bStep 4: Setting up users and SSH auth \n"
+source ~/.commcare-cloud/load_config.sh
+commcare-cloud $env_name update-local-known-hosts
+commcare-cloud $env_name bootstrap-users -c local
+
+printf "\nEverything is setup to install CommCareHQ now!\n"
+printf "Please see below a summary of what this script has setup so far!\n"
+printf "1. Installed commcare-cloud, the tool to deploy and manage your CommCareHQ instance.\n"
+printf "2. Users ansible and $ssh_username are created with SSH and sudo access.\n"
+printf "3. A configuration directory created for your CommCareHQ environment at: $HOME/environments/env_name.\n"
```
proteusvacuum, line 61:
```suggestion
printf "3. A configuration directory created for your CommCareHQ environment at: $HOME/environments/$env_name.\n"
```

oneclick_monolith_install/install-config.yml
```diff
@@ -0,0 +1,18 @@
+# Name to uniquely identify your CommCareHQ instance/environment.
+env_name: ""
+
+# Web address of your HQ instance
+site_host: ""
+
+# Ansible inventory name for this server (e.g. hqserver1)
+server_inventory_name: ""
```
proteusvacuum, line 8: is this necessary? we could just default this to `commcare` or the `env_name` from above?

oneclick_monolith_install/install-config.yml
```diff
@@ -0,0 +1,18 @@
+# Name to uniquely identify your CommCareHQ instance/environment.
```
proteusvacuum, line 1: This file: ❤️

oneclick_monolith_install/install-config.yml
```diff
@@ -0,0 +1,18 @@
...
+env_name: ""
+
+# Web address of your HQ instance
```
proteusvacuum, line 4:
```suggestion
# Web address of your HQ instance. e.g. "example.com"
```

oneclick_monolith_install/sample-environment/vault.yml.j2
```diff
@@ -0,0 +1,48 @@
...
+  DROPBOX_SECRET: ''
+  EMAIL_LOGIN: ''
+  EMAIL_PASSWORD: ''
+  FORMPLAYER_INTERNAL_AUTH_KEY: '{{ lookup('password', '/dev/null length=32 chars=ascii_letters,digits') }}'
+  GMAPS_API_KEY: ''
+  GOOGLE_ANALYTICS_API_ID: ''
+  HUBSPOT_API_ID: ''
+  HUBSPOT_API_KEY: ''
+  JAR_KEY_PASS: ''
+  JAR_STORE_PASS: ''
```
proteusvacuum, line 45: Are these still used in the codebase?

oneclick_monolith_install/sample-environment/public.yml.j2
```diff
@@ -0,0 +1,116 @@
...
+    'region': "MA"
+    'postal_code': "02139"
+    'country': "US"
+    'phone_number': "(617) 649-2214"
+    'email': "accounts@dimagi.com"
+    'website': "http://www.dimagi.com"
+  INVOICE_PREFIX: INC-
+  INVOICE_STARTING_NUMBER: 5000
+  J2ME_ADDRESS: ''
+#  KAFKA_HOST:
```
proteusvacuum, line 104: Can we just remove these commented entries?

oneclick_monolith_install/sample-environment/public.yml.j2
```diff
@@ -0,0 +1,116 @@
...
+  #  PILLOWTOP_MACHINE_ID:
+  PILLOW_RETRY_QUEUE_ENABLED: True
+  REDIS_DB: '0'
+  REDIS_HOST: "localhost"
+  REDIS_PORT: '6379'
+  REMINDERS_QUEUE_ENABLED: False
+  SMS_GATEWAY_URL: 'http://gw1.promessaging.com/sms.php'
```
proteusvacuum, line 111: I don't think this is used any more

proteusvacuum (approved):
Wow @sravfeyn this is great!
A few comments:
- I would remove reference to "one click" - we don't click anything, and I think this should just be the default "monolith install"
- In the docs, we should add some links to "optional set up steps" that detail adding things like the BITLY API key 
- We should find a way of keeping the settings files up to date with our internal settings file, since I already see things that are out of sync (out of scope for now, and you might have already thought of this)

oneclick_monolith_install/README.md
sravfeyn: Yeah, re. the name, we should definitely use another name. It's also not a 'wizard'. I was originally [using](https://commcare-cloud.readthedocs.io/en/latest/installation/index.html) `Quick Install`. I will update this.

Currently I have positioned this as below (per https://commcare-cloud.readthedocs.io/en/latest/installation/index.html)
> Most of the install is done automatically with configuration set to sensible defaults. This is useful to quickly get started, test or preview what’s involved in a working CommCareHQ environment.

The main reason why I am not comfortable positioning this as a production install is because it's not tested (other than on my local setup). The other less important reason is that the developer using this quick install would have no understanding of what happened in the install (which seems a bit necessary for production). 

(will be good to talk about this involving Mary too)

oneclick_monolith_install/cchq-install.sh
sravfeyn: Good catch!

oneclick_monolith_install/install-config.yml
sravfeyn: :) I had exactly the same! I removed it by mistake.

oneclick_monolith_install/install-config.yml
sravfeyn: My reasoning to offer this in the config was this.

This is a name that will be used/visible frequently when referring the host via ansible or in the sentry/datadog logs, and in the email alerts etc. Also, updating this in the future takes more effort too (needs ansible redeploy).  So, I thought better to have the user pick a name.

oneclick_monolith_install/sample-environment/public.yml.j2
sravfeyn: Yes!

oneclick_monolith_install/sample-environment/public.yml.j2
sravfeyn: True, will remove it

oneclick_monolith_install/sample-environment/vault.yml.j2
sravfeyn: It doesn't look like so, will remove it

oneclick_monolith_install/install-config.yml
proteusvacuum: That's fair, although I think if this is the only machine the name here is somewhat irrelevant. If they end up moving to a cluster type deployment, they will need to redeploy ansible anyway.

quick_monolith_install/cchq-install.sh
```diff
@@ -0,0 +1,105 @@
+# # Run as bash cchq-install.sh install-config.yml
+
+# #!/bin/sh
+set -e
+
+printf "Please note that you may be prompted for sudo permission for some of the scommands.\n"
```
kaapstorm, line 6:
```suggestion
printf "Please note that you may be prompted for sudo permission for some of the commands.\n"
```

quick_monolith_install/cchq-install.sh
```diff
@@ -0,0 +1,105 @@
...
+sudo touch /var/log/ansible.log && sudo chmod 666 /var/log/ansible.log
+
+printf "\n"
+printf "#################################################"
+printf "\nStep 3: Initializing environments directory for "
+printf "storing your CommCareHQ instance's configuration \n"
+printf "#################################################"
+printf "\n"
+# VENV should have been set by init.sh
+echo "asdasdasdasd$VENV"
```
kaapstorm, line 63: Is this for debugging?

kaapstorm (approved):
This is very cool!

quick_monolith_install/bootstrap-env-playbook.yml
```diff
@@ -12,15 +12,46 @@
...
+          {% for interface in ansible_interfaces %}
+            IP Address: {{hostvars[inventory_hostname]['ansible_%s' | format(interface)]['ipv4']['address'] }}, Interface Name: {{ interface}}
+          {% endfor%}
+
+          Please enter the private interface name to be used from above three.
+          This will be used to set firewall rules
+      register: result_internal_interface
+      until: "result_internal_interface.user_input in ansible_interfaces"
+      retries: 3
+      delay: 0
```
proteusvacuum, line 50: This is great!

quick_monolith_install/cchq-install.sh
```diff
@@ -29,47 +31,67 @@ config_file_path=$(get_abs_filename $1)
...
+printf "\nStep 4: Setting up users and SSH auth \n"
+printf "#################################################"
+printf "\n"
 source ~/.commcare-cloud/load_config.sh
 commcare-cloud $env_name update-local-known-hosts
-commcare-cloud $env_name bootstrap-users -c local
+commcare-cloud $env_name bootstrap-users -c local --quiet
 
-printf "\nEverything is setup to install CommCareHQ now!\n"
+printf "\nEverything is setup to install CommCareHQ now! Would you like to instal CommCareHQ now?\n"
```
proteusvacuum, line 78: instal -> install

quick_monolith_install/cchq-install.sh
sravfeyn: Oh, yes! Thanks, removing it.

PR #5104, opened 2022-01-21
Adjust pgformplayer0-production pgbouncer_pool_size and max_wal_size

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-12952

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -50,6 +50,7 @@ dbs:
   formplayer:
     host: rds_pgformplayer0
     pgbouncer_endpoint: pgformplayer_nlb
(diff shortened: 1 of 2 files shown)


skodde (approved):
You may want to account for the reserve pool as per https://github.com/dimagi/commcare-cloud/pull/5099#issuecomment-1014726746

dannyroberts:
Yeah I think it's okay. I'm not trying to target a very precise upper/lower bound, I just want to roughly cut the max in half, so that it's significantly lower but still ~2x as high as it goes during normal operation

PR #5112, opened 2022-01-26
Revert "Lower the pgbouncer reserve pool size for auditcare db"

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-12885

This reverts commit a765a68e24fee1fda7d02b753786ac4a0e2ab77b.

This commit from https://github.com/dimagi/commcare-cloud/pull/5111 appears to have caused serious problems for reasons I don't fully understand. Reverting it should make the rest of that PR safe to apply.

##### ENVIRONMENTS AFFECTED
Production

### DIFF
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -80,7 +80,6 @@ dbs:
     host: rds_pgauditcare0
     pgbouncer_endpoint: pgmain_nlb
     pgbouncer_pool_size: 8
-    pgbouncer_reserve_pool_size: 2
     pgbouncer_hosts:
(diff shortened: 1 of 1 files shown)


dannyroberts:
@prem-fissionhq FYI

millerdev (approved):
I gather that postgres would not start after this change was applied in production. Sounds like possibly a syntax error in the config file. Do we have access to the logs when postgres would not start?

dannyroberts:
My understanding is that it could still start, but wasn't able to connect to the auditcare db anymore. I'll drop more notes in the ticket

PR #5136, opened 2022-02-03
Reduce total connections to pgucr0

skodde:
https://dimagi-dev.atlassian.net/browse/SAAS-13135

##### ENVIRONMENTS AFFECTED
Production

### DIFF
--- a/environments/production/postgresql.yml
+++ b/environments/production/postgresql.yml
@@ -62,8 +62,8 @@ dbs:
   ucr:
     host: rds_pgucr0
     pgbouncer_endpoint: pgucr_nlb
-    pgbouncer_pool_size: 9
-    pgbouncer_reserve_pool_size: 3
+    pgbouncer_pool_size: 10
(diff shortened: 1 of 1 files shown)


environments/production/postgresql.yml
```diff
@@ -62,6 +62,8 @@ dbs:
   ucr:
     host: rds_pgucr0
     pgbouncer_endpoint: pgucr_nlb
+    pgbouncer_pool_size: 9
+    pgbouncer_reserve_pool_size: 3
```
dannyroberts, line 66: Since each pgbouncer machine has two pgbouncer processes, 3 here actually means `3 // 2`, which is `1` (and the same as `2 // 2`). So I might suggest just putting 2 here rather than 3.

skodde: Good point. I'll do 10 and 2 then, to have the same total. Thanks!

dannyroberts (approved):

PR #5137, opened 2022-02-03
Add pillow instance

skodde:
https://dimagi-dev.atlassian.net/browse/SAAS-13135

##### ENVIRONMENTS AFFECTED
Production


environments/production/terraform.yml
```diff
@@ -242,6 +242,15 @@ servers:
     os: ubuntu_pro_bionic
     count: 2
 
+  - server_name: "pillow_b{i}-production"
+    server_instance_type: r5a.8xlarge
+    network_tier: "app-private"
+    az: "b"
```
dannyroberts, line 248: Is the idea that we'll have a machine in zone b, so that in case there's serious issue with zone a, we can at least manually move processors over to the machine in zone b quickly?

skodde: Right now we just need a new node to allocate additional resources for the issue described in the ticket, but it seemed like a good idea to create it in a new AZ and then possibly keep it by removing one of the old ones instead, given our multi-AZ effort to prepare for possible disruptions such as the one you described.

dannyroberts: Okay cool. That sounds good. Unlike some other types of workers, pillows don't allow for any kind of automatic failover (since processes are statically defined per machine), but it's still very nice to give us a chance to resolve quickly with manual intervention.

dannyroberts (approved):

calellowitz (approved):

calellowitz (approved):

PR #5162, opened 2022-02-17
Add new django_manage machine

skodde:
https://dimagi-dev.atlassian.net/browse/SAAS-13181

##### ENVIRONMENTS AFFECTED
Production


environments/production/terraform.yml
```diff
@@ -21,6 +21,14 @@ servers:
     group: "control"
     os: ubuntu_pro_bionic
 
+  - server_name: "djangomanage_a1-production"
+    server_instance_type: m6a.xlarge
```
dannyroberts, line 25: m6a seems good

environments/production/terraform.yml
```diff
@@ -21,6 +21,14 @@ servers:
     group: "control"
     os: ubuntu_pro_bionic
 
+  - server_name: "djangomanage_a1-production"
```
dannyroberts, line 24: Curious to hear your reasoning, but my first instinct is that we should stick with the numbering system we use for "singleton" machines and use `djangomanage3-production` here.

skodde: I'd like for us to get in the habit of uniforming a bit the names we use, and even for single machines to use one that corresponds to the AZ they're supposed to live in. This should also make it easier to add/delete/replace them in the future while still keeping consistent names, and the inventory in general more maintainable.

dannyroberts: I'm not against the naming system you chose in particular (and having the AZ in the name is nice), as long as there's a clear decision that we're moving to it. What I don't want is to just have everyone on the team name machines differently. I'll offline with you and the team

dannyroberts (approved):

dannyroberts (approved):

shyamkumarlchauhan (approved):

PR #5169, opened 2022-02-23
kafka verify download URL checksum

prem-fissionhq:
##### ENVIRONMENTS AFFECTED

Staging. 

@dannyroberts performing SHA512 verification for Kafka download URL. Please review and let me know your suggestions. 
cc: @shyamkumarlchauhan

### DIFF
--- a/src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
@@ -14,6 +14,8 @@
 
 - name: Download Kafka
   get_url: url="http://{{ apache_mirror }}/kafka/{{ kafka_version }}/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz" dest="/opt/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
+  checksum: "{{ kafka_download_sha512 }}"
+  when: kafka_version == "2.6.1"
 
 - name: Extract Kafka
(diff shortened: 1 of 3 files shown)


dannyroberts (approved):

src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
```diff
@@ -13,7 +13,11 @@
     state: "{{ 'present' if ansible_distribution_version == '14.04' else 'absent' }}"
 
 - name: Download Kafka
-  get_url: url="http://{{ apache_mirror }}/kafka/{{ kafka_version }}/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz" dest="/opt/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
+  get_url:
+    url: "http://{{ apache_mirror }}/kafka/{{ kafka_version }}/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
+    dest: "/opt/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
+    checksum: "{{ kafka_download_sha512 }}"
+  when: kafka_version == '2.6.1'
```
sravfeyn, line 20: @prem-fissionhq 
This is breaking installs on other envs, since kafka gets downloaded only when `kafka_version == '2.6.1'` but the task below `Extract Kafka` expects the file to be there regardless of the version so it fails for the default version which is `0.8.2.2` per above [file](https://github.com/dimagi/commcare-cloud/pull/5169/files#diff-a581db887657987789d2007a5254612e45b9f49f934102418ca8eb3426814e41R3).

PR #5171, opened 2022-02-24
Add workers to ush_background_tasks

minhaminha:
##### ENVIRONMENTS AFFECTED
Staging

### DIFF
--- a/environments/staging/app-processes.yml
+++ b/environments/staging/app-processes.yml
@@ -34,10 +34,8 @@ celery_processes:
       concurrency: 1
     ucr_queue:
       concurrency: 1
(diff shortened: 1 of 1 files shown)


environments/staging/app-processes.yml
```diff
@@ -34,10 +34,8 @@ celery_processes:
       concurrency: 1
     ucr_queue:
       concurrency: 1
-    user_import_queue:
+    user_import_queue, ush_background_tasks:
```
gherceg, line 37: could you keep these queues defined on separate lines? Minimizes diffs which is a good thing in this context

PR #5182, opened 2022-03-01
Cs/sc 1918 restore docs

Charl1996:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-1918)

Although the changes to the docs are minimal (fixing spelling mistakes and expounding a bit on BlobDB restore), the process to get to that point has taken me a significant amount of time. The main reason being that I, in order to test the restore process, had to 
- spin up a monolith instance
- import some data to it; used my production (dummy) data
- ensure backups are made
- finally restore PostgreSQL, CouchDB and BlobDB (didn't go through Elasticsearch as it can easily be rebuilt)

By going through this more elaborate process above, I was able to uncover a few other issues along the way.

##### ENVIRONMENTS AFFECTED
This only affects the docs.

### DIFF
--- a/docs/commcare-cloud/backup.md
+++ b/docs/commcare-cloud/backup.md
@@ -229,8 +229,12 @@ The `blobdb` is our binary data store.
 BlobDB backups create a compressed version of the blobdb data directory.
 
 ### Restoring BlobDB Backups
-The BlobDB restore process is the same as the couchdb restore process in that it involves extracting the backed up data
-to the data directory.
+The BlobDB restore process depends on what BlobDB system you're using.
+- If you're using the default file system BlobDB, the restore process is the same as the couchdb restore process in that it involves extracting the backed up data 
+to the data directory. 
+- If you're using some other (distributed) system you should follow that service's provided instructions on restoration.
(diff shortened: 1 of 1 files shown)


docs/commcare-cloud/backup.md
```diff
@@ -229,10 +229,25 @@ The `blobdb` is our binary data store.
 BlobDB backups create a compressed version of the blobdb data directory.
 
 ### Restoring BlobDB Backups
+The BlobDB restore process is the same as the couchdb restore process in that it involves extracting the backed up data
```
proteusvacuum, line 232: Maybe add a note that this is true for the file system blob db (is that the default?), but that if you are using some other distributed S3 compatible blob system (e.g. Minio), that you should follow the instructions provided by that system.

proteusvacuum (approved):

docs/commcare-cloud/backup.md
Charl1996: @proteusvacuum Ah, good point.

proteusvacuum (approved):

PR #5188, opened 2022-03-02
Fix format statement for COMMCARE_CLOUD_DEFAULT_USERNAME_ENV_VAR_MESSAGE

minhaminha:
While trying run `cchq <env> terraform` commands I noticed that if your username on the AWS machines is different than your dev workstation username, the message that prompts you to run `export COMMCARE_CLOUD_DEFAULT_USERNAME={AWS machine username}` would give an error and exit the process. 

##### ENVIRONMENTS AFFECTED

None


millerdev (approved):
Thanks for fixing :hammer_and_wrench:

PR #5062, opened 2021-12-27
Saas 12875/upgrade postgres plproxy

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-12875
##### ENVIRONMENTS AFFECTED

Staging
--
Ansible changes will affect all environments.

--
Upgrading Postgres version on pl/proxy server to "10" (specific: 10.19).

PR #5199, opened 2022-03-07
changing postgres_version default value to 9.6 from 10 in ansible Roles

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-12875
##### ENVIRONMENTS AFFECTED

All.

Since my previous modification to default values will impact third-party users who have not yet updated the PostgreSQL version to 10. Therefore reverting back my changes to default values and as we can also control this variable from public.yml `postgresql_version: '9.6'` updated India files and production will do when performing it. 

Old PR: https://github.com/dimagi/commcare-cloud/pull/5062

### DIFF
--- a/environments/india/postgresql.yml
+++ b/environments/india/postgresql.yml
@@ -11,6 +11,7 @@ pgbouncer_override:
   pgbouncer_reserve_pool: 5
 
 postgres_override:
+  postgresql_version: '10'
   postgresql_max_connections: 300
   postgresql_hba_entries:
     # required for backup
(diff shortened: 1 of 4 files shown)


joxl (commented):
Would it be good to add a changelog item with this revert that outlines the issue and explains how to adjust your env files (explicitly override `postgresql_version`) in case there are other parties who were negatively affected by this?  Since this was merged for several weeks, it is in theory _possible_ that this PR could once again break something for a third-party hoster.

dannyroberts (approved):

PR #5203, opened 2022-03-09
net-tools not in Ubuntu 18.04.6 LTS by default

kaapstorm:
From Ubuntu 18.04.6 LTS, users must install the net-tools package to use `ifconfig`. It is deprecated in favor of `ip`.

This change uses `ip` instead of `ifconfig`.

Stupidly, I tested the happy path first, and now I can't go back to test the unhappy path (an incorrectly configured interface) on that particular VM. I have not yet tested the unhappy path on a new VM. But I figured I'd open the PR in the meantime. (I do expect it to work though. `ip addr show dev invaliddevice` returns nothing to STDOUT, and an error message to STDERR.)

##### ENVIRONMENTS AFFECTED

Affects new environments, and provisioning new VMs in existing environments, when VMs have a fresh Ubuntu 18.04 image.

### DIFF
--- a/src/commcare_cloud/ansible/roles/ufw/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/ufw/tasks/main.yml
@@ -1,6 +1,6 @@
 - name: "make sure the specified private interface actually exists"
   # a bit of a hack, but just hopefully prevents turning off access on all interfaces!
-  shell: "ifconfig | grep {{ ufw_private_interface }}"
+  shell: "ip addr show dev {{ ufw_private_interface }}"
   changed_when: no
 
 - name: allow incoming traffic on private interface
--- a/src/commcare_cloud/ansible/roles/ufw/tasks/proxy_ufw.yml
+++ b/src/commcare_cloud/ansible/roles/ufw/tasks/proxy_ufw.yml
@@ -1,6 +1,6 @@
 - name: "make sure the specified private interface actually exists"
   # a bit of a hack, but just hopefully prevents turning off access on all interfaces!
-  shell: "ifconfig | grep {{ ufw_private_interface }}"
+  shell: "ip addr show dev {{ ufw_private_interface }}"
   changed_when: no
 
 # Rules for proxies


proteusvacuum:
How has this worked up until now? 
Also, if this change is merged, do we also need to remove `net-tools` from here: https://github.com/dimagi/commcare-cloud/blob/655c27b610eead6d2be9e9de531a40e2e51fc2ed/docs/setup/new_environment.md#L126 ?

kaapstorm:
> I see our control machine is running a different version of ubuntu

Yeah, I think so. I tried to find out when net-tools stopped being installed by default, but I couldn't.

> Also, if this change is merged, do we also need to remove net-tools from here: 

https://github.com/dimagi/commcare-cloud/blob/655c27b610eead6d2be9e9de531a40e2e51fc2ed/docs/setup/new_environment.md#L126

That's a good question. Why is it there in the first place, if it used to be installed by default?

And if it wasn't installed by default, how did `deploy-stack` work in my own test cluster a few months ago?

I don't know. @sravfeyn do you have more clue than I do? Should we pull in some reviewers who have a deeper understanding of commcare-cloud than me?

kaapstorm:
Confirming that the unhappy path behaves as intended:

    TASK [ufw : make sure the specified private interface actually exists] *************************************************************
    fatal: [192.168.122.7]: FAILED! => {"changed": false, "cmd": "ip addr show dev enp1XXXXs0", "delta": "0:00:00.003967", "end": "2022-03-11 13:41:55.950882", "msg": "non-zero return code", "rc": 1, "start": "2022-03-11 13:41:55.946915", "stderr": "Device \"enp1XXXXs0\" does not exist.", "stderr_lines": ["Device \"enp1XXXXs0\" does not exist."], "stdout": "", "stdout_lines": []}

kaapstorm:
@snopoke or @shyamkumarlchauhan do you maybe know why we install **net-tools** on the Control VM? Or why `ifconfig` has been working on other machines, where **net-tools** is not explicitly installed?

snopoke:
> @snopoke or @shyamkumarlchauhan do you maybe know why we install **net-tools** on the Control VM? Or why `ifconfig` has been working on other machines, where **net-tools** is not explicitly installed?

Unfortunately I don't have any insight.

skodde:
To recap our offline discussion:

- The difference seems to arise from set of base packages in the AWS AMI we use vs. the one generated by the installer with your specified target.
- `ifconfig` is indeed deprecated and we should replace it with `ip` (net-tools vs. iproute2).
- We should note in the documentation what we test against and make others aware of possible divergences, but we should fix similar issues and make our code more generic/robust, like in this case.

skodde (approved):

PR #5211, opened 2022-03-15
executable: pip3

kaapstorm:
This change allows me to install Supervisor 4.0.4 on a VM where the only installed service is Formplayer.

Without this change I get the following error:

    TASK [install supervisor 4.0.4] ****************************************************************************************************
    fatal: [192.168.122.7]: FAILED! => {"changed": false, "msg": "Unable to find any of pip2, pip to use.  pip needs to be installed."}

I just copied the `pip` parameters used in **common_installs/tasks/main.yml**.

##### ENVIRONMENTS AFFECTED

Only affects new environments. Existing environments will already have Supervisor 4.0.4 present.


dannyroberts:
Perhaps you could check for the existence of pip, and if it doesn't exist, alias it to pip3? Not sure what causes the issue in the environment you're working in; if it's using ubuntu 18.04 it must just have a different set of things installed by default or something.

sravfeyn (approved):

kaapstorm:
> FAILED! => {"changed": false, "msg": "Unable to find any of pip3 to use.  pip needs to be installed."}

That makes me think that 
* Supervisor is getting installed _before_ [common_installs](https://github.com/dimagi/commcare-cloud/blob/8b8919aad41c882d7f420796f87d7ebc4caa490f/src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml#L121) is run.
* Or common_installs should be installing `python{{ python_version }}-pip` and it is not.

I'll do some more investigating.

src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
```diff
@@ -4,8 +4,19 @@
...
+    state: present
+  become: yes
+
 - name: install supervisor 4.0.4
-  pip: name=supervisor state=present version=4.0.4
+  pip:
+    name: supervisor
+    state: present
+    version: 4.0.4
+    executable: python3 -m pip
```
skodde, line 18:
```suggestion
    executable: pip3
```
`/usr/bin/pip3` is what `python3-pip` provides, and what we use in similar instances. Is there a specific reason to change it in this case?

src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
```diff
@@ -4,8 +4,19 @@
   apt: name=supervisor state=absent
   become: yes
 
+- name: Ensure pip is available
+  apt:
+    name: python3-pip
+    state: present
+  become: yes
+
```
skodde, line 12:
```suggestion
```
I think we should put `python3-pip` in the task `Install python packages` or `Install common packages` in `src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml` instead.

src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
```diff
@@ -4,8 +4,19 @@
...
 - name: install supervisor 4.0.4
-  pip: name=supervisor state=present version=4.0.4
+  pip:
+    name: supervisor
+    state: present
```
skodde, line 16:
```suggestion
```
This is the default and we don't explicitly state it in similar instances.

src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
kaapstorm: No good reason. I wanted to make sure it was using what I had just installed. But I should have checked `dpkg -L`: Seems `pip` and `pip3` would both work:

    $ dpkg -L python3-pip
    ...
    /usr/bin/pip
    /usr/bin/pip3
    ...

I'll go with your suggestion if that's what we do in other places.

src/commcare_cloud/ansible/roles/supervisor/tasks/main.yml
kaapstorm: Thanks. https://github.com/dimagi/commcare-cloud/pull/5211/commits/f47650e40a67517def5b6fe52cfd661f4161d7bd

I still find it difficult to track imports vs dependencies, but your comments have been useful pointers on how to do that.

kaapstorm:
Hmmm. [Travis is not happy](https://app.travis-ci.com/github/dimagi/commcare-cloud/builds/248115902).

> FAILED! => {"changed": false, "msg": "Unable to find any of pip3 to use.  pip needs to be installed."}

(Not very tactful about it either:

> :x: The build **failed**, just like the previous build.

Gee, Travis. Kick a guy while he's down, why don't you?)

So the build on my previous, incorrect, commit also failed -- "Unable to find any of python3 -m pip to use." That seems a little nuts. 

I'll keep messing around, but if you've got any more thoughts @skodde you could probably save me hours.

kaapstorm:
So `python{{ python_version }}-pip` does not exist for deadsnakes Python 3.9, but it does exist for stock universe python3.

Travis is running Ansible in virtualenv /home/travis/virtualenv/python3.6.10/ but a real monolith or cluster is using stock python3. [Our docs](https://commcare-cloud.readthedocs.io/en/latest/installation/2-manual-install.html#prepare-all-machines-for-automated-deploy) instruct admins to install `python3-pip` on monoliths / their control machine.

:thinking: I think I'll change common_installs to use `python3-pip`.

skodde:
I'll make some tests, we're probably missing something simple.

kaapstorm:
Closing this. I'll open a follow-up PR that resolves this and a related problem.

PR #5232, opened 2022-04-04
Support two terraform versions (0.12, 0.13) concurrently

dannyroberts:
and allow each environment to specify a terraform_version

https://dimagi-dev.atlassian.net/browse/SAAS-13313

##### ENVIRONMENTS AFFECTED
SaaS aws only


skodde (approved):

src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
```diff
@@ -5,7 +5,8 @@ terraform {
     region  = {{ state_bucket_region|tojson }}
     encrypt = true
   }
-  required_version = ">= 0.12.0, <= 0.13.7"
+  {#- guaranteed to be a valid string; see comment in terraform_version_range_string definition #}
+  required_version = "{{ terraform_version_range_string }}"
 }
 
 provider "aws" {
```
prem-fissionhq, line 12: @dannyroberts aws provider 4.8.0 is particular for terraform13 version it will not support for terraform12. T12 supports >3.75.1. 
So we need to make a small change here, Instead of using ~> 4.8.0 we have to use ~> 3.75.1

prem-fissionhq: image

dannyroberts: Does `~> 3.75.1` work for both terraform 0.12 and 0.13?

dannyroberts: https://github.com/dimagi/commcare-cloud/pull/5241

PR #5233, opened 2022-04-04
Attempt to fix build

kaapstorm:
Prompted by the fact that the last successful build was 6 days ago.

##### ENVIRONMENTS AFFECTED

should only affect Travis


skodde:
Proper fix here: https://github.com/dimagi/commcare-cloud/pull/5237
See also: https://github.com/cloudalchemy/ansible-prometheus/issues/387

PR #5240, opened 2022-04-05
PgBouncer various fixes and cleanups

skodde:
Changes prompted by https://dimagi-dev.atlassian.net/browse/SAAS-12182

 - Bump the minimum required version to 1.13 so that we can use `Type=notify` with systemd
 - Add a `pgbouncer_version` variable that can be set per environment if necessary, and set the default to the currently latest available version (1.17)
 - Remove unnecessary `become` keywords and "pgbouncer" tags, which are already set in the playbook
 - Reorder the handlers to avoid unnecessary multiple restarts
 - Remove unnecessary hardcoded pgbouncer "classic" restart
 - Remove outdated after-reboot actions (open fd limit is set in the systemd unit)
 - Uniform configurations for "classic" and "multiprocess"
 - Correct the start/stop commands for monit, which were inverted
 - Adjust `with_sequence` to avoid "to count backwards make stride negative" errors (`with_sequence` is always parsed regardless of the `when` condition, even if applied only when that is true)
 - General format and syntax cleanups

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
@@ -21,4 +21,5 @@ postgresql_port: 5432
 
 pgbouncer_processes: 1
 pgbouncer_current_processes: [0]
-pgbouncer_min_version: "1.12"
+pgbouncer_version: "1.17.0-1.pgdg18.04+1"
+pgbouncer_min_version: "1.13"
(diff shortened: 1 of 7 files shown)


dannyroberts:
@skodde can you share more context (description, link to ticket that motivated it if applicable, etc)?

dannyroberts (approved):

PR #5241, opened 2022-04-05
Change aws provider version down to 3.75.1

dannyroberts:
to support terraform 0.12 and 0.13

https://dimagi-dev.atlassian.net/browse/SAAS-13313

##### ENVIRONMENTS AFFECTED
AWS envs

### DIFF
--- a/src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
+++ b/src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
@@ -10,7 +10,7 @@ terraform {
 }
 
 provider "aws" {
-  version = "~> 4.80.0"
+  version = "~> 3.75.1"
   region  = {{ region|tojson }}
 }
 


dannyroberts:
Tested and this appears to work with both terraform versions

dannyroberts:
@prem-fissionhq

skodde (commented):
`4.80.0` appears to have been a typo for `4.8.0`, which still supports terraform `0.12`.
Any reason to use `3.x` instead?

dannyroberts:
Thanks for pointing that out. No, I think was just a misunderstanding of what the error was on my part. I'll update to use `4.8.0`. For context, before yesterday this was set to `~> 3.22`, so even 3.75 is an upgrade. If we have any issues with the newer version for unrelated reasons, we can always fall back to 3.75 while we try to remediate the (hypothetical) issue.

skodde (approved):

dannyroberts:
I removed https://github.com/dimagi/commcare-cloud/commit/7d141533b0d4f5163c8d4b450dc55a22996fce30 from this PR because @prem-fissionhq pointed out to me that there are some minor incompatibilities between our code and aws provider 4.8.0. I'll let him open up a PR when that's ready

skodde (approved):

PR #5249, opened 2022-04-07
Automate swiss vault rotation

mjriley:
Associated ticket: https://dimagi-dev.atlassian.net/browse/SAAS-13405.

Our current process for vault rotation is manual and error prone. This script aims to automate the process -- the intention is that the genie can run this one script in one tab, use another tab to create a PR with the generated vault changes, and then come back to the initial tab to commit those changes to the Keepass. Potential improvements could be automating the PR process or putting in password requirements.


scripts/echo_stdin-client.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# This script just aims to provide a way for ansible-vault to read standard input.
+# ansible-vault doesn't let you specify passwords on the command line -- you either
+# need to provide a password file or a password script. This functions as the latter,
+# reading passwords piped in
+read -r line
+echo $line
```
joxl, line 8: If unquoted, `$line` will get "unpacked" into an argument vector (munging any whitespace, and possibly other special characters).
```suggestion
echo "$line"
```

joxl (changes requested):

millerdev (commented):
We are in the process of switching from KeePass to 1Password. Are you planning to port this script when that change is made?

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+    fi
+}
+
+prompt_for_password () {
+    echo "Enter password to unlock $KEE_PASS_DB:"
+    read -s KEE_PASS_DB_PASS
+}
+
+store_password () {
+    local password=$1
```
joxl, line 23:
```suggestion
    local password="$1"
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+    echo "  <keepass file> should be the full path to the keepass database containing the swiss vault password"
+    echo "  --resume should be specified after the rotation PR has been merged"
+}
+
+if [ $# -lt 1 ] || [ $# -gt 2 ]; then
+    display_help
+    exit 1
+fi
+
+KEE_PASS_DB=$1
```
joxl, line 133:
```suggestion
KEE_PASS_DB="$1"
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+
+init_keepass () {
+    # Locate keepassxc-cli or fail with error
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        KEE_PASS=/Applications/KeePassXC.app/Contents/MacOS/keepassxc-cli
+    elif [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        KEE_PASS=`which keepassxc-cli`
+    fi
+
+    if [ ! -x $KEE_PASS ]; then
```
joxl, line 11:
```suggestion
    if [ ! -x "$KEE_PASS" ]; then
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+prompt_for_password () {
+    echo "Enter password to unlock $KEE_PASS_DB:"
+    read -s KEE_PASS_DB_PASS
+}
+
+store_password () {
+    local password=$1
+
+    printf "$KEE_PASS_DB_PASS\n$KEE_PASS_DB_PASS" | $KEE_PASS db-create "$TEMP_DATABASE" -p > /dev/null 2>&1
+    printf "$KEE_PASS_DB_PASS\n$password" | $KEE_PASS add "$TEMP_DATABASE" $TEMP_KEY -p -q
```
joxl, line 26:
```suggestion
    printf "$KEE_PASS_DB_PASS\n$KEE_PASS_DB_PASS" | "$KEE_PASS" db-create "$TEMP_DATABASE" -p > /dev/null 2>&1
    printf "$KEE_PASS_DB_PASS\n$password" | "$KEE_PASS" add "$TEMP_DATABASE" "$TEMP_KEY" -p -q 
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+
+store_password () {
+    local password=$1
+
+    printf "$KEE_PASS_DB_PASS\n$KEE_PASS_DB_PASS" | $KEE_PASS db-create "$TEMP_DATABASE" -p > /dev/null 2>&1
+    printf "$KEE_PASS_DB_PASS\n$password" | $KEE_PASS add "$TEMP_DATABASE" $TEMP_KEY -p -q 
+}
+
+retrieve_password () {
+    echo "$KEE_PASS_DB_PASS"|$KEE_PASS show "$TEMP_DATABASE" "$TEMP_KEY" -a password -q
```
joxl, line 30:
```suggestion
    echo "$KEE_PASS_DB_PASS"| "$KEE_PASS" show "$TEMP_DATABASE" "$TEMP_KEY" -a password -q
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+}
+
+update_keepass () {
+    if [ -z "$GENERATED_PASS" ]; then
+        GENERATED_PASS=`retrieve_password`
+    fi
+
+    CURRENT_TIME=`date -u`
+    printf "$KEE_PASS_DB_PASS\n$GENERATED_PASS" | $KEE_PASS edit "$KEE_PASS_DB" "$KEY_PATH" -p \
```
joxl, line 39:
```suggestion
    printf "$KEE_PASS_DB_PASS\n$GENERATED_PASS" | "$KEE_PASS" edit "$KEE_PASS_DB" "$KEY_PATH" -p \
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+    fi
+}
+
+rotate_vault_key () {
+    init_keepass
+    check_for_previous_run
+    prompt_for_password
+
+    # Read the current vault password from the keepass file
+    CURRENT_PASS=`echo "$KEE_PASS_DB_PASS"|$KEE_PASS show "$KEE_PASS_DB" "$KEY_PATH" -a password -q`
```
joxl, line 81:
```suggestion
    CURRENT_PASS=`echo "$KEE_PASS_DB_PASS" | "$KEE_PASS" show "$KEE_PASS_DB" "$KEY_PATH" -a password -q`
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+
+    if [ -z $CURRENT_PASS ]; then
```
joxl, line 83:
```suggestion
    if [ -z "$CURRENT_PASS" ]; then
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+        echo "Invalid password"
+        exit 1
+    fi
+
+    # Generate a new password
+    GENERATED_PASS=`$KEE_PASS generate`  # Can add specific complexity requirements as args
```
joxl, line 89:
```suggestion
    GENERATED_PASS=`"$KEE_PASS" generate`  # Can add specific complexity requirements as args
```

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+        --new-vault-id scripts/echo_stdin-client.sh \
+        environments/swiss/vault.yml > /dev/null 2>&1
+
+    if [ $? -ne 0 ]; then
+        echo "KeePass password does not match existing vault password"
+        exit 1
+    fi
+
+    # Store the generated password in a temporary database for retrieval later
+    store_password $GENERATED_PASS
```
joxl, line 103:
```suggestion
    store_password "$GENERATED_PASS"
```

joxl (changes requested):
Mostly extra quotes needed.  I have more concerns about the `echo_stdin-client.sh` script, but I'm curious if `cat` can solve that.

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+
+prompt_for_password () {
+    echo "Enter password to unlock $KEE_PASS_DB:"
+    read -s KEE_PASS_DB_PASS
+}
+
+store_password () {
+    local password=$1
+
+    printf "$KEE_PASS_DB_PASS\n$KEE_PASS_DB_PASS" | $KEE_PASS db-create "$TEMP_DATABASE" -p > /dev/null 2>&1
```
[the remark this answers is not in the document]
mjriley: I think I'm opting against this. I don't think it encourages good practices to expose the password before the main entry is updated. From a practical standpoint, it also makes it more difficult to determine if there was a previous run to resume from.

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+        exit 1
+    fi
+
+    # Generate a new password
+    GENERATED_PASS=`$KEE_PASS generate`  # Can add specific complexity requirements as args
+
+    # Update the ansible vault
+    printf "$CURRENT_PASS\n$GENERATED_PASS\n" | ansible-vault rekey \
+        --vault-id scripts/echo_stdin-client.sh \
+        --new-vault-id scripts/echo_stdin-client.sh \
```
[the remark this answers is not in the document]
mjriley: I'll be honest, I hadn't thought about using this. That said, after trying, it doesn't work. There are two potential issues here -- I only have one stdin feed, but need two passwords. `cat` echoes out everything in stdin, so I only get the first password, not the second. The second, lesser issue, is that `ansible-vault` actually changes its behavior based on whether or not the id script ends in `-client` -- if it does, it passes a `--vault-id` parameter to the script. It's not used here, but it's still the 'most correct' behavior according to [the ansible docs](https://docs.ansible.com/ansible/latest/user_guide/vault.html)

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
joxl: Maybe you're confusing `set -x` (print commands and their arguments as they are executed) with `set -e` (exit immediately if a command exits non-zero)?  I think it's important to use `set -e` in "production" scripts like this to guard against weird edge cases in the event a command unexpectedly fails (exits non-zero).

scripts/rotate_vault.sh
```diff
@@ -0,0 +1,147 @@
...
+init_keepass () {
+    # Locate keepassxc-cli or fail with error
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        KEE_PASS=/Applications/KeePassXC.app/Contents/MacOS/keepassxc-cli
+    elif [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        KEE_PASS=`which keepassxc-cli`
+    fi
```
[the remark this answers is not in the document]
joxl: I'm fine with either. Supporting a configurable `KEE_PASS` env var might be virtuous. I think the scenario that stuck out to me the most was a system that is neither macOS or Linux.  In this scenario, the entire `if` block will be an effective no-op, and any value that happens to exist in the env's `KEE_PASS` var (if one exists) will be attempted.  For 100% safety, I think at least this is warranted:

```bash
else
    echo "unsupported OS: $OSTYPE"
    exit 1
fi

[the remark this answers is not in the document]
Comment (joxl) [reply scripts/rotate_vault.sh]: With the added user-configurable `KEE_PASS` var now supported (and most importantly, `echo`'d if present), I'm not sure this `else` clause is so important anymore, although it's still more "strict", which I do like (in theory it prevents a situation that might otherwise be confusing to a user). That being said, if someone is trying to run this on OpenBSD/`<insert your Unix variant here>`/etc, I think they'll likely have other bigger fish to fry.

File: scripts/rotate_vault.sh (hunk 14 above)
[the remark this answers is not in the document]
Comment (mjriley) [reply scripts/rotate_vault.sh]: No, the issue is that the internal commands being executed have their own output. I'm piping that to /dev/null in some cases so it is hidden from the user and I can control the messaging. If use `set -e`, anything that fails and is piped to /dev/null is just going to be completely unknown to the user, because I have no chance to catch it, right?

File: scripts/rotate_vault.sh (hunk 13 above)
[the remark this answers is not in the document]
Comment (joxl) [reply scripts/rotate_vault.sh]: > changes its behavior based on whether or not the id script ends in `-client`

Ugh, this is too "clever" in my book (would be much better to have a separate dedicated commandline arg for this, but I digress). Thank you for pointing this out, I was actually worried why/how this works at all.

Being as you can't get away from the `-client.<ext>`, I'm curious if this works:

```diff
-read -r line
-echo "$line"
+head -n1
```

scripts/echo-stdin-client.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# This script just aims to provide a way for ansible-vault to read standard input.
+# ansible-vault doesn't let you specify passwords on the command line -- you either
+# need to provide a password file or a password script. This functions as the latter,
+# reading passwords piped in
+read -r line
+echo "$line"
```
joxl, line 8: nit: missing newline at end of file.

scripts/rotate-swiss-vault.sh
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env bash
+
+init_keepass () {
+    # Locate keepassxc-cli or fail with error
+    if [[ -x "$KEE_PASS" ]]; then
+        echo "Running with explicit KeePass binary at: $KEE_PASS"
```
joxl, line 6: `echo` :star:

joxl (approved):

scripts/rotate_vault.sh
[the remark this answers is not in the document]
joxl: You could catch such a command with something like:

```bash
if ! mission_critical_command --opt foo >/dev/null; then
    echo "mission failed for foo"
    exit 1
fi

**Comment by mjriley** [comment]
> We are in the process of switching from KeePass to 1Password. Are you planning to port this script when that change is made?

It appears that 1Password has its own command line tools, so yeah, I don't think porting this should pose a challenge when we do make the switch

File: scripts/rotate-swiss-vault.sh [hunk 18]
```diff
@@ -0,0 +1,148 @@
...
+    elif [[ "$OSTYPE" == "darwin"* ]]; then
+        KEE_PASS=/Applications/KeePassXC.app/Contents/MacOS/keepassxc-cli
+    elif [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        KEE_PASS=`which keepassxc-cli`
+    fi
+
+    if [ ! -x "$KEE_PASS" ]; then
+        echo "KeePassXC was not found. Please go to https://keepassxc.org to install it"
```
millerdev, line 16: nit
```suggestion
        echo "KeePassXC was not found. Please go to https://keepassxc.org to install it"
        echo "or set the KEE_PASS environment variable to the path of keepassxc-cli"
```

millerdev (commented):
Looks good, pending @joxl's final review.

scripts/rotate-swiss-vault.sh
```diff
@@ -0,0 +1,151 @@
...
+    read -s KEE_PASS_DB_PASS
+}
+
+store_password () {
+    local password="$1"
+
+    { echo "$KEE_PASS_DB_PASS"; echo "$KEE_PASS_DB_PASS"; } | \
+        "$KEE_PASS" db-create "$TEMP_DATABASE" -p > /dev/null 2>&1
+    { echo "$KEE_PASS_DB_PASS"; echo "$password"; } | \
+        "$KEE_PASS" add "$TEMP_DATABASE" "$TEMP_KEY" -p -q
```
joxl, line 33: Does this mean that these passwords _cannot_ have an actual newline in them (these commands would certainly fail if the db password contained a newline).  Maybe guard against that scenario? Could be done here, but might be better to check immediately after each password is first known to the script.

```bash
if [ $(echo "$KEE_PASS_DB_PASS" | wc -l) -ne 1 ]; then
    echo 'invalid db password (passwords may not contain newlines)'
    exit 1
fi

# snip ...

if [ $(echo "$password" | wc -l) -ne 1 ]; then
    echo 'invalid vault password (passwords may not contain newlines)'
    exit 1
fi
```

mjriley: My testing works fine with newlines. I don't think multiple echo command emit a '\n' character

joxl: > a developer would have to go out of their way to intentionally troll us

Or `keepassxc-cli generate` (or whatever binary the user provides in its place) emits more than one line of text on STDOUT.  Since the value comes from an executable outside of this script's "control", I think it's important to employ the robustness principle here (i.e. this script should refuse to encrypt the vault with value that the user does not know, and it can detect is a malformed password).  If the "password generate" command for some reason wrote multiple lines of output, the end user would have no idea something other than what they expected actually happened.

Consider the following where:
- `KEE_PASS_DB_PASS='something the user typed'`
- `password='multi-line↲output↲'` (Note: `↲` represents a newline (ASCII character 10) -- I know, that's the unicode symbol for a carriage return, but you get the idea).

```bash
    { echo "$KEE_PASS_DB_PASS"; echo "$password"; } | \
        "$KEE_PASS" add "$TEMP_DATABASE" "$TEMP_KEY" -p -q
```
The key used to encrypt the vault is `multi-line↲output`, and the value saved in the temporary database is `multi-line`, but the user has no way of knowing this. In this scenario, all of the following are true:
- the vault is encrypted with a new key
- the user can unlock the keepass db with the password they provided (i.e. the db appears fine)
- the value stored in the keepass db is incorrect (cannot unlock the vault)
- the user does not know that the value in the keepass db is incorrect

joxl: Another solution could be (if we're willing to assume the _first_ line of output is always a "good password".
```bash
GENERATED_PASS=`"$KEE_PASS" generate | head -n1`
```

mjriley: I still feel like this sounds like over-engineering for a situation that will never happen, and should never happen. I can't find any option within KeePass to generate a password with a newline (apart from perhaps manually adding that character in the 'Additional Characters' field). I configured 1Pass to look at its API, and it provides even less configurability with passwords than KeePass does. A password with a newline character is going to cause issues all over the place -- it's not a coincidence that these generators don't give you the option to include that character.

Sure, I could add this check to make it more robust, but if that scenario should never come up, it's additional code that has to be understood and maintained. In javascript, for example, I don't tend to check that my string parameters are actually strings and not numbers, unless they could actually be passed as numbers by mistake. I just don't think there is any realistic case where a password contains a newline.

That said, if my assumptions are incorrect, and this does happen, it is recoverable. The encrypted vault can be rolled back to its previous version in GitHub, and both KeePass and 1Pass have records of their previous passwords, so we can restore the version prior to the faulty one.

joxl: > it's additional code that has to be understood and maintained
:heavy_plus_sign: 

I'm satisfied to leave it as-is.

joxl (approved):

millerdev:
This should be updated to use 1Password now since we are no longer updating passwords in KeePass.

millerdev (changes requested):
This should be updated to use 1Password now since we are no longer updating passwords in KeePass.

mjriley:
That mode should be supported, but I don't believe SaaS is currently ready to use 1Pass for these passwords. Should this hold up this specific PR? It might be cleaner to merge this as is, and open a second PR to address 1Pass compatibility

millerdev (approved):
Offlined with Matt. Makes sense to merge this now and then update to 1Password soon.

PR #5250, opened 2022-04-08
Add alert to old docs, remove alert on new docs

sravfeyn:
https://dimagi-dev.atlassian.net/browse/SC-2021

I have added a warning to the old docs at http://dimagi.github.io/commcare-cloud/ that it will be removed and added a link to the new docs. I will merge this after the signoff from other divisions. 

Note that, there's no way to display a sitewide banner at the top, so I have updated the title.

Screenshot 2022-04-08 at 11 21 58 AM
,


sravfeyn:
@dannyroberts @proteusvacuum

proteusvacuum (approved):
🚀

dannyroberts:
> after the signoff from other divisions
Thanks I have some (I think minor?) concerns that I'd like to offline about before this goes forward and make sure all our teams are on the same page before this change goes out.

sravfeyn:
Now that we discussed and resolved the [concerns](https://docs.google.com/document/d/1kifgKiOQ-3ZCN4K-4h2N6K0B77MPDiqEfgvF65Tz5rs/edit#), this is how we will announce. @dannyroberts

dannyroberts:
@sravfeyn this seems like an appropriate PR in which to switch any links that point to the old docs to point to the new docs instead. For example I can see in the README.md of this repo we point people to https://dimagi.github.io/commcare-cloud/. Even a bunch of pages on the _new_ docs point to the old docs. I think we want to sort that out in this PR so that all the changes that officially mark the new one as the correct one and the old one as the "deprecated" one happen at the same time.

sravfeyn:
Good point, I have updated the links here.

sravfeyn:
@dimagi/team-commcare-hq FYI, goingforward if you are updating docs on commcare-cloud, please update them for https://commcare-cloud.readthedocs.io/ instead of https://dimagi.github.io/commcare-cloud/. There is a README.md [here](https://github.com/dimagi/commcare-cloud/blob/master/hosting_docs/README.md) that explains how to update the docs going forward.

dannyroberts:
I missed this during my review in July, but just a reminder to always wait for PR approval before merging.

sravfeyn:
There was an [approval](https://github.com/dimagi/commcare-cloud/pull/5250#pullrequestreview-936436233) and I think this was also discussed over voice?

dannyroberts:
Oh, I missed Farid's approval check mark because it wasn't the green color I was expecting. I actually don't know why it's grey rather than green: according to https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/approving-a-pull-request-with-required-reviews a gray check means it was approved by someone without write permissions, but that seems incorrect for Farid at the time. Regardless, thanks for following up to point it out.

The only thing I want to add now that I've been corrected, is that even though you're saying the final commits were discussed over voice, it does not _document_ the approval, so in a situation like that, remember to get the person who verbally agreed to actually approve those final commits explicitly in the PR. Other than that, this is all set, no further action required.

sravfeyn:
Well noted Danny, thanks.

PR #5252, opened 2022-04-11
[EOL] Update node.js to 14.x and npm to 7.x

biyeun:
This PR will upgrade the version of Node.js that we use to `14.19.1` and `npm` to `7.24.2` (the two libraries are tightly coupled). Node.js `v12` is reaching it's EOL on April 30th.

https://dimagi-dev.atlassian.net/browse/SAAS-13399

We will also need to roll out changes to commcare-cloud to ensure that node.js is upgraded in our build process and that third party developers are notified.

see corresponding HQ PR here https://github.com/dimagi/commcare-hq/pull/31408

##### ENVIRONMENTS AFFECTED
all environments

There will also be a PR with a changelog alerting third parties of the upgrade.

### DIFF
--- a/src/commcare_cloud/ansible/roles/nodejs/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/nodejs/tasks/main.yml
@@ -30,7 +30,7 @@
   npm: name="{{ item.name }}" version="{{ item.version }}" state=present global=yes
   become: yes
   with_items:
-    - {name: 'n', version: '14.19.1'}
+    - {name: 'n', version: '2.1.12'}
     - {name: 'less', version: '3.10.3'}
     - {name: 'yarn', version: '1.22.4'}
     - {name: 'bower', version: '1.8.4'}


src/commcare_cloud/ansible/roles/nodejs/tasks/main.yml
```diff
@@ -48,6 +48,6 @@
   npm: name="{{ item.name }}" version="{{ item.version }}" state=present global=yes
   become: yes
   with_items:
-    - {name: 'npm', version: '6.14.5'}
+    - {name: 'npm', version: '7.24.2'}
```
dannyroberts, line 51: I see npm is updated both here and in commcare-hq `package.json`/`yarn.lock`. Is that because this manages the global npm install and that (`package.json`) manages a local `npm` install, and we need both? Curious if you know what we need each for

dannyroberts (approved):

src/commcare_cloud/ansible/roles/nodejs/tasks/main.yml
biyeun: yes, npm is needed for the `yarn` install and we also ask `yarn` to build packages from certain `npm` repos

PR #5145, opened 2022-02-08
Create Google Oauth Config variable to overwrite settings.py

RynhardPietersen:
[link](https://github.com/dimagi/commcare-hq/pull/30982) to pr which prompted this change

I'm simply adding some secrets for use in localsettings.py in order to overwrite the client secrets dict in settings.py

PR #5255, opened 2022-04-12
added redirect uris to google_oauth_config setting

RynhardPietersen:
forgot to add the redirect uris in [this pr](https://github.com/dimagi/commcare-cloud/pull/5145)

##### ENVIRONMENTS AFFECTED

staging

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -349,7 +349,8 @@ GOOGLE_OAUTH_CONFIG = {
         "auth_uri": "https://accounts.google.com/o/oauth2/auth",
         "token_uri": "https://oauth2.googleapis.com/token",
         "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
-        "client_secret": '{{ GOOGLE_OAUTH_CLIENT_SECRET }}'
+        "client_secret": '{{ GOOGLE_OAUTH_CLIENT_SECRET }}',
+        "redirect_uris": ["https://staging.commcarehq.org/oauth/google_sheets_oauth/callback/"],
     }
 }
 


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -349,7 +349,8 @@ GOOGLE_OAUTH_CONFIG = {
         "auth_uri": "https://accounts.google.com/o/oauth2/auth",
         "token_uri": "https://oauth2.googleapis.com/token",
         "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
-        "client_secret": '{{ GOOGLE_OAUTH_CLIENT_SECRET }}'
+        "client_secret": '{{ GOOGLE_OAUTH_CLIENT_SECRET }}',
+        "redirect_uris": ["https://staging.commcarehq.org/oauth/google_sheets_oauth/callback/"],
```
gherceg, line 353: Just curious why this needs to be a list? Is it a config constraint or is there a plan to add more uris to this in the future?

gherceg (approved):

src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
RynhardPietersen: I will need to add the prod uri once we push gsheets to prod, and then probably like india and swiss if gsheets is going to be used on those environments.

biyeun: is there any way to add this outside of the localsettings config here? It would be nice to build this url from referencing the actual view. I feel like this URL does not need to live here like this

biyeun (changes requested):
I think the `redirect_uris` should not be specified in settings, but added to the config when needed

RynhardPietersen:
closing this because it would be better to specify the redirect uri before sending the config to google!

PR #5266, opened 2022-04-21
Fix deploy notice (staging.yml)

joxl:
Staging branches YAML filename was changed in dimagi/commcare-hq#31446

##### ENVIRONMENTS AFFECTED
- staging


orangejenny (approved):
thanks!

PR #5268, opened 2022-04-25
adding virtualenv path for upload script. It is failing with default python

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-13494
##### ENVIRONMENTS AFFECTED

All

@dannyroberts @shyamkumarlchauhan 
It seems like the Previous swiss backup scripts were edited on the server directly. Because of that when we ran the `deploy-stack` command on the swiss server recently, it has replaced the existing backup scripts, and backups were not `uploaded` to the s3 bucket. 

As a fix, I'm adding the cchq python virtualenv path to the python script as the using the default Python (2.7) the scripts are failing. Tested this process on the server and it is working as expected. 
image

--
Need Suggestions: 
adding {{ deploy_env }} is right option or {{ virtualenv_home }} which one is the right option to choose. Please suggest.

### DIFF
--- a/src/commcare_cloud/ansible/roles/backups/templates/create_blobdb_backup.sh.j2
+++ b/src/commcare_cloud/ansible/roles/backups/templates/create_blobdb_backup.sh.j2
@@ -3,11 +3,13 @@ BACKUP_TYPE=$1
 MINUTES_TO_RETAIN_BACKUPS=$2
 DAYS_TO_RETAIN_BACKUPS=$2
 HOSTNAME=$(hostname)
+ENV="{{ env_name }}"
 TODAY=$(date +"%Y_%m_%d")
 HOUR=$(date +"%Y_%m_%d_%H")
 BACKUP_FILE="blobdb_${BACKUP_TYPE}_${TODAY}.tar.gz"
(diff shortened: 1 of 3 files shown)


dannyroberts (approved):

src/commcare_cloud/ansible/roles/backups/templates/create_blobdb_backup.sh.j2
```diff
@@ -3,11 +3,13 @@ BACKUP_TYPE=$1
 MINUTES_TO_RETAIN_BACKUPS=$2
 DAYS_TO_RETAIN_BACKUPS=$2
 HOSTNAME=$(hostname)
+ENV="{{ env_name }}"
```
dannyroberts, line 6: @prem-fissionhq it looks like this resulted an an error for someone else trying to run this today. 

https://forum.dimagi.com/t/ansible-sudo-password-not-being-passed-to-deploy-commcare-hq-code/9015/2

I'm going to revert this PR and let you test it again to see if you can reproduce.

skodde:
Corrected in #5458

PR #5271, opened 2022-04-25
SSH directly to any machine via SSM

millerdev:
This removes the need for the `--control` parameter in `cchq --control <env> ssh <host>`

:blowfish: Review by commit.

##### ENVIRONMENTS AFFECTED
AWS SSM environments

### DIFF
--- a/src/commcare_cloud/commands/ansible/helpers.py
+++ b/src/commcare_cloud/commands/ansible/helpers.py
@@ -64,18 +64,18 @@ def get_common_ssh_args(environment, use_factory_auth=False):
     return cmd_parts_with_common_ssh_args
 
 
-def get_default_ssh_options(environment, use_aws_ssm_with_instance_id=None):
+def get_default_ssh_options(environment, aws_ssm_target=None):
     default_ssh_options = []
 
     strict_host_key_checking = environment.public_vars.get('commcare_cloud_strict_host_key_checking', True)
-    if not strict_host_key_checking or use_aws_ssm_with_instance_id:
+    if not strict_host_key_checking or aws_ssm_target:
         default_ssh_options.append(('StrictHostKeyChecking', 'no'))
 
     known_hosts_filepath = environment.paths.known_hosts
-    if os.path.exists(known_hosts_filepath) and not use_aws_ssm_with_instance_id:
+    if os.path.exists(known_hosts_filepath) and not aws_ssm_target:
         default_ssh_options.append(('UserKnownHostsFile', known_hosts_filepath))
 
-    if use_aws_ssm_with_instance_id:
+    if aws_ssm_target:
         sso_command = (
             'aws ssm start-session --target %h '
             '--document-name AWS-StartSSHSession '
@@ -86,14 +86,14 @@ def get_default_ssh_options(environment, use_aws_ssm_with_instance_id=None):
             ProxyCommand below if it becomes necessary to put quotation
             marks in sso_command. Otherwise it is more readable as is."""
         default_ssh_options.append(('ProxyCommand', f'sh -c "{sso_command}"'))
-        default_ssh_options.append(('HostName', use_aws_ssm_with_instance_id))
+        default_ssh_options.append(('HostName', aws_ssm_target))
 
     return default_ssh_options
 
 
-def get_default_ssh_options_as_cmd_parts(environment, original_ssh_args=(), use_aws_ssm_with_instance_id=None):
+def get_default_ssh_options_as_cmd_parts(environment, original_ssh_args=(), aws_ssm_target=None):
     ssh_args = []
-    options = get_default_ssh_options(environment, use_aws_ssm_with_instance_id)
+    options = get_default_ssh_options(environment, aws_ssm_target)
     for option_name, default_option_value in options:
         arg_names = ('{}='.format(option_name), "-o{}=".format(option_name))
         if not any(a.startswith(arg_names) for a in original_ssh_args):
--- a/src/commcare_cloud/commands/ansible/helpers.py
+++ b/src/commcare_cloud/commands/ansible/helpers.py
@@ -76,7 +76,16 @@ def get_default_ssh_options(environment, use_aws_ssm_with_instance_id=None):
         default_ssh_options.append(('UserKnownHostsFile', known_hosts_filepath))
 
     if use_aws_ssm_with_instance_id:
-        default_ssh_options.append(('ProxyCommand', 'sh -c "aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters portNumber=%p"'))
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -64,28 +64,41 @@ def get_common_ssh_args(environment, use_factory_auth=False):
...
+            # hostname is aws_ssm_target (rather than an IP address)
+            f'aws ssm start-session --target {aws_ssm_target} '
+            '--document-name AWS-StartSSHSession '
+            '--parameters portNumber=%p'
+        )
+        assert '"' not in sso_command, """
+            Use {shlex_quote(sso_command)} and remove quotes (") in the
+            ProxyCommand below if it becomes necessary to put quotation
+            marks in sso_command. Otherwise it is more readable as is."""
+        default_ssh_options.append(('ProxyCommand', f'sh -c "{sso_command}"'))
```
millerdev, line 91: Is there a reason why we have to use this `sh -c "..."` wrapper for the proxy command? Would it work just as well to do this?
```suggestion
        default_ssh_options.append(('ProxyCommand', sso_command))
```

dannyroberts: I believe this comes straight from the AWS documentation https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-getting-started-enable-ssh-connections.html:

```
# SSH over Session Manager
host i-* mi-*
    ProxyCommand sh -c "aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'"
```

so that's where I got it from initially. I'm not sure whether it can be removed or not and why they chose to put it in there.

millerdev: I noticed that it was in AWS docs as well, but I'd like to know why it's there and if it's required?

joxl: The command (`aws sso ...`) doesn't use any shell-specific functionality as we're using it here. I have seen examples of it used like `ProxyCommand sh -c 'AWS_PROFILE=<profile>:session aws ssm start-...'` where the shell is needed for the environment it provides.  As long as all of this is setup as desired by the thing calling `ssh`, I would assume removing the sh "wrapper" should be fine.

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -74,22 +74,30 @@ class _Ssh(Lookup):
...
         cmd_parts = [self.command, address, '-t'] + ssh_args
         cmd = ' '.join(shlex_quote(arg) for arg in cmd_parts)
         if not args.quiet:
             print_command(cmd)
         return subprocess.call(cmd_parts, **({'env': env_vars} if env_vars else {}))
 
+    def get_address_tuple(self, args):
+        address = self.lookup_server_address(args)
+        if ':' in address:
+            address, port = address.split(':')
```
joxl, line 98: The scenario of `some.host:` will not raise an exception here and will return an an empty non-None port (i.e. `port == ''`). Is that ok (in the scope of this function)?

src/commcare_cloud/commands/ansible/helpers.py
millerdev: Removed. We can always add it back if needed.

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
```diff
@@ -118,24 +126,21 @@ def run(self, args, ssh_args):
...
+                ))
+                puts(
+                    f"{color_notice('See ')}{color_link(self.INSTALL_SSM)}"
+                    f"{color_notice(' for instructions.')}"
+                )
+                return -1
+            _, host, _ = self.get_address_tuple(args)
+            use_aws_ssm_with_instance_id = environment.get_host_vars(host)['ec2_instance_id']
+            env_vars = os.environ.copy()
+            env_vars.update({'AWS_PROFILE': aws_sign_in(environment)})
```
joxl, line 151: Is there a reason we can't use the aws `--profile` option instead? (I've been meaning to look into this for a while now.)

```bash
$ aws ssm -h
usage: aws [-h] [--profile PROFILE] [--debug]
...
```

src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -84,11 +84,7 @@ def get_default_ssh_options(environment, aws_ssm_target=None):
             '--document-name AWS-StartSSHSession '
             '--parameters portNumber=%p'
         )
-        assert '"' not in sso_command, """
-            Use {shlex_quote(sso_command)} and remove quotes (") in the
-            ProxyCommand below if it becomes necessary to put quotation
-            marks in sso_command. Otherwise it is more readable as is."""
-        default_ssh_options.append(('ProxyCommand', f'sh -c "{sso_command}"'))
+        default_ssh_options.append(('ProxyCommand', sso_command))
```
joxl, line 87: The `sso_command` arg should be quoted.

joxl (commented):
I'm really excited for this. Thanks!

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
millerdev: I'm not sure. This code did not change other than the if-block was un-indented.

src/commcare_cloud/commands/ansible/helpers.py
millerdev: The previous `sh -c ...` command was not quoted here. The option gets quoted later by `subprocess.call()`

https://github.com/dimagi/commcare-cloud/blob/f44bb73689131adaf04592c258e15128d7118aab/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py#L89-L93

The `print_command()` output looks like this:

`-o 'ProxyCommand=aws ssm start-session --target ... --document-name AWS-StartSSHSession --parameters portNumber=%p'`

joxl: Ah, brilliant!

joxl (approved):

src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
millerdev: I should have pointed out earlier that the commit where this "changed" is easier to review with whitespace changes hidden:

https://github.com/dimagi/commcare-cloud/pull/5271/commits/e9ba2770aa2765d58cf5183c5f466934a249e382?diff=unified&w=1

dannyroberts (approved):

PR #5274, opened 2022-04-26
Revert "Update public.yml file ALTERNATE_HOSTS"

dannyroberts:
Reverts dimagi/commcare-cloud#5114

This was creating a hard-to-resolve terraform diff. Offlined with @Charl1996; looks like this was a little premature and the project team can redo this change when they're ready.

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -25,7 +25,6 @@ soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
 new_domain_email: inquiries@dimagi.com
 ALTERNATE_HOSTS:
   - rec-mobile.sante.gov.bf
-  - commcare.health.gov.mw
 
 DATADOG_ENABLED: True
 


Charl1996 (approved):

sravfeyn:
Hey @dannyroberts why was this reverted?

It looks like the change is still live on the proxy machine, anyway. I would like to update this again, so that if someone else deploys the change doesn't get lost.

dannyroberts:
> Offlined with @Charl1996; looks like this was a little premature and the project team can redo this change when they're ready.

You might want to take this up with Charl. At the time I asked him whether to we wanted to try and resolve the conflict or just revert and let him reimplement it when it was ready, and he chose the latter.

sravfeyn:
Cool, thanks for the context. I have added a PR as the migration is being planned this week. https://github.com/dimagi/commcare-cloud/pull/5357 if there is a conflict, I will resolve it.

PR #5282, opened 2022-05-02
Ce/redis read replica

calellowitz:
This allows us to specify multiple redis hosts in localsettings, where the first is the primary and the others are read replicas, to share load across our cluster. It also configures staging to use that setup.

##### ENVIRONMENTS AFFECTED

Staging

### DIFF
--- a/environments/staging/public.yml
+++ b/environments/staging/public.yml
@@ -158,7 +158,7 @@ localsettings:
   SES_CONFIGURATION_SET: staging-email-events
   SNS_EMAIL_EVENT_SECRET: "{{ SNS_EMAIL_EVENT_SECRET }}"
   REDIS_ENDPOINTS:
-    - host: "redis.production.commcare.local"
+    - host: "redis.staging.commcare.local"
       port: '6379'
     - host: "staging-elasticache-cluster-ro.ormqor.ng.0001.use1.cache.amazonaws.com"
       port: '6379'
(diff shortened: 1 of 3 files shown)


dannyroberts (approved):
Before merging it would be good to check that this does not make an unwanted diff in production/india/any environment where this hasn't been changed

snopoke:
Should we be concerned about this statement in the docs:
> WARNING: Replication setup is not heavily tested in production environments.
https://github.com/jazzband/django-redis#default-client

calellowitz:
> Should we be concerned about this statement in the docs:
> 
> > WARNING: Replication setup is not heavily tested in production environments.
> > https://github.com/jazzband/django-redis#default-client

I am not super concerned by that statement. I read the code that handled the relevant bits and it seemed fine, but this is also far less urgent now since we fixed the root cause of the redis network issue, so I would be fine to hold off on rolling this out further. Notably in django 4.0, this is supported by the new native redis backend, so worst case, we could wait until then https://docs.djangoproject.com/en/4.0/topics/cache/#redis-1, but I think the template changes to support this are worth leaving around.

calellowitz:
Closing this for now. There are places in HQ where we write to redis and read back what we have written in the same request, so even milliseconds of replication lag can can cause errors. Since we got to the root cause of the bandwidth issues, I am not sure this is worth the effort of fixing the code in those places.

Its also a bit hard to tell what current lag is since for our version of redis in elasticache they only report lag in 1 second resolution, and current lag is too low to report. Later versions report in millisecond resolution, so once we have that data we may be better able to figure out how easy this will be.

snopoke:
Can you link to one of those places? If they are using quickcache then it's also cached in memory on the webworker.

calellowitz:
> Can you link to one of those places? If they are using quickcache then it's also cached in memory on the webworker.

https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/case_importer/views.py#L116

This method both puts the upload file from the case import into redis and reads it from redis. There may be other places that do the same, but when I realized it wasn't going to be super straightforward to find out, I decided to table this work for now.

PR #5284, opened 2022-05-10
only fetch the required branch

calellowitz:
https://dimagi-dev.atlassian.net/browse/SAAS-13513

Followup from a recent deploy issue where a bad branch in a submodule caused the repo size to baloon, but also just good practice. This may even make creating releases marginally faster by cutting down what needs to be fetched over the network from github each time.

##### ENVIRONMENTS AFFECTED

All

### DIFF
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -75,7 +75,7 @@ def _update_code_from_previous_release(code_repo, subdir, git_env):
             sudo('git remote set-url origin {}'.format(code_repo))
     else:
         with shell_env(**git_env):
-            sudo('git clone {} {} --single-branch'.format(code_repo, code_root))
+            sudo('git clone --single-branch {} {}'.format(code_repo, code_root))
 
 
 def _get_submodule_list(path):
@@ -125,7 +125,7 @@ def _clone_code_from_local_path(from_path, to_path, run_as_sudo=True):
     ]
 
     with cd(from_path):
-        cmd_fn('git clone {}/.git {}'.format(
+        cmd_fn('git clone --single-branch {}/.git {}'.format(
             from_path,
             to_path
         ))
@@ -134,7 +134,7 @@ def _clone_code_from_local_path(from_path, to_path, run_as_sudo=True):
         cmd_fn('git config receive.denyCurrentBranch updateInstead')
         if git_local_submodule_config:
             cmd_fn(' && '.join(git_local_submodule_config))
-        cmd_fn('git submodule update --init --recursive')
+        cmd_fn('git submodule update --init --recursive --single-branch')
         if git_remote_submodule_config:
             cmd_fn(' && '.join(git_remote_submodule_config))
 
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -68,8 +68,6 @@ def _update_code_from_previous_release(code_repo, subdir, git_env):
         code_root = os.path.join(code_root, subdir)
 
     if files.exists(code_current, use_sudo=True):
-        with cd(code_current), shell_env(**git_env):
-            sudo('git submodule foreach "git fetch origin"')
         _clone_code_from_local_path(code_current, code_root)
         with cd(code_root):
             sudo('git remote set-url origin {}'.format(code_repo))
--- a/src/commcare_cloud/fab/operations/release.py
+++ b/src/commcare_cloud/fab/operations/release.py
@@ -47,11 +47,11 @@ def update_code(full_cluster=True):
             sudo('git remote prune origin')
             # this can get into a state where running it once fails
             # but primes it to succeed the next time it runs
-            sudo('git fetch origin --tags -q || git fetch origin --tags -q')
+            sudo(f'git fetch origin {env.code_branch} --tags -q || git fetch origin {env.code_branch} --tags -q')
             sudo('git checkout {}'.format(git_tag))
             sudo('git reset --hard {}'.format(git_tag))
             sudo('git submodule sync')
-            sudo('git submodule update --init --recursive -q')
+            sudo('git submodule update --init --recursive --single-branch -q')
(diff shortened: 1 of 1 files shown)


snopoke (approved):
Does this also work with Git commits and tags?

joxl:
> Does this also work with Git commits and tags?

For submodules, yes. A `git clone` command always leaves HEAD at the top of the default remote branch, so commits and tags aren't relevant for that (as in, those need to be checked out after a clone even without `--single-branch`).

millerdev:
> Does this also work with Git commits and tags?

Good question. My hunch is that it will probably work for tags. As for commits, I think will only work for commits that are ancestors of `env.code_branch`, which is probably not enough in most cases.

calellowitz:
> Does this also work with Git commits and tags?

I think the replies here are talking past each other a bit (one reply appears to be about submodules while the other looks like it is about the hq deploy ref), but assuming you are asking if the `--commcare-rev` argument passed to the deploy command is a tag or a commit, will this code work, the answer is yes. Specifically the `env.code_branch` argument @millerdev mentions is the value of that argument, and `git fetch origin <code_branch> --tags` will successfully grab that commit (and any missing ancestor commits) from github whether the value passed in is a branch, commit hash, or tag. However, we do have a different env variable, `env.deploy_ref`, that converts whatever is passed in to a commit hash, so if people are more comfortable knowing that value will always be a commit hash, it is a very easy switch to make.

> I think will only work for commits that are ancestors of env.code_branch, which is probably not enough in most cases.

@millerdev 
Can you explain a bit more what you mean here? What do you mean by it working for commits that are ancestors of `code_branch`? You are correct that it will only pull down commits that are ancestors of `code_branch`, but the deploy is to the commit marked by `code_branch`, so I am not sure how that could be an issue.

As far as submodules, since those are always set to a commit hash, this will work there as well.

@snopoke @millerdev 
Does that resolve any concerns?

millerdev:
@calellowitz I was mostly concerned that `git fetch origin <ref>` would not work when `<ref>` is a commit hash, but a quick local test reveals that it does work. You can ignore my comment. I have no further concerns.

calellowitz:
> @calellowitz I was mostly concerned that `git fetch origin <ref>` would not work when `<ref>` is a commit hash, but a quick local test reveals that it does work. You can ignore my comment. I have no further concerns.

Yeah, I did not know it could do that until yesterday. I think it's a new(ish) feature (2015, but who really keeps on top of these things).

snopoke:
@calellowitz yes, my question was about arguments passed to `--commcare-rev`. So long as that works ✅

dannyroberts (approved):
Looks good to me. Cal your suggestion about potentially using `env.deploy_ref` instead sounds good, and logically clearer, but I think not strictly necessary. I.e. if you use `env.code_branch` is `master` and master changes during the deploy such that `env.deploy_ref` is now behind master, there's very little harm in fetch latest master anyway, which will contain `env.deploy_ref`.

If you follow along and agree with my reasoning above then I'm happy either way (keeping as is or changing)

src/commcare_cloud/fab/operations/release.py
```diff
@@ -68,8 +68,6 @@ def _update_code_from_previous_release(code_repo, subdir, git_env):
         code_root = os.path.join(code_root, subdir)
 
     if files.exists(code_current, use_sudo=True):
-        with cd(code_current), shell_env(**git_env):
-            sudo('git submodule foreach "git fetch origin"')
```
calellowitz, line 72: This removal is the one concern I have. This fetch appears to be unnecessary and grabs all branches, but it was added in a standalone PR (https://github.com/dimagi/commcare-cloud/pull/2787) so ostensibly was necessary at one point. However, it should never be necessary to fetch in the submodule since `git submodule update` handles that.

joxl: Agreed. This is what `submodule update` does for you. Furthermore, performing a `foreach 'git fetch origin'` will do exactly what this PR intends to stop (indiscriminately fetch _all_ remote branches), so it simply needs to go, one way or another.

src/commcare_cloud/fab/operations/release.py
```diff
@@ -47,11 +47,11 @@ def update(git_tag, subdir=None, code_repo=None, deploy_key=None):
             sudo('git remote prune origin')
             # this can get into a state where running it once fails
             # but primes it to succeed the next time it runs
-            sudo('git fetch origin --tags -q || git fetch origin --tags -q')
+            sudo(f'git fetch origin {env.code_branch} --tags -q || git fetch origin {env.code_branch} --tags -q')
             sudo('git checkout {}'.format(git_tag))
             sudo('git reset --hard {}'.format(git_tag))
             sudo('git submodule sync')
-            sudo('git submodule update --init --recursive -q')
+            sudo('git submodule update --init --recursive --single-branch -q')
```
joxl, line 54: Has this been tested and confirmed to avoid fetching ancillary branches on a submodule where:

- parent repo existed (was cloned) prior to _this PR_, and
- the submodule existed (was cloned into the the parent repo) prior to _this PR_

I'm interested in this specific scenario because my understanding of `--single-branch` is that it only has effect on a `git clone` operation (which is only ever performed once per repo; naturally). In my testing, it appears that the _only_ thing that `--single-branch` does is configure the `remote.<name>.fetch` setting on a repo to be restrictive (i.e. only fetch the default remote branch) _at clone time_.

I think this means that any repos that were cloned _prior to this PR_ will continue to indiscriminately fetch _all_ branches, even after this PR is "live", which means that this PR won't help any HQ repos (parent or submodules) until they are either "recycled" (by decommission, manual deletion, etc) or a manual correction with something like:

```shell
git config remote.origin.fetch '+refs/heads/master:refs/remotes/origin/master'
git submodule foreach "git config remote.origin.fetch '+refs/heads/master:refs/remotes/origin/master'"
```

snopoke:
Closing this since deploy is now moved to Ansible. Can re-open if necessary but might be easier to create a new branch.

PR #5290, opened 2022-05-17
Ignore install-config so the installer doesn't complain we aren't on latest master

proteusvacuum:
I keep getting the following message when running the quick installer:

```
Aborting. You have unstaged changes
```

##### ENVIRONMENTS AFFECTED

None


kaapstorm (approved):

sravfeyn:
👍

PR #5291, opened 2022-05-18
Update docs to not use sudo

sravfeyn:
Not all commands need sudo. This needed to be updated which I forgot earlier.


sravfeyn:
@proteusvacuum

PR #5293, opened 2022-05-20
Correct domain migration docs

sravfeyn:
This PR updates the [migration docs](https://commcare-cloud.readthedocs.io/en/latest/installation/migration/1-migrating-project.html#import-the-data-to-the-new-environment) specifically with the understanding that an env needs to be code-deployed before migration and the DBs to be reset only if there is test data that the user wants to be deleted. There are also other updates to rebuild UCRs and to remove cleanliness_flag command (it's not needed anymore).

@proteusvacuum @Charl1996

### DIFF
--- a/hosting_docs/source/installation/migration/1-migrating-project.rst
+++ b/hosting_docs/source/installation/migration/1-migrating-project.rst
@@ -98,7 +98,7 @@ During the downtime, mobile users will still be able to collect data, but they
 will be unable to submit forms or sync with the server.
 
 
-* Block data access by turning on the ``DATA_MIGRATION`` feature flag.
+* Block data access by turning on the ``DATA_MIGRATION`` feature flag (via HQ Web UI).
 * Print information about the numbers in the database for later reference.
   This will take a while (15 mins) even on small domains. Tip: add ``--csv`` to
   the command to save the output in a csv file.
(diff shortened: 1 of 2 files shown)


hosting_docs/source/installation/migration/1-migrating-project.rst
```diff
@@ -117,25 +117,15 @@ will be unable to submit forms or sync with the server.
 3. Prepare the new environment to be populated
 ----------------------------------------------
 
+* Setup a new environment by following :ref:`deploy-commcarehq`
+* Do a commcare-hq deploy using :ref:`operations/2-deploys:Deploying CommCare HQ code changes`
+* And now proceed to step 4.
```
Charl1996, line 122: Very nit and very non blocking of nature:
```suggestion
* Proceed to step 4.
```
The fact that it's listed in the next bullet point implies the "And now" (I have a weird thing about these types of stuff)

hosting_docs/source/reference/howto/wipe_persistent_data.rst
```diff
@@ -69,42 +74,25 @@ Rebuilding environment
...
+#. Remove the "wipe_environment_enabled: True" line in your `public.yml` file.
 
-#. 
-   Run Ansible playbook to recreate databases
+#. Run Ansible playbook to recreate databases
 
    .. code-block::
 
-      $ cchq monolith ap deploy_db.yml
+      $ cchq <env_name> ap deploy_db.yml
```
Charl1996, line 83: Has this worked for you? In the past I have had to add the `--skip-check` flag (if I understand correctly, the `--skip-check` flag specifies that the command is not a dry run).

hosting_docs/source/reference/howto/wipe_persistent_data.rst
```diff
@@ -69,42 +74,25 @@ Rebuilding environment
...
-#. 
-   Recreate a superuser (where you substitute your address in place of
-   "you@your.domain").
+#. Recreate a superuser (where you substitute your address in place of
+   "you@your.domain"). (Optional)
 
    .. code-block::
 
-      $ cchq monolith django-manage make_superuser you@your.domain
+      $ cchq <env_name> django-manage make_superuser you@your.domain
```
Charl1996, line 98: I have wondered about this step, especially when doing a domain import; should this superuser be created when the plan is to do a domain import? In the past I have created a superuser, but then when I imported the data (containing a superuser with same username), it failed because the user already existed. 

I think we should keep this step here in the spirit of the "rebuilding environment" document context, but if the reader comes to this document with the intent of importing data, this step might be confusing...any thoughts?

hosting_docs/source/installation/migration/1-migrating-project.rst
sravfeyn: 👍

sravfeyn:
@Charl1996 This is updated now, thanks for the review

Charl1996 (approved):
Awesome!

PR #5294, opened 2022-05-20
Add autogenerated changelog/commands to readthedocs

sravfeyn:
This adds the target for autogenerated docs for changelog/commands to commcare-cloud.readthedocs.io 

This is open for review, it will be merged only after the [proposal](https://dimagi-dev.atlassian.net/browse/SC-2021) to migrate from Jekyll based docs to Readthedocs is approved. So this is mostly to receive feedback on the approach.

To review, you can exclude the last commit (which is the result of make).

##### ENVIRONMENTS AFFECTED
None, only documentation changes.

@dannyroberts @proteusvacuum


src/commcare_cloud/manage_commcare_cloud/commands.md.j2
```diff
@@ -1,16 +1,28 @@
...
 
-* TOC
-{:toc}
+This page explains how to run commcare-cloud commands to perform various
+actions on your environment and list of all commcare-cloud commands and their
+usage.
 
 ## Running Commands with `commcare-cloud`
 
+To run any commcare-cloud you need to install commcare-cloud first
```
proteusvacuum, line 18: "commcare-cloud _command_"

proteusvacuum (approved):

Makefile
```diff
@@ -15,16 +15,15 @@ src/commcare_cloud/help_cache/ansible-playbook.txt: export ANSIBLE_CONFIG=src/co
...
 
-
-docs/changelog/%.md : changelog/%.yml src/commcare_cloud/manage_commcare_cloud/*
+hosting_docs/source/changelog/%.md : changelog/%.yml src/commcare_cloud/manage_commcare_cloud/*
 	manage-commcare-cloud make-changelog $< > $@
 
-docs/changelog/index.md : changelog/*.yml src/commcare_cloud/manage_commcare_cloud/*
-	manage-commcare-cloud make-changelog-index > docs/changelog/index.md
+hosting_docs/source/changelog/index.md : changelog/*.yml src/commcare_cloud/manage_commcare_cloud/*
+	manage-commcare-cloud make-changelog-index > hosting_docs/source/changelog/index.md
```
dannyroberts, line 26: For these three changes, It looks like `make` is set up to now only write to the new location (`hosting_docs/source/`).

I think the PR in which we stop writing to the old location (`docs/`) from `make` should also be the PR in which we delete the docs in the old location.

My preference would be to have it write to both in this PR, and then later when we sunset the old docs location we can make `make` not write to the old location and in that same PR we can delete the files in the old location.

If you agree with that, then my suggestion here in this PR would be to keep the lines that were there previously and simply _add_ these new entries as well, making `make` write to both the old and the new locations.

src/commcare_cloud/manage_commcare_cloud/changelog-index.md.j2
```diff
@@ -3,11 +3,13 @@
 <!--See https://github.com/dimagi/commcare-cloud/blob/master/changelog/README.md for instructions-->
 # About this changelog
 
-The following changes to `commcare-cloud` that require your attention,
-newest first.
+Below are the list of changes (the newest first) to `commcare-cloud` and CommCareHQ that
+need to be applied on your environemtn to keep it upto date.
```
dannyroberts, line 7:
```suggestion
need to be applied on your environment to keep it up to date.
```

dannyroberts (commented):
This looks cool @sravfeyn. A few high level questions to help orient me mixed in with some suggestions:
- I see there's a new file `src/commcare_cloud/manage_commcare_cloud/commands.rst.j2` but I can't see how it's used. Is it used in some magical way, or is it left over from a previous iteration and can be removed?
- I see that the main thrust of this PR is to change where the autogenerated output of changelog and commands are going, but mixed in with that are formatting changes. Are the formatting changes integral to the change or could you PR them separately? I wouldn't want to hold this up because of typos in the text or hold up the text changes because we're figuring out how exactly to roll the bigger part out.

Makefile
sravfeyn: Done

sravfeyn:
@dannyroberts 
- I have removed the `src/commcare_cloud/manage_commcare_cloud/commands.rst.j2` file (it's not needed)
- I have kept the autogen target for jekyll docs too.
- I have addressed typos
- The formatting changes are very minimal, I don't think a separate PR is a must, is it?

dannyroberts:
> The formatting changes are very minimal, I don't think a separate PR is a must, is it?

Yup that's fine, that was a just a general suggestion for this or future PRs if you think it might help

Makefile
dannyroberts: Awesome, I tested this out locally and it seems to work as desired! Unrelated to your changes, while I was looking at it, I found a stray file in docs/changelog we can remove, which I did in https://github.com/dimagi/commcare-cloud/pull/5313

dannyroberts (approved):
This looks good to me now! I think you can even merge it now independent of the switchover timeline, since it now implements both methods and doesn't break the old one

sravfeyn:
Thanks Danny,

PR #5298, opened 2022-05-23
Update celery and django runner scripts to exit with their child's code

joxl:
I observed that `supervisord` logs don't reflect the actual celery or django exit codes because of these wrapper "runner" scripts. This PR fixes that.

##### ENVIRONMENTS AFFECTED
- **all**


calellowitz (approved):

joxl:
FYI, I did test this with the use of `trap` (both of these scripts use trap, although I didn't look at if it affects the final "child" service). Using `trap` to do a non-error action when a script fails does not change the code it exits with, so the change in this PR will cause the "runner" script to exit with its child's (the backgrounded and "waited on" process) code.

PR #5303, opened 2022-05-25
Revert "adding virtualenv path for upload script. It is failing with default python"

dannyroberts:
Reverts dimagi/commcare-cloud#5268


dannyroberts:
@prem-fissionhq FYI

PR #5308, opened 2022-05-26
Port travis tests to GitHub Actions

dannyroberts:
This PR ports our existing Travis tests to GitHub Actions. There's very little difference, except that I had to install virtualenv manually because it doesn't come pre-installed on the github actions VM. I checked the output to make sure that the tests are indeed running all the way through.

I also removed usages of "travis" in the code and file paths, mostly opting to use the more generic "tests" or similar instead. We've migrated a number of repos to GitHub Actions already (including commcare-hq) and my experience (and I think our experience) with it in general has been pretty positive.

This is tangentially related to the recurring frustration with the paucity of tests on commcare-cloud. Knowing we probably would not continue to invest in that knowing we are about to move to GitHub Actions, I wanted to get that out of the way so we can start diving into the more interesting problem of how to make this run a more comprehensive test. (I think giving keeping it at a reasonable run time and making it not be flaky are going to be the big challenges!)

##### ENVIRONMENTS AFFECTED
None


gherceg:
> Knowing we probably would not continue to invest in that knowing we are about to move to GitHub Actions

Could you explain why staying on Travis would be a blocker on improving the test suite? I don't follow unless there is an advantage to GitHub Actions that I'm overlooking

dannyroberts:
It's not strictly speaking a blocker. I just know we're planning to eventually move to GitHub Actions, which makes me hesitate to build more stuff out in our current build system. It's not super rational, honestly.

.tests/tests.sh
```diff
@@ -4,10 +4,14 @@ set -ve
...
 eval PYVER=$(grep python_version: $GROUP_VARS | head -n1 | cut -d' ' -f2)
-# Link Travis-installed python where ansible virtualenv will find it
+# Link GitHub Actions-installed python where ansible virtualenv will find it
 [ -z $(which python${PYVER}) ] && sudo ln -s /opt/python/${PYVER}/bin/*${PYVER}* /usr/local/bin/
 which python${PYVER}
 python${PYVER} --version
+# link virtualenv to a place where it will be found in ansible
+which /opt/pipx_bin/virtualenv && sudo ln -s /opt/pipx_bin/virtualenv /usr/local/bin/
+which virtualenv
+virtualenv --version
```
skodde, line 14: I'm confused by the use of `which` here.
In the first case a shell test seems more appropriate.
The second one appears to be useful for debugging, but the result should never vary.
Unless the idea was to perform a check similar to the previous one for `python${PYVER}`?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+        sudo apt-get update -qq
+        source control/init.sh
+        # Note: the installs must be done in the following order, otherwise the installation of commcare-cloud
+        # tries to install from the commcare-cloud-bootstrap.egg_info directory
+        pip install -e .[test]
+
+        # Python for deploy-stack in .tests/tests.sh
+        # May be removed when this file is updated to match global_vars/all.yml:python_version
+        curl -sSf --retry 5 -o python-3.9.tar.bz2 'https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/18.04/x86_64/python-3.9.tar.bz2'
+        sudo tar xjf python-3.9.tar.bz2 --directory /
```
skodde, line 44: We're currently targeting python 3.6, why does this part require 3.9?
If that's necessary, I would prefer that we install it via apt by e.g. using the deadsnakes repository like we currently do for the deployment of HQ, instead of relying on an obscure Travis binary.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.6"
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get clean
+        sudo apt-get update -qq
```
skodde, line 35: What's the purpose of these two operations? I don't see any package being installed via apt, which could benefit from a prior `update`, and `clean` is normally used to "clear out the local repository of retrieved package files", freeing up disk space after a package installation.

.tests/tests.sh
dannyroberts: Yeah the lines
```
which virtualenv
virtualenv --version
```
are just for debugging. I saw that we did that for python a few lines above and I thought it seemed useful.

.github/workflows/tests.yml
dannyroberts: The way these tests are currently run, we run our cchq `deploy-stack` command, limited to a small set that includes `localsettings.py` changes, _on the local VM_ using the ansible `local` connection in the test environment inventory. Even though `commcare-cloud` currently runs on 3.6, it requires python 3.9 to be running on the "remote" machine. In this case, the remote machine is itself, so it needs both 3.6 and 3.9.

This is copied straight from the `.travis.yml` this file is replacing.

dannyroberts: > If that's necessary, I would prefer that we install it via apt by e.g. using the deadsnakes repository like we currently do for the deployment of HQ, instead of relying on an obscure Travis binary.

This is a good suggestion. Eventually I would like to see us put the target environment in a container within this container, if that's practical, to isolate the environment. Then we could just use our ansible scripts themselves to do it without worrying about clobbering other things in the runner VM itself.

.github/workflows/tests.yml
dannyroberts: I think again this is for the later ansible run that targets `local`. If my plan of moving to a container works out I think we could get rid of this.

This too was copied from `.travis.yml`

.tests/tests.sh
skodde: Then what about the same approach?

```suggestion
# link virtualenv to a place where it will be found in ansible
[ -z $(which virtualenv) ] && sudo ln -s /opt/pipx_bin/virtualenv /usr/local/bin/
which virtualenv
virtualenv --version
```

dannyroberts: That may not work because `/opt/pipx_bin/` isn't necessarily on the $PATH the way it's set up. That's why I was using the full path to the executable.

So I can learn, what's the benefit here of `[ -z $(...)] &&` over just `... &&`? Does it have nicer behavior like suppressing the output or something, or is it just for readability i.e. easier to tell I'm using it as a condition?

skodde: In the suggested change `/opt/pipx_bin/virtualenv` is linked to `/usr/local/bin/` only if a `virtualenv` is not already present in the PATH (same as what the block before does for `python${PYVER}`).
If this is unnecessary and you only want to check for the existence of `/opt/pipx_bin/virtualenv` before linking it, then I think `[ -e /opt/pipx_bin/virtualenv ] && ...` would be preferable, because, yes, it's a more standard/readable approach and won't produce the redundant output from `which`.
But in this case is the check necessary at all? Is there a scenario in which `/opt/pipx_bin/virtualenv` wouldn't be available?

dannyroberts: I think the benefit I was trying to get out of using `which /opt/pipx_bin/virtualenv && ` is that I wanted this line to fail hard (cause tests to halt and report error) if this was not found. This could happen if we change something about the corresponding setup or if a default behavior changes out from under us over time. That would let us get the error here rather than much later when we try to use `virtualenv` but `/usr/local/bin/virtualenv` is a broken link.

I think `[ -e /opt/pipx_bin/virtualenv ] && ` will have that same behavior (failing hard if that file is not an executable). So I like the suggestion and I'll use that.

gherceg (approved):
Looks good to me. @skodde are there any remaining questions on your end?

.github/workflows/tests.yml
millerdev: As the one who added this to `.travis.yml`, I agree, it would be nice to install this in a more standard way.

millerdev (approved):
I noticed a bunch of files were moved from `.travis/...` to `.tests/...`. Were we using that hidden directory because Travis expected it? In any case, I'm wondering if it would make sense to move stuff from `.tests/` to `tests/` (or maybe a non-hidden subdirectory of `tests/`) so all test files are in the same tree?

.github/workflows/tests.yml
dannyroberts: I'm going to punt this to a followup and keep this PR focused on switching CI systems, but yeah I agree there's a clear better way to do this

dannyroberts:
I think `.travis` was a convention we used, not something pushed on us by Travis CI. I see the `.` as just moving this file to the top and marking it as a little bet "outside" of the main contents of the codebase as much as anything. I thought of this directory (now called `.tests`) as being a place to keep scripts and context related to things we only do in CI. Like running `./.tests/tests.sh` locally isn't the best idea, because it'll run portions of our ansible script against your local machine and not clean up after itself.

I'm hoping to changes this as I alluded to above to have these tests create an isolated container that it runs ansible "remotely" against, solving that problem. Once that's done, I think your suggestion of merging this directory with `/tests/` starts to make a lot of sense. I'm thinking again I will leave that to a future PR. There are also some other bits of cleaning I want to PR independently of this that include deleting some potentially unused stuff in `.tests` and deleting the long-unused `commcare-cloud-bootstrap` thing.

skodde (approved):

PR #5314, opened 2022-05-31
Update sravan's public key

sravfeyn:
My laptop had been reset recently


akashkj (approved):

snopoke:
Verified over voice that this belongs to the correct individual.

PR #5316, opened 2022-06-01
Update github actions to include testing on Python 3.10

gherceg:
Update the yml to include running tests on 3.10. Python 3.9 still needs to be installed so deploy-stack is successful, and should match the version of HQ.

More updates:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,24 +26,12 @@ jobs:
       TEST: main
 
     steps:
-    # setup 3.9 for deploy-stack
-    # may be removed when this file is updated to match global_vars/all.yml:python_version
-    - name: Setup HQ Python Version
-      uses: actions/setup-python@v3
-      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -28,16 +28,18 @@ jobs:
...
+          python-version: "3.9"
 
-    - uses: actions/checkout@v3
-    - name: Set up Python 3.x
+    - name: Set up Cloud Python Version
       uses: actions/setup-python@v3
       with:
         python-version: ${{ matrix.versions.python }}
 
+    - uses: actions/checkout@v3
```
millerdev, line 41: Is this valid?

millerdev: When I looked at this before I wasn't sure if it was valid, but I can't remember why I thought that.

gherceg:
Closing in favor of https://github.com/dimagi/commcare-cloud/pull/5379

PR #5325, opened 2022-06-06
tf_14upgrade_warnings_cleanup

pvisweswar:
**Terraform-v0.14.11-upgrade** : **tf_14upgrade_warnings_cleanup**

### DIFF
--- a/src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
+++ b/src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
@@ -10,16 +10,16 @@ terraform {
 }
 
 provider "aws" {
-  version = "~> 4.8.0"
+  #version = "~> 4.8.0"
   region  = {{ region|tojson }}
 }
 
 provider "archive" {
-  version = "~> 2.2.0"
+  #version = "~> 2.2.0"
 }
 
 provider "random" {
-  version = "~> 3.1.2"
+  #version = "~> 3.1.2"
 }
 
 // for global accelerator which can only be configured in us-west-2
--- a/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
+++ b/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
@@ -13,33 +13,33 @@ variable "tf_s3_bucket" {
 
 module "network" {
   source            = "./modules/network"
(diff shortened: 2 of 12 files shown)


src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
```diff
@@ -10,16 +10,16 @@ terraform {
...
 }
 
 provider "archive" {
-  version = "~> 2.2.0"
+  #version = "~> 2.2.0"
 }
 
 provider "random" {
-  version = "~> 3.1.2"
+  #version = "~> 3.1.2"
```
dannyroberts, line 22: I don't understand the changes in this file, but they look like they're supposed to be a temporary local change. Would it make sense to remove these changes from this PR?

dannyroberts (commented):
Looks good besides my one comment. Can you confirm that these changes are compatible with both v0.13 and v0.14?

src/commcare_cloud/commands/terraform/templates/terraform.tf.j2
pvisweswar: we have commented out the version atrribute from provider block as Terraform deprecated mentioning version in provider block,
instead it suggested to mention the provider version in terraform block directly like below.

terraform {
  required_providers {
    mycloud = {
      source  = "mycorp/mycloud"
      version = "~> 1.0"
    }
  }
}

provider "mycloud" {
  # ...
}


we haven't mentioned the version info in terraform block as part of the upgrade task, as the current configuration is ok to download the latest provider plugin as part of terraform init.

**Link:** https://www.terraform.io/language/providers/requirements

pvisweswar: Please review the below PR,  the one you were looking is old one which doesn't have python file changes.

**PR:** https://github.com/dimagi/commcare-cloud/pull/5328

dannyroberts: Oh okay. If you don't intend to merge this PR separately then do you want to just close it?

pvisweswar: Yes Danny we're not going to merge this PR. So, we can close this.

PR #5315, opened 2022-05-31
upgrading ansible to 4.2.0

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-12902
##### ENVIRONMENTS AFFECTED

All.

PR #5326, opened 2022-06-06
upgrade ansible by uninstall existing if version is 2.9

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-12902
##### ENVIRONMENTS AFFECTED

We would like to upgrade the ansible version to 4.2.0. In order to install in a clean manner, I'm uninstalling the existing ansible version under cchq env and installing the one mentioned in requirements.txt. 
please review my PR. 
@dannyroberts @millerdev 
thanks


control/init.sh
```diff
@@ -73,6 +80,13 @@ else
         COMMCARE=
         cd ${COMMCARE_CLOUD_REPO}
         pip install --quiet --upgrade pip-tools
+        # uninstall if ansible version is less than 4.0
+        ANSIBLE_VERSION=`pip show ansible | grep Version | awk '{print $2}'`
+        echo "installed version of ansible is: ${ANSIBLE_VERSION:0:3}"
+        if [[ ${ANSIBLE_VERSION:0:3} < "4.0" ]]; then
+            echo "ansible version ${ANSIBLE_VERSION} is uninstalling"
+            pip uninstall ansible --yes
+        fi
```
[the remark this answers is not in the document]
prem-fissionhq: @dannyroberts update the ticket with testing scenarios. please review and let me know. 
I have moved the commands into a function to re-use it.

dannyroberts (approved):

dannyroberts:
This change looks good (but should be coupled with the actual upgrade and the changelog change)

control/init.sh
```diff
@@ -60,11 +60,23 @@ if [ -d ${COMMCARE_CLOUD_REPO}/commcare-cloud ]; then
...
+        pip uninstall ansible --yes
+    fi
+}
+
 if [ -z "$(which manage-commcare-cloud)" ]; then
     # first time install need requirements installed in serial
     # installs strictly what's in requirements.txt, so versions are pre-pinned
     cd ${COMMCARE_CLOUD_REPO}
     pip install --upgrade pip-tools
+    # calling function to check and uninstall if ansible version < 4.0
```
millerdev, line 78: nit: This comment seems unnecessary. Same below.

prem-fissionhq:
Merging this PR: #[5315](https://github.com/dimagi/commcare-cloud/pull/5315) into this.

control/init.sh
```diff
@@ -60,11 +60,22 @@ if [ -d ${COMMCARE_CLOUD_REPO}/commcare-cloud ]; then
...
 fi
 
+function uninstall-lowerversion-ansible() {
+    ANSIBLE_VERSION=`pip show ansible | grep Version | awk '{print $2}'`
+    if [[ ${ANSIBLE_VERSION:0:3} < "4.0" ]] && [[ ! -z ${ANSIBLE_VERSION} ]]; then
+        echo "installed version of ansible is: ${ANSIBLE_VERSION:0:3}"
+        echo "ansible version ${ANSIBLE_VERSION} is uninstalling"
+        pip uninstall ansible --yes
+    fi
+}
```
dannyroberts, line 70: This should work great for updating commcare-cloud on the control machine, but it won't help people with their own dev environment. I suggest we include a similar check in commcare-cloud that tells people to upgrade ansible. (It doesn't have to block them, just alert people that to upgrade they have to first run this uninstall step.)

dannyroberts (commented):
One note that came up in our sync today

setup.py
```diff
@@ -3,8 +3,8 @@
 from setuptools import setup, find_packages
 
 install_deps = [
-    'ansible-vault==1.1.1',
-    'ansible~=2.9.7',
+    'ansible-vault==2.1.0',
+    'ansible>=2.9.7,<4.3.0',
```
dannyroberts, line 7: Since this PR forces an upgrade, I suggest changing this to say
```suggestion
    'ansible~=4.3',
```

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -106,6 +107,16 @@ def get_limit():
...
 
     def ansible_playbook(environment, playbook, *cmd_args):
+        # verifying cchq env ansible version before execution
+        required_ansible_version = environment.public_vars['ansible_version']
+        venv_ansible_version = subprocess.run(["ansible", "--version", "|", "head", "-n1", "|", "awk", "'{print $2}'"], stdout=subprocess.PIPE, text=True)
+        ansible_version = str(venv_ansible_version.stdout).split()[1]
+
+        if (version.parse(ansible_version) < version.parse(str(required_ansible_version))):
+            puts(color_error("Please uninstall current Version "+ansible_version+" and install ansible version "+required_ansible_version+"\n"
+                              "in Cchq env. Else the you ran into issues in fetching credentials from secret manager"))
```
dannyroberts, line 117: How about this as an error message/formatting:

```suggestion
            puts(color_error(f"The version of ansible you have installed ({ansible_version}) is no longer supported."))
            puts(color_notice(f"To upgrade from ansible {ansible_version} to {required_ansible_version} or above you will first have to uninstall the current version (due to an idiosyncratic issue)"))
            puts(color_code("  pip uninstall ansible"))
            puts(color_notice("before re-installing the supported version using your standard method."))
```

environments/staging/public.yml
```diff
@@ -1,6 +1,8 @@
 DATADOG_ENABLED: True
 monitor_celery_heartbeat: False
 
+ansible_version: 4.2.0
```
dannyroberts, line 4: I don't like the idea of this being a per-env setting, since it's already managed by `setup.py` and `requirements.txt`. Is this needed for anything other than the newly added error message?

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -106,6 +107,16 @@ def get_limit():
             return ()
 
     def ansible_playbook(environment, playbook, *cmd_args):
+        # verifying cchq env ansible version before execution
+        required_ansible_version = environment.public_vars['ansible_version']
```
dannyroberts, line 111: Can this just be hard-coded? Taking a step back, the only reason we are going the extra mile and adding in this warning is that due to what you might call a mistake by the ansible maintainers, the normal steps for updating ansible from version X to version Y don't work, and you have to do this additional uninstall step first.

If you just hard code `required_ansible_version` to what today (Aug 5) is the version you're trying to upgrade to that is fine, because that's the version before which this upgrade process requires this manual step.

dannyroberts (changes requested):

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -106,6 +108,24 @@ def get_limit():
...
+                ansible_version = ansible.__version__
+                if (version.parse(ansible_version) < version.parse(str(required_ansible_version))):
+                    puts(color_error(f"The version of ansible you have installed ({ansible_version}) is no longer supported."))
+                    puts(color_notice(f"To upgrade from ansible {ansible_version} to {required_ansible_version} or above you will first have to uninstall the current version (due to an idiosyncratic issue)"))
+                    puts(color_code("  pip uninstall ansible"))
+                    puts(color_notice("before re-installing the supported version using your standard method."))
+                    return 2
+            # else:
+            #     puts(color_notice("please activate the cchq virtualenv."))
+            #     return 2
```
dannyroberts, line 125: I think you meant to remove these lines

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -106,6 +108,24 @@ def get_limit():
             return ()
 
     def ansible_playbook(environment, playbook, *cmd_args):
+        # verifying cchq env ansible version before execution
+        required_ansible_version = "4.2.0"
+        #venv_ansible_version = subprocess.run(["ansible", "--version", "|", "head", "-n1", "|", "awk", "'{print $2}'"], stdout=subprocess.PIPE, universal_newlines=True)
```
dannyroberts, line 113: I think you meant to remove this line

dannyroberts (commented):
This looks great!

dannyroberts (approved):

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -106,6 +108,20 @@ def get_limit():
...
+        try:
+            if os.environ['VIRTUAL_ENV']:
+                if (version.parse(ansible_version) < version.parse(str(required_ansible_version))):
+                    puts(color_error(f"The version of ansible you have installed ({ansible_version}) is no longer supported."))
+                    puts(color_notice(f"To upgrade from ansible {ansible_version} to {required_ansible_version} or above you will first have to uninstall the current version (due to an idiosyncratic issue)"))
+                    puts(color_code("  pip uninstall ansible"))
+                    puts(color_notice("before re-installing the supported version using your standard method."))
+                    return 2
+        except KeyError:
+            puts(color_notice("please activate the cchq virtualenv."))
```
millerdev, line 123: Few issues here:

1. The subprocess call is unnecessary, use `ansible.__version__` instead.
2. The `try`/`except` block is unnecessary and will do the wrong thing if `os.environ['VIRTUAL_ENV']` evaluates to false.
3. This function should return immediately if the virtualenv is not active.

```suggestion
        if 'VIRTUAL_ENV' not in os.environ:
            puts(color_notice("please activate the cchq virtualenv."))
            return 2
        min_ansible_version = "4.2.0"
        if version.parse(ansible.__version__) < version.parse(min_ansible_version):
            puts(color_error(f"The version of ansible you have installed ({ansible.__version__}) is no longer supported."))
            puts(color_notice(f"To upgrade from ansible {ansible.__version__} to {min_ansible_version} or above you will first have to uninstall the current version (due to an idiosyncratic issue)"))
            puts(color_code("  pip uninstall ansible"))
            puts(color_notice("before re-installing the supported version using your standard method."))
            return 2
```

millerdev (changes requested):

src/commcare_cloud/commands/ansible/ansible_playbook.py
dannyroberts: I think the point of the most recent commits is that `ansible.__version__` is returning a different version number from the one we're looking for. Prem says more about this in the ticket, but I think the deal is that it's returning the "ansible-core" version number instead of the version number associated with the `ansible` pip package, which is what we're looking for. (He was using `ansible.__version__` here previously and faced this issue.)

The other points stand though

millerdev: I looked at the ticket and was not able to understand the problem with `ansible.__version__`, so for now I remain unconvinced (more details requested).

millerdev: I see that since 4.0 (ansible-core 2.10.0), `ansible.__version__` reports the core version, and the version used to install from pip is completely different. I haven't figured out the reason for that yet, and it's super weird to `pip install` one version and then get something completely different from the package's `__version__`! Indeed, the `4.x` version does not appear to be anywhere in the ansible source code in `site-packages`. Why do you do that, ansible?

However, the [ansible 2.10 porting guide](https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.10.html#porting-guide-for-v2-10-0) says the uninstall is necessary when upgrading from versions prior to 2.10, so I think we can use `2.10.0` as the minimum version string (which can be compared to `ansible.__version__`), before which we will show the uninstall message. I have updated the suggestion above.

millerdev (approved):

millerdev:
The test is failing with `please activate the cchq virtualenv.` Seems that tests do not run with a virtualenv. We could set the `VIRTUAL_ENV` environment variable in the test environment to get past that check. Not sure if that would have any other side effects, but I doubt it.

src/commcare_cloud/commands/ansible/ansible_playbook.py
dannyroberts: Thanks @millerdev this is a great find

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -106,6 +108,17 @@ def get_limit():
             return ()
 
     def ansible_playbook(environment, playbook, *cmd_args):
+        if 'VIRTUAL_ENV' not in os.environ:
+            puts(color_notice("please activate the cchq virtualenv."))
+            return 2
```
dannyroberts, line 113: Is this assertion (that we're in a virtualenv) still necessary? I think the original motivation for adding it was because we had to use `pip` to get the ansible package version; since we aren't going to do that anymore, I think this will run fine not in a vitualenv as well. Typically this codebase doesn't require being in a virtualenv.

src/commcare_cloud/commands/ansible/ansible_playbook.py
```diff
@@ -106,6 +108,17 @@ def get_limit():
...
+        min_ansible_version = "2.10.0"
+        if version.parse(ansible.__version__) < version.parse(min_ansible_version):
+            puts(color_error(f"The version of ansible you have installed ({ansible.__version__}) is no longer supported."))
+            puts(color_notice(f"To upgrade from ansible {ansible.__version__} to {min_ansible_version} or above you will first have to uninstall the current version (due to an idiosyncratic issue)"))
+            puts(color_code("  pip uninstall ansible"))
+            puts(color_notice("before re-installing the supported version using your standard method."))
```
dannyroberts, line 119:
```suggestion
            puts(color_error(f"The version of ansible-core you have installed ({ansible.__version__}) is no longer supported."))
            puts(color_notice(f"To upgrade from ansible-core {ansible.__version__} to {min_ansible_version} or above you will first have to uninstall the current version of ansible (due to an idiosyncratic issue)"))
            puts(color_code("  pip uninstall ansible"))
            puts(color_notice("before re-installing the supported version using your standard method."))
```

src/commcare_cloud/commands/ansible/ansible_playbook.py
prem-fissionhq: @dannyroberts yes, we can remove it.

millerdev (approved):

dannyroberts (approved):

dannyroberts:
Merging this + https://github.com/dimagi/commcare-cloud/pull/5380 now!

PR #5328, opened 2022-06-07
upgrade_to_terraform_v0.14

pvisweswar:
**Ticket:** https://dimagi-dev.atlassian.net/browse/SAAS-13628 

**ENVIRONMENTS AFFECTED:** Production, India and Staging


dannyroberts:
Can you confirm that these changes are compatible with both v0.13 and v0.14?

dannyroberts (approved):
Looks good, assuming the `.tf` changes are compatible with both 0.13 and 0.14

dannyroberts:
And as long as these changes are compatible with 0.13 then this can be merged. If they're compatible with 0.13 but not yet quite ready for 0.14, then I would suggest leaving `terraform_version: '0.13'` on staging until you're ready for _everyone_ to use 0.14 on staging. If it's compatible with both 0.13 and 0.14 and you're ready for everyone to use 0.14 on staging, then we can merge this as-is.

If it's not quite ready for 0.14, then I still prefer merging these changes if they work for 0.13, because then you can focus on the few remaining changes without having more and more pile up in one single PR.

pvisweswar:
Yes Danny, the .tf changes are compatible with both 0.13 and 0.14.
But, only state files are not compatible after upgrade, in case we want to revert to 0.13 for any reason then we need to make 0.13 state file version as current version in s3.

shyamkumarlchauhan (approved):

dannyroberts:
@pvisweswar is splitting this into 2 PRs

PR #5332, opened 2022-06-09
Upgrade couchdb-cluster-admin to 0.7.2

gherceg:
This makes couchdb-cluster-admin compatible with jsonobject>=2.0.0, which means we could then upgrade commcare-cloud to reference a newer version of jsonobject which will work with python 3.10.

Note I separated this into 2 commits. The first was the changes made to the requirements file after running `pip-compile --output-file=requirements.txt setup.py`. The second was just upgrading the couchdb-cluster-admin package.


joxl (approved):
Excellent. Thanks for making two commits, it made the changes much easier to review.

gherceg:
As an explanation to the `importlib-resources` netaddr dependency, that appears to depend on the version of python used. I assume commcare-cloud still only officially supports 3.6, so we should continue to generate requirements with that version of python.

dannyroberts (approved):

PR #5334, opened 2022-06-12
Bump jsonobject to 2.0.0

gherceg:
Need to upgrade to at least v1.0.0 of jsonobject which added support for python 3.10. couchdb-cluster-admin also depends on jsonobject, and has been upgraded to support v2.0.0 so this is the last step.

As far as safety of this goes, the main change to jsonobject is how [ListPropertys accept iterables now](https://github.com/dimagi/jsonobject/blob/master/CHANGES.md) (my understanding). There are lots of ListProperty usages in commcare-cloud, and it is hard to say with confidence that we do not depend on this previous BadValueError exception being raised for iterables. Welcoming any suggestions for gaining a better understanding there.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -78,7 +78,7 @@ jmespath==0.10.0
     # via
     #   boto3
     #   botocore
-jsonobject==0.9.10
+jsonobject==2.0.0
     # via
     #   commcare-cloud (setup.py)
     #   couchdb-cluster-admin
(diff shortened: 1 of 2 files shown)


millerdev (approved):
> There are lots of ListProperty usages in commcare-cloud, and it is hard to say with confidence that we do not depend on this previous BadValueError exception being raised for iterables. Welcoming any suggestions for gaining a better understanding there.

I don't think this is a thing we need to worry about. See https://github.com/dimagi/jsonobject/pull/200 for specific details about the context in which this became an issue. I don't think commcare-cloud does anything like what Django 3.1 was doing, and almost certainly does not do that expecting `BadValueError` will be raised.

PR #5340, opened 2022-06-16
Don't forget to deploy Formplayer

kaapstorm:
I noticed this step was missing from our instructions.

##### ENVIRONMENTS AFFECTED
None. Documentation only.


kaapstorm:
Tiny. Docs only.

kaapstorm:
The Formplayer deploy happens automatically when the HQ deploy succeeds (because the instructions don't explicitly "deploy commcare"). ... So this documentation change is unnecessary and misleading. Closing.

PR #5343, opened 2022-06-17
Remove report cases and forms

esoergel:
The commcare-cloud portion of https://github.com/dimagi/commcare-hq/pull/31803
This has docs changes, updates to inactive environments files, a settings change, and a changelog entry.

##### ENVIRONMENTS AFFECTED
All enviroments will see a change to their localsettings file next time `update-config` is run

```python
-ES_SETTINGS = {'case_search': {'number_of_shards': 16}, 'default': {'number_of_replicas': 1}, 'hqapps': {}, 'hqcases': {}, 'hqdomains': {}, 'hqgroups': {}, 'hqusers': {}, 'ledgers': {}, 'report_cases': {}, 'report_xforms': {}, 'smslogs': {}, 'xforms': {}}
+ES_SETTINGS = {'case_search': {'number_of_shards': 16}, 'default': {'number_of_replicas': 1}, 'hqapps': {}, 'hqcases': {}, 'hqdomains': {}, 'hqgroups': {}, 'hqusers': {}, 'ledgers': {}, 'smslogs': {}, 'xforms': {}}
```
That is, `'report_cases': {}, 'report_xforms': {}, ` is removed.  [This shouldn't have any effect on its own](https://github.com/dimagi/commcare-hq/blob/a45119ce99cbd40cc9e234f9d7beb3d406747875/corehq/ex-submodules/pillowtop/es_utils.py#L101-L105).  

What's the protocol for this type of thing?  Am I expected to manually apply this change even if it's a noop?

These three deprecated environments have their list of pillows changed:
64-test
enikshay-reference (sorta, it's commented out anyways)
development

(is it really useful for us to leave those environments in this repo?)


esoergel:
Bumping for review - don't be alarmed by the `+206 −37` - it's actually pretty small, that's mostly the changelog entry in triplicate

calellowitz:
I made a comment on the linked PR. I don't love the idea of automatic deletion of these indices, but generally am on board with the rest of this changelog.

changelog/0050-remove-report-indices.yml
```diff
@@ -0,0 +1,49 @@
...
+  officially deprecated in late 2018 in changelog 0007.
+  If enabled, these two pillows do not send data to elasticsearch except for a
+  hardcoded set of project spaces on Dimagi's production environment.
+
+  If your environment does reference these pillows or has index-specific
+  settings configured for those indices, those references will need to be
+  removed.
+
+update_steps: |
+  1. Search your environment's app_processes.yml for `ReportCaseToElasticsearchPillow` and `ReportXFormToElasticsearchPillow`.  If in there, it should look something like this:
```
calellowitz, line 26: I don't think we need this step. I think the only supported option at this point is the unified `case-pillow`

esoergel: That's what I thought to, but they would still work as far as I can tell, and we never updated the docs appropriately.  We list all the old pillows including these two, and then say:
> For all features to work, each of these ETL processors must appear at least once

esoergel:
@calellowitz any objections to me merging and announcing this?

calellowitz (approved):
Looks great

PR #5348, opened 2022-06-23
Use COMMCARE_CLOUD_DEFAULT_USERNAME also for SSH

skodde:
If set, COMMCARE_CLOUD_DEFAULT_USERNAME will be used also for SSH.

Let's avoid suggesting two different places to configure the same value, while
still respecting the local SSH configuration, if present, with a lower
precedence.

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/user_utils.py
+++ b/src/commcare_cloud/user_utils.py
@@ -5,6 +5,7 @@ import subprocess
 import six
 from clint.textui import puts
 from memoized import memoized
+from paramiko import SSHConfig
 
 from commcare_cloud.colors import color_error, color_notice
 from commcare_cloud.environment.main import get_environment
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/user_utils.py
```diff
@@ -59,28 +60,26 @@ def get_default_ssh_username(host):
     If is_guess is False, the caller should assume the user wants this value
     and should not give them a chance to change their choice of user interactively.
     """
-    for line in subprocess.check_output(['ssh', host, '-G']).decode('utf8').splitlines():
-        if line.startswith("user "):
-            return StringIsGuess(line.split()[1], is_guess=False)
+    ssh_config_file = "~/.ssh/config"
```
millerdev, line 63: I don't love this hard-coded path because it feels like the beginning of re-implementing part of `ssh`. Was the switch from ssh subprocess to paramiko strictly necessary? (I also don't love the subprocess, but in this case it feels a little less messy because it uses ssh rather than re-implementing part of it.)

skodde: I'm not too happy with it myself, but to account only for the most standard path, avoiding added complexity and a series of additional assumptions, seems like a good compromise. We could also parse another environment variable that could be used to specify a different path.

The switch was necessary because `ssh -G` would **always** return a user name, which would be the system user name if one was not matched for that host in the configuration. The advantage of the subprocess is clear, it will correctly parse the entire local ssh configuration regardless of customizations, but unfortunately I wasn't able to find a way to get the value we were thinking we were getting in the original function, which instead would always hit the return inside the conditional statement, rendering it useless for its original purpose.

dannyroberts:
I looked over this in realtime with Marco. @skodde, summarizing what we talked about here. I know of at least one example where someone (for historical reasons) has a different ssh username from their AWS SSO username (possibly both different from their laptop's user's name, not sure), so having one variable that overrides both with no way to define them differently would likely be a practical problem. You mentioned that it might be possible to change the order of precedence, so that if there's a user defined in `~/.ssh/config` then that is used, otherwise `COMMCARE_CLOUD_DEFAULT_USERNAME` is used, otherwise ssh is left to use its default. I agree that might work. I'll leave it up to you to figure out if the details work out there.

As for hard-coding the path `~/.ssh/config`, I both see Daniel's point, and think that if what we're trying to do is insert another variable into the order of precedence list of places SSH looks for username, I don't think there's another way to do that. So unless we want to give up on this change entirely, I don't think I see a way around that.

PR #5357, opened 2022-07-04
Add a migration domain for ONSE

sravfeyn:
https://dimagi-dev.atlassian.net/browse/SC-2145
This is requried for migration of a project from Prod to a self hosted env.

This is already deployed onto the proxy machine.


dannyroberts (approved):
I've offlined with you @sravfeyn about the involvement needed to actually make the https cert valid

sravfeyn:
Thanks Danny, I will work with Shyam to get this safely in.

PR #5358, opened 2022-07-05
Fix kafka download script when on default versions

sravfeyn:
This fixes the bug introduced in this PR https://github.com/dimagi/commcare-cloud/pull/5169/files#r911632281

@dannyroberts 
@prem-fissionhq

### DIFF
--- a/src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
@@ -16,6 +16,12 @@
   get_url:
     url: "http://{{ apache_mirror }}/kafka/{{ kafka_version }}/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
     dest: "/opt/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
+  when: kafka_version != '2.6.1'
+
+- name: Download Kafka 2.6.1
+  get_url:
+    url: "http://{{ apache_mirror }}/kafka/{{ kafka_version }}/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/kafka/tasks/main.yml
```diff
@@ -13,6 +13,12 @@
     state: "{{ 'present' if ansible_distribution_version == '14.04' else 'absent' }}"
 
 - name: Download Kafka
+  get_url:
+    url: "http://{{ apache_mirror }}/kafka/{{ kafka_version }}/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
+    dest: "/opt/kafka_{{ kafka_scala_version }}-{{ kafka_version }}.tgz"
+  when: kafka_version != '2.6.1'
```
dannyroberts, line 19: It seems like the relevant condition here is whether `kafka_download_sha512` is defined or not, if I'm reading this correctly that these two tasks are identical except for the `checksum: "{{ kafka_download_sha512 }}"` param. It seems like anytime `kafka_download_sha512` is defined, regardless of `kafka_version` it ought to be checked. Does that sound right to you?

sravfeyn: That's right. But the value of `kafka_download_sha512` is set to that of `2.6.1`, even though `kafka_version` is not `2.6.1` ([here](https://github.com/dimagi/commcare-cloud/pull/5169/files)). So, it would fail with checksum error for non 2.6.1 versions. So, we would also have to fix the value of `kafka_download_sha512` if we want to use conditional on `kafka_download_sha512` instead of `kafka_version`.

dannyroberts (approved):

PR #5363, opened 2022-07-07
[swiss] Adjust celery processes

skodde:
##### ENVIRONMENTS AFFECTED
Swiss


minhaminha (approved):
Thanks! What is the added `celery` after the analytics_queue btw?

skodde:
> Thanks! What is the added `celery` after the analytics_queue btw?

It's the `celery` queue that I separated from the original `repeat_record_queue,celery,export_download_queue` and moved there.

PR #5369, opened 2022-07-14
configure sentry with formplayer relase name

snopoke:
Current we create the release in Sentry when we deploy using this name but all events from formplayer are tagged with a different release name (the git commit hash). This will tie the events to the correct release.

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
+++ b/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
@@ -15,6 +15,7 @@ sentry.dsn={{ formplayer_sentry_dsn }}
 {% endif %}
 sentry.environment={{ env_monitoring_id }}
 sentry.tags.HQHost={{ host }}
+sentry.release={{ formplayer_release_name }}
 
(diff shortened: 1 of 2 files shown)


skodde (approved):

calellowitz (approved):

src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
```diff
@@ -15,6 +15,7 @@ sentry.dsn={{ formplayer_sentry_dsn }}
 {% endif %}
 sentry.environment={{ env_monitoring_id }}
 sentry.tags.HQHost={{ host }}
+sentry.release={{ formplayer_release_name }}
```
dannyroberts, line 18: I noticed this gets set to something new and arbitrary when we run `cchq <env> update-config`. Not quite sure what we can do about that, just wanted to note it 🤔

snopoke: That's unfortunate but probably not a big deal.

PR #5370, opened 2022-07-15
publish a single deploy event instead of multiple

snopoke:
This changes the deploy events from 'one per component' to 'one per deploy'
which will prevent triggering unnecessary workflows when both components
are deployed.

### DIFF
--- a/src/commcare_cloud/commands/deploy/command.py
+++ b/src/commcare_cloud/commands/deploy/command.py
@@ -13,6 +13,7 @@ from commcare_cloud.commands.deploy.commcare import deploy_commcare
 from commcare_cloud.commands.deploy.formplayer import deploy_formplayer
 from commcare_cloud.environment.main import get_environment
 from commcare_cloud.environment.paths import get_available_envs
+from commcare_cloud.events import publish_deploy_event
 from commcare_cloud.fab.utils import retrieve_cached_deploy_env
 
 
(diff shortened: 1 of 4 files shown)


src/commcare_cloud/events.py
```diff
@@ -19,7 +19,7 @@ def publish_deploy_event(name, component, environment):
     data = json.dumps({
         "event_type": name,
         "client_payload": {
-            "component": component,
+            "components": components,
```
snopoke, line 22: this is a breaking change but the 'component' portion of the payload is not currently used anywhere

FYI @kiran-tripathy to confirm

kiran-tripathy: didn't find it being used anywhere

src/commcare_cloud/commands/deploy/command.py
```diff
@@ -100,6 +101,9 @@ def run(self, args, unknown_args):
                 print(color_error("Skipping formplayer because commcare failed"))
             else:
                 rc = deploy_formplayer(environment, args)
+
+        if rc == 0:
+            publish_deploy_event("deploy_success", deploy_component, environment)
```
millerdev, line 106: The event will not be published if `commcare` deploy succeeds, but `formplayer` fails. Is that intentional?

millerdev:
I remembered that the _Deploy Process_ instructions recommend deploying commcare and formplayer with separate commands:

> 7. Use CommCare-Cloud to deploy the environment.
> ```
> # Deploy CommCareHQ (see NOTE below)
> (cloud)$ cchq <env> --control deploy commcare
> 
> # Deploy formplayer (all envs)
> (cloud)$ cchq <env> --control deploy formplayer
> ```

That of course will result in two separate events being published.

PR #5373, opened 2022-07-26
Add old to URL table map to 404 page

sravfeyn:
This will make it easy for people to find the new link

@joxl 

Screenshot 2022-07-26 at 12 48 31 PM


docs/404.md
```diff
@@ -2,4 +2,75 @@
...
+|                                                                                                                                                                                                 |                                                                                                                                                                                                                                                                                  |
+| [https://dimagi.github.io/commcare-cloud/firefighting/index](https://dimagi.github.io/commcare-cloud/firefighting/index)                                                                        | [https://commcare-cloud.readthedocs.io/en/latest/reference/firefighting/index.html](https://commcare-cloud.readthedocs.io/en/latest/reference/firefighting/index.html)                                                                                                           |
+                                        |
```
joxl, line 29: Is this row malformed? When I view the rendered version of this file the rows below here aren't in a table format.

joxl (commented):
Thanks for this @sravfeyn!

docs/404.md
sravfeyn: I have just fixed it.

sravfeyn:
I haven't explored too deeply, but if the github pages allows a site-wide javascript snippet, we could query this table for the current URL and auto redirect to the new pages without user having to do anything!

joxl:
> we could query this table for the current URL and auto redirect to ...

That would be fancy! Unless it takes only very little effort, I'm inclined to say we it's probably not _that_ important yet. Definitely good to keep on the table as something to explore if needed, though.

joxl (approved):

PR #5375, opened 2022-07-27
install phonetic plugin

Robert-Costello:
https://dimagi-dev.atlassian.net/browse/USH-1722
##### ENVIRONMENTS AFFECTED

Staging
(Production at a later date if testing on staging goes well)

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml
@@ -20,4 +20,4 @@ elasticsearch_node_name: "{{ hostname }}"
 elasticsearch_delete_old_logs: False
 elasticsearch_log_retention_days: 7
 # to test plugin on staging only
-is_staging: False
+es_plugins_phonetic_enabled: False
--- a/src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml
@@ -19,3 +19,5 @@ elasticsearch_service_name: elasticsearch
 elasticsearch_node_name: "{{ hostname }}"
 elasticsearch_delete_old_logs: False
 elasticsearch_log_retention_days: 7
+# to test plugin on staging only
+is_staging: False
--- a/src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
@@ -40,3 +40,8 @@
   become: yes
   shell: "{{ elasticsearch_home }}/bin/plugin install https://github.com/dimagi/elasticsearch-repository-swift/releases/download/swift-repository-plugin-{{ swift_plugin_version }}/repository-swift-3.0.0-es{{ swift_plugin_version }}.zip"
   when: (backup_es_swift) and (installed_plugins.stdout is defined) and ('repository-swift' not in installed_plugins.stdout) and elasticsearch_version == '2.4.6'
+
+- name: Install elasticsearch phonetic plugin
+  become: yes
+  shell: "{{ elasticsearch_home }}/bin/plugin install analysis-phonetic"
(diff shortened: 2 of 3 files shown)


src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
```diff
@@ -40,3 +40,8 @@
   become: yes
   shell: "{{ elasticsearch_home }}/bin/plugin install https://github.com/dimagi/elasticsearch-repository-swift/releases/download/swift-repository-plugin-{{ swift_plugin_version }}/repository-swift-3.0.0-es{{ swift_plugin_version }}.zip"
   when: (backup_es_swift) and (installed_plugins.stdout is defined) and ('repository-swift' not in installed_plugins.stdout) and elasticsearch_version == '2.4.6'
+
+- name: Install elasticsearch phonetic plugin
+  become: yes
+  shell: "{{ elasticsearch_home }}/bin/plugin install analysis-phonetic"
+  when: (backup_es_s3) and (installed_plugins.stdout is defined) and ('analysis-phonetic' not in installed_plugins.stdout) and elasticsearch_version == '2.4.6'
```
snopoke, line 47: this isn't related to S3 backup so no need to check that var. It could be worth adding a var for this plugin specifically so that this can be merged but only enabled on staging.

```suggestion
  when: (installed_plugins.stdout is defined) and ('analysis-phonetic' not in installed_plugins.stdout) and elasticsearch_version == '2.4.6'
```

snopoke: Also, the ES version check is redundant since this whole file is excluded if ES version ~= 2.4.6: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml#L79-L80

Robert-Costello: gotcha, I'll remove those checks. As far as the checking to only enable on staging would that be something like `and (deploy_env is staging)` ?

snopoke: You can add a var here defaulted to `false`: https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml

And then override it in https://github.com/dimagi/commcare-cloud/blob/master/environments/staging/public.yml

src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml
```diff
@@ -19,3 +19,5 @@ elasticsearch_service_name: elasticsearch
 elasticsearch_node_name: "{{ hostname }}"
 elasticsearch_delete_old_logs: False
 elasticsearch_log_retention_days: 7
+# to test plugin on staging only
+is_staging: False
```
snopoke, line 23: I think this should be renamed to `es_plugins_phonetic_enabled` or something like that.

src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
```diff
@@ -40,3 +40,8 @@
   become: yes
   shell: "{{ elasticsearch_home }}/bin/plugin install https://github.com/dimagi/elasticsearch-repository-swift/releases/download/swift-repository-plugin-{{ swift_plugin_version }}/repository-swift-3.0.0-es{{ swift_plugin_version }}.zip"
   when: (backup_es_swift) and (installed_plugins.stdout is defined) and ('repository-swift' not in installed_plugins.stdout) and elasticsearch_version == '2.4.6'
+
+- name: Install elasticsearch phonetic plugin
+  become: yes
+  shell: "{{ elasticsearch_home }}/bin/plugin install analysis-phonetic"
+  when: (installed_plugins.stdout is defined) and ('analysis-phonetic' not in installed_plugins.stdout) and is_staging
```
snopoke, line 47:
```suggestion
  when: (installed_plugins.stdout is defined) and ('analysis-phonetic' not in installed_plugins.stdout) and es_plugins_phonetic_enabled
```

snopoke (approved):

PR #5379, opened 2022-08-01
Add support for python 3.10

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-13500

First step in upgrading commcare-cloud to work with python 3.10. Once tests pass, we should be able to test this on staging.

##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,11 +9,11 @@ PIP_COMPILE = pip-compile --output-file ${REQUIREMENTS} setup.py
 ANSIBLE_ENV=COLUMNS=80
 src/commcare_cloud/help_cache/ansible.txt: export ANSIBLE_CONFIG=src/commcare_cloud/ansible/ansible.cfg
 src/commcare_cloud/help_cache/ansible.txt:
-	$(ANSIBLE_ENV) ansible -h > src/commcare_cloud/help_cache/ansible.txt
+	$(ANSIBLE_ENV) ansible -h | sed 's/options:/optional arguments:/' > src/commcare_cloud/help_cache/ansible.txt
 
 src/commcare_cloud/help_cache/ansible-playbook.txt: export ANSIBLE_CONFIG=src/commcare_cloud/ansible/ansible.cfg
 src/commcare_cloud/help_cache/ansible-playbook.txt:
-	$(ANSIBLE_ENV) ansible-playbook -h > src/commcare_cloud/help_cache/ansible-playbook.txt
+	$(ANSIBLE_ENV) ansible-playbook -h | sed 's/options:/optional arguments:/' > src/commcare_cloud/help_cache/ansible-playbook.txt
 
 hosting_docs/source/reference/1-commcare-cloud/commands.md : src/commcare_cloud/* src/commcare_cloud/*/* src/commcare_cloud/*/*/*
 	manage-commcare-cloud make-docs > hosting_docs/source/reference/1-commcare-cloud/commands.md
--- a/Makefile
+++ b/Makefile
@@ -9,11 +9,11 @@ PIP_COMPILE = pip-compile --output-file ${REQUIREMENTS} setup.py
 ANSIBLE_ENV=COLUMNS=80
 src/commcare_cloud/help_cache/ansible.txt: export ANSIBLE_CONFIG=src/commcare_cloud/ansible/ansible.cfg
 src/commcare_cloud/help_cache/ansible.txt:
-	$(ANSIBLE_ENV) ansible -h | sed 's/^options:/optional arguments:/' > src/commcare_cloud/help_cache/ansible.txt
+	$(ANSIBLE_ENV) ansible -h | sed 's/^optional arguments:/options:/' > src/commcare_cloud/help_cache/ansible.txt
 
 src/commcare_cloud/help_cache/ansible-playbook.txt: export ANSIBLE_CONFIG=src/commcare_cloud/ansible/ansible.cfg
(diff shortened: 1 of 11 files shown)


src/commcare_cloud/parse_help.py
```diff
@@ -2,19 +2,27 @@
...
-ANSIBLE_HELP_OPTIONS_PREFIX='optional arguments:'
 _HELP_CACHE = os.path.join(os.path.dirname(__file__), 'help_cache')
 _AVAILABLE_HELP_CACHES = {
     'ansible -h': os.path.join(_HELP_CACHE, 'ansible.txt'),
     'ansible-playbook -h': os.path.join(_HELP_CACHE, 'ansible-playbook.txt'),
 }
 
 
+def get_ansible_help_options_prefix():
+    """Assumes Python 3"""
```
millerdev, line 19: nit
```suggestion
    assert sys.version_info.major == 3, sys.version_info
```

gherceg:
There are two entry points in our build that depend on the generated help files from the `ansible -h` and `ansible-playbook -h` commands. As of 3.10, the output for `<ansible-command> -h` changed ever so slightly  from "optional arguments:" to "options:". This is important because we use [this syntax](https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/parse_help.py#L10) to key off when parsing help messages ([here](https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/commands/ansible/run_module.py#L66) and [here](https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/commands/ansible/ansible_playbook.py#L62))

The first entrypoint occurs when running `commcare-cloud testenv deploy-stack --skip-check --quiet --tags=py3,commcarehq`. In this case, the "cached" files are used as is from the repository which at the moment means they were generated with python 3.6 and have "optional arguments:" syntax.

The second entrypoint occurs when running the `test_autogen_docs.sh` script, which starts by running `make clean` and removing all "cached" files, and then regenerating them. This means that when running with 3.10, the files contain the new "options:" syntax.

The logic to resolve the parsing is straightforward, and could either be the current implementation which is to grab the correct syntax based on the python version, or just use a regex pattern to match against that contains both strings (leaning towards changing to this).

I think the more important factor is how to handle the git diff that results from generating these files on an env with 3.10. The change to the ansible.txt and ansible-playbook.txt files on 3.10 are flagged by the `fail_if_there_is_a_git_diff.sh` script as not being logged properly in our changelog. This would make it tricky to support multiple versions of python currently (assuming that is desirable), so we would either:
- remove these files from git and leave it up to specific environments to generate them
- continue to only target one version of python at a time

millerdev:
Could we do something to eliminate the diffs in those files? Like post-process the ansible help output to standardize lines with expected diffs before saving it in the "cache" file?

src/commcare_cloud/parse_help.py
```diff
@@ -20,7 +20,7 @@ def _get_help_text(command):
         with open(_AVAILABLE_HELP_CACHES[command], 'r', encoding='utf-8') as f:
             return f.read()
     else:
-        return subprocess.check_output(command, shell=True)
+        return subprocess.check_output(command, shell=True).decode('utf-8')
```
gherceg, line 23: I added this when playing around with removing the cache entirely as I noticed this returned a byte string previously. At the moment it looks like we never actually invoke this code since only the two ansible commands use this method, and definitely hit the cache. I think this is a safe change, but wanted to flag since it isn't directly related to any 3.10 compatibility.

setup.py
```diff
@@ -32,7 +32,7 @@
 ]
 test_deps = [
     'modernize',
-    'nose>=1.3.7',
+    'nose @ git+https://github.com/dimagi/nose.git@06dff28bbe661b9d032ce839ea0ec8e9eaf6f337',
```
dannyroberts, line 35: Looks like `nose` hasn't received an update since 2015. Is it time to remove it from our test stack here? (Not blocking this PR.)

dannyroberts (approved):

Makefile
```diff
@@ -9,11 +9,11 @@ PIP_COMPILE = pip-compile --output-file ${REQUIREMENTS} setup.py
 ANSIBLE_ENV=COLUMNS=80
 src/commcare_cloud/help_cache/ansible.txt: export ANSIBLE_CONFIG=src/commcare_cloud/ansible/ansible.cfg
 src/commcare_cloud/help_cache/ansible.txt:
-	$(ANSIBLE_ENV) ansible -h > src/commcare_cloud/help_cache/ansible.txt
+	$(ANSIBLE_ENV) ansible -h | sed 's/^options:/optional arguments:/' > src/commcare_cloud/help_cache/ansible.txt
```
millerdev, line 12: Would it be disruptive to do the opposite replacement, ending up with `options:`, since that is the format supported by newer versions like Python 3.10? That way this could be removed at some point when we drop support for older versions of Python that output `optional arguments:`.

setup.py
millerdev: This is a bigger topic, so I agree that it should not block this PR. My vote is to keep nose until pytest has a solution for https://github.com/pytest-dev/pytest/issues/3834

millerdev (approved):

Makefile
gherceg: That does make more sense. Will update shortly.

gherceg: See 88f370a999a0d720f5bf99ca35ea980a71d9fe52

millerdev (approved):

PR #5384, opened 2022-08-03
Cs/add option to delete kafka

Charl1996:
TdH has recently experienced some odd issue where they couldn't run the `wipe_kafka.yml` playbook successfully, as the script tried to delete an internal topic (`__consumer_offsets`). As a result of this, we've thought it best to also include the option to reinstall kafka when running the playbook (the _"when your hammer is too small, use a bigger hammer"_ philosophy).

(After thinking about it, is it enough to just reinstall kafka? Should I also be removing Zookeeper, since Zookeeper is the one that actually [keeps track of the list of topics](https://www.cloudkarafka.com/blog/cloudkarafka-what-is-zookeeper.html#:~:text=Zookeeper%20keeps%20track%20of%20status%20of%20the%20Kafka%20cluster%20nodes%20and%20it%20also%20keeps%20track%20of%20Kafka%20topics%2C%20partitions%20etc)?)

##### ENVIRONMENTS AFFECTED
All (I guess)

### DIFF
--- a/hosting_docs/source/reference/howto/wipe_persistent_data.rst
+++ b/hosting_docs/source/reference/howto/wipe_persistent_data.rst
@@ -62,6 +62,9 @@ in the sequence given below, so you shouldn't proceed to next steps until the pr
 
       $ cchq <env_name> ap wipe_kafka.yml
 
+   .. note::
+
+      If you encounter any issues executing the wipe_kafka.yml playbook, you can optionally append `--tags reinstall` to the command, which will reinstall Kafka rather than just clearing its topics.
 
    You can check they have been removed by confirming that the following shows
    no output:
--- a/src/commcare_cloud/ansible/wipe_kafka.yml
+++ b/src/commcare_cloud/ansible/wipe_kafka.yml
@@ -2,15 +2,15 @@
 - name: Delete all Kafka topics
   become: true
   hosts: zookeeper  # Doesn't need to run on all Kafka hosts
-  vars_prompt:
-    - name: confirm_wipe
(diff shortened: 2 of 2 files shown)


snopoke (commented):
why not just filter the topics being deleted to exclude internal ones?

Charl1996:
@snopoke 

> why not just filter the topics being deleted to exclude internal ones?

That did came to mind, but won't that potentially create issues down the line, especially considering the `__consumer_topics`? I can imaging the chaos that could ensure when especially that topic still have some residual data.

snopoke:
> @snopoke
> 
> > why not just filter the topics being deleted to exclude internal ones?
> 
> That did came to mind, but won't that potentially create issues down the line, especially considering the `__consumer_topics`? I can imaging the chaos that could ensure when especially that topic still have some residual data.

That topic is not important and does not impact our usage of Kafka so you can effectively ignore it. It is used by Kafka if you are using consumer groups.

Charl1996:
@snopoke Sorry, I was actually referring to `__consumer_offsets` (not sure where I got `__consumer_topics` from). 
Am I correct in assuming that you are referring to `__consumer_offsets` and not my (probably made-up) `__consumer_topics` when you say we're not actually using it?

snopoke:
> @snopoke Sorry, I was actually referring to `__consumer_offsets` (not sure where I got `__consumer_topics` from). Am I correct in assuming that you are referring to `__consumer_offsets` and not my (probably made-up) `__consumer_topics` when you say we're not actually using it?

Correct

Charl1996:
Closing in favour of https://github.com/dimagi/commcare-cloud/pull/5386

PR #5391, opened 2022-08-09
Create playbook to bootstrap python3.10 on deploy_stack

gherceg:
With this change, running `deploy_stack` on a machine will result in python 3.10 being used instead of 3.6. We are achieving 2 goals with this setup. 1) control machines (aka machines that run commcare-cloud) will run with 3.10 2) non-control machines will use 3.10 when executing ansible roles. 

As far as the installations in addition to python 3.10, this is caused due to an incompatibility with 3.10 in the current pip/setuptools installed on the machine. To borrow from this [SO post](https://stackoverflow.com/a/69573368):
> The problem is caused by an old version of pyparsing that has been vendored into pkg_resources, which is now part of setuptools.

Rather than take the approach recommended in that post which includes cloning the setuptools repo and updating python3.10's pip using `curl`, we instead use the current installation of python3.6/pip to upgrade pip, setuptools, and wheel at the location that python3.10 will reference. This ensures that pip will execute successfully when invoked from python3.10.

##### ENVIRONMENTS AFFECTED

None

### DIFF
--- a/src/commcare_cloud/ansible/bootstrap_python.yml
+++ b/src/commcare_cloud/ansible/bootstrap_python.yml
@@ -1,0 +1,26 @@
+---
+- hosts: all
+  gather_facts: no
+  tasks:
+
+    - name: Check for Python 3.10
+      raw: test -e /usr/bin/python3.10
+      register: check_python
+      check_mode: no
+      failed_when: false
+      changed_when: false
+
+    - name: Install Python 3.10
+      raw: add-apt-repository -y ppa:deadsnakes/ppa && apt update && apt-get -y install python3.10
+      when: check_python.rc != 0
+
+    - name: Check symlink is setup
+      raw: python3 -c "import platform; print(platform.python_version())"
+      register: check_symlink
+      check_mode: no
+      failed_when: false
+      changed_when: false
+
+    - name: Create symlink for 3.10
+      raw: ln -sf /usr/bin/python3.10 /usr/bin/python3
+      when: (check_symlink.stdout | regex_replace('[\\r\\n]+','')) is version('3.10', '<', strict=True)
--- a/src/commcare_cloud/ansible/bootstrap_python.yml
+++ b/src/commcare_cloud/ansible/bootstrap_python.yml
@@ -1,5 +1,5 @@
 ---
-- hosts: control
+- hosts: all
   gather_facts: no
   tasks:
 
--- a/src/commcare_cloud/ansible/bootstrap_python.yml
+++ b/src/commcare_cloud/ansible/bootstrap_python.yml
@@ -11,7 +11,11 @@
       changed_when: false
 
     - name: Install Python 3.10
-      raw: add-apt-repository -y ppa:deadsnakes/ppa && apt update && apt-get -y install python3.10
+      raw: add-apt-repository -y ppa:deadsnakes/ppa && apt update && apt-get -y install python3.10 python3.10-distutils python3.10-dev
+      when: check_python.rc != 0
+
+    - name: Update pip, setuptools, and wheel in 3.10 target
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/deploy_stack.yml
```diff
@@ -2,6 +2,7 @@
 # deploy-stack.yml
 
 - import_playbook: update_apt_cache.yml
+- import_playbook: bootstrap_python.yml
```
millerdev, line 5: Am I understanding correctly that this will put Python 3.10 on every machine, not just the control machine? Is that necessary? I think we only need it on the control machine where commcare-cloud/ansible runs.

gherceg: Yeah that's correct, and also correct that it is not necessary to install on every machine, only machines where commcare-cloud runs. I believe ansible does execute on all machines (which is why the raw ansible module is significant), but ansible makes an effort to maintain compatibility with a wide range of python versions so it isn't necessary to target a specific version.

With that in mind, I'm pretty sure only the control machine runs commcare-cloud, and therefore we could limit this playbook to control hosts only? I'm operating with the understanding that "control" is a standard across commcare-cloud env setups.

gherceg: Confirmed with Danny that I can assume all envs will have a "control" group, so I think bb604efd67566cf6d0abe8a02cb8b9a9977e12a8 effectively limits this playbook to running on servers within that group only.

gherceg: Ok finally got this into a working state thanks to @skodde! Marco brought up a good point as well related to this thread, which is that even though commcare-cloud does not run on the non-control machines, and ansible supports a large range of python versions, it could still be good to upgrade these non-control machines to 3.10 now and see if our regular running of ansible roles are impacted at all. This would be one less concern when upgrading to ubuntu 22.04 which at that point would force these machines onto 3.10.

src/commcare_cloud/ansible/bootstrap_python.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    - name: Check symlink is setup
+      raw: python3 -c "import platform; print(platform.python_version())"
+      register: check_symlink
+      check_mode: no
+      failed_when: false
+      changed_when: false
+
+    - name: Create symlink for 3.10
+      raw: ln -sf /usr/bin/python3.10 /usr/bin/python3
```
millerdev, line 25: Is this supported on Ubuntu? My guess is it's not.

For some OSes it is unsupported and strongly discouraged to change the system Python installation in any way because scripts that are built in to the OS may not work on the new version of Python.

Here's [one example](https://cloudbytes.dev/snippets/upgrade-python-to-latest-version-on-ubuntu-linux#step-3-set-python-310-as-default) (Ubuntu) and [here's another](https://superuser.com/a/770697/86694) (MacOS).

It's fine to install new versions of Python, but I think we will need to refer to them by either reference them via full version paths like `/usr/bin/python3.10` (best option) or by putting symlinks in places where the system will not reference them (like `/usr/local/bin`).

gherceg: That's a good point. It sounds like a mixture of using this playbook to install python 3.10 with root privileges, and then using `init.sh` so decide which version of python to use would be the ideal solution then? How does that sound? And @skodde feel free to add any thoughts you may have.

millerdev: [`ansible_python_interpreter`](https://docs.ansible.com/ansible/latest/reference_appendices/interpreter_discovery.html) can be used to tell ansible to use a non-default python on target nodes.

gherceg:
No longer taking this approach. Closing in favor of [a manual approach](https://github.com/dimagi/commcare-cloud/pull/5424).

PR #5411, opened 2022-08-22
Fix TypeError: 'NoneType' object is not iterable

millerdev:
Follow-up for https://github.com/dimagi/commcare-cloud/pull/5410


gherceg (approved):

gherceg:
Looks like builds won't pass for another hour and a half due to https://github.com/actions/runner-images/issues/6002, how do we feel about force merging this since it appears to be a blocker for using `cchq`?

PR #5415, opened 2022-08-23
Add option to use python3.10 in init.sh

gherceg:
We want to upgrade commcare-cloud to support 3.10, and this is a step needed to test cchq commands on 3.10.

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -11,7 +11,7 @@ fi
 
 
 function realpath() {
-    python3 -c "import os,sys; print(os.path.realpath(sys.argv[1]))" $1
+    python -c "import os,sys; print(os.path.realpath(sys.argv[1]))" $1
 }
 
 
--- a/control/init.sh
+++ b/control/init.sh
@@ -1,5 +1,6 @@
 #! /bin/bash
 CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq}
+CCHQ_PYTHON=$([ "$CCHQ_VIRTUALENV" == cchq ] && echo 36 || echo 310)
 VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
 
 if [[ $_ == $0 ]]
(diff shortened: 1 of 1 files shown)


control/init.sh
```diff
@@ -10,20 +11,29 @@ then
...
+          python3 -m virtualenv $VENV
+      fi
     fi
     source $VENV/bin/activate
 fi
 
+
+function realpath() {
+    python -c "import os,sys; print(os.path.realpath(sys.argv[1]))" $1
+}
```
gherceg, line 34: might have been unnecessary to move this method until after the virtualenv was activated, but wanted to ensure `python` was referencing the correct binary.

dannyroberts (approved):

control/init.sh
millerdev: This should work with any version of Python.

millerdev (approved):

gherceg:
Ended up dropping the commit that moved `function realpath()` since it was unnecessary, hence the force push.

control/init.sh
```diff
@@ -11,15 +11,19 @@ fi
 
 
 function realpath() {
-    python3 -c "import os,sys; print(os.path.realpath(sys.argv[1]))" $1
+    python -c "import os,sys; print(os.path.realpath(sys.argv[1]))" $1
```
millerdev, line 14: Was this change necessary? If a virtualenv is activated it will work, but if a virtualenv is not activated I think this will fail on any modern Ubuntu.

gherceg: I mainly am aiming for consistency, whether that is using `python` or `python3`. Maybe I'm misunderstanding how this would be used, but I assumed it would only be used when a virtualenv is activated, otherwise the operator does not have a valid setup. We use `python` [here](https://github.com/dimagi/commcare-cloud/blob/9826971801730c3b603791247260dfebb5a7ab7f/control/init.sh#L98) and [here](https://github.com/dimagi/commcare-cloud/blob/9826971801730c3b603791247260dfebb5a7ab7f/control/init.sh#L127). Maybe those are different cases than this though?

millerdev: Sure, if the virtualenv is always expected to be activated when this is called then it's fine. It's also fine from a consistency perspective. This utility may have been written and possibly copied from a place where it was expected to work outside of a virtualenv context.

Incidentally, when this was [originally added](https://github.com/dimagi/commcare-cloud/commit/50cceea14a54ba4639907adde80e39884117e69b) it was called before the virtualenv was activated. However when it was later [changed from `python` to `python3`](https://github.com/dimagi/commcare-cloud/commit/a5d69d593722e5bd55148aa60d2b33e0466c68cf) it was also called after the virtualenv was activated, so that change was not strictly necessary (and that commit was inconsistent in changing `python` to `python3`).

PR #5421, opened 2022-08-25
adding new pgauditcare db to replace with smaller disk instance

prem-fissionhq:
https://dimagi-dev.atlassian.net/browse/SAAS-13888
##### ENVIRONMENTS AFFECTED

Production 
- auditcareDB

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -525,11 +525,18 @@ rds_instances:
     backup_retention: 7
     enable_cross_region_backup: true
     params:
-      shared_preload_libraries: pg_stat_statements
+      shared_preload_libraries: pg_stat_statements,pg_transport
       log_min_duration_statement: 1000
       max_wal_size: 8GB
       wal_buffers: 32MB
       shared_buffers: 12GB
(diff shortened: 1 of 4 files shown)


shyamkumarlchauhan (approved):

environments/production/terraform.yml
```diff
@@ -516,6 +516,21 @@ rds_instances:
...
       shared_buffers: 12GB
 
+  - identifier: "pgauditcare1-production"
+    instance_type: "db.t3.2xlarge"
+    storage: 6000
+    multi_az: true
+    engine_version: "10.17"
+    backup_window: "23:00-01:00"
+    backup_retention: 7
+    enable_cross_region_backup: true
```
dannyroberts, line 526: To save some costs during the switchover, can you set this to `false` to begin with, and then the day before the switchover, set it to `true`? Unless we think the total amount of time both will be up is just a few days

prem-fissionhq: @dannyroberts I'm planning to launch the new instance before a couple of hours of the maintenance window.

dannyroberts (approved):

PR #5431, opened 2022-08-29
Use venv instead of virtualenv on 3.10

gherceg:
This should allow us to switch from `virtualenv` to `venv` as users move onto 3.10, without breaking any existing 3.6 workflows.

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -17,13 +17,14 @@ function realpath() {
 
 if [ -z ${CI_TEST} ]; then
     if [ ! -f $VENV/bin/activate ]; then
-        python_version="python3"
-        if [[ "$CCHQ_VIRTUALENV" == *"3.10"* ]]; then
-          python_version="python3.10"
+        if [[ $CCHQ_VIRTUALENV == *3.10* ]]; then
+          # use venv because 3.10 setup includes installing python3.10-venv
+          python3.10 -m venv $VENV
+        else
+          # use virtualenv because `python3 -m venv` requires python3-venv
+          python3 -m pip install --user --upgrade virtualenv
+          python3 -m virtualenv $VENV
         fi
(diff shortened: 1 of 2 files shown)


millerdev (approved):

dannyroberts (approved):

gherceg:
Just confirming with reviewers that it makes sense to update the monolith install script as well in this case. I'll be updating the documentation for the monolith install in https://github.com/dimagi/commcare-cloud/pull/5424, which will specify to set the `CCHQ_VIRTUALENV` environment variable before installing if you want to use 3.10 instead. This way in our transition period, any new self hosters have the opportunity to jump straight to 3.10, otherwise they can follow the normal steps to upgrade from 3.6.

millerdev (approved):

sravfeyn:
People who may manually activate the virtualenv need to be aware of this change, so that they start activating using venv?

How does this work for people with existing installs when they initialise commcare-cloud or update requirements? I think it should just switch to venv without issues?

gherceg:
> People who may manually activate the virtualenv need to be aware of this change, so that they start activating using venv?

Yeah my understanding is the process to activate and deactivate a virtual environment is the same between `venv` and `virtualenv`. `source <path/to/virtualenv>/bin/activate` and `deactivate`.

> How does this work for people with existing installs when they initialise commcare-cloud or update requirements? I think it should just switch to venv without issues?

Part of the instructions for upgrading to 3.10 include installing `python3.10-venv` to ensure it exists on the system. So `venv` will only become necessary once upgrading to 3.10, otherwise `virtualenv` will continued to be use on 3.6.

Also, both of these changes remain completely optional in this PR, and serve mainly as a way for us to continue testing 3.10. 

@sravfeyn

PR #5432, opened 2022-08-29
Removed NIC password settings

orangejenny:
@akashkj I'm fairly certain this is leftover from CAS, but can you confirm that India division doesn't need this setting?

##### ENVIRONMENTS AFFECTED
india


akashkj (approved):
@orangejenny Curious, what motivated this and the other cleanup PR?

orangejenny:
@akashkj I am always looking for code to delete 🙂  I think I stumbled on this while looking at the sql dump code.

PR #5446, opened 2022-09-01
Fix issue updating es conf

dannyroberts:
https://forum.dimagi.com/t/elasticsearch-installation-error/9247

The issue was caused by the recent ansible upgrade and a behavior change in ansible
between versions on how |map("extract", ...) handles undefined values

##### ENVIRONMENTS AFFECTED
all


dannyroberts:
@prem-fissionhq FYI

PR #5447, opened 2022-09-02
Elasticsearch small fixes

skodde:
https://forum.dimagi.com/t/error-when-running-commcare-cloud-monolith-django-manage-preindex-everything/9253

##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
@@ -2,5 +2,5 @@ check file elasticsearch with path {{ elasticsearch_data_dir }}/logs/{{ elastics
    group database
    group elasticsearch
    start program = "/bin/systemctl start {{ elasticsearch_service_name }}"
-   stop program = "/bin/systemctl stop {{ elasticsearch_service_name }}"
+   stop program = "/usr/bin/pkill -f {{ elasticsearch_service_name }};/bin/sleep 5;/bin/systemctl stop {{ elasticsearch_service_name }}"
    if content = "java.lang.OutOfMemoryError" then restart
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
```diff
@@ -2,5 +2,5 @@ check file elasticsearch with path {{ elasticsearch_data_dir }}/logs/{{ elastics
    group database
    group elasticsearch
    start program = "/bin/systemctl start {{ elasticsearch_service_name }}"
-   stop program = "/usr/bin/pkill -f {{ elasticsearch_service_name }};/bin/sleep 5;/bin/systemctl stop {{ elasticsearch_service_name }}"
+   stop program = "/bin/systemctl stop {{ elasticsearch_service_name }}"
```
dannyroberts, line 5: Can you confirm you saw https://github.com/dimagi/commcare-cloud/pull/3981? If so are you saying it was the wrong fix at the time or that it was the right fix at that time but something else has changed and now it's not necessary? I don't have any further context on this, I just want to make sure we're not reintroducing the problem this was meant to solve.

skodde: I don't expect any differences with this change, `pkill` will send a `SIGTERM` which is what stopping the service will use, and a 5 seconds timeout should not be relevant at all as I don't see any set in the systemd unit file.
I also see only one process that matches the provided name, which is the same managed by the service.
If there is still a problem and we need to make adjustments, we should make those in the systemd unit file for this service.

dannyroberts (approved):

dannyroberts (approved):

src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
skodde: We can continue on #5448

PR #5448, opened 2022-09-02
Elasticsearch monit remove unnecessary pkill (2)

skodde:
##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/monit.elasticsearch.j2
@@ -2,5 +2,5 @@ check file elasticsearch with path {{ elasticsearch_data_dir }}/logs/{{ elastics
    group database
    group elasticsearch
    start program = "/bin/systemctl start {{ elasticsearch_service_name }}"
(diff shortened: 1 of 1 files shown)


dannyroberts:
Poking around the other place I see something similar is https://github.com/dimagi/commcare-cloud/blob/055ef8b287513aedd01cb9d16bb1b83c3bfe50d7/src/commcare_cloud/ansible/roles/elasticsearch/tasks/rolling_restart.yml#L26-L29. I vaguely recall that at some point we had issues with elasticsearch processes not dying when we issued a normal stop/start to them, and that this was meant to deal with that. Perhaps @shyamkumarlchauhan you have a clearer memory.

skodde:
> Poking around the other place I see something similar is
> 
> https://github.com/dimagi/commcare-cloud/blob/055ef8b287513aedd01cb9d16bb1b83c3bfe50d7/src/commcare_cloud/ansible/roles/elasticsearch/tasks/rolling_restart.yml#L26-L29
> . I vaguely recall that at some point we had issues with elasticsearch processes not dying when we issued a normal stop/start to them, and that this was meant to deal with that. Perhaps @shyamkumarlchauhan you have a clearer memory.

Also here this should not be necessary, this is probably from a time before systemd or with a systemd unit file for elasticsearch not configured properly.
Do we use this task at all?

PR #5455, opened 2022-09-06
install phonetic analysis plugin on prod

Robert-Costello:
https://dimagi-dev.atlassian.net/browse/USH-1700
##### ENVIRONMENTS AFFECTED

Production

### DIFF
--- a/environments/staging/public.yml
+++ b/environments/staging/public.yml
@@ -198,6 +198,3 @@ use_chrony: True
 
 python_version: '3.9'
 use_deadsnakes: '{{ ansible_distribution_version <= "18.04" }}'
-
-# to test plugin on staging only
-es_plugins_phonetic_enabled: True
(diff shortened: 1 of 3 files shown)


esoergel (approved):
This will install it on all environments next time they deploy ES, right?  I think that's fine.  Do we need to advise any manual steps to prevent that from breaking something?  My understanding is that it should just install the plugin, but it won't actually do anything until they migrate the index over.

Robert-Costello:
>it should just install the plugin, but it won't actually do anything until they migrate the index over.

Yes, it should only install the plugin with no side effects or changes to functionality until settings and mapping are changed.

snopoke:
> > it should just install the plugin, but it won't actually do anything until they migrate the index over.
> 
> Yes, it should only install the plugin with no side effects or changes to functionality until settings and mapping are changed.

Also, just to confirm that it won't restart ES either?

Robert-Costello:
That's right it won't restart ES. The docs instruct you to restart each node after installation. I had to do that manually to make the plugin available locally and on staging we had run the command to deploy es to get a restart.

PR #5458, opened 2022-09-07
Use python from cchq venv for backup scripts

skodde:
##### ENVIRONMENTS AFFECTED
All

### DIFF
--- a/src/commcare_cloud/ansible/roles/backups/tasks/set_up_check_s3_backups.yml
+++ b/src/commcare_cloud/ansible/roles/backups/tasks/set_up_check_s3_backups.yml
@@ -12,7 +12,7 @@
   become: yes
   cron:
     name: "Check {{ service }} s3 backups"
-    job: "/usr/local/bin/check_s3_backup.py {{ service }} > {{ service_backup_dir }}/s3_backup_status.txt || { cat {{ service_backup_dir }}/s3_backup_status.txt | mail -s 'Recent {{ service }} backups missing' {{ check_s3_backups_email }}; }"
+    job: "{{ virtualenv_home }}/bin/python3 /usr/local/bin/check_s3_backup.py {{ service }} > {{ service_backup_dir }}/s3_backup_status.txt || { cat {{ service_backup_dir }}/s3_backup_status.txt | mail -s 'Recent {{ service }} backups missing' {{ check_s3_backups_email }}; }"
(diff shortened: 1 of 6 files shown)


shyamkumarlchauhan (approved):

src/commcare_cloud/ansible/roles/backups/tasks/set_up_check_s3_backups.yml
```diff
@@ -12,7 +12,7 @@
   become: yes
   cron:
     name: "Check {{ service }} s3 backups"
-    job: "/usr/local/bin/check_s3_backup.py {{ service }} > {{ service_backup_dir }}/s3_backup_status.txt || { cat {{ service_backup_dir }}/s3_backup_status.txt | mail -s 'Recent {{ service }} backups missing' {{ check_s3_backups_email }}; }"
+    job: "{{ virtualenv_home }}/bin/python3 /usr/local/bin/check_s3_backup.py {{ service }} > {{ service_backup_dir }}/s3_backup_status.txt || { cat {{ service_backup_dir }}/s3_backup_status.txt | mail -s 'Recent {{ service }} backups missing' {{ check_s3_backups_email }}; }"
```
millerdev, line 15: What is this script doing that it needs to use the commcare-hq virtualenv? Is the change to the shebang line in the scripts not good enough? Why change both the interpreter used here as well as the shebang line? Seems like only one or the other should be necessary, and if the interpreter must be specified here, why have a shebang line in the script at all?

In case not clear, I think it would be best not to specify the Python interpreter here, and instead use the one specified in the shebang line of the script. Also, I don't think we want these backup scripts to be tied to the set of dependencies installed by HQ since those may change without warning. If these scripts need some extra dependencies that are not available with default with the system Python then we should probably install those dependencies in a virtualenv that is specifically setup for them.

skodde: The shebang change is to follow best practices and to make explicit that these scripts now require python3 and are incompatible with python2.

They require boto3, which is in the base requirements for the commcare-hq virtualenv. I don't expect this to change anytime soon.
Specifying the interpreter allows us to avoid installing this dependency separately, either with a system package or by setting up another virtualenv, both of which would be a separate maintenance burden and cause potential hurdles during transitions to newer versions.

I think we should just consider these scripts part of HQ and maintain them and their dependencies accordingly.

PR #5475, opened 2022-09-14
Add missing pyOpenSSL dependency

skodde:
Add missing pyOpenSSL dependency (22.x), as we've been relying on a system
package that gets installed indirectly, and bump cryptography to 38.x.

The current system provided version of pyOpenSSL (python3-openssl) is no
longer compatible with the authorized_key ansible module.

##### ENVIRONMENTS AFFECTED
All


gherceg:
This is a huge version jump in `cryptography`. Just curious what the plans to test that out are?

millerdev (approved):

PR #5476, opened 2022-09-14
Replace lowercase with uppercase environment variable PROMETHEUS_MULTIPROC_DIR

gherceg:
Followup to https://github.com/dimagi/commcare-hq/pull/31684

This includes a new changelog entry to detail instructions for applying this change.


changelog/0056-update-prometheus-variable.yml
```diff
@@ -0,0 +1,15 @@
+title: Update Prometheus variable
+key: update-prometheus-variable
+date: 2022-09-14
+optional_per_env: no
```
snopoke, line 4: caveat here is that it is only needed if you are using 'prometheus'

millerdev (approved):

PR #5478, opened 2022-09-15
Create and use 3.10 venv if 3.10 is installed

gherceg:
This is a minor update that I was initially going to make on https://github.com/dimagi/commcare-cloud/pull/5424, but it felt more fitting to keep it separate as it leaves 3.10 optional (easier for internal testing).

This will make it so that when we do upgrade to 22.04, because 3.10 will already exist, the virtualenvs will be named 'cchq', rather than 'cchq-3.10' which is minor, but convenient.

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -2,7 +2,7 @@
 CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq}
 VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
 NO_INPUT=0
-USE_SYSTEM_PYTHON=${USE_SYSTEM_PYTHON:-false}
+BIONIC_USE_SYSTEM_PYTHON=${BIONIC_USE_SYSTEM_PYTHON:-false}
 
 if [[ $_ == $0 ]]
 then
@@ -18,14 +18,14 @@ function realpath() {
 
 
 if [ -z ${CI_TEST} ]; then
-    if [[ $USE_SYSTEM_PYTHON == true ]] && hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
-        # if on 18.04 with 3.10 installed, use cchq-3.10 unless $USE_SYSTEM_PYTHON is true
+    if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
+        # if on 18.04 with 3.10 installed, use cchq-3.10 unless $BIONIC_USE_SYSTEM_PYTHON is true
         CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
         VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
     fi
     # check if a virtualenv at $VENV exists yet, and create if not
     if [[ ! -f $VENV/bin/activate ]]; then
-        if hash python3.10 2>/dev/null; then
+        if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null; then
             echo "Creating a python3.10 virtual environment named ${CCHQ_VIRTUALENV}"
             # use venv because 3.10 setup includes installing python3.10-venv
             python3.10 -m venv $VENV
--- a/control/init.sh
+++ b/control/init.sh
@@ -17,7 +17,7 @@ function realpath() {
 
 
 if [ -z ${CI_TEST} ]; then
-    if [[ $(cat /etc/os-release | grep '^VERSION=' | tr -dc '0-9.') == *18.04* && hash python3.10 2>/dev/null ]]; then
+    if [[ $(cat /etc/os-release | grep '^VERSION=' | tr -dc '0-9.') == *18.04* ]] && hash python3.10 2>/dev/null; then
         # if on 18.04 and 3.10 is installed, figure out which name to use for a 3.10 virtualenv
         if [[ -f $VENV/bin/python3.6 ]]; then
             # if a 3.6 environment with the $CCHQ_VIRTUALENV name already exists, append '-3.10'
--- a/control/init.sh
+++ b/control/init.sh
@@ -17,8 +17,8 @@ function realpath() {
 
 
 if [ -z ${CI_TEST} ]; then
-    if hash python3.10 2>/dev/null; then
-        # figure out which name to use for a 3.10 virtualenv, defaulting to $CCHQ_VIRTUALENV
+    if [[ $(cat /etc/os-release | grep '^VERSION=' | tr -dc '0-9.') == *18.04* && hash python3.10 2>/dev/null ]]; then
+        # if on 18.04 and 3.10 is installed, figure out which name to use for a 3.10 virtualenv
         if [[ -f $VENV/bin/python3.6 ]]; then
             # if a 3.6 environment with the $CCHQ_VIRTUALENV name already exists, append '-3.10'
             CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
--- a/control/init.sh
+++ b/control/init.sh
@@ -23,6 +23,7 @@ if [ -z ${CI_TEST} ]; then
             if [[ -f $VENV/bin/python3.6 ]]; then
                 # if a 3.6 environment with the $CCHQ_VIRTUALENV name already exists, append '-3.10'
                 CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
+                VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
             fi
             # check if a virtualenv at $VENV exists yet, and create if not
             if [[ ! -f $VENV/bin/activate ]]; then
--- a/control/init.sh
+++ b/control/init.sh
@@ -17,13 +17,10 @@ function realpath() {
 
 
 if [ -z ${CI_TEST} ]; then
-    if [[ $(cat /etc/os-release | grep '^VERSION=' | tr -dc '0-9.') == *18.04* ]] && hash python3.10 2>/dev/null; then
-        # if on 18.04 and 3.10 is installed, figure out which name to use for a 3.10 virtualenv
-        if [[ -f $VENV/bin/python3.6 ]]; then
-            # if a 3.6 environment with the $CCHQ_VIRTUALENV name already exists, append '-3.10'
-            CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
-            VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
-        fi
+    if hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
+        # if on 18.04 with 3.10 installed, use cchq-3.10
+        CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
+        VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
     fi
     # check if a virtualenv at $VENV exists yet, and create if not
     if [[ ! -f $VENV/bin/activate ]]; then
--- a/control/init.sh
+++ b/control/init.sh
@@ -38,6 +38,11 @@ if [ -z ${CI_TEST} ]; then
     source $VENV/bin/activate
 fi
 
+if [[ $BIONIC_USE_SYSTEM_PYTHON == true ]]; then
+    echo "BIONIC_USE_SYSTEM_PYTHON should only be used temporarily when it is necessary to use Python 3.6 on 18.04."
+    echo "Please remove this variable from your environment when you are able to use Python 3.10 again."
+fi
+
 if [ -n "${BASH_SOURCE[0]}" ] && [ -z "${BASH_SOURCE[0]##*init.sh*}" ]
 then
     # this script is being run from a file on disk, presumably from within commcare-cloud repo
--- a/control/init.sh
+++ b/control/init.sh
@@ -2,6 +2,7 @@
 CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq}
 VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
 NO_INPUT=0
+USE_SYSTEM_PYTHON=${USE_SYSTEM_PYTHON:-false}
 
 if [[ $_ == $0 ]]
 then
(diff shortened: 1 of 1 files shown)


control/init.sh
```diff
@@ -17,10 +17,20 @@ function realpath() {
 
 
 if [ -z ${CI_TEST} ]; then
-    if [ ! -f $VENV/bin/activate ]; then
-        if [[ $CCHQ_VIRTUALENV == *3.10* ]]; then
-          # use venv because 3.10 setup includes installing python3.10-venv
-          python3.10 -m venv $VENV
+    if [[ ! -f $VENV/bin/activate ]]; then
```
millerdev, line 20: Will this always use the existing (3.6) `cchq` virtualenv if it exists?

millerdev (commented):
> when we do upgrade to 22.04, because 3.10 will already exist

I could be wrong, but I don't think we're going to be "upgrading" many (any?) machines from 18.04 to 22.04. In most cases we will create a new 22.04 control machine and then decommission the old 18.04 control machine after it is setup.

millerdev:
Instead of using a different name if an existing `cchq` virtualenv already exists, what do you think of always using the name `cchq-3.10` on 18.04 hosts? On other hosts we will use the system Python, and the virtualenv will be `cchq`.

control/init.sh
```diff
@@ -17,10 +17,20 @@ function realpath() {
...
+        if [[ -f /usr/bin/python3.10 ]]; then
```
joxl, line 21: I assume the actual installed location of the `python3.10` binary isn't important? If it must be exactly `/usr/bin/python3.10` then it would be good to document why that is. Otherwise, using `hash` to check for binaries on PATH is clearer that _"any Python 3.10 binary will do"_.

```suggestion
        if hash python3.10 2>/dev/null; then
```

gherceg:
> I could be wrong, but I don't think we're going to be "upgrading" many (any?) machines from 18.04 to 22.04. In most cases we will create a new 22.04 control machine and then decommission the old 18.04 control machine after it is setup.

Poor choice of words on my part, but yes by upgrade I meant create new machines. The point was that the virtualenv would be named `cchq`, not `cchq-3.10` on 22.04 machines (or new 18.04 machines with 3.10 installed though I know you addressed this in another comment).

> Instead of using a different name if an existing cchq virtualenv already exists, what do you think of always using the name cchq-3.10 on 18.04 hosts? On other hosts we will use the system Python, and the virtualenv will be cchq.

I think that behavior basically exists at the moment with the exception of new 18.04 machines right? Are you suggesting changing the logic to check the ubuntu version first, which in the case of 22.04 wouldn't need to do anything, but with 18.04 would need the additional logic to check which version of python to use?

control/init.sh
gherceg: Oh good catch. I think the `if [[ ! -f $VENV/bin/activate ]]; then` check is completely unnecessary here as the script hasn't even decided which venv to use yet. Will update that!

control/init.sh
gherceg: Yeah I wasn't thinking of cases where it would be installed elsewhere, but agreed that isn't important so this is a better check. Thanks!

control/init.sh
gherceg: See 42925c49685db2325c02596168cb1649b50f9c78

millerdev:
> Are you suggesting changing the logic to check the ubuntu version first, which in the case of 22.04 wouldn't need to do anything, but with 18.04 would need the additional logic to check which version of python to use?

Correct.

control/init.sh
joxl: > I wasn't thinking of cases where it would be installed elsewhere

:+1: Yep, this is why I asked. Some further elaborations if you care to understand the perspective I was coming from:

If this script is worried about the presence of "unexpected" python3.10 installations, then it (this script) would need to both:

1. Check for the existence of the version it cares about (presumably the source line of this thread), and
2. explicitly use the expected version everywhere it calls `python3.10 ...`.

Alternatively, it (what you've done here) is a much cleaner "separation of interests" for this script to "not care" which python3.10 it uses, and just assume that another "machine setup scope" has taken care to ensure the installed version of python3.10 is "safe to use without further consideration".

gherceg:
@millerdev see 518907e6a983ef941f1d6e0746523bfb66d1d167. Notably can't refactor the logic to actually create the virtualenvs just yet because 18.04 relies on `virtualenv`, whereas we are moving to `venv` with 22.04.

control/init.sh
```diff
@@ -17,24 +17,24 @@ function realpath() {
 
 
 if [ -z ${CI_TEST} ]; then
+    if [[ $(cat /etc/os-release | grep '^VERSION=' | tr -dc '0-9.') == *18.04* ]] && hash python3.10 2>/dev/null; then
+        # if on 18.04 and 3.10 is installed, figure out which name to use for a 3.10 virtualenv
+        if [[ -f $VENV/bin/python3.6 ]]; then
+            # if a 3.6 environment with the $CCHQ_VIRTUALENV name already exists, append '-3.10'
+            CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
+            VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
+        fi
```
millerdev, line 26: What do you think of always using `cchq-3.10` on 18.04? It simplifies the logic.
```suggestion
        # if on 18.04 and 3.10 is installed, cchq-3.10 is the default virtualenv name
        if [[ "$CCHQ_VIRTUALENV" == cchq ]]; then
            CCHQ_VIRTUALENV=cchq-3.10
            VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
        fi
```

control/init.sh
```diff
@@ -17,14 +17,24 @@ function realpath() {
 
 
 if [ -z ${CI_TEST} ]; then
-    if [ ! -f $VENV/bin/activate ]; then
-        if [[ $CCHQ_VIRTUALENV == *3.10* ]]; then
-          # use venv because 3.10 setup includes installing python3.10-venv
-          python3.10 -m venv $VENV
+    if [[ $(cat /etc/os-release | grep '^VERSION=' | tr -dc '0-9.') == *18.04* ]] && hash python3.10 2>/dev/null; then
```
joxl, line 20: I think `VERSION_ID` is the one you want. The fact that this file is meant to be a shell-sourceable makes grepping it annoying (you have to contend with quote characters that may or may not be present). Perhaps just source it in a subshell?

```suggestion
    if hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
```

gherceg: See 38a57d19c00901b8085204834bff68134cf46e0f

control/init.sh
gherceg: See 38a57d19c00901b8085204834bff68134cf46e0f

gherceg:
So the big con of this is that once 3.10 is installed, there is no way to specify to use the 3.6 environment if issues are faced. The only way to enable that would be exporting an environment variable. This can't be done with the `cchq` command because the environment variables there do not take effect until commcare-cloud is run, whereas the init script is run before that.

I think adding an environment variable that when set skips the logic to update the venv to path to 3.10 is the best approach. I know Daniel has advised against making this public knowledge in documentation since an environment variable is not something we can cleanup with commcare-cloud on a user's system, meaning it'll be up to them to unset it when the time is right.

control/init.sh
```diff
@@ -17,14 +17,21 @@ function realpath() {
...
-          # use venv because 3.10 setup includes installing python3.10-venv
-          python3.10 -m venv $VENV
+    if hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
+        # if on 18.04 with 3.10 installed, use cchq-3.10
+        CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
+        VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
+    fi
+    # check if a virtualenv at $VENV exists yet, and create if not
+    if [[ ! -f $VENV/bin/activate ]]; then
+        if hash python3.10 2>/dev/null; then
```
millerdev, line 27: Replace the first line of this file with:
```py
if hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
    # if on 18.04 with 3.10 installed, use cchq-3.10
    CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq-3.10}
else
    CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq}
fi
```

And apply this:
```suggestion
    # check if a virtualenv at $VENV exists yet, and create if not
    if [[ ! -f $VENV/bin/activate ]]; then
        if [[ "$CCHQ_VIRTUALENV" =~ -3\.10$ ]]; then
```

This way arbitrary custom 3.10 environments can be created by using a suffix of `-3.10` on the name. Virtualenv names not ending with `-3.10` will continue to use the default system python3.

millerdev (commented):
> So the big con of this is that once 3.10 is installed, there is no way to specify to use the 3.6 environment if issues are faced.

See comment below.

control/init.sh
gherceg: Hmm I am surprised we have to activate the virtualenv there. I was under the impression that the init script would be run after logging into the machine which would activate the script. Though this seems to have stemmed from https://github.com/dimagi/commcare-cloud/commit/4725f8130a4e78882075ff3b2a6be1482f9f6c43 which aimed to optimize this. @dannyroberts what do you think of partially reverting this by running `source ~/init-ansible` instead of `source ~/.virtualenvs/${{CCHQ_VIRTUALENV:-cchq}}/bin/activate` to allow better control over the virtualenv used?

millerdev: One reason for virtualenv activation vs running the init script is to skip the things in the init script that take a long time (updating pip and ansible galaxy requirements, etc.). It's a performance optimization that is very valuable when all you want to do is ssh to a machine, for example.

gherceg: So alternatively, we could just rely on an environment variable being set on the control machine right? I know it isn't the best approach because we lose control, but we could log a message to the user saying they are overriding commcare-cloud's default behavior and that they should remove the variable (assuming we haven't encouraged custom virtualenv names in the past).

gherceg:
I added a variable `USE_SYSTEM_PYTHON` that defaults to false, but can be set to true in the user's environment if they wish to revert back to 3.6 on 18.04. This is merely a safety mechanism in the event that something goes wrong once users upgrade to 3.10.

My intention is that once this is merged, staging will be setup to use 3.10 for all users by default.

control/init.sh
```diff
@@ -17,8 +18,8 @@ function realpath() {
 
 
 if [ -z ${CI_TEST} ]; then
-    if hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
-        # if on 18.04 with 3.10 installed, use cchq-3.10
+    if ! $USE_SYSTEM_PYTHON && hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
```
millerdev, line 21: This will evaluate the contents of this variable as a command, right? I haven't seen that technique used before, and not sure if it's a security risk? In any case, seems like it would be more conventional to use string comparison
```suggestion
    if [[ $USE_SYSTEM_PYTHON == true ]] && hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
```

millerdev (commented):
Is there a way to pass `USE_SYSTEM_PYTHON` to the control machine when executing a command with the form `cchq --control ...`?

gherceg:
> Is there a way to pass USE_SYSTEM_PYTHON to the control machine when executing a command with the form cchq --control ...?

My understanding is no, because the init script on the control machine runs prior to any commcare-cloud code, and therefore cannot make use of the environment variables passed via that command. So to make use of this variable, the user would have to add it to their profile ahead of time so that it is set when the init script is executed.

control/init.sh
gherceg: That's a good point. Thanks for flagging.

control/init.sh
```diff
@@ -17,14 +18,21 @@ function realpath() {
...
-          # use venv because 3.10 setup includes installing python3.10-venv
-          python3.10 -m venv $VENV
+    if [[ $USE_SYSTEM_PYTHON == true ]] && hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
+        # if on 18.04 with 3.10 installed, use cchq-3.10 unless $USE_SYSTEM_PYTHON is true
+        CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
+        VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
+    fi
+    # check if a virtualenv at $VENV exists yet, and create if not
+    if [[ ! -f $VENV/bin/activate ]]; then
+        if hash python3.10 2>/dev/null; then
```
millerdev, line 28: This will still use Python 3.10, even if `$USE_SYSTEM_PYTHON == true`, and will create a 3.10 env named `cchq` on Ubuntu 18.04 if python 3.10 is installed and a 3.6 virtualenv with that name does not exist. I think that would be a confusing state to get into.

control/init.sh
```diff
@@ -2,6 +2,7 @@
 CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq}
 VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
 NO_INPUT=0
+USE_SYSTEM_PYTHON=${USE_SYSTEM_PYTHON:-false}
```
millerdev, line 5: The meaning of this will be wrong on Ubuntu 22.04, right? Maybe it should be renamed to `BIONIC_USE_SYSTEM_PYTHON`?

millerdev (commented):
> > Is there a way to pass USE_SYSTEM_PYTHON to the control machine when executing a command with the form cchq --control ...?
>
> My understanding is no...

I think the steps to revert to using the system Python are going to be less than ideal if we need to do that. Additionally, we'll need to either (indefinitely) ignore that variable in the future or ensure that anyone who ever sets it later removes it from wherever they set it (or risk confusing and probably unsupported run modes).

gherceg:
> I think the steps to revert to using the system Python are going to be less than ideal if we need to do that

At this point I'm at a loss for other options. I know that there are cons to having users export a variable into their environment outside of the scope of commcare-cloud's control, but it also will only be used in the event that someone runs into issues, and commcare-cloud can incessantly bug a user if it sees that the variable is set to the non-default value. Maybe I'm undervaluing how much of a potential headache this could cause?

gherceg:
Daniel and I offlined and the decision was to continue on this path of using an environment variable, but carry the awareness that this solution could result in user's getting into states that are out of our control. To prevent this from happening, I think we need to have clear instructions communicated to users who intend to revert to 3.6, and only share these instructions if absolutely necessary. The fewer users that make use of this, the easier it till be to transition.

Additionally, everyone should be on 3.10 prior to the ubuntu upgrade, so we should be able to remove this variable and ensure no users are relying on it prior to any additional upgrades.

millerdev (approved):

gherceg:
Made a very minor text change to the user message after Daniel's approval, so I feel like it is safe to still merge.

dannyroberts (approved):

control/init.sh
gherceg: See d7b796ffecf8df3ef2b5635456424a572a1c79fe

control/init.sh
gherceg: See d7b796ffecf8df3ef2b5635456424a572a1c79fe

PR #5479, opened 2022-09-15
Remove argparse14.py for 2.7 compatibility

gherceg:
My understanding is that this is no longer needed. I've updated the imports that referenced this special file, replacing `argparse14` with `argparse`.


gherceg:
I'm realizing there are some diffs outside of the Optional Arguments -> Options change. I've added those diffs in the most recent commit so reviewers can see, but I'm not sure it is worth trying to address those the same way we addressed the Optional Arguments diff. I know we floated the idea of only running these tests on one version of python which might be necessary now?

gherceg:
Just making a note, will be able to pick this up in a few weeks once 3.6 support is dropped.

gherceg:
Rebased on master to include the update to no longer run tests on python 3.6.

millerdev (approved):

PR #5482, opened 2022-09-16
Adding_3_proxy_servers_in_terraform.yml_and_inventory.ini.j2_files

pvisweswar:
**Ticket:** https://dimagi-dev.atlassian.net/browse/SAAS-13945

**ENVIRONMENTS AFFECTED:** Production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -367,7 +367,7 @@ proxy_servers:
     group: "proxy"
     os: ubuntu_pro_bionic
 
-  - server_name: "proxy_a001-production"
+  - server_name: "proxy_a000-production"
     server_instance_type: "t3.large"
     network_tier: "app-private"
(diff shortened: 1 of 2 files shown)


environments/production/terraform.yml
```diff
@@ -367,6 +367,33 @@ proxy_servers:
     group: "proxy"
     os: ubuntu_pro_bionic
 
+  - server_name: "proxy_a001-production"
```
dannyroberts, line 370: I think our naming convention would suggest that we use either `proxy_a000-production` (adopt the naming we use for machines that use `count` in case we add that) or `proxy_a2-production` (increment 1 from the machines we're replacing). If you have another convention in mind this isn't a big deal, but I thought I would share.

shyamkumarlchauhan:
raise another PR #5488

PR #5495, opened 2022-09-26
install phonetic plugin for elasticsearch

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/USH-1700

##### ENVIRONMENTS AFFECTED

staging
swiss
india
production

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
@@ -44,4 +44,4 @@
 - name: Install elasticsearch phonetic plugin
   become: yes
   shell: "{{ elasticsearch_home }}/bin/plugin install analysis-phonetic"
-  when: (installed_plugins.stdout is defined) and ('analysis-phonetic' not in installed_plugins.stdout)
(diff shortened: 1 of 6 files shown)


src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml
```diff
@@ -44,4 +44,4 @@
 - name: Install elasticsearch phonetic plugin
   become: yes
   shell: "{{ elasticsearch_home }}/bin/plugin install analysis-phonetic"
-  when: (installed_plugins.stdout is defined) and ('analysis-phonetic' not in installed_plugins.stdout)
+  when: (installed_plugins.stdout is defined) and ('analysis-phonetic' not in installed_plugins.stdout) and es_plugins_phonetic_enabled
```
Robert-Costello, line 47: @shyamkumarlchauhan Thanks for looking into this. It was my understanding that the changes [here](https://github.com/dimagi/commcare-cloud/blob/4378ed208ee030e7153bd07da002df8eda800afd/src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v2.yml#L47), where I had removed the var that was limiting the install to staging, would install the plugin on all environments. Was that not correct?

Robert-Costello (approved):

PR #5500, opened 2022-09-26
add user emapson and public key

Robert-Costello:
Adds emapson, new USH developer, user and public ssh key.


Robert-Costello:
@nospame FYI this is the PR for adding your user/public key in case you need to do the same for someone in the future.

stephherbers (approved):

PR #5520, opened 2022-10-12
Removed CLOUDCARE_BASE_URL

orangejenny:
See https://github.com/dimagi/commcare-hq/pull/32244

~~Marking don't merge until https://github.com/dimagi/commcare-hq/pull/32244 is approved by product.~~ Approved.

This isn't directly dependent on the other PR since the feature in question is already 500ing.

##### ENVIRONMENTS AFFECTED
All, but it's just removing dead code.


dannyroberts (approved):

orangejenny:
@dannyroberts Do I need to deploy this?

dannyroberts:
Offlined, but given that this just cleanup that doesn't need to be coordinated with anything (no repercussions to it being applied or not being applied), you can just let this go out with the next change.

PR #5524, opened 2022-10-18
terraform update to launch servers

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-13980

##### ENVIRONMENTS AFFECTED
production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -42,6 +42,16 @@ servers:
     count: 6
     server_auto_recovery: true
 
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):
What do you think about using `c6a.2xlarge` instead of `c5.2xlarge`? I think it should be similar performance for about 10% less

shyamkumarlchauhan:
> 

Updated the instance type to be `c6a.2xlarge`

PR #5527, opened 2022-10-18
Save 3.10 venv to a new var

gherceg:
Fix venv naming issue. See https://github.com/dimagi/commcare-cloud/pull/5526 for more context.

PR #5529, opened 2022-10-18
Fix 3.10 venv var typo

skodde:
Fix typo in #5527.


control/init.sh
```diff
@@ -1,6 +1,6 @@
 #! /bin/bash
 CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq}
-CCHQ_VIRTUALENV_310=$CCHQ_CCHQ_VIRTUALENV-3.10
+CCHQ_VIRTUALENV_310=$CCHQ_VIRTUALENV-3.10
```
millerdev, line 3: This feedback was intended for https://github.com/dimagi/commcare-cloud/pull/5527, but it got merged too quickly.

> If `CCHQ_VIRTUALENV` is already set to `cchq-3.10` when the script is run, it will still create a new virtualenv `cchq-3.10-3.10`

I don't see how this change resolves the double suffix issue. If `CCHQ_VIRTUALENV` contains `cchq-3.10` then `CCHQ_VIRTUALENV_310` will be `cchq-3.10-3.10`. I think this would need to be made conditional to avoid that. Something like
```suggestion
if [[ $CCHQ_VIRTUALENV == *3.10* ]]; then
    CCHQ_VIRTUALENV_310=$CCHQ_VIRTUALENV
else
    CCHQ_VIRTUALENV_310=$CCHQ_VIRTUALENV-3.10
fi
```

But this is getting ridiculous. The extra variable is not necessary to resolve that issue, the check could have been added below under the `# if on 18.04 with 3.10 installed, use cchq-3.10` comment.

gherceg: I created https://github.com/dimagi/commcare-cloud/pull/5530 to implement it the way you suggested @millerdev. I was in a rush to go offline, and needlessly applied pressure to Marco to finish what should have been my responsibility. Apologies for the rushed solution.

skodde: I don't think the conditional is needed, my understanding is that `CCHQ_VIRTUALENV` can be used to set a base name for the venv, but we don't populate it if not with the default value of `cchq`, so it's always safe to use as a prefix for the 3.10 venv name.
I'm ok however with the new solution.

PR #5533, opened 2022-10-19
Announce Python 3.10 Transition

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-13500

This PR is responsible for communicating to users of commcare-cloud that they need to upgrade to Python 3.10. This is a re-write of https://github.com/dimagi/commcare-cloud/pull/5424, with the irrelevant refactor pulled out into a separate PR.

The commits are organized as follows:
- d9d1451cc0990a11fd3d4ff4455252a3be423d72: adds instructions to commcare-cloud installation docs for how to install and start using Python 3.10
- ba0b1c84afcfe565df95049e61c57799df537be0: adds a check to the init.sh script to inform users who are still using Python 3.6 that they need to upgrade
- 8a5f15e0526521c0619fed740e46fc6fce05e36e: creates a changelog entry detailing the upgrade, timeline, and instructions

##### ENVIRONMENTS AFFECTED

None


changelog/0058-upgrade-to-python-3-10.yml
```diff
@@ -0,0 +1,34 @@
...
+update_steps: |
+  Follow the steps below to properly install Python 3.10 and its dependencies.
+
+  $ sudo add-apt-repository -y ppa:deadsnakes/ppa
+  $ sudo apt update
+  $ sudo apt-get -y install python3.10 python3.10-dev python3.10-distutils python3.10-venv libffi-dev
+
+  Once Python 3.10 is installed on your control machine, run:
+  
+  $ update-code
```
skodde, line 24: Can we make explicit that this next step has to be run from an active cchq venv?

changelog/0058-upgrade-to-python-3-10.yml
```diff
@@ -0,0 +1,34 @@
...
+
+  which will pull the latest version of commcare-cloud and trigger the creation and activation of a Python 3.10 virtual environment.
+
+  Confirm the active virtual environment is using Python 3.10
+
+  $ python --version
```
skodde, line 30: I'd like us to start using and referencing `python3`, but I'm open to discuss this.

skodde (approved):

changelog/0058-upgrade-to-python-3-10.yml
skodde: I think it would be best practice based on https://peps.python.org/pep-0394/ and something we should get in the habit of doing (same as `python3 -m pip` instead of `pip`), since we moved away from 2.7 and only support 3,

millerdev: That PEP pertains to naming of the system python on Unixes. In a context such as this one where a virtualenv is expected to be activated, using `python` is preferred. From the PEP:

- While far from being universally available, `python` remains the preferred spelling for explicitly invoking Python, as this is the spelling that virtual environments make consistently available across different platforms and Python installations.
- When a virtual environment (...) is active, the `python` command should refer to the virtual environment’s interpreter and should always be available.
- As of 2019, activating a Python virtual environment (or its functional equivalent) prior to script execution is one way to obtain a consistent cross-platform and cross-distribution experience.
- If these conventions are adhered to, it will become the case that the `python` command is only executed in an interactive manner as a user convenience, or else when using a virtual environment or similar mechanism.

millerdev (approved):

changelog/0058-upgrade-to-python-3-10.yml
skodde: I see, I missed these updates. I see in "History of this PEP" that the rationale has changed in the presence of virtualenvs and also following the deprecation of Python 2.
I'm still uncertain about the "hybrid" state of our documentation and scripts, as some rely on the system version which, being from 18.04 with 2.7 still installed (yes, I know :-) ), creates some inconsistencies (but that would be true also for newer systems, where the presence of `python` is not guaranteed and its behavior undefined).
Also, I think the recommendation for the shebang (i.e. set it explicitly to `python3`) remains, and I'd probably prefer us to be consistent in this usage/nomenclature, as we're not dealing only with end users (as defined in the PEP) but also with a more complex system/setup.

millerdev: Sure, for non-virtualenv dependent scripts the best practice is to use `/usr/bin/python3`. In a virtualenv context, the PEP is explicit:
> For scripts that are only expected to be run in an activated virtual environment, shebang lines can be written as `#!/usr/bin/env python`, as this instructs the script to respect the active virtual environment.

skodde: Ok, so we should use/reference `python`/`#!/usr/bin/env python` for everything related to our venv (almost everything) and `python3`/`#!/usr/bin/python3` for the few things we have outside of it?

changelog/0058-upgrade-to-python-3-10.yml
gherceg: Marco and I talked offline and feel this is no longer necessary. Resolving, but please correct me if I'm wrong Marco 👍🏻

gherceg:
The plan is to merge this on October 31st, giving users 6 weeks (December 12th) to transition to Python 3.10.

skodde (approved):

quick_monolith_install/cchq-install.sh
```diff
@@ -40,13 +40,12 @@ sudo apt --assume-yes -qq update
...
-  sudo apt update
-  sudo apt-get --assume-yes -q install python3.10 python3.10-dev python3.10-distutils python3.10-venv libffi-dev
-else
-  sudo -H pip -q install ansible virtualenv --ignore-installed six
+if ! hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
+    echo "Starting December 19th, 2022, commcare-cloud will require Python 3.10."
+    echo "To upgrade, follow the instructions in:"
+    echo "   https://commcare-cloud.readthedocs.io/en/latest/installation/2-manual-install.html#upgrade-to-python-3-10"
 fi
+sudo -H pip -q install ansible virtualenv --ignore-installed six
```
gherceg, line 48: This is unnecessary on 3.10, but also harmless as far as I am aware so will remove once we drop support for 3.6.

gherceg:
@millerdev @skodde last batch of updates and targeting to merge this today.

millerdev (approved):

PR #5536, opened 2022-10-19
ucr pgbouncers

calellowitz:
https://dimagi-dev.atlassian.net/browse/USH-2498

The changes to sqlalchemy pooling had an unexepect effect on the other databases that shared a pgbouncer instance with the UCR db. This PR adds new pgbouncer machines and sets them as the targets for the UCR nlb. It will be rolled out in two phases, first adding the new targets to the nlb, then removing the old ones, so that it can be done with no downtime.

##### ENVIRONMENTS AFFECTED

Production


environments/production/postgresql.yml
```diff
@@ -148,3 +166,5 @@ dbs:
           - pgbouncer7
           - pgbouncer8
           - pgbouncer9
+          - pgbouncer10
+          - pgbouncer11
```
snopoke, line 170: Confirm that these added to all DB's not for the sake of the rollout? And they will be removed from most DBs in stage 2?

calellowitz: This seems to be our convention, which allows them to access all dbs, but does not affect the routing. All other pgbouncers are listed under every DB, so I tried to keep that consistent.

snopoke (approved):

shyamkumarlchauhan (approved):

PR #5542, opened 2022-10-21
archive deploy

calellowitz:
This is a first pass at creating deploy artifacts to enable autoscaling. It involves a minor change to the deploy process (adding another step to package up and store the release), and a lot of terraform code to create an s3 bucket to store that. I also added a script to fetch and unpackage the release, but did not include code to put it on the server, since I think it's exact location will be determined by how the rest of autoscaling is set up.

Opening now for review, and to make sure this doesn't conflict with Daniel's work on moving deploy to ansible, while I do further testing.

### DIFF
--- a/environments/staging/fab-settings.yml
+++ b/environments/staging/fab-settings.yml
@@ -7,3 +7,5 @@ deploy_event_url: https://api.github.com/repos/dimagi/dimagi-qa/dispatches
 generate_deploy_diffs: False
 custom_deploy_details:
   View staging branches here: https://github.com/dimagi/commcare-hq/blob/autostaging/scripts/staging.yml
+archive_release: True
+release_bucket: dimagi-commcare-staging-release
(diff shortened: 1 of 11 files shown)


snopoke (approved):
Seems OK as a starting point. A few thoughts I had while reviewing:

* Rather than save it to a 'current' file I think it would be better to save the release using the full release name since that gives us a history and also makes it easier to know which release is being installed (though it does mean we need to do cleanup).

* S3 doesn't support aliases or symlinks but there are a few strategies for doing more manual redirects e.g. https://stackoverflow.com/a/35043462/632517

* it is also very handy to store some metadata with the release like date, git commit hash. Especially if you can get that metadata without fetching the full release. This could be used when initiating a deploy for doing a deploy to generate the git diff etc. (we do a similar thing with [formplayer](https://github.com/dimagi/commcare-cloud/blob/b44a0476ce3172fca14198de2a007bca4fafce95/src/commcare_cloud/commands/deploy/formplayer.py#L193-L207)).

I assume that you're doing this at the end of a normal deploy just for the sake of testing?

PR #5544, opened 2022-10-23
Remove commented out USE_KAFKA_SHORTEST_BACKLOG_PARTITIONER: true

gherceg:
Mainly surfacing so that we can take action on this, whether that is removing it or creating a ticket to investigate. 

https://github.com/dimagi/commcare-cloud/pull/3737 commented this out in what was supposed to be temporary, but must have gotten dropped. @dannyroberts I'm curious to hear your thoughts on if this is still something worth looking into or not. 
##### ENVIRONMENTS AFFECTED

prod

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -212,7 +212,6 @@ localsettings:
         - 'ews-ghana'
         - 'ils-gateway'
         - 'ils-gateway-train'
-#  USE_KAFKA_SHORTEST_BACKLOG_PARTITIONER: True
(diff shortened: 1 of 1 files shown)


calellowitz (approved):

dannyroberts (approved):
Yeah, in fact if there's any code related to this feature you can delete that too. It was a spike attempt at something that was live for a few days, immediately caused issues, and was rolled back and abandoned.

gherceg:
Cool. Created https://github.com/dimagi/commcare-hq/pull/32305 in HQ to remove the code that runs under this setting.

PR #5547, opened 2022-10-24
Incorporate BackupPlan tag for EBS volumes into terraform

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-11932

##### ENVIRONMENTS AFFECTED
staging, production


dannyroberts:
I successfully applied this by running

```
cchq staging terraform apply $(for server_name in $(grep server_name: environments/staging/terraform.yml | cut -d'"' -f2); do echo "-target module.server__${server_name}"; done)
```
twice. (Terraform has two different ways you can handle tags on ec2 instance volumes and this PR switches which one we use, and I found that in order to properly switch it, I have to run this twice before it's properly updated.) After running a third time, I could see it was properly applied.

skodde (approved):

PR #5551, opened 2022-10-25
Remove airflow deployment code

millerdev:
We no longer use airflow. However, [india appears to still have an airflow database](https://github.com/dimagi/commcare-cloud/commit/fa08da143d146c55a1a1549762dde9ccfe0d74d8#diff-9523db7b7a952483237f255c40436f6b2f4813386f5188de039c4ca5473cbf28). Do we need to implement something to remove that before merging this PR?

##### ENVIRONMENTS AFFECTED

India?


snopoke (approved):
We could check the DB on india but I don't think it has anything useful in it and could likely just be removed (or backed up and then removed).

PR #5554, opened 2022-10-25
Clean up aws backup plan settings

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-11932

This has been applied already to staging. By my logic, none of these changes result in any meaningful functional changes.

##### ENVIRONMENTS AFFECTED
staging, production

### DIFF
--- a/src/commcare_cloud/terraform/modules/backup/main.tf
+++ b/src/commcare_cloud/terraform/modules/backup/main.tf
@@ -82,15 +82,13 @@ resource "aws_backup_vault_policy" "business_continuity_local_vault_policy" {
     "Version": "2012-10-17",
     "Statement": [
         {
+            "Sid": "Allow ${var.outside_account_id} to copy into ${aws_backup_vault.business_continuity_local_vault.name}",
             "Effect": "Allow",
-            "Principal": "*",
+            "Principal": {
+                "AWS": "arn:aws:iam::${var.outside_account_id}:root"
+            },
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/terraform/modules/backup/main.tf
```diff
@@ -106,28 +104,12 @@ resource "aws_backup_vault_policy" "business_continuity_remote_vault_policy" {
...
             },
-            "Action": [
-                "backup:CopyIntoBackupVault",
-                "backup:CopyFromBackupVault",
-                "backup:DeleteRecoveryPoint",
-                "backup:ListRecoveryPointsByBackupVault"
-            ],
+            "Action": "backup:CopyIntoBackupVault",
             "Resource": "*"
         }
```
skodde, line 112: Should we drop this entirely? I don't think the outside account needs to be able to copy anything into this vault.
The original policy was made I think to allow it to copy from this vault to its own and then delete what was copied, and the `CopyIntoBackupVault` permission was just a requirement for the others.

dannyroberts: I think the idea here is to allow us to copy a snapshot from the outside account into the main account, as part of a recovery step for example. I agree the way things are currently set up it's not likely to be useful, but I think it also doesn't hurt. Happy to delete it, but I'm also trying to keep these changes minimal to start with.

skodde (approved):

PR #5560, opened 2022-10-26
Revert "add backfill pillow to india"

snopoke:
Reverts dimagi/commcare-cloud#5557


snopoke:
@millerdev it was only a temporary pillow needed during the reindex process. Now that the new index is the primary index it is being written to by the `case-pillow` and this one is no longer needed.

PR #5573, opened 2022-10-31
Rollback formplayer with ansible instead of fabric

millerdev:
Formplayer deploy is currently implemented with ansible, but the rollback command was still implemented with fabric. This replaces the fabric implementation with ansible. The new rollback command is:

```sh
cchq <env> ansible-playbook rollback_formplayer.yml --tags=rollback
``` 

Confluence [rollback docs](https://confluence.dimagi.com/display/GTD/Deploy+Process#DeployProcess-RollingBack) will be updated with the new command after this is merged.

##### ENVIRONMENTS AFFECTED
None.


src/commcare_cloud/ansible/roles/formplayer/tasks/rollback.yml
```diff
@@ -34,3 +34,10 @@
       state: link
       force: yes
       follow: no
+
+  - name: Re-read supervisor config
+    command: supervisorctl reread
+
+  - name: Restart formplayer service
+    supervisorctl: name=all state=restarted
+    # on a monolith this actually restarts all services
```
millerdev, line 43: The commands used to restart here are a simplified version of what was done previously
https://github.com/dimagi/commcare-cloud/blob/b28a81770b673a7cca4298a44b25f85592740b06/src/commcare_cloud/fab/operations/supervisor.py#L104-L118

That code was originally added in https://github.com/dimagi/commcare-cloud/commit/010b20d80852c6ddf4c4dd3ad09037c80cbb56c1, which has no explanation of why it used so many commands to perform the restart.

dannyroberts (approved):

millerdev:
Added one more commit to remove `clean_formplayer_releases` fabric operation since formplayer release cleanup is done as part of the formplayer deploy.

gherceg (approved):

src/commcare_cloud/ansible/roles/formplayer/filter_plugins/filters.py
```diff
@@ -0,0 +1,14 @@
+from os.path import basename
+
+
+class FilterModule:
+
+    def filters(self):
+        return {'release_names': self.release_names}
+
+    def release_names(self, hostvars, play_hosts):
```
snopoke, line 9: A docstring would be useful. Why do you need to do a set intersection? Is it so that you only get 'global' releases and not 'limited releases?

snopoke (approved):
:+1:

src/commcare_cloud/ansible/roles/formplayer/filter_plugins/filters.py
millerdev: f403d6d1ddee568ce2e7179331680e594b2b05aa

Yes, the intersection is used to exclude releases that are not present on all hosts. It does not seem useful to roll back to a release that only present on a subset of hosts.

PR #5574, opened 2022-10-31
Inform user that old virtualenv still exists

gherceg:
Manish raised concern around the old virtualenv taking up unnecessary space which is totally fair. The only reason for not deleting the old directory was to facilitate reverting in urgent situations, but realistically the old virtualenv can be recreated easily, only adding a few additional minutes of setup.

I'm still in favor of keeping the name cchq-3.10 for the 3.10 venvs on 18.04 as a way to distinguish the unique setup that exists, and reserving `cchq` for default system python versions.

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -31,7 +31,7 @@ if [ -z ${CI_TEST} ]; then
     if [[ ! -f $VENV/bin/activate ]]; then
         if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null; then
             echo "Creating a python3.10 virtual environment named ${CCHQ_VIRTUALENV}"
-            if [ $OLD_VENV ]; then
+            if [ -n "$OLD_VENV" ]; then
                 echo "Your old virtual environment will remain at ${OLD_VENV}"
                 echo "If you wish to delete it, run 'rm -rf ${OLD_VENV}'"
             fi
--- a/control/init.sh
+++ b/control/init.sh
@@ -3,6 +3,7 @@ CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq}
 VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
 NO_INPUT=0
 BIONIC_USE_SYSTEM_PYTHON=${BIONIC_USE_SYSTEM_PYTHON:-false}
+OLD_VENV=''
 
 if [[ $_ == $0 ]]
 then
@@ -22,7 +23,8 @@ if [ -z ${CI_TEST} ]; then
     if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
         # only append 3.10 if it is not already in the name
         if [[ $CCHQ_VIRTUALENV != *"3.10"* ]]; then
-          CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
+            OLD_VENV=$VENV
+            CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
         fi
         VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
     fi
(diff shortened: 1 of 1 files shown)


mkangia:
Thanks @gherceg 

Don't have the expertise to review this but can confirm that staging control machine is close to full and I believe anyone who tries to set up new virtualenv would face a challenge with disk space going full and removing stale env seems like the right thing to do as well.

No preference on what it should be called.

skodde:
I'd prefer to avoid this kind of automatic operation. We can perform some manual maintenance if needed but we should leave it up to the user to decide.
Maybe we can add a notice or suggestion when setting up the new virtualenv instead.

gherceg:
Yeah I imagine from a disc space perspective, this disproportionally impacts us since we have so many users on a control machine. I agree that forcing the venv deletion upon users doesn't seem like the best approach, and a notice about the old virtualenv is better. I'll make those changes shortly.

control/init.sh
```diff
@@ -22,14 +23,19 @@ if [ -z ${CI_TEST} ]; then
...
         VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
     fi
     # check if a virtualenv at $VENV exists yet, and create if not
     if [[ ! -f $VENV/bin/activate ]]; then
         if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null; then
             echo "Creating a python3.10 virtual environment named ${CCHQ_VIRTUALENV}"
+            if [[ $OLD_VENV != '' ]]; then
+                echo "Your old virtual environment will remain at ${OLD_VENV}"
+                echo "If you wish to delete it, run 'rm -rf ${OLD_VENV}'"
+            fi
```
skodde, line 38:
```suggestion
            if [ -n "$CCHQ_VENV_OLD" ]; then
                echo "Your old virtual environment will remain at ${CCHQ_VENV_OLD}"
                echo "If you wish to delete it, run 'rm -rf ${CCHQ_VENV_OLD}'"
            fi
```

control/init.sh
```diff
@@ -22,14 +23,19 @@ if [ -z ${CI_TEST} ]; then
     if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
         # only append 3.10 if it is not already in the name
         if [[ $CCHQ_VIRTUALENV != *"3.10"* ]]; then
-          CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
+            OLD_VENV=$VENV
```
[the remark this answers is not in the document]
gherceg: I'm not too strongly attached to any naming convention here, but seeing as we set the path of the current virtualenv to `VENV`, I don't see why `OLD_VENV` isn't good enough. (An explanation along with the suggestion would be helpful)

control/init.sh
gherceg: Ah thanks. It looks like `[ $OLD_VENV ]` is sufficient and more similar to python so I'm going to go with that.

control/init.sh
```diff
@@ -3,6 +3,7 @@ CCHQ_VIRTUALENV=${CCHQ_VIRTUALENV:-cchq}
 VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
 NO_INPUT=0
 BIONIC_USE_SYSTEM_PYTHON=${BIONIC_USE_SYSTEM_PYTHON:-false}
+OLD_VENV=''
```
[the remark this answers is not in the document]
gherceg: See e564f4dcd3d1ed50d8af11e0a715e609170735e9

control/init.sh
gherceg: See e564f4dcd3d1ed50d8af11e0a715e609170735e9

skodde: Note that quoting the variable is recommended because otherwise we're making assumption about its content, and also that `test` and `test -n` are not exactly the same thing, and `test -n` would be more clear and explicit in this case.

control/init.sh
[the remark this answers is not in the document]
skodde: I don't have a strong opinion here either. Since I had other suggestions it made sense to me to change the variable to something that would be more specific to us, avoiding possible conflicts in the user env.

control/init.sh
gherceg: What about [these docs](https://tldp.org/LDP/Bash-Beginners-Guide/html/sect_07_01.html)?

> [ -n STRING ] or [ STRING ] | True if the length of "STRING" is non-zero.

Assuming I will wrap it in quotes, it guarantees the content will be a string so the `[ "$CCHQ_VENV_OLD" ]` check would be clear right?

gherceg: Offlined with Marco and added the `-n` option even though `[ STRING ]` and `[ -n STRING ]` do behave the same.

control/init.sh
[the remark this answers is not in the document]
gherceg: Cool. I updated the var in 895c590c8591c01c0b1c5595059894944ce5d1c9 to `CCHQ_VENV_PATH_OLD`.

skodde (approved):

PR #5587, opened 2022-11-03
Updates for rolling out ES phonetic plugin

snopoke:
This is now a requirement for CommCare (assuming you are using the case search index) hence the default has been changed to True.

This PR also adds a changelog entry for installing the  ES plugin and reindexing the index.

##### ENVIRONMENTS AFFECTED
3rd party envs


sravfeyn:
Is `es_plugins_phonetic_enabled: True` needed only for envs that need case-search index. If so, should the same be added in the 'Enable Case Search' related documentation as well?

snopoke:
> Is es_plugins_phonetic_enabled: True needed only for envs that need case-search index. If so, should the same be added in the 'Enable Case Search' related documentation as well?

Yes, it's only needed for the case-search index but I don't think it's possible to deploy commcare without creating that index (and without the plugin the preindex command will fail when trying to create the index).

snopoke:
> the 'Enable Case Search' related documentation

Where are these docs? I don't see them anywhere.

sravfeyn:
> but I don't think it's possible to deploy commcare without creating that index

Cool, in that case it doesn't matter whether an env is using case-search or not.

> the 'Enable Case Search' related documentation

I meant this https://confluence.dimagi.com/display/USH/Case+Search+Configuration#CaseSearchConfiguration-Overview

sravfeyn (approved):

shyamkumarlchauhan:
@snopoke there was a changelog which was pending for approval which has got the same entry as off yours , so either I need to merge ffirst so that you can raise a new changelog record. or you need to merge first so that I can raise a new changelog record. https://github.com/dimagi/commcare-cloud/pull/5578

PR #5598, opened 2022-11-08
Replace PrivilegedCommand with ansible

millerdev:
Also fix a few bugs in the `copy-files` command.

:blowfish: Review by commit.

##### ENVIRONMENTS AFFECTED
None.

### DIFF
--- a/src/commcare_cloud/commands/migrations/copy_files.py
+++ b/src/commcare_cloud/commands/migrations/copy_files.py
@@ -26,10 +26,12 @@ FILE_MIGRATION_RSYNC_SCRIPT = 'file_migration_rsync.sh'
 TEMPLATE_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'templates')
 REMOTE_MIGRATION_ROOT = 'file_migration'
 
+
 @attr.s
 class Plan(object):
     source_env = attr.ib()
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/commands/migrations/copy_files.py
```diff
@@ -218,12 +220,10 @@ def prepare_file_copy_scripts(target_host, source_file_configs, script_root):
 
 def copy_scripts_to_target_host(target_host, script_root, environment, ansible_context):
     local_files_path = os.path.join(script_root, target_host)
-
     destination_path = os.path.join('/tmp', REMOTE_MIGRATION_ROOT)
```
gherceg, line 223: I don't love that `destination_path` is defined both here and within `remove_scripts` as it seems prone to error. Maybe adding it as an optional parameter to `remove_scripts` could work?

gherceg (commented):
Holding off on approval until one comment is resolved.

src/commcare_cloud/commands/migrations/copy_files.py
millerdev: 27d6a00732dd3d9013db72eb2fa3721651959b16

gherceg: Even better!

PR #5602, opened 2022-11-09
Ansible connect via AWS SSM through control host

millerdev:
Improve the efficiency of the test/development cycle for features that use ansible by allowing ansible commands to be executed against staging hosts directly from the development machine.

Previously ansible commands could not connect to AWS hosts unless they were run from the _control_ machine. `ProxyCommand` options can be used to route ansible SSH connections through an SSM session on _control_ machine (as a jump host) and then on to the target host. This feature is only enabled if the `COMMCARE_CLOUD_PROXY_SSM` environment variable is set to a non-empty string.

:blowfish: Review by commit.

##### ENVIRONMENTS AFFECTED
None.

### DIFF
--- a/src/commcare_cloud/commands/ansible/run_module.py
+++ b/src/commcare_cloud/commands/ansible/run_module.py
@@ -14,7 +14,6 @@ from commcare_cloud.commands.ansible.helpers import (
     get_common_ssh_args,
     get_user_arg, run_action_with_check_mode)
 from commcare_cloud.commands.command_base import CommandBase, Argument
-from commcare_cloud.environment.main import get_environment
 from commcare_cloud.parse_help import ANSIBLE_HELP_OPTIONS_PREFIX, add_to_help_text, filtered_help_message
 
 NON_POSITIONAL_ARGUMENTS = (
@@ -80,9 +79,9 @@ class RunAnsibleModule(CommandBase):
         ))
 
     def run(self, args, unknown_args):
-        environment = get_environment(args.env_name)
-        environment.create_generated_yml()
         ansible_context = AnsibleContext(args)
+        environment = ansible_context.environment
+        environment.create_generated_yml()
 
         def _run_ansible(args, *unknown_args):
             return run_ansible_module(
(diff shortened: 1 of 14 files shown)


src/commcare_cloud/commands/ansible/run_module.py
```diff
@@ -138,6 +138,7 @@ def run_ansible_module(ansible_context, inventory_group, module, module_args,
             '-e', '@{}'.format(environment.paths.generated_yml),
         )
         cmd_parts += environment.secrets_backend.get_extra_ansible_args()
+    if needs_secrets or environment.secrets_backend.has_extra_env_vars():
```
dannyroberts, line 141: I think I see what you're doing—insert a special `ANSIBLE_SSH_EXTRA_ARGS` value if this is an AWS environment—but the secrets_backend seems like kind of an awkward place to graft that in. The secrets backend's job is to mediate fetching/storing environment secrets, not connection settings. Anything I'm missing about how this is logically tied to e.g. the choice to use AWS Secrets Manager vs ansible-vault for storing secrest?

If there's not a nuance here I'm missing, I might suggest putting these methods directly on `environment`—even though that's already somewhat cluttered, I think these methods go better there. I'm imagining the env vars would be additive—so first you'd add the env vars related to the secrets backend, and then separately you'd add the env vars related to the connection.

Does that make sense?

dannyroberts (commented):
I like how straightforward this ended up being! I had one separation of concerns question/comment, but I'm looking forward to having this

src/commcare_cloud/commands/ansible/run_module.py
millerdev: Yes, that makes sense, and I like the idea.

millerdev: https://github.com/dimagi/commcare-cloud/pull/5602/commits/8fc4a9ab31f663f24b3f6882085d96a6d085ce72 - best viewed with [this diff](https://github.com/dimagi/commcare-cloud/compare/09fb356b9761bfd3b0ee6442b3899e99c01cfd2f..8fc4a9ab31f663f24b3f6882085d96a6d085ce72), ignoring changes to `hosting_docs/source/reference/1-commcare-cloud/commands.md` from https://github.com/dimagi/commcare-cloud/pull/5602/commits/0e3047f84e5aaafdc0560a7bddefb9478bf9d2d2

PR #5618, opened 2022-11-15
Remove more fabric and convert some to ansible

millerdev:
Additionally, various small cleanups. The main fabric -> ansible work is in the last four commits.

:blowfish: Review by commit.

##### ENVIRONMENTS AFFECTED
None.

### DIFF
--- a/src/commcare_cloud/ansible/perform_system_checks.yml
+++ b/src/commcare_cloud/ansible/perform_system_checks.yml
@@ -1,12 +1,35 @@
 - hosts: django_manage
+
+  # uncomment these and comment the "become" lines below for local debugging
+  #connection: local
+  #vars:
+  #  code_home: '{{ "~/code/commcare-hq" | expanduser}}'
+  #  virtualenv_home: ../virtualenvs/hq
+
   become: true
   become_user: cchq
+  gather_facts: no
   tasks:
   - shell: 'cd {{ code_home }}; {{ virtualenv_home }}/bin/python manage.py check --deploy'
     register: deploy_check
+    failed_when: false
+    changed_when: deploy_check.rc
+  - name: "display output: manage.py check --deploy"
+    debug: var=deploy_check.stdout_lines
+    when: deploy_check.stdout_lines
+  - name: "display errors: manage.py check --deploy"
+    debug: var=deploy_check.stderr_lines
+    when: deploy_check.rc or deploy_check.stderr_lines
+    failed_when: deploy_check.rc
+
   - shell: 'cd {{ code_home }}; {{ virtualenv_home }}/bin/python manage.py check --deploy -t database'
     register: database_check
-  - name: "manage.py check --deploy"
-    debug: msg={{ deploy_check.stdout }}
-  - name: "manage.py check --deploy -t database"
-    debug: msg={{ database_check.stdout }}
+    failed_when: false
+    changed_when: database_check.rc
+  - name: "display output: manage.py check --deploy -t database"
+    debug: var=database_check.stdout_lines
+    when: database_check.stdout_lines
+  - name: "display errors: manage.py check --deploy -t database"
+    debug: var=database_check.stderr_lines
+    when: database_check.rc or database_check.stderr_lines
+    failed_when: database_check.rc
--- a/src/commcare_cloud/ansible/perform_system_checks.yml
+++ b/src/commcare_cloud/ansible/perform_system_checks.yml
@@ -1,0 +1,12 @@
+- hosts: django_manage
+  become: true
+  become_user: cchq
+  tasks:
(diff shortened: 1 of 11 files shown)


src/commcare_cloud/commands/ansible/run_module.py
```diff
@@ -136,8 +138,8 @@ def run_ansible_module(ansible_context, inventory_group, module, module_args,
 
     env_vars = ansible_context.build_env(need_secrets=become)
     if run_command is ansible_json:
-        env_vars.setdefault("ANSIBLE_LOAD_CALLBACK_PLUGINS", "1")
-        env_vars.setdefault("ANSIBLE_STDOUT_CALLBACK", "json")
+        env_vars["ANSIBLE_LOAD_CALLBACK_PLUGINS"] = "1"
+        env_vars["ANSIBLE_STDOUT_CALLBACK"] = "json"
```
dannyroberts, line 141: To check my understanding. The idea here is what you described as

> - Always set callback environment variables (it would break if they were not set properly).

i.e. the command these env vars feed into will break if `ANSIBLE_LOAD_CALLBACK_PLUGINS` and `ANSIBLE_STDOUT_CALLBACK` are set to anything other than these values, and so giving the caller the ability to (accidentally) override them is just asking for trouble without providing any value. Is that right?

src/commcare_cloud/fab/fabfile.py
```diff
@@ -192,6 +192,7 @@ def env_common():
         'staticfiles': staticfiles,
         'staticfiles_primary': [staticfiles[0]],
         'lb': [],
+        # deploy contains one server to which commcare-hq is deployed
```
dannyroberts, line 195: Took me a couple reads to understand the comment as worded here. I struggled to make it clearer on first read, but this is what I came up with to suggest.

```suggestion
        # 'deploy' contains a single server, one that commcare-hq is deployed to.
```

src/commcare_cloud/ansible/perform_system_checks.yml
```diff
@@ -0,0 +1,12 @@
+- hosts: django_manage
+  become: true
+  become_user: cchq
+  tasks:
+  - shell: 'cd {{ code_home }}; {{ virtualenv_home }}/bin/python manage.py check --deploy'
+    register: deploy_check
+  - shell: 'cd {{ code_home }}; {{ virtualenv_home }}/bin/python manage.py check --deploy -t database'
+    register: database_check
+  - name: "manage.py check --deploy"
+    debug: msg={{ deploy_check.stdout }}
```
dannyroberts, line 10: Nit: is it worth moving this task to before the previous one? That way if `manage.py check --deploy ` runs fine but the `manage.py check --deploy -t database` task fails hard for some reason, we still see the output from the first. It's not the most common thing, but that's my rationalization for feeling like these tasks should switch places. Feel free to ignore.

src/commcare_cloud/fab/fabfile.py
```diff
@@ -241,10 +241,7 @@ def send_email(subject, message, use_current_release=False):
 
 @task
 def kill_stale_celery_workers():
-    """
-    Kills celery workers that failed to properly go into warm shutdown
-    """
-    execute(release.kill_stale_celery_workers)
+    print("OBSOLETE use 'kill-stale-celery-workers' instead")
```
dannyroberts, line 244: Just thought of this, but if you put this message ("OBSOLETE ...") in the docstring as well, instead of removing it, then it would show up in the fab help output and documentation, which will go just a tad further into passively communicating the deprecation to people. Not something I'd insist on, just a passing thought.

dannyroberts (approved):
Looks great! Left a few comments for your perusal, nothing blocking

src/commcare_cloud/commands/ansible/run_module.py
millerdev: Correct.

src/commcare_cloud/ansible/perform_system_checks.yml
millerdev: Thanks for pointing this out. Turns out these debug tasks will ever get called if either of the shell tasks fails. This needed a lot more debugging and major rework. 69fef4efbab2a27d68bdcd85a8fc326368ee8b9

src/commcare_cloud/fab/fabfile.py
millerdev: 46233a411e34f5cbee01c71e19748c61c5ead34f

gherceg (commented):
Looked over the commits to stay up to date, but will leave to Danny to approve. Thanks for all of the clean up Daniel!

dannyroberts (approved):
Nice! Looks great

PR #5620, opened 2022-11-15
Create and use script for activating cchq virtualenv

gherceg:
In the case where we want to activate a virtualenv without going through the typical control setup process (aka invoking `init.sh`), we currently attempt to activate `CCHQ_VIRTUALENV` and default to `cchq` if that environment variable is not set. See [here](https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/commcare_cloud.py#L113-L116) for more context on where this occurs. 

Assuming `CCHQ_VIRTUALENV` variable is not exported in the user's bash profile (which it shouldn't be), that variable is not set until `init.sh` is run. This means we attempt to default to `cchq`. With the shift to `cchq-3.10`, we want this to activate `cchq-3.10` if it exists rather than attempt to activate `cchq`.

The solution is to pull the existing logic from `init.sh` that determines which virtualenv to activate, and create a new script that can be invoked from `init.sh` as well as directly. In the event that the targeted virtualenv does not exist, the script returns with an error. The `--quiet` arg is used from the `init.sh` script since that will handle creating the missing virtualenv, whereas the direct usage expects a failure to output info to the user.


control/activate_venv.sh
```diff
@@ -0,0 +1,44 @@
...
+    fi
+    VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
+fi
+
+# activate virtualenv if it exists, otherwise exit with error
+if [ -f $VENV/bin/activate ]; then
+    source $VENV/bin/activate
+    exit 0
+else
+    if [ "$quiet_mode" = "false" ] then;
```
joxl, line 39: Two things going on here:

1. The previous semicolon position is a syntax error.
2. **Suggestion**: when using an internal "bool" var (whose value you know is either `true` or `false` because your script controls it), it's simpler to just execute the value of the var because the `true` and `false` binaries are always present on all POSIX-compliant OSes.


```suggestion
    if $quiet_mode; then
```

joxl (changes requested):
Change requested for syntax error.

control/activate_venv.sh
```diff
@@ -0,0 +1,44 @@
...
+    echo "The variable BIONIC_USE_SYSTEM_PYTHON is set in your environment."
+    echo "This variable should only be used temporarily when it is absolutely necessary to use Python 3.6 on 18.04."
+    echo "Please remove this variable from your environment when you are able to use Python 3.10 again."
+fi
+
+# if on 18.04 with 3.10 installed, use cchq-3.10 unless $BIONIC_USE_SYSTEM_PYTHON is true
+if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null && [[ $( source /etc/os-release; echo $VERSION_ID ) == 18.04 ]]; then
+    # only append 3.10 if it is not already in the name
+    if [[ $CCHQ_VIRTUALENV != *"3.10"* ]]; then
+        CCHQ_VENV_PATH_OLD=$VENV
```
joxl, line 28: nit: prevent head-scratching bugs if whitespace shows up in a path somewhere

```suggestion
        CCHQ_VENV_PATH_OLD="$VENV"
```

control/activate_venv.sh
```diff
@@ -0,0 +1,44 @@
...
+        CCHQ_VIRTUALENV=$CCHQ_VIRTUALENV-3.10
+    fi
+    VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
+fi
+
+# activate virtualenv if it exists, otherwise exit with error
+if [ -f $VENV/bin/activate ]; then
+    source $VENV/bin/activate
```
joxl, line 36: Same as above (whitespace safety).

```suggestion
if [ -f "$VENV/bin/activate" ]; then
    source "$VENV/bin/activate"
```

control/init.sh
```diff
@@ -25,38 +25,26 @@ if [ -z ${CI_TEST} ]; then
...
-    fi
-    # check if a virtualenv at $VENV exists yet, and create if not
-    if [[ ! -f $VENV/bin/activate ]]; then
-        if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null; then
-            echo "Creating a python3.10 virtual environment named ${CCHQ_VIRTUALENV}"
-            if [ -n "$CCHQ_VENV_PATH_OLD" ]; then
-                echo "Your old virtual environment will remain at ${CCHQ_VENV_PATH_OLD}"
-                echo "If you wish to delete it, run 'rm -rf ${CCHQ_VENV_PATH_OLD}'"
+    # attempt to activate
+    if ! source ${COMMCARE_CLOUD_REPO}/control/activate_venv.sh --quiet; then
```
joxl, line 29: Whitespace safety.

```suggestion
    if ! source "${COMMCARE_CLOUD_REPO}/control/activate_venv.sh" --quiet; then
```

control/init.sh
```diff
@@ -25,38 +25,26 @@ if [ -z ${CI_TEST} ]; then
...
+        # check if a virtualenv at $VENV exists yet, and create if not
+        if [[ ! -f $VENV/bin/activate ]]; then
+            if [[ $BIONIC_USE_SYSTEM_PYTHON == false ]] && hash python3.10 2>/dev/null; then
+                echo "Creating a python3.10 virtual environment named ${CCHQ_VIRTUALENV}"
+                if [ -n "$CCHQ_VENV_PATH_OLD" ]; then
+                    echo "Your old virtual environment will remain at ${CCHQ_VENV_PATH_OLD}"
+                    echo "If you wish to delete it, run 'rm -rf ${CCHQ_VENV_PATH_OLD}'"
+                fi
+                # use venv because 3.10 setup includes installing python3.10-venv
+                python3.10 -m venv $VENV
```
joxl, line 39: Whitespace safety.

```suggestion
                python3.10 -m venv "$VENV"
```

control/init.sh
```diff
@@ -25,38 +25,26 @@ if [ -z ${CI_TEST} ]; then
...
+            else
+                # use virtualenv because `python3 -m venv` requires python3-venv
+                python3 -m pip install --user --upgrade virtualenv
+                python3 -m virtualenv $VENV
```
joxl, line 43: Whitespace safety.

```suggestion
                python3 -m virtualenv "$VENV"
```

control/init.sh
```diff
@@ -25,38 +25,26 @@ if [ -z ${CI_TEST} ]; then
...
             fi
-            # use venv because 3.10 setup includes installing python3.10-venv
-            python3.10 -m venv $VENV
-        else
-            # use virtualenv because `python3 -m venv` requires python3-venv
-            python3 -m pip install --user --upgrade virtualenv
-            python3 -m virtualenv $VENV
         fi
+        source $VENV/bin/activate
```
joxl, line 46: Whitespace safety.

```suggestion
        source "$VENV/bin/activate"
```

joxl (commented):
A few more suggestions to improve robustness.

control/activate_venv.sh
gherceg: > The previous semicolon position is a syntax error.

Distinctly remember writing this and thinking to myself "that's not right" but I clearly did not go back to check 😅 

@skodde and I had [a discussion](https://github.com/dimagi/commcare-cloud/pull/5574#discussion_r1011013702) about bool string values in bash not too long ago. I think you may have differing opinions and my goal is not to ruffle feathers, but to see if we can come to a consensus for consistency's sake. If we chalk it up to personal preference that is also fine!

I do like the idea of `if $quiet_mode; then` in cases where we are absolutely sure the variable will only be "true" or "false", but my question is are we able to guarantee that? Based on my understanding, I'd assume yes but maybe that's wrong.

control/init.sh
gherceg: See 577430651879a4e1a537484e6bf7822489082ccf

control/init.sh
gherceg: See 577430651879a4e1a537484e6bf7822489082ccf

control/init.sh
gherceg: See 577430651879a4e1a537484e6bf7822489082ccf

control/init.sh
gherceg: See 577430651879a4e1a537484e6bf7822489082ccf

control/activate_venv.sh
gherceg: Just realized the thread I linked doesn't discuss bools, so I totally made that. It makes sense why we would want to wrap the content of non-bools in quotations since we cannot make assumptions about its value. Sorry for the unnecessary ping!

joxl (approved):

control/activate_venv.sh
```diff
@@ -0,0 +1,44 @@
...
+    fi
+    VENV=~/.virtualenvs/$CCHQ_VIRTUALENV
+fi
+
+# activate virtualenv if it exists, otherwise exit with error
+if [ -f "$VENV/bin/activate" ]; then
+    source "$VENV/bin/activate"
+    exit 0
+else
+    if $quiet_mode; then
```
millerdev, line 39: Should this expression be negated? If not quiet mode, then show a message.

gherceg: Whoops! See 46c3888819a1d9eaab85f778477a88026ce8ca1b

gherceg:
Needed 4f5f138a9d61781288ad38c432986b1b42dc53b1 after testing this, since `BIONIC_USE_SYSTEM_PYTHON` was unset when calling `activate_venv.sh` directly.

08c119f57709d34552c96fc1c0db38101ad893b6 is mainly cleanup.

control/activate_venv.sh
```diff
@@ -35,7 +35,9 @@ fi
 
 # activate virtualenv if it exists, otherwise exit with error
 if [ -f "$VENV/bin/activate" ]; then
-    source "$VENV/bin/activate"
+    if [ $( which python ) != "$VENV/bin/python" ]; then
```
millerdev, line 38: I'm a little nervous about this technique since it's dependent on implementation details of the virtualenv provider. For example, I did a similar test in my dev environment:
```sh
$ which python
/home/dimagi/.pyenv/shims/python
$ echo "$VIRTUAL_ENV/bin/python"
/home/dimagi/.pyenv/versions/3.9.12/envs/cchq/bin/python
```
I think typical way to tell if a virtualenv is activated is to test if `$VIRTUAL_ENV` has a value. Alternately, you can just go ahead and `source "$VENV/bin/activate"` unconditionally—activating twice should be fine.

gherceg: Yeah I misunderstood an issue and chalked it up to this, but you can indeed attempt to activate an active virtualenv without issues. See 343a0d014be3826eb8692b5fc957eb8484cb2fc0

gherceg:
Addressing the test failure in https://github.com/dimagi/commcare-cloud/pull/5626 as it is unrelated to these changes.

joxl (approved):

src/commcare_cloud/commcare_cloud.py
```diff
@@ -121,7 +121,7 @@ def run_on_control_instead(args, argv, force_latest_code):
         )
     else:
         bash_commands_template = (
-            '{env}cd ~/commcare-cloud && source ~/.virtualenvs/${{CCHQ_VIRTUALENV:-cchq}}/bin/activate '
+            '{env}source ~/commcare-cloud/control/activate_venv.sh --fail-on-error'
```
millerdev, line 124: Currently `env` will always be non-empty (if `CCHQ_VIRTUALENV` is not set to an empty string) since `default_virtualenv` always has a value. I think some lines above need to change as well:

https://github.com/dimagi/commcare-cloud/blob/7a6622a601c7efc4e3810d73b5b67854384ef34f/src/commcare_cloud/commcare_cloud.py#L115-L116

Line 115 can be removed entirely since we do not support Python 2 anymore. Line 116 should change to
```py
    venv = os.environ.get("CCHQ_VIRTUALENV")
```

millerdev (changes requested):

src/commcare_cloud/commcare_cloud.py
millerdev: Update: I forgot that the activation script appends `-3.10` to the value of `CCHQ_VIRTUALENV` unless that suffix is already there (or unless `BIONIC_USE_SYSTEM_PYTHON=true`), so my comment was wrong and therefore not a blocker. However, the cleanup would still be nice.

millerdev (approved):

src/commcare_cloud/commcare_cloud.py
gherceg: Thanks. See 9e0ec6c10fffc0507a69a8eb4c5a5ce8d89a0296.

Offlined with Daniel and clarified that we still want to allow callers to pass a value in for `CCHQ_VIRTUALENV` if it is set, but agreed there is no need to provide a default since the `activate_venv.sh` script handles that, and the `ansible` venv has long been irrelevant.

gherceg:
My last commit just added a space in the command and does not impact any behavior, so I'm not going to wait for approval on that.

millerdev (approved):

src/commcare_cloud/commcare_cloud.py
```diff
@@ -112,22 +112,21 @@
...
+    venv = os.environ.get("CCHQ_VIRTUALENV")
     if force_latest_code:
         bash_commands_template = (
             '{env}cd ~/commcare-cloud && git fetch --prune && git checkout {branch} '
             '&& git reset --hard origin/{branch} && source ~/init-ansible && {cchq} {cchq_args}'
         )
     else:
         bash_commands_template = (
-            '{env}cd ~/commcare-cloud && source ~/.virtualenvs/${{CCHQ_VIRTUALENV:-cchq}}/bin/activate '
+            '{env}source ~/commcare-cloud/control/activate_venv.sh --fail-on-error '
```
joxl, line 123: I think this causes a "discontinuity problem" when executing `cchq --control ...` _from_ a system with this change (this PR) _against_ a control environment that does not have this change (yet).  In that scenario, commands are executed against the remote control system that depend on this file (`activate_venv.sh`) being present, but since it isn't, the command crashes.

gherceg: Thanks for flagging this Joel. My initial thought is that a change like this should have had a changelog/commcare-cloud announcement with it as well. I'll put that together now.

gherceg: See https://github.com/dimagi/commcare-cloud/pull/5649

PR #5629, opened 2022-11-18
Fix return

mkangia:
Fix bug introduced in https://github.com/dimagi/commcare-cloud/pull/5625/files#diff-43d45817760a8bf46406bde4e52babc6acf9847e159de7c806cc2da931ab2776R204

tested fix on staging that it works okay


AmitPhulera (approved):
Thanks 😄

millerdev (commented):
Thanks for fixing.

PR #5630, opened 2022-11-18
Detect changelogs since last deploy and alert on deploy

sravfeyn:
https://dimagi-dev.atlassian.net/browse/SC-1711 See background [here](https://docs.google.com/document/d/1EXVoCheX44bVs_GxVL2d7LOTsDZybKPbQ8l4kWU2v8M/edit)

This fetches the changelogs published since the last deploy and if there are alerts the user to review them. This is a useful feature for third-party hosting teams who may not be on top of the changelogs as we are.


sravfeyn:
@dannyroberts @kaapstorm @snopoke This affects the deploy process, so wanted to make sure this looks good to you all.

src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -51,14 +53,26 @@ def deployed_commit_matches_latest_commit(self):
...
 
+    def get_changelogs(self):
+        def get_commit_date(commit):
+            return datetime.strptime(
+                self.repo.get_commit(commit).last_modified,
+                "%a, %d %b %Y %H:%M:%S GMT"
+            )
+        return get_changelogs_in_date_range(
+            get_commit_date(self.current_commit),
+            get_commit_date(self.deploy_commit)
```
sravfeyn, line 64: It doesn't look like there is a way in Github API to get both the commits in one request, so making two separate requests.

tests/test_fab_utils.py
```diff
@@ -0,0 +1,17 @@
+from datetime import datetime
+from nose.tools import assert_equal
+
+from commcare_cloud.fab.utils import get_changelogs_in_date_range
+
+
+def test_get_changelogs():
+    assert_equal(
+        set(get_changelogs_in_date_range(datetime(2022, 11, 1), datetime(2022, 11, 11))),
```
snopoke, line 9: Since this relies on fetching actual data I suspect it may be flaky - it might be better to mock it to return the content from the local repo itself instead of from the web.

snopoke (commented):
Looks good. One comment re the tests

tests/test_fab_utils.py
sravfeyn: I have modified it as as suggested.
 
But, while the current downside is the test duration, the downside if we mock it is that if the Github API or its Python client has a bug the tests will no longer catch that issue.

I think this is a general problem as well with any code that needs to get data from external APIs. May be the ideal setup is to have a separate external API test matrix which we can test on a sparse weekly schedule or such?

Curious to hear your take on this problem.

snopoke: > I think this is a general problem as well with any code that needs to get data from external APIs. May be the ideal setup is to have a separate external API test matrix which we can test on a sparse weekly schedule or such?

For a critical function I would say yes but in this case a breaking change will get is not critical.

On that note, do you think we should catch any exceptions here to prevent blocking the deploy if it fails?

sravfeyn: Since the intention is to allow the deploy only after changelogs is reviewed, I think blocking it on exception is fine here. The tests ensures an exception free deploy process, but if there is an exception it will be good to hear about for which a hard fail is necessary. Otherwise, this feature might never work and we will never know.

snopoke: We could still show a message to indicate that there was an error and direct the user to check the changelog manually.

>  Otherwise, if this feature breaks we will never know.

Since this will impact our own deploys as well I think we'll notice if there is an error message.

Also, this is just for convenience so I really don't think it should block deploy. We don't block deploy if we aren't able to generate the deploy diff either.

sravfeyn: Fair point, I have added that.

src/commcare_cloud/fab/diff_templates/console.txt.j2
```diff
@@ -16,11 +16,16 @@ New version details:
 {{ mgs|warning }}
 {%- endfor %}
 
-{%- if changelogs -%}
-There have been some changelogs since last deploy, you must make sure that these are applied on your environment before proceeding to avoid getting your environment into a broaken state.
+{%- if changelogs.error -%}
+Changelogs:
+    Please review any pending changelogs from https://commcare-cloud.readthedocs.io/en/latest/ and make sure that these are applied on your environment before deploying to avoid getting your environment into a broaken state.
```
snopoke, line 21: nit: what about wrapping these long lines at 80 chars?

snopoke (approved):

src/commcare_cloud/fab/diff_templates/console.txt.j2
```diff
@@ -16,6 +16,19 @@ New version details:
 {{ mgs|warning }}
 {%- endfor %}
 
+{%- if changelogs.error -%}
+Changelogs:
+    Please review any pending changelogs from https://commcare-cloud.readthedocs.io/en/latest/ and make sure that these are applied on your environment before deploying to avoid getting your environment into a broaken state.
+
+{%- else -%}
+Changelogs:
+    There have been some changelogs since last deploy, you must make sure that these are applied on your environment before deploying to avoid getting your environment into a broaken state.
```
millerdev, line 25: @sravfeyn I'm seeing this notice when deploying staging, but I don't think it applies since no change logs are listed below. It looks like this is printed unconditionally, and I think that's a bug that was introduced in https://github.com/dimagi/commcare-cloud/commit/3391641f3e901eb3c781e5b1cf739a6050ca759a. Do you know why this is printed when there are no change logs, and is that the intended behavior?

sravfeyn: It looks like you have already addressed it, thank you!

PR #5649, opened 2022-11-28
Create changelog for activate_venv.sh script

gherceg:
Now that I've made this I will admit it feels strange, but I think it makes sense. A new file was added to commcare-cloud that requires that local and remote hosts of commcare-cloud are on a certain version or greater to work together properly. This changelog entry details that issue, and provides a solution.


millerdev (approved):
I agree, it's a little strange to do this, but I can't think of another way that isn't much more complicated.

PR #5651, opened 2022-11-29
update documentation for postgres upgrade process

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-13373


snopoke (approved):
It is worth trying to limit the scope to just the 'common_installs' role by adding `--tags=common_installs`?

PR #5654, opened 2022-11-29
Add section to PR template to announce new releases

gherceg:
After not doing this a couple of times after creating new changelogs, I think a reminder in the PR template would be helpful.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,9 @@
 <!--- Provide a link to the ticket or document which prompted this change -->
 
-##### ENVIRONMENTS AFFECTED
+##### Environments Affected
 <!--- list which environments are affected by this change or None if this doesn't change any environment files -->
+
+##### Announce New Release
+<!-- Delete this section if the PR does not contain any new changelog entries -->
+<!-- https://confluence.dimagi.com/display/saas/Announcing+changes+affecting+third+parties -->
+- [ ] I plan to announce this new release on the forum after merging this PR
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,6 +4,10 @@
 <!--- list which environments are affected by this change or None if this doesn't change any environment files -->
 
 ##### Announce New Release
+<!-- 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,9 @@
 <!--- Provide a link to the ticket or document which prompted this change -->
 
-##### ENVIRONMENTS AFFECTED
+##### Environments Affected
 <!--- list which environments are affected by this change or None if this doesn't change any environment files -->
+
+##### Announce New Release
+<!-- Delete this section if the PR does not contain any new changelog entries -->
```
millerdev, line 7: I suppose there is no way to conditionally add something to the template if the PR meets a condition (like contains a file with a specific path pattern)?

millerdev (approved):
Unfortunately this section will be deleted for almost all commcare-cloud PRs.

.github/PULL_REQUEST_TEMPLATE.md
joxl: Suggestion: alternatively use a label workflow like HQ's QA labels? The `labels.yml` file could be configured to add an **Announce** label or something similar.

joxl (approved):

.github/PULL_REQUEST_TEMPLATE.md
calellowitz: This is the same workflow as migrations in HQ (most delete, but it is always there). I think it's nice to be prompted with that question each time, forcing the conscious act of deletion, because it may cause the author to realize they should add a changelog, whereas the a label workflow requires them to already have thought of that. One alternative is to make it yes/no, check either "no release needed" or "yes release needed", and then it would not need to be deleted.

calellowitz (approved):

.github/PULL_REQUEST_TEMPLATE.md
gherceg: > I think it's nice to be prompted with that question each time, forcing the conscious act of deletion, because it may cause the author to realize they should add a changelog

As someone who has not added a changelog when I should have, I agree with this. I think the labels could only be used to supplement this, not replace it. I made one more change on 0dbfd87bfcad5b98b4aa50d6b2ed7891a2c48206 to include docs to how to determine if a changelog is necessary.

joxl (approved):

PR #5656, opened 2022-11-30
Remove bad code from flake8 config

czue:
Same as https://github.com/dimagi/django-prbac/pull/68

##### ENVIRONMENTS AFFECTED

n/a


millerdev (approved):
In case interested, this was originally added in https://github.com/dimagi/commcare-hq/pull/25152, and the warning [will never occur on Python 3+](https://pylint.pycqa.org/en/1.6.0/features.html?highlight=W1618#messages)

PR #5663, opened 2022-12-07
couchdb disk shrink

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-14163

##### Environments Affected
Production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -187,7 +187,7 @@ servers:
     volume_size: 90
     volume_encrypted: yes
     block_device:
-      volume_size: 8000
+      volume_size: 2700
       encrypted: yes
     group: "couchdb2"
     os: bionic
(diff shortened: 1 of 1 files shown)


shyamkumarlchauhan:
As disk used % will cross threshold level . can. we consider 2900 GB instead of 2700 GB

dannyroberts (approved):

dannyroberts:
> As disk used % will cross threshold level . can. we consider 2900 GB instead of 2700 GB
Had this discussion on slack, but to summarize: because we already know how we're going to further reduce disk usage in the next month or so (eventually removing the couch fixutres db now that that workload has been migrated to postgres), and because 80% is the "first think about how to address this without adding more disk" threshold, we're comfortable being over 80% for a short period of time while these plans play out.

PR #5664, opened 2022-12-07
Update manual install steps for Python 3.10

joxl:
No ticket, internal discussion that lead to realization that this needs updated.

##### Environments Affected

None.

### DIFF
--- a/hosting_docs/source/reference/1-commcare-cloud/1-installation.rst
+++ b/hosting_docs/source/reference/1-commcare-cloud/1-installation.rst
@@ -117,12 +117,12 @@ If you get to this point, congratulations! ``commcare-cloud`` is installed.
 Manual Installation
 -------------------
 
-You will need python 3.6 installed to follow these instructions (for
-Ubuntu 18.04; other operating systems may differ):
+You will need python 3.10 installed to follow these instructions (for
+Ubuntu 22.04; other operating systems may differ):
 
 .. code-block::
 
-   sudo apt-get install git python3-dev python3-pip
+   sudo apt-get install git python3.10-dev python3-pip
 
 Setup
 ^^^^^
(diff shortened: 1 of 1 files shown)


RynhardPietersen (approved):
Looking good! It's more readable and up to date!

hosting_docs/source/reference/1-commcare-cloud/1-installation.rst
```diff
@@ -117,12 +117,12 @@ If you get to this point, congratulations! ``commcare-cloud`` is installed.
 Manual Installation
 -------------------
 
-You will need python 3.6 installed to follow these instructions (for
-Ubuntu 18.04; other operating systems may differ):
+You will need python 3.10 installed to follow these instructions (for
+Ubuntu 22.04; other operating systems may differ):
```
gherceg, line 121: I'm not sure we want to reference 22.04 in public docs just yet, but I do agree with the 3.10 change.

millerdev, line 121: As far as I know we do not support 22.04 yet, or is there news I have not heard?

millerdev (commented):
I'm interested to see @gherceg's review. Is this documentation that was missed in the Python 3.10 upgrade, or was the plan to update it at some point in the future?

hosting_docs/source/reference/1-commcare-cloud/1-installation.rst
```diff
@@ -152,23 +152,42 @@ Manual setup
...
+
+**NOTE**: *The virtualenv name and location may be customized, below example uses ``cchq``
+and ``~/.virtualenvs/cchq``. Adjust according to your preferred configuration.*
+
+.. code-block:: sh
+
+   # using virtualenv
+   python3 -m pip install --user --upgrade virtualenv
+   python3 -m virtualenv ~/.virtualenvs/cchq
+   source ~/.virtualenvs/cchq/bin/activate
```
gherceg, line 166: a) I think we need to be explicit about python3.10 because this is outside of a virtual environment, and we don't know for sure what version python3 references. 
b) We are shifting to removing the dependency on `virtualenv` since python has `venv` built-in. 
```suggestion
   # using venv
   python3.10 -m venv ~/.virtualenvs/cchq
```

hosting_docs/source/reference/1-commcare-cloud/1-installation.rst
```diff
@@ -117,12 +117,12 @@ If you get to this point, congratulations! ``commcare-cloud`` is installed.
...
 
-You will need python 3.6 installed to follow these instructions (for
-Ubuntu 18.04; other operating systems may differ):
+You will need python 3.10 installed to follow these instructions (for
+Ubuntu 22.04; other operating systems may differ):
 
 .. code-block::
 
-   sudo apt-get install git python3-dev python3-pip
+   sudo apt-get install git python3.10-dev python3-pip
```
gherceg, line 125: I'm wondering if this section should point to the [instructions] for installing 3.10 since that ensures _everything_ that is necessary is installed. I don't like where that section on installing 3.10 currently lives though, so maybe this could be the impetus to move to a more logical spot that both sets of directions can reference?

(a) I think we need to be explicit about python3.10 because this is outside of a virtual environment, and we don't know for sure what version python3 references. 
b) We are shifting to removing the dependency on `virtualenv` since python has `venv` built-in. 
```suggestion
   python3.10 -m venv ~/.virtualenvs/cchq
```) for installing 3.10 given that the current config that we support is 3.10 on 18.04.

gherceg:
I missed @millerdev 's review before I submitted mine, but to answer:
> Is this documentation that was missed in the Python 3.10 upgrade, or was the plan to update it at some point in the future?

I missed this, and based on my understanding it is an appropriate time to update these docs since we now support and encourage 3.10.

hosting_docs/source/reference/1-commcare-cloud/1-installation.rst
joxl: > this section should point to the [instructions] for installing 3.10

I like that idea, can you point me to those instructions?

gherceg: 🤦🏻 updated my initial comment to actually have the link, but here are [the instructions](https://commcare-cloud.readthedocs.io/en/latest/installation/2-manual-install.html#upgrade-to-python-3-10).

millerdev: I think it's [here](https://github.com/dimagi/commcare-cloud/blob/master/hosting_docs/source/installation/2-manual-install.rst#upgrade-to-python-310). As an example, I think `python3.10-venv` (apt package) is needed for `python3.10 -m venv ...`

joxl: See latest changes.

millerdev (approved):

PR #5672, opened 2022-12-09
Install JDK 17 and set as the version to Staging

avazirna:
[SAAS-14201](https://dimagi-dev.atlassian.net/browse/SAAS-14201)

##### Environments Affected

Staging

##### Announce New Release

- [X] After merging, I will follow [these instructions](https://confluence.dimagi.com/display/saas/Announcing+changes+affecting+third+parties#Announcingchangesaffectingthirdparties-announcing)


shyamkumarlchauhan:
Though practically we can install multiple versions of java on a ubuntu machine but when we call the java for use to run different applications by default the /usr/bin/java will be called.
so it might break the other services which uses java like kafka, elasticsearch , kinesis agent . another approach of this can be  use var defined in public.yml and use that var to be called to install and run only the relevant service and this way it won't break any other service .

shyamkumarlchauhan (approved):

shyamkumarlchauhan (approved):

PR #5673, opened 2022-12-09
increase max_content_length

calellowitz:
A domain in prod is submitting forms larger than 100mb (the ES doc size limit), which are piling up in the pillow errors table, and painfully slow to retry. This increases the size limit so that those forms will successfully make it into ES.

##### Environments Affected

All


dannyroberts (approved):
impressive

snopoke (commented):
Is it reasonable to accept forms of this size?

PR #5675, opened 2022-12-12
Update occurrences of 'zookeeper' service name

skodde:
'zookeeper' service was renamed to 'zookeeper-server' in a previous PR (#5135).
Update our code accordingly so that we reference the correct service.

##### Environments Affected
All

### DIFF
--- a/src/commcare_cloud/ansible/roles/kafka/templates/kafka-server.service.j2
+++ b/src/commcare_cloud/ansible/roles/kafka/templates/kafka-server.service.j2
@@ -2,7 +2,7 @@
 Description=Apache Kafka server (broker)
 Documentation=http://kafka.apache.org/documentation.html
 Requires=network.target remote-fs.target
-After=network.target remote-fs.target kafka-zookeeper.service
+After=network.target remote-fs.target zookeeper-server.service
 
 [Service]
 Environment=JMX_PORT={{ kafka_jmx_port }}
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/ansible/roles/kafka/templates/kafka-server.service.j2
```diff
@@ -2,7 +2,7 @@
 Description=Apache Kafka server (broker)
 Documentation=http://kafka.apache.org/documentation.html
 Requires=network.target remote-fs.target
-After=network.target remote-fs.target kafka-zookeeper.service
+After=network.target remote-fs.target zookeeper-server.service
```
mkangia, line 5: what is that change?

mkangia (commented):
Thanks for taking this up @skodde 
i can't think of the impact of this change fully so i will let others review for approval.

src/commcare_cloud/ansible/roles/kafka/templates/kafka-server.service.j2
skodde: `kafka-zookeeper` does not exist, I'm not sure if previously it was called that way when/if bundled with kafka, but the scope of that line is to make sure that `kafka-server` is started only after those other services (on the same machine), including now the correct `zookeeper-server`.

calellowitz (approved):

PR #5683, opened 2022-12-14
Staging release bucket

calellowitz:
This pulls the terraform work out of https://github.com/dimagi/commcare-cloud/pull/5542 so that it can be merged while I work on the archive commands.

##### Environments Affected

Staging


src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
```diff
@@ -248,6 +248,12 @@ module "logshipping" {
   account_id = {{ account_id|tojson }}
 }
 
+module "autoscaling" {
+  source = "./modules/autoscaling"
+  release_bucket = var.release_bucket
+  commcare_server_role_id = module.server_iam_role.commcare_server_role_id
+}
```
dannyroberts, line 255: If we merge this as is, I think this will error on production, because this variable is `None`. I think probably you want to "hide" this addition from environments other than staging until you're ready to apply it, by wrapping it in

```
{% if release_bucket %}
...
{% endif %}
```

dannyroberts (commented):
I had just the one comment. I'd recommend seeing what `terraform plan` looks like on on production or india with your final change, to prevent this from causing an unwanted diff there.

dannyroberts:
@shyamkumarlchauhan I'll let you review/approve, and merge

dannyroberts (approved):

dannyroberts (approved):

PR #5709, opened 2023-01-10
Fix rebuild commands

mkangia:
Couple of issues came up when we were following the documentation for rebuild on a test environment.
1. PG was not restarted post shut down to kill connections to execute data wipe. So that is now added
2. An initial migration is needed post wipe to setup tables before deploy. If this isn't done, deploy fails at `ensure_checkpoints_safe` task which needs `pillowtop_kafkacheckpoint` table to be present. So that is resolved by running the migrate_on_fresh_install playbook

##### Environments Affected

None


sravfeyn (approved):
Thank you!

PR #5716, opened 2023-01-11
Display Maintenance Warnings for 3rd party envs

sravfeyn:
https://dimagi-dev.atlassian.net/browse/SC-1765
Also see the [spec](https://www.google.com/url?sa=t&rct=j&esrc=s&source=appssearch&uact=8&cd=8&cad=rja&q=&sig2=7WzMn9579V0TT9Z1QBJgaA&ved=0ahUKEwiXp9f5mL_8AhUXEtUKHRV1BQY4ABABKAgwCA&url=https://drive.google.com/a/dimagi.com/open%3Fid%3D10NCEdVLFjIlodHTQyM-dwr4jOeyHmdcdtJQ-ocMPJ28&usg=AOvVaw3fomtpFRWnwedGbKXXspaC)

##### Announce New Release

- [x] After merging, I will follow [these instructions](https://confluence.dimagi.com/display/saas/Announcing+changes+affecting+third+parties#Announcingchangesaffectingthirdparties-announcing)


src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -82,6 +82,32 @@ def get_commit_date(commit):
                 "error": True
             }
 
+    def get_maintenance_pr_context(self):
+        from github import Github
+
+        now = datetime.now().strftime( '%Y-%m-%d')
+        sixweeks_ago = (datetime.now() - timedelta(days=6*7)).strftime( '%Y-%m-%d')
+        created = f"{sixweeks_ago}..{now}"
+        query = f"repo:dimagi/commcare-hq is:pr is:open created:{created} label:maintenance0,maintenance1"
```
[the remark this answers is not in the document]
sravfeyn: The closed/merged PRs are already available under `prs_by_label` variable. The relevant warnings are displayed [here](https://github.com/dimagi/commcare-cloud/pull/5716/files#diff-fc37883b9b748511bd5debb09c5317b4471055ba3f6becbc82db7ea3ca07bca4R57) for the merged PRs.

The `prs_by_label` has only merged ones, so the above util gets the open ones with last six weeks that have the maintenance labels for which a maintenance should be performed within next six weeks (maximum)

src/commcare_cloud/fab/diff_templates/console.txt.j2
```diff
@@ -29,6 +29,16 @@ Changelogs:
 {% endfor %}
 {%- endif %}
 
+{%- if not (is_dimagi_env and maintenance_prs.error) and maintenance_prs.maintenance_prs -%}
```
snopoke, line 32: shouldn't this be `or`:
```suggestion
{%- if not (is_dimagi_env or maintenance_prs.error) and maintenance_prs.maintenance_prs -%}
```

src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -82,6 +82,32 @@ def get_commit_date(commit):
                 "error": True
             }
 
+    def get_maintenance_pr_context(self):
+        from github import Github
+
+        now = datetime.now().strftime( '%Y-%m-%d')
+        sixweeks_ago = (datetime.now() - timedelta(days=6*7)).strftime( '%Y-%m-%d')
+        created = f"{sixweeks_ago}..{now}"
```
snopoke, line 90: nit: slightly simpler:
```suggestion
        created = f">{sixweeks_ago}"
```

src/commcare_cloud/environment/main.py
```diff
@@ -66,6 +66,10 @@ def check(self):
             self.proxy_config
         self.create_generated_yml()
 
+    @property
+    def is_dimagi_env(self):
+        return self.name in ["production", "staging", "india"]
```
snopoke, line 71: instead of doing this, what about having variable in the environment that determines whether to display the info? e.g. 'show_maintenance_info_on_deploy'?

src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -90,7 +116,8 @@ def get_diff_context(self):
             "LABELS_TO_EXPAND": LABELS_TO_EXPAND,
             "errors": [],
             "warnings": [],
-            "changelogs": self.get_changelog_context()
+            "changelogs": self.get_changelog_context(),
+            "maintenance_prs": self.get_maintenance_pr_context()
```
snopoke, line 120: It would be better to skip this method call entirely if `is_dimagi_env` is True and just return a static value.

src/commcare_cloud/fab/diff_templates/console.txt.j2
sravfeyn: This message is to be displayed if it's not a dimagi env AND if there are no errors, so it should be `and`

src/commcare_cloud/fab/deploy_diff.py
sravfeyn: True!

src/commcare_cloud/environment/main.py
sravfeyn: True, actually however it appears there is already a public var called `IS_DIMAGI_ENVIRONMENT` so I can use that. It is unlikely we need this on Dimagi envs since we do all the maintenance proactively, so an extra var is not super necessary.

sravfeyn: Or may be some it will be nice to have it enabled even if it is a Dimagi env such as PNA or SWISS. I will update

src/commcare_cloud/fab/deploy_diff.py
sravfeyn: True

sravfeyn:
Thanks for the review @snopoke I have updated the code per your feedback

src/commcare_cloud/fab/diff_templates/console.txt.j2
snopoke: The challenge is that the `not` is outside the brackets which invokes de Morgans law of boolean algebra:

https://ryanstutorials.net/boolean-algebra-tutorial/boolean-algebra-laws.php#demorgan

```
NOT(p AND k) = NOT(p) OR NOT(k)

and

NOT(p OR k) = NOT(p) AND NOT(k)
```

I think the expanded logic should be (correct me if this is wrong):
```
not is_dimagi_env and not maintenance_prs.error and maintenance_prs.maintenance_prs
```

which means that the contracted expression is:
```
not (is_dimagi_env or maintenance_prs.error) and maintenance_prs.maintenance_prs
```

snopoke: With the latest commit this is moot but probably still useful to know

snopoke (approved):
minor comments remaining but nothing blocking

src/commcare_cloud/fab/deploy_diff.py
```diff
@@ -82,6 +83,37 @@ def get_commit_date(commit):
...
             }
 
+    def get_maintenance_pr_context(self):
+        if not self.show_maintenance_updates_on_deploy:
+            return {
+                "maintenance_prs": [],
+                "error": False
+            }
+
+        from github import Github
```
snopoke, line 93: why is this import here and not at the module level?

snopoke (approved):

hosting_docs/source/reference/1-commcare-cloud/commands.md
```diff
@@ -30,7 +30,7 @@ commcare-cloud [--control] [--control-setup {yes,no}] <env> <command> ...
 
 server environment to run against
 
-## Options
+## Optional Arguments
```
millerdev, line 33: Do you know why this changed? It doesn't seem like that should happen in this PR.

sravfeyn: Not sure why I can't see my comment that I made a week ago, but I had resolved this issue. I think originally the issue was due to a difference in my local version of argparse.

PR #5718, opened 2023-01-13
Add Java 17 bin path parameter

avazirna:
Considering that we are now supporting multiple versions of Java, this PR introduces a new parameter to facilitate setting an environment's Formplayer Java version. 

##### Environments Affected
Staging


environments/staging/public.yml
```diff
@@ -208,4 +208,4 @@ python_version: '3.9'
 use_deadsnakes: '{{ ansible_distribution_version <= "18.04" }}'
 allow_aws_ssm_proxy: true
 
-formplayer_java_version: /usr/lib/jvm/java-1.17.0-openjdk-amd64/bin/java
+formplayer_java_version: "{{ java_17_bin_path }}/java"
```
snopoke, line 211: @avazirna 

Was this change tested? The `java_17_bin_path` variable isn't defined in a place that is accessible from here.

```
ansible.errors.AnsibleUndefinedVariable: {{ formplayer_java_version | default('java') }}: {{ java_17_bin_path }}/java: 'java_17_bin_path' is undefined
```

avazirna: @snopoke I relied mainly on Github tests, which were successful. However, it seems that these don't cover the affected part as it turns out the setting needed to be defined elsewhere to be accessible to Ansible, this was corrected with this [PR](https://github.com/dimagi/commcare-cloud/pull/5728).

PR #5729, opened 2023-01-18
increase pgmain1 disk size by 500GB

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-14294

##### Environments Affected
Production


dannyroberts:
Let's hold off on this. Cal's working on a way to quickly reduce usage so we won't require more disk; it's something he's been aware of for some time but prioritizing today.

PR #5731, opened 2023-01-20
Collect OS level info for audit

SmittieC:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-2462)

### Description
Added the collection of some basic info from the OS that might be good to know. The idea here is to check the permissions/ownership of some files/directories that might be a cause of some issue. I _think_ the only one that we need to check is the `/opt/data` directory? If there's any others that might be good to include, or other stats of each file/dir, please shout (and then message me)

##### Environments Affected

All

## Example output (so far) of info.txt file after this change
```json
{
    "environment":"monolith",
    "commcare_cloud":{
       "commit":"67fbd722f800b03947a3e55837279047dbeecb47"
    },
    "commcare_hq":{
       "commit":"9bb2e466ea790d4f23982f06c18aaa85baecf49c"
    },
    "settings_validation":{
       "passed":true,
       "failure_reason":null
    },
    "os_data":{
       "os_distrib":{
          "DISTRIB_ID":"Ubuntu",
          "DISTRIB_RELEASE":"18.04",
          "DISTRIB_CODENAME":"bionic",
          "DISTRIB_DESCRIPTION":"\"Ubuntu 18.04.6 LTS\""
       },
    }
 }
```
The new audit directory (for a cluster having two machines with hostnames `control-worker` and `worker1` respectively):

Screenshot from 2023-02-01 15-41-02

In the file `disk_usage.txt` is the output of the `df -h` command and the output of the `opt_data_file_permissions.txt` file is like the template introduced in this PR.


src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -502,6 +504,7 @@ def run(self, args, unknown_args):
         self.collect_commcare_cloud_details()
         self.collect_commcare_hq_details()
         self.validate_environment_settings()
+        self.collect_control_machine_os_level_info()
```
mkangia, line 507: nit: I reckon all of these could be private functions?

mkangia (commented):
what is the vision of when the snapshot will be used? I believe that will help understand what information will be useful to have in the snapshot.

The documentation says "it good to run this before making any changes to the environment". When would OS level info saved here come in handy? I believe only if one updates their OS version which isn't something one would do through commcare cloud.

src/commcare_cloud/commands/ansible/ops_tool.py
SmittieC: True. I'll update these to private

SmittieC:
> what is the vision of when the snapshot will be used? I believe that will help understand what information will be useful to have in the snapshot.

The vision is to make the troubleshooting of issues easier. If a 3rd party is having issues and we're engaging in a back-and-forth with them to give us "this and that" info, it might help to have them give us this "audit" information rather. The dream is to have this reveal something to us (or whoever is doing the support work).

> When would OS level info saved here come in handy?

My thought here was along the lines of "rather having this information than not". It might be that - during troubleshooting - we spot a 3rd party with an OS version that's unsupported. I do suppport the notion of not including info for something that _might_ come in handy in a corner case, but it's low hanging fruit in this case, which is why I included it. I'd be happy to take it out if there's a reason why we should not include it.

mkangia:
thanks @SmittieC 
No objection on adding this.

mkangia (approved):

src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -517,20 +520,51 @@ def _create_audit_dir(self):
...
+        os_data["os_distrib"] = os_distrib
+        
+        # File/Dir stats
+        paths = ["/opt/data"]
+        file_stats = []
+        for path_raw in paths:
+            path = Path(path_raw)
+            stats = {}
+            path_stats = path.stat()
+            stats["path"] = path_raw
```
Charl1996, line 563: Is this really necessary to log? I get the OS info, but not really the motivation for the `path_stats`?

Charl1996: I just think it's not contributing any information as opposed to the OS info, so rather leave it out.

SmittieC: It's technically not needed. That said, I did realize this was done wrong. I'm working on an update for this branch

SmittieC:
I'm adding new changes to this PR. I realized I missed the fact that this the checks in this PR should happen on all hosts

SmittieC:
Big update: Previously I naively checked the OS info (disk space and /opt/data dir permissions) only for the control machine. I added an ansible role to do these checks for all hosts

mkangia:
> for all hosts

Now, that is what we call "All hands"

src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -487,50 +486,83 @@ class AuditEnvironment(_AnsiblePlaybookAlias):
...
         except Exception as exception:
             settings_validaton["passed"] = False
             settings_validaton["failure_reason"] = str(exception)
         self.env_info_dict["settings_validation"] = settings_validaton
+
+    def _collect_control_machine_os_level_info(self):
+        os_data = {}
+        
+        os_distrib = {}
+        with open("/etc/lsb-release", "r") as file:
```
sravfeyn, line 551: Can't this be captured in `commcare-audit.yml` as well?

sravfeyn (approved):

src/commcare_cloud/commands/ansible/ops_tool.py
SmittieC: Yes you are correct. I'm going to create a separate PR for this change/refactor

PR #5739, opened 2023-01-26
Collect service status for audit

SmittieC:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-2461)

##### Description
We want to collect the status of each service and save that in the audit directory. I initially wanted to just capture the output of the `django-manage check_services` command, but boy that proved to be a challenge (still unsolved), since we need to capture the stdout of the remote server and save that on the control server. I went with just using the equivalent of the `cchq <env> service <service name> status` command and redirected the ansible log - which shows the status - to the audit directory for each status check. This doesn't require a lot of new dev work and seems to be the simplest solution. It does the job.

##### Environments Affected

All (Any)

This is what the `~/.commcare-cloud/audits` directory will look like after this change. Notice the files with the service names. Each file contains the status for that service at that time as per ansible. It's not as clean as the django manage command, but this will do the job for now.
Screenshot from 2023-01-25 16-51-26


src/commcare_cloud/commands/ansible/ops_tool.py
```diff
@@ -568,3 +576,21 @@ def _collect_control_machine_os_level_info(self):
...
         self.env_info_dict["os_data"] = os_data
+
+    def _collect_service_status_info(self):
+        """Collects service information for each service in SERVICE_NAMES. This method changes the
+        ansible context for each check so that ansible logs the status to a service-specific file.
+        """
+        self.service_status_directory = os.path.join(self.curr_audit_directory, "service_statuses")
+        self._ensure_dir_exists(self.service_status_directory)
+
+        for service_name in SERVICE_NAMES:
```
snopoke, line 587: You can ignore one of `elasticsearch` or `elasticsearch-classic` since they both do the same status check (`elasticsearch` calls `elasticsearch-classic` for status checks).

snopoke (approved):

kaapstorm (approved):

SmittieC:
Closing this one out in favour of [this new PR](https://github.com/dimagi/commcare-cloud/pull/5744/files)

PR #5744, opened 2023-02-01
Collect service status for audit

SmittieC:
[Ticket](https://dimagi-dev.atlassian.net/browse/SC-2461)
Note: The [previous PR](https://github.com/dimagi/commcare-cloud/pull/5739) for these changes are closed due to rebasing. I thought it might be easier to just open a new one.

##### Description
We want to collect the status of each service and save that in the audit directory. I initially wanted to just capture the output of the `django-manage check_services` command, but boy that proved to be a challenge (still unsolved), since we need to capture the stdout of the remote server and save that on the control server. I went with just using the equivalent of the `cchq <env> service <service name> status` command and redirected the ansible log - which shows the status - to the audit directory for each status check. This doesn't require a lot of new dev work and seems to be the simplest solution. It does the job.

##### Environments Affected

All (Any)

This is what the `~/.commcare-cloud/audits` directory will look like after this change. Notice the files with the service names. Each file contains the status for that service at that time. It's not as clean as the django manage command, but this will do the job for now.
Screenshot from 2023-01-25 16-51-26


src/commcare_cloud/commands/ansible/helpers.py
```diff
@@ -30,6 +30,7 @@ def __init__(self, args, environment=None):
             assert args.env_name == environment.name, (args.env_name, environment.name)
         self.environment = environment
         self._stdout_callback = getattr(args, 'stdout_callback', None)
+        self.ansible_logfile = ansible_logfile
```
SmittieC, line 33: This allows us to redirect the ansible output to a specific file (other than ansible.log)

Charl1996 (approved):
I have no objections.

PR #5751, opened 2023-02-04
requirements.txt can just use setup.py

kaapstorm:
Instead of using pip-compile to duplicate the requirements from setup.py, just use `-e .` to use requirements from setup.py directly.

What do you think @millerdev ?

##### Environments Affected

None


millerdev (commented):
The requirements file pins indirect requirements, which are not listed in setup.py. We would lose those pins. Example:
```
paramiko==2.11.0
    # via fabric3
```

PR #5774, opened 2023-02-14
Do not stop celery and pillowtop before db migrate

millerdev:
All database migrations must be backward-compatible, so stopping services is unnecessary. Webworkers are not stopped, and they can do the same things that celery or pillow process do, so it does not make sense to treat pillow and celery processes specially. This should help to avoid deploy-related downtime that causes queue backups.

Service restart on error is no longer necessary since no services have been stopped at that point.

##### Environments Affected
All.

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -311,17 +311,6 @@ def _setup_release(keep_days=2, full_cluster=True):
     print(blue(env.code_root))
 
 
-def conditionally_stop_pillows_and_celery_during_migrate():
-    """
-    Conditionally stops pillows and celery if any migrations exist
-    """
-    if all(execute(db.migrations_exist).values()):
-        execute_with_timing(supervisor.stop_pillows)
-        execute(db.set_in_progress_flag)
-        execute_with_timing(supervisor.stop_celery_tasks)
-    execute_with_timing(db.migrate)
-
-
 def deploy_checkpoint(command_index, command_name, fn, *args, **kwargs):
     """
     Stores fabric env in redis and then runs the function if it shouldn't be skipped
@@ -344,8 +333,6 @@ def _deploy_without_asking(skip_record):
             "fab {environment} deploy:resume=yes.".format(environment=env.env_name),
             traceback_string()
         )
-        # hopefully bring the server back to life
-        silent_services_restart()
         raise
     else:
         execute(check_servers.perform_system_checks)
@@ -540,7 +527,7 @@ ONLINE_DEPLOY_COMMANDS = [
     staticfiles.collectstatic,
     staticfiles.compress,
     staticfiles.update_translations,
-    conditionally_stop_pillows_and_celery_during_migrate,
+    db.migrate,
     db.create_kafka_topics,
     staticfiles.pull_manifest,
     staticfiles.pull_staticfiles_cache,
--- a/src/commcare_cloud/fab/operations/db.py
+++ b/src/commcare_cloud/fab/operations/db.py
@@ -50,24 +50,6 @@ def set_in_progress_flag(use_current_release=False):
 
 
 @roles(ROLES_DEPLOY)
-def migrations_exist():
-    """
-    Check if there exists database migrations to run
-    """
-    with cd(env.code_root):
-        result = sudo(f'{env.virtualenv_root}/bin/python manage.py showmigrations | grep "\\[ ]" | wc -l')
-        try:
-            # This command usually returns some logging and then then number of migrations
-            result = result.splitlines()
-            n_migrations = int(result[-1])
-        except Exception:
-            # If we fail on this, return True to be safe. It's most likely cause we lost connection and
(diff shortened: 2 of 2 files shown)


src/commcare_cloud/fab/fabfile.py
```diff
@@ -344,8 +333,6 @@ def _deploy_without_asking(skip_record):
             "fab {environment} deploy:resume=yes.".format(environment=env.env_name),
             traceback_string()
         )
-        # hopefully bring the server back to life
-        silent_services_restart()
```
calellowitz, line 348: why is this not needed anymore?

src/commcare_cloud/fab/fabfile.py
```diff
@@ -311,17 +311,6 @@ def _setup_release(keep_days=2, full_cluster=True):
     print(blue(env.code_root))
 
 
-def conditionally_stop_pillows_and_celery_during_migrate():
-    """
-    Conditionally stops pillows and celery if any migrations exist
-    """
-    if all(execute(db.migrations_exist).values()):
-        execute_with_timing(supervisor.stop_pillows)
-        execute(db.set_in_progress_flag)
```
calellowitz, line 320: Do we need to also change the place where this flag gets unset?

src/commcare_cloud/fab/fabfile.py
millerdev: > Service restart on error is no longer necessary since no services have been stopped at that point.

src/commcare_cloud/fab/fabfile.py
millerdev: It doesn't get unset. It sets a redis key with an automatic timeout of 5 minutes. This same management command is also run later before the final service restart. It prevents service checks from failing (by ignoring failures) during that 5-minute period.

https://github.com/dimagi/commcare-cloud/blob/bd5ee374de19108520829d074aa6580d740466a5/src/commcare_cloud/fab/fabfile.py#L475

dannyroberts:
I'm not against this, but I want to share more about my understanding of why we had this there in the first place, to make sure it's part of the consideration.

I believe the concern was a quantitative rather than qualitative one. That is, we stopped celery and pillows out of a belief (or possibly from a particular experience, though it was so long ago that I'm not sure I can point to it specifically) that the less total concurrent load on the postgres databases there was, the less likely it was that a migration would meet with any kind of contention that would significantly increase the time it took to apply. While there may also be some qualitative differences between the type of things that happen on webworkers vs celery and pillowtop (maybe more bulk update in the latter, for example), the main reason we stopped celery and pillowtop but not webworkers is that we could do so without any system downtime, and that a 1-5m delay in asynchronous processing was well within the acceptable range, but 1-5m downtime of web requests would of course be unacceptable. By shutting off what we acceptably could while migrations run, the idea was just to keep concurrent activity at the minimum acceptable level, in order to give migrations the best environment in which they could be applied with the least potential complications.

I don't know whether today we still believe this is a concern or not; just wanted to make sure that that was part of the consideration.

millerdev:
@dannyroberts Thanks for the background info and thoughts. I still think it's worth making this change since we do not have any hard evidence of why the functionality is needed anymore. We can safely and easily revert if it has a negative effect.

Note that long-running migrations are typically applied outside of deploy, and celery and pillowtop processes are usually not stopped for them, so I think it will be fine to keep them running during normal deploy migrations as well.

dannyroberts (approved):
Okay, makes sense to me!

calellowitz:
> Note that long-running migrations are typically applied outside of deploy
I am still not opposed to this change, but I don't think this addresses the issue Danny flagged and I am leaning more toward the idea that we may not want this. The issue it sounds like we are trying to avoid by shutting down celery and piilows is having them block a migration, specifically by holding a connection on a table when a migration needs an exclusive lock (as most schema migrations do). It is not that the slow, out of band, migrations we are worried about, but that small ones gets blocked by pillows or celery, both of which can hold long running connections to postgres. Having less activity on postgres, and exclusively activity we expect to be short lived (as web requests are), makes these issues far less likely.

snopoke:
> > Note that long-running migrations are typically applied outside of deploy
> 
> I am still not opposed to this change, but I don't think this addresses the issue Danny flagged and I am leaning more toward the idea that we may not want this. The issue it sounds like we are trying to avoid by shutting down celery and piilows is having them block a migration, specifically by holding a connection on a table when a migration needs an exclusive lock (as most schema migrations do). It is not that the slow, out of band, migrations we are worried about, but that small ones gets blocked by pillows or celery, both of which can hold long running connections to postgres. Having less activity on postgres, and exclusively activity we expect to be short lived (as web requests are), makes these issues far less likely.

I don't think we have anything in pillows that will hold lock for a long time. The largest queries are likely the batch fetches in the case / form pillows but even those are only in the order of 100 rows per batch.

If we do happen to find anything that causes migrations to hang then we should probably address that directly.

snopoke (approved):
Approving because I think this is low risk and if there are issues we can revert.

PR #5777, opened 2023-02-15
tf_hashes for all platforms in hashicorp/aws provider

pvisweswar:
**Environments Affected :** Staging, India and Production


pvisweswar:
@dannyroberts @shyamkumarlchauhan The **terraform.lock.hcl** should have hashes for three platforms **linux, mac and windows.** When you made some modifications they generated the hashes for mac system. For  Linux and windows machines those are blocking to run terraform plan. 

In the terraform.lock.hcl file  h1 indicates hash values. So for **hashicorp/archive  and hashicorp/random** we have 3 h1 values but for hashicorp/aws  we have only 1 h1 value. So we need to create hash values for both linux and windows.

So for that I created this PR with  hash value for both linux and windows. Please review the  PR and suggest me the changes

dannyroberts (approved):

PR #5789, opened 2023-02-21
Test containers

millerdev:
Test container environment and automation script, which uses `podman-compose` for container orchestration. An alias to
`docker-compose` should work if `podman-compose` is not available.

##### Environments Affected
None.


tests/cloud/environments/test/public.yml
```diff
@@ -0,0 +1,41 @@
...
+commcare_cloud_root_user: ansible
+commcare_cloud_strict_host_key_checking: no
+commcare_cloud_use_vault: no
+commcare_cloud_ssh_args:
+- -o UserKnownHostsFile=/dev/null
+- >-
+  -o ProxyCommand=ssh
+  -o UserKnownHostsFile=/dev/null
+  -o StrictHostKeyChecking=no
+  -W %h:%p -p 2222 admin@localhost
```
dannyroberts, line 41: I think I don't understand why the first `-o UserKnownHostsFile=/dev/null` stands alone as its own list item, whereas all the other options are included together in a single list item. What's the difference?

dannyroberts (approved):
Nice!

tests/cloud/environments/test/public.yml
millerdev: There are two arguments for the commcare-cloud SSH session:
```
ssh -o UserKnownHostsFile=/dev/null -o ProxyCommand=...
```

`ProxyCommand` has a separate SSH command with its own arguments:

```
ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -W %h:%p -p 2222 admin@localhost
```
The proxy command sets up a SSH connection to the "portal" container, and connections to "worker" containers are tunneled through the portal connection.

dannyroberts: Oooohhh, okay, right. I got confused by the way the the `-o`s in `-o ProxyCommand=...` and `-o UserKnownHostFile=...`, etc. are visually aligned, whereas all but the first are conceptually nested under the first. Makes sense, thanks!

PR #5816, opened 2023-03-08
Default release_name to NO_RELEASE

millerdev:
To allow "fab" commands to run without setting a release name. The release name they would have used previously was not meaningful.

##### Environments Affected
All.

### DIFF
--- a/src/commcare_cloud/fab/fabfile.py
+++ b/src/commcare_cloud/fab/fabfile.py
@@ -102,7 +102,7 @@ def _setup_path():
     env.log_dir = posixpath.join(env.home, 'www', env.deploy_env, 'log')
     env.releases = posixpath.join(env.root, 'releases')
     env.code_current = posixpath.join(env.root, 'current')
-    env.code_root = posixpath.join(env.releases, env.release_name)
+    env.code_root = posixpath.join(env.releases, env.ccc_environment.release_name)
     env.project_root = posixpath.join(env.code_root, env.project)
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/fab/fabfile.py
```diff
@@ -118,7 +118,7 @@ def _setup_path():
 
 def load_env():
     env.ccc_environment = get_environment(env.env_name)
-    env.ccc_environment.release_name = env.release_name
+    env.ccc_environment.release_name = getattr(env, "release_name", "NO_RELEASE")
```
mkangia, line 121: isn't it confusing to have a release named "NO_RELEASE"?

millerdev: That name should never be used to refer to a real release. The confusing thing is that there are code paths that require this variable (and others that depend on it) to be set when it is not valid and should not be used.

Prior to https://github.com/dimagi/commcare-cloud/pull/5806 this would have been set to a name (a date string) with the same format as other valid release names, but it would not refer to a real release. In that scenario, code that referenced `env.code_root` or other similar variables would have referenced a release that did not exist.

mkangia: With that context, it makes sense.
I am just thinking if someone would read only this code and get confused.
possibly, `NO_NAME` would be less confusing.
(this is not a blocking feedback)

mkangia (approved):

PR #5820, opened 2023-03-09
Ansible deploy HQ: part 3 - remaining setup release tasks

millerdev:
The `deploy` command argument `--setup-release` has been renamed to `--private`. Private releases are intended for running management commands, and therefore do not include static JavaScript files.

`preindex-views` now runs the management command in a tmux session rather than kicking it off in the background and returning immediately.

Tested locally in containers and with automated tests. Also tested on staging:
- private release (in which `./manage.py check_services` was run)
- `preindex-views`
- a full deploy.

:blowfish: Review by commit.

##### Environments Affected
All.

### DIFF
--- a/src/commcare_cloud/ansible/roles/deploy_hq/tasks/copy_release_files.yml
+++ b/src/commcare_cloud/ansible/roles/deploy_hq/tasks/copy_release_files.yml
@@ -1,0 +1,19 @@
+- name: Check if {{ item }} directory exists in previous release
+  stat:
+    path: '{{ code_home }}/{{ item }}'
+  register: src_directory
+
+- name: Create {{ item }} directory
+  file:
+    path: "{{ code_source }}/{{ item }}"
+    state: directory
+    mode: 0755
+
+- name: Copy {{ item }} from previous release
+  copy:
+    src: "{{ code_home }}/{{ item }}/"
+    dest: "{{ code_source }}/{{ item }}/"
+    remote_src: true
+    local_follow: false
+    mode: preserve
+  when: src_directory.stat.exists
--- a/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
+++ b/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
@@ -36,3 +36,10 @@
   loop:
     - node_modules
     - bower_components
+
+- name: Mark keep until
+  file:
+    path: "{{ code_source }}/KEEP_UNTIL__{{ keep_until }}"
+    mode: 0644
+    state: touch
+  when: keep_until | default(None)
--- a/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
+++ b/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
@@ -1,3 +1,11 @@
+# Negated tag checks for "private_release" are used instead of normal
+# tags because the "private_release" tag is inherited from deploy_hq.yml
+# to all tasks in this file. Tag inheritance is required because of the
+# "include_role" task in main.yml. In other words, the tasks in this
+# file would not run if they were tagged "private_release" unless the
+# "include_role" statement that loads them was also tagged, and that is
+# not desirable because main.yml is generic.
+
 - name: Clone source code
   git_setup_release:
     repo: "{{ commcarehq_repository }}"
@@ -14,3 +22,17 @@
     python_version: "{{ python_version }}"
     requirements_file: "requirements/prod-requirements.txt"
     http_proxy: "{{ http_proxy | default(omit, true) }}"
+
+- name: Copy localsettings.py
+  copy:
+    src: "{{ code_home }}/localsettings.py"
+    dest: "{{ code_source }}/localsettings.py"
+    remote_src: true
+    mode: preserve
+
+- name: Copy release files
+  include_tasks: copy_release_files.yml
+  when: '"private_release" not in ansible_run_tags'
+  loop:
+    - node_modules
+    - bower_components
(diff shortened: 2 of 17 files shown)


src/commcare_cloud/commands/deploy/commcare.py
```diff
@@ -44,24 +45,27 @@ def deploy_commcare(environment, args, unknown_args):
...
-            else:
-                exit("--limit only supports 'django_manage' at this time")
-        else:
-            fab_command = "setup_release"
+    if args.private and not args.keep_days:
+        args.keep_days = 1
+    if args.keep_days:
+        until = (datetime.utcnow() + timedelta(days=args.keep_days)).strftime(DATE_FMT)
+        ansible_args.extend(["-e", f"keep_until={until}"])
+    if args.private:
```
snopoke, line 59: Just to confirm my understanding, if no `--limit` arg is provided this will create a release on all machines right? What do you think about defaulting the limit to 'django_manage' which is the common use case. If you want the release on all machines you must provide `--limit=[all|*]` (https://docs.ansible.com/ansible/latest/inventory_guide/intro_patterns.html#common-patterns)

snopoke (approved):
Some general comments but nothing blocking.

src/commcare_cloud/ansible/roles/deploy_hq/tasks/copy_release_files.yml
```diff
@@ -0,0 +1,19 @@
+- name: Check if {{ item }} directory exists in previous release
```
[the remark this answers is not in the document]
millerdev: This script is used in two places.

https://github.com/dimagi/commcare-cloud/blob/6f145b197f22972f850f71913a54d6a436647e9f/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml#L33-L38

https://github.com/dimagi/commcare-cloud/blob/6f145b197f22972f850f71913a54d6a436647e9f/src/commcare_cloud/ansible/deploy_hq.yml#L21-L24

src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
```diff
@@ -14,3 +22,53 @@
     python_version: "{{ python_version }}"
     requirements_file: "requirements/prod-requirements.txt"
     http_proxy: "{{ http_proxy | default(omit, true) }}"
+
+- name: Copy localsettings.py
```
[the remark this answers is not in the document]
millerdev: That's a reasonable idea. I'd like to do that in a separate PR.

src/commcare_cloud/commands/deploy/commcare.py
millerdev: I like that idea.

src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
```diff
@@ -14,3 +22,53 @@
...
+  when: '"private_release" not in ansible_run_tags'
+  run_once: true
+
+- name: Unsafe checkpoints error
+  fail:
+    msg: >-
+      Deploy failed, likely because kafka checkpoints were not available.
+      You can rerun with --ignore-kafka-checkpoint-warning to prevent this
+      error from blocking the deploy.
+  when: '"private_release" not in ansible_run_tags and checkpoints_result.rc'
```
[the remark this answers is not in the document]
millerdev: https://github.com/dimagi/commcare-cloud/pull/5820/commits/0bf15b814fc43638d2c64730946c785ee1c7a41d

src/commcare_cloud/commands/deploy/commcare.py
millerdev: https://github.com/dimagi/commcare-cloud/pull/5820/commits/c8ce50e706b38a4be35147c8cbce44a5b132ff40

snopoke (approved):

src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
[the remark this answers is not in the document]
millerdev: https://github.com/dimagi/commcare-cloud/pull/5824

src/commcare_cloud/fab/fabfile.py
```diff
@@ -240,50 +236,16 @@ def rollback_formplayer():
...
-
-@incomplete_task("deploy commcare --setup-release ...")
-def setup_release(keep_days=0):
-    """OBSOLETE. Use deploy commcare --setup-release
-                                     [--keep-days=N] [--commcare-rev=HQ_BRANCH]
-    """
-    _setup_release(parse_int_or_exit(keep_days), full_cluster=True)
-
-
-def _setup_release(keep_days=2, full_cluster=True):
```
gherceg, line 267: @millerdev do you happen to remember your thoughts on this default of keeping releases for 2 days when switching over to the ansible deploy? From what I can tell, there is no default for keep_days anymore unless [it is a private release](https://github.com/dimagi/commcare-cloud/blob/be8fbb8add1c32e6af77048a29b31648901f3693/src/commcare_cloud/commands/deploy/commcare.py#L49-L53). My understanding is that the default of 2 days was actually [important](https://github.com/dimagi/commcare-cloud/pull/4206#discussion_r494327913) to ensure any stale celery processes still have access to the release they are running on.

millerdev: I don't remember clearly, but I think I concluded that it was unnecessary. A typical deploy remains active for a week, so the `KEEP_UNTIL` date assigned by `keep_days=2` would be long past by the time a release usually becomes eligible for clean up.

The [`clean_releases` module](https://github.com/dimagi/commcare-cloud/blob/c9a7003050ebf9b668543ed45d0d05661e306dad/src/commcare_cloud/ansible/library/clean_releases.py) retains the most recent 2 releases in addition to the current release. Unless there are more than 3 deploys in a two-day period, any given release should be kept around for plenty of time beyond when it is no longer current.

Is this inquiry about the edge case where there are more than 3 deploys in a two-day period? I don't think a default `keep_days=2` addresses that scenario fully since the resulting `KEEP_UNTIL` date on a release could be past by the time the release is _demoted from current_. I think the time a deploy is demoted from current is the point at which we want to start the timer for celery tasks. I don't think we record that date currently, but we could do that and update `clean_releases` to retain any releases for some period of time beyond that date.

PR #5822, opened 2023-03-13
reduce ram for es heap

calellowitz:
https://dimagi-dev.atlassian.net/browse/SAAS-14151

ES timeouts are linked to large and slow GC events. While the long term plan to reshard the case index still seems like the best way to prevent these memory spikes, ES [recommends](https://www.elastic.co/blog/a-heap-of-trouble) reducing heap size as a way to help keep GC events smaller, and heap usage on prod nodes never exceeds 24GB. Switching to G1GC may also help but ES does not recommend that until a higher version of the JDK than we currently run ES on, so it may be something to look into after the upgrade.

##### Environments Affected

Production (could be all, but that is the only one with machines large enough to be affected)

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml
@@ -1,8 +1,8 @@
 ---
 elasticsearch_version: 2.4.6
 # https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html#_give_half_your_memory_to_lucene
-# Either half the machines RAM or 30 GB. Never want to go over 30 GB
-elasticsearch_memory: "{{ [ansible_memory_mb.real.total // 2, 30000] | min }}m"
+# Either half the machines RAM or 24 GB. If changing in the future, never want to go over 30 GB
+elasticsearch_memory: "{{ [ansible_memory_mb.real.total // 2, 24576] | min }}m"
 elasticsearch_download_sha256: 5f7e4bb792917bb7ffc2a5f612dfec87416d54563f795d6a70637befef4cfc6f.
 elasticsearch_home: "/opt/elasticsearch-{{ elasticsearch_version }}"
 elasticsearch_conf_dir: "/etc/elasticsearch-{{ elasticsearch_version }}"


dannyroberts (approved):

src/commcare_cloud/ansible/roles/elasticsearch/defaults/main.yml
```diff
@@ -1,8 +1,8 @@
 ---
 elasticsearch_version: 2.4.6
 # https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html#_give_half_your_memory_to_lucene
-# Either half the machines RAM or 30 GB. Never want to go over 30 GB
-elasticsearch_memory: "{{ [ansible_memory_mb.real.total // 2, 30000] | min }}m"
+# Either half the machines RAM or 24 GB. If changing in the future, never want to go over 30 GB
+elasticsearch_memory: "{{ [ansible_memory_mb.real.total // 2, 24576] | min }}m"
```
millerdev, line 5: Why 24GB? Is that custom tuning for our production cluster, or is it a generic limit that makes sense on all clusters? If it's specific to production, maybe it makes sense to make this setting parameterized per environment?

dannyroberts: I'll let @calellowitz confirm, but my understanding of the rationale is basically (1) our production instance is the biggest ES cluster of any CommCare instance so if it's okay for us to use 24GB heap, it's probably okay for anyone (this is a bit handwavy about different possible configurations, etc, but most third party instances are monoliths where presumably you'd already want ES to take up less space) and (2) if we're running into this GC issue, then other people who got to somewhere near our size and usage might also run into it, so they'd benefit from lowering the heap size for the same reason as we do. I don't feel super strongly about it, but I think that rationale makes reasonable sense.

calellowitz: Yeah, basically what Danny said. That we believe this to be the largest cluster by far, and that this is an easy mistake to make but a hard one to find, so if we can use defaults to prevent people from over committing, that might be helpful (and this whole setting can still be set per environment in the public yaml  files). I don't feel particularly strongly though, so if you do, happy to change it.

millerdev: Thanks. Sounds fine if it can be overridden per env as needed.

PR #5824, opened 2023-03-13
Generate localsettings.py from template on each deploy

millerdev:
As suggested in PR feedback https://github.com/dimagi/commcare-cloud/pull/5820#discussion_r1132165786

- `update-config` is no longer required to update `localsettings.py`. However, the existing `update-config` workflow can still be used to preview and check the validity of changes to `localsettings.py` if desired.
- As implemented with [`no_log: true`](https://github.com/dimagi/commcare-cloud/blob/595cb85504d60b3a527a405d192784c4f9023bdf/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml#L25), `localsettings.py` content is not printed on the console, reducing the risk of copy/paste exposing sensitive information.

##### Environments Affected
All.

### DIFF
--- a/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
+++ b/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
@@ -15,12 +15,14 @@
     requirements_file: "requirements/prod-requirements.txt"
     http_proxy: "{{ http_proxy | default(omit, true) }}"
 
-- name: Copy localsettings.py
-  copy:
-    src: "{{ code_home }}/localsettings.py"
+- name: Generate localsettings.py
+  template:
+    src: roles/commcarehq/templates/localsettings.py.j2
     dest: "{{ code_source }}/localsettings.py"
-    remote_src: true
-    mode: preserve
+    owner: "{{ cchq_user }}"
+    group: "{{ cchq_user }}"
+    mode: 0600
+  no_log: true
 
 - name: Mark keep until
   file:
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml
```diff
@@ -15,12 +15,14 @@
...
+- name: Generate localsettings.py
+  template:
+    src: roles/commcarehq/templates/localsettings.py.j2
     dest: "{{ code_source }}/localsettings.py"
-    remote_src: true
-    mode: preserve
+    owner: "{{ cchq_user }}"
+    group: "{{ cchq_user }}"
+    mode: 0600
+  no_log: true
```
snopoke, line 25: :+1:

snopoke (approved):

snopoke:
This should probably be communicated out quite clearly (internally & externally) since it is a significant change.

millerdev:
I'm leaning away from merging this, particularly because

> [`no_log: true`](https://github.com/dimagi/commcare-cloud/blob/595cb85504d60b3a527a405d192784c4f9023bdf/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml#L25), `localsettings.py` content is not printed on the console

It seems less than ideal to not know what changed if there is a difference between localsettings in the current and new releases. With another task or two I could show the diff (if there is one), but that's still not quite as good as the current `update-code` workflow, which allows a review before apply. Input is welcome.

snopoke:
> I'm leaning away from merging this, particularly because
> 
> > [`no_log: true`](https://github.com/dimagi/commcare-cloud/blob/595cb85504d60b3a527a405d192784c4f9023bdf/src/commcare_cloud/ansible/roles/deploy_hq/tasks/setup_release.yml#L25), `localsettings.py` content is not printed on the console
> 
> It seems less than ideal to not know what changed if there is a difference between localsettings in the current and new releases. With another task or two I could show the diff (if there is one), but that's still not quite as good as the current `update-code` workflow, which allows a review before apply. Input is welcome.

That's fine. Rather leave it as is for now then

PR #5826, opened 2023-03-13
Make changes necessary to build webworkers on both Bionic and Jammy

dannyroberts:

I have tested this on both Bionic and Jammy and it runs through clean from scratch from the beginning to the end on both.

##### Environments Affected
all environments including third parties (but there should be no change in behavior)


dannyroberts:
Okay I worked through the following feedback from you @skodde:

- [x] is https://github.com/dimagi/commcare-cloud/commit/5e11f0ebe5eabbc12eb7498cb767d157082f78a0 needed at all? Can we remove the apt-install of `python-psycopg2` entirely?
- [x] use ansible_distribution_release in apt repo tasks
- [x] simplify https://github.com/dimagi/commcare-cloud/commit/95c0bdf90efc1a7775a095c85d4ab8654380ff35 to just always install `pkg-config`
- [x] no need to specify `become_user` in https://github.com/dimagi/commcare-cloud/commit/218ba1166ff32a41a1519165069a911ab11a732b

I also went through with the rest of the process to integrate these webworkers (running a deploy) and now they're receiving traffic on staging like the others, so ✔️

src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
```diff
@@ -112,9 +106,25 @@
...
+      - ufw
   when: ansible_distribution_version == '18.04'
 
+- name: Install common packages (Jammy)
+  become: yes
+  apt:
+    name:
+      - acl
+      - software-properties-common
+      - ufw
```
skodde, line 125: Can we remove already ufw at least on 22.04?

skodde (approved):

src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
dannyroberts: To me that's a separate question from the Ubuntu upgrade. It's also a bigger change than I'd want to sneak into a PR like this—to the extent that any third party relies at all on this configuration to manage their port accessibility (which some almost certainly do) this is a change we would need to have very clear communication about ahead of time.

src/commcare_cloud/ansible/deploy_commcarehq.yml
```diff
@@ -110,11 +110,14 @@
           - "{{ mgmt_cmd_managed_files|default([]) }}"
           - "{{ prometheus_managed_files|default([]) }}"
           - "{{ prometheus_django_runner_managed_files|default([]) }}"
-    - shell: find {{ service_home|quote }} -type f
+    - name: "Find supervisor config files"
+      become: true
```
millerdev, line 114: The commit message says "Run some tasks as cchq..." but this appears to be running as root. Is `become_user: '{{ cchq_user }}'` specified somewhere that I missed?

The same applies to the next task as well.

src/commcare_cloud/ansible/roles/deploy_hq/tasks/main.yml
```diff
@@ -1,4 +1,5 @@
 - name: 'Check if done: {{ if_not_done }}'
+  become: true
```
millerdev, line 2: This should also have `become_user: '{{ cchq_user }}'`

src/commcare_cloud/ansible/deploy_commcarehq.yml
dannyroberts: Ah, yes. This was a git commit cleanup/interactive rebase error. I originally did do it as `cchq`, marco pointed out that it wasn't necessary to run as cchq specifically, and so I made a change and melded it into the same commit. The code ended up how I intended it, but the commit message ended up not correctly describing it.

PR #5830, opened 2023-03-14
Make changes necessary to build proxy machines on both Bionic and Jammy

dannyroberts:

##### Environments Affected
all environments including third parties (but there should be no change in behavior)


dannyroberts:
I'm still doing a final run through of the ansible code (both Jammy and Bionic), after which I'll do another test of update-config + deploy, which failed during the deploy on my previous attempt.

dannyroberts:
Okay this ran all the way through, start to finish, including the deploy, with no issues. The two new proxy machines are now part of the staging cluster and functioning.

dannyroberts:
(Now I'm going to do some rebasing and cleanup)

skodde (approved):

src/commcare_cloud/ansible/deploy_logstash.yml
```diff
@@ -22,6 +22,17 @@
         name: net.core.rmem_max
         value: '26214400'
         state: present
+    - name: 'Accommodate poorly functioning task within sansible.users_and_groups'
+      # Task 'Enable include of /etc/sudoers.d'
+      # Forces the line '#includedir /etc/sudoers.d' to be included in /etc/sudoers.
+      # The up-to-date syntax is @includedir, and it's in the default file that comes with Jammy
+      # but since there's no simple way to get sansible *not* to add its version of this line,
+      # I'm removing the more correct line so that the dir is not included twice.
```
millerdev, line 30: Would it work to remove the less correct line?

dannyroberts: I thought of that, but decided not to because it would be like unstably idempotent, in that every time you ran the playbook (to update a machine for instance) it would first add back in the `#includedir` line and then remove it again here.

millerdev: Sounds reasonable, at least for now. What would it take for us to fix this more correctly? For example, could we submit a fix to "sansible"? (I'm not even sure what that is.)

dannyroberts: Yeah I think we'd have to submit a fix to `sansible.logstash` (or really the dependency it calls, `sansible.users_and_groups`). (`sansible` is just someone's username on ansible galaxy or something like that.) It's the module we use to install logstash which is used for something which I can't quite remember on just the proxy machines. Possibly for reporting stats form the nginx logs to datadog?

PR #5845, opened 2023-03-20
Ansible deploy HQ: part 5 - clean releases

millerdev:
Tested locally and on staging. The new `clean_releases` Ansible module is also covered by automated tests.

:blowfish: Review by commit.

##### Environments Affected
All.

### DIFF
--- a/src/commcare_cloud/ansible/deploy_hq.yml
+++ b/src/commcare_cloud/ansible/deploy_hq.yml
@@ -9,6 +9,7 @@
     - pillowtop
     - proxy
     - django_manage
+  any_errors_fatal: true
   tasks:
     - name: Get code version
       include_role:
@@ -28,6 +29,7 @@
 
 - name: Copy static JS
   hosts: proxy
+  any_errors_fatal: true
   tasks:
     - include_role: {name: deploy_hq}
       vars:
@@ -36,6 +38,7 @@
 
 - name: Install javascript dependencies
   hosts: [webworkers, celery, proxy]
+  any_errors_fatal: true
   tasks:
     - import_role: {name: deploy_hq}
       vars:
@@ -43,6 +46,7 @@
 
 - name: Collect static files
   hosts: [webworkers, proxy]
+  any_errors_fatal: true
   tasks:
     - import_role: {name: deploy_hq}
       vars:
@@ -50,6 +54,7 @@
 
 - name: Compress static files
   hosts: proxy[0]
+  any_errors_fatal: true
   tasks:
     - import_role: {name: deploy_hq}
       vars:
@@ -62,6 +67,7 @@
     - pillowtop
     - proxy
     - django_manage
+  any_errors_fatal: true
   tasks:
     - import_role: {name: deploy_hq}
       vars:
@@ -69,6 +75,7 @@
 
 - name: Run migrations
   hosts: webworkers[0]
+  any_errors_fatal: true
   tasks:
     - import_role: {name: deploy_hq}
       vars:
@@ -76,6 +83,7 @@
 
 - name: Pull staticfiles manifest
   hosts: webworkers
+  any_errors_fatal: true
   tasks:
     - import_role: {name: deploy_hq}
       vars:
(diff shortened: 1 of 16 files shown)


src/commcare_cloud/commands/deploy/commcare.py
```diff
@@ -88,8 +88,9 @@ def deploy_commcare(environment, args, unknown_args):
             '--set', ','.join(fab_settings), branch=args.branch, *unknown_args
         )
     if rc != 0:
+        resume_option = color_success(f"--resume={environment.release_name}")
```
dannyroberts, line 91: nit: success doesn't feel like the right semantic category to me. Maybe `color_notice`?

src/commcare_cloud/ansible/deploy_hq.yml
```diff
@@ -9,6 +9,7 @@
     - pillowtop
     - proxy
     - django_manage
+  any_errors_fatal: true
```
dannyroberts, line 12: I'm surprised this is needed; I thought this was the default behavior in ansible. I thought if any task failed (unless it's specifically handled) the ansible run would end. But the documentation for `any_errors_fatal` matches what you're saying (https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#aborting-on-the-first-error-any-errors-fatal), so now I'm confused as to what the default behavior is.

dannyroberts (approved):

src/commcare_cloud/commands/deploy/commcare.py
millerdev: I was going for green, as in

"Deploy failed." (red, stop)  
"Add --resume=... to retry." (green, go)

I could use `clint.textui.colored.green` instead of `color_success`. I considered `color_notice`, but yellow often looks pretty washed out by default for folks who don't use a dark background.

Alternately, I could make it red. That would also stand out nicely.

src/commcare_cloud/ansible/deploy_hq.yml
snopoke: The default behaviour is: "Ansible continues to execute tasks as long as there are hosts that have not yet failed" (https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#setting-a-maximum-failure-percentage)

So if a task fails on a host then it will be ignored for the rest of the playbook but other hosts will continue to have tasks executed on them.

snopoke (approved):

src/commcare_cloud/ansible/deploy_hq.yml
millerdev: Yes, from the top of the [page you linked](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_error_handling.html#error-handling-in-playbooks):

> When Ansible receives a non-zero return code from a command or a failure from a module, by default it stops executing on that host and continues on other hosts.

Although [loops appear to be an exception](https://github.com/dimagi/commcare-cloud/blob/5118f0aa33d13e020c86ce02ade273e2928e8f3c/src/commcare_cloud/ansible/roles/deploy_hq/tasks/staticfiles_collect.yml#L21-L26). Execution does not stop until the loop is exhausted.

I wish there was a way to set `any_errors_fatal: true` globally when running a playbook rather than adding it to every play, but I haven't found a way to do that.

src/commcare_cloud/commands/deploy/commcare.py
dannyroberts: 🤷 I think maybe if that's the problem (`color_notice` looks bad on a light background) we should change the color scheme. That must be a general annoyance for you anywhere `color_notice` is used. I'm not sure, but maybe there's a way to detect the terminal background color, and have one palate for a light background and another one for a dark background. Just throwing out ideas. Obviously very tangential to the purpose of this PR

millerdev: `color_notice` is fine. https://github.com/dimagi/commcare-cloud/pull/5849/commits/2e4d8d9c87406037612d342489be263457d37a2e

I can reconfigure my terminal to use a color that is more visible than the default yellow color (actually, I had forgotten that I did this long ago). It surprises me that the default yellow color is so bad against white. Maybe I'm the only person who doesn't like dark mode?

PR #5847, opened 2023-03-21
Do not print changelogs alert if there are none

millerdev:
https://github.com/dimagi/commcare-cloud/pull/5630#pullrequestreview-1342336933

##### Environments Affected
All.


gherceg (approved):
Looks good. One out of scope question.

The difference between the `changelogs.error` message and the default `changelogs` message aren't very clear to me. It seems like either way, there are changelogs that need to be applied before deploying? Additionally, it seems worth linking to https://commcare-cloud.readthedocs.io/en/latest/changelog/index.html directly for _both_ messages.

millerdev:
^ @sravfeyn please note feedback

PR #5852, opened 2023-03-23
Ansible deploy HQ: part 8 - rollback deploy and service restart commands

millerdev:
`fab rollback` was probably not a commonly used command, and the extra functionality it provided did not seem worth implementing. The replacement (`list-releases` followed by `deploy --resume=RELEASE_NAME`) leaves the choice of which release to "resume" up to the operator.

Tested locally and on staging.

:blowfish: Review by commit.

##### Environments Affected
All.


src/commcare_cloud/commands/ansible/run_module.py
```diff
@@ -296,3 +296,40 @@ def run(self, args, unknown_args):
             become_user='cchq',
             extra_args=['-e', f'@{group_vars}'] + unknown_args,
         )
+
+
+class ListReleases(CommandBase):
+    command = 'list-releases'
+    help = 'List names that can be passed to `deploy --resume=RELEASE_NAME`'
```
snopoke, line 303: would be nice if this filtered out ones that were 'complete' (or perhaps just included that info in the list).

snopoke (approved):

src/commcare_cloud/commands/ansible/run_module.py
millerdev: :+1: I'll implement that in a follow-up PR.

FWIW, any "incomplete" build should build to completion on `--resume`, and if you're watching the terminal output you should have time to cancel if a particular resume starts doing some unexpected build operations. Any step that "changes" prior to "Check release" would indicate the build had not previously completed. But granted, this requires intimate knowledge of the build, so not a reasonable expectation for everyone.

millerdev: https://github.com/dimagi/commcare-cloud/pull/5854

src/commcare_cloud/ansible/restart_webworker_services.yml
```diff
@@ -0,0 +1,29 @@
+- name: Rolling restart webworker services
+  hosts: webworkers
+  serial: ["10%", "40%", "100%"]
```
dannyroberts, line 3: Just writing it out to make sure I understand. From https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_strategies.html it sounds like this means "Do 10%, then another 40%, then the rest (50%)" and not "Do up to 10%, then up to 40% (so 30% in this batch), and then up to 100% (so 60% in this batch)". This is my first time seeing this syntax, but it seems great!

millerdev: Yeah, I think it's 10% followed by the next 40% (cumulative 50%), then the remaining 50%. We can easily change it to use smaller percentages if this turns out to be too aggressive.

dannyroberts: I bet this is fine. If 50% down at a time turns out to strain the remaining webworkers during this period, then the alternative I'd suggest (if it comes to that) would be `["10%", "30%"]` (which I believe from those docs would do 10% in the first batch, and then keep doing 30% until they've all been restarted). This way we get the 10% canary, and then restart the remaining in three equal parts such that at any given time at least 70% of webworkers are up.

snopoke: I don't think the 10% can be considered a canary unless we are actively checking their status before rolling out the rest (just mentioning).

millerdev: It's a very basic check where the restart is expected to come back with a zero return code. If the restart fails at that level then the Ansible task will fail and the next batch will not be attempted. I think it's a reasonable if not comprehensive check. Of course there are many things that could go wrong beyond that, but checking and monitoring for every possible failure mode is beyond the scope of this task.

PR #5857, opened 2023-03-27
Update 2-manual-install.rst with python setup that works on Jammy

dannyroberts:

##### Environments Affected
None

### DIFF
--- a/hosting_docs/source/installation/2-manual-install.rst
+++ b/hosting_docs/source/installation/2-manual-install.rst
@@ -147,7 +147,7 @@ our example cluster, the control machine is named “control1”.
     ::
 
         $ sudo apt update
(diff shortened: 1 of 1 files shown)


hosting_docs/source/installation/2-manual-install.rst
```diff
@@ -163,23 +163,9 @@ our example cluster, the control machine is named “control1”.
 
         $ sudo update-alternatives --install /usr/bin/python python /usr/bin/python3 10
 
-5.  Now update pip; you might encounter installation issues otherwise.
```
dannyroberts, line 166: Jammy didn't like these steps; it printed out a warning about not running pip as root or you'll get permissions issues, and then later on there were permissions issues.

skodde (approved):

PR #5861, opened 2023-03-28
Hide git mirrors directory

millerdev:
Reduce the number of places that need to know about the "git mirrors" directory:
- `clean_releases` and `list-releases` ignore hidden directories.
- People looking in the releases directory will be less likely to be confused because `.git_mirrors` is hidden from commands like `ls` by default.

This will have the side effect of breaking the git repositories in existing releases when `clean_releases` is run the first time after this is merged (it will delete the old `git_mirrors` directory). The only instance where this would likely matter is in a private release where someone was using git to update the release to a new version of the code. Probably the easiest  workaround there is to create a new private release and discard the old one. However, the following command can be used if a release gets broken that way and it becomes necessary to (temporarily, until the next `clean_releases` run) rehabilitate its git repo:

    sudo -u cchq ln -s .git_mirrors /home/cchq/www/<env>/releases/git_mirrors

##### Environments Affected
All.

### DIFF
--- a/src/commcare_cloud/ansible/library/clean_releases.py
+++ b/src/commcare_cloud/ansible/library/clean_releases.py
@@ -60,7 +60,6 @@ EXAMPLES = """
   clean_releases:
     path: "/path/to/releases"
     exclude:
-      - git_mirrors
       - 2023-03-07_13.16
 """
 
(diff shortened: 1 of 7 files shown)


snopoke (approved):
👍🏻

PR #5862, opened 2023-03-28
Improve safety of .generated.yml

millerdev:
- Do not add `release_name` when it is not needed.
- Delete `.generated.yml` after running Ansible. It's dangerous to let generated files like that lay around on the filesystem. For example, incorrect/outdated values could be used if a command erroneously did not generate it, but a copy was left from a previous command.

Set `COMMCARE_CLOUD_KEEP_GENERATED_YML=1` to keep `.generated.yml` after Ansible terminates. This might be handy when iterating on Ansible commands that depend on generated variables.

:blowfish: Review by commit.

##### Environments Affected
All.

### DIFF
--- a/src/commcare_cloud/environment/main.py
+++ b/src/commcare_cloud/environment/main.py
@@ -5,6 +5,7 @@ import re
 import shlex
 import sys
 from collections import Counter
+from contextlib import contextmanager
 from io import open
 
 import yaml
@@ -63,7 +64,7 @@ class Environment(object):
             self.fab_settings_config
             self.postgresql_config
             self.proxy_config
-        self.create_generated_yml()
+        self._create_generated_yml(save=False)
 
     def check_known_hosts(self):
         if not os.path.exists(self.paths.known_hosts):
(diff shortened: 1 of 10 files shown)


millerdev:
Tests [passed on jammy, cancelled on bionic](https://github.com/dimagi/commcare-cloud/actions/runs/4543676572/attempts/1). I tried to re-run them today, but it appears to be stalled on something. In any case, I don't think there is much worry about bionic if they passed on jammy.

src/commcare_cloud/environment/main.py
```diff
@@ -63,7 +64,7 @@ def check(self):
             self.fab_settings_config
             self.postgresql_config
             self.proxy_config
-        self.create_generated_yml()
+        self._create_generated_yml(save=False)
```
gherceg, line 67: If we aren't saving the generated.yml file, why is it necessary to call this?

gherceg (approved):
Pair reviewed with @AmitPhulera. Left a clarifying question, but besides that looks good! 

We were also curious if you could elaborate more on what data the `.generated.yml` file contains.

src/commcare_cloud/environment/main.py
millerdev: I think this is just a check to ensure there are no errors when generating variables.

millerdev:
`.generated.yml` is used to pass values from the `Environment` instance to Ansible. Some values are included unconditionally, and others are only included when certain conditions are met.

https://github.com/dimagi/commcare-cloud/blob/2bb8d4bdd71ad8efc3c150c8fc03eeb5591ec47d/src/commcare_cloud/environment/main.py#L369-L404

PR #5863, opened 2023-03-28
remove ubuntu 18.04 from github runner tests

shyamkumarlchauhan:
https://github.com/actions/runner-images/issues/6002
Ubuntu 18.04 action runner image is deprecated from github

##### Environments Affected
All

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,7 +16,7 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):
For context, this image has a blackout today per github's schedule, and then on Apr 1 it will be permanently unsupported. So we can revert this for a couple days starting tomorrow if we really want, but then starting Saturday it'll be broken forever, so we can probably just keep it this way now.

PR #5867, opened 2023-03-28
Remove the addition of smspro intermediate cert

skodde:
smspro.mtn.ci seems to have finally corrected their TLS configuration,
so we can remove the task that added the missing intermediate
certificate to the certifi CA bundle.

See also #4609

https://dimagi-dev.atlassian.net/browse/SAAS-12017

Note that this change won't actively remove the intermediate certificate
where installed, and I think it's ok to leave it until the next certifi update
and/or server rebuilding, which would remove it.
This avoids added complexity and another task to then go back to and
revisit/remove.

##### Environments Affected
All


millerdev (approved):

gherceg (approved):
🙌🏻

PR #5874, opened 2023-03-30
remove old webworkers

shyamkumarlchauhan:

##### Environments Affected
production


mkangia (approved):
Syntax wise looks okay. I trust with Shyam that only redundant servers were removed.

PR #5875, opened 2023-03-30
post message to slack with diff

snopoke:
Add a slack message that includes the deploy diff.

This will save having to copy the diff output from the console into Slack. It also allows us to do better formatting for the slack message.

##### Environments Affected
prod, india


millerdev (approved):
Love it! :watch: :floppy_disk:

PR #5886, opened 2023-04-04
Add default release_name to formplayer role

millerdev:
Introduced by https://github.com/dimagi/commcare-cloud/pull/5862/commits/54c291cd175f522d130c6ba399d134b0b9041967 of https://github.com/dimagi/commcare-cloud/pull/5862 

The scope where `release_name` is needed includes the `formplayer` role anywhere `formplayer_release_dir` or `_formplayer_target_dir` are referenced, which is currently only done by the main tasks file. This fix seems better than reverting the commit that introduce the issue since it maintains a more limited scope where `release_name` is set.

Fixes error on `cchq <env> deploy-stack`:
```
TASK [formplayer: Formplayer build dir] ***********
task path: .../commcare-cloud/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml:98
fatal: [...]: FAILED! => {
    "msg": "{{ formplayer_build_dir }}/releases/{{ release_name}}: 'release_name' is undefined"
}
```

##### Environments Affected
All.

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/defaults/main.yml
@@ -11,3 +11,7 @@ formplayer_purge_time_spec: '2d'
 _should_update_formplayer_in_place: no
 formplayer_data_dir: "{{ encrypted_root }}/formplayer"
 use_monit_for_formplayer: false
+# Default release name for cases like deploy-stack where it is not
+# explicitly set. This is overridden by environment.release_name (from
+# Python via .generated.yml) on formplayer deploy.
+release_name: "{{ ansible_date_time.date }}_{{ ansible_date_time.hour }}.{{ ansible_date_time.minute }}"


shyamkumarlchauhan (approved):

src/commcare_cloud/ansible/roles/formplayer/defaults/main.yml
```diff
@@ -11,3 +11,7 @@ formplayer_purge_time_spec: '2d'
 _should_update_formplayer_in_place: no
 formplayer_data_dir: "{{ encrypted_root }}/formplayer"
 use_monit_for_formplayer: false
+# Default release name for cases like deploy-stack where it is not
+# explicitly set. This is overridden by environment.release_name (from
+# Python via .generated.yml) on formplayer deploy.
+release_name: "{{ ansible_date_time.date }}_{{ ansible_date_time.hour }}.{{ ansible_date_time.minute }}"
```
dannyroberts, line 17: Do you know if this is evaluated once and then `release_name` is "set" to that value, or if release name is set to the template string, which is re-evaluated every time it is used? The latter would be an issue since if it's used more than once in the same run, it would evaluate to a different string if it crosses a minute boundary

millerdev: From docs on [Ansible facts](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_vars_facts.html#ansible-facts):

> Because `ansible_date_time` is created and cached when Ansible gathers facts before each playbook run, it can get stale with long-running playbooks. If your playbook takes a long time to run, use the `pipe` filter (for example, `lookup('pipe', 'date +%Y-%m-%d.%H:%M:%S'))` or [`now()`](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_templating_now.html#templating-now) with a Jinja 2 template instead of `ansible_date_time`.

This is also the same syntax used by the `commcarehq` role.

https://github.com/dimagi/commcare-cloud/blob/7dd7667b7f261e45e2baf58b5492ee38134f4ed1/src/commcare_cloud/ansible/roles/commcarehq/vars/main.yml#L2

dannyroberts (approved):

PR #5896, opened 2023-04-07
ubuntu22 upgrade work for rabbitmq

shyamkumarlchauhan:

##### Environments Affected
staging


millerdev (approved):
This looks fine, but I've only been following the ubuntu22 upgrade from afar and don't have a lot of knowledge around how we use RabbitMQ. I don't really know what to look that might be problematic. However, I'm approving to unblock @shyamkumarlchauhan since this only affects staging.

environments/staging/terraform.yml
```diff
@@ -246,25 +246,25 @@ servers:
...
     volume_size: 80
     volume_encrypted: yes
     volume_type: "gp3"
     group: "rabbitmq"
-    os: bionic
+    os: jammy
 
-  - server_name: "rabbit_b1-staging"
+  - server_name: "rabbit_b2-staging"
     server_instance_type: "t3.large"
```
dannyroberts, line 260: I just noticed in reviewing these PRs that the staging rabbit instances are larger (t3.large, 80GB disk) than the india instances (t3.small, 30 GB disk) which is surprising! I'll make a note of it in our cost saving sheet, but keep an eye out for these kinds of opportunistic improvements!

PR #5898, opened 2023-04-10
ubuntu22 upgrade for rabbitmq india

shyamkumarlchauhan:

##### Environments Affected
india


dannyroberts (approved):
Do you want to take the opportunity to change anything about these machines from a cost-savings perspective? Fine if the answer's no, also long as I know you've considered it.

PR #5899, opened 2023-04-10
ubuntu22 upgrade for rabbitmq production

shyamkumarlchauhan:

##### Environments Affected
production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -239,6 +239,24 @@ servers:
     group: "rabbitmq"
     os: bionic
 
+  - server_name: "rabbit12-production"
+    server_instance_type: m5.2xlarge
+    network_tier: "db-private"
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):

dannyroberts (commented):
Anything you want to take the opportunity to change about these instances from a cost savings perspective? I had written down an idea here, and I think Viswa brought up the idea of switching over to t3a (over t3) wherever we can, but you don't have to limit yourself to those suggestions. Give it a fresh look and decide what changes you think are worth making. Volume size could be another thing to revisit, for example (or maybe 80 is still right for us).

dannyroberts (approved):

dannyroberts:
I know you fixed this right after, but remember to check the build ❌ / ✅  before merging!

PR #5903, opened 2023-04-10
Add missing dependencies to installation docs

skodde:


skodde:
> Can we remove the 3.10 specific instructions now as well?

I think we should wait until we officially drop 18.04 for that (July 1).

hosting_docs/source/installation/2-manual-install.rst
```diff
@@ -147,7 +147,7 @@ our example cluster, the control machine is named “control1”.
     ::
 
         $ sudo apt update
-        $ sudo apt install python3-pip python3-venv sshpass net-tools
+        $ sudo apt install python3-dev python3-distutils python3-venv libffi-dev sshpass net-tools
```
dannyroberts, line 150: Did you remove python3-pip because it's installed already by something else? `python3-dev` maybe?

skodde: Oops. That was actually a mistake. I'll make a new PR.

skodde: #5904

PR #5917, opened 2023-04-18
Add warning for hosts running Ubuntu != 22.04

skodde:
##### Environments Affected
All

### DIFF
--- a/src/commcare_cloud/ansible/deploy_hq.yml
+++ b/src/commcare_cloud/ansible/deploy_hq.yml
@@ -2,6 +2,12 @@
 # similarly named deploy_commcarehq.yml playbook is used to configure machines
 # to run commcare-hq, and is expected to be run less frequently than this one.
 
+- name: Check hosts Ubuntu version
+  hosts:
+    - all
+  tasks:
+    - import_role: {name: upgrade_warning}
+
 - name: Deploy CommCare HQ
   hosts:
     - webworkers
--- a/src/commcare_cloud/ansible/deploy_hq.yml
+++ b/src/commcare_cloud/ansible/deploy_hq.yml
@@ -7,6 +7,7 @@
     - all
   tasks:
     - import_role: {name: upgrade_warning}
+      tags: upgrade_warning
 
 - name: Deploy CommCare HQ
   hosts:
--- a/src/commcare_cloud/ansible/roles/upgrade_warning/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/upgrade_warning/tasks/main.yml
@@ -10,11 +10,17 @@
   ansible.builtin.group_by:
     key: ubuntu_{{ ansible_distribution_version | replace(".", "_") }}
 
+- name: Create group for non 22.04 hosts
+  add_host:
+    name: '{{ item }}'
+    groups: non_2204_hosts
+  with_inventory_hostnames:
+    - all:!ansible_skip:!ubuntu_22_04
+
 - name: Show 'distribution_version' for non 22.04 hosts
   debug:
-    msg: "{{ item }}"
-  register: warning_present
-  loop: "{% if 'ubuntu_22_04' not in group_names %}{{ warning_host_message }}{% else %}[]{% endif %}"
+    msg: "WARNING: Host {{ inventory_hostname }} is running Ubuntu {{ ansible_distribution_version }}"
+  loop: "{% if 'non_2204_hosts' in group_names %}['{{ inventory_hostname }}']{% else %}[]{% endif %}"
 
 - name: Show warning if non 22.04 hosts are present
   pause:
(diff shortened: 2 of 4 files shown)


src/commcare_cloud/ansible/deploy_hq.yml
```diff
@@ -2,6 +2,12 @@
 # similarly named deploy_commcarehq.yml playbook is used to configure machines
 # to run commcare-hq, and is expected to be run less frequently than this one.
 
+- name: Check hosts Ubuntu version
+  hosts:
+    - all
+  tasks:
+    - import_role: {name: upgrade_warning}
```
millerdev, line 9: I think we should make it possible to disable this check. One way to do that would be to use `include_role` here and add a `when: ...` condition that checks for the presence of a flag (file) on the control machine or something like that.

src/commcare_cloud/ansible/roles/upgrade_warning/tasks/main.yml
```diff
@@ -0,0 +1,34 @@
...
+
+- name: Create groups based on 'distribution_version'
+  ansible.builtin.group_by:
+    key: ubuntu_{{ ansible_distribution_version | replace(".", "_") }}
+
+- name: Show 'distribution_version' for non 22.04 hosts
+  debug:
+    msg: "{{ item }}"
+  register: warning_present
+  loop: "{% if 'ubuntu_22_04' not in group_names %}{{ warning_host_message }}{% else %}[]{% endif %}"
```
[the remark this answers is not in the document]
skodde: Based on our discussion, I removed the confusing and unnecessary variable and simplified the loop, but I'd like to keep it (instead of using `when:`) so that the output will not include the skipped hosts.
I reference the generated list in the warning message, and I think it would be helpful to keep it more lean/readable.

src/commcare_cloud/ansible/roles/upgrade_warning/tasks/main.yml
```diff
@@ -36,5 +37,12 @@
        - https://forum.dimagi.com/t/commcare-cloud-release-notes-2023-03-29/9739
        - https://commcare-cloud.readthedocs.io/en/latest/changelog/0071-upgrade_to_ubuntu_22.04.html
 
+      If you really know what you're doing, you can silence this
+      warning by adding
+
+      --skip-tags "upgrade_warning"
```
millerdev, line 43: nit: quotes not needed
```suggestion
      --skip-tags upgrade_warning
```

src/commcare_cloud/ansible/roles/upgrade_warning/tasks/main.yml
```diff
@@ -19,15 +20,15 @@
...
   loop: "{% if 'non_2204_hosts' in group_names %}['{{ inventory_hostname }}']{% else %}[]{% endif %}"
 
 - name: Show warning if non 22.04 hosts are present
   pause:
     prompt: |-
 
-      The hosts listed above are running a version of Ubuntu which
-      we will no longer support after July 1, 2023.
+      The hosts listed above are running an unsupported OS version.
+      We will only support Ubuntu 22.04 after July 1, 2023.
```
millerdev, line 31: :+1:

src/commcare_cloud/ansible/deploy_hq.yml
skodde: I added the tag `upgrade_warning` to this role when called by the playbook, and instructions in the prompt message to add `--skip-ansible "upgrade_warning"` to the command used in order to skip the warning.
Does this solution work?

millerdev: The purpose of this override is to allow third parties who know what they are doing to disable the warning. What do you think of providing a more permanent way to disable it, such as an environment variable? Something like adding `export COMMCARE_CLOUD_IGNORE_UNSUPPORTED_OS=1` to bash profile, for example. Forcing them to always add an extra argument to the deploy command will probably get annoying.

millerdev (approved):
Noting for reference that the override argument depends on https://github.com/dimagi/commcare-cloud/pull/5917

src/commcare_cloud/ansible/roles/upgrade_warning/tasks/main.yml
skodde: They seem to be always used in official examples (e.g. https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tags.html). But, yes, it should work either way.

src/commcare_cloud/ansible/deploy_hq.yml
skodde: I think we may actually want it to be a bit annoying, at least until the deadline.
If the option is used, we know it will be visible, reminding the user of the potential risk.
After the deadline, we can definitely add an environment variable or similar solution to avoid using an explicit option, but in the meantime I think the purpose of this is also to be a bit bugging.

millerdev: I see no need to be annoying to people who know what they are doing. But I'm not going to block this PR over that disagreement.

PR #5942, opened 2023-04-26
update COMMCAREHQ_XML_POST_URLS_REGEX

Robert-Costello:
##### Environments Affected

production
india
staging


src/commcare_cloud/commands/terraform/constants.py
```diff
@@ -20,6 +20,7 @@
 ^/a/([\w\.:-]+)/apps/view/([\w-]+)/languages/bulk_app_translations/upload/$
 ^/a/([\w\.:-]+)/cloudcare/api/readable_questions/$
 ^/a/([\w\.:-]+)/cloudcare/apps/report_formplayer_error
+^/a/([\w\.:-]+)/cloudcare/apps/report_sentry_error
```
Robert-Costello, line 23: noting that this change is unrelated to the work I was doing for `answer_media` requests.

gherceg: Looks like Simon added this back in https://github.com/dimagi/commcare-hq/commit/5ee02c55a188fb9c5e317e52e8dc0ad51ad45401

gherceg (approved):
India is also an impacted environment

PR #5946, opened 2023-04-28
Create jekyll-gh-pages.yml

sravfeyn:
Followup on https://github.com/dimagi/commcare-cloud/pull/5929

It [looks](https://github.com/orgs/community/discussions/23073) like we can't have any source dir other than docs and root for the github pages setting, so this adds a new github action to build the old docs at https://dimagi.github.io/commcare-cloud.

### DIFF
--- a/.github/workflows/jekyll-gh-pages.yml
+++ b/.github/workflows/jekyll-gh-pages.yml
@@ -1,0 +1,51 @@
+# Sample workflow for building and deploying a Jekyll site to GitHub Pages
+name: Deploy Jekyll with GitHub Pages dependencies preinstalled
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["master"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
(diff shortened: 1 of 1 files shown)


sravfeyn:
@millerdev

.github/workflows/jekyll-gh-pages.yml
```diff
@@ -0,0 +1,51 @@
+# Sample workflow for building and deploying a Jekyll site to GitHub Pages
+name: Deploy Jekyll with GitHub Pages dependencies preinstalled
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["master"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
millerdev, line 10: I don't see a value for this key. Is this valid YAML?

Hmm, maybe it [is valid](https://stackoverflow.com/a/836370/10840). Next question, is `workflow_dispatch: null` valid input for github actions? I'm guessing this was copied from somewhere, so probably?

millerdev (approved):

.github/workflows/jekyll-gh-pages.yml
sravfeyn: Yeah, I think this is valid. This was auto-created when I clicked on the button linked in the github [docs](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site).

PR #5949, opened 2023-05-02
add bootstrap5 build process to build_requirejs

biyeun:
Will be needed after this HQ branch is merged:
https://github.com/dimagi/commcare-hq/pull/32907

##### Environments Affected
all


biyeun:
@millerdev as long as the non bootstrap5 command based on running locally. I don't think the babel translation has too much of an impact, especially because there is only one module affected

PR #5951, opened 2023-05-17
add control machine

shyamkumarlchauhan:
##### Environments Affected
production


dannyroberts (approved):

dannyroberts:
Make sure to confirm with all devs before destroing the old django manage, as someone might have an important process running on it that they need to let finish or transfer.

PR #5952, opened 2023-05-17
fix or deploy diffs when using 'resume'

snopoke:
This fixes the deploy diffs when using `--resume=X` by fetching the Git commit from the resumed release folder instead of using the remote HEAD commit.

With this change now in the pre-deploy code I have also removed the same logic from the Ansible tasks (`get_code_version.yml)` and rather pass in the commit as a variable.

This has the added advantage of failing early with an error message when a consistent version for 'resume' can't be found.

##### Environments Affected
None


src/commcare_cloud/commands/deploy/commcare.py
```diff
@@ -239,24 +239,23 @@ def call_record_deploy_success(environment, context, end_time):
...
         run_command=ansible_json,
     )
-    result = next(iter(res.values()), {"stderr": "no result for host"})
-    if "stdout" in result:
-        return result["stdout"]
-    error = result["stderr"] if "stderr" in result else repr(result)
-    if "rc" in result:
-        error += f"\n\nreturn code: {result['rc']}"
-    raise BadAnsibleResult(error)
+    versions = {host_result.get('stdout') for host_result in res.values() if host_result.get('stdout')}
```
millerdev, line 259: ~Looks like this will skip results that do not contain `'stdout'`, which implies it skips hosts that had an error? That seems like the wrong thing to do in the case of an error.~

After reviewing further I think I understand. This facilitates

https://github.com/dimagi/commcare-cloud/blob/ad54a8081488aede6852e28b9992475ff5fdc323/src/commcare_cloud/ansible/roles/deploy_hq/tasks/get_code_version.yml#L29-L30

Would it make sense to use a return code check here to be more like the code this is replacing:
```suggestion
    versions = {host_result.get('stdout') for host_result in res.values() if host_result.get('rc') == 0}
    if not versions or len(versions) > 1:
```
as in

https://github.com/dimagi/commcare-cloud/blob/ad54a8081488aede6852e28b9992475ff5fdc323/src/commcare_cloud/ansible/roles/deploy_hq/tasks/get_code_version.yml#L19

src/commcare_cloud/ansible/roles/deploy_hq/tasks/get_code_version.yml
```diff
@@ -1,33 +0,0 @@
...
-          | unique
-          }}
-
-    - name: Set code_version={{ code_versions[0] }}
-      # Set code version to the single unique version found among
-      # selected play hosts (when such version exists). Allows
-      # --resume=RELEASE_NAME where the git clone previously failed on
-      # some hosts or where --limit was changed to include additional
-      # hosts. A private release can be converted to a full release
-      # using --resume=RELEASE_NAME without --private.
```
millerdev, line 30: I think it would be useful to have this comment preserved at a relevant place in the new code.

millerdev (approved):

src/commcare_cloud/ansible/roles/deploy_hq/tasks/get_code_version.yml
snopoke: 8fae3a053a279160de252f4f45c690f59e787667

src/commcare_cloud/commands/deploy/commcare.py
snopoke: 559af0d6ae9a83478c3a8fc2ac22b8dfb62b5609

PR #5954, opened 2023-05-17
remove one set of node from ES cluster

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-14585

##### Environments Affected
production


environments/production/terraform.yml
```diff
@@ -160,7 +160,7 @@ servers:
       enable_cross_region_backup: yes
     group: "elasticsearch"
     os: jammy 
-    count: 7
+    count: 6
```
dannyroberts, line 163: I think we talked about removing only one node on just one AZ, so that if there was any stress it caused we would see it but it would only affect one zone. So you still want to use that approach (and then only roll out the other one the next day if a we're expecting there are no issues)?

dannyroberts (approved):

environments/production/terraform.yml
shyamkumarlchauhan: Okay , while rolling out it will be done for one AZ at a time (single day) and repeat the same for another AZ (another day)

dannyroberts (approved):

PR #5962, opened 2023-05-24
Remove UCR_COMPARISONS variable

gherceg:
Relates to https://github.com/dimagi/commcare-hq/pull/32969

This feature is no longer used and can be removed. The static UCRs that are removed in this PR no longer exist in commcare-hq which is why it is safe to remove.

### DIFF
--- a/environments/india/public.yml
+++ b/environments/india/public.yml
@@ -173,11 +173,6 @@ localsettings:
 #  STATIC_ROOT:
   WS4REDIS_CONNECTION_HOST: "redis.india.commcare.local"
   USER_REPORTING_METADATA_BATCH_ENABLED: True
-  UCR_COMPARISONS:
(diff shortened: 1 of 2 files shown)


millerdev (approved):

gherceg:
Am I overlooking how this would impact self-hosters? Do I need to make a release about this? And if so, should there be a deprecation period first?

millerdev:
I'm guessing it's unlikely that any third parties would be using this since it was a custom experiment developed for ICDS. @snopoke or others who worked on it might have a better sense?

snopoke:
> I'm guessing it's unlikely that any third parties would be using this since it was a custom experiment developed for ICDS. @snopoke or others who worked on it might have a better sense?

Correct, I don't expect this to be in use anywhere else.

snopoke (approved):

PR #5969, opened 2023-05-26
reverse order of requirejs build so that all the summary build.js files default to bootstrap 3 settings

biyeun:
I believe this will fix the symptoms in https://dimagi-dev.atlassian.net/browse/SAAS-14605 while I dig deeper into `build_requirejs` for a more longterm fix for the bootstrap5 version of the build

##### Environments Affected
all


millerdev (approved):
This looks fine if the theory about how it works is correct.

PR #5975, opened 2023-05-31
Only grab from first djangomanage machine

gherceg:
https://dimagi.slack.com/archives/C0B44C0Q2/p1685566977608219

By grabbing the latest release from one django machine, we avoid the issue of a newly deploy machine with a different rev conflicting with the most recent deploy rev. My only hesitation is the potential for the newly deployed machine to be equivalent to django_manage[0] in certain scenarios, which would result in an incorrect diff.

### DIFF
--- a/src/commcare_cloud/commands/deploy/commcare.py
+++ b/src/commcare_cloud/commands/deploy/commcare.py
@@ -246,7 +246,7 @@ def get_deployed_version(environment, from_source=False):
         hosts = "webworkers,celery,proxy,pillowtop,django_manage"
     else:
         release = environment.remote_conf.code_current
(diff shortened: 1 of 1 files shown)


skodde (approved):

skodde:
> My only hesitation is the potential for the newly deployed machine to be equivalent to django_manage[0] in certain scenarios, which would result in an incorrect diff.

I think that's unlikely, I'd imagine the new machine to always be added last in the django_manage group, but it certainly makes sense to improve the documentation regarding the deployment of new machines, especially in this group, to try and avoid this situation in the first place.

snopoke (approved):
Approving since this only affects the diff generation

PR #5977, opened 2023-06-01
Remove ddtrace from local apps

AmitPhulera:
While looking into previous changelogs of `ddtrace`, I found that they have updated the way in which `ddtrace` should be used. Even for the version that we are using i.e 0.44.0, `ddtrace` can work without being configured in settings. See the changelog [here](https://github.com/DataDog/dd-trace-py/blob/0.44/ddtrace/contrib/django/__init__.py#L131)

So I am creating this PR to remove `ddtrace` from local_apps
So removing it from staging.
##### Environments Affected

Staging


gherceg (approved):
🥇

gherceg:
Oh actually, is there a reason to not follow the other steps listed in that migration? Like what to do with `DATADOG_TRACE` configs? Looks like we use that on production

gherceg:
Additionally, it seems like step 3:
> 3. Enable Django tracing automatically via `ddtrace-run`` or manually by
   adding ``ddtrace.patch_all()`` to ``settings.py``.
   
  Might be needed on staging?

AmitPhulera:
Thanks for pointing it out. I think the change for production can be added here but the settings update need to happen in `commcare-hq`, right?

gherceg:
Ah right of course! Forgot which repo we were in. I was going to ask if order of changes matters in this case (even though it is just staging), but I guess if you merge this PR, and deploy the commcare-hq changes, there will be minimal disruption since the settings changes will be applied alongsisde the code changes. Does that sound right?

AmitPhulera:
@calellowitz https://github.com/dimagi/commcare-hq/pull/33186 is the link HQ PR.

PR #5995, opened 2023-06-13
Update the security policy to support TLS 1.3 for ALB and for Nginx

pvisweswar:
****Ticket :** https://dimagi-dev.atlassian.net/browse/SAAS-14600

**Environments Affected :** Staging


pvisweswar:
Closing this PR as we decided to change the SSL policy to another policy which supports as same as previous security policy.

PR #6002, opened 2023-06-15
Fix super-linter

Charl1996:
This change is for
1) Using the correct super-linter version
2) Telling super-linter to use the linter rule configuration files in the root of the repo (where .flake8 is situated).

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -22,11 +22,10 @@ jobs:
           fetch-depth: 0
 
       - name: Lint Code Base
-        uses: github/super-linter@v5
+        uses: super-linter/super-linter@v5
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/linter.yml
```diff
@@ -22,11 +22,10 @@ jobs:
           fetch-depth: 0
 
       - name: Lint Code Base
-        uses: github/super-linter@v5
+        uses: super-linter/super-linter@v5
         env:
           VALIDATE_ALL_CODEBASE: false
-          DEFAULT_BRANCH: master
```
Charl1996, line 28: The default values for these env vars are the same as the set values here. See [here](https://github.com/super-linter/super-linter#environment-variables).

snopoke (approved):

PR #6003, opened 2023-06-15
Upgrade Elasticsearch to version 5 on staging.

AmitPhulera:
https://dimagi-dev.atlassian.net/browse/SAAS-14657

A re-attempt of https://github.com/dimagi/commcare-cloud/pull/5992

After failing to upgrade on ES due to improper configs, @shyamkumarlchauhan  and I looked into and update the things required to make ES5 work.

We have tested it on a new machine where we were able to validate that the es was running after that changes in the PR.

We are going to attempt to upgrade staging with these changes again.

##### Environments Affected

Staging

Review by 🐟


shyamkumarlchauhan (approved):

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -13,7 +13,7 @@ node.name: "{{ elasticsearch_node_name }}"
 {% if elasticsearch_version == '2.4.6'%}
 node.zone: "{{ elasticsearch_node_zone }}"
 {% endif %}
-{% if elasticsearch_version is version('7.0.0', '>=') %}
+{% if elasticsearch_version is version('5.0.0', '>=') %}
```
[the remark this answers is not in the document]
AmitPhulera: I think it works fine. I found [this](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tests.html#comparing-versions) in ansible documentation.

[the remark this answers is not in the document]
millerdev: Indeed. It's annoying that Ansible would violate Python semantics that way. That documentation even references a PEP about version semantics, so it is definitely in a Python context. They must be doing it by parsing the expression because [it's not possible to override the behavior of the `is` operator in Python](https://stackoverflow.com/a/3993244/10840).

src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v5.yml
```diff
@@ -0,0 +1,35 @@
...
+    owner: elasticsearch
+    group: elasticsearch
+  register: copy_elasticsearch_logging_result
+  tags:
+    - es_conf
+
+- name: Install elasticsearch python client
+  become: yes
+  pip:
+    name: elasticsearch>=5.0.0,<6.0.0
```
millerdev, line 26: Does this use a virtualenv, or does it install globally? What uses this library?

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -31,20 +31,25 @@ http.max_content_length: 200mb
 path.data: {{ elasticsearch_data_dir }}/data
 path.logs: {{ elasticsearch_data_dir }}/logs
 
-{% if elasticsearch_version == '2.4.6'%}
+{% if elasticsearch_version is version('5.6.16', '<=') %}
```
millerdev, line 34: Would it make more sense to say `version('5.0.0', '>=')` here? As is, if we ever rolled out some older release of v5 for some reason, it would do the 2.4 thing, which would most likely be wrong.
```suggestion
{% if elasticsearch_version is version('5.0.0', '<') %}
```
The same applies to other version checks below.

millerdev (approved):

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
AmitPhulera: 5.6.16 is the last version that is released in 5 version series so 5.6.16 would cover every 5 version. And these changes are compatible with ES version < 5.6.16 including version 2. So I don't think that even if we roll out some older 5 version we would have issues. It should work fine as it is.

I have updated the template file in way that it can produce correct working config for both ES 5 and ES2 based on the variable `elasticsearch_variable`

src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v5.yml
AmitPhulera: I am not sure about it, I kept it because it was there in the previous version and there was a es 7 config that had it too. 
Maybe @dannyroberts or @calellowitz might have some idea where it is used.

AmitPhulera:
@millerdev since the comments seems non blocking, I am going ahead and merging this PR inorder to be sure that no-one accidentally applies new changes on staging without these branches. You are free to continue the discussions in the open comments above, I would be happy to create a followup PR if required.

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
millerdev: Oh, I was reading the condition incorrectly. But now I wonder if it should say something like `if elasticsearch_version is version('6', '<')`? Will we want this `if` block to execute if we install some newer point release, for example a security patch?

AmitPhulera: That would make sense but given that ES 5 is no longer maintained so chanced of them releasing a new patch version are minimal. So I am not sure if changing that would benefit us in any way.

PR #6005, opened 2023-06-20
Default ES reindex settings to False for dimagi envs

AmitPhulera:
https://dimagi-dev.atlassian.net/browse/SAAS-12152
As a part of ES reindex process, we would be turning on multiplexer on index by index basis. So we would be setting all the reindex settings to False and turn them on as we move ahead with the reindex on any Environment.
##### Environments Affected

India, Production, Swiss


calellowitz:
Do these need default values so self hosters can properly apply them?

millerdev (approved):
The defaults have already been set in settings.py, right?

https://github.com/dimagi/commcare-hq/blob/bd290b78840dd1df85f368b145fc40b9af364dab/settings.py#L862-L881

AmitPhulera:
Yes. Daniel is correct, they have already been added to settings before.

PR #6007, opened 2023-06-22
Turn On Multiplexer on ES india

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-14657

##### Environments Affected
india


AmitPhulera (approved):
Just an FYI for others - We will reindex Apps, Domains, Groups, SMS and Users Index first because they are very small. And then we will do rest of the indexes one by one.

millerdev (approved):
Do we need to upgrade cluster resources for future/larger reindex tasks, given that these are the small ones and they are triggering alerts on the India cluster?

AmitPhulera:
Yes. I think that might be required. Shared my analysis in [#sass-dev](https://dimagi.slack.com/archives/C05939NA5C7/p1687516497148729)

AmitPhulera:
@shyamkumarlchauhan I am merging this PR as it would create possibilities of some accidentally running update-config against master which would not have the changes from this PR. Then we would have to go through the entire reindex process again. 
We can raise a followup PR to revert the variable values.

PR #6009, opened 2023-06-26
Add ability to set maxmemory-policy=allkeys-lru and other redis params in Elasticache

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-14659

##### Environments Affected
staging materially (india, production superficially by replacing the default parameter group with a custom parameter group that inherits all the same defaults with no changes)

PR #4367, opened 2020-12-01
Added elasticache-cluster terraform module ref: saas-11511

rameshganne:
Reference to the ticket id [SAAS-11511](https://dimagi-dev.atlassian.net/browse/SAAS-11511), I am adding terraform modules (elasticache-cluster, r53-private-zone-create-update & r53-record-create-update ). 

Information about modules:
elasticache-cluster: It can create AWS elasticache service with cluster mode. 
r53-private-zone-create-update: It can create Route-53 private hosted-zone and records.
r53-record-create-update: It can create new records if the Route-53 hosted-zone already exist.

Extend to the above modules, I am updating terraform template files, environment and environment schema file to get terraform variable values from environment based files by using code.

**src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2**  -- In this template file, provided terraform modules (elasticache-cluster & r53-private-zone-create-update) in jinja template format to create commcarehq.tf file with required data to execute elasticache-cluster and r53-private-zone-create-update module
**environments/staging/terraform.yml** -- In this environment file, provided required variable values for staging environment.
**src/commcare_cloud/environment/schemas/terraform.py** -- In this schema file, added code to create terraform file by reading template file & environment file. 

Here we can create AWS Elasticache cluster, Route-53 hosted zone and add record-set with cluster end-point. These changes can apply for staging environment only. At this level, we are not adding integration part changes(application & redis [elasticache endpoint] ).

PR #6014, opened 2023-06-27
Remove unused cache_prameter_group (sic) and old "elasticache" module

dannyroberts:
Since https://github.com/dimagi/commcare-cloud/pull/6009, `cache_prameter_group` became unused. This PR removes it to clean up the code, with no effect.

The other change is that when we first switched to AWS & Terraform, the contractors we were working with set up an `elasticache` module which we didn't end up using. Much later, when we decided to use elasticache, we found what they'd left there wasn't directly usable and just wrote a separate module, `elasticache-cluster` https://github.com/dimagi/commcare-cloud/pull/4367. We should have deleted `elasticache` at that time, or really long before that time, but instead it was just left there. This PR cleans that up as well.


millerdev (commented):
Looks fine to me, but I don't have enough context on how this was used to do a meaningful review.

dannyroberts:
Sorry, you're right my description of removing the "elasticache" module was totally inadequate. Updated with better context.

millerdev (approved):

PR #6017, opened 2023-06-27
Downsize staging webworkers

dannyroberts:
t3a.xlarge => t3a.large
There's no big story behind this; I just came across it and noticed these machines were larger than they needed to be.

##### Environments Affected
staging

### DIFF
--- a/environments/staging/terraform.yml
+++ b/environments/staging/terraform.yml
@@ -35,7 +35,7 @@ servers:
     os: jammy 
 
   - server_name: "web13-staging"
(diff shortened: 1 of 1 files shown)


dannyroberts:
I already applied this, and then I also ran

```
cchq --control staging update-supervisor-confs
```
to update the worker count based on the smaller machine size.

dannyroberts:
@shyamkumarlchauhan or @AmitPhulera not urgent but it should be a pretty quick review

millerdev (approved):

PR #6018, opened 2023-06-27
Revert & Correct "Set india case pillow to not process cases in chunks"

mkangia:
Reverts dimagi/commcare-cloud#6011

Using 0 resulted for `processor-chunk-size` resulted in the param to be ignored due to if condition [here](https://github.com/dimagi/commcare-cloud/blob/a6ee9daf0d05009c75872da1e4e9897425153bc8/src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_pillowtop.conf.j2#L12).
So 1 should be used instead

### DIFF
--- a/environments/india/app-processes.yml
+++ b/environments/india/app-processes.yml
@@ -64,6 +64,7 @@ pillows:
     case-pillow:
       num_processes: 2
       dedicated_migration_process: True
+      processor_chunk_size: 1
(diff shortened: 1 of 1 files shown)


calellowitz (approved):

mkangia:
Thanks @calellowitz 
This was run on India

PR #6022, opened 2023-06-28
Use maxmemory-policy=allkeys-lru on production redis and downsize

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-14659

This has been applied on staging and india. This PR does the same for production

##### Environments Affected
production


millerdev (commented):
Which is applied first, the `allkeys-lru` parameter, or `xlarge` node type? If `xlarge` is applied first is there any risk that a lot of keys could be evicted that we want to keep? In other words, would it make sense to apply these in two separate steps so `allkeys-lru` becomes effective first?

millerdev (approved):
Saw the [note](https://dimagi-dev.atlassian.net/browse/SAAS-14659?focusedCommentId=273646) in https://dimagi-dev.atlassian.net/browse/SAAS-14659. It sounds like my concerns are valid, but elasticache will not allow the RAM to be downsized until memory usage is low enough that it will fit the smaller machine, which means the evictions must happen first, and that will be done after `allkeys-lru` has been applied.

dannyroberts:
Good question. I have a more detailed description of the steps in the ticket (https://dimagi-dev.atlassian.net/browse/SAAS-14659), but it turns out that if keys _would_ be evicted by the scale down, then it simply refuses to scale down. There's then a simple workaround to make it evict keys to fit into the target level of memory, after which you can retry.

But the takeaway from that process is that if and when it would make a difference, the `allkeys-lru` definitely gets applied first. (Also, my experience on india is that it gets applied first no matter what—i.e. terraform first applied that change, and then attempted and failed to apply the instance size change.) Also I found that contrary to what I previously thought, changing these params does not require restarting redis, so it takes effect immediately. (Changing the instance type does still of course require restarting redis.)

shyamkumarlchauhan (approved):

PR #6025, opened 2023-06-29
add support for inline update for India env

AmitPhulera:
Got the same errors that Cal faced while reindexing forms on production.
Need to turn on inline update for groovy scripts elasticsearch.
##### Environments Affected

India


dannyroberts (approved):

shyamkumarlchauhan (approved):

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -87,6 +87,7 @@ indices.fielddata.cache.size:  {{ elasticsearch_fielddata_cache_size }}
 # somewhat of a security risk but required by pact custom reports
 script.engine.groovy.inline.aggs: true
 script.engine.groovy.inline.search: true
+script.engine.groovy.inline.update: true
```
millerdev, line 90: This looks like it will affect all environments, not just India, since `elasticsearch_enable_inline_groovy_scripts` is enabled by default. Did I miss how this only affects India?

AmitPhulera: You are correct. This will affect all the environments, this was merged because it had to be applied on production again for reindex. We plan to revert it once the reindexing on Production is completed.

millerdev: The PR description says.

> Environments Affected
> 
> India

Was that an oversight?

AmitPhulera: No. It was intentional. When I raise the PR, I thought that we will apply it using --branch but @shyamkumarlchauhan recommended that we have to do the same thing on Prod as well. So we should merge it for now and will revert when we are done with reindex on production.

millerdev: I think the PR description should be updated if/when that happens so it's easy to see which envs the PR affects.

PR #6042, opened 2023-07-13
Upgrade pyOpenSSL and cryptography, make requirements

millerdev:
https://dimagi-dev.atlassian.net/browse/SAAS-14733

##### Environments Affected
All

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ install_deps = [
     'boto3>=1.9.131',
     'clint',
     'couchdb-cluster-admin',
-    'cryptography~=39.0',
+    'cryptography~=41.0',
     'datadog>=0.2.0',
     'dimagi-memoized>=1.1.0',
     'dnspython',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -26,7 +26,7 @@
     'passlib',
     'pycryptodome>=3.6.6',  # security update
     'PyGithub>=1.43.3',
-    'pyOpenSSL~=23.0',
+    'pyOpenSSL~=23.2',
```
mjriley, line 29: I see no references to `OpenSSL` in our codebase. Is this a necessary dependency anymore?

millerdev: Great question. Let's try removing it and see what it breaks.

millerdev: Update: It looks like Marco added this in d1810fbbbb85ab4bddc744de2e4ef27caf6d4fdc because we had previously been relying on a system package that gets installed indirectly. I think we need to keep it.

mjriley (approved):

PR #6055, opened 2023-07-19
Set shards and replica count to 1 for hqdomains and hqgroups

AmitPhulera:
For production we want to set replicas and shard count to 1 for both domain and groups indexes as data in these indexes are just in order of MBs and it does not makes sense to keep the default shard count of 5 for them. 
##### Environments Affected

Production


millerdev (approved):

environments/production/elasticsearch.yml
```diff
@@ -3,5 +3,11 @@ settings:
     number_of_shards: 96
   xforms:
     number_of_shards: 64
+  hqdomains:
+    number_of_shards: 1
+    number_of_replicas: 1
+  hqgroups:
+    number_of_shards: 1
+    number_of_replicas: 1
```
dannyroberts, line 11: Not blocking but curious: what was the number of replicas before?

AmitPhulera: It was set to default value which was also 1 for these indexes.

PR #6059, opened 2023-07-25
Remove unused ICDS metric

esoergel:
I came across this when looking for datadog metrics in elasticsearch code. Looks like it was built for ICDS, and it's not enabled for any Dimagi envs or documented anywhere that I could find, so it feels pretty safe to remove.

See also https://github.com/dimagi/commcare-hq/pull/33281

##### Environments Affected

None that I'm aware of, though it's possible third party environments have turned this on if they grepped through the codebase.


Jtang-1:
There's another instance of it here, can this be removed also? Why are there actually two of the seemingly same files?
https://github.com/dimagi/commcare-cloud/blob/1e5e64d909e6acf6f5b13085ce935203c5720f19/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2#L1030

esoergel:
@Jtang-1 isn't that the same file?  I only see one in my repo, [likewise on github](https://github.com/search?q=repo%3Adimagi%2Fcommcare-cloud%20TRACK_ES_REPORT_LOAD&type=code)

Jtang-1:
@esoergel Oh ya it is, oops not sure what I was seeing. My bad!

esoergel:
No worries, thanks for checking!

PR #6061, opened 2023-07-26
revert the auto restart script

pvisweswar:
Ticket: https://dimagi-dev.atlassian.net/browse/SAAS-14342

Environments Affected: Staging


src/commcare_cloud/ansible/roles/devops_scripts/templates/uptime.sh.j2
amerry19, line 1: @pvisweswar I suggest we DRY up the code (and get rid of unnecessary variables/functions) and simplify significantly. And, as Danny pointed out, since this script is called by a cron job there is no need for an internal while loop. Something like this is more concise and should do the job:

```shell
uptime_limit=90

# get uptime in days
uptime_days=$(uptime | awk '/days?/ {print $3; next}; {print 0}')

if [ "$uptime_days" -ge "$uptime_limit" ]
then
    logger '[Maintenance] Rebooting server as uptime is greater than or equal to ninety days'
    echo "Starting reboot"
    shutdown -r +1
fi
```

src/commcare_cloud/ansible/roles/devops_scripts/tasks/main.yml
```diff
@@ -8,29 +8,6 @@
...
-    dest: /usr/local/sbin/uptime.sh
-    owner: ansible
-    group: ansible
-    mode: "u=rwx,g=rwx,o=r"
-  when: "'formplayer' in group_names or 'webworkers' in group_names"
-
-- name: Creates a cron to check the uptime details
-  become: yes
-  cron:
-    name: Run uptime for every minute
```
amerry19, line 23: Since we only restart every 90 days, can we consider only running this once per day to match our uptime_limit unit? At the very least, every hour seems sufficient. Since you're calling ``` shutdown -r +1 ``` in uptime.sh it seems like we should avoid running this script every minute.

amerry19 (commented):
Left some thoughts to simply and DRY out this code for @pvisweswar.

src/commcare_cloud/ansible/roles/devops_scripts/tasks/main.yml
pvisweswar: Sure @amerry19 Thankyou for your advice and I'll look into it.

dannyroberts (approved):

PR #6072, opened 2023-08-01
adding script to auto restart task

pvisweswar:
**Ticket :** https://dimagi-dev.atlassian.net/browse/SAAS-14342

**Environments Affected :** Staging


src/commcare_cloud/ansible/roles/devops_scripts/templates/autoreboot.sh.j2
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+current_availability_zone=$(ec2metadata --availability-zone | awk -F'-' '{print $3}')
```
dannyroberts, line 3: This would be more general if you pull of just `a` instead of `1a`, because the `1` is part of the region

pvisweswar: modifying the command as suggested above **ec2metadata --availability-zone | awk -F'-' '{print $3}' | cut -c2-** 
it uses cut to remove the character from the output **-c, --characters=LIST   select only these characters**. Inour case it pulling of just a from 1a.

dannyroberts:
I just noticed, this should say **Staging** not production

> **Environments Affected** : Production

in the description. This should not affect production. My approval is assuming that this will only affect staging

pvisweswar:
> I just noticed, this should say **Staging** not production
> 
> > **Environments Affected** : Production
> 
> in the description. This should not affect production. My approval is assuming that this will only affect staging

Danny, you are correct; I set the environment to **staging**

AmitPhulera (approved):
Went through the dry-run of the command on staging. It looked good. Approved to apply the changes on staging.

AmitPhulera (approved):

PR #6076, opened 2023-08-02
Add action-eslint and lint-action

Charl1996:
This adds the new linter(s) to the repo, lint-action and action-eslint.

See [this commcarehq PR](https://github.com/dimagi/commcare-hq/pull/33280) as example usage. 

In addition to the linked PR I also verified the following scenarios:

1. No .py or .js file changes should not run the linters (the jobs does run until it determines there's not relevant files to lint over).
2. If a linter is not run it should green check that particular linting job.
3. If only .js files changed only eslint should run and flake8 should not be invoked (green check it by default).
4. If only .py files changed only flake8 should run and eslint should not be invoked (green check it by default).
5. If a linter fails the PR can still be merged.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -31,7 +31,7 @@ jobs:
         uses: actions/setup-python@v1
         if: ${{ steps.changed-files.outputs.all_changed_files }}
         with:
-          python-version: 3.8
+          python-version: 3.10
 
       - name: Install Python dependencies
         if: ${{ steps.changed-files.outputs.all_changed_files }}
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -45,3 +45,4 @@ jobs:
           flake8_auto_fix: false
           flake8_args: ${{ steps.changed-files.outputs.all_changed_files }}
           continue_on_error: false
+
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,31 +1,71 @@
----
-name: Lint Code Base
+name: Lint
 
 on:
   pull_request:
-    branches: [ master ]
+    branches:
+      - master
+
+permissions:
+  checks: write
+  contents: write
 
 jobs:
-  build:
-    name: Lint Code Base
+  lint-python:
+    name: Lint Python
     runs-on: ubuntu-latest
 
-    permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/linter.yml
```diff
@@ -1,31 +1,71 @@
...
+        uses: tj-actions/changed-files@v24
+        with:
+          files: |
+            **/*.py
+
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        if: ${{ steps.changed-files.outputs.all_changed_files }}
+        with:
+          python-version: 3.8
```
millerdev, line 34: commcare-cloud requires Python 3.10. Would it make more sense to use that version?

.github/workflows/linter.yml
```diff
@@ -1,31 +1,71 @@
...
+        if: ${{ steps.changed-files.outputs.all_changed_files }}
+        uses: wearerequired/lint-action@v2
+        with:
+          flake8: true
+          flake8_auto_fix: false
+          flake8_args: ${{ steps.changed-files.outputs.all_changed_files }}
+          continue_on_error: false
+
+  # If this is in the same step as python linting the annotations don't show up
+  lint-javascript:
```
esoergel, line 50: Can we drop this section entirely?  I don't believe commcare-cloud contains any javascript files

esoergel (commented):
I don't see the linter running on this PR - is it not functioning properly?

.github/workflows/linter.yml
Charl1996: You're right, I'll make the update.

Charl1996:
> I don't see the linter running on this PR - is it not fully set up yet?

Ah, I see the workflow was disabled manually. I enabled it and pushed a commit to trigger it.

millerdev (approved):

.github/workflows/linter.yml
```diff
@@ -45,3 +45,4 @@ jobs:
           flake8_auto_fix: false
           flake8_args: ${{ steps.changed-files.outputs.all_changed_files }}
           continue_on_error: false
+
```
esoergel, line 48: I don't wanna know how this solves the problem :see_no_evil: 
must've been frustrating to figure out

esoergel (approved):

PR #6084, opened 2023-08-09
consolidate celery workers

shyamkumarlchauhan:
https://dimagi-dev.atlassian.net/browse/SAAS-14816

##### Environments Affected
production

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -14,7 +14,7 @@ management_commands:
   pillow_a2000:
     run_pillow_retry_queue:
 celery_processes:
-  celery_b2000,celery_b2001,celery_c2001:
+  celery_b2000,celery_b2001,celery_c2000,celery_c2001:
     analytics_queue:
       pooling: gevent
       concurrency: 4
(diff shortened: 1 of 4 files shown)


dannyroberts (commented):
We offlined, but just to note here too, I think this change requires a bit more thought. There are two main things:

1. Right now celery_c2000, awkwardly, has a different workload from the rest of the machines, so we don't want to double its size, but operationally in this configuration file the its instance type is tied to the instance types of the other machines. Ideally it would stand as an entry in terraform.yml on its own
2. In order to take advantage of the extra CPU and RAM, we would probably need to double all of the concurrencies for the shared queues (but not the ones on celery_c2000). With half the machines but double the same concurrency-per-machine, the total concurrency will be unchanged.

shyamkumarlchauhan:
> We offlined, but just to note here too, I think this change requires a bit more thought. There are two main things:
> 
> 1. Right now celery_c2000, awkwardly, has a different workload from the rest of the machines, so we don't want to double its size, but operationally in this configuration file the its instance type is tied to the instance types of the other machines. Ideally it would stand as an entry in terraform.yml on its own
> 2. In order to take advantage of the extra CPU and RAM, we would probably need to double all of the concurrencies for the shared queues (but not the ones on celery_c2000). With half the machines but double the same concurrency-per-machine, the total concurrency will be unchanged.

Updated the PR with required changes.

PR #6092, opened 2023-08-16
Upgrade Terraform to version 1.5 on staging

pvisweswar:
**Ticket: ** https://dimagi-dev.atlassian.net/browse/SAAS-14730

**Environments Affected :** Staging

PR #6093, opened 2023-08-16
Fix python version specification in linter action

dannyroberts:
3.10 was being interpreted as a number, 3.1, instead of a string "3.10"

Noticed this error on https://github.com/dimagi/commcare-cloud/pull/6092. Not sure why this wouldn't have been pervasive on all PRs in the last two weeks.

##### Environments Affected
None


Charl1996 (approved):
Woahw, thanks!

PR #6094, opened 2023-08-17
Updates to avoid big last batch

mkangia:
Current HQ deploy fails due to rolling restart trying a restart on the full set of webworkers under `production_commcare_submissions` block.
This change would help with a smaller blocks
This would just cause a slightly longer deploy (~1 minute)

Using https://github.com/dimagi/commcare-cloud/pull/5852/files/f66001d7386ea30437e99a2093415f2a31d6702b#diff-3d7593c6d6f4d6c05a0a5c9a194301e3eaec3cc436f730dfa64b233fa5102054R3 as the guide to ensure that the change should be safe

##### Environments Affected

All

### DIFF
--- a/src/commcare_cloud/ansible/restart_webworker_services.yml
+++ b/src/commcare_cloud/ansible/restart_webworker_services.yml
@@ -1,6 +1,6 @@
 - name: Rolling restart webworker services
   hosts: webworkers
-  serial: ["10%", "40%", "75%", "100%"]
+  serial: ["10%", "30%", "50%", "100%"]
   become: true
   any_errors_fatal: true
   pre_tasks:
--- a/src/commcare_cloud/ansible/restart_webworker_services.yml
+++ b/src/commcare_cloud/ansible/restart_webworker_services.yml
@@ -1,6 +1,6 @@
 - name: Rolling restart webworker services
   hosts: webworkers
-  serial: ["10%", "40%", "100%"]
+  serial: ["10%", "40%", "75%", "100%"]
   become: true
   any_errors_fatal: true
   pre_tasks:


AmitPhulera (approved):

mkangia:
Some more details,

Currently there are total 10 servers in the production nginx configuration in two blocks
```
upstream production_commcare {
  zone production_commcare 64k;

    least_conn;

  server 10.202.11.206:9010;
  server 10.202.11.136:9010;
  server 10.202.11.87:9010;
  server 10.202.10.119:9010;
  server 10.202.10.74:9010;
  server 10.202.10.148:9010;
}
upstream production_commcare_submissions {
  zone production_commcare_submissions 64k;

    least_conn;

  server 10.202.11.46:9010;
  server 10.202.11.57:9010;
  server 10.202.10.110:9010;
  server 10.202.10.73:9010;
}
```

As per the configuration `["10%", "40%", "100%"]`  before this change, the comment-restart-uncomment process would do the following number of servers
1 - 4 - rest (5)
where last 5 were these,
```
changed: [10.202.10.148 -> 10.202.10.83] => (item=10.202.10.83)
--- before: /etc/nginx/sites-available/production_commcare
+++ after: /etc/nginx/sites-available/production_commcare
@@ -12,17 +12,17 @@
   server 10.202.11.87:9010;
   server 10.202.10.119:9010;
   server 10.202.10.74:9010;
-  server 10.202.10.148:9010;
+#  server 10.202.10.148:9010;
 }
 upstream production_commcare_submissions {
   zone production_commcare_submissions 64k;
 
     least_conn;
   
-  server 10.202.11.46:9010;
-  server 10.202.11.57:9010;
-  server 10.202.10.110:9010;
-  server 10.202.10.73:9010;
+#  server 10.202.11.46:9010;
+#  server 10.202.11.57:9010;
+#  server 10.202.10.110:9010;
+#  server 10.202.10.73:9010;
 }
```
This caused the block `production_commcare_submissions` to have all servers commented out which caused the nginx configuration to be invalid with error "no servers are inside upstream".

When I made the [first commit](https://github.com/dimagi/commcare-cloud/pull/6094/commits/718701da5ad16bac61547cd8c57535aa369379b9) to update to `["10%", "40%", "75%", "100%"]` it didn't change anything for production and it still did 1-4-all, which was surprising.

So i updated it to `["10%", "30%", "50%", "100%"]` which finally did it as it updated 1 server, then 3, then 5 and then pending 1.

I initially misunderstood those numbers.
When I did `["10%", "40%", "75%", "100%"]`,
I thought it was cumulative and it meant to do, 10% of the servers, then 30% more (40-10) , then 35% more (75-40) and then the pending ones but it did 10% of *total* servers so 1, then 40% of *total* servers so 4 *more* and then 75% was 7 more and then none were left to do the pending 100%. So it again did 1-4-7 (or 5 since we had total 10)

So the new configuration `["10%", "30%", "50%", "100%"]`, did 1 server, then 3 more, then 5 more and then pending 1.

Pasting the log of each server updates block for records
```
-  server 10.202.11.206:9010;
+#  server 10.202.11.206:9010;
   server 10.202.11.136:9010;
   server 10.202.11.87:9010;
   server 10.202.10.119:9010;
```

```
   server 10.202.11.206:9010;
-  server 10.202.11.136:9010;
-  server 10.202.11.87:9010;
-  server 10.202.10.119:9010;
+#  server 10.202.11.136:9010;
+#  server 10.202.11.87:9010;
+#  server 10.202.10.119:9010;
   server 10.202.10.74:9010;
   server 10.202.10.148:9010;
```
```
   server 10.202.11.136:9010;
   server 10.202.11.87:9010;
   server 10.202.10.119:9010;
-  server 10.202.10.74:9010;
-  server 10.202.10.148:9010;
+#  server 10.202.10.74:9010;
+#  server 10.202.10.148:9010;
 }
 upstream production_commcare_submissions {
   zone production_commcare_submissions 64k;
 
     least_conn;
   
-  server 10.202.11.46:9010;
+#  server 10.202.11.46:9010;
   server 10.202.11.57:9010;
-  server 10.202.10.110:9010;
-  server 10.202.10.73:9010;
+#  server 10.202.10.110:9010;
+#  server 10.202.10.73:9010;
 }
```
```
   
   server 10.202.11.46:9010;
-  server 10.202.11.57:9010;
+#  server 10.202.11.57:9010;
   server 10.202.10.110:9010;
   server 10.202.10.73:9010;
 }

```

I guess a better approach would be to update each block in nginx configuration instead of the full group proxy in [this task](https://github.com/dimagi/commcare-cloud/blob/c144448846c0fe29249ac953c10496d64fb7d9f4/src/commcare_cloud/ansible/roles/deploy_hq/tasks/commission_hosts.yml#L9) if that is possible because with the way it is currently written, its possible we could run into this issue again.

src/commcare_cloud/ansible/restart_webworker_services.yml
```diff
@@ -1,6 +1,6 @@
 - name: Rolling restart webworker services
   hosts: webworkers
```
dannyroberts, line 2: Amazing find, @mkangia and @AmitPhulera!

I think the best solution would be to break this up into two groups like you noted @mkangia, so we hav one with `hq_webworkers` and one with `mobile_webworkers`. To make matters more complicated, it would need to detect whether that split exists (as in SaaS production for example) or we just have one big `webworkers` group (as possibly in some self-hosted setups).

I think we could achieve that by passing hosts (e.g. `webworker_hosts`) as a variable into this playbook and move this logic one level up to https://github.com/dimagi/commcare-cloud/blob/f66001d7386ea30437e99a2093415f2a31d6702b/src/commcare_cloud/ansible/restart_commcare_services.yml#L28 with something like this:

```yaml
# here we use groups['hq_webworkers'] being truthy or not
# to determine whether webworkers are split (into hq_webworkers versus mobile_webworkers) or not

- name: "Restart all webworkers (if not split)"
  import_playbook: restart_webworker_services.yml
  vars:
    webworker_hosts: 'webworkers'
  when: not groups['hq_webworkers']

- name: "Restart hq_webworkers (if split)"
  import_playbook: restart_webworker_services.yml
  vars:
    webworker_hosts: 'hq_webworkers'
  when: groups['hq_webworkers']

- name: "Restart mobile_webworkers (if split)"
  import_playbook: restart_webworker_services.yml
  vars:
    webworker_hosts: 'mobile_webworkers'
  when: groups['hq_webworkers']
```

I haven't tested this, but in theory something like that should work well. (I'm often wrong about what ansible constructs are available in what contexts, so that's an example of something I could have gotten wrong!)

mkangia: Thanks @dannyroberts, good to see that something like this is possible.
I am okay with holding off merging this PR if someone should give that solution a try. I believe we are okay till next deploy is needed on prod.

dannyroberts: I think if this is a meaningful stop-gap measure we should go ahead and merge it, to give any deploy in the interim the best chance of succeeding. I'll go ahead and do that and then also make a note to consider reverting once we have the "better" solution.

mkangia: Cool. Thanks for the merge @dannyroberts

PR #6098, opened 2023-08-18
Fix batching logic during restart servers

AmitPhulera:
https://dimagi-dev.atlassian.net/browse/SAAS-14835

We observed during the deploy that our current configuration to restart webworkers in batches is putting us in a condition where all the servers are getting commented in `production_commcare_submissions`. This is happening due to reasons explained in https://github.com/dimagi/commcare-cloud/pull/6094#issuecomment-1682101763

This PR tries to fix it as suggested in https://github.com/dimagi/commcare-cloud/pull/6094#discussion_r1297316040

##### Environments Affected

All

### DIFF
--- a/src/commcare_cloud/ansible/restart_webworker_services.yml
+++ b/src/commcare_cloud/ansible/restart_webworker_services.yml
@@ -1,6 +1,6 @@
 - name: Rolling restart webworker services - ({{ webworker_hosts }})
   hosts: "{{ webworker_hosts }}"
-  serial: ["10%", "30%", "50%", "100%"]
+  serial: ["10%", "40%", "100%"]
   become: true
   any_errors_fatal: true
   pre_tasks:


mkangia (commented):
I am not aware how the inventory is defined, so just checking, Can't we just do a restart for webworkers, hq_webworkers and mobile_webworkers or it would lead to a double restart for hq_webworkers and mobile_webworkers.

dannyroberts (approved):
Confirmed with Amit that this has been run on staging already (including the last commit).

dannyroberts:
> I am not aware how the inventory is defined, so just checking, Can't we just do a restart for webworkers, hq_webworkers and mobile_webworkers or it would lead to a double restart for hq_webworkers and mobile_webworkers.

In some environments, only `webworkers` is defined; in others, `hq_webworkers` and `mobile_webworkers` are defined (with servers assigned only to one or the other), and `webworkers` is defined as the union of the two. In that latter case, if you ran it for each of the three groups, then like you said it would lead to a double restart and the restart associated with `webworkers` would be subject to the same issues that you ran into that motivated this change.

src/commcare_cloud/ansible/restart_webworker_services.yml
```diff
@@ -1,5 +1,5 @@
-- name: Rolling restart webworker services
-  hosts: webworkers
+- name: Rolling restart webworker services - {{ host_type }}
```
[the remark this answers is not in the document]
AmitPhulera: I got rid of the `host_type` variable and I assume that  `({{ webworker_hosts }})` was used to display the list of hosts that would be restarted by the task?
If that is the case, the syntax did not seem to work, I tried to google for a way to do that, did not find a straight forward way. Any suggestion on what can be done here?

src/commcare_cloud/ansible/restart_webworker_services.yml
```diff
@@ -1,4 +1,4 @@
-- name: Rolling restart webworker services - {{ webworker_hosts }}
+- name: Rolling restart webworker services - ({{ webworker_hosts }})
```
dannyroberts, line 1: I'm not sure I understand this change and how it relates to the commit message

AmitPhulera: I thought that adding the brackets would make ansible make interpret it as a variable which it did not. And after reading your last comment I understand why.

src/commcare_cloud/ansible/restart_webworker_services.yml
```diff
@@ -1,6 +1,6 @@
 - name: Rolling restart webworker services - ({{ webworker_hosts }})
   hosts: "{{ webworker_hosts }}"
-  serial: ["10%", "30%", "50%", "100%"]
+  serial: ["10%", "40%", "75%", "100%"]
```
dannyroberts, line 3: I believe it was previously
```
["10%", "40%", "100%"]
```
just looking at https://github.com/dimagi/commcare-cloud/pull/6094/files#diff-3d7593c6d6f4d6c05a0a5c9a194301e3eaec3cc436f730dfa64b233fa5102054L3.

Also based on the logic of what ansible does with these batch sizes, I think actually both what you have here and what we had before will result in exactly the same behavior: three batches of 10%, then 40%, then 50%.

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_strategies.html#rolling-update-batch-size describes this. Since after doing 10% and then 40%, there is 50% remaining, whether the next batch size is 75% or 100% (both larger than 50%), it will just do all of the remaining 50%. Then it's done so if there's another batch size in the list (100%) then it won't get used because it's already done by that point.

dannyroberts (approved):
Looks fine in terms of safety, but had one comment (on the batch sizes) which should probably be addressed just for readability.

src/commcare_cloud/ansible/restart_webworker_services.yml
mkangia: I guess we could revert to the original numbers now since we have fixed the underlying problem.
10-40-100 or 10-40-50 are both okay and we don't need to do this in more than 3 batches

AmitPhulera: Thanks @dannyroberts.
I think I compared it to previous commit's changes and got the values from there.  I forgot that the values were changed twice.

dannyroberts (approved):

mkangia:
Would be good to remove unnecessary commits before merging this, since we went back to the earlier implementation.

AmitPhulera:
@millerdev Yes that is correct. We would be performing the tasks defined in restart playbook 6 times rather than 3.

AmitPhulera:
@mkangia Cleaned up the commit history, just kept three relevant commits.

dannyroberts (approved):

mkangia:
Thanks for seeing this through @AmitPhulera and @dannyroberts

PR #6101, opened 2023-08-23
Swap form index

AmitPhulera:
https://dimagi-dev.atlassian.net/browse/SAAS-14740

We have reindexed form index on production, we now need to switch the reads from older index to the newer index. 
The changes in PR would do that. 

##### Environments Affected

Production


dannyroberts (approved):
I assume this is to be applied after adding the new shard replicas is complete

AmitPhulera:
Yes that is correct.

AmitPhulera:
Replica shards have been assigned. Merging this now.

PR #6102, opened 2023-08-24
Turn multiplexer off for forms index

AmitPhulera:
We have swapped forms index today now we would be turning the multiplexer off for the forms index.

The changes will be only applied after we have kept the indexes swapped for one EST working day.
##### Environments Affected

Production


AmitPhulera:
Looked in Sentry, found nothing related to changes.

Going ahead and turning multiplexer off on Forms index. Merging this.

PR #3796, opened 2020-04-30
prometheus supervisor process

rohit-dimagi:
* https://dimagi-dev.atlassian.net/browse/IIO-770
##### ENVIRONMENTS AFFECTED

* ICDS

PR #6113, opened 2023-09-01
Fix supervisor service file management

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-14853?focusedCommentId=284526

This change prevents an issue that can happen when moving certain processes between machines; types of processes affected by this issue include: celery beat, celery flower, and the management commands, at least. The issue was that on the machine that is no longer supposed to have the process, the corresponding supervisor service file is simply _abandoned_, rather than deleted, which means it ends up running in both the old and new locations.

To fix this, this PR takes logic that was strewn through multiple files and centralizes it. Critically, it centralizes the logic dictating whether each individual service file is supposed to exist on a given machine or not. This allows the process that sweeps the service file directory at the end for unexpected files to have a complete understanding of what files are supposed to be there. Previously, the logic available to this task presumed, basically, that if the machine was a celery machine then any celery beat service file (for example) was supposed to be there. This PR updates that logic so that it's exactly the same logic as the task uses to decide whether to create that file on a given machine or not in the first place.

This issue was first introduced in https://github.com/dimagi/commcare-cloud/pull/3796/commits/58004f054720ae399c921ba9dbcaad12ff3d1363 in mid 2020 (though from the look of that PR https://github.com/dimagi/commcare-cloud/pull/3796, that change itself was introduced to fix a different problem).

##### Safety Story
This PR is backwards compatible, in the sense that any environment already set up correctly won't experience a diff when `update-supervisor-confs` is next applied. Of course, if it has stray service files that weren't previously detected, those will be deleted in the diff after this update. I have confirmed this (in check mode of course) on staging, india, and production.

I have also confirmed that it is still true to the previous behavior of applying changes made to `inventory.ini` and `app-processes.yml` through some ad hoc spot testing. Given that the changes were all very mechanical and had a limited impact on the overall logic, I think this is sufficient to allay concerns that these changes broke the fundamental logic in some important way.

##### Environments Affected
all (no environment files changed, though)

### DIFF
--- a/src/commcare_cloud/ansible/group_vars/all.yml
+++ b/src/commcare_cloud/ansible/group_vars/all.yml
@@ -212,3 +212,7 @@ supervisor_service_files:
     file_path: "{{ service_home }}/{{ deploy_env }}_supervisor_management_commands.conf"
     template: ../templates/supervisor_management_commands.conf.j2
     should_exist: "{{ app_processes_config.management_commands.get(inventory_hostname, {}) }}"
+  pillowtop:
+    file_path: "{{ service_home }}/{{ deploy_env }}_supervisor_pillowtop.conf"
+    template: ../templates/supervisor_pillowtop.conf.j2
+    should_exist: "{{ inventory_hostname in groups['pillowtop'] }}"
--- a/src/commcare_cloud/ansible/group_vars/all.yml
+++ b/src/commcare_cloud/ansible/group_vars/all.yml
@@ -186,3 +186,7 @@ supervisor_service_files:
     file_path: "{{ service_home }}/{{ deploy_env }}_supervisor_django.conf"
     template: ../templates/supervisor_django.conf.j2
     should_exist: "{{ inventory_hostname in groups['webworkers'] }}"
+  celery_beat:
+    file_path: "{{ service_home }}/{{ deploy_env }}_supervisor_celery_beat.conf"
+    template: ../templates/supervisor_celery_beat.conf.j2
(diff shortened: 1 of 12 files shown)


src/commcare_cloud/ansible/group_vars/all.yml
```diff
@@ -169,3 +169,63 @@ ansible_become_password: "{{ ansible_sudo_pass }}"
...
+#  ./src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_celery_beat.conf.j2
+#  ./src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_prometheus.conf.j2
+#  ./src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_pillowtop.conf.j2
+#  ./src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_celery_workers.conf.j2
+#  ./src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_websockets.conf
+#  ./src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_formplayer_spring.conf.j2
+#  ./src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_management_commands.conf.j2
+#  ./src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_celery_flower.conf.j2
+#
+# The only ones not matching this pattern is celery_bash_runner and django_bash_runner
```
dannyroberts, line 185:
```suggestion
# The only ones not matching this pattern are celery_bash_runner and django_bash_runner
```

millerdev (approved):

PR #6108, opened 2023-08-29
Add placeholder couchdb nodes to production that we can enlarge when ready

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-14844

This is "step 0" in a multi-step process to entirely rebuild the couch cluster, migrating one node at a time. Each PR represents one self-contained maintenance procedure. This PR simply adds very small "placeholder" nodes to the existing couch cluster without assigning them any data or sending them any traffic. Later steps will update these nodes to their full size when their time has come. The maintenance procedures are described in https://confluence.dimagi.com/display/saas/Replace+a+Couch+node+in+Couch+cluster.

The purpose of adding placeholders first is to be able to do each step without time pressure. This allows us to add all the nodes up front without incurring the huge costs of having 6 full size nodes until we're done with all the maintenance.

- **Step 0: https://github.com/dimagi/commcare-cloud/pull/6108**
- Step 1: https://github.com/dimagi/commcare-cloud/pull/6116
- Step 2: https://github.com/dimagi/commcare-cloud/pull/6117
- Step 3: https://github.com/dimagi/commcare-cloud/pull/6118
- Step 4: https://github.com/dimagi/commcare-cloud/pull/6110

Upon PR approval, I will create these machines, push the file updates that require machine info, and run the procedure to add these nodes to the cluster.

##### Environments Affected
production

PR #6116, opened 2023-09-06
Replace production couch node 1

dannyroberts:
https://dimagi-dev.atlassian.net/browse/SAAS-14844

- Step 0: https://github.com/dimagi/commcare-cloud/pull/6108
- **Step 1: https://github.com/dimagi/commcare-cloud/pull/6116**
- Step 2: https://github.com/dimagi/commcare-cloud/pull/6117
- Step 3: https://github.com/dimagi/commcare-cloud/pull/6124
- Step 4: https://github.com/dimagi/commcare-cloud/pull/6110

This PR will be applied using Phase 2 of https://confluence.dimagi.com/display/saas/Replace+a+Couch+node+in+Couch+cluster (applied to just this one node replacement).

##### Environments Affected
production

### DIFF
--- a/environments/production/migrations/0009-move-to-3-jammy-nodes/couchdb-step-1.yml
+++ b/environments/production/migrations/0009-move-to-3-jammy-nodes/couchdb-step-1.yml
@@ -1,0 +1,2 @@
+target_allocation:
+  - couch11,couch12,couch_a0:3
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -191,17 +191,13 @@ servers:
     os: bionic
 
   - server_name: "couch_a0-production"
-    ## Increase once ready for migration
-    # server_instance_type: c6a.8xlarge
-    server_instance_type: c6a.large
+    server_instance_type: c6a.8xlarge
     network_tier: "db-private"
     az: "a"
     volume_size: 80
     volume_encrypted: yes
     block_device:
-      ## Increase once ready for migration
-      # volume_size: 3150
-      volume_size: 1
+      volume_size: 3150
       encrypted: yes
       ## Enable once migration is complete
       # enable_cross_region_backup: yes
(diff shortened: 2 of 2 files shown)


environments/production/migrations/0009-move-to-3-jammy-nodes/couchdb-step-1.yml
```diff
@@ -0,0 +1,2 @@
+target_allocation:
```
AmitPhulera, line 1: Curious what this file is used for?

AmitPhulera (approved):

environments/production/migrations/0009-move-to-3-jammy-nodes/couchdb-step-1.yml
dannyroberts: In https://confluence.dimagi.com/display/saas/Replace+a+Couch+node+in+Couch+cluster, it's the PLAN_FILE that is used in phase 2.

gherceg: Taking a stab at answering this cause why not 😅 I believe we reference this migration file when running `cchq ${ENV} migrate-couchdb ${PLAN_FILE} plan` in phase 2 of adding this node. And then it looks like that file is used to [allocate shards properly](https://github.com/dimagi/commcare-cloud/blob/2bb8d4bdd71ad8efc3c150c8fc03eeb5591ec47d/src/commcare_cloud/commands/migrations/couchdb.py#L321-L323) via couchdb-cluster-admin.

So then the syntax here is saying that we have three nodes and should make three copies of data across these nodes, which might be the same as saying 1 copy per node?

gherceg (approved):

environments/production/migrations/0009-move-to-3-jammy-nodes/couchdb-step-1.yml
dannyroberts: Yeah but you said it better, Graham :)

AmitPhulera: Thanks @dannyroberts and @gherceg. That was super helpful :)

PR #6120, opened 2023-09-09
Add jvm option path

AmitPhulera:
The ES 5 was not respecting the config path and jvm options need to be specifically defined using Environment variable ES_JVM_OPTIONS
##### Environments Affected

All

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
@@ -9,6 +9,9 @@ Type=simple
 Environment=ES_HOME={{ elasticsearch_home }}
 Environment=CONF_DIR={{ elasticsearch_conf_dir}}
 Environment=ES_PATH_CONF={{ elasticsearch_conf_dir}}
+{% if elasticsearch_version == '5.6.16'%}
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):

src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
```diff
@@ -9,6 +9,9 @@ Type=simple
 Environment=ES_HOME={{ elasticsearch_home }}
 Environment=CONF_DIR={{ elasticsearch_conf_dir}}
 Environment=ES_PATH_CONF={{ elasticsearch_conf_dir}}
+{% if elasticsearch_version == '5.6.16'%}
```
millerdev, line 12: Assuming the same issue exists with newer versions of ES as well, would be nice if this was more flexible. Something along the lines of `elasticsearch_version > 2`

AmitPhulera: I think this should be fixed in ES 6 - See https://github.com/elastic/elasticsearch/issues/25964#issuecomment-318925717

millerdev: That's good. Will we ever run a different _5.x.x_ version? Will this need to be updated if we do?

PR #6128, opened 2023-09-18
Allow --check and be less strict about --diff

millerdev:
It's confusing to get a "that option is managed automatically" error when attempting to run with --check. The operator is forced to guess how to accomplish a --check run when lacking a thorough understanding of how the automatic option management works. It seems appropriate to perform a single --check run if that was requested by the operator.

The situation is similar for --diff, although a duplicate option does not cause harm and does not change the run mode.

##### Environments Affected
None.


gherceg (approved):
I can't decide if there is some refactoring that could be done here with `run_action_with_check_mode`, but it seems out of scope of this PR anyway. Looks good!

AmitPhulera (approved):

PR #6129, opened 2023-09-18
Use subprocess.run instead of check_output

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-14935

AWS sso login recently changed to include a verification code to ensure the device that requested the login is the same as the one you are allowing authentication for. We used `check_output` to call `aws sso login`, which swallows the output and does not wait for the command to finish, meaning users could not verify the matching code. By switching to `subprocess.run`, we are able to see the command output as if running it directly, and therefore can verify the code.
##### Environments Affected

None

### DIFF
--- a/src/commcare_cloud/commands/terraform/aws.py
+++ b/src/commcare_cloud/commands/terraform/aws.py
@@ -40,6 +40,19 @@ def check_output(cmd_parts, env, silent=False):
     return subprocess.check_output(cmd_parts, env=env_vars)
 
 
+def run(cmd_parts, env, silent=False):
+
+    env_vars = os.environ.copy()
+    env_vars.update(env)
+    if not silent:
+        cmd = ' '.join(shlex_quote(arg) for arg in cmd_parts)
+        print_command('{} {}'.format(
+            ' '.join('{}={}'.format(key, value) for key, value in env.items()),
+            cmd,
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -40,6 +40,20 @@ def check_output(cmd_parts, env, silent=False):
...
+
+    env_vars = os.environ.copy()
+    env_vars.update(env)
+    if not silent:
+        cmd = ' '.join(shlex_quote(arg) for arg in cmd_parts)
+        print_command('{} {}'.format(
+            ' '.join('{}={}'.format(key, value) for key, value in env.items()),
+            cmd,
+        ))
+    # check=True to raise error if results in non-zero exit status
```
esoergel, line 52: :star:

esoergel (approved):
Looks great - do you happen to have an example of what this looks like in the console when run?

gherceg:
Ah good idea @esoergel, here's an example. I left that last error message in there because that was a part of this before and still is but does not impact the success of this (we haven't gotten to the bottom of how to fix it). 

Screenshot 2023-09-18 at 3 49 01 PM

millerdev (approved):

PR #6133, opened 2023-09-25
Re-enable compressing formplayer DBs on production

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-14929

During COVID times, we [disabled formplayer DB compression](https://github.com/dimagi/commcare-cloud/pull/3997) as a tradeoff to increase performance (decompressing the database takes time). We now find ourselves in a steadier state, but disk usage across formplayer machines is nearing its limit. There are quite a few very large user DBs, so the hope is that by reintroducing this compression, we will free up enough disk space to make a meaningful difference.

We chose 3 days as the inactive threshold to ensure a more random distribution of decompressing DBs. We wanted to avoid "stampede effects", where users log on at the same time every morning and have to wait for their DB to be decompressed, or similarly log on Monday morning after the weekend. 3 days is hopefully large enough to avoid this, but small enough to actually have an impact.

##### Environments Affected

Production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -67,6 +67,7 @@ filebeat_inputs:
     tags: nginx-error
 
 formplayer_java_version: "{{ java_17_bin_path }}/java"
+formplayer_archive_time_spec: '3d'
 formplayer_purge_time_spec: '10d'
 formplayer_sensitive_data_logging: true
 formplayer_forward_ip_proxy: true


dannyroberts:
Code change looks good to me, but before we can apply, I'm looking for a basic test on staging that the rollout is smooth and some test that all the supporting functionality (like formplayer's ability to decrompess) still works.

gherceg:
Yeah I should have been more explicit about this, but I had seen that `formplayer_archive_time_spec` was set to 10 minutes on staging (https://github.com/dimagi/commcare-cloud/blob/ccc2e7030ae02298b3251c26382bc8e6c83cc97e/environments/staging/public.yml#L46) and given that formplayer on staging has been operational since that change (2017?), that was enough to ensure compressing and de-compressing DBs still behaves as expected.

That being said, I can still test the rollout by changing this value on staging, confirming my user db is compressed after X minutes, and confirming my user db is decompressed and usable after that. Am I overlooking something with that `formplayer_archive_time_spec` value in staging/public.yml?

dannyroberts:
That all sounds right—that would be sufficient testing for me!

dannyroberts (approved):

PR #6150, opened 2023-10-23
Update gevent (used by couchdb-cluster-admin)

gherceg:
https://dimagi-dev.atlassian.net/browse/SAAS-15034

gevent is only used by couchdb-cluster-admin which is used in the `couchdb-cluster-info` command. I ran `cchq --control staging couchdb-cluster-info --branch=gh/upgrade-gevent` on staging to make sure it still works, which it does.

##### Environments Affected

None


dannyroberts (approved):
excellent

PR #6158, opened 2023-10-30
Fix Readthedocs build

AmitPhulera:
Fixes - 

Screenshot 2023-10-30 at 8 03 22 PM


gherceg (approved):
Do you know what change caused this to happen? I assume it was on readthedocs' end?

AmitPhulera:
Yep. This https://blog.readthedocs.com/use-build-os-config/

PR #6159, opened 2023-10-30
Fix readthedocs - Add python version

AmitPhulera:
I thought tools might not be required in https://github.com/dimagi/commcare-cloud/pull/6158, but it seems it is required in the doc config.


gherceg (approved):

AmitPhulera:
This fixed the builds 🎉

PR #6183, opened 2023-12-15
Temporarily double number of formplayer machines in production

dannyroberts:
at USH's request

https://dimagi-dev.atlassian.net/browse/SAAS-15177

To be applied with the standard maintenance procedure https://confluence.dimagi.com/display/saas/Add+new+formplayer+node.

##### Environments Affected
production


environments/production/terraform.yml
```diff
@@ -299,6 +299,19 @@ servers:
     count: 10
     server_auto_recovery: true
 
+  - server_name: "formplayer_a2{i}-production"
+    server_instance_type: r6a.2xlarge
+    network_tier: "app-private"
+    az: "a"
+    volume_size: 50
```
gherceg, line 306: I assume the volume size doesn't need to be as large as existing machines since this will distribute users across twice the number of machines? Just curious how you settled on 50.

gherceg (approved):

environments/production/terraform.yml
dannyroberts: Yeah that was my logic. Just save a bit of money (but also really not very much money now that I think of it). But yeah, I expect disk usage to approximately halve when we double the number of machines.

gherceg (approved):

gherceg (approved):

PR #6192, opened 2023-12-21
Reuse maintenance window check for formplayer deploys

gherceg:
I deployed formplayer at a non ideal time, and realized the existing maintenance window check we have for commcare deploys is not used for formplayer. This PR reuses that code for formplayer deploys as well.

Tested on production:
image

##### Environments Affected

None

### DIFF
--- a/src/commcare_cloud/commands/deploy/commcare.py
+++ b/src/commcare_cloud/commands/deploy/commcare.py
@@ -1,8 +1,6 @@
 import shlex
 from datetime import datetime, timedelta
 
-import pytz
-
 from commcare_cloud.alias import commcare_cloud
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/commands/deploy/commcare.py
```diff
@@ -1,8 +1,6 @@
 import shlex
 from datetime import datetime, timedelta
 
-import pytz
```
gherceg, line 4: pytz is already imported in deploy/utils.py if you happen to notice that there isn't a diff adding it to the file.

dannyroberts (approved):

millerdev (approved):

PR #6204, opened 2024-01-17
A few small documentation tweaks

kaapstorm:
Some documentation tweaks, based on recently deploying a monolith a couple of times.

##### Environments Affected

None. Documentation only.

### DIFF
--- a/docs/source/reference/howto/wipe_persistent_data.rst
+++ b/docs/source/reference/howto/wipe_persistent_data.rst
@@ -23,56 +23,28 @@ in the sequence given below, so you shouldn't proceed to next steps until the pr
 
 #. Add "wipe_environment_enabled: True" to `public.yml` file.
 
-
-#. Stop CommCare and close Postgres sessions.
+#. Stop CommCare
 
    .. code-block::
 
       $ cchq <env_name> service commcare stop
-      $ cchq <env_name> service postgresql stop
-      $ cchq <env_name> service postgresql status
 
-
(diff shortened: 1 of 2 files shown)


docs/source/reference/howto/wipe_persistent_data.rst
```diff
@@ -23,56 +23,28 @@ in the sequence given below, so you shouldn't proceed to next steps until the pr
...
-#. Wipe PostgreSQL databases
+#. Wipe PostgreSQL data
 
-   Check status & start postgres if NOT running.
+   Check status. Once status is "OK", wipe PostgreSQL data
 
    .. code-block::
 
       $ cchq <env_name> service postgresql status
-
```
Charl1996, line 54: What is the reason for removing this? Or is it simply that we're expecting the user to use the other parts of the docs for further debugging in case there's issues (which I think is OK since handling issues is probably out of scope here)

kaapstorm: > is it simply that we're expecting the user to use the other parts of the docs for further debugging in case there's issues

Yeah, pretty much this.

It turns out that the user will need to reset Postgres & PgBouncer anyway, so I removed these steps because they aren't useful.

(I'm opening follow-up tickets for commands given in this document, because there are at least two bugs (maybe more) that need to be fixed. So there will be more testing here, and this doc may change in the near future.)

Charl1996 (approved):

PR #6210, opened 2024-01-23
WAF whitelist XSS_BODY for logo upload

akashkj:
Add WAF rule for project basic settings in which a custom logo is uploaded.

Ticket: https://dimagi-dev.atlassian.net/browse/USH-4025
Commcare-HQ PR: https://github.com/dimagi/commcare-hq/pull/34017

##### Environments Affected

Production, India and Staging

### DIFF
--- a/src/commcare_cloud/commands/terraform/constants.py
+++ b/src/commcare_cloud/commands/terraform/constants.py
@@ -39,6 +39,7 @@ COMMCAREHQ_XML_POST_URLS_REGEX = r"""
 ^/a/([\w\.:-]+)/receiver/submission/?$
 ^/a/([\w\.:-]+)/reports/export/(case_list_explorer|duplicate_cases)/$
 ^/a/([\w\.:-]+)/settings/locations/import/$
+^/a/([\w\.:-]+)/settings/project/basic/$
 ^/a/([\w\.:-]+)/settings/users/commcare/fields/$
 ^/a/([\w\.:-]+)/settings/users/commcare/upload/$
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/terraform/constants.py
```diff
@@ -55,5 +56,5 @@
 """.strip().split('\n')
 
 COMMCAREHQ_XML_QUERYSTRING_URLS_REGEX = """
```
dannyroberts, line 58: I would suggest fixing the lint error not by escaping the `\` within the string, but by making this an r-string (starting with `r"""`) as described here https://docs.python.org/3/library/re.html:
> The solution is to use Python’s raw string notation for regular expression patterns; backslashes are not handled in any special way in a string literal prefixed with 'r'. So r"\n" is a two-character string containing '\' and 'n', while "\n" is a one-character string containing a newline. Usually patterns will be expressed in Python code using this raw string notation.

That's also how the block of regular expressions above is defined.

dannyroberts:
@akashkj can you update the PR description to include a ticket, list the environments it affects ("saas aws environments"), and remove the section about announcing a new release (since there's no announcement needed with this one because it doesn't affect third parties)?

dannyroberts:
Besides my two comments, this looks good to me. Ping me when it's ready for re-review

dannyroberts (approved):

akashkj:
@dannyroberts Other(hq) branch is merged.

PR #6211, opened 2024-01-24
Add Internal Feedback Email to LocalSettings

zandre-eng:
Link to ticket [here](https://dimagi-dev.atlassian.net/browse/SC-3317).

Adds a new internal feedback email to the local settings of all environments.

##### Environments Affected

All environments.

### DIFF
--- a/docs/source/reference/6-settings.rst
+++ b/docs/source/reference/6-settings.rst
@@ -117,3 +117,6 @@ soft_assert_email:
 new_domain_email:
     This address is notified when a user requests a project space or
     organization.
+
+internal_feedback_email:
+    Feedback by staff members for features/changes sent to this address.
--- a/docs/source/reference/6-settings.rst
+++ b/docs/source/reference/6-settings.rst
@@ -119,4 +119,8 @@ new_domain_email:
     organization.
 
 internal_feedback_email:
-    Feedback by staff members for features/changes sent to this address.
+    Internal feedback by staff members for features or changes is sent
+    to this address. For Dimagi-hosted environments this feedback is
+    sent to Dimagi's Solutions Division. Self-hosted environments can
+    choose to send feedback to a product manager or appropriate team.
+    Set to ``None`` in order to hide this option.
--- a/environments/64-test/public.yml
+++ b/environments/64-test/public.yml
@@ -88,6 +88,8 @@ feedback_email: droberts+feedback@dimagi.com
 eula_change_email: droberts+eula-notifications@dimagi.com
 contact_email: droberts+info@dimagi.com
 soft_assert_email: droberts+commcarehq-ops+soft_asserts@dimagi.com
+internal_feedback_email: solutions-feedback@dimagi.com
+
 daily_deploy_email: null
 
 commcare_cloud_root_user: ubuntu
--- a/environments/backup-production/public.yml
+++ b/environments/backup-production/public.yml
@@ -23,7 +23,7 @@ eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
 new_domain_email: inquiries@dimagi.com
-
+internal_feedback_email: solutions-feedback@dimagi.com
 DATADOG_ENABLED: False
 
 DATADOG_TRACE:
--- a/environments/india/public.yml
+++ b/environments/india/public.yml
@@ -108,6 +108,7 @@ eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
 new_domain_email: inquiries@dimagi.com
+internal_feedback_email: solutions-feedback@dimagi.com
 
 TWO_FACTOR_GATEWAY_ENABLED: True
 
--- a/environments/pna/public.yml
+++ b/environments/pna/public.yml
@@ -65,6 +65,7 @@ eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
 new_domain_email: inquiries@dimagi.com
+internal_feedback_email: solutions-feedback@dimagi.com
 
 KSPLICE_ACTIVE: yes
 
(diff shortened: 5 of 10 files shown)


docs/source/reference/6-settings.rst
```diff
@@ -117,3 +117,6 @@ soft_assert_email:
 new_domain_email:
     This address is notified when a user requests a project space or
     organization.
+
+internal_feedback_email:
```
mkangia, line 121: Hey @zandre-eng 

Just thinking about this change. It is very solutions division specific. If we think about this from an independent HQ environment perspective, this is a good change, though it's motivation is coming from a division specific feature request.
If two teams need separate email addresses, this would not be able to support it.

I suggest we don't add this to commcare-cloud as a generic/configurable email address but add a clear `solutions_internal_feedback_email` directly to localsettings if that is allowed or change the name to make it specific to solutions.
Additionally, you could limit settings this on dimagi environments only where `IS_DIMAGI_ENVIRONMENT` is true.

Let me know if that makes sense

mkangia (changes requested):

docs/source/reference/6-settings.rst
zandre-eng: I do like this idea as it would make it more obvious that this is specific to the Solutions division. The only thing that concerns me however, is that self-hosters would also have this variable in their HQ `settings.py` file, and so I thought it might be weird for them to have a variable specific to Dimagi.

I'm happy to change it though if, given the above, you still think it makes more sense to have it specific to Solutions?

mkangia: If there was a requirement to add an internal feedback email for environments, this change would make sense.
I am conflicted by how this is being used for a request meant to be sent to a specific sub team.

> The only thing that concerns me however, is that self-hosters would also have this variable in their HQ settings.py file

I hear you, that is why I suggested to limit to dimagi environments
>  Additionally, you could limit settings this on dimagi environments only where IS_DIMAGI_ENVIRONMENT is true.

Something like
```
if IS_DIMAGI_ENVIRONMENT:
    SOLUTIONS_FEEDBACK_EMAIL = "something@dimagi.com"
else:
    SOLUTIONS_FEEDBACK_EMAIL = NONE // or equivalent of blank to be used
```

kaapstorm: I am not a big fan of variable name debates. I'm not yet convinced that the cost of the time spent on a debate like this is worth the value of the variable name. Is the name accurate? It's okay. Is the name wrong? I don't think so.

But I do have a solution. I am keeping in mind that what we're looking at here is documentation, to be hosted on ReadTheDocs, and written for self-hosters. I looked at both lines:
```
internal_feedback_email:
    Feedback by staff members for features/changes sent to this address.
```

I think we could solve our conundrum by changing the description instead of the name. Maybe something like:
```
internal_feedback_email:
    Internal feedback by staff members for features or changes is sent
    to this address. For Dimagi-hosted environments this feedback is
    sent to Dimagi's Solutions Division. Self-hosted environments can
    choose to send feedback to a product manager or appropriate team.
    Set to ``None`` in order to hide this option.
```

docs/source/reference/6-settings.rst
```diff
@@ -117,3 +117,6 @@ soft_assert_email:
...
+    Feedback by staff members for features/changes sent to this address.
```
kaapstorm, line 122: (Adding as a suggestion to save you from a little copying and pasting.)
```suggestion
    Internal feedback by staff members for features or changes is sent
    to this address. For Dimagi-hosted environments this feedback is
    sent to Dimagi's Solutions Division. Self-hosted environments can
    choose to send feedback to a product manager or appropriate team.
    Set to ``None`` in order to hide this option.
```

kaapstorm (approved):

docs/source/reference/6-settings.rst
zandre-eng: +1 to changing the description. I think this is a nice middle ground between going very specific and being very generic with the variable name.

mkangia: Hey @kaapstorm 

thanks for jumping in.
Just confirming on one thing to ensure we are aligned.

My sense here was that we don't need this change that this PR is doing entirely. The motivation behind this PR is to allow adding an email address to be used for requests submitted on HQ to be sent to solutions team just like we have the generic "Report an issue" that is sent to a dimagi email.
My concern was that we are adding a generic internal feedback email on an environment and use that to send email for a specific sub team.
so my suggestion was to not add this customizable email but instead directly add a `SOLUTIONS_FEEDBACK_EMAIL`.

The naming of the variable was not the concern here.

Charl1996 (approved):

docs/source/reference/6-settings.rst
kaapstorm: Okay. So just to make sure I understand you correctly, what you're suggesting is:
1. Close this PR
2. Set the value of the constant (name irrelevant) to the email address of the Solutions team _directly in the `commcare-hq` codebase_, so that it will be the same in all environments, including non-Dimagi environments, similar to what we do with soft asserts.
3. (Optionally) Disable the functionally in non-Dimagi environments.

:thinking: It is simpler.

I don't have a problem with that. I think it meets the spec, which states,
> The purpose of this work is to ... enable Internal Dimagi Users ... to make product feature suggestions / changes.

mkangia: that is correct.

> (Optionally) Disable the functionally in non-Dimagi environments.

Good point. We can disable it for non-dimagi environments in HQ itself by checking on `IS_DIMAGI_ENVIRONMENT`. It is same level of check as I was suggesting above with a benefit of no changes to CommCare Cloud.

mkangia: @zandre-eng let me know if this sits well with you or you wish to continue with current approach. happy to hear it out.

zandre-eng: Ah, I must admit I also slightly misunderstood the approach you were suggesting @mkangia, but on the same page now. Definitely in agreement over closing this PR off and having the email sit inside CommCare HQ itself.

> (Optionally) Disable the functionally in non-Dimagi environments.

If we are going to hard code it to be the Solutions email, then definitely +1 for disabling in non-Dimagi environments. This will help prevent self-hosters from potentially spamming the feedback report.

zandre-eng:
Given [this](https://github.com/dimagi/commcare-cloud/pull/6211#discussion_r1465922233) comment thread, I'll be closing off this PR as we will no longer be storing the Solutions email variable in the `localsettings.py` file, but rather as a const in CommCare HQ itself.

docs/source/reference/6-settings.rst
mkangia: Great! Glad we were eventually able to come on the same page with this.

PR #6214, opened 2024-01-29
Update documentation around migrating data

gherceg:
After going through this process a number of times, this is where I've landed. I've tried to simplify the steps (take out what seemed unnecessary), and added/updated anything that caused issues but was needed.

##### Environments Affected

None

PR #6216, opened 2024-01-31
More doc updates applicable to migrating data to a new instance

gherceg:
Followup to https://github.com/dimagi/commcare-cloud/pull/6214
##### Environments Affected

None

### DIFF
--- a/docs/source/installation/2-manual-install.rst
+++ b/docs/source/installation/2-manual-install.rst
@@ -588,7 +588,11 @@ initially.
    ::
 
        $ commcare-cloud cluster deploy
-       $ Or if you need to deploy a specific version of CommCare HQ as opposed to the latest:
+
+   Or if you need to deploy a specific version of CommCare HQ as opposed to the latest:
+
+   ::
+
        $ commcare-cloud cluster deploy --commcare-rev=<commit-hash>
 
    When prompted for the ``sudo`` password, enter the
--- a/docs/source/installation/2-manual-install.rst
+++ b/docs/source/installation/2-manual-install.rst
@@ -588,6 +588,8 @@ initially.
    ::
 
        $ commcare-cloud cluster deploy
+       $ Or if you need to deploy a specific version of CommCare HQ as opposed to the latest:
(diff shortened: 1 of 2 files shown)


mkangia (approved):

docs/source/installation/2-manual-install.rst
```diff
@@ -588,6 +588,8 @@ initially.
    ::
 
        $ commcare-cloud cluster deploy
+       $ Or if you need to deploy a specific version of CommCare HQ as opposed to the latest:
```
kaapstorm, line 591:
```suggestion

   Or if you need to deploy a specific version of CommCare HQ as opposed to the latest
   
   ::

```

kaapstorm (approved):
Just the one formatting suggestion (cos otherwise the instruction looks like a command for the user to type at the command prompt).

docs/source/installation/2-manual-install.rst
gherceg: 92056e653c92d2b3fe15e311ef081a1dd6654595

docs/source/installation/2-manual-install.rst
```diff
@@ -589,6 +589,12 @@ initially.
 
        $ commcare-cloud cluster deploy
 
+   Or if you need to deploy a specific version of CommCare HQ as opposed to the latest:
+
+   ::
+
+       $ commcare-cloud cluster deploy --commcare-rev=<commit-hash>
```
minhaminha, line 596: not blocking but might be good to clarify the specific scenarios in which you'll need to deploy with a specific version, unless they read this already knowing they have to (like if they were redirected from the project migration docs or something)

minhaminha (approved):

mkangia (approved):

docs/source/installation/2-manual-install.rst
gherceg: Yeah the docs for a data migration mention that they will need to deploy from a specific commit, and we will share a commit hash with them which should also remind them that they will need to use it. I'd rather not add too much to the manual install docs about this, but I imagine there is still room for improvement.

PR #6219, opened 2024-02-06
Add removing kafka data for debugging kafka issues

gherceg:
This line is also necessary when "resetting" kafka once it has gotten into a broken state. I don't love that I hard code the paths here since theoretically a self hoster could setup a different path, but hopefully they would be aware of that and able to make the necessary change for their own system.

##### Environments Affected

None

### DIFF
--- a/docs/source/reference/howto/wipe_persistent_data.rst
+++ b/docs/source/reference/howto/wipe_persistent_data.rst
@@ -134,4 +134,5 @@ Issues with check_services
     
      $ cchq monolith service kafka stop
      $ rm -rf /var/lib/zookeeper/*
+     $ rm -rf /opt/data/kafka/data/*
      $ cchq monolith service kafka restart


calellowitz (approved):

docs/source/reference/howto/wipe_persistent_data.rst
```diff
@@ -134,4 +134,5 @@ Issues with check_services
     
      $ cchq monolith service kafka stop
      $ rm -rf /var/lib/zookeeper/*
+     $ rm -rf /opt/data/kafka/data/*
```
millerdev, line 137: > don't love that I hard code the paths here

Would it be worth adding a comment that the paths may need to be adjusted, depending on how Kafka is configured?

gherceg: Yeah probably: https://github.com/dimagi/commcare-cloud/pull/6221

PR #6225, opened 2024-02-15
Update COMMCARE_ANALYTICS_HOST for staging to point to the staging host

SmittieC:
No ticket, I just noticed that the link to the analytics server is broken. We have a staging instance, but staging doesn't point to that. This update should fix this.

##### Environments Affected
Staging

##### Announce New Release
No announcement needed, since this is staging

### DIFF
--- a/environments/staging/public.yml
+++ b/environments/staging/public.yml
@@ -127,7 +127,7 @@ localsettings:
   BANK_NAME: "RBS Citizens N.A."
   BANK_SWIFT_CODE: 'CTZIUS33'
   CELERY_PERIODIC_QUEUE: 'celery_null'
-  COMMCARE_ANALYTICS_HOST: "http://commcare-analytics-demo.dimagi.com"
+  COMMCARE_ANALYTICS_HOST: "https://commcare-analytics-staging.dimagi.com"
   COUCH_CACHE_DOCS: True
   COUCH_CACHE_VIEWS: True
   COUCH_PASSWORD: "{{ COUCH_PASSWORD }}"


Charl1996 (approved):

mkangia (approved):
The code change looks good.
thanks for additional https change.

Do we need to test this to make sure the integration with staging is actually working? I assume there might be other factors to update like some tokens/keys?

Charl1996:
> Do we need to test this to make sure the integration with staging is actually working?

Good idea

> I assume there might be other factors to update like some tokens/keys?

No, it's only the link.

mkangia:
So we update this via ansible or manually on the server? good to test the integration before merging it.

SmittieC:
> So we update this via ansible or manually on the server? good to test the integration before merging it.

The interaction between the staging analytics server (https://commcare-analytics-staging.dimagi.com) and the HQ staging server is already working, so the integration is already tested. [This link in the code](https://github.com/dimagi/commcare-hq/blob/e4a0f4aed43d19b857937c70a69630360efd0ad2/corehq/apps/export/templates/export/commcare_analytics.html#L20) is the only thing that needs to be updated. It gets its value from the COMMCARE_ANALYTICS_HOST env variable, specified by CC. It's just a link that the user can click on to go to the staging analytics server.

mkangia:
Cool. So we would just update the localsettings on staging by running commcare cloud command.

kaapstorm (approved):

SmittieC:
Thanks for merging @kaapstorm. I forgot about this PR :eyes:

PR #6229, opened 2024-02-17
[India] Set broker heartbeat to match timeout specified in rabbitmq.conf

gherceg:
https://dimagi.atlassian.net/browse/SAAS-15294

We've been seeing timeout issues on the India environment, likely due to some long running tasks on the `reminder_queue`. My understanding is that [this code](https://github.com/dimagi/commcare-cloud/blob/3c579e8885e72346cfb91062fbd0e8436db470ae/src/commcare_cloud/ansible/roles/rabbitmq/templates/rabbitmq.conf.j2#L6-L6) sets the timeout on the server side (from rabbitmq's perspective), but the client (celery) also specifies a timeout value, and rabbit and celery agree to use the lower of the two values (docs [here](https://www.rabbitmq.com/heartbeats.html#heartbeats-timeout)).

So this PR's aim to specify the [broker_heartbeat](https://docs.celeryq.dev/en/v5.2.7/userguide/configuration.html#broker-heartbeat) value from celery's side so that rabbit and celery agree on a value of 600 seconds.

##### Environments Affected

India

### DIFF
--- a/environments/india/public.yml
+++ b/environments/india/public.yml
@@ -122,7 +122,7 @@ localsettings:
   BANK_ADDRESS: { 'first_line': "1 Citizens Drive", 'city': "Riverside", 'region': "RI", 'postal_code': "02915" }
   BANK_NAME: "RBS Citizens N.A."
   BANK_SWIFT_CODE: 'CTZIUS33'
-  BROKER_HEARTBEAT: 600
+  CELERY_BROKER_HEARTBEAT: 600
 #  COUCH_CACHE_DOCS:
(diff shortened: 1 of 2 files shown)


millerdev (approved):

gherceg:
Updated to reflect that we namespace celery settings with "CELERY" (code [here](https://github.com/dimagi/commcare-hq/blob/0add2cfe6167bd9c2b701ec888edaa601e5af98b/corehq/apps/celery/__init__.py#L25-L25)), so made a change to support that.

millerdev (approved):

PR #6237, opened 2024-03-04
Increase Indicator Queue Count Variable

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/SC-3498).

This change is part of an urgent request to rebuild a UCR datasource. To help speed up the process, we are increasing the amount of indicators that can be queued to 20k (default is 10k).

##### Environments Affected

production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -232,6 +232,7 @@ localsettings:
   USER_REPORTING_METADATA_BATCH_ENABLED: True
   WS4REDIS_CONNECTION_HOST: "redis.production.commcare.local"
   CONNECTID_URL: 'https://connectid.dimagi.com/o/userinfo'
+  ASYNC_INDICATORS_TO_QUEUE: 20000
 
 # comment these two lines out to make a new rackspace machine
 commcare_cloud_root_user: ubuntu
@@ -256,5 +257,3 @@ shared_dir_gid: 1500
 # for better time syncronization on aws servers
 # replace ntp with chrony.
 use_chrony: True
-
-ASYNC_INDICATORS_TO_QUEUE: 20000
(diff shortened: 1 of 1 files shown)


mkangia:
Hey @calellowitz 
Based on our recent discussions, we are trying to run the async indicators with a bumped up number while not pushing it too far ahead as well.
Let me know if this sounds like a safe update to you.

mkangia (approved):

environments/production/public.yml
```diff
@@ -256,3 +256,5 @@ shared_dir_gid: 1500
 # for better time syncronization on aws servers
 # replace ntp with chrony.
 use_chrony: True
+
+ASYNC_INDICATORS_TO_QUEUE: 20000
```
calellowitz, line 260: This needs to be part of the `localsettings` object for it to be included in the localsettings https://github.com/dimagi/commcare-cloud/blob/master/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2#L882

calellowitz (changes requested):

environments/production/public.yml
zandre-eng: Ah, good catch. I've moved it under the localsettings object ([c8c14a8](https://github.com/dimagi/commcare-cloud/pull/6237/commits/c8c14a8a42f6420f5173fc0bf6559b9fe5d51338)).

calellowitz (approved):
nit: these are mostly in alphabetical order and it would be nicer to add this in its alphabetical place, but i recognize there is another that has been added out of order at the end so not blocking

environments/production/public.yml
mkangia: Thanks for catching that @calellowitz

PR #6240, opened 2024-03-06
Various documentation updates for data migration

gherceg:
I forgot I had https://github.com/dimagi/commcare-cloud/pull/6216 open which has some overlapping changes. I'm going to move forward with this one first, and then resolve the conflicts on that PR separately.

There were a couple of defective links that I resolved as well as making it more clear to readers that services should not be started back up until after loading data into the new environment. I also replaced the deploy step with a simple downtime end call because a deploy step is now included when starting the steps for rebuilding an environment.

##### Environments Affected

None

### DIFF
--- a/docs/source/installation/migration/1-migrating-project.rst
+++ b/docs/source/installation/migration/1-migrating-project.rst
@@ -152,6 +152,8 @@ will be unable to submit forms or sync with the server.
 * Proceed to step 4.
 
 
+.. _import-data-into-environment:
+
 4. Import the data to the new environment
 -----------------------------------------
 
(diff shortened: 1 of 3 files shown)


Charl1996 (approved):

minhaminha (approved):

gherceg:
I made another change to make it even more clear that Start new environment should not be run if migrating, so now it looks like:
image

Does this seem like overkill?

minhaminha:
> Does this seem like overkill?

You can never be too careful with user facing docs!

minhaminha (approved):

gherceg:
Sorry one more change. Made the text for "not" bold.

minhaminha (approved):

PR #6256, opened 2024-04-09
add custom properties to fomplayer config for DB connection pool

snopoke:
Change request: https://dimagi.atlassian.net/browse/SAAS-15447

Update DB connection pool configuration on production:

- maximum-pool-size
  - Change from the default of 10 to 20 to better handle peak usage
- minimum-idle
  - Change from the default of 10 to 5 to reduce off peak DB connections and connections from instances receiving less requests

Config options docs: https://github.com/brettwooldridge/HikariCP?tab=readme-ov-file#gear-configuration-knobs-baby

##### Environments Affected
Production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -75,6 +75,9 @@ formplayer_forward_ip_proxy: true
 formplayer_detailed_tags:
   - form_name
   - module_name
+formplayer_custom_properties:
+  spring.datasource.hikari.maximum-pool-size: 20
+  spring.datasource.hikari.minimum-idle: 5
 
 KSPLICE_ACTIVE: yes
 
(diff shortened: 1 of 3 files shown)


dannyroberts (approved):

gherceg:
Can you talk more about what sparked the decrease in minimum-idle? Based on docs:
> for maximum performance and responsiveness to spike demands, we recommend not setting this value and instead allowing HikariCP to act as a fixed size connection pool. Default: same as maximumPoolSize

I can understand that it is a waste to maintain unused db connections when usage is lower, but if the goal is to better handle peak usage, why decrease the minimum number of idle connections?

snopoke:
> Can you talk more about what sparked the decrease in minimum-idle? Based on docs:
> 
> > for maximum performance and responsiveness to spike demands, we recommend not setting this value and instead allowing HikariCP to act as a fixed size connection pool. Default: same as maximumPoolSize
> 
> I can understand that it is a waste to maintain unused db connections when usage is lower, but if the goal is to better handle peak usage, why decrease the minimum number of idle connections?

The guidance is aimed at applications requiring very low latency. For Formplayer I do not expect the overhead of creating new connections on occasion to be an issue. New connections will remain in the pool until they have been idle for 10 minutes so during peak there should only be 10 requests that bear the cost of creating a new connection.

I'm also cautious to increase the total number of connections across all the 20 FP machines in prod. The traffic is not spread evenly across all nodes so it doesn't make sense to have all of them open max connections which may impact the DB's performance since each connections get's memory assigned to it.

PR #6260, opened 2024-04-12
Remove references to Ubuntu 18.04 and installation of Python 3.10

gherceg:
This is long overdue, but we don't have any control machines running 18.04 and self hosters shouldn't either, so we should cleanup our relevant scripts to only handle 22.04.

I also noticed the test container Daniel added was using 18.04, and updated that, but admittedly don't have context on that change. @millerdev if that is outside of the scope of this PR let me know and I can remove it. 

Before merging:

- test this (maybe use vagrant locally, or a new machine on staging since we have other control setup issues to test)
- is there anything to announce? We already said we wouldn't support 18.04 as of July 1st, 2023, so I think no announcement is needed.

##### Environments Affected

None

##### Announce New Release


docs/source/reference/1-commcare-cloud/1-installation.rst
```diff
@@ -117,10 +117,7 @@ If you get to this point, congratulations! ``commcare-cloud`` is installed.
 Manual Installation
 -------------------
 
-You will need python 3.10 installed to follow these instructions. See
-:ref:`installation/2-manual-install:Upgrade to Python 3.10` for instructions on
```
millerdev, line 121: Did the target of this link get removed earlier? I could not find it.

millerdev (approved):
> I think no announcement is needed.

I think I agree, although what would happen if someone was still running 18.04? Will their control machine still work? If ~yes~ not, should we add a check for that and print an error about unmet requirements?

docs/source/reference/1-commcare-cloud/1-installation.rst
gherceg: yeah I think it was removed [here](https://github.com/dimagi/commcare-cloud/pull/6070/files)

PR #6274, opened 2024-04-26
Disable Formplayer APM Traces for Memory Testing on Prod

ctsims:
Since enabling APM in February we've seen a change to the memory pattern on the formplayer workers
image

Temporarily disables this deep APM Tracing and Profiling  so we can evaluate load tests to see if this pattern reverts.

##### Environments Affected
Production


ctsims:
Duplicate of #6273

PR #6276, opened 2024-04-29
Change mobile worker instance type to c6a.8xlarge

AmitPhulera:
Temporarily change the instance type to have more CPUs
##### Environments Affected

Production


gherceg (approved):
Just noting that we compared this to other instance families to see if there was a better option, but this was the most cost friendly option to increase CPUs.

PR #6284, opened 2024-05-05
Added webpack to the HQ deploy process

mjriley:
Associated ticket: https://dimagi.atlassian.net/browse/SAAS-15485

Adds support for running webpack while deploying HQ, if a webpack configuration file is found. This is not currently required by any live code, but allows us to test react and vue (and other javascript features) in the future

##### Announce New Release

### DIFF
--- a/src/commcare_cloud/ansible/deploy_hq.yml
+++ b/src/commcare_cloud/ansible/deploy_hq.yml
@@ -48,10 +48,10 @@
   hosts: [webworkers, proxy]
   any_errors_fatal: true
   tasks:
-    - import_role: {name: deploy_hq}
+    - include_role: {name: deploy_hq}
       vars:
         if_not_done: run_webpack
-      when: use_webpack is defined and use_webpack
+      when: use_webpack | default(false)
   
 
 - name: Collect static files
--- a/src/commcare_cloud/ansible/deploy_hq.yml
+++ b/src/commcare_cloud/ansible/deploy_hq.yml
@@ -45,7 +45,7 @@
         if_not_done: yarn_install
 
 - name: Run webpack
-  hosts: [proxy]
+  hosts: [webworkers, proxy]
   any_errors_fatal: true
   tasks:
     - import_role: {name: deploy_hq}
--- a/src/commcare_cloud/ansible/deploy_hq.yml
+++ b/src/commcare_cloud/ansible/deploy_hq.yml
@@ -45,7 +45,7 @@
         if_not_done: yarn_install
 
 - name: Run webpack
-  hosts: [webworkers]
+  hosts: [proxy]
   any_errors_fatal: true
   tasks:
     - import_role: {name: deploy_hq}
--- a/src/commcare_cloud/ansible/deploy_hq.yml
+++ b/src/commcare_cloud/ansible/deploy_hq.yml
@@ -44,6 +44,14 @@
       vars:
         if_not_done: yarn_install
 
+- name: Run webpack
+  hosts: [webworkers]
+  any_errors_fatal: true
+  tasks:
+    - import_role: {name: deploy_hq}
+      vars:
+        if_not_done: run_webpack
+
 - name: Collect static files
   hosts: [webworkers, proxy]
   any_errors_fatal: true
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/deploy_hq.yml
```diff
@@ -44,6 +44,14 @@
       vars:
         if_not_done: yarn_install
 
+- name: Run webpack
+  hosts: [webworkers, proxy]
+  any_errors_fatal: true
+  tasks:
+    - import_role: {name: deploy_hq}
+      vars:
+        if_not_done: run_webpack
```
millerdev, line 53: What do you think of adding the tasks in this file to `staticfiles_collect.yml` file? It would reduce deploy overhead to group them together since they are applied to the same set of hosts. This is doing the same kind of thing as other staticfiles collection tasks, so I think it makes sense.

mjriley: Is the primary motivation to reduce overhead or to create natural groups? From running this deploy process on staging, I don't think overhead should be a concern -- the webworkers task executed very quickly. The bottleneck is the `compilejs18n` command from _staticfiles_collect.yml_. If the motivation is instead to create natural groups, I'd argue that this grouping has the potential to create a lot of confusion. Webpack doesn't collect any static files. Instead, it is generally known as the step to compress/minify, and our dedicated "Compress static files" task lives outside of "Collect static files".  I also wouldn't recommend combining tasks just because they use the same hosts. I suppressed a warning by including webworkers here, but in my head, this should really only be deployed to proxy servers.

If we wanted to split off a playbook dedicated to javascript tasks, that would make sense, but I think dedicated tasks for webpack make sense

millerdev: The overhead I'm speaking of is SSH connections made by Ansible, not the time it takes to actually run webpack.

`staticfiles_collect.yml` is where we do other js-related build steps (example: `build_requirejs`), so I think it makes sense to put webpack in there too since that is the same type of task.

mjriley: I've updated the associated ticket with logs from 2 runs to compare what changes when webpack is run from the deploy_hq task vs from within staticfiles_collect. I don't believe there is any difference with respect to SSH connections. From what I can tell, ansible will create connections for every task, and it is thus advised to configure SSH to [re-use connections](https://docs.ansible.com/ansible/latest/reference_appendices/faq.html#how-do-i-get-ansible-to-reuse-connections-enable-kerberized-ssh-or-have-ansible-pay-attention-to-my-local-ssh-config-file).

I continue to believe that moving webpack to _staticfiles_collect_ would be confusing. Collecting into static is [django terminology](https://docs.djangoproject.com/en/5.0/ref/contrib/staticfiles/#collectstatic). I think it's unintuitive that we're building requireJS in this playbook, and I don't think we should double down here. Webpack deserves its own task file because it is capable of much more than collecting, and those responsibilities may not align wit the same hosts that collecting does. In addition, it is essentially in feature-flag status at the moment. Including it as I'm doing minimizes log output for runs that do not use webpack.

millerdev: Looks like there is one extra SSH connection (`Check if done: run_webpack`) when it's included in a separate block in this file. Admittedly not much overhead there, but it's something.

For me putting it together with other tasks doing similar things (currently in `staticfiles_collect.yml`) is the more important point since that's where people who are familiar with the existing structure would go to look for that type of task. I could see an argument for renaming `staticfiles_collect.yml` to something more generic that includes JS bundling. Adding another top-level block in this file seems overkill for a feature flagged experiment.

millerdev: Noting that `build_requirejs` is the last task in `staticfiles_collect.yml`, I think because it depends `collectstatic`. Is there any chance (now or in the future) that webpack could also depend on `collectstatic` being run first?

mjriley: I looked at this, and I'm not entirely sure why requireJS runs after `collectstatic`. That said, webpack is creating artifacts that django will later serve, so those artifacts should be created before static assets are collected. I previously modified the webpack configuration such that webpack is able to follow django app paths, thus it looks for the location of files _before_ collecting them into a single static folder

PR #6293, opened 2024-05-10
Changed webpack command to be more generic

mjriley:
##### Environments Affected
Staging (only if webpack is being used)

Changed the webpack command from `yarn build:react` to `yarn build:webpack` to be more accommodating if we use some framework other than React.


gherceg (approved):

gherceg:
Just clarifying that the only reason this is limited to staging is because that is where a webpack config exists?

PR #6294, opened 2024-05-11
Swap case search index

AmitPhulera:
swaps case search index on production.
##### Environments Affected

Prod


AmitPhulera:
This is applied on production.

PR #6296, opened 2024-05-15
Stop multiplexing on case-search index on Prod

AmitPhulera:
After prod has been successfully reindexed, we are turning off multiplexer on case search index
##### Environments Affected

Production


environments/production/public.yml
```diff
@@ -171,7 +171,7 @@ localsettings:
   ES_MULTIPLEX_TO_VERSION: '6'
   # Index Multiplexer Settings  
   ES_APPS_INDEX_MULTIPLEXED: False
-  ES_CASE_SEARCH_INDEX_MULTIPLEXED: True
+  ES_CASE_SEARCH_INDEX_MULTIPLEXED: False # Index swapped
```
millerdev, line 174: Should `ES_CASE_SEARCH_INDEX_SWAPPED: True` be reset to `False` now that this is done?

AmitPhulera: Nope. We will need keep Swapped as `True` until we change primary and secondary index names in consts file in es app.

PR #6299, opened 2024-05-16
Add changelog for copying "Web User " Permission to "Manage Tableau Config" Permission

Jtang-1:
[PR](https://github.com/dimagi/commcare-hq/pull/34468) creates a migration the depends on an imported model. The historical model can not be used since the migration also depends on a custom method. Therefore, it potentially needs to be manually run.

##### Environments Affected

Only environments that use the [TABLEAU SYNC FF ](https://www.commcarehq.org/hq/flags/edit/tableau_user_syncing/)

##### Announce New Release

- [X] After merging, I will follow [these instructions](https://confluence.dimagi.com/display/saas/Announcing+changes+affecting+third+parties#Announcingchangesaffectingthirdparties-announcing)


snopoke (commented):
This seems like a bad way to deal with the limitation of custom methods on historical models.

I also doubt it will be the last time we need to do this sort of migration so I'd much prefer to see a method that allows us to do it safely in a Django migration.

See my comment on the HQ PR: https://github.com/dimagi/commcare-hq/pull/34468#pullrequestreview-2063122492

PR #6304, opened 2024-05-22
Give consultant access to staging

jingcheng16:
Ticket: https://dimagi.atlassian.net/browse/SAAS-15439

##### Environments Affected

Staging

Will run `cchq --control staging update-users` after this is merged.


environments/staging/meta.yml
```diff
@@ -7,5 +7,6 @@ users:
   - g2
   - fissionlabs
   - fission-warehouse
+  - staging
```
millerdev, line 10: I don't understand why `staging` would be added here. Should this be `frener`?

jingcheng16: Hi Daniel, because I add frener in [staging.yml](https://github.com/dimagi/commcare-cloud/blob/136b671ef02f07e3d9e05753c08f826fe43b31ea/environments/_users/staging.yml)

millerdev: Oh, I see, this is a list of yaml files (without `.yml` extensions), not actual users. Makes sense now.

PR #6305, opened 2024-05-23
Starting syncing writes to sub-index of case-search on staging

AmitPhulera:
Adding this PR to test a separate index for `co-carecoordination-test` on staging env.
##### Environments Affected

staging


AmitPhulera:
This is already applied on staging.

PR #6313, opened 2024-06-07
Ensure that SMTP message body is not empty

kaapstorm:
Context: [Forum](https://forum.dimagi.com/t/email-error-when-deploying/10812)

Email notifications of both the start of a deploy, and of a failed deploy, have empty message bodies. Azure Communication Services rejects emails with empty message bodies with the following exception:

> smtplib.SMTPDataError: (501, b'5.6.0 Request body validation error. Need either non-empty Html or PlainText body to be present.')

##### Environments Affected

Only affects environments that use Azure Communication Services for SMTP. No Dimagi environments are affected.

### DIFF
--- a/src/commcare_cloud/commands/deploy/utils.py
+++ b/src/commcare_cloud/commands/deploy/utils.py
@@ -99,16 +99,19 @@ def announce_deploy_success(environment, context):
     )
 
 
-def send_email(environment, subject, message='', to_admins=True, recipients=None):
+def send_email(environment, subject, message, to_admins=True, recipients=None):
     """
     Call a Django management command to send an email.
 
-    :param environment: The Environement object
+    :param environment: The Environment object
     :param subject: Email subject
-    :param message: Email message
+    :param message: Email message body
     :param to_admins: True if mail should be sent to Django admins
     :param recipients: List of additional addresses to send mail to
     """
+    if not message:
+        raise ValueError('Some cloud hosting providers require a message body')
+
     if environment.fab_settings_config.email_enabled:
         print(color_summary(f">> Sending email: {subject}"))
         args = [
--- a/src/commcare_cloud/commands/deploy/utils.py
+++ b/src/commcare_cloud/commands/deploy/utils.py
@@ -100,7 +100,7 @@ def send_email(environment, subject, message='', to_admins=True, recipients=None
     """
     Call a Django management command to send an email.
 
-    :param environment: The Environment object
+    :param environment: The Environement object
     :param subject: Email subject
     :param message: Email message
     :param to_admins: True if mail should be sent to Django admins
@@ -108,11 +108,6 @@ def send_email(environment, subject, message='', to_admins=True, recipients=None
     """
     if environment.fab_settings_config.email_enabled:
         print(color_summary(f">> Sending email: {subject}"))
-
-        if not message:
-            # Azure Communication Services require a message body
-            message = subject
-
         args = [
             message,
             '--subject', subject,
--- a/src/commcare_cloud/commands/deploy/utils.py
+++ b/src/commcare_cloud/commands/deploy/utils.py
@@ -60,27 +60,30 @@ def send_deploy_start_email(environment, context):
         and context.revision is not None
     )
     env_name = environment.meta_config.deploy_env
-    subject = f"{context.user} has initiated a {context.service_name} deploy to {env_name}"
+    message = f"{context.user} has initiated a {context.service_name} deploy to {env_name}"
     prefix = ""
     if is_nonstandard_deploy_time:
-        subject += " outside maintenance window"
+        message += " outside maintenance window"
         prefix = "ATTENTION: "
     if is_non_default_branch:
-        subject += f" with non-default branch '{context.revision}'"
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/commands/deploy/utils.py
```diff
@@ -100,14 +100,19 @@ def send_email(environment, subject, message='', to_admins=True, recipients=None
...
     :param subject: Email subject
     :param message: Email message
     :param to_admins: True if mail should be sent to Django admins
     :param recipients: List of additional addresses to send mail to
     """
     if environment.fab_settings_config.email_enabled:
         print(color_summary(f">> Sending email: {subject}"))
+
+        if not message:
+            # Azure Communication Services require a message body
```
mkangia, line 113: Oh, its only for Azure.

I'd say we don't make any change for this.
To cater to a restriction to on email provider, we are populating emails for all providers.
& I don't believe that empty email body is unusual, so its not clear why Azure isn't happy about it.

If we absolutely need to fix this, we could just catch the exception from provider and give a nicer message though the exception itself seems totally okay too since this is not because of any issues with commcare-cloud.

So, I'd say we don't address this error.

kaapstorm: > Oh, its only for Azure.

According to [HG Insights](https://hginsights.com/blog/microsoft-azure-market-share-2024):
> In 2024, Azure market share reached 24% of the global cloud computing market and their customer base grew by 14.2% from 2023 to 2024.

...

> I don't believe that empty email body is unusual, so its not clear why Azure isn't happy about it.

Microsoft's implementations of Internet standards is an old and interesting discussion, but it seems only academic. I think we have to be pragmatic. What is the best way to deal with the situation we have?

I think there are two facts that persuade me that this change is good:

1. _Azure has a quarter of the cloud market_, so it is not unrealistic to expect that some of Dimagi's self-hosting partners will use Azure to host HQ.
2. An important partner _is_ using Azure to host HQ _right now_.

I don't understand why you are recommending that we prevent our self-hosting partners from receiving notifications. That sounds like a footgun to me.

Alternatively, if you feel that _this specific_ implementation of a solution is incorrect, but that in general helping partners is a good idea, then would you be amenable to changing the notifications for start and failure to have message bodies, since the notification for success has a message body already?

kaapstorm:
I'm curious what others think, @dannyroberts @gherceg ? How do you feel about setting a message body if it is empty? Alternatively, should we pass a message body to `send_email()` in the two places where it is called without one? Or is it best not to send notifications when HQ is hosted on Azure?

kaapstorm:
Bump. @snopoke and @AmitPhulera I noticed you've worked in this module. Any thoughts or opinions on this change?

src/commcare_cloud/commands/deploy/utils.py
mkangia: Hey @kaapstorm 

Solid facts about the Azure usage, I was not aware its that widely used.

> I don't understand why you are recommending that we prevent our self-hosting partners from receiving notifications. That sounds like a footgun to me.

Definitely not saying we should not notify. I see that commcare cloud is attempting to send the notification, but its failing due to restrictions at the email provider. I am aligned that we should make reasonable changes to cater for requirements of different integrations we have. My hesitation was to modify implementations for all email providers to cater for one email provider.

> would you be amenable to changing the notifications for start and failure to have message bodies

I see that would resolve the concern currently for these emails. In future, if another email is added without a message body, the same error is going to appear. And I see that the change you are making takes care of that.

If here we acknowledge that some email providers block emails that don't have a body & in commcare cloud we want to take care of that generally & to cater for that we would use the subject as the email body if its missing, I am okay with that.

An alternate could be to
1. add email body for the emails that don't have one
2. raise an error if message body is missing so that the dev can add one

Moreover, happy to go with whatever the team decides.

AmitPhulera:
@kaapstorm The change looks reasonable to me. I also looked briefly into it while going through the forum and I had a similar approach in mind for the fix.
One question, do we know how it changes the deploy emails? Like how they used to appear before and how they would appear now after this change is applied.

src/commcare_cloud/commands/deploy/utils.py
kaapstorm: I like that idea. https://github.com/dimagi/commcare-cloud/pull/6313/commits/358dcce7f83f2328c2c9596e07b21f23402c5ee5 and https://github.com/dimagi/commcare-cloud/pull/6313/commits/19547c9701d4d680ca6c55bbf22abe70ce12bd34

gherceg:
Feels a bit weird to have the same subject and body for an email, but otherwise I think it is a simple enough change that it is worth doing for the reasons you outlined above.

mkangia (approved):
Thanks for the redirection @kaapstorm 

Also, I will be cool if you would like to drop the commit and the revert commit to keep this PR shorted with relevant commits only

kaapstorm:
> drop the commit and the revert commit to keep this PR shorted with relevant commits only

GitHub makes this easy with "Squash and merge" :) It did squash the 3rd and 4th commits too, but I think that's okay.

PR #6326, opened 2024-06-24
Route /hq/admin/phone/restore requests to mobile webworkers

gherceg:
https://dimagi.atlassian.net/browse/SAAS-15627

This may not be as targeted as we want. We want to route all admin/phone/restore requests to mobile webworkers. Currently, the only admin/phone/ requests defined in admin/urls.py are the two restore requests (app aware and not).

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -67,7 +67,15 @@ nginx_sites:
       proxy_buffers: 8 64k
       proxy_buffer_size: 64k
       client_body_buffer_size: 512k
-    - name: '~ /a/[^/]+/(receiver|phone|admin/phone)/'
+    - name: '~ /a/[^/]+/(receiver|phone)/'
+      balancer: "{{ deploy_env }}_commcare_submissions"
+      proxy_redirect: "http://{{ SITE_HOST }} https://{{ SITE_HOST }}"
+      proxy_next_upstream_tries: 1
+      proxy_read_timeout: 900s
+      proxy_buffers: 8 64k
+      proxy_buffer_size: 64k
+      client_body_buffer_size: 512k
+    - name: '~ /hq/admin/phone/restore/'
       balancer: "{{ deploy_env }}_commcare_submissions"
       proxy_redirect: "http://{{ SITE_HOST }} https://{{ SITE_HOST }}"
       proxy_next_upstream_tries: 1
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -67,15 +67,7 @@ nginx_sites:
       proxy_buffers: 8 64k
       proxy_buffer_size: 64k
       client_body_buffer_size: 512k
-    - name: '~ /a/[^/]+/(receiver|phone)/'
-      balancer: "{{ deploy_env }}_commcare_submissions"
-      proxy_redirect: "http://{{ SITE_HOST }} https://{{ SITE_HOST }}"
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -67,7 +67,7 @@ nginx_sites:
       proxy_buffers: 8 64k
       proxy_buffer_size: 64k
       client_body_buffer_size: 512k
-    - name: '~ /a/[^/]+/(receiver|phone)/'
+    - name: '~ /a/[^/]+/(receiver|phone|admin/phone)/'
```
dannyroberts, line 70: Is this the right URL? I'm seeing `/hq/admin/phone/restore/`. Does the one you have here also work?

dannyroberts: Matches https://dimagi.atlassian.net/wiki/spaces/GS/pages/2144317344/Restore+File+and+Configuration+Options (internal docs) too.

dannyroberts: In the code I also see `/a/<domain>/phone/admin_restore/` here https://github.com/dimagi/commcare-hq/blob/44b56b86f8e750f9e0a816ef61e2565a185ce430/corehq/apps/ota/urls.py#L18-L19

gherceg: Oh yeah duh. Do we actually need both of those admin restore endpoints? I suppose the admin_restore endpoint has the domain readily available in a way that the hq/admin/phone/restore might not? I need to understand how a domain is chosen when not restoring a mobile worker since the domain isn't included in the username.

dannyroberts: I think it's also available to domain admins, whereas /hq/admin/... is only available to site admins (superusers)

src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -75,6 +75,14 @@ nginx_sites:
       proxy_buffers: 8 64k
       proxy_buffer_size: 64k
       client_body_buffer_size: 512k
+    - name: '~ /hq/admin/phone/restore/'
```
dannyroberts, line 78: Nit: I see value in both the way it was before and the way it is now. This makes the paths easier to parse visually, but then makes it look like the related configuration variables (`balancer`, `proxy_redirect`, etc.) may be different when in fact they are (and are intended to stay) the same. I think I would probably go with just having a single

```
    - name: '~ (/a/[^/]+/(receiver|phone)/|/hq/admin/phone/restore/)'
```
above. That way the two different patterns are clearly differentiated (both for anyone familiar with regex willing to look at it for a couple seconds, as well as for some likely text searches).

As for the `/a/{domain}/phone/admin_restore/` endpoint (is that the right URL?), I think your reasoning (which seems right to me) is that the existing pattern already covers that. Just wanted to call that out explicitly in case you disagree.

gherceg: > As for the /a/{domain}/phone/admin_restore/ endpoint (is that the right URL?), I think your reasoning (which seems right to me) is that the existing pattern already covers that. Just wanted to call that out explicitly in case you disagree.

Yeah that is the reasoning. Thanks for calling that out explicitly.

And yeah confirmed that the `admin_restore` URL is `/a/{domain}/phone/admin_restore/`

dannyroberts (approved):

millerdev (approved):
@gherceg was it an oversight that this is not merged yet?

gherceg:
Yep thanks for raising. I'll roll this out to staging now to test that these endpoints remain functional.

PR #6327, opened 2024-06-26
Launch celery nodes in India environment

pvisweswar:
**Ticket :** https://dimagi.atlassian.net/browse/SAAS-15391

**Environments Affected :** India

PR #6330, opened 2024-06-26
Revert attempt to replace india celery machines (#6327 and #6329)

dannyroberts:
https://dimagi.atlassian.net/browse/SAAS-15391

This had to be reverted due to an issue running `deploy` with these machines included (for the first time), in the `TASK [deploy_hq : Setup virtualenv]` step. Something related to a particular version of a dependency being incompatible—it seemed fundamentally unrelated to the maintenance we were trying to do, but also not easily solvable in the moment. More info forthcoming in the ticket.

##### Environments Affected
india


millerdev (approved):
It's not clear to me what's happening with `celerybeat_a1`. Is that a new machine in addition to `celery12` and `celery13`?

Answering my own question, I think it is, but was not supposed to be part of https://github.com/dimagi/commcare-cloud/pull/6327 as noted in https://github.com/dimagi/commcare-cloud/pull/6327#discussion_r1654791631. Looks like it was not entirely removed from #6327 as intended?

PR #6335, opened 2024-07-03
Add New Celery Queue for Custom Script (Temporary)

zandre-eng:
We are planning on running [this](https://github.com/dimagi/commcare-hq/pull/34588) custom script for a domain to move cases and mobile workers to different locations. A new queue will be used so as to not clog up existing queues due to the high number of cases that we expect to move (~1.6m). This queue will only be added temporarily and will be removed again once the script has finished executing successfully for all cases.

For additional context, please refer to [this](https://dimagi.atlassian.net/browse/SAAS-15527?focusedCommentId=335643) support ticket.

##### Environments Affected

This new queue will only be available to the production environment, where we plan to run the script.

### DIFF
--- a/commcare-cloud-bootstrap/environment/app-processes.yml.j2
+++ b/commcare-cloud-bootstrap/environment/app-processes.yml.j2
@@ -5,7 +5,7 @@ celery_processes:
     flower: {}
     background_queue,case_rule_queue,celery,analytics_queue,case_import_queue:
       concurrency: 1
-    email_queue,reminder_case_update_queue,export_download_queue,linked_domain_queue,malt_generation_queue:
+    email_queue,reminder_case_update_queue,export_download_queue,linked_domain_queue,malt_generation_queue, soltech_custom_script_queue:
       concurrency: 1
     reminder_rule_queue,repeat_record_queue,saved_exports_queue:
       concurrency: 1
--- a/commcare-cloud-bootstrap/environment/app-processes.yml.j2
+++ b/commcare-cloud-bootstrap/environment/app-processes.yml.j2
@@ -5,7 +5,7 @@ celery_processes:
     flower: {}
     background_queue,case_rule_queue,celery,analytics_queue,case_import_queue:
       concurrency: 1
-    email_queue,reminder_case_update_queue,export_download_queue,linked_domain_queue,malt_generation_queue, soltech_custom_script_queue:
+    email_queue,reminder_case_update_queue,export_download_queue,linked_domain_queue,malt_generation_queue:
       concurrency: 1
     reminder_rule_queue,repeat_record_queue,saved_exports_queue:
       concurrency: 1
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -72,6 +72,8 @@ celery_processes:
       prefetch_multiplier: 1
     ush_background_tasks:
       concurrency: 8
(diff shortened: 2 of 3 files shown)


commcare-cloud-bootstrap/environment/app-processes.yml.j2
```diff
@@ -5,7 +5,7 @@ celery_processes:
     flower: {}
     background_queue,case_rule_queue,celery,analytics_queue,case_import_queue:
       concurrency: 1
-    email_queue,reminder_case_update_queue,export_download_queue,linked_domain_queue,malt_generation_queue:
+    email_queue,reminder_case_update_queue,export_download_queue,linked_domain_queue,malt_generation_queue, soltech_custom_script_queue:
```
mkangia, line 8: is this for the monolith setup? if yes, I don't think we need this change here?

environments/production/app-processes.yml
```diff
@@ -72,6 +72,8 @@ celery_processes:
       prefetch_multiplier: 1
     ush_background_tasks:
       concurrency: 8
+    soltech_custom_script_queue:  # TODO: Remove after executing script
+    concurrency: 4
```
mkangia, line 76: indentation is off here?

mkangia: so this will add the queue to 4 machines. That is what we want?

commcare-cloud-bootstrap/environment/app-processes.yml.j2
zandre-eng: This seems to be redudant code used to set up a new environment. Agreed it's then not necessary here, will remove.

zandre-eng: [77c363f](https://github.com/dimagi/commcare-cloud/pull/6335/commits/77c363f8ff6952386812d6eb3fd87a87f1ad7493)

environments/production/app-processes.yml
zandre-eng: Ah, my silly IDE auto-formatted that incorrectly. Fixed in [f33980f](https://github.com/dimagi/commcare-cloud/pull/6335/commits/f33980f665ccf85669371c84ccd2a4a775a4a90e)

environments/production/app-processes.yml
```diff
@@ -72,6 +72,8 @@ celery_processes:
       prefetch_multiplier: 1
     ush_background_tasks:
       concurrency: 8
+    soltech_custom_script_queue:  # TODO: Remove after executing script
+      concurrency: 4
```
zandre-eng, line 76: An open question I have is whether a concurrency of 4 is enough? This totals to 16 processes (4 * 4 machines) which should definitely speed up the runtime, but I'm wondering if this should be taken up any higher?

zandre-eng:
Given that we will be running the script directly (not making use of Celery), I will be closing off this PR.

environments/production/app-processes.yml
zandre-eng: I agree, the estimate was more so just a very basic linear estimation. I believe things like rate limiting and pillow lag would greatly affect the actual speed at which the Celery processes can get through the cases.

Just an FYI that I've closed off this PR since we'll be running the script without Celery.

mkangia: okay,

I believe rate limiting won't apply here since we are submitting system forms directly.

PR #6336, opened 2024-07-03
change instance type from t3a.medium to t3a.xlarge

AmitPhulera:
We started seeing staging deploy failure due to CPU constraints, the machine would become unresponsive during the deploy. So we are scaling the machines which get CPU starved on deploy.

[Context](https://dimagi.slack.com/archives/C0WLJ3HRP/p1719494057721179)

This is already applied on staging.

##### Environments Affected

Staging

### DIFF
--- a/environments/staging/terraform.yml
+++ b/environments/staging/terraform.yml
@@ -314,7 +314,7 @@ servers:
 
 proxy_servers:
   - server_name: "proxy6-staging"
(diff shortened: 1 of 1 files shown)


AmitPhulera:
I totally forgot about this PR. 
@dannyroberts or @gherceg can you have a look when you have time. This is already applied on staging.

gherceg (approved):

PR #6344, opened 2024-07-15
Increase RAM to double the number of production case-pillow processors

dannyroberts:
https://dimagi.atlassian.net/browse/SAAS-15747

##### Environments Affected
production


jingcheng16 (approved):
Looks fine......

PR #6351, opened 2024-07-31
add delcroip's ssh key to swiss

AmitPhulera:
Add `delcroip`'s ssh keys to swiss 

https://dimagi.atlassian.net/browse/SAAS-15789
##### Environments Affected

Swiss


environments/_users/swiss.yml
```diff
@@ -3,6 +3,7 @@ dev_users:
     - ddobre
     - pdelcroix
```
dannyroberts, line 4: This looks like the same person, FYI. Maybe just change their keys?

AmitPhulera: Yeah. That looks right. Thanks for pointing it out. Would update the PR.

AmitPhulera: The keys for the user already exists. Closing this PR.

PR #6352, opened 2024-08-10
staging repeat_record_queue prefetch_multiplier: 1

kaapstorm:
Currently Staging is able to create many (many) prefetched tasks on the repeat record queue. This change sets the behavior to the same as Production.

##### Environments Affected

Staging


kaapstorm:
(Oh, and it also sets the concurrency from 4 to 8, which matches the number of tasks spawned by the new `process_repeater()` task.)

millerdev (approved):

kaapstorm:
(Rebased)

millerdev (approved):

PR #6353, opened 2024-08-13
add epic secret variables

Robert-Costello:
This is part of an Epic integration related to the SUDCare project.
[OCS - CommCare - SUCare doc](https://docs.google.com/document/d/1BElCNLTXWNUzubS1oVmew8PLUOGvk-pNrMnjwKYi-5w/edit#heading=h.t3i54g959ol5)
##### Environments Affected

This will affect staging initially for testing using the Epic sandbox. 
For the actual study the variables will be updated on production in order to access a pre-existing Epic application.


snopoke (commented):
Does EPIC use oath? Just wondering if we can't use the existing oath application models in HQ (https://django-oauth-toolkit.readthedocs.io/en/latest/models.html#oauth2_provider.models.Application)

snopoke:
> Does EPIC use oath? Just wondering if we can't use the existing oath application models in HQ (https://django-oauth-toolkit.readthedocs.io/en/latest/models.html#oauth2_provider.models.Application)

Looking at your other code I see that the private key is an RSA key, not just an API key so this is probably the better approach. Are these settings 'per environment' or would we need different ones if we integrated with more projects?

Robert-Costello:
> Are these settings 'per environment' or would we need different ones if we integrated with more projects?

They are specific to each Epic App so we'd need new variables for any additional projects.

snopoke:
> > Are these settings 'per environment' or would we need different ones if we integrated with more projects?
> 
> They are specific to each Epic App so we'd need new variables for any additional projects.

OK - It doesn't seem ideal to have these here in that case but probably fine for now until we have more projects.

Robert-Costello:
@snopoke are you able to give this an approval. I'm looking to get things on staging to test on Monday.

snopoke (approved):

PR #6360, opened 2024-08-15
Remove remaining motech references

gherceg:
The first commit just removes motech and motech2 from the `special_sites` var in prod and backup prod, which is used when deciding which sites to install in the nginx config (no longer relevant for these urls).

The second commit, which I'm less sure of, just removes the motech environment. It sounds like we no longer support that env based on the url it points to, but best to get confirmation from @dannyroberts or @AmitPhulera.

##### Environments Affected

Production, backup production, motech

### DIFF
--- a/environments/backup-production/proxy.yml
+++ b/environments/backup-production/proxy.yml
@@ -21,5 +21,3 @@ special_sites:
   - tableau
   - wiki
   - wiki_http
-  - motech
-  - motech2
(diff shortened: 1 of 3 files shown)


millerdev:
Going to allow Danny or Amit sign off on this one.

dannyroberts (approved):

AmitPhulera (approved):

PR #6363, opened 2024-08-15
Remove ICDS Tableau server references and unused vars

nospame:
Removes references related to `tableau2.internal.commcarehq.org` and `icds.commcarehq.org` as the servers it pointed to were no longer active and DNS records have been removed as part of https://dimagi.atlassian.net/browse/SAAS-15848. Also removes `TABLEAU_HOST` and related variables that now appear unused.

##### Environments Affected
India


millerdev (commented):
Should this do something similar to https://github.com/dimagi/commcare-cloud/pull/6359 since it is also removing files in `

PR #6364, opened 2024-08-15
Remove J2ME site references and variables

nospame:
Removes configs and variables for J2ME sites since they are no longer supported, and any servers pointed to were determined to be inactive and their DNS records removed as part of https://dimagi.atlassian.net/browse/SAAS-15848

##### Environments Affected
All of them potentially, I think -- but in practice I don't think this could have been functional anywhere.


millerdev (commented):
Should this do something similar to https://github.com/dimagi/commcare-cloud/pull/6359 since it is also removing a file in `

PR #6366, opened 2024-08-15
Remove PNA environment and variable references

nospame:
Removes the entire `pna` environment and references to it, as the servers it pointed to were no longer active and DNS records have been removed as part of https://dimagi.atlassian.net/browse/SAAS-15848

I am less confident about the changes to ansible tasks here, since there was a small chunk of code referencing `PNA_SITE_HOST`, unlike the removal of other environments that appears a bit more straightforward.

##### Environments Affected
Removes `pna` environment, var removal likely affects production.


src/commcare_cloud/ansible/deploy_common.yml
```diff
@@ -17,14 +17,14 @@
   tags: common
 
 - name: ufw (firewall)
-  hosts: 'all_commcarehq:!proxy:!cas_proxy:!pna_proxy:!couchdb2_proxy:!reach_proxy'
+  hosts: 'all_commcarehq:!proxy:!cas_proxy:!couchdb2_proxy:!reach_proxy'
```
millerdev, line 20: I assume `cas_proxy` is also obsolete. Same below.

millerdev (commented):
Should this do something similar to https://github.com/dimagi/commcare-cloud/pull/6359 since it is also removing a file in `

src/commcare_cloud/ansible/deploy_common.yml
nospame: Yes, good call. I missed that on #6358

PR #6370, opened 2024-08-20
Update Webpack Build Script on Deploy

biyeun:
This updates the webpack build process, where applicable. This is in preparation of the new `webpack` build process coming to CommCare HQ. 

The webpack build should run after collect staticfiles. Additionally, the webpack build itself has been updated to only build
when the `webpack` directory is present. As before, this command will not run unless `use_webpack` is set to `true`.

##### Environments Affected
`staging` — no other environments are affected due to `use_webpack` environment check, which is only set to `true` for staging.


src/commcare_cloud/ansible/roles/deploy_hq/tasks/run_webpack.yml
```diff
@@ -1,10 +1,20 @@
...
     chdir: '{{ code_source }}'
-  when: config_file.stat.exists
+  when: webpack_config_directory.stat.exists
+
+- name: Build Webpack Bundles
+  # Builds all the webpack bundles
+  command:
+    cmd: yarn build
+    chdir: '{{ code_source }}'
+  when: webpack_config_directory.stat.exists
```
mjriley, line 20: The duplication of the `webpack_config_directory.stat.exists` conditional doesn't seem ideal. Is this something we can get around with [blocks](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_blocks.html)? Another link would be https://stackoverflow.com/a/34541646

src/commcare_cloud/ansible/roles/deploy_hq/tasks/run_webpack.yml
```diff
@@ -1,10 +1,20 @@
...
+    path: "{{ code_source }}/webpack"
+  register: webpack_config_directory
 
-- name: Run webpack
+- name: Generate Webpack Settings
+  # generates webpack/settings.json so that webpack has the same path
+  # to STATIC_ROOT as django
   command:
-    cmd: yarn run build:webpack
+    cmd: ./manage.py generate_webpack_settings
```
mjriley, line 11: So, if we were to want to do this process manually, we'd need to call `./manage.py generate_webpack_settings` followed by `yarn build`? I realize you've created `generate_webpack_settings` to split off this work, but is this not something that can be done as part of `yarn build`?

mjriley (commented):
I don't think my comments are blocking, but did have some questions

src/commcare_cloud/ansible/roles/deploy_hq/tasks/run_webpack.yml
biyeun: i wanted to separate a management command from django with the `yarn build` process.

millerdev (approved):
Seems like the `run_webpack.yml` could be merged with `staticfiles_collect.yml` since the tasks are doing very similar things and apply to the same hosts.

src/commcare_cloud/ansible/deploy_hq.yml
```diff
@@ -44,23 +44,22 @@
...
-    - include_role: {name: deploy_hq}
+    - import_role: {name: deploy_hq}
       vars:
-        if_not_done: run_webpack
-      when: use_webpack | default(false)
-  
+        if_not_done: staticfiles_collect
 
-- name: Collect static files
+- name: Webpack Build
```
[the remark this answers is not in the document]
biyeun: it has to come after collect static files because the `staticfiles` directory is removed.
`build_requirejs` happens after `collectstatic` fyi. when tested on `staging`, all of the files made it to the CDN just fine, if that was the concern?

biyeun:
@millerdev thanks daniel! I originally had them together (if you see the first commit), but then I noticed the separate webpack placeholder already there. I'm happy to merge the two again. There will be an update to this process once I have the HQ PR ready...this is just to unblock testing on staging

src/commcare_cloud/ansible/deploy_hq.yml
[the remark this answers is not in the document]
mjriley: Gotcha, thanks

PR #6372, opened 2024-08-21
Bump repeat record processes temporarily

mkangia:
In order to test repeaters integration with commcare analytics, bumping staging repeaters to match closer to that on production(288) for testing purposes temporarily.

##### Environments Affected

Staging


gherceg (approved):

mkangia:
This was deployed for each change on staging and then in the end staging was reverted to its original state by running `update-supervisor-conf` from master.

PR #6376, opened 2024-08-22
Add repeat_record_datasource_queue to staging

gherceg:
https://dimagi.atlassian.net/browse/SAAS-15906

Goes with this HQ PR: https://github.com/dimagi/commcare-hq/pull/35022

Adding a new queue to process a specific type of repeater. This is only meant as a temporary solution to give solutions time to improve CCA, and give SaaS time to improve repeat record infrastructure.

##### Environments Affected

Staging


environments/staging/app-processes.yml
```diff
@@ -43,7 +43,7 @@ celery_processes:
     reminder_case_update_queue,reminder_queue:
       pooling: gevent
       concurrency: 2 
-    repeat_record_queue:
+    repeat_record_queue, repeat_record_datasource_queue:
```
millerdev, line 46: nit: The repeater is named `DataSourceRepeater`. What do you think of renaming this to `datasource_repeat_record_queue`?

Same for all other similar names in this PR.

millerdev (approved):

environments/staging/app-processes.yml
gherceg: 😅 I had it that way but decided to change it so that alphabetically, these two queues would be next to each other.

millerdev: Fair enough. It was only a nitpick.

millerdev (approved):

environments/staging/public.yml
```diff
@@ -127,6 +127,7 @@ localsettings:
   BANK_NAME: "RBS Citizens N.A."
   BANK_SWIFT_CODE: 'CTZIUS33'
   CELERY_PERIODIC_QUEUE: 'celery_null'
+  CELERY_REPEAT_RECORD_DATASOURCE_QUEUE: "repeat_record_datasource_queue"
```
mkangia, line 130: we are not doing this for production? I was expecting it to be available there so we can have this on the live environment.

gherceg: I know we offlined but just so others are aware, I'm testing it on staging first and plan to put up a separate PR enabling this on production.

gherceg:
Tested on staging by running the following:

```
# create new workers for new queue
cchq --control staging update-supervisor-confs --limit celery --branch gh/repeaters/datasource-queue 

# update django settings
cchq --control staging update-config --branch gh/repeaters/datasource-queue

# restart commcare services to ensure updated settings are used when sending tasks to rabbit (could be accomplished via deploy)
cchq --control staging service commcare restart
```

Things look good on staging.

PR #6377, opened 2024-08-23
Use dedicated datasource queue on production

gherceg:
https://dimagi.atlassian.net/browse/SAAS-15906

Followup to https://github.com/dimagi/commcare-cloud/pull/6376

This creates a new celery worker for the repeat_record_datasource_queue, which will handle forwarding records specifically for datasource repeaters.
##### Environments Affected

Production

PR #6378, opened 2024-08-27
Bump up data source queue concurrency

mkangia:
Following up on the work done in https://github.com/dimagi/commcare-cloud/pull/6377, to now bump up the queue's concurrency step by step to asses its impact and adapt for load on CommCare Analytics, which is the receiving end for the requests sent via this queue.
There are 4 celery servers, so with concurrency of 1, there were 4 requests being sent in parallel.
Bumping that to 4 for each server, so at max 16 requests will be sent at once.

##### Environments Affected

Production

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -58,7 +58,7 @@ celery_processes:
       prefetch_multiplier: 1
     repeat_record_datasource_queue:
       pooling: gevent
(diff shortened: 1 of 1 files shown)


gherceg (approved):

mkangia:
This was applied to production successfully.

PR #6381, opened 2024-09-02
Add email as environment variable

Charl1996:
[Ticket](https://dimagi.atlassian.net/browse/SC-3844)

This PR simply adds an environment variable containing an email address that is used in HQ. This is needed so that local hosting projects' statistics don't end up spamming the inbox of this email address. See related [HQ PR here](https://github.com/dimagi/commcare-hq/pull/35073).

##### Environments Affected
Staging/Production

No announcements needed.

### DIFF
--- a/environments/backup-production/public.yml
+++ b/environments/backup-production/public.yml
@@ -23,6 +23,7 @@ eula_change_email: eula-notifications@dimagi.com
 contact_email: info@dimagi.com
 soft_assert_email: commcarehq-ops+soft_asserts@dimagi.com
 new_domain_email: inquiries@dimagi.com
+solutions_aes_email: solutions-tech-app-engineers@dimagi.com
 
 DATADOG_ENABLED: False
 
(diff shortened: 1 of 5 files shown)


gherceg (approved):

zandre-eng (approved):

gherceg:
@Charl1996 did you run `update-config` to apply this change? I should have mentioned that in my review, but that is necessary to get local settings updated on the targeted environment. A deploy won't actually update local settings.

Charl1996:
@gherceg 
> did you run update-config to apply this change? 

Nope, but thanks for highlighting this! 

> it has since been updated, but now a restart would be necessary for the settings to take effect, which a deploy will do

Ok, then I'll just leave it. The change is only needed from next Monday, so the deploy this week should handle this. Thus nothing for me to do now.

Thanks for bringing this up, though. Now I know.

mkangia:
Thanks for the follow up @gherceg

PR #6383, opened 2024-09-06
Lower max connections in postgres from 300 to 125 on India

gherceg:
https://dimagi.atlassian.net/browse/SAAS-15954
https://dimagi.atlassian.net/browse/SAAS-16026

We saw a spike in db connections on India that almost resulted in the instance running out of memory. We typically sit somewhere around ~50 connections, and the spike was up to 156, so we certainly couldn't handle 300 connections with the current machine size. I think we should decrease the number of max connections in this env to accommodate the RDS instance size.

I think because of our usage of pgbouncer in front of the postgres instance, we are already setup in a way that this change won't have an impact on users. If we do reach 125 connections, a client might need to wait a little longer for a request to complete since pgbouncer will wait for a connection to be freed up, but given we aren't hitting 125 connections regularly I don't anticipate this being an issue. We can monitor [this graph](https://app.datadoghq.com/dashboard/68v-xyf-tni/pgbouncer-overview?fromUser=false&fullscreen_end_ts=1725635098000&fullscreen_paused=true&fullscreen_refresh_mode=paused&fullscreen_section=overview&fullscreen_start_ts=1725592172056&fullscreen_widget=199883477&refresh_mode=paused&tpl_var_environment%5B0%5D=india&from_ts=1725592172056&to_ts=1725635098000&live=false) to see if there is an increase in wait times.

##### Environments Affected

India


environments/india/postgresql.yml
```diff
@@ -4,15 +4,15 @@ REPORTING_DATABASES:
   aaa-data: aaa-data
 
 pgbouncer_override:
-  pgbouncer_default_pool: 290
+  pgbouncer_default_pool: 115
```
gherceg, line 7: This leaves room for the 5 reserved connections and 5 additional connections if we ever need direct connections to a database for debugging/firefighting.

gherceg:
Will need to run `cchq --control india ap deploy_postgres.yml --tags=pgbouncer --limit=pgbouncer` to apply this change.

millerdev (approved):

millerdev:
Do we have any way of measuring if the memory size per connection reasonably consistent, or is there wide variation? If they are consistent then this seems reasonable, but if they are not then maybe there are circumstances where the previous limit of 300 could be useful and reasonable? Or possibly the inverse where maybe a lower number of connections could cause an OOM?

Seems like the ideal would be to have a setting to limit based on free memory (reserving enough to establish a few emergency connections) and otherwise allow unlimited connections as memory permits. But I'm guessing that's not possible.

gherceg:
Yeah that is a good point, and good reminder that this change won't necessarily solve any potential problems we may encounter in the future. I can't say how much memory on average was held by each connection when we saw the spike to 156, but given the limited memory on this machine, it seems prudent to lower it at the very least. But this is a lot of hand waving. Based on [historical trends](https://app.datadoghq.com/dashboard/6kr-yqd-dsy/rds-dashboard?fromUser=true&fullscreen_end_ts=1725917514000&fullscreen_paused=false&fullscreen_refresh_mode=sliding&fullscreen_section=overview&fullscreen_start_ts=1717968714000&fullscreen_widget=405554796&refresh_mode=sliding&tpl_var_environment%5B0%5D=india&from_ts=1725312700000&to_ts=1725917500000&live=true), I don't think there is any need for a limit higher than 125 connections, so I'm content with this change, but am acknowledging that it does not guarantee we will avoid OOMs on this RDS instance in the future.

millerdev (approved):

PR #6380, opened 2024-08-28
Bump up data source queue concurrency to 6

mkangia:
Just a follow up to https://github.com/dimagi/commcare-cloud/pull/6378 which went well.

##### Environments Affected

Production

PR #6385, opened 2024-09-11
Bump up data source queue concurrency to 9

mkangia:
A follow up to https://github.com/dimagi/commcare-cloud/pull/6380 to push more concurrency after the deploy of https://github.com/dimagi/commcare-hq/pull/35080 which should reduce the load on CCA to process each request.

##### Environments Affected

Production


gherceg (approved):
Reviewed memory and cpu usage on celery machines on prod and we have more than enough headroom.

mkangia:
Applied on production with
`cchq production update-supervisor-confs --limit=celery`

PR #6388, opened 2024-09-12
Add UAT domain to dedicated index

esoergel:
Add third BHA domain to the dedicated case search index, in the same vein as https://github.com/dimagi/commcare-cloud/pull/6308

https://dimagi.atlassian.net/browse/USH-4878

##### Environments Affected

Production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -198,6 +198,10 @@ localsettings:
     'co-carecoordination-perf': {
         'index_cname': 'case_search_bha',
         'multiplex_writes': True,
+    },
+    'co-carecoordination-uat': {
+        'index_cname': 'case_search_bha',
+        'multiplex_writes': True,
     }
   }
   IS_DIMAGI_ENVIRONMENT: True


gherceg (commented):
Do you have an idea of how much additional data is going to be included in the case_search_bha index?

MartinRiese:
@gherceg if I ran a count query on the current index would that give you the answer you are looking for?

MartinRiese:
@gherceg if I run the case search with out any other queries [here](https://www.commcarehq.org/a/co-carecoordination-uat/case/search/) I get 123435 hits. My understanding is that all those would end up in the bha index.

gherceg:
Yeah that understanding is correct. For good measure I just did a `CaseSearchES().domain('co-carecoordination-uat').count()` and got the same result. This would increase the BHA index by ~25%. I checked current shard sizes for this index which are all below 1GB (there are 4 shards), so current storage on ES nodes is more than enough to accommodate this increase.

gherceg (approved):

Issue #6401, opened 2024-10-10
Fresh install fails on task "Get current formplayer release"


zwets:
Merged pull request #6395 from yesterday breaks fresh installs.

See this comment: [6fe6169](https://github.com/dimagi/commcare-cloud/commit/6fe616921bc4a3f31861a1cc6fba768dc05ae11f#commitcomment-147827812)

Best wishes,
Marco

gherceg:
Hi @zwets. Thanks for creating the issue.

Requoting your initial comment.

> This commit breaks a fresh install. Line 170 in main.yml has
>
>readlink {{ formplayer_current_dir }})-{{ env_monitoring_id }}
>
>On a fresh install, {{ formplayer_current_dir }} is empty (there isn't a current symlink yet), so this evaluates to
>
>readlink -something
>
>So readlink will parse the -something as a commandline option and fail.

Based on the [main vars file](https://github.com/dimagi/commcare-cloud/blob/bc5cf52613bf0c401c869c00f5bfc084273188a0/src/commcare_cloud/ansible/roles/formplayer/vars/main.yml#L11-L11), I'd expect `formplayer_current_dir` to never be empty, which would mean you shouldn't run into an issue with the `readlink` command failing due to strange formatting (i.e. `readlink -something`). However given that the symlink has not yet been created, `readlink /path/to/current/release` would return an empty string, causing `basename` to fail. Am I missing something there? 

I've put up [a PR](https://github.com/dimagi/commcare-cloud/pull/6402) to resolve my understanding of the issue, but please let me know if you still have concerns.

gherceg:
@zwets [the PR](https://github.com/dimagi/commcare-cloud/pull/6402) has been merged into the master branch. Thanks again for creating this issue.

zwets:
Thanks Graham, happy to contribute!

PR #6402, opened 2024-10-11
Only set current release if symlink is set

gherceg:
https://github.com/dimagi/commcare-cloud/issues/6401

Followup to https://github.com/dimagi/commcare-cloud/pull/6395

The previous solution didn't take into consideration a fresh install where the symlink for the current release has not been setup yet. This registers the output of `readlink {{ formplayer_current_dir }}` to a variable and sets the current release name based on if that variable is an empty string or not, in addition to the `_should_update_formplayer_in_place` flag.

##### Environments Affected

None

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
+++ b/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
@@ -16,7 +16,7 @@ sentry.dsn={{ formplayer_sentry_dsn }}
 {% endif %}
 sentry.environment={{ env_monitoring_id }}
 sentry.tags.HQHost={{ host }}
-sentry.release={{ formplayer_current_release_name if _should_update_formplayer_in_place else formplayer_release_name }}
+sentry.release={{ current_formplayer_release }}
 
 spring.datasource.driver-class-name=org.postgresql.Driver
 spring.datasource.url=jdbc:postgresql://{{ postgresql_dbs.formplayer.pgbouncer_endpoint }}:{{ postgresql_dbs.formplayer.port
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
```diff
@@ -16,7 +16,7 @@ sentry.dsn={{ formplayer_sentry_dsn }}
 {% endif %}
 sentry.environment={{ env_monitoring_id }}
 sentry.tags.HQHost={{ host }}
-sentry.release={{ formplayer_current_release_name if _should_update_formplayer_in_place else formplayer_release_name }}
+sentry.release={{ current_formplayer_release }}
```
millerdev, line 19: Prior to https://github.com/dimagi/commcare-cloud/pull/6395 this was `formplayer_release_name`. I assume it is true, but will the new `current_formplayer_release` variable always be set when this template is rendered? In other words, is this template used in any other context, particularly ones where  `current_formplayer_release` is not set?

millerdev (approved):

src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
gherceg: Good point. We only reference the application.properties.j2 template in the task that also sets the `current_formplayer_release` var. I don't know if it is better to default to `formplayer_release_name` in the template if `current_formplayer_release` isn't defined, or just leave this as is so that any future attempt at rendering this file without the var defined will fail hard. I am leaning towards the latter. What do you think?

millerdev: I agree, the latter seems better. I was concerned about potential other usages, but it sounds like there are none so that's fine.

PR #6406, opened 2024-10-14
Remove duplicate 'when' in Create shared mount dir

zwets:
- Remove duplicate when that overrides the proper condition (shared_drive_enabled)
 - Remove spurious check for /opt/data being mounted, is unrelated to the task
 - FTR, this undoes changes introduced in commit 6fbac06754


millerdev (approved):

millerdev:
@zwets Thanks for the fix/cleanup.

PR #6409, opened 2024-10-14
Fix formplayer cron prune job

zwets:
This fixes the formplayer cron job "delete empty sqlite db dirs"

The previous job produced this diagnostic (emailed daily to `cchq`): `find: The -delete action automatically turns on -depth, but -prune does nothing when -depth is in effect.  If you want to carry on anyway, just explicitly use the -depth option.`

### DIFF
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -40,8 +40,8 @@
     # delete any empty directories so they don't just build up (to literally 100,000+)
     # but leaving any that have been modified in the last 20m
     # (to avoid a race condition on newly created dirs)
-    # Also never delete {{ formplayer_access_log_dir }} or {{ formplayer_data_dir }}
-    job: 'find {{ formplayer_data_dir }} -mindepth 1 -not -path {{ formplayer_access_log_dir }} -empty -type d -mmin +20 -delete'
+    # and skip {{ formplayer_access_log_dir }} and its contents
+    job: "find {{ formplayer_data_dir }} -mindepth 1 -not -path {{ formplayer_access_log_dir }} -not -path '{{ formplayer_access_log_dir }}/*' -empty -type d -mmin +20 -delete"
     user: '{{ cchq_user }}'
     state: present
   tags:
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -41,7 +41,7 @@
     # but leaving any that have been modified in the last 20m
     # (to avoid a race condition on newly created dirs)
     # Also never delete {{ formplayer_access_log_dir }}
-    job: 'find {{ formplayer_data_dir }} -path {{ formplayer_access_log_dir }} -prune -o -empty -type d -mmin +20 -delete'
+    job: 'find {{ formplayer_data_dir }} -not -path {{ formplayer_access_log_dir }} -empty -type d -mmin +20 -delete'
     user: '{{ cchq_user }}'
     state: present
   tags:
--- a/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
@@ -40,8 +40,8 @@
     # delete any empty directories so they don't just build up (to literally 100,000+)
     # but leaving any that have been modified in the last 20m
     # (to avoid a race condition on newly created dirs)
-    # Also never delete {{ formplayer_access_log_dir }}
-    job: 'find {{ formplayer_data_dir }} -not -path {{ formplayer_access_log_dir }} -empty -type d -mmin +20 -delete'
+    # Also never delete {{ formplayer_access_log_dir }} or {{ formplayer_data_dir }}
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -41,7 +41,7 @@
     # but leaving any that have been modified in the last 20m
     # (to avoid a race condition on newly created dirs)
     # Also never delete {{ formplayer_access_log_dir }}
-    job: 'find {{ formplayer_data_dir }} -path {{ formplayer_access_log_dir }} -prune -o -empty -type d -mmin +20 -delete'
+    job: 'find {{ formplayer_data_dir }} -not -path {{ formplayer_access_log_dir }} -empty -type d -mmin +20 -delete'
```
millerdev, line 44: I think the `-path` option will not match in this new command. A wildcard may fix it though.
```suggestion
    job: "find {{ formplayer_data_dir }} -not -path '{{ formplayer_access_log_dir }}/*' -empty -type d -mmin +20 -delete"
```

zwets: No, adding the `*` will delete the access log directory if it is empty:

```bash
$ mkdir dontdelete delete
$ find . -not -path ./dontdelete -empty -type d
./delete

$ find . -not -path ./dontdelete/* -empty -type d
./delete
./dontdelete
```

The `-path` predicate tests for a literal end-to-end match, so its argument should be the exact path we do not want to delete. All else (below `{{ formplayer_data_dir }}`) that is an empty directory is deleted.

zwets: I just added a commit that tackles the not-so-corner case that `{{ formplayer_data_dir }}` itself is empty and would get deleted.

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
```diff
@@ -40,8 +40,8 @@
     # delete any empty directories so they don't just build up (to literally 100,000+)
     # but leaving any that have been modified in the last 20m
     # (to avoid a race condition on newly created dirs)
-    # Also never delete {{ formplayer_access_log_dir }}
-    job: 'find {{ formplayer_data_dir }} -path {{ formplayer_access_log_dir }} -prune -o -empty -type d -mmin +20 -delete'
+    # Also never delete {{ formplayer_access_log_dir }} or {{ formplayer_data_dir }}
+    job: 'find {{ formplayer_data_dir }} -mindepth 1 -not -path {{ formplayer_access_log_dir }} -empty -type d -mmin +20 -delete'
```
millerdev, line 44: New suggested fix:

```suggestion
    job: "find {{ formplayer_data_dir }} -mindepth 1 -not -path {{ formplayer_access_log_dir }} -not -path '{{ formplayer_access_log_dir }}/*' -empty -type d -mmin +20 -delete"
```

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
millerdev: > All else below `formplayer_data_dir` that is an empty directory will be deleted.

I don't think that was the original intent. The documentation for `-prune` says

> if  the file is a directory, do not descend into it. ... For example,  to  skip  the  directory  `src/emacs` and all files and directories under it, and print the names of the other files found, do something like this:
>
>     find . -path ./src/emacs -prune -o -print

With the previous code, `formplayer_access_log_dir` and all subdirectories would have been ignored (except they were not because the `-delete` command implies `-depth`, which renders `-prune` ineffective). Some examples to illustrate:

```sh
## setup directories for testing
$ mkdir -p temp/data/logs/elk
$ mkdir -p temp/data/deer

## -prune example : does not descend into "logs" dir
$ find temp/data -path temp/data/logs -prune -o -empty -type d -print
temp/data/deer

## -not -path example : descends into "logs" dir
$ find temp/data -mindepth 1 -not -path temp/data/logs -empty -type d -print
temp/data/deer
temp/data/logs/elk
```

However, you are correct that simply adding `/*` to the path will cause an empty logs directory to be deleted. To avoid that we need another `-not -path ...` predicate. Adding a suggestion in a separate comment...

src/commcare_cloud/ansible/roles/formplayer/tasks/main.yml
zwets: Hi Daniel, the reason I didn't bother excluding what's underneath the access log dir is that I assumed the point was to spare the access log directory itself, and there weren't going to be (empty) directories underneath it anyway. If that's not the case then, yes, that extra exclusion is needed.

zwets: I'll patch in your change in another commit.

Another thing: I was looking for the actual cron file in `/etc/cron.*` but couldn't find it anywhere. Then remembered the old days, and indeed: it is in `/var/spool/crontab/cchq`, I suppose because its ansible task lacks the `cron_file` attribute.

I considered adding a patch to add the attribute (for consistency with the other jobs - this is the single one that uses classic crontab) but I suppose we'll then end up with both jobs when deploying on an existent system?

zwets: Talking about cleanup: neither this cron job nor the one below is in `formplayer/tasks/cleanup.yml`, whereas the cron jobs above it are.

millerdev (approved):

PR #6410, opened 2024-10-14
Update WAF regex

orangejenny:
Motivated by https://github.com/dimagi/commcare-hq/pull/35216, also makes a couple of updates from recent HQ PRs that missed it.

##### Environments Affected
Planning to run on staging, prod, and india


gherceg (approved):

orangejenny:
I've now run this on staging, india, and prod, and have updated the prod maintenance feed.

PR #6416, opened 2024-10-18
Disable termination protection on backup prod env

AmitPhulera:
we destroy it with terraform destroy after bcp is done and have to manually disable termination protection from aws console. So its better not to have it while we are creating the machines

##### Environments Affected

Backup Production

### DIFF
--- a/src/commcare_cloud/terraform/modules/pgbouncer_nlb/main.tf
+++ b/src/commcare_cloud/terraform/modules/pgbouncer_nlb/main.tf
@@ -4,7 +4,7 @@ resource "aws_lb" "this" {
   load_balancer_type = "network"
   subnets = var.subnets
 
-  enable_deletion_protection = true
+  enable_deletion_protection = var.environment == "bk-production" ? false : true
 
   tags = {
     Environment = var.environment
--- a/src/commcare_cloud/terraform/modules/postgresql/main.tf
+++ b/src/commcare_cloud/terraform/modules/postgresql/main.tf
@@ -36,7 +36,7 @@ module "postgresql" {
   password = var.rds_instance["password"]
   port     = var.rds_instance["port"]
 
-  deletion_protection = true
+  deletion_protection = var.environment == "bk-production" ? false : true
 
   multi_az = var.rds_instance["multi_az"]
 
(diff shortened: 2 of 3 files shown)


gherceg (approved):
I assume there is no way to make this an option that we can define in an environment specific file that terraform can then reference? Might be slightly nicer/feel slightly safer but I'm fine with this.

AmitPhulera:
There is a straightforward way to make it configurable. The reason why I did not do it is because it won't make sense on any other environment and we would not want to set this accidentally to true on any environment. If there is a usecase in future we would be able to update it.

The steps would be - 

Add a new attribute `termination_protection = jsonobject.BooleanProperty(default=True)` in following classes
- `ServerConfig` class https://github.com/dimagi/commcare-cloud/blob/843489bc77e028694fdbfb20b37ca5ba78c37285/src/commcare_cloud/environment/schemas/terraform.py#L110
- `RDSInstanceConfig` class https://github.com/dimagi/commcare-cloud/blob/843489bc77e028694fdbfb20b37ca5ba78c37285/src/commcare_cloud/environment/schemas/terraform.py#L173
- `PgbouncerNlbs` class https://github.com/dimagi/commcare-cloud/blob/843489bc77e028694fdbfb20b37ca5ba78c37285/src/commcare_cloud/environment/schemas/terraform.py#L209

And use the value you have added in - 

- Set `disable_api_termination = var.disable_api_termination`
https://github.com/dimagi/commcare-cloud/blob/843489bc77e028694fdbfb20b37ca5ba78c37285/src/commcare_cloud/terraform/modules/server/main.tf#L12

- Set `deletion_protection = var.disable_api_termination`
 https://github.com/dimagi/commcare-cloud/blob/843489bc77e028694fdbfb20b37ca5ba78c37285/src/commcare_cloud/terraform/modules/postgresql/main.tf#L39

- Set `enable_deletion_protection = var.disable_api_termination`
https://github.com/dimagi/commcare-cloud/blob/843489bc77e028694fdbfb20b37ca5ba78c37285/src/commcare_cloud/terraform/modules/pgbouncer_nlb/main.tf#L7

After that update `terraform.yml` for your environment to set the following values to `false`. Again I did not wanted to make it configurable, atleast not now.

PR #6419, opened 2024-10-18
Pass directory permissions as string

AmitPhulera:
While running 
`cchq backup-proudction ap deploy_efs_mount_shared_dir.yml`,  I encountered following error . The reason is that jinja is interpreting the permission starting with 0 as octal number? changing it to string fixes the issue.

See - https://github.com/ansible/ansible/issues/56047#issuecomment-492293723

##### Environments Affected

All ?


gherceg (approved):
Nice find.

PR #6420, opened 2024-10-18
Use -daystart option in the `find` command when removing old backups

gherceg:
From the commit message
> The -daystart option effectively makes the -mtime option measure calendar days instead of 24 hour increments from the current timestamp. Due to unintuitive behavior of the -mtime option, this ensures that backups modified close to 3 days ago at the time the find command is run are returned since they were last modified 3 calendar days ago.

The aim of this PR is to make our daily backup logic behave more intuitively. Currently, on an environment where we keep daily backups for 2 days, at any given point, the machine will have 4 daily backups. This is because of rounding done in `find ... -mtime "+2" ...`. This rounding makes it so that a file as to have been last modified over 3 days ago in order for it to be found when providing the `-mtime "+2"` option. In other words, a file modified 2.99999 days ago would still be considered as modified exactly 2 days ago, and therefore not modified _more than_ 2 days ago.

By adding the `-daystart` option, we are telling `find` to look for files modified more than 2 days ago...starting from the beginning of today. This ensures that at the time the cron job runs the backup cleanup logic (after creating backups), it will consider anything created 3 calendar days ago as ready for cleanup, which is the behavior we want.

Given this could impact self hosters, I think it makes sense to make an announcement about this change. To be overly cautious, it is also probably best to allow time for self hosters to see this change before merging in case they have shortened their backup retention window to account for the current behavior.

### Referencing docs

Looking at docs for [find](https://man7.org/linux/man-pages/man1/find.1.html):

> -mtime n
File's data was last modified less than, more than or
exactly n*24 hours ago.  See the comments for -atime to
understand how rounding affects the interpretation of file
modification times.

Looking at `atime`, the important part to understanding rounding is:
> When find figures out how many 24-hour
periods ago the file was last accessed, any fractional
part is ignored, so to match -atime +1, a file has to have
been accessed at least two days ago.

And to finish it off:
> -daystart
Measure times (for -amin, -atime, -cmin, -ctime, -mmin,
and -mtime) from the beginning of today rather than from
24 hours ago.  This option only affects tests which appear
later on the command line.

##### Announce New Release

### DIFF
--- a/src/commcare_cloud/ansible/roles/backups/templates/create_blobdb_backup.sh.j2
+++ b/src/commcare_cloud/ansible/roles/backups/templates/create_blobdb_backup.sh.j2
@@ -27,7 +27,7 @@ else
 	tar -Pzcf "{{ blobdb_backup_dir }}/${BACKUP_FILE}" "{{ blobdb_dir_path }}"
 
 	# Remove old backups of this backup type
-	find {{ blobdb_backup_dir }} -mtime "+${DAYS_TO_RETAIN_BACKUPS}" -name "blobdb_${BACKUP_TYPE}_*" -delete;
+	find {{ blobdb_backup_dir }} -daystart -mtime "+${DAYS_TO_RETAIN_BACKUPS}" -name "blobdb_${BACKUP_TYPE}_*" -delete;
 fi
 
 {% if blobdb_s3 %}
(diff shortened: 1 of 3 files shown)


AmitPhulera (approved):
Good find 👏 ! 
Curious how did you stumble upon this?

gherceg:
Backups were a source of high disk usage on a machine which led me here. I've deployed this to the target environment and will monitor in the coming days to see if this behaves as expected.

PR #6421, opened 2024-10-21
Changes required to install Elasticsearch 6 on staging

AmitPhulera:
https://dimagi.atlassian.net/browse/SAAS-16062

Changes required to install Elasticsearch 6 on staging

##### Environments Affected

Staging


src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
```diff
@@ -7,8 +7,10 @@ After=network-online.target
 [Service]
 Type=simple
 Environment=ES_HOME={{ elasticsearch_home }}
-Environment=CONF_DIR={{ elasticsearch_conf_dir}}
 Environment=ES_PATH_CONF={{ elasticsearch_conf_dir}}
+{% if elasticsearch_version <= '5.6.16'%}
+Environment=CONF_DIR={{ elasticsearch_conf_dir}}
+{% endif %}
```
gherceg, line 13: Just trying to follow along. Why don't we need to define `CONF_DIR` for ES 6?

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -39,12 +39,12 @@ discovery.zen.fd.ping_interval: 10s
 discovery.zen.fd.ping_retries: 20
 {% endif %}
 
-{% if elasticsearch_version == '5.6.16' %}
+{% if elasticsearch_version is version('6.0.0', '>=') %}
```
gherceg, line 42: Why does this no longer apply to 5.6.16?

zwets:
If I may chime in on this change, I would suggest also fixing this:

```
diff --git a/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/jvm.options.j2 b/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/jvm.options.j2
index 36761736b..50ba49a01 100644
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/jvm.options.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/config/jvm.options.j2
@@ -54,7 +54,7 @@
 14-:-XX:InitiatingHeapOccupancyPercent=30
 
 ## JVM temporary directory
--Djava.io.tmpdir=${ES_TMPDIR}
+-Djava.io.tmpdir=tmp
 
 ## heap dumps
 
@@ -67,7 +67,7 @@
 -XX:HeapDumpPath={{ elasticsearch_data_dir }}/logs/heapdump.hprof
 
 # specify an alternative path for JVM fatal error logs
--XX:ErrorFile={{ elasticsearch_data_dir }}logs/hs_err_pid%p.log
+-XX:ErrorFile={{ elasticsearch_data_dir }}/logs/hs_err_pid%p.log
 
 ## JDK 8 GC logging
 8:-XX:+PrintGCDetails
```

Regarding the fixes: `jvm.options` is a config file for the JVM, not a shell script, so  `-Djava.io.tmpdir=${ES_TMPDIR}` results in a directory that is _literally_ called `/opt/elasticsearch-5.6.16/${ES_DIR}` :laughing: 
 
I did not submit [my local fix](https://github.com/zwets/commcare-cloud/tree/fix-es-jvm-options)) of simply commenting the line out, because on many systems the default `java.io.tmpdir` (`/tmp`, under which ES creates a unique directory `jna--{number}`) is mounted `noexec`, whereas apparently ES wants to execute scripts from it.

Clearly then this is an ES bug: rather than make deployers redirect the JVM tmpdir, they should do their JNA execution in some directory of their own, for instance simply `tmp` under their data directory. My fix above does exactly that.

Regardless of whether you stick to the directory called `$ES_TMPDIR` or call it `tmp` as I suggest, if it is not under `/tmp` you'll probably need to take care of its clean-up. Not sure if that's currently taken care of?

The second change in my diff above seemed obvious to me by just looking at it.

src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
AmitPhulera: Its not longer supported `ES_PATH_CONF` is all we need. 
https://www.elastic.co/guide/en/elasticsearch/reference/6.8/breaking-changes-6.0.html#_literal_conf_dir_literal_is_no_longer_supported

src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml
```diff
@@ -32,6 +32,15 @@
     dest: "/home/{{ cchq_user }}/downloads/elasticsearch-{{ elasticsearch_version }}.tar.gz"
   when: elasticsearch_version == "5.6.16"
 
+- name: Download Elasticsearch 6
+  become: yes
+  get_url:
+    url: https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-{{ elasticsearch_version }}.tar.gz
+    dest: "/home/{{ cchq_user }}/downloads/elasticsearch-{{ elasticsearch_version }}.tar.gz"
+    sha256sum: "{{ elasticsearch_download_sha256 }}"
+  when: elasticsearch_version is version('6.0.0', '>=') and elasticsearch_version is version('7.0.0', '<')
```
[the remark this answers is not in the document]
AmitPhulera: I also had the same thought while adding the task I looked for it and realised that after 7 the distribution of artefacts is standardised. So we can have a single task when we are fully on ES 7.

src/commcare_cloud/ansible/roles/elasticsearch/tasks/misc_v6.yml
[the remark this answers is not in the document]
AmitPhulera: Done in https://github.com/dimagi/commcare-cloud/pull/6421/commits/fda8d7fa5a8b7fc03d7bf70652f70c7655b8c1ef

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
AmitPhulera: Ahh! This was a typo. Thanks for catching it. Fixed in https://github.com/dimagi/commcare-cloud/pull/6421/commits/086eba47d387e7a14cd0b99195814e54cd8a3f95

AmitPhulera:
@zwets Thank you for your suggestions. They were definitely an oversight from our end. I have incorporated them in https://github.com/dimagi/commcare-cloud/pull/6421/commits/eb11aa55358c1040002fa38cddb3ba4de598ed53.

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
```diff
@@ -69,13 +63,6 @@ network.host: "{{ lookup('dig', inventory_hostname, wantlist=True)[0] }}"
 indices.fielddata.cache.size:  {{ elasticsearch_fielddata_cache_size }}
 {% endif %}
 
-{% if elasticsearch_enable_inline_groovy_scripts %}
-{% if elasticsearch_version is version('5.6.16', '<=') %}
```
gherceg, line 70: Just going off of this, these settings seemed to also apply to ES5 and therefore aren't strictly an ES2 setting? Are they still safe to remove?

gherceg (commented):
One small comment but otherwise looks good

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/elasticsearch.yml.j2
AmitPhulera: Yes this is safe to remove and is intentionally removed. This was there to in ES 5 to support if someone was using groovy scripts which are deprecated. 
If someone installs commcare now they will not need es5 with groovy scripts so it is safe to remove for ES5 as well.

PR #6423, opened 2024-10-22
Update pid path from tmp to services dir

AmitPhulera:
Main concern was if pid file is deleted by some automated process then it would affect the log rotation until gunicorn is restarted. So adding it to a more permanent place

##### Environments Affected

All


gherceg (approved):
Seems like a better place anyway!

PR #6428, opened 2024-10-31
Reduce Kafka logging

zwets:
PR for added parsimony:

 - 24ec2008b653 bumps log4j thresholds from `TRACE` to `INFO` for Zookeeper and Kafka. Rationale: `DEBUG` and lower shouldn't be on in production, and anyway we wouldn't be debugging that code (it's Zookeeper and Kafka's)
 - 7eb112295af7 removes `stdout` from the root logger, as this merely duplicates what is written to the log files (by the `KafkaAppender`) in systemd's journal. (Try `journalctl -u kafka-server`.)

### DIFF
--- a/src/commcare_cloud/ansible/roles/kafka/files/log4j_2.X.X.properties
+++ b/src/commcare_cloud/ansible/roles/kafka/files/log4j_2.X.X.properties
@@ -1,4 +1,4 @@
-log4j.rootLogger=INFO, stdout, kafkaAppender
+log4j.rootLogger=INFO, kafkaAppender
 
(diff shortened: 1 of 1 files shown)


zwets:
Thanks @millerdev, also for the nitpick. TBH I had forgotten that this commit was still in an open PR. I was just rebasing my old patches on latest master to start a new deployment after a 16mo hiatus.

So far, it's been smooth sailing. The one issue I ran into (redis not starting) turned out to be [a niche LXC issue](https://github.com/lxc/distrobuilder/issues/988). (My test monolith is an LXC container.)

PR #6430, opened 2024-11-04
Add Geospatial Celery Queue

zandre-eng:
Jira ticket [here](https://dimagi.atlassian.net/browse/SC-3864).

A new Celery queue for the Geospatial feature has been added. Currently, there is only one task planned to be run on this queue, which is related to [this](https://github.com/dimagi/commcare-hq/pull/35126) PR. This task was originally set to run as a serial task on the `background_queue`, however it takes very long to process a lot of cases. Given how long it takes, and that it is a serial task, any domains that are queued up for the task are highly likely to fail from getting locked out of the task for too long.

Adding a new Celery queue will resolve the issue mentioned above by allowing the Geospatial Celery task to be run in parallel without the risk of blocking up something like the `background_queue`. A concurrency of 1 has been chosen since we do not expect more than a few of the Celery tasks to be fired off at any given time. Additionally, this is a task that we expect to be run quite infrequently since it is only run when the Geospatial feature is enabled for a domain.

##### Environments Affected

This touches all environments.

### DIFF
--- a/src/commcare_cloud/environment/schemas/app_processes.py
+++ b/src/commcare_cloud/environment/schemas/app_processes.py
@@ -137,7 +137,8 @@ CELERY_PROCESSES = [
     CeleryProcess("ucr_indicator_queue", required=False, blockage_threshold=60 * 60),
     CeleryProcess("ucr_queue", required=False, blockage_threshold=60 * 60),
     CeleryProcess("user_import_queue", required=False, blockage_threshold=60 * 60),
-    CeleryProcess("ush_background_tasks", required=False, blockage_threshold=3 * 60 * 60)
+    CeleryProcess("ush_background_tasks", required=False, blockage_threshold=3 * 60 * 60),
+    CeleryProcess("geospatial_queue", required=False, blockage_threshold=3 * 60 * 60)
 ]
 
 
--- a/src/commcare_cloud/environment/schemas/app_processes.py
+++ b/src/commcare_cloud/environment/schemas/app_processes.py
@@ -138,7 +138,7 @@ CELERY_PROCESSES = [
     CeleryProcess("ucr_queue", required=False, blockage_threshold=60 * 60),
     CeleryProcess("user_import_queue", required=False, blockage_threshold=60 * 60),
     CeleryProcess("ush_background_tasks", required=False, blockage_threshold=3 * 60 * 60),
-    CeleryProcess("geospatial_queue", required=False, blockage_threshold=3 * 60 * 60)
+    CeleryProcess("geospatial_queue", required=False, blockage_threshold=6 * 60 * 60)
 ]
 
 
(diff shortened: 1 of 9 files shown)


src/commcare_cloud/environment/schemas/app_processes.py
```diff
@@ -137,7 +137,8 @@ def __new__(cls, name, required=True, is_queue=True, blockage_threshold=None,
     CeleryProcess("ucr_indicator_queue", required=False, blockage_threshold=60 * 60),
     CeleryProcess("ucr_queue", required=False, blockage_threshold=60 * 60),
     CeleryProcess("user_import_queue", required=False, blockage_threshold=60 * 60),
-    CeleryProcess("ush_background_tasks", required=False, blockage_threshold=3 * 60 * 60)
+    CeleryProcess("ush_background_tasks", required=False, blockage_threshold=3 * 60 * 60),
+    CeleryProcess("geospatial_queue", required=False, blockage_threshold=3 * 60 * 60)
```
mkangia, line 141: nit: Do you want to update this threshold considering we would have long running tasks?

mkangia (approved):
Looks good.

I appreciate the thoughtful clubbing of queue for different environments.

Good to wait for a go ahead from @gherceg

src/commcare_cloud/environment/schemas/app_processes.py
zandre-eng: I was thinking of perhaps taking this up to 6 hours, however I'm not sure what the potential consequences are for raising the threshold. Then again, this queue will only be running the Geospatial indexing task, so I can't think there would be any major negative implications in raising this limit.

Tagging @gherceg to get his thoughts on whether raising this limit to 6 hours sounds fine.

mkangia: > I'm not sure what the potential consequences are for raising the threshold

I believe this is used to check if there is any issue with the queues. Considering we expect the queue to be blocked with one task for long, 3 hours of threshold for wait period for a new task seems low.

For this queue, this could probably be the time we expect a large data load to take.

zandre-eng: On the production envrionment, a task processing the max case limit takes around 1.5 hours. Given this, I feel a time limit of 3-4 hours should then be fine. 

The time taken is higher on smaller instances however (e.g. staging takes ~3.5 hours), so if we want to consider these instances as well then something closer to 4-6 hours would make more sense.

mkangia: Alright :+1:

zandre-eng: Blockage threshold has been increased to 6 hours ([01dfb8d](https://github.com/dimagi/commcare-cloud/pull/6430/commits/01dfb8d20ea6b7d09696753afc4dd5ed1d74506b)).

gherceg (approved):
I think this is a good way to artificially limit the load these migrations can incur on the system at any given time. I do still wonder if it is worth chunking up the celery tasks that run on this queue to avoid long running tasks since that is still a general good practice. Though I recognize that is not in the scope of this PR.

zandre-eng:
> I do still wonder if it is worth chunking up the celery tasks that run on this queue to avoid long running tasks since that is still a general good practice.

I completely agree here. I left this change out for now as it will increase the complexity of the current solution which I agree is a bit out of scope for this PR. But this is definitely a great next improvement to make in followup work!

PR #6451, opened 2024-12-16
Double block storage on kafka machines

gherceg:
There was an increase in case load on staging due to some testing, and the kafka machines do not have enough disk to handle it. I think this small increase should be enough to handle the additional load (trying to be conservative since it is annoying to downsize).

##### Environments Affected

Staging

### DIFF
--- a/environments/staging/terraform.yml
+++ b/environments/staging/terraform.yml
@@ -253,7 +253,7 @@ servers:
     volume_encrypted: yes
     volume_type: "gp3"
     block_device:
-      volume_size: 5
+      volume_size: 10
       encrypted: true
       volume_type: "gp3"
     group: "kafka"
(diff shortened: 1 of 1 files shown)


AmitPhulera (approved):

environments/staging/terraform.yml
```diff
@@ -253,7 +253,7 @@ servers:
     volume_encrypted: yes
     volume_type: "gp3"
     block_device:
-      volume_size: 5
+      volume_size: 10
```
AmitPhulera, line 256: Are you aware if log rotation is enabled on these machines?

gherceg: This block storage is solely for kafka data (/opt/data/) whereas kafka logs are in /var/log/kafka, so I don't think that was the issue here. I don't know the details of kafka's log rotation, but it looks like we have a default of 28 days (kafka_log_retention) so I would guess that the answer is still that log rotation is enabled on staging.

PR #6458, opened 2025-01-03
Update commcare cloud setup instructions

gherceg:
I've run into various issues while setting up monoliths and am attempting to clarify those instructions in this PR.

This also notably updates documentation to be aligned with our removal of the suggestion to copy a private key to the server, and instead use `ssh -A` when/where appropriate.

Review by commit.
##### Environments Affected

None

### DIFF
--- a/docs/source/installation/2-manual-install.rst
+++ b/docs/source/installation/2-manual-install.rst
@@ -456,13 +456,7 @@ Install and Configure CommCare Cloud
 
     ::
 
-        $ nano ~/.profile
-
-    Paste the following before the line that sources the init-ansible script.
-
-    ::
-
-        source ~/.commcare-cloud/load_config.sh
+       $ echo "source ~/.commcare-cloud/load_config.sh" >> ~/.profile
 
     Load the commcare-cloud config so it takes effect immediately:
 
--- a/docs/source/installation/2-manual-install.rst
+++ b/docs/source/installation/2-manual-install.rst
@@ -514,19 +514,31 @@ will tell you more about how we use this vault file.
 Deploy CommCare HQ services
 ---------------------------
 
-When you run the “commcare-cloud deploy-stack”, you will be prompted for
-the vault password from earlier. You will also be prompted for an SSH
-password. This is the root user’s password. After this step, the root
-user will not be able to log in via SSH.
+The first step is to setup the expected user configuration. You will be prompted for
+the vault password from earlier and the SSH password, which is the root user's password.
+After this step, the root user will not be able to log in via SSH.
 
 ::
 
-   $ commcare-cloud cluster deploy-stack --first-time -e 'CCHQ_IS_FRESH_INSTALL=1'
+    $ commcare-cloud cluster bootstrap-users
 
-   This command will apply without running the check first. Continue? [y/N]y
-   ansible-playbook /home/jbloggs/commcare-cloud/src/commcare_cloud/ansible/deploy_stack.yml -i /home/jbloggs/environments/cluster/inventory.ini -e @/home/jbloggs/environments/cluster/vault.yml -e @/home/jbloggs/environments/cluster/public.yml -e @/home/jbloggs/environments/cluster/.generated.yml --diff --tags=bootstrap-users -u root --ask-pass --vault-password-file=/home/jbloggs/commcare-cloud/src/commcare_cloud/ansible/echo_vault_password.sh --ask-pass --ssh-common-args -o=UserKnownHostsFile=/home/jbloggs/environments/cluster/known_hosts
-   Vault Password for 'cluster': <ansible_sudo_pass>
-   SSH password: <root user's password>
+
+Once this completes successfully, you will now be able to ssh into this machine from your previously created user (e.g., jbloggs).
+You should exit your current ssh session, and ssh back into the machine using the "-A" option to enable agent forwarding.
+This is necessary to escalate privileges when running commcare-cloud commands, as well as for executing commands on other machines if
+you are setting up a cluster.
+
+::
+
+    $ exit  # exit until no longer connected to the machine
+    $ ssh -A jbloggs@control1
+
+
+You are now ready to deploy CommCare HQ services.
+
+::
+
+   $ commcare-cloud cluster deploy-stack -e 'CCHQ_IS_FRESH_INSTALL=1'
 
 This will run a series of Ansible commands that will take quite a long
 time to run.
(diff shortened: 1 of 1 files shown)


docs/source/installation/2-manual-install.rst
```diff
@@ -111,6 +111,12 @@ reference.
 
         PasswordAuthentication yes
 
+    To allow keyboard interactive authentication, ensure
```
mkangia, line 114: I guess this is needed in specific cases and not all interactive authentication. I definitely could do an install without this change.

Found this useful reference: https://serverfault.com/questions/189643/sshd-blocking-password-authentication

docs/source/installation/2-manual-install.rst
```diff
@@ -126,46 +132,14 @@ To be used in the installation process.
...
-
-    If your default version is not 3.x or if the “python” command was
-    not found, make python3 your default by running the command below,
-    otherwise skip it.
-
-    ::
-
-        $ sudo update-alternatives --install /usr/bin/python python /usr/bin/python3 10
-
-
```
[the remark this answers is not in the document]
mkangia: Oh I see, its moved.

docs/source/installation/2-manual-install.rst
```diff
@@ -211,15 +201,23 @@ Install CommCare Cloud
         Do you want to have the CommCare Cloud environment setup on login?
         (y/n): y
 
-3.  Clone the sample CommCare Cloud “environments” folder into your home
+    If the input times out before entering 'y', you can follow the prompt instructions
+    to setup CommCare Coud on login:
```
AmitPhulera, line 211: nit
```suggestion
    to setup CommCare Cloud on login:
```

docs/source/installation/2-manual-install.rst
```diff
@@ -467,13 +464,13 @@ Install CommCare Cloud
 
        $ cp ~/commcare-cloud/src/commcare_cloud/config.example.py ~/commcare-cloud/src/commcare_cloud/config.py
 
-    Update the known hosts file
+    Update the known hosts file (substituting your environment name if necessary)
 
     ::
 
        $ commcare-cloud cluster update-local-known-hosts
```
AmitPhulera, line 483: I know we have used cluster as environment name in the entire doc but would it be clearer if we use <env> which would indicate that this has to be replaced?
```suggestion
       $ commcare-cloud <env> update-local-known-hosts
```

btw I am totally fine with the way it is right now it was just a thought that I wanted to share with you.

docs/source/installation/2-manual-install.rst
```diff
@@ -450,7 +450,13 @@ Install and Configure CommCare Cloud
...
     ::
 
-       $ echo "source ~/.commcare-cloud/load_config.sh" >> ~/.profile
+        $ nano ~/.profile
+
+    Paste the following before the line that sources the init-ansible script.
+
+    ::
+
+        source ~/.commcare-cloud/load_config.sh
```
AmitPhulera, line 465: Any specific case where this approach better than the previous one?

AmitPhulera (commented):
Great improvements, clearly makes things lot clearer. Thanks for the PR. Just have a few questions/suggestions otherwise this looks gtg.

docs/source/installation/2-manual-install.rst
gherceg: Yeah I do want to make this change. I also want to make a sweeping change to the `$` syntax before each command as it makes copy/pasting annoying, so was planning to do this in a separate PR if that is ok with you.

docs/source/installation/2-manual-install.rst
gherceg: Based on the order of instructions, `load_config.sh` was being run _after_ `init-ansible`, but load_config plays a critical role in exporting the COMMCARE_CLOUD_ENVIRONMENTS variable to point to the right environments directory. I know this change complicates the setup a bit, but hopefully results in fewer issues down the road. Does that answer your question?

docs/source/installation/2-manual-install.rst
gherceg: Yeah agreed it isn't necessary all of the time. Just checking that you agree that it isn't worth specifying that in the instructions given that it will be reverted in a later setup step anyway?

docs/source/installation/2-manual-install.rst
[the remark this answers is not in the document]
gherceg: Given that we specify ubuntu 22.04 at the moment, 3.10 _will_ be the system version, so it seems fine to simplify the step to just run this rather than specify the condition.

gherceg:
Possibly a more extreme change, but I'm tempted to remove support for running `bootstrap-users` as part of the deploy-stack command entirely, since bootstrap users is a one-off command that will change the state of user permissions on the system, and inevitably result in needed to re-run deploy-stack from a different user config, at least in my experience.

millerdev (approved):

docs/source/installation/2-manual-install.rst
millerdev: Another option would be to have the `init-ansible` script `source ~/.commcare-cloud/load_config.sh` if it exists, and skip the instruction to add it to `.profile`.

gherceg: Yeah upon closer inspection, I may have incorrectly assumed this was the cause of an issue I was facing. It seems fair to say that the init-ansible script does not use the COMMCARE_CLOUD_ENVIRONMENTS variable for anything other than how it is used in `update_code.sh`. I think I know where I went wrong, but I'll try to run through the setup steps without this change just to make sure.

gherceg: Ok it seems fine. Reverted in 17240b4e210f6b9867fc35e9cd6f03f559100815.

millerdev (approved):

docs/source/installation/2-manual-install.rst
AmitPhulera: Yeah. Totally fine by me.

AmitPhulera (approved):

docs/source/installation/2-manual-install.rst
mkangia: I think if someone gets stuck at this step like you did, they are doing something special.

Seems fine to leave it if you think it could be useful for someone to make the initial root password step work, with a comment that "If needed you can enable keyboard interactive authentication with", though there are so many steps here that one would just go about copying pasting commands without really reading the instructions well, which is why it would be nice to keep commands limited and exclude anything special.
So, This could go either also go [here](https://commcare-cloud.readthedocs.io/en/latest/installation/troubleshooting.html)

Agreed that this later gets cleaned up.

PR #6460, opened 2025-01-10
Make slack deploy diff work for longer diffs as well

AmitPhulera:
We sometimes do not receive diffs during deploy so they have to be manually copied. On looking I found out it is because of the 3000 character limit on the message. The fix that is made here is chunking the longer messages and sending as separate messages on slack

[This](https://dimagi.slack.com/archives/C0B44C0Q2/p1736501423146569) is how the diffs will look like. 

There was another approach that I experimented with, which was sending different blocks in the same message like [this](https://dimagi.slack.com/archives/C0B44C0Q2/p1736502263670369)

In both the approaches the messages are getting truncated under `See More`. For the second approach the `See More` is hidden in the diff and hard to find. Although first approach is still not the best but it provides a good differentiation between the messages.

##### Environments Affected

Prod and India

### DIFF
--- a/src/commcare_cloud/commands/deploy/slack.py
+++ b/src/commcare_cloud/commands/deploy/slack.py
@@ -67,11 +67,31 @@ class SlackClient:
         response = self._post_message(message, blocks)
         context.set_meta_value('slack_thread_ts', response["ts"])
         if self.environment.fab_settings_config.generate_deploy_diffs:
-            diff_blocks = [{
-                "type": "section",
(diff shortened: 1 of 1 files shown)


snopoke (approved):
Another option would be to send the diff as a file: https://api.slack.com/messaging/files#upload

AmitPhulera:
That could have been one of the approach but I was not sure if Slack would render markdown files in the same way it renders markdown messages.

millerdev:
:star: :trophy: :star:

PR #6469, opened 2025-01-31
[Not actually enforced] Increase DEVICE_LIMIT_PER_USER to 50

gherceg:
https://dimagi.atlassian.net/browse/SAAS-16355

This rate limiter isn't actually enabled yet, so this change is only to help us get better insight in our metrics, since we have more insight into metrics before exceeding the limit.

##### Environments Affected

Production


orangejenny (approved):
Would it be useful to have this on india and/or staging?

gherceg:
Fair question. My initial reaction was no, but I suppose it makes sense to keep this limit consistent to start.

orangejenny (approved):

PR #6472, opened 2025-02-15
Soft kill celery processes via celery_bash_runner

gherceg:
https://dimagi.atlassian.net/browse/SAAS-16545

This is an improvement to an existing band-aid solution. The ticket outlines in detail the underlying issue, but basically we've seen celery processes, especially repeat_record_queue workers, fail to shutdown properly when receiving a SIGTERM. The investigation into this is ongoing, but in the meantime we should make this change since in the current state, if a celery process runs into an issue during shut down, supervisor is unaware of it and does not fire up a new worker. This way, supervisor will create a new worker. From what I've seen, the process that fails to shut down stops consuming from the queue, and there is still the hard kill that will come along to clean the rogue process up eventually.

You might ask "why do we do this in the first?". My understanding is that we periodically shut down celery processes to prevent them from outliving the release they are running on. There is some expectation that it might take awhile for a celery process to do a warm shutdown if there are any long running tasks in progress, but I'm not convinced that grace period should be 36 hours (currently our soft kill threshold is 36 hours, and our hard kill threshold is 72 hours, so 72 - 36 = 36).

Writing this out, I'm not convinced we shouldn't just remove these scripts altogether, but it feels like while this rogue celery process issue exists we are better off keeping these scripts to prevent them from building up over time.

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/celery_bash_runner_kill.sh.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/celery_bash_runner_kill.sh.j2
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# Process to soft kill old celery jobs via supervisor managed process.
+#
+
+HOURS=$1
+MAX_MINUTES=`expr ${HOURS} \* 60 \* 60`
+
+SIG_NAME=$2
+
+PARENT_PIDS=$(ps -eo comm,ppid,etimes | awk '/celeryd:/ { if ( $4 > '"${MAX_MINUTES}"' ) { print $3 } }');
+
+if [ "$PARENT_PIDS" != "" ]; then
+  echo "Shutting down celery bash runners:"
+  echo "Process information with start and elapsed time:"
+  for pid in ${PARENT_PIDS}
+  do
+     if ps -p ${pid} -o pid,cmd | grep -qE 'celery_bash_runner'; then
+        echo "`ps -p ${pid} -o pid,start,etime,cmd`"
+        kill -${SIG_NAME} ${pid}
+  done
+fi
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/celery_bash_runner_kill.sh.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/celery_bash_runner_kill.sh.j2
@@ -15,8 +15,9 @@ if [ "$PARENT_PIDS" != "" ]; then
   echo "Process information with start and elapsed time:"
   for pid in ${PARENT_PIDS}
   do
-     if ps -p ${pid} -o pid,cmd | grep -qE 'celery_bash_runner'; then
-        echo "`ps -p ${pid} -o pid,start,etime,cmd`"
-        kill -${SIG_NAME} ${pid}
+    if ps -p ${pid} -o pid,cmd | grep -qE 'celery_bash_runner'; then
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/ansible/roles/commcarehq/templates/celery_bash_runner_kill.sh.j2
```diff
@@ -0,0 +1,22 @@
...
+
+# Process to soft kill old celery jobs via supervisor managed process.
+#
+
+HOURS=$1
+MAX_MINUTES=`expr ${HOURS} \* 60 \* 60`
+
+SIG_NAME=$2
+
+PARENT_PIDS=$(ps -eo comm,ppid,etimes | awk '/celeryd:/ { if ( $4 > '"${MAX_MINUTES}"' ) { print $3 } }');
```
AmitPhulera, line 11: If we are asking for only 3 columns in `ps -eo comm,ppid,etimes`, why do we have $4 and $3 in the if condition?

gherceg: This is a great question. In testing, this somehow works but I don't have a good understanding of why.

If you run this command, you get:

And if you run a similar command without the if conditional in the awk statement, a fourth argument does not exist:
```
~/dimagi/code/commcare-cloud - (master) ps -eo comm,pid,etimes | awk '{ if ( $4 > '"${MAX_MINUTES}"' ) { print $3 } }'
11676
11678
2269837
```
```
~/dimagi/code/commcare-cloud - (master) ps -eo comm,pid,etimes | awk '{print "Column 1: "$1, "Column 2: "$2, "Column 3: "$3, "Column 4: "$4}'                              
Column 1: COMMAND Column 2: PID Column 3: ELAPSED Column 4: 
Column 1: systemd Column 2: 1 Column 3: 2055023 Column 4: 
```

@millerdev any chance you have an idea of the reason for the difference here? Clearly accessing the $4 has been working for us since this is also what we use in the existing script to kill celery processes, but it would be nice to understand why.

millerdev: I don't know, and that looks like a bug. Does it actually work?

gherceg: Oh yeah it does. Figured out it was due to whitespace in the output for the command column. For example, look at these two:
```
[celeryd: celer 3351773       7
ps              3351794       0
```
So `awk` effectively sees 4 columns since "[celeryd: celer" is separated by whitespace. So while this is "correct", it feels weirdly dependent on the command output for celery.

gherceg: While I don't love it, in the worst case (the celery command no longer has whitespace and we don't change this logic), we just wouldn't kill anything, rather than kill something we don't want to. So I'm in favor of just moving forward with this as is.

millerdev (approved):
> I'm not convinced we shouldn't just remove these scripts altogether

It seems like a good idea to investigate if we can do that. However, I'm fine with moving forward with the change in this PR if it is an improvement.

gherceg:
See [my comment](https://dimagi.atlassian.net/browse/SAAS-16545?focusedCommentId=366053) in the ticket to see the testing process carried out on staging. This is good to go.

PR #6477, opened 2025-02-27
[Production] Enable device rate limiter at 100

gherceg:
https://dimagi.atlassian.net/browse/SAAS-16559

We want to have this turned on at a high enough limit that it doesn't impact typical usage (or come anywhere close), but would help mitigate issues we've seen in the past. Based on datadog metrics outlined in [this comment](https://dimagi.atlassian.net/browse/SAAS-16559?focusedCommentId=367361), I feel 100 is an adequately high number such that this change does not require any formal announcement.

[ ] Create public facing documentation on this limit before merging

##### Environments Affected

Production


gherceg:
To roll this out, run `cchq --control production update-config` and `cchq --control service commcare restart`. If helpful, limiting to `--limit=webworkers` for both commands should be sufficient for enforcing this limit where applicable.

millerdev (approved):

gherceg:
Replaced by https://github.com/dimagi/commcare-hq/pull/35781. This setting is no longer controlled via settings but in the SystemLimit SQL table instead.

PR #6478, opened 2025-03-03
Removing es_b3001 from es cluster on production

pvisweswar:
**Ticket :** https://dimagi.atlassian.net/browse/SAAS-16652

**Environments Affected :** Production


environments/production/terraform.yml
```diff
@@ -162,7 +162,7 @@ servers:
       volume_size: 2800
     group: "elasticsearch"
     os: jammy
-    count: 2
+    count: 1
```
AmitPhulera, line 165: We should remove one more machine from `es_a3{i}-production`. We want to ensure that both azs have the same number of machines.

AmitPhulera:
Approving to be applied!

PR #6443, opened 2024-11-26
Multiplex Indices on India and on Prod

AmitPhulera:
https://dimagi.atlassian.net/browse/SAAS-16251
https://dimagi.atlassian.net/browse/SAAS-16250

To start reindexing Prod and India, I am enabling multiplexer in both the environments.

##### Environments Affected

India 
Prod

PR #6480, opened 2025-03-05
Swap all the remaining indices that have been Reindexed

AmitPhulera:
Swaps the following indices on prod
- domains
- groups
- users
- sms
##### Environments Affected

Production


mkangia (approved):
Offlined with @AmitPhulera to understand the overall process being followed and the change seems safe considering these indexes were already marked for multiplexing earlier in https://github.com/dimagi/commcare-cloud/pull/6443/files#diff-58eae44b16842253fae0d183a1b4aabe6ccf2b851248777a5a765fdd8557330d

PR #6484, opened 2025-03-11
EU env setup

AmitPhulera:
The PR has the initial work to bootstrap EU servers
##### Environments Affected

EU

### DIFF
--- a/environments/eu/terraform.yml
+++ b/environments/eu/terraform.yml
@@ -1,0 +1,400 @@
+account_alias: commcare-eu
+terraform_version: "1.5"
+state_bucket: dimagi-terraform
+state_bucket_region: "us-east-1"
+region: "eu-west-1"
+environment: "eu"
+ssl_policy: "ELBSecurityPolicy-TLS13-1-2-Res-2021-06"
+azs:
+  - "eu-west-1a"
+  - "eu-west-1b"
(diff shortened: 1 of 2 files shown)


environments/eu/terraform.yml
```diff
@@ -0,0 +1,400 @@
...
+terraform_version: "1.5"
+state_bucket: dimagi-terraform
+state_bucket_region: "us-east-1"
+region: "eu-west-1"
+environment: "eu"
+ssl_policy: "ELBSecurityPolicy-TLS13-1-2-Res-2021-06"
+azs:
+  - "eu-west-1a"
+  - "eu-west-1b"
+  - "eu-west-1c"
```
dannyroberts, line 11: Any particular to have 3 zones when we really just want to use two? (Or am I forgetting somewhere / some reason we want 3?) Otherwise, if we're just going to use two, it might be better to not even create the option of using a third one.

dannyroberts (approved):

environments/eu/terraform.yml
AmitPhulera: I think 2 is good. All the severs are only in zone a and b. I thought this might be required to ensure that we support this region? 
But yeah I agree to your point of not creating the option if we are not using it.

PR #6487, opened 2025-03-12
Update pygithub to latest version

AmitPhulera:
This is just used in deploys and I have tested it with deploy on staging and deploy went smooth.

##### Environments Affected

All

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from setuptools import setup, find_packages
+from setuptools import find_packages, setup
 
 install_deps = [
     'ansible-vault==2.1.0',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -22,7 +22,7 @@
     'netaddr',
     'passlib',
     'pycryptodome>=3.6.6',  # security update
-    'PyGithub>=1.43.3',
+    'PyGithub',
```
millerdev, line 25: This pin didn't really need to be removed since it was a lower bound, right?

millerdev (approved):

setup.py
AmitPhulera: Yeah. But I thought it might have been added to fix a security issue years back, since its a long time, no need to keep that pin around.

PR #6489, opened 2025-03-13
settings to upgrade to es 6 on india

AmitPhulera:
Upgrade to ES 6 on India
##### Environments Affected

India


environments/india/public.yml
```diff
@@ -6,9 +6,9 @@ DATADOG_ENABLED: True
 
 elasticsearch_endpoint: '{{ groups.elasticsearch.0 }}'
 #elasticsearch_memory: '4096m'
-elasticsearch_cluster_name: 'indiahqes-1.x'
-elasticsearch_version: 5.6.16
-elasticsearch_download_sha256: 6b035a59337d571ab70cea72cc55225c027ad142fbb07fd8984e54261657c77f.
+elasticsearch_cluster_name: 'indiahqes'
```
gherceg, line 9: Can you elabroate on the cluster name? Mainly how it is used/why the format changed?

gherceg (approved):

gherceg (approved):

environments/india/public.yml
gherceg: We offlined about this. The cluster name used to be used in the ES data path, but that is no longer the case, so this name doesn't really matter much. It hasn't been updated since we were on the first version of ES, and given it doesn't need to to change, this name makes sense.

PR #6490, opened 2025-03-13
Move creating /tmp out of other directories

AmitPhulera:
Since this option is configured in `java.opts` and is recently changed so if anyone who will update the config might face issue after es restarts. The commit creates another task and tags it as `es_conf ` to ensure that the directory mentioned in `jvm.opts` exists

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml
@@ -142,6 +142,17 @@
 - debug: msg="If you're just updating the ES conf then you sill need to restart the cluster. Use  the 'es_rolling_restart' playbook."
   when: copy_elasticsearch_conf_result.changed
 
+- name: Create Elasticsearch tmp directory
+  become: yes
+  file:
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml
```diff
@@ -150,7 +161,6 @@
     - "{{ elasticsearch_data_dir }}"
     - "{{ elasticsearch_data_dir }}/data"
     - "{{ elasticsearch_data_dir }}/logs"
-    - "{{ elasticsearch_data_dir }}/tmp"
```
millerdev, line 153: Curious why the `es_conf` tag was not added to this task? Does the conf file not reference data and logs directories?

millerdev (approved):

src/commcare_cloud/ansible/roles/elasticsearch/tasks/main.yml
AmitPhulera: I want just thinking about java options file and we recently updated 🤔 
I guess having it here would also make sense since it can also be updated by updating es_conf.

PR #6498, opened 2025-03-24
Stop multiplexing apps index on prod

AmitPhulera:
This is the last older index on prod. This was swapped a couple of days back and we haven't seen any issues with the newer index. After this PR is applied we would be stopping writes to older apps index and will later delete the index.

##### Environments Affected

Production

### DIFF
--- a/environments/production/public.yml
+++ b/environments/production/public.yml
@@ -170,7 +170,7 @@ localsettings:
   # When to consider multiplex settings
   ES_MULTIPLEX_TO_VERSION: '6'
   # Index Multiplexer Settings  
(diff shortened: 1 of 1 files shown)


mkangia:
Hey @AmitPhulera 

Changes look good. 
Just checking how much do we usually wait before stopping the writes to the old index?

AmitPhulera:
@mkangia  The process is to wait minimum of 1 EST working day and the index was swapped on Friday around 1:00 PM IST, so it has been an entire IST and EST working day.

mkangia (approved):
Kudos on completion of the swapping!!! :medal_sports:

PR #6508, opened 2025-03-27
Limit reprocessing concurrency to 1

gherceg:
This is a bit hacky, but we've seen this queue cause memory issues. Seems like limiting processing to one thing at a time could help (maybe). If not, we could also try moving this to the other celery machines which have much more memory headroom, but again, is a bit hacky and we are better off solving the underlying problem.

##### Environments Affected

Production


nospame:
I don't think I have the context behind this particular queue causing issues, but the change seems safe enough.

nospame (approved):

PR #6510, opened 2025-03-28
Change from m6a.4xlarge to 8xlarge

gherceg:
Temporary change to ensure trainings go smoothly.

On demand usage per day is for 4xlarge is $16.59, and for 8xlarge is $33.18. That is a difference of $16.59, which for 4 machines is 4 x 16.59 = $66.36 per day. That is just under $200 for 3 days which is how long we plan to have this rolled out for.

##### Environments Affected

Production


dannyroberts (commented):
Alongside this, probably best in this same PR, could you also update `app-processes.yml` to include lower the number of gunicorn workers per CPU, such that the absolute number of gunicorn processes stays the same, or at least doesn't double?

dannyroberts (approved):

PR #6511, opened 2025-03-28
Update WAF exceptions

mjriley:
The work prompting this change was https://dimagi.atlassian.net/browse/SAAS-16494. However, it appears there were other exceptions that were added previously that had not been added here.

##### Environments Affected
Should affect any environments which use the AWS WAF.


nospame (commented):
This seems fine based on context in Confluence about how we handle WAF exceptions, but not sure about the test failure.

dannyroberts (approved):

PR #6519, opened 2025-04-03
Route API requests to API webworkers if specified

gherceg:
In theory this PR alone shouldn't change anything, but just updates the nginx config to route API requests to the api_webworkers group if that group exists, otherwise default to webworkers, similar to how mobile webworkers operates.

##### Environments Affected

All

### DIFF
--- a/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
+++ b/src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
@@ -8,6 +8,9 @@ nginx_sites:
     - name: "{{ deploy_env }}_commcare_submissions"
       hosts: "{{ groups['mobile_webworkers'] | default('webworkers') }}"
       port: "{{ django_port }}"
+    - name: "{{ deploy_env }}_commcare_api"
+      hosts: "{{ groups['api_webworkers'] | default('webworkers') }}"
+      port: "{{ django_port }}"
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/roles/nginx/vars/cchq_ssl.yml
```diff
@@ -8,6 +8,9 @@ nginx_sites:
     - name: "{{ deploy_env }}_commcare_submissions"
       hosts: "{{ groups['mobile_webworkers'] | default('webworkers') }}"
       port: "{{ django_port }}"
+    - name: "{{ deploy_env }}_commcare_api"
+      hosts: "{{ groups['api_webworkers'] | default('webworkers') }}"
+      port: "{{ django_port }}"
```
dannyroberts, line 13: I'll point out that if you have hq_webworkers and mobile_webworkers defined, but not api_webworkers, then API requests will be spread across ALL webworkers—hq_webworkers and mobile_webworkers.

gherceg: Yeah that was the reason for the revert. I should have updated the description here.

dannyroberts: Oh sorry 😂 I missed that

PR #6526, opened 2025-04-07
[Take 2] Route API requests to api group if available

gherceg:
https://dimagi.atlassian.net/browse/SAAS-17337

Redo of https://github.com/dimagi/commcare-cloud/pull/6519, but updated the syntax to fallback to hq_webworkers _first_, and then all webworkers if that group doesn't exist.

I tested this on staging and it worked, and my understanding is that jinja is inspired by python syntax which this is.
##### Environments Affected

All


millerdev (approved):

gherceg:
Tested on staging and this behaved as expected.

PR #6528, opened 2025-04-08
Add more RAM/CPU to case pillow machine

gherceg:
https://dimagi.atlassian.net/browse/SAAS-16892

~This is a difference of $23 per day. This increases RAM from 386 to 512. At 48 processes, we are consuming 250 GB of RAM so doubling will put us just around 500GB. I'm not sure if that is quite enough headroom to feel comfortable.~

Decided to increase to r6a.24xlarge instead. So this will be an added $65 per day. 
##### Environments Affected

Production

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -277,7 +277,7 @@ servers:
     count: 1
 
   - server_name: "pillow_b2{i}-production"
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):

gherceg:
Going to merge this, and plan to make the update in a few hours.

PR #6530, opened 2025-04-09
Create command to check deploy diff

gherceg:
Just a minor `cchq` command to view the deploy diff, aka the PRs set to go out if we were to deploy right now. It is nice to get the list of PRs without interacting with the `deploy` command directly. I thought about adding a `--diff` option to the deploy command that clearly said "don't actually deploy, just show the diff", but there are already so many arguments for that command that creating a simple new command seemed fine.

##### Environments Affected

None

### DIFF
--- a/src/commcare_cloud/commands/deploy/command.py
+++ b/src/commcare_cloud/commands/deploy/command.py
@@ -112,6 +112,7 @@ class DeployDiff(CommandBase):
     def run(self, args, unknown_args):
         check_branch(args)
         environment = get_environment(args.env_name)
+        environment.release_name = 'deploy-diff'
         if args.component == 'formplayer':
             diff = get_formplayer_deploy_diff(environment)
         else:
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/commands/deploy/command.py
```diff
@@ -96,6 +96,31 @@ def run(self, args, unknown_args):
...
+        shared_args.BRANCH_ARG,
+    )
+
+    def run(self, args, unknown_args):
+        check_branch(args)
+        environment = get_environment(args.env_name)
+        environment.release_name = 'deploy-diff'
+        if args.component == 'formplayer':
+            diff = get_formplayer_deploy_diff(environment)
+        else:
```
millerdev, line 119: nit
```suggestion
        else:
            assert args.component == 'commcare', args.component
```

millerdev (approved):

millerdev (changes requested):
The test failure is due to docs not being rebuilt.

PR #6533, opened 2025-04-09
Use Java 17 with Elasticsearch 6

gherceg:
Use Java 17 on Elasticsearch 6.8.23.

Upgrading the java version allows us to use a garbage collection strategy that is a better fit for the size of our infra.

As outlined in [this matrix](https://www.elastic.co/support/matrix#matrix_jvm), Elasticsearch 6 clearly supports this version of Java, and we already have this version of Java installed on machines which makes it really convenient to switch.

This was tested on staging.
##### Environments Affected

Production, India, Staging (environments with Elasticsearch 6.8.23)

### DIFF
--- a/src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
+++ b/src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
@@ -8,6 +8,9 @@ After=network-online.target
 Type=simple
 Environment=ES_HOME={{ elasticsearch_home }}
 Environment=ES_PATH_CONF={{ elasticsearch_conf_dir}}
+{% if elasticsearch_version == '6.8.23'%}
+Environment=JAVA_HOME={{ java_17_home_path }}
+{% endif %}
 {% if elasticsearch_version <= '5.6.16'%}
 Environment=CONF_DIR={{ elasticsearch_conf_dir}}
 {% endif %}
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
```diff
@@ -8,6 +8,9 @@ After=network-online.target
 Type=simple
 Environment=ES_HOME={{ elasticsearch_home }}
 Environment=ES_PATH_CONF={{ elasticsearch_conf_dir}}
+{% if elasticsearch_version == '6.8.23'%}
```
dannyroberts, line 11: Not super urgent but worth considering what behavior we think we want for higher versions. Right now it will just revert to the default java install, which 🤷. I wonder if the right answer is to just make it hard fail if the version is 7 or higher to force someone to decide what version of java they want or something like that. But if I were to guess, we probably want whatever java we've been using in which case we kind of want `>=` here. (Also `>=` doesn't semantically understand version strings so isn't 100% appropriate, but for the actual versions that will be encountered in this scenario, it will be right.)

dannyroberts (approved):

src/commcare_cloud/ansible/roles/elasticsearch/templates/systemd/elasticsearch.service.j2
gherceg: Yeah automatically failing to force us to use the bundled version of java on versions of ES 7 and up sounds good to me.

PR #6544, opened 2025-04-16
Bump up to a reasonable value for prod env

mkangia:
We are using more of the async processing for UCRs on production now.
This bumps up the number of records we are processing in a certain time which is way low for what the environment can process.
Based on readings from datadog right now, the task finishes queuing up 20k records in less than a minute and then more records are picked up in next run of the task.
And it does appear that the records are then being processed fairly quickly as well.
This number is 30k on India. This should not increase the load on the environment at a given moment but just reduce the idle time for the queues added for processing the records.

##### Environments Affected

Production


gherceg (approved):

mkangia:
Rolled out
```
cchq production update-config --limit=celery  --branch=mk/bump-async-indicator-queue-size-prod
cchq production service celery restart --only=celery_periodic
```

PR #6555, opened 2025-04-28
increase iops to 18000

AmitPhulera:
Increase Pgmain iops to 18000
https://dimagi.atlassian.net/browse/SAAS-17469
##### Environments Affected

Prod

### DIFF
--- a/environments/production/terraform.yml
+++ b/environments/production/terraform.yml
@@ -420,6 +420,7 @@ rds_instances:
     engine_version: "14"
     backup_window: "23:00-01:00"
     backup_retention: 7
+    iops: 18000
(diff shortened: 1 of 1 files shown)


environments/production/terraform.yml
```diff
@@ -420,6 +420,7 @@ rds_instances:
     engine_version: "14"
     backup_window: "23:00-01:00"
     backup_retention: 7
+    iops: 18000
```
mkangia, line 423: This is per second?
18k, oh wow, that seems high.

AmitPhulera: Yeah. This is per second.

gherceg (approved):

PR #6568, opened 2025-05-08
add new public key

Robert-Costello:
##### Environments Affected

This is to reestablish my connection to cc-cloud with a stronger key pair.


AmitPhulera (approved):
Confirming that you are aware of the next steps on how to get this key out to all the machines?

Robert-Costello:
> Confirming that you are aware of the next steps on how to get this key out to all the machines?
Thanks, @AmitPhulera 
yes, I'll need to run update-users for each environment. In this case I'll need someone with current access to run that. I believe that covers it, but let me know if I'm missing anything.

PR #6574, opened 2025-05-09
Save gunicorn.pid file in www_home

gherceg:
A very late followup to https://dimagi.atlassian.net/browse/SAAS-15962

The service directory is "managed" by commcare-cloud in that [we will clean up non supervisor config files](https://github.com/dimagi/commcare-cloud/blob/7c1e0703ca9f6d4696c5e76718507b89905c2252/src/commcare_cloud/ansible/deploy_commcarehq.yml#L107-L113) in that directory. 

Rather than attempt to whitelist gunicorn.pid, it is easier to just move it up one level into the www_home directory.

##### Environments Affected

None

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_django.conf.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_django.conf.j2
@@ -12,7 +12,7 @@ command={{ app_processes_config.django_command_prefix }}{{ virtualenv_home }}/bi
     --log-file {{ log_home }}/{{ project }}.gunicorn.log
     --log-level debug
     --statsd-host localhost:8125
-    --pid {{ service_home }}/gunicorn.pid
+    --pid {{ www_home }}/gunicorn.pid
 user={{ cchq_user }}
 autostart=true
 autorestart=true
--- a/src/commcare_cloud/ansible/deploy_webworker.yml
+++ b/src/commcare_cloud/ansible/deploy_webworker.yml
@@ -35,7 +35,7 @@
             - notifempty
             - sharedscripts
           scripts:
-            postrotate: "[ -s {{ service_home }}/gunicorn.pid ] && kill -USR1 `cat {{ service_home }}/gunicorn.pid`"
+            postrotate: "[ -s {{ www_home }}/gunicorn.pid ] && kill -USR1 `cat {{ www_home }}/gunicorn.pid`"
 
 # There used to be some code here for updating workers one by one
 # but it's not functionally useful until we're off fab deploy and makes things


dannyroberts (approved):

AmitPhulera (approved):

src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_django.conf.j2
```diff
@@ -12,7 +12,7 @@ command={{ app_processes_config.django_command_prefix }}{{ virtualenv_home }}/bi
     --log-file {{ log_home }}/{{ project }}.gunicorn.log
     --log-level debug
     --statsd-host localhost:8125
-    --pid {{ service_home }}/gunicorn.pid
+    --pid {{ www_home }}/gunicorn.pid
```
millerdev, line 15: Is there a possibility of collision? Guessing not since there should only be one supervisor-managed django process on a given machine, but wanted to bring it to your attention in case I'm missing something.

gherceg: My cop out answer is there isn't any more of a possibility than there was before this PR. My understanding is that gunicorn uses a pre-fork worker model and therefore has one central process and a bunch of worker processes underneath it. So the collision would be between two different central processes if we were to ever want to run distinct services on the same machine? I suppose a monolith is a good example of a case where this could/would be an issue right? Are there instances where someone would want distinct django processes for things like APIs and mobile workers even on a monolith?

millerdev: My question was more around the difference between `{{ service_home }}`, which I assumed was unique for each supervisor-managed service, and `{{ www_home }}`, which would be the same across all supervisor-managed services. If any two service configurations on a given machine stored their PID in `gunicorn.pid` then they would collide.

Monolith is a separate consideration, and a good one to think about. I assume we only ever have a single `supervisor_django.conf` per monolith. Maybe it's worth checking that assumption?

gherceg: Ah I see. My understanding is that `{{ service_home }}` is the parent directory for all supervisor-managed services, not unique for each service (see [here](https://github.com/dimagi/commcare-cloud/blob/d9cc95a9962cb8d985e658c32ac67c33cf523a65/src/commcare_cloud/ansible/group_vars/all.yml#L8-L8)). Which is why I figured this change doesn't actually change the potential for a collision.

Yeah for the monolith, the default example env just has the server listed under "webworkers", but I'd assume if a user changed that to have hq_webworkers and mobile_webworkers for example, we might end up trying to run two supervisor managed django processes on the monolith?

PR #6576, opened 2025-05-12
Fixes issues es6 upgrade changelog

AmitPhulera:
Add a missed variable in changelog and also updates the update-config and restart commands 
##### Environments Affected

All

### DIFF
--- a/changelog/0087-upgrade-to-es-6.yml
+++ b/changelog/0087-upgrade-to-es-6.yml
@@ -86,8 +86,8 @@ update_steps: |
   5. Run update config and restart the services to apply the changes.
 
       ```sh
-      cchq <env> django-manage update_config
-      cchq <env> django-manage restart_services
+      cchq <env> update_config
+      cchq <env> service commcare restart
       ```
   6. Run the reindexing process.
     
@@ -175,8 +175,8 @@ update_steps: |
   2. Run update config and restart the services to apply the changes.
 
       ```sh
-      cchq <env> django-manage update_config
-      cchq <env> django-manage restart_services
+      cchq <env> update_config
+      cchq <env> service commcare restart
       ```
 
   3. Ensure that CommCare HQ is running fine with the new index. You can do it by testing the reports in your hosted CommCare instance. Optional but recommended: You can keep CommCare HQ in this state for a few working hours to ensure that the new index is stable.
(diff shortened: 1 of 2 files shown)


changelog/0087-upgrade-to-es-6.yml
```diff
@@ -47,6 +47,12 @@ update_steps: |
       ELASTICSEARCH_MAJOR_VERSION: 6
       ```
 
+      ##### Add `ES_MULTIPLEX_TO_VERSION` which tells when to consider the below multiplex settings ###
+
+      ```yaml
+      ES_MULTIPLEX_TO_VERSION: '6'
```
mkangia, line 53: I see single quotes here but not for `ELASTICSEARCH_MAJOR_VERSION` above. Are both okay?

AmitPhulera: Yes. They are both okay.
Both are used in different context. The updated variable is used in [get_es_reindex_setting_value(name, default):](https://github.com/dimagi/commcare-hq/blob/c53a7c794d2fa81c147bd7c94c22c7f3905bc186/corehq/apps/es/utils.py#L184-L195).
`ES_MULTIPLEX_TO_VERSION` which uses [ES_REINDEX_LOG](https://github.com/dimagi/commcare-hq/blob/c53a7c794d2fa81c147bd7c94c22c7f3905bc186/corehq/apps/es/const.py#L69-L71) to compare for which version reindex settings are for and values in `ES_REINDEX_LOG` are strings.
Whereas `ELASTICSEARCH_MAJOR_VERSION` is used to figure out which elasticsearch client to use in HQ which expects int https://github.com/dimagi/commcare-hq/blob/c53a7c794d2fa81c147bd7c94c22c7f3905bc186/corehq/util/es/elasticsearch.py#L16-L53

Let me know if this answers your question.

AmitPhulera:
I think we have done it before and I think it is fine - a recent example I remember is https://github.com/dimagi/commcare-cloud/pull/6543

changelog/0087-upgrade-to-es-6.yml
```diff
@@ -80,8 +86,8 @@ update_steps: |
   5. Run update config and restart the services to apply the changes.
 
       ```sh
-      cchq <env> django-manage update_config
-      cchq <env> django-manage restart_services
+      cchq <env> update_config
```
millerdev, line 89: I think this still has a typo.
```suggestion
      cchq <env> update-config
```
There are a couple of other similar lines below.

AmitPhulera: Thanks @millerdev for catching it. I don't know what I was thinking while adding them 😓 

Made a fix in - https://github.com/dimagi/commcare-cloud/pull/6580

PR #6579, opened 2025-05-13
EU Env Setup 

AmitPhulera:
The PR has configuration for entire EU setup.  This is redo of the draft branch - https://github.com/dimagi/commcare-cloud/pull/6484 with cleaned commit history.

##### Environments Affected

EU


environments/eu/public.yml
```diff
@@ -0,0 +1,223 @@
...
+
+firehose_endpoint: "firehose.eu-west-1.amazonaws.com"
+kinesis_endpoint: "kinesis.eu-west-1.amazonaws.com"
+kinesis_flows:
+  - file_pattern: "/opt/data/formplayer/log/request_response.*log"
+    delivery_stream: "formplayer-request-response-logs-partitioned-eu"
+
+enable_cloudwatch_logs: true
+
+efs_shared_mount_dir: "shared{{ '_' ~ deploy_env if deploy_env != 'production' else '' }}"
```
millerdev, line 216: This references `production`. Is that intentional?

millerdev (approved):
One small comment. If that's intentional then LGTM.

environments/eu/public.yml
AmitPhulera: Yes. That is intentional. I think we have append env name in shared path for all environments other than production.

PR #6585, opened 2025-05-19
Do not create python_env-3.9 link to 3.13 venv

millerdev:
https://dimagi.atlassian.net/browse/SAAS-16584

##### Environments Affected
all

### DIFF
--- a/src/commcare_cloud/ansible/library/setup_virtualenv.py
+++ b/src/commcare_cloud/ansible/library/setup_virtualenv.py
@@ -111,7 +111,12 @@ def main():
             if (dest / "pyproject.toml").exists():
                 uv_sync(dest, proxy, module)
                 assert (dest / ".venv").is_dir(), "uv did not create .venv"
-                (dest / full_env_name).symlink_to(".venv")
+                if (dest / f".venv/bin/python{python_version}").exists():
+                    # Only symlink if python_version matches uv python version.
+                    # Prevents virtualenv-clone of this env to an older python
+                    # version, which could happen if trying to install an old
+                    # commcare from before the switch to uv.
+                    (dest / full_env_name).symlink_to(".venv")
                 full_env_name = ".venv"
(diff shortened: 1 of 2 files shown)


gherceg (approved):

src/commcare_cloud/ansible/library/setup_virtualenv.py
```diff
@@ -111,7 +111,14 @@ def main():
             if (dest / "pyproject.toml").exists():
                 uv_sync(dest, proxy, module)
                 assert (dest / ".venv").is_dir(), "uv did not create .venv"
-                (dest / full_env_name).symlink_to(".venv")
+                if (dest / f".venv/bin/python{python_version}").exists():
+                    # Only symlink if python_version matches uv python version.
+                    # Prevents virtualenv-clone of this env to an older python
+                    # version, which could happen if trying to install an old
+                    # commcare from before the switch to uv.
```
gherceg, line 118: Reading this again, is it possible to hit this code when attempting to install a version of commcare before the switch to uv? Mainly thinkg of the check above on line 111 to ensure that a pyproject.toml exists, which I think was merged the same time as the switch to uv?

millerdev: I don't think it's likely. By the time we are rolling out Python 3.13 on non-test environments, we will be well beyond the point where everyone is required to be running a version of HQ that uses uv. In other words, it will be at least 6 weeks after https://commcare-cloud.readthedocs.io/en/latest/changelog/0088-install-uv.html was announced.

millerdev: HQ started using uv around the same time that change log was announced, so by 6 weeks later every third-party environment should be running a version of HQ that has pyproject.toml and uses uv.

gherceg (approved):

PR #6587, opened 2025-05-20
cleanup existing keys for apt

AmitPhulera:
When I added newer apt keys for Datadog in https://github.com/dimagi/commcare-cloud/pull/6570, it worked fine for newer machines. 
But when I tried to run it for production machines, I got following error - 

```
E:Conflicting values set for option Signed-By regarding source http://apt.datadoghq.com/ stable: /etc/apt/keyrings/DATADOG_APT_KEY_F14F620E.gpg != /etc/apt/keyrings/DATADOG_APT_KEY_C0962C7D.gpg, E:The list of sources could not be read.
```

The error was because `apt del <key>` was not cleaning the keys and also not cleaning the source lists. I tried the steps mentioned here in `django-manage` machine and was successful. 

So I added them to the playbook which worked fine.

https://dimagi.atlassian.net/browse/SAAS-17510
##### Environments Affected

All

### DIFF
--- a/src/commcare_cloud/ansible/roles/datadog/tasks/main.yml
+++ b/src/commcare_cloud/ansible/roles/datadog/tasks/main.yml
@@ -20,6 +20,23 @@
     repo: 'deb http://apt.datadoghq.com/ stable main'
     state: absent
 
+# the GPG keys were not getting cleaned up by the apt_key module, so we need to do it manually
+- name: Remove all old datadog key files
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - /etc/apt/keyrings/DATADOG_APT_KEY_382E94DE.gpg
+    - /etc/apt/keyrings/DATADOG_APT_KEY_F14F620E.gpg
+    - /etc/apt/keyrings/DATADOG_APT_KEY_C0962C7D.gpg
+
+# the repository configuration was not getting cleaned up by the apt_repository module, so we need to do it manually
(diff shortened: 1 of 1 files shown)


gherceg (approved):

src/commcare_cloud/ansible/roles/datadog/tasks/main.yml
```diff
@@ -20,6 +20,23 @@
...
 
+# the GPG keys were not getting cleaned up by the apt_key module, so we need to do it manually
+- name: Remove all old datadog key files
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - /etc/apt/keyrings/DATADOG_APT_KEY_382E94DE.gpg
+    - /etc/apt/keyrings/DATADOG_APT_KEY_F14F620E.gpg
+    - /etc/apt/keyrings/DATADOG_APT_KEY_C0962C7D.gpg
```
gherceg, line 31: @AmitPhulera in hindsight, `DATADOG_APT_KEY_C0962C7D.gpg` is not yet an old key right? It gets added back in "Add new datadog apt key" below. Was this intentional?

AmitPhulera: Yes. This was intentional. I think the task should be called remove all existing keys instead of old keys. We are doing it to start from a clean slate.
And if I remember correctly not removing that key added same key twice in soucrce list which again caused issues.

PR #6592, opened 2025-05-28
Add idle_in_transaction_session_timeout setting

pvisweswar:
**Ticket :** https://dimagi.atlassian.net/browse/SAAS-17641

**Environments Affected :**  ALL

### DIFF
--- a/src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
+++ b/src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
@@ -166,7 +166,7 @@ lc_time = 'en_US.UTF-8'
 
 default_text_search_config = 'pg_catalog.english'
 
-{% if postgresql_idle_in_transaction_session_timeout is defined %}
+{% if postgresql_idle_in_transaction_session_timeout %}
 idle_in_transaction_session_timeout = '{{ postgresql_idle_in_transaction_session_timeout }}'
 {% endif %}
 


src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
```diff
@@ -166,6 +166,10 @@ lc_time = 'en_US.UTF-8'
 
 default_text_search_config = 'pg_catalog.english'
 
+{% if postgresql_idle_in_transaction_session_timeout is defined %}
```
gherceg, line 169: I realized we also have this `postgresql_base/defaults/main.yml` file that sets "defaults" for postgres options commcare-cloud supports. A lot of them use `~` which appears to ensure the default postgres setting is used. Given that is how we do this for vars like [postgresql_autovacuum_vacuum_scale_factor](https://github.com/dimagi/commcare-cloud/blob/6d6c6fc342658d04851b00e199513a3ecd55f1a3/src/commcare_cloud/ansible/roles/postgresql_base/defaults/main.yml#L31-L31), I'd be in favor of adding this setting there as well, setting it to `~`, and simplifying the check on this end to just check for a truthy value.

I'm suggesting this because this is how other code in this area is structured, and it is preferable to be consistent when possible. How does that sound?

pvisweswar: Thanks for pointing that out, that makes sense to me.

Yes, I agree adding `postgresql_idle_in_transaction_session_timeout: ~` to `postgresql_base/defaults/main.yml` would help keep things consistent with how we handle other optional PostgreSQL settings (like postgresql_autovacuum_vacuum_scale_factor), and it avoids repeating logic.

src/commcare_cloud/ansible/roles/postgresql_base/defaults/main.yml
```diff
@@ -29,6 +29,7 @@ postgresql_autovacuum_multixact_freeze_max_age: ~
 postgresql_autovacuum_max_workers: ~
 postgresql_autovacuum_analyze_scale_factor: ~
 postgresql_autovacuum_vacuum_scale_factor: ~
+postgresql_idle_in_transaction_session_timeout: ~
```
pvisweswar, line 32: @gherceg Included support for idle_in_transaction_session_timeout.

src/commcare_cloud/ansible/roles/postgresql_base/templates/postgresql.conf.j2
gherceg, line 169: I think you can remove this thanks to the change you made in the most recent commit. Again thinking of how to make this look like other settings in this file.
```suggestion
{% if postgresql_idle_in_transaction_session_timeout %}
```

pvisweswar: Added this change to the PR.

millerdev (commented):
:+1: LGTM

gherceg (approved):

gherceg (approved):

PR #6597, opened 2025-06-11
Disable Kissmetrics, Drift, and Appcues

dannyroberts:
https://dimagi.atlassian.net/browse/SAAS-17704

This disconnects any API keys for these services, even for environments that have them defined. Can be merged independently from the commcare-hq commits that actually remove support for this functionality from that codebase.

Some quick testing shows that codebase already supports these keys being absent (as opposed to just set to `''` or something), but I will also apply this to staging first and test.

##### Environments Affected
all


dannyroberts:
This appears to work as expected on staging. DEBUG output in the console also verifies it understands that it's not initializing these three analytics tools in js (DEBUG-level console output with the text "Failed to initialize because apiId was not provided"). This seems very safe from my perspective

gherceg (approved):
I assume we are not concerned about self hosters using any analytics tools and accidentally disabling this for them?

Also if I understand correctly, by rolling this out before cleaning up HQ code, we should expect to make the JS console debug output a bit noisier until HQ removes references to these keys right?

Neither of those are blocking questions as this looks good and sounds safe.

dannyroberts:
> I assume we are not concerned about self hosters using any analytics tools and accidentally disabling this for them?

Yeah I'm not concerned

> Also if I understand correctly, by rolling this out before cleaning up HQ code, we should expect to make the JS console debug output a bit noisier until HQ removes references to these keys right?

Yeah that's true. Also not too concerned by that; and easy to even clean up those debug messages first before doing a larger cleanup if that becomes a pressing concern.


Thanks!

PR #6598, opened 2025-06-16
Enable 100% APM Trace sampling

Jtang-1:
This https://github.com/dimagi/formplayer/pull/1712 and the corresponding [ticket](https://dimagi.atlassian.net/browse/USH-5218) introduced a feature flag to control whether FP tracing data is sent to Datadog. Only domains with the flag enabled will have their traces reported.

The motivation for this change was to allow us to set the sampling rate to 100%, ensuring we capture all traces — especially the slower ones that are most valuable for debugging performance issues.

Now that tracing can be toggled per domain, this PR sets the sampling rate to 100% for production. (This has already been done on staging) - see this [PR](https://github.com/dimagi/commcare-cloud/pull/6594)

##### Environments Affected

Production

### DIFF
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -5,9 +5,7 @@ gunicorn_workers_static_factor: 0
 gunicorn_workers_factor: 2
 formplayer_memory: "31g"
 formplayer_g1heapregionsize: "16m"
-formplayer_command_args: ''
-# Use the following to enable datadog tracing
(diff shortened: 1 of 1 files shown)


gherceg (approved):
Don't have much context on the changes up to this point, but assume this will only send traces for domains with the flag enabled, as the description says. ✅

gherceg (approved):
I was wondering about the previous syntax 😅 should have mentioned something.

PR #6602, opened 2025-06-18
Remove HUBSPOT_API_KEY from secrets

gherceg:
Followup to https://github.com/dimagi/commcare-hq/pull/32678

The `HUBSPOT_API_KEY` is no longer referenced in commcare-hq and has been replaced by `HUBSPOT_ACCESS_TOKEN`.

##### Environments Affected

All


dannyroberts (approved):

gherceg:
To say it explicitly, given this initial change was rolled out 2 years ago, I think it is reasonable to expect that all self hosters are on a version of HQ that does not reference the old keys, and therefore we don't need a changelog announcement. If anyone disagrees with that, I'm happy to do one.

PR #6604, opened 2025-06-19
Installation Improvements

mkangia:
Some findings from the recent exercise done to test a fresh cluster install of HQ using commcare cloud.
Open to push back on any of the changes done here.

##### Environments Affected

None

### DIFF
--- a/src/commcare_cloud/ansible/roles/shared_dir/vars/main.yml
+++ b/src/commcare_cloud/ansible/roles/shared_dir/vars/main.yml
@@ -12,7 +12,6 @@ shared_mount_address: "{{ groups.shared_dir_host[0] }}:{{ shared_data_dir }}"
 restore_payload_dir_host: "{{ shared_data_dir }}/{{ restore_payload_dir_name }}"
 transfer_payload_dir_host: "{{ shared_data_dir }}/{{ transfer_payload_dir_name }}"
 shared_temp_dir_host: "{{ shared_data_dir }}/{{ shared_temp_dir_name }}"
-blobdb_dir_host: "{{ shared_data_dir }}/{{ blobdb_dir_name }}"
 
 #needed for cleanup tasks
 old_shared_dir_name: "shared_data{{ '_' ~ deploy_env if deploy_env != 'production' else '' }}"
--- a/src/commcare_cloud/ansible/group_vars/all.yml
+++ b/src/commcare_cloud/ansible/group_vars/all.yml
@@ -29,7 +29,6 @@ shared_drive_enabled: '{{ not is_monolith }}'
 transfer_payload_dir_name: "transfer_payloads"
 restore_payload_dir_name: "restore_payloads"
 shared_temp_dir_name: "temp"
-blobdb_dir_name: "blobdb"
 public_site_path: /opt/commcare-hq-public/site/output
 cchq_user: cchq
 cchq_group: cchq
--- a/src/commcare_cloud/ansible/roles/shared_dir/tasks/setup_host.yml
+++ b/src/commcare_cloud/ansible/roles/shared_dir/tasks/setup_host.yml
@@ -62,7 +62,6 @@
     - "{{ restore_payload_dir_host }}"
     - "{{ transfer_payload_dir_host }}"
     - "{{ shared_temp_dir_host }}"
-    - "{{ blobdb_dir_host }}"
   tags:
     - shared_files
 


kaapstorm (approved):

mkangia:
Hey @kaapstorm 
Added some more changes post your review. Please have a look

mkangia:
Hey @gherceg @millerdev 

Tagging you both for review, please share inputs.

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/jvm.options.j2
```diff
@@ -57,7 +57,7 @@
 {% if elasticsearch_jvm_tmp_dir is defined %}
 -Djava.io.tmpdir={{ elasticsearch_jvm_tmp_dir }}
 {% else %}
--Djava.io.tmpdir={{ elasticsearch_data_dir }}/tmp
+-Djava.io.tmpdir=/tmp
```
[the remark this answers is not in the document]
mkangia: I could try that.

I believe the contents of this directory are fine to be visible outside elasticsearch.

src/commcare_cloud/ansible/roles/shared_dir/vars/main.yml
```diff
@@ -12,6 +12,7 @@ shared_mount_address: "{{ groups.shared_dir_host[0] }}:{{ shared_data_dir }}"
 restore_payload_dir_host: "{{ shared_data_dir }}/{{ restore_payload_dir_name }}"
 transfer_payload_dir_host: "{{ shared_data_dir }}/{{ transfer_payload_dir_name }}"
 shared_temp_dir_host: "{{ shared_data_dir }}/{{ shared_temp_dir_name }}"
+blobdb_dir_host: "{{ shared_data_dir }}/{{ blobdb_dir_name }}"
```
millerdev, line 15: I don't know if a shared directory provides the proper filesystem guarantees needed by the fsdb variant of blob db. It will probably work most of the time, but may have edge cases where things are broken.

mkangia: Thanks for the inputs @millerdev 
I assume you are referring to any new installations that would be done. Any existing setups won't get impacted by this. right?

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/jvm.options.j2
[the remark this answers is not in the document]
gherceg: Ah I see. Yeah I'm not sure about defaulting to the system tmp dir, though if it clearly reproducible for you when setting up an env then it does seem like the right thing to do. Can you wait on defaulting it to /tmp until Amit returns next week as he may have some more context to share?

src/commcare_cloud/ansible/roles/shared_dir/vars/main.yml
millerdev: I haven't really considered whether this change would impact existing installations, but I would expect not since we previously did not provide an easy way to put the blob db in a shared directory. And I'm not sure it's a good idea to make it easy since we do not know if it will work. The recommended blobdb configuration for multi-server environments is to either use a third-party S3 service like AWS, or use a self-hosted service like minio.

mkangia: > The recommended blobdb configuration for multi-server environments is to either use a third-party S3 service like AWS, or use a self-hosted service like minio.

I will add that to the recommendations/documents.

The installation currently fails which is not great, so i am going to go ahead with this change unless this is a blocking feedback, then I am happy to reconsider.

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/jvm.options.j2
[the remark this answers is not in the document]
mkangia: @kaapstorm @gherceg meanwhile I have reverted this change.

src/commcare_cloud/ansible/roles/shared_dir/vars/main.yml
millerdev: I've been trying to say I don't think this is a supported configuration, so I don't think it's a good idea to use it for anything other than a test environment. It's hard for me to tell what this PR are geared toward. For example, should these changes only affect test environment setup, or may they also affect production setups? If this could affect production environment setup then I thing we should not make this change.

src/commcare_cloud/ansible/roles/elasticsearch/templates/config/jvm.options.j2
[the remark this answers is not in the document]
mkangia: Thanks @AmitPhulera 
I have reverted the change but added this in the firefighting section, https://github.com/dimagi/commcare-cloud/pull/6619/commits/c20a60150ec1e70561a2c925dccb4dea032f02a9. Feel free to share any inputs for that if you feel like we should add

PR #6605, opened 2025-06-19
HQ Python version agnostic

millerdev:
Remove Python 3.9 installation automation since it is now handled by `uv`. The deadsnakes apt repo and related Python installations will remain on existing machines.

This is now safe because six weeks have elapsed since the move to `uv`  was [announced](https://github.com/dimagi/commcare-cloud/blob/ce5fb2f297b9c9da963c7ba407ef0d6eb56ab095/docs/source/changelog/0088-install-uv.md).

:blowfish: Review by commit.

##### Environments Affected
all

### DIFF
--- a/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
+++ b/src/commcare_cloud/commands/inventory_lookup/inventory_lookup.py
@@ -305,7 +305,7 @@ class DjangoManage(CommandBase):
     ```
     commcare-cloud <env> django-manage --tmux --release 2018-04-13_18.16 shell
     ```
-    
+
     To do this on a specific server
 
     ```
@@ -413,7 +413,8 @@ class ForwardPort(CommandBase):
         local_port = self.get_random_available_port()
 
         while not self.is_loopback_address_set_up(loopback_address):
-            puts(color_error('To make this work you will need to run set up a special loopback address on your local machine:'))
+            puts(color_error('To make this work you will need to run set up a '
+                             'special loopback address on your local machine:'))
             puts(color_notice(f'  - Mac: Run `sudo ifconfig lo0 alias {loopback_address}`.'))
             puts(color_notice(f'  - Linux: Run `sudo ip addr add {loopback_address}/8 dev lo`.'))
             if not ask("Follow the instructions above or type n to exit. Ready to continue?"):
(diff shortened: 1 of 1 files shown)


src/commcare_cloud/ansible/library/setup_virtualenv.py
```diff
@@ -92,42 +77,21 @@ def main():
...
-            else:
-                if not (next_env / "bin/python").exists():
-                    if not prev_env.exists():
-                        module.fail_json(msg=f"virtualenv not found: {prev_env}")
-                        return
-                    clone_virtualenv(prev_env, next_env, module)
-                pip_sync(dest, next_env, module, proxy)
-            python_env.symlink_to(full_env_name)
+            uv_sync(dest, proxy, module)
+            assert next_env.is_dir(), f"uv did not create {next_env}"
```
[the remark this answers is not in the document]
millerdev: This would be reasonable if the error were ever expected to happen, but an assertion seems reasonable since it is never expected to happen as a result of user input variation.

tests/test_deploy/test_setup_virtualenv.py
```diff
@@ -33,105 +31,41 @@ def setUp(self):
...
-                    "dest": str(self.release),
-                    "env_name": "venv",
-                    "python_version": PYTHON_VERSION,
-                })
-        assert (self.release / "venv-3.9/bin/python").exists()
-        assert not (self.release / "venv").exists(), listdir(self.release)
-        with patch.object(setup_virtualenv, "clone_virtualenv") as mock:
-            mock.side_effect = Fail("unexpected: env already created")
-            self.test_setup_virtualenv()
+        self.test_setup_virtualenv()  # should not fail
```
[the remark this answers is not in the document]
millerdev: Too nitpicky.

tests/test_deploy/test_setup_virtualenv.py
```diff
@@ -31,82 +31,31 @@ def setUp(self):
...
-            mock.side_effect = clone_fail
+        with patch.object(setup_virtualenv, "uv_sync", uv_sync):
             with self.assertRaises(Fail):
                 ansible.run(setup_virtualenv, {
-                    "src": str(self.previous_release),
                     "dest": str(self.release),
                     "env_name": "venv",
                 })
+        assert not (self.release / ".venv").exists(), listdir(self.release)
         assert not (self.release / "venv").exists(), listdir(self.release)
```
gherceg, line 53: Is this line still necessary?

tests/test_deploy/test_setup_virtualenv.py
```diff
@@ -31,82 +31,31 @@ def setUp(self):
...
-                ansible.run(setup_virtualenv, {
-                    "src": str(self.previous_release),
-                    "dest": str(self.release),
-                    "env_name": "venv",
-                })
-        assert not (self.release / "venv").exists(), listdir(self.release)
-        with patch.object(setup_virtualenv, "clone_virtualenv") as mock:
-            mock.side_effect = Fail("unexpected: env already created")
-            self.test_setup_virtualenv()
+        self.test_setup_virtualenv()  # should not fail
```
gherceg, line 55: I don't really understand why we need this line. Presumably this test runs on its own already, and I guess the reason to run it here is to ensure there isn't a knock on effect from the failed setup? But it seems like at that point we are testing the validity of the context manager or something.

.tests/tests.sh
```diff
@@ -2,16 +2,7 @@
...
-eval PYVER=$(grep python_version: $GROUP_VARS | head -n1 | cut -d' ' -f2)
-# Link GitHub Actions-installed python where ansible virtualenv will find it
-[ -z $(which python${PYVER}) ] && sudo ln -s /opt/python/${PYVER}/bin/*${PYVER}* /usr/local/bin/
-which python${PYVER}
-python${PYVER} --version
-# link virtualenv to a place where it will be found in ansible
-[ -e /opt/pipx_bin/virtualenv ] && sudo ln -s /opt/pipx_bin/virtualenv /usr/local/bin/
-which virtualenv
-virtualenv --version
+python --version
```
gherceg, line 5: 😍

gherceg (approved):
Beautiful cleanup! Nothing blocking, just a couple of comments on a test.

tests/test_deploy/test_setup_virtualenv.py
millerdev: Yes, this tests that the `env_name` symlink was created.

tests/test_deploy/test_setup_virtualenv.py
millerdev: It's to verify that virtualenv setup works properly after a failure.

snopoke (commented):
Just confirming that we're comfortable using the python distributions from [python-build-standalone](https://gregoryszorc.com/docs/python-build-standalone/main/). These are not official python packages and do have some quirks and security consiertions.

There's short writeup with some thought from Oct 2024 here: https://pythonspeed.com/articles/uv-python-production/

I think it's probably fine but just wanting to make sure someone on SaaS has given this proper thought.

PR #6607, opened 2025-06-24
Add ENABLE_BHA_CASE_SEARCH_ADAPTER setting

millerdev:
Previously this was bundled under the umbrella `IS_SAAS_ENVIRONMENT` setting, but it should only be enabled on a subset of SaaS environments. See also: https://dimagi.atlassian.net/browse/SAAS-11593 and https://github.com/dimagi/commcare-hq/pull/36458#discussion_r2161879108

This will cause the BHA ES index to no longer be enabled on the EU environment. Is any follow-up maintenance needed there (maybe delete the index)?

The second commit is cleanup of a duplicate setting, and completely unrelated to the primary purpose of this PR.

##### Environments Affected
eu, production, staging


gherceg:
> This will cause the BHA ES index to no longer be enabled on the EU environment. Is any follow-up maintenance needed there (maybe delete the index)?

I only recently merged that PR and haven't run anything on EU to create the index, so no follow-up maintenance necessary. I just confirmed that the index does not exist on EU as well.

gherceg (approved):

PR #6613, opened 2025-06-26
Add mriley user for new laptop

mjriley:
Adds 'mriley' user for Matt's new laptop. During the previous setup, I had been set up with 'matthewriley' which I don't want to replicate again.

##### Environments Affected
None, just adds an alias to an existing user


mkangia:
I assume you would add `matthewriley` in absent section as a follow up?

mjriley:
> I assume you would add `matthewriley` in absent section as a follow up?

Yep. I'm not quite done with my mac, so I was planning on retaining `matthewriley` until I was finished there, but I intend to remove the user afterwards

PR #6628, opened 2025-07-08
Remove python 2 related cruft

gherceg:
🐟 Review by commit

Just a cleanup that shouldn't have any change in behavior. Tried to keep commits bite sized until 57f1ae5f86349219daf3369317577af9acd4d5c4 where I just decided to do it all since those imports aren't necessary on python 3.

##### Environments Affected

None


src/commcare_cloud/environment/schemas/role_defaults.py
```diff
@@ -18,7 +18,7 @@ def get_defaults_jsonobject(role, **kwargs):
     :return: The new JsonObject subclass.
     """
     cls = type(jsonobject.JsonObject)(
-        str('{}_Defaults'.format(role)),  # bytes in Python 2, text in Python 3
+        str("{}_Defaults".format(role)),
```
millerdev, line 19: nit
```suggestion
        "{}_Defaults".format(role),
```

millerdev (approved):

PR #6630, opened 2025-07-08
Remove attrs

gherceg:
I don't see any references to attrs in commcare-cloud. It was first added in [this commit](https://github.com/dimagi/commcare-cloud/commit/6a5a086), and imagine it may have been the case that it didn't end up being used as expected but slipped into requirements?

##### Environments Affected

None


gherceg:
Just kidding, it was `import attr` :)

PR #6633, opened 2025-07-08
Put current GuardDuty configuration in terraform

dannyroberts:
https://dimagi.atlassian.net/browse/SAAS-17846

##### Environments Affected
eu, india (no diffs on staging, production)

### DIFF
--- a/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
+++ b/src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
@@ -43,6 +43,27 @@ locals {
   }
 }
 
+resource "aws_guardduty_detector" "detector" {
+  enable = true
+  finding_publishing_frequency = "SIX_HOURS"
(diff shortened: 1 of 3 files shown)


src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
```diff
@@ -43,6 +43,27 @@ locals {
...
+            ebs_volumes {
+                enable = false
+            }
+        }
+    }
+    s3_logs {
+        enable = false
+    }
+  }
+}
```
gherceg, line 66: For my own understanding, where/how do you figure out what goes here?

gherceg (approved):
Should there be terraform imports for india and eu as well?

dannyroberts:
> Should there be terraform imports for india and eu as well?
Good question. No, but to see why, check the ticket, and feel free to follow up there

PR #6641, opened 2025-07-21
Default to elasticsearch 6.8.23

gherceg:
ES 5 is no longer supported by commcare, so we should be defaulting to 6 for new setups.
##### Environments Affected

None

### DIFF
--- a/quick_monolith_install/sample-environment/public.yml.j2
+++ b/quick_monolith_install/sample-environment/public.yml.j2
@@ -2,7 +2,7 @@
 {% endraw %}
 elasticsearch_memory: '2048m'
 elasticsearch_cluster_name: '{{env_name}}-es'
-elasticsearch_version: 5.6.16
+elasticsearch_version: 6.8.23
 elasticsearch_download_sha256: 6b035a59337d571ab70cea72cc55225c027ad142fbb07fd8984e54261657c77f.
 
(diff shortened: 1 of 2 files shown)


millerdev (approved):
Do these files need to be updated every time we upgrade to a new patch release of ES? That seems less than ideal.

millerdev:
Maybe it's fine? Are these files sort of like uv.lock or yarn.lock for ES?

gherceg:
@AmitPhulera would know better, but I do think we would need to update these any time we upgrade even a patch release for ES. Agreed it doesn't seem ideal.

quick_monolith_install/sample-environment/public.yml.j2
```diff
@@ -2,7 +2,7 @@
 {% endraw %}
 elasticsearch_memory: '2048m'
 elasticsearch_cluster_name: '{{env_name}}-es'
-elasticsearch_version: 5.6.16
+elasticsearch_version: 6.8.23
 elasticsearch_download_sha256: 6b035a59337d571ab70cea72cc55225c027ad142fbb07fd8984e54261657c77f.
```
AmitPhulera, line 6: We would need to update `elasticsearch_download_sha256` too.

AmitPhulera: Updated in https://github.com/dimagi/commcare-cloud/pull/6644

PR #6646, opened 2025-07-29
encrypted = true for CouchDB EBS volume

pvisweswar:
**Ticket :** https://dimagi.atlassian.net/browse/SAAS-18008

**Environment :** India

**Note :**

Recently, we [downsized and encrypted the disk on the Couch servers](https://dimagi.atlassian.net/browse/SAAS-15878). During that process, we missed adding the volume_encrypted: yes parameter in the terraform.yml file. As a result, running terraform plan was marking the volumes for deletion. We've now added the parameter to ensure the volumes remain safe, and everything is running smoothly.

### DIFF
--- a/environments/india/terraform.yml
+++ b/environments/india/terraform.yml
@@ -99,7 +99,7 @@ servers:
     network_tier: "db-private"
     az: "a"
     volume_size: 60
-    volume_encrypted: no
+    volume_encrypted: yes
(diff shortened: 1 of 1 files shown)


AmitPhulera:
@pvisweswar Can you add details on the PR why we need this change?

AmitPhulera (approved):

PR #6656, opened 2025-08-11
instead of AZ `c` use AZ `a` which is similar to that of production

AmitPhulera:
It gets slightly confusing when restoring the volumes from backups, good to keep it consistent with the production.

Do not merge it until all the backup env is destroyed after test.

##### Environments Affected

Backup Production

### DIFF
--- a/environments/backup-production/terraform.yml
+++ b/environments/backup-production/terraform.yml
@@ -6,11 +6,11 @@ region: "us-east-2"
 environment: "bk-production"
 ssl_policy: 'ELBSecurityPolicy-FS-1-2-Res-2020-10'
 azs:
+  - "us-east-2a"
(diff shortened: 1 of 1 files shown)


gherceg (approved):
Just double checking that there wasn't a strategic reason we setup AZs like this in the first place? Or if there was, if we think that reason isn't enough to justify the confusion this can lead to when restoring the backup environment.

AmitPhulera:
> Just double checking that there wasn't a strategic reason we setup AZs like this in the first place? Or if there was, if we think that reason isn't enough to justify the confusion this can lead to when restoring the backup environment.

I missed responding to it. No, I don't think there is any strategic reason and it should not affect cost.

PR #6660, opened 2025-08-14
Update PR template to include a checkbox to notify devs to ensure to be cognizant of changes that affects multiple envs

AmitPhulera:
This came up during a discussion with @gherceg  where we can easily miss that a particular change might be affecting multiple environments. This shows up when we try to apply some other changes and its hard to find context on them. This is a small step to nudge people to think about other environments.


gherceg (approved):
🚀

PR #6666, opened 2025-08-20
Gh/elasticache/celery broker

gherceg:
Prototyping using elasticache for celery broker on staging.

##### Environments Affected

Staging


src/commcare_cloud/commands/terraform/templates/commcarehq.tf.j2
```diff
@@ -231,6 +231,32 @@ module "elasticache-module-base" {
 }
 {% endif %}
 
+{% if celery_broker %}
+module "elasticache-module-base" {
+  source                = "./modules/elasticache-cluster"
```
gherceg, line 236: Feels like this is a key line that I assume will not work out of the box? I'm concerned about overlapping resources used for the redis cluster and this cluster like [this](https://github.com/dimagi/commcare-cloud/blob/d420758afddfc260b072c2a20253eeb9c64d1eee/src/commcare_cloud/terraform/modules/elasticache-cluster/main.tf#L1-L5). @AmitPhulera curious to hear your thoughts

PR #6668, opened 2025-08-21
add credentials url to settings

zandre-eng:
Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1372).
Link to related HQ PR [here](https://github.com/dimagi/commcare-hq/pull/36803).

##### Environments Affected

- production
- staging
- backup-production

- [X] If the changes affect multiple environments, I will ensure they are rolled out consistently across all environments.


zandre-eng:
> Do we need to set it for EU and India environment as well?

@AmitPhulera Good question. Given that the related HQ PR includes a Celery task that will run on all environments we should add this setting to EU and India as well. I'll push a commit for this.

AmitPhulera (approved):

PR #6678, opened 2025-08-28
Revert "tempororily push couch11 backups to BCP Vault"

AmitPhulera:
Reverts dimagi/commcare-cloud#6670 

We have copied the backups, now this is no longer required.


AmitPhulera:
Forgot to merge this earlier but this is not required anymore.

PR #6693, opened 2025-09-30
Don't use second_line for invoice_from_address

nospame:
It's supported by the Address class, which is also used for customer billing addresses, but not by the Invoice PDF template where we put the invoice "from "address, see:
image

Condensing it into a single line does fit the Invoice PDF template:
image

Follow-up to #6692

##### Environments Affected
production, eu, india, backup-production, staging


gherceg (approved):
Not blocking, but seems like we should make a note of this "bug" with `second_line` in HQ.

PR #6696, opened 2025-10-06
Remove localsettings_private namespace

gherceg:
Followup to https://github.com/dimagi/commcare-cloud/pull/6643.

##### Environments Affected

All

My understanding is this blocked `update-config` from running successfully, so it isn't that anything needs to be run on all environments, but this does affect all environments.

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -415,8 +415,8 @@ REPORT_CACHE = 'redis'
 redis_cache = {
     'BACKEND': 'django_redis.cache.RedisCache',
 {% set redis_url = 'redis://' ~ localsettings.REDIS_HOST ~ ':' ~ localsettings.REDIS_PORT %}
-{% if localsettings_private.get('REDIS_PASSWORD') %}
-{% set redis_url = 'redis://:' ~ localsettings_private.REDIS_PASSWORD ~ '@' ~ localsettings.REDIS_HOST ~ ':' ~ localsettings.REDIS_PORT %}
+{% if REDIS_PASSWORD %}
+{% set redis_url = 'redis://:' ~ REDIS_PASSWORD ~ '@' ~ localsettings.REDIS_HOST ~ ':' ~ localsettings.REDIS_PORT %}
 {% endif %}
 {% if 'REDIS_DB' in localsettings %}
     'LOCATION': '{{ redis_url }}/{{ localsettings.REDIS_DB }}',
--- a/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
+++ b/src/commcare_cloud/ansible/roles/formplayer/templates/application.properties.j2
@@ -51,8 +51,8 @@ redis.clusterString={{ groups.redis_cluster_master | join(":" + localsettings.RE
 redis.hostname={{ localsettings.REDIS_HOST }}
 {% endif %}
 
-{% if localsettings_private.get('REDIS_PASSWORD') %}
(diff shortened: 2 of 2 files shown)


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -415,8 +415,8 @@ REPORT_CACHE = 'redis'
 redis_cache = {
     'BACKEND': 'django_redis.cache.RedisCache',
 {% set redis_url = 'redis://' ~ localsettings.REDIS_HOST ~ ':' ~ localsettings.REDIS_PORT %}
-{% if localsettings_private.get('REDIS_PASSWORD') %}
-{% set redis_url = 'redis://:' ~ localsettings_private.REDIS_PASSWORD ~ '@' ~ localsettings.REDIS_HOST ~ ':' ~ localsettings.REDIS_PORT %}
+{% if REDIS_PASSWORD %}
```
millerdev, line 418: Prior to https://github.com/dimagi/commcare-cloud/pull/6643, this was `localsettings.get('REDIS_PASSWORD')`. Was that just always wrong?

Looks like [`REDIS_PASSWORD` is defined in a `locasettings_private` namespace](https://github.com/dimagi/commcare-cloud/pull/6643/files#diff-9105ce9cf2682b136266267595ce85567d0b070fb945bfa75797bdf09c1297b7) in the linked PR. The non-obvious part is the thing that copies it from that namespace to the global namespace. My understanding of how this works now is via the _secrets backend_ and _generated variables_:

https://github.com/dimagi/commcare-cloud/blob/1e3fb2bc02b44c05dd1b582f9d86c43557894442/src/commcare_cloud/environment/secrets/secrets_schema.py#L56-L63

https://github.com/dimagi/commcare-cloud/blob/1e3fb2bc02b44c05dd1b582f9d86c43557894442/src/commcare_cloud/environment/secrets/secrets.yml#L132-L133

https://github.com/dimagi/commcare-cloud/blob/1e3fb2bc02b44c05dd1b582f9d86c43557894442/src/commcare_cloud/environment/main.py#L395

I think the _generated variables_ bit is how it ends up in the global namespace. Do I have that right?

millerdev (approved):

mkangia:
Thanks for catching & fixing it @gherceg 

I am a bit surprised by this. I believe I did run a full cluster install after my changes. 
May be there is a difference in how the different level of variables are available during update config as compared to full install?

src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
mkangia: > I'm having trouble wrapping my head around what a self hoster would have had to do to make that work as is

So this is needed for a cluster. This won't be needed for a monolith. I believe that is why this was not reported before by any independent hoster, which are all monolith I think.

src/commcare_cloud/ansible/roles/redis/templates/redis_is_master.j2
```diff
@@ -3,8 +3,8 @@
 # This is a helper script to check if the current redis instance is the master of the replication scheme.
 # This accounts for cases when the local redis instance is down. Since that means this node can't be the master, we handle
 # that gracefully.
-{% if localsettings_private.get('REDIS_PASSWORD') %}
-REPLICATION_DATA=`redis-cli -a {{ localsettings_private.REDIS_PASSWORD }} info replication 2> /dev/null`
+{% if REDIS_PASSWORD %}
```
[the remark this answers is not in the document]
mkangia: >  if REDIS_PASSWORD is not defined (assuming that is possible)

I think not having the password defined is possible.

PR #6700, opened 2025-10-07
Overhaul Datadog APM configuration for commcare processes

millerdev:
Unify environment variable configuration for the main process groups (celery, management command, pillow, webworker). The motivation is to simplify configuration of Datadog APM on specific processes.

Remove app_processes_config.datadog_pythonagent, which was problematic because its scope was too broad. For example, enabling it in _app-processes.yml_ caused an effective setting of `DD_AGENT_ENABLED=True` (which is the default when that variable is not set) on all commcare processes.

`DD_TRACE_ENABLED=False` is set as the default to prevent "failed to send, dropping N traces" logs in environments that do not use Datadog. See https://github.com/dimagi/commcare-cloud/pull/6404 for more on that.

APM can be enabled by setting `app-processes.yml:...env_vars.DD_TRACE_ENABLED=True`. See _roles/commcare/tasks/*.yml_ for specific variable paths.

##### Environments Affected
production/all

##### Announce New Release (probably not needed)


gherceg:
@millerdev we had talked about a changelog for this, but I forgot to include that in my review. I still think that is necessary here just to give self hosters notice of this change.

src/commcare_cloud/ansible/roles/commcarehq/tasks/celery.yml
```diff
@@ -38,6 +38,4 @@
         PROMETHEUS_PUSHGATEWAY_HOST: "{% if prometheus_celery_pushgateway is defined %}{{ prometheus_celery_pushgateway }}{% endif %}"
         TMPDIR: '{{ encrypted_tmp }}'
         DD_TRACE_ENABLED: "{% if not app_processes_config.datadog_pythonagent %}False{% endif %}"
-        DD_TRACE_DJANGO_ENABLED: "{% if not app_processes_config.datadog_pythonagent %}False{% endif %}"
-        DD_TRACE_DEBUG: "{% if not app_processes_config.datadog_pythonagent %}False{% endif %}"
```
millerdev, line 42: @zwets re: https://github.com/dimagi/commcare-cloud/pull/6404

After reading the ddtrace source code and consulting Datadog support (see [commit message](https://github.com/dimagi/commcare-cloud/pull/6700/commits/d0ae7b15d715b8130f1618e6188757af40c119a3)) I concluded that these were unnecessary. Do you agree that these seem unnecessary?

PR #6858, opened 2026-04-14
Ansible module to start/stop/restart ec2 instances

AmitPhulera:
https://dimagi.atlassian.net/browse/SAAS-19382

As part of our ongoing efforts to automate the restarts, we would need ansible to have a module to start/stop ec2 services. Having it in ansible as opposed to python scripts would give us the advantage of using it directly into the playbooks instead of calling a python command from ansible.

I have had a boilorplate from where claude started the implementation. After that it asked couple of good questions to reach the final implementation. 

I have kept spec and implementation plan in first and second commit but will remove it before merging the PR..

##### Environments Affected

All


src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -0,0 +1,179 @@
...
+description:
+    - Manages the running state of EC2 instances given an explicit list of
+      instance IDs. Supports four states - started, stopped, restarted,
+      described - and is idempotent (no API call is made if the instance is
+      already in the requested state).
+    - Designed to run with delegate_to localhost. AWS credentials and the
+      target region are picked up from the standard boto3 credential chain;
+      in the commcare-cloud workflow the AWS_PROFILE and AWS_REGION
+      environment variables are exported automatically before ansible runs.
+    - The restarted state is implemented as stop-then-start. The stop phase
```
gherceg, line 24: Any particular reason to issue a stop-then-start as opposed to a "reboot"? My understanding is that a reboot is faster since it doesn't put the VM on new hardware, but I know there are times where we want to do a full stop-then-start.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -0,0 +1,179 @@
...
+      stopping instance cannot be started); the start phase respects wait.
+
+version_added: "1.0.0"
+
+options:
+    instance_ids:
+        description: List of EC2 instance IDs to act on.
+        required: true
+        type: list
+        elements: str
```
gherceg, line 35: Is it possible to make this more flexible? Could it accept a name for "group" or "hostname" and translate that to the instance id(s) under the hood?

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -0,0 +1,179 @@
...
+            "AWS region not provided. Pass 'region' to the module, "
+            "or set the AWS_REGION environment variable."
+        ))
+    return region
+
+
+def _get_ec2_client(region):
+    """Return a boto3 EC2 client. Defined as a module-level function so tests can patch it."""
+    try:
+        import boto3
```
gherceg, line 136: Is this typically installed with ansible or is it a special requirement for this module?

tests/test_ec2_instance_state.py
```diff
@@ -0,0 +1,134 @@
...
+# Add the library directory to sys.path so we can import the module by name.
+# We must also ensure that the tests/ directory does NOT appear before the
+# library dir on sys.path, because tests/ansible.py would otherwise shadow
+# the real `ansible` package when ec2_instance_state does
+# `from ansible.module_utils.basic import AnsibleModule`.
+import os
+LIBRARY_DIR = os.path.abspath(os.path.join(
+    os.path.dirname(__file__), '..', 'src', 'commcare_cloud', 'ansible', 'library'
+))
+TESTS_DIR = os.path.abspath(os.path.dirname(__file__))
```
gherceg, line 16: nit: Can you define `TESTS_DIR` before `LIBRARY_DIR` and reference `TESTS_DIR` in the construction of `LIBRARY_DIR`?

tests/test_ec2_instance_state.py
```diff
@@ -0,0 +1,134 @@
...
+        if self._orig_region is not None:
+            os.environ['AWS_REGION'] = self._orig_region
+
+    def test_missing_region_fails(self):
+        result = run_module({
+            'instance_ids': ['i-0123456789abcdef0'],
+            'state': 'described',
+        })
+        self.assertTrue(result['failed'])
+        self.assertIn('region', result['result']['msg'].lower())
```
gherceg, line 97: nit: use pytest simple assertions (assert result['failed'], assert 'region' in result...)

tests/test_ec2_instance_state.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def test_invalid_state_fails(self):
+        os.environ['AWS_REGION'] = 'us-east-1'
+        result = run_module({
+            'instance_ids': ['i-0123456789abcdef0'],
+            'state': 'bogus',
+        })
+        self.assertTrue(result['failed'])
```
gherceg, line 105: This should also assert that the message is as expected similar to the other tests.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -141,6 +141,69 @@ def _get_ec2_client(region):
...
+        raw_by_id = _describe_instances(client, instance_ids)
+    except ClientError as e:
+        module.fail_json(msg="AWS DescribeInstances failed: {}".format(e))
+    # Order output to match input.
+    formatted = []
+    states = {}
+    for iid in instance_ids:
+        raw = raw_by_id.get(iid)
+        if raw is None:
+            module.fail_json(msg="Instance {} missing from DescribeInstances response.".format(iid))
```
gherceg, line 188: Is it possible to fail just for this instance rather than the whole group?

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -141,6 +141,69 @@ def _get_ec2_client(region):
...
+        'instance_type': raw.get('InstanceType'),
+        'availability_zone': (raw.get('Placement') or {}).get('AvailabilityZone'),
+        'private_ip': raw.get('PrivateIpAddress'),
+        'public_ip': raw.get('PublicIpAddress'),
+        'tags': tags,
+        'launch_time': launch_time,
+    }
+
+
+def _describe_and_format(client, instance_ids, module):
```
gherceg, line 175: Could you rename "format" in this context to avoid conflicting with the `_format_instance` helper defined above? We expected to see that called in here.

tests/test_ec2_instance_state.py
```diff
@@ -131,5 +131,166 @@ def test_valid_instance_id_pattern(self):
...
+    def test_described_terminated_is_reported_not_failed(self):
+        fake = FakeEC2Client(instances_by_id={
+            'i-0123456789abcdef0': {'state': 'terminated'},
+        })
+        result = run_module(
+            {'instance_ids': ['i-0123456789abcdef0'], 'state': 'described'},
+            fake_client=fake,
+        )
+        self.assertFalse(result['failed'])
+        self.assertEqual(result['result']['instances'][0]['current_state'], 'terminated')
```
gherceg, line 283: Why do we need this test? We don't validate the `described` response so why would we expect terminated to be special?

gherceg (commented):
Pair reviewed with @millerdev. We reviewed commits before 15354b0ab2dc2ecafd2519b011c758e362ab0364

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: This will essentially replace our restart processes which requires us to start and stop the instances rather than issuing reboot. This is because it is required control for APN and also the reason you mentioned of probably getting a better hardware.
I think we can add a ec2_reboot that does that but as now control machine doesn't have permission to do that. So I can probably create a followup PR with the restart permissions and updating the module?

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: It is possible but this is just the module and there will be a wrapper command that will call the module which will do the translation. What I am envisioning is a command like

`cchq <env> ec2_stop <instance_id|host|host_group>`

Do you think that should be enough or do you think there is a requirement that would need us to include the group and hostname in the module itself?

fwiw there is an example on how to do use `host` or `host_group` when you try to call it from a playbook.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: This is a special requirement for the module but given this module will always be called from `cchq` venv `boto3` should be present. 

`boto3` is imported in this way that it can be independently called as well and should elegantly exit if dependency is not there.

src/commcare_cloud/ansible/library/ec2_instance_state.py
gherceg: Cool that makes sense. No need to worry about that followup PR until we find that we need it.

gherceg: Do you think it is worth renaming this state to "stopped and started" just to be clear? This avoids needing to distinguish between restarted and rebooted in the future, and makes it clear what this state is doing without needing to read docs.

src/commcare_cloud/ansible/library/ec2_instance_state.py
gherceg: It seems ideal to put all logic related to translating groups/hosts to instance ids inside the module. Do we need the ability to pass in instance ids to this module, or would ansible names (groups/hosts) be sufficient?

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -174,18 +175,19 @@ def _format_instance(raw, previous_state, current_state):
 
 def _describe_and_format(client, instance_ids, module):
     """Describe + format. Fails the module on AWS errors."""
-    from botocore.exceptions import ClientError
     try:
         raw_by_id = _describe_instances(client, instance_ids)
     except ClientError as e:
         module.fail_json(msg="AWS DescribeInstances failed: {}".format(e))
+        return
```
gherceg, line 182:
```suggestion
        return None, None
```

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -194,6 +194,101 @@ def _describe_and_format(client, instance_ids, module):
...
+def _check_no_terminal(formatted, action_state, module):
+    """Fail the module if any instance is terminated/shutting-down."""
+    bad = [(iid, state) for (iid, _raw, state) in formatted if state in TERMINAL_STATES]
+    if bad:
+        module.fail_json(msg=(
+            "Cannot {} terminated/shutting-down instances: {}".format(
+                action_state, ', '.join('{}={}'.format(i, s) for i, s in bad)
+            )
+        ))
+        return
```
gherceg, line 206: This comment applies to more than just this function, but it is hard to follow the `formatted` list of tuples getting passed around. The name does not adequately describe what is contained in the data structure. It would be nicer to pass around a list of `Instance` objects. Having an `Instance` class could also simplify this code. For example, `_describe_instances` would then return a dictionary of `id: Instance(...)` items, and the `Instance` class would provide attributes to obtain whatever info you need.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -194,6 +194,101 @@ def _describe_and_format(client, instance_ids, module):
...
+
+    refresh=True: re-describe to compute after_states from live data.
+    refresh=False: use the supplied after_states (used for check_mode and no-op fast paths).
+    """
+    if refresh:
+        formatted, after_states = _describe_and_format(client, instance_ids, module)
+        raw_by_id = {iid: raw for (iid, raw, _state) in formatted}
+    else:
+        formatted, _ = _describe_and_format(client, instance_ids, module)
+        raw_by_id = {iid: raw for (iid, raw, _state) in formatted}
```
gherceg, line 238: This doesn't make sense. Regardless of `refresh`, we call `_describe_and_format` which calls `_describe_instances` again. So `refresh=False` isn't going to be faster, it is just going to call describe again but ignore the `after_states`. 

This appears to be AI generated code that shouldn't be part of a PR up for review.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -194,6 +194,101 @@ def _describe_and_format(client, instance_ids, module):
     return formatted, states  # list of (id, raw, state); dict id->state
 
 
+def _check_no_terminal(formatted, action_state, module):
```
gherceg, line 197:
```suggestion
def _check_not_terminated(formatted, action_state, module):
```
We are aware that this checks for both terminated and shutting down states, but this name is easier to understand at a glance.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -194,6 +194,101 @@ def _describe_and_format(client, instance_ids, module):
...
+              'after': {'states': after_states}},
+    )
+
+
+def _do_start(module, client, instance_ids, wait, timeout):
+    formatted, before_states = _describe_and_format(client, instance_ids, module)
+    _check_no_terminal(formatted, 'start', module)
+
+    targets_needing_start = [iid for (iid, _raw, state) in formatted
+                             if state in ('stopped', 'stopping')]
```
gherceg, line 259: Would it be worth adding constants for these states? This comment applies throughout the PR.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -194,6 +194,101 @@ def _describe_and_format(client, instance_ids, module):
...
+    skipped = [iid for iid in instance_ids if iid not in targets_needing_start]
+
+    # If any instance is currently 'stopping' and we'll need to start it, wait for stopped first.
+    stopping_now = [iid for (iid, _raw, state) in formatted if state == 'stopping']
+    if wait and stopping_now:
+        _wait_for(client, 'instance_stopped', stopping_now, timeout, module)
```
gherceg, line 265: If in check mode, should this be waiting?

tests/test_ec2_instance_state.py
```diff
@@ -291,5 +291,97 @@ def test_described_invalid_id_fails(self):
         self.assertIn('NotFound', result['result']['msg'])
 
 
+class TestStarted(unittest.TestCase):
+
+    def setUp(self):
+        os.environ['AWS_REGION'] = 'us-east-1'
+
+    def tearDown(self):
+        os.environ.pop('AWS_REGION', None)
```
gherceg, line 300: Should this restore the original value if set? This applies throughout the PR.

tests/test_ec2_instance_state.py
```diff
@@ -291,5 +291,97 @@ def test_described_invalid_id_fails(self):
         self.assertIn('NotFound', result['result']['msg'])
 
 
+class TestStarted(unittest.TestCase):
```
gherceg, line 294: Should include a test for check mode behavior, specifically checking wait calls.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -261,6 +261,14 @@ def _do_start(module, client, instance_ids, wait, timeout):
...
     # If any instance is currently 'stopping' and we'll need to start it, wait for stopped first.
     stopping_now = [iid for (iid, _raw, state) in formatted if state == 'stopping']
+    if stopping_now and not wait:
+        module.fail_json(msg=(
+            "Cannot start instances that are currently 'stopping' with wait=False: "
+            "{}. Either set wait=true (so the module can wait for them to reach "
+            "'stopped' first), or retry once the instances have finished stopping."
+            .format(', '.join(stopping_now))
+        ))
+        return
```
gherceg, line 271: Should the wait parameter only apply to waiting to start? As written, you have to know the state of your instances to safely use `wait=False`.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -297,6 +297,51 @@ def _do_start(module, client, instance_ids, wait, timeout):
...
+                            if state in ('running', 'pending')]
+    # 'stopping' instances are mid-transition; we wait for them but don't initiate.
+    already_stopping = [iid for (iid, _raw, state) in formatted if state == 'stopping']
+    skipped = [iid for iid in instance_ids
+               if iid not in targets_needing_stop and iid not in already_stopping]
+
+    # If any instance is 'pending', wait for it to be 'running' before stopping.
+    pending_now = [iid for (iid, _raw, state) in formatted if state == 'pending']
+    if wait and pending_now:
+        _wait_for(client, 'instance_running', pending_now, timeout, module)
```
gherceg, line 314: Should not wait in check mode.

gherceg, line 314: Also similar to the previous commit/comment, when stopping an instance, should we always wait for pending instances to start regardless of `wait`?

src/commcare_cloud/ansible/library/ec2_instance_state.py
gherceg, line 107: Skipped has a different meaning in Ansible. If  no action is needed on a target, Ansible would consider that "OK". Maybe `unchanged_instance_ids` would be better to avoid conflicting with the Ansible meaning of skipped?

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -223,13 +223,9 @@ def _wait_for(client, waiter_name, instance_ids, timeout, module):
...
-def _emit_result(module, client, instance_ids, before_states, after_states,
-                 state, changed, skipped, refresh):
-    """Build and emit the module's exit_json payload.
-
-    refresh=True: re-describe to compute after_states from live data.
-    refresh=False: use the supplied after_states (used for check_mode and no-op fast paths).
-    """
+def _result_or_emit(module, client, instance_ids, before_states, after_states,
+                    state, changed, skipped, refresh, emit):
+    """Build the result payload. If emit is True, call exit_json; else return the dict."""
```
gherceg, line 228: This change feels like a refactor that should be in a separate commit from implementing restart.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -344,8 +344,57 @@ def _do_stop(module, client, instance_ids, wait, timeout):
...
+    formatted, _ = _describe_and_format(client, instance_ids, module)
+    raw_by_id = {iid: raw for (iid, raw, _state) in formatted}
+
+    instances = [
+        _format_instance(raw_by_id[iid], before_states[iid], after_states[iid])
+        for iid in instance_ids
+    ]
+
+    # 'changed' is True if either phase mutated state.
+    changed = bool(stop_payload['changed'] or start_payload['changed'])
```
gherceg, line 383:
```suggestion
    changed = True
    assert start_payload['changed'], start_payload
```
If doing a restart, there is no way for the start_payload to not have changed if things were successful.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -344,8 +344,57 @@ def _do_stop(module, client, instance_ids, wait, timeout):
...
             _wait_for(client, 'instance_stopped', wait_for, timeout, module)
 
-    _emit_result(module, client, instance_ids, before_states, after_states=None,
-                 state='stopped', changed=changed, skipped=skipped, refresh=True)
+    return _result_or_emit(module, client, instance_ids, before_states, after_states=None,
+                           state='stopped', changed=changed, skipped=skipped,
+                           refresh=True, emit=emit)
+
+
+def _do_restart(module, client, instance_ids, wait, timeout):
```
gherceg, line 352: Should this command support check mode?

tests/test_ec2_instance_state.py
```diff
@@ -59,8 +65,9 @@ def fake_fail(**kwargs):
 
     with patch.object(sys, 'argv', ['ec2_instance_state']), \
          patch.object(basic, '_ANSIBLE_ARGS', encoded_args), \
-         patch.object(ec2_instance_state.AnsibleModule, 'exit_json', side_effect=fake_exit), \
-         patch.object(ec2_instance_state.AnsibleModule, 'fail_json', side_effect=fake_fail):
+         patch.object(ec2_instance_state.AnsibleModule, 'exit_json', new=fake_exit), \
+         patch.object(ec2_instance_state.AnsibleModule, 'fail_json', new=fake_fail), \
```
gherceg, line 69: What was the reason for changing from `side_effect` to `new`?

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -224,14 +224,21 @@ def _wait_for(client, waiter_name, instance_ids, timeout, module):
 
 
 def _result_or_emit(module, client, instance_ids, before_states, after_states,
-                    state, changed, skipped, refresh, emit):
-    """Build the result payload. If emit is True, call exit_json; else return the dict."""
+                    state, changed, skipped, refresh, emit, raw_by_id=None):
```
gherceg, line 227: A lot of parameters are being added to this function and it is getting hard to reason about.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -350,16 +359,17 @@ def _do_stop(module, client, instance_ids, wait, timeout, emit=True):
...
-    if not wait:
+    # Phase 1: stop, always wait (required because StartInstances rejects
+    # still-stopping instances; we must let stop complete before starting).
+    stop_payload = _do_stop(module, client, instance_ids, wait=True,
+                            timeout=timeout, emit=False)
+
+    if not wait and stop_payload['changed']:
         module.warn(
-            "wait=False ignored for the stop phase of 'restarted'; "
+            "wait=False was overridden to wait=True for the stop phase of 'restarted'; "
```
gherceg, line 369: We think it would be simpler to not warn and instead document that `restart` always waits for instances to stop.

tests/test_ec2_instance_state.py
```diff
@@ -584,5 +584,86 @@ def test_restarted_terminated_fails(self):
         self.assertIn('terminated', result['result']['msg'].lower())
 
 
+class TestCheckMode(unittest.TestCase):
```
gherceg, line 587: Should assert that waiter is not invoked when expected too.

tests/test_ec2_instance_state.py
```diff
@@ -584,5 +584,86 @@ def test_restarted_terminated_fails(self):
...
+            {'instance_ids': ['i-0aaaaaaaaaaaaaaaa'],
+             'state': 'stopped', '_ansible_check_mode': True},
+            fake_client=fake,
+        )
+        self.assertFalse(result['failed'])
+        self.assertTrue(result['result']['changed'])
+        self.assertNotIn('stop_instances', [c[0] for c in fake.calls])
+
+
+class TestInvalidIdMutating(unittest.TestCase):
```
gherceg, line 622: Does this need to be a new test class? Should each command have this test?

gherceg (commented):
Paired again with @millerdev. This took a long time to review, partially because we spent time asking questions on earlier commits that were addressed later.

I think we'd be most interested in seeing this module use a cleaner architecture for encapsulating instance state.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Fixed in https://github.com/dimagi/commcare-cloud/pull/6905/commits/036229a25878e0b0eba23733175cfdfc91a72956

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: I thought it would be possible earlier and when I tried to implement it I found out the way ansible works, its not possible to directly reference inventory files inside the modules. The ansible module is shipped to remote host over ssh and run there. A [stackoverflow question](https://stackoverflow.com/questions/32102988/access-playbook-variables-from-custom-module-in-ansible) regarding the same. 

There are ways but not straightforward, so I kept it as it is.

tests/test_ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/commits/6e7ea8fe73fbb0551308ef803c06ede9eb029801

tests/test_ec2_instance_state.py
AmitPhulera: All the tests in the new PR used pytest simple assertions.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: I would like to keep this functionality, I would like us to fail hard if describe fails. As we call this in every command, it would give us immediate feedback on something being wrong.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: This function was removed in the new PR.

tests/test_ec2_instance_state.py
AmitPhulera: Agreed. It is removed.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: The return statements are just for sanity on tests that they actually return. `module.fail_json` actually calls exit  so in production this would never be called. But for tests we mock `fail_json` and if we don't exit there the function will keep running. 
And the reason None was not returned was that this should not give an impression that the function returns something.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Create an Instance class as you suggested - https://github.com/dimagi/commcare-cloud/pull/6905/commits/29991c28d1cec0ae5451505b11964944f6b1f8fb and also a _Ctx class pass around shared context to all the functions - https://github.com/dimagi/commcare-cloud/pull/6905/commits/1cc0e41ca722c4c6098ead643b44dbdf25c39ff3

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: This was already refactored in this PR itself. Apologies for having you review this. I would keep it in mind on how I should structure commits. I tried to do a better job in the newer PR.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/changes/5b264673a5a5fcfcdfabbffe8c2b73106aa304e2#diff-c6913cbc417347f1f21fdf62868aba7c3a544d1561404dc8c4ff1c5581760b6bR302

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/changes/ba1b46a4b75ba417264fe3c94794b0933420c8d9#diff-c6913cbc417347f1f21fdf62868aba7c3a544d1561404dc8c4ff1c5581760b6bR101-R118

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Added a check like https://github.com/dimagi/commcare-cloud/pull/6905/changes/5b264673a5a5fcfcdfabbffe8c2b73106aa304e2 in every command to ensure that no waiting happens in check mode.

tests/test_ec2_instance_state.py
AmitPhulera: Yes. Was missed for this test. It is fixed.

tests/test_ec2_instance_state.py
AmitPhulera: Added a test for this for each command - 

An example - https://github.com/dimagi/commcare-cloud/pull/6905/changes/5b264673a5a5fcfcdfabbffe8c2b73106aa304e2#diff-76c2e039133e62f2ec8bf7a361c858732854bdf3c6e08f5d45dd3184a51618a6R409-R426

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: We are waiting now if the instance is in transitional state - an example - https://github.com/dimagi/commcare-cloud/pull/6905/changes/5b264673a5a5fcfcdfabbffe8c2b73106aa304e2#diff-c6913cbc417347f1f21fdf62868aba7c3a544d1561404dc8c4ff1c5581760b6bR276

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Fixed.

AmitPhulera: Fixed - https://github.com/dimagi/commcare-cloud/pull/6905/changes/ea6ed5fc293418fae72148c7ebdf4f2206e8f35b#diff-c6913cbc417347f1f21fdf62868aba7c3a544d1561404dc8c4ff1c5581760b6bR345-R347

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/changes/ba1b46a4b75ba417264fe3c94794b0933420c8d9#diff-c6913cbc417347f1f21fdf62868aba7c3a544d1561404dc8c4ff1c5581760b6bR90-R93

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: I have added a `fail_json` instead of assert to get a better output if this happens - https://github.com/dimagi/commcare-cloud/pull/6905/changes/036229a25878e0b0eba23733175cfdfc91a72956#diff-c6913cbc417347f1f21fdf62868aba7c3a544d1561404dc8c4ff1c5581760b6bR394

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Fixed.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Have an instance class now in the new PR and this situation does not exist.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Addressed in https://github.com/dimagi/commcare-cloud/pull/6905/changes/036229a25878e0b0eba23733175cfdfc91a72956

tests/test_ec2_instance_state.py
AmitPhulera: The tests in new PR address it.

tests/test_ec2_instance_state.py
AmitPhulera: This is now being tested in - https://github.com/dimagi/commcare-cloud/pull/6905/changes/6e7ea8fe73fbb0551308ef803c06ede9eb029801

AmitPhulera:
@millerdev,  @gherceg  It took me a while to come back to this PR and address the comments, really sorry about that.

While addressing the comments, the PR took a different shape than it was before and since a lot of time has passed since this was reviewed, I felt it would be good to have a new PR rather than adding commits here. So I created https://github.com/dimagi/commcare-cloud/pull/6905 which addresses the suggestions given here. I have also responded to each suggestion in the PR to close the loop here. 
Let me know if you need any more information on it. I am closing this PR now.

PR #6859, opened 2026-04-21
Fetch code_version explicitly during deploy to fix git checkout race

dannyroberts:
Fixes a race during staging deploys where `git checkout <code_version>` fails with `fatal: unable to read tree (<sha>)` even though the SHA exists on GitHub. Happens when an autostaging branch is rebuilt (via \`rebuildstaging\`) between the moment \`code_version\` is resolved from the branch tip (GitHub API, \`commcare.py:154\`) and the moment the target host's mirror runs \`git remote update --prune\`. The old tip is no longer reachable from any ref, so it's never fetched, and the subsequent \`git clone --reference\` can't supply it either.

The fix: after the mirror update, check whether the requested commit is present; if not, fetch it explicitly by SHA. GitHub's uploadpack accepts a fetch-by-SHA for commits still in the object db after a force-push.

Split into two commits:
1. Prefactor `Release.run` to return a `CommandResult` namedtuple so callers can read `.rc` / `.stdout` / `.stderr` at the call site — no behavior change.
2. Add `has_commit` guard + explicit `git fetch --no-tags origin <version>` after the mirror update. Test update for `test_bad_version`: the failure mode shifts from "pathspec did not match" (at checkout) to "couldn't find remote ref bad-version" (at fetch), which is strictly more informative for debugging.

##### Environments Affected

Affects deploy on all environments, but invisibly so other than to fix a race condition. Does not change any actual environment setup.

### DIFF
--- a/tests/test_deploy/test_git_setup_release.py
+++ b/tests/test_deploy/test_git_setup_release.py
@@ -181,7 +181,7 @@ class TestGitSetupRelease(TestCase):
     def test_bad_version(self):
         release = self.releases / "one"
         refs = self.releases / "refs"
-        with self.assertRaisesRegex(ansible.Fail, f"error: pathspec 'bad-version' did not match"):
+        with self.assertRaisesRegex(ansible.Fail, "couldn't find remote ref bad-version"):
             ansible.run("git_setup_release", {
                 "repo": self.src_repo,
                 "version": "bad-version",
--- a/src/commcare_cloud/ansible/library/git_setup_release.py
+++ b/src/commcare_cloud/ansible/library/git_setup_release.py
@@ -176,6 +176,13 @@ class Release:
         else:
             self.run(["git", "remote", "update", "--prune"], cwd=repo_mirror)
 
+        # Ensure the requested commit is present even if its branch moved
+        # mid-deploy (e.g. autostaging rebuilt between SHA resolution and
+        # the mirror update), which would leave the SHA unreachable from
+        # any ref and absent after `remote update`.
+        if not self.has_commit(repo_mirror, version):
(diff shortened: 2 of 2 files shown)


tests/test_deploy/test_git_setup_release.py
```diff
@@ -181,7 +181,7 @@ def test_incomplete_release(self):
     def test_bad_version(self):
         release = self.releases / "one"
         refs = self.releases / "refs"
-        with self.assertRaisesRegex(ansible.Fail, f"error: pathspec 'bad-version' did not match"):
+        with self.assertRaisesRegex(ansible.Fail, "couldn't find remote ref bad-version"):
```
millerdev, line 184: Did this change because of the difference between `git clone` and `git fetch` output when given a bad ref? If not, can you explain why it changed?

millerdev (approved):
Thanks, this looks like it should resolve the issue I was seeing earlier.

dannyroberts:
Testing update: I tested that this doesn't cause any problems for the happy path of deploy, which would be the main failure mode. (Worse than if the fix is harmless but doesn't work for the edge case.)

tests/test_deploy/test_git_setup_release.py
dannyroberts: You got it right. I had the choice of letting the extra specific-commit fetch from origin fail silently so that it fails later with the same error message, or failing faster right then with a different error message. I chose to have it fail quickly with the new error message, because it seemed better on both fronts: the error comes from the earliest place where we know the deploy will fail, making debugging easier, and the new error message was actually a slightly more informative.

PR #6865, opened 2026-04-28
Push deploy tags via SSH instead of GitHub API

dannyroberts:
https://dimagi.atlassian.net/browse/SAAS-19625
## Summary
- Replaces the GitHub-API `create_git_ref` call in `create_release_tag` with a native `git push` over SSH that uses the forwarded SSH agent on the control machine
- Pins github.com's SSH host keys in a bundled `github_known_hosts` file (with a `manage-commcare-cloud refresh-github-known-hosts` subcommand for rotation) so the first SSH push from a new control machine doesn't hit an interactive TOFU prompt
- Drops the write-scope PAT machinery from `github.py`: `require_write_permissions`, the `repo.permissions.push` check, `GithubException`, and the long-form prompt function. Tag creation no longer needs a write-scope token.
- Replaces `require_write_permissions` with a new `prompt_if_missing` kwarg on `github_repo()`. Deploy callers gate it on `fab_settings_config.generate_deploy_diffs`, so production deploys still prompt for a (read-only) token to populate the deploy diff with PR details, while staging (which doesn't generate diffs) doesn't prompt.
- Modernizes `github.py`: `@memoized` on the credential lookup, legacy-location warning extracted into a memoized helper that fires only when `fab/config.py` defines a truthy `GITHUB_APIKEY`, tuple return simplified to a plain token.
- Refreshes `config.example.py` to describe the token as an optional read-only rate-limit lever (no longer required for tagging, no longer prompted each time).

## Test plan

Manually tested:
- [x] Do a staging deploy to confirm a tag lands on GitHub without a PAT prompt

## Side notes
- Test stdout shows git progress lines from the real-git integration test (cosmetic CI noise; can be silenced by adding `stderr=subprocess.DEVNULL` to the helper if it becomes annoying)


src/commcare_cloud/github.py
```diff
@@ -1,66 +1,16 @@
...
-    global GITHUB_TOKEN
-
-    token, found_in_legacy_location = get_github_credentials_no_prompt()
-
-    if found_in_legacy_location:
-        print(color_notice(f"[Deprecation Warning] Config file has moved."))
-        print(color_notice(f"New location is {PROJECT_ROOT}/config.py or else use the "
-                           f"'GITHUB_TOKEN' environment variable."))
-        print(color_notice(f"\nYou can move the config to the new location as follows:"))
-        print(color_notice(f"    $ mv {PROJECT_ROOT}/fab/config.py {PROJECT_ROOT}/config.py\n"))
```
dannyroberts: This message is restored in a later commit, pushed into `get_github_credentials_no_prompt`.

dannyroberts:
Just one review tip, there's one commit where there are some indentation changes, so it'll look a lot better if you add `?w=1` to the URL: https://github.com/dimagi/commcare-cloud/pull/6865/changes/b3561ce9a695382e1391a9201a921104e593efc2?w=1

dannyroberts:
As an aside, the original commit chunking and sequencing was a mess and it took me quite a while to get the changes broken up into these nice consumable chunks with a (hopefully!) clear narrative. Feel free to share feedback in your review about how it was to review—I'm trying but I want to know how it's landing!

src/commcare_cloud/manage_commcare_cloud/refresh_github_known_hosts.py
```diff
@@ -0,0 +1,38 @@
...
+class RefreshGithubKnownHosts(CommandBase):
+    command = 'refresh-github-known-hosts'
+    help = (
+        "Refetch github.com's SSH host keys via ssh-keyscan and write them "
+        "to the bundled known_hosts file used for deploy tag pushes. "
+        "Run this if GitHub rotates its host keys."
+    )
+
+    def make_parser(self):
+        pass
```
nospame: What is this for?

nospame (commented):
The code all looks good here, but I feel like I might not understand enough about how this actually works to approve it. Specifically, why do we need to create a temporary repo to push a tag? Is it just that the tag just needs to be created in _a_ repo, and then once created can be pushed to _any_ repo there is access to?

src/commcare_cloud/commands/deploy/utils.py
```diff
@@ -36,16 +40,40 @@ def get_meta_value(self, key):
...
+            f"-o StrictHostKeyChecking=yes"
+        ),
+    }
+    try:
+        subprocess.run(
+            ["git", "-C", tmp, "init", "--bare", "-q"],
+            check=True,
+        )
+        subprocess.run(
+            ["git", "-C", tmp, "fetch", "--depth=1", "--no-tags", remote_url, sha],
```
millerdev: Can these two be replaced with
```sh
git clone --bare --filter=blob:none --depth=1 --revision={sha} {remote_url} {tmp}
```
In addition to reducing the number of commands, it also does not fetch file content. My local test resulted in a 624K clone for a recent HQ commit.

src/commcare_cloud/commands/deploy/utils.py
```diff
@@ -36,16 +40,40 @@ def get_meta_value(self, key):
...
+            check=True, env=env,
+        )
+        subprocess.run(
+            ["git", "-C", tmp, "push", remote_url, f"{sha}:refs/tags/{tag_name}"],
+            check=True, env=env,
+        )
```
millerdev: nit: What do you think of using the `sh` library, which has a cleaner API for running subprocesses? We use it in other places.

src/commcare_cloud/manage_commcare_cloud/refresh_github_known_hosts.py
```diff
@@ -0,0 +1,38 @@
+import subprocess
+
+from commcare_cloud.commands.command_base import CommandBase
+from commcare_cloud.github import GITHUB_KNOWN_HOSTS
+
+
+class RefreshGithubKnownHosts(CommandBase):
+    command = 'refresh-github-known-hosts'
```
millerdev, line 14: Can we (possibly prompt to) run this automatically on deploy if the github hosts file is missing?

src/commcare_cloud/github.py
```diff
@@ -45,8 +34,6 @@ def get_github_credentials(repo_name, require_write_permissions):
     if token is None:
         print(color_warning("Github credentials not found!"))
         print(f"Github token is required for repository {repo_name}.")
-        if require_write_permissions:
-            print("The token must have write permissions to the repository to create release tags.")
```
millerdev, line 49: Should we tell devs to generate new tokens that do not have write permissions now that it is no longer required for deploy?

millerdev (commented):
From b61a8e65e48b9aa44e071c32312af961aed7f2c1
> the prompt is still useful: a token unlocks PR details in the deploy diff and avoids the unauthenticated rate limit.

Does this mean we still need to supply a PAT when deploying environments other than staging? Would it make sense to use the github API to create tags if a token is present since that seems far more efficient and probably faster than creating a local clone?

Edit: clicked submit too early. This review is incomplete...

src/commcare_cloud/config.example.py
```diff
@@ -1,8 +1,6 @@
-# To generate a Github apikey follow these instructions: https://github.com/blog/1509-personal-api-tokens
-# Minimally you will need a key with repo -> public_repo
-
-# On a control machine shared with others,
-# leave this as None to be prompted for a token each time instead,
-# to avoid storing private github token where others could access it.
+# Optional GitHub API token used to get a higher rate limit on read-only
+# diff/compare calls during deploy. The GITHUB_TOKEN environment variable
+# is also honored. A token is no longer required for deploy tagging,
+# which is now done via SSH using the user's forwarded agent credentials.
```
millerdev, line 4: Is it still recommended to use a token? If yes, the link to instructions (removed) still seems useful.

src/commcare_cloud/manage_commcare_cloud/refresh_github_known_hosts.py
millerdev: Additionally, if github ever changes its keys I'm guessing the error message will not point to this command as the correct course of action. Maybe the known hosts file that this command writes could include a comment that it was generated by this command?

src/commcare_cloud/github.py
```diff
@@ -3,87 +3,72 @@
...
-    return token or None
-
-
-def get_github_credentials_no_prompt():
-    """
-    :return: tuple(token, found_in_legacy_location)
+    The optional token authenticates API calls (5,000/hr instead of 60/hr
+    unauthenticated, and unlocks repo permissions metadata used by deploy
+    diff). When ``prompt_if_missing`` is True and no token is found via
+    env or config, ask interactively. Empty input continues unauthenticated.
```
millerdev: nit: This docstring reads very AI. Suggest to rephrase

- "5,000/hr instead of 60/hr unauthenticated" -> "higher rate limit"
- "ask interactively. Empty input continues unauthenticated." -> "ask for token and optionally continue without authentication.

millerdev (commented):
As noted in my earlier review, it is unclear when a PAT is required (if ever). Is it fine to run a production deploy without one? Sounds like some details will be missing if that is done, but not quite sure what that means... less informative Slack messages?

src/commcare_cloud/config.example.py
dannyroberts: It's only still recommended to use a token when `generate_deploy_diffs` is set to `True`. On staging it's set to `False`.

This is because this is the only API get intensive part of the deploy that could conceivably be rate limited. Staging doesn't benefit from the diffs anyway, so it has long since been disabled there.

However, I think we also want to de-emphasize writing the github token to a local file, since we don't encourage that practice on our team, where "local" would mean on the shared control server. I'm keeping it mostly for third parties in case they are already set up this way, and to not have to think through whether it's worth deprecation notices and such.

src/commcare_cloud/manage_commcare_cloud/refresh_github_known_hosts.py
dannyroberts: > Can we (possibly prompt to) run this automatically on deploy if the github hosts file is missing?

I think it should never be missing, because it's bundled with the code base (at `src/commcare_cloud/github_known_hosts`). Let me know if I misunderstood the question.

> Maybe the known hosts file that this command writes could include a comment that it was generated by this command?

This is not a bad idera. This happens rarely: I believe they're quite stable month to month (published [here](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints) as per the message below), but they can and do change at a years timescale. I'll see if there's an easy way for me to surface that proactively. I think if it's a lot of overhead it's probably not worth it, but otherwise it's quite nice.

src/commcare_cloud/github.py
dannyroberts: Yes thanks, I'll share on our team. The prompt output does include a reminder that read-only is fine, but I'll proactively share the guidance that we'd now prefer people to revisit those permissions and use read-only.

src/commcare_cloud/commands/deploy/utils.py
dannyroberts: Good idea. The exact form you proposed requires --revision, which [was added in git 2.49](https://github.blog/open-source/git/highlights-from-git-2-49/), about a year ago. I checked and not all of our control machines are on that version. *But* it seems that `--filter=blob:none` is where the real speedup is and there's no similar version issue with that. Updated in 0b8f96208.

Here's the timing result locally (I'm taking Claude Code's word for it, but seems plausible):
 
> |        | size | time |
> |--------|------|------|
> | before | 22M  | 11s  |
> | after  | 552K | 3s   |

I think I'm okay with 3-5 seconds instead of ~1 second or whatever with the API.

src/commcare_cloud/commands/deploy/utils.py
dannyroberts: That would be a tad nicer, I agree. When I looked into it (responding to your feedback here), I saw that it's currently a test-dep only when it comes to the `commcare-cloud` CLI itself. (It's used in an ansible module we wrote, so it gets installed in our ansible requirements.) For the `commcare-cloud` CLI, it's currently set under `test_deps`.

I think it would be fine to promote to `install_deps` (especially since it's already broadly a production dependency of the CLI via the ansible requirements), but it does add a small wrinkle that makes me pause.

I went ahead with it, but just wanted to validate that thought process with you. Happy to take it or leave it. 6b1dafd47.

dannyroberts:
Thanks for review. Just noting that I'll make some changes and make the rest of my replies in "review comments".

src/commcare_cloud/config.example.py
dannyroberts: And regarding
> Would it make sense to use the github API to create tags if a token is present since that seems far more efficient and probably faster than creating a local clone?

I think it wouldn't be worth it. It's nice to not have to require write permissions, even when we need a token. Adding the branching and the two ways of implementing, and also the branching on whether write permissions are present or not seems like added complexity without all that much value. It's true that it's faster, but I think we can easily get this fast enough the one way that we do it that it's not a meaningful part of the deploy time.

src/commcare_cloud/manage_commcare_cloud/refresh_github_known_hosts.py
dannyroberts: Oh, good catch. This was copied from other similar commands that have this (probably for vestigial historical reasons), but I confirmed it's unnecessary and doesn't do anything. Removed: 013856bc5

src/commcare_cloud/github.py
dannyroberts: Fair enough. 2a220b67f

dannyroberts:
@nospame
> Why do we need to create a temporary repo to push a tag? Is it just that the tag just needs to be created in a repo, and then once created can be pushed to _any_ repo?

Yeah almost. The local repo needs to have local knowledge of the commit we're trying to tag. If we did just the init part and tried to push to the remote it would fail with something like

> fatal: bad object <SHA>

based on my testing. That's why we have to do the `fetch` command I have here. (Daniel's suggestion would be a further optimization of doing the init + fetch in a single `clone` command that grabs just the commit we care about, but that's only available in the most recent version of git, so not available everywhere we'd run this.)

src/commcare_cloud/config.example.py
dannyroberts: And to more fully address

> As noted in my earlier review, it is unclear when a PAT is required (if ever). Is it fine to run a production deploy without one? Sounds like some details will be missing if that is done, but not quite sure what that means... less informative Slack messages?

If a PAT is missing when it's expected (when `generate_deploy_diffs` is set to `True`), even after prompting, you'll get this warning message https://github.com/dimagi/commcare-cloud/blob/bf29f0c65aa44a1a78425f110b624416a2a2f7e6/src/commcare_cloud/commands/deploy/deploy_diff.py#L143-L147
and the deploy email  and slack message won't have that richer diff information that says what PRs are in the deploy, etc.

src/commcare_cloud/manage_commcare_cloud/refresh_github_known_hosts.py
dannyroberts: Okay added a comment at the top of the generated `github_known_hosts` file. I briefly looked at making the error itself get surfaced at an earlier place, but for something that gets exercised every few years, it seemed more likely that the code around raising it proactively in real time was pretty likely to silently break before it ever gets exercised, so the cost of maintaining that code path didn't seem worth it. fc3109e1d

dannyroberts:
Confirming that I reteseted with another deploy, so this is ready for another pass at review

src/commcare_cloud/commands/deploy/utils.py
millerdev: I didn't realize we only use it in testing in commcare-cloud. I guess we use it for non-test things in HQ, which was why I thought it would work here. I like the change, and am fine with keeping it.

src/commcare_cloud/commands/deploy/utils.py
millerdev: > The exact form you proposed requires --revision, which [was added in git 2.49](https://github.blog/open-source/git/highlights-from-git-2-49/), about a year ago. I checked and not all of our control machines are on that version.

We specify what software gets installed on our control machines. Why not upgrade git and then use the simpler command? I think it could be done by [adding a version specifier](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/apt_module.html#parameter-name) like `git>=2.49` on this line:

https://github.com/dimagi/commcare-cloud/blob/77c6774150e1d6b4bb886f978a3fb913b3b60be4/src/commcare_cloud/ansible/roles/git/tasks/main.yml#L13

The down-side is this PR would require a maintenance command to be run on control machines, but maybe that's worth it in the long run since it simplifies usage and makes our control machines more alike. Fine to do that in a separate PR if you like the idea.

millerdev (approved):

src/commcare_cloud/commands/deploy/utils.py
dannyroberts: Yeah I went through basically this same chain of thought, and it felt like a lot to hitch this change onto. I think even putting the specific `--revision` thing aside, it's probably good for us to set a minimum version here and bump machines up, so I can just go ahead and make that PR. If we want to actually rely on it though (as with this `--revision` flag), we need to add a changelog entry and wait six weeks. So working backwards, I think what I'll do is make a PR with just that minimum git version and a changelog entry, merge that, and announce it, and make a ticket to roll that out that also says once it's rolled out make this change.

PR #6867, opened 2026-05-05
Skip staging deploy prompt and post diff URL to Slack thread

millerdev:
Reduce friction when deploying to staging:

- Staging deploys no longer pause to ask for confirmation when there is new code to deploy.
- The bot now posts a link to the GitHub compare URL in the deploy thread so lurkers can see what's being deployed.

Tested by running a staging deploy.

:blowfish: Review by commit.

##### Environments Affected

staging

### DIFF
--- a/src/commcare_cloud/commands/deploy/slack.py
+++ b/src/commcare_cloud/commands/deploy/slack.py
@@ -81,6 +81,12 @@ class SlackClient:
                     "text": {"type": "mrkdwn", "text": chunked_diff_text}
                 }]
                 self._post_message("Deploy diff", diff_blocks, thread_ts=response["ts"])
+        else:
+            diff_blocks = [{
+                "type": "section",
+                "text": {"type": "mrkdwn", "text": "Complete diff: " + context.diff.url}
+            }]
+            self._post_message("Deploy diff", diff_blocks, thread_ts=response["ts"])
 
     def _chunk_diff(self, diff_text, chunk_size=2800):
         # Slack has a limit of 3000 characters per message block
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/commands/deploy/slack.py
```diff
@@ -81,6 +81,12 @@ def send_deploy_start_message(self, context):
                     "text": {"type": "mrkdwn", "text": chunked_diff_text}
                 }]
                 self._post_message("Deploy diff", diff_blocks, thread_ts=response["ts"])
+        else:
+            diff_blocks = [{
+                "type": "section",
+                "text": {"type": "mrkdwn", "text": "Complete diff: " + context.diff.url}
+            }]
+            self._post_message("Deploy diff", diff_blocks, thread_ts=response["ts"])
```
gherceg, line 89: This isn't really useful on staging right? My understanding is that it isn't a diff from one deploy to the next, but rather from master to autostaging?

gherceg: Oh though looking at the example in the slack channel, maybe it is useful? Idk, I'll wait to hear from you but otherwise no issues with this!

millerdev: Snarky response: if it's not useful, why do we [print it in the console](https://github.com/dimagi/commcare-cloud/blob/e34fa64c867a6d0a1a57ccb6832b05861bcb4912/src/commcare_cloud/commands/deploy/diff_templates/console.txt.j2#L68) when running a deploy?

Real response: I think it is useful to be able to quickly answer the question "Did change X get deployed in this run?"

It's true, change X could have also gone out in an earlier deploy, so it does not answer the question "When did change X first get deployed?"

gherceg (approved):

PR #6868, opened 2026-05-05
Require git >= 2.49 on managed machines

dannyroberts:
Followup to https://github.com/dimagi/commcare-cloud/pull/6865#discussion_r3189395222. Rollout tracked in https://dimagi.atlassian.net/browse/SAAS-19703.

##### Environments Affected

All

##### Announce New Release
- [x] (Dimagi only) After merging, I will follow [these instructions](https://confluence.dimagi.com/display/saas/Announcing+changes+affecting+third+parties#Announcingchangesaffectingthirdparties-announcing) to announce a new commcare-cloud release.

## Summary
- Pin a minimum git version (`>=1:2.49.0`) on the `install_git` task so commcare-cloud can rely on recent git features without per-host probing.
- Add a changelog entry directing environments to run `cchq <env> deploy-stack --tags=install_git` to bring all hosts up to the new minimum.

This is the prerequisite I committed to in https://github.com/dimagi/commcare-cloud/pull/6865#discussion_r3189395222 — once this is merged, announced, and rolled out (per our usual six-week window), we can switch the deploy-tag push in #6865 to the simpler `git clone --bare --filter=blob:none --depth=1 --revision={sha} ...` form @millerdev originally proposed.

## Why `1:2.49.0` and not `2.49.0`
The `1:` here is the "epoch", which in debian packaging, starts at / defaults to 0, and then can be changed to 1 (and then 2 and so on) as a hard-reset when versioning schemes change, in order to preserve ordering. For historical reasons, the `ppa:git-core/ppa` PPA (already configured by the role) uses epoch 1, e.g. `1:2.49.0-2~ppa1~ubuntu22.04.1`. Without the explicit `1:`, dpkg would compare against epoch 0, which would make `>=` falsely match all versions starting with epoch `1:` (basically, all versions). (I think `1:` is used to match the default non-ppa `git`, which itself uses `1:` because the `git` name used to refer to a completely different package, GNU Interactive Tools, and they had to reset version numbers which they decided to give that name to the eventually much more popular `git` version control system it refers to today; but I kind of just pieced this together from scraps, so this is not the authoritative story.)

## Test plan
(with `--branch=dmr/min-git-version` since I'll test on this branch):
- [x] Roll out on staging first: `cchq staging deploy-stack --tags=install_git`
- [x] Verify the applied diff shows only this one install command, or at least no unrelated/unnecessary commands—we don't want this accidentally rolling other stuff out
- [x] Confirm `cchq staging run-shell-command all 'git --version'` reports `2.49.x` or later on every host


millerdev (approved):

dannyroberts:
Ansible's `apt` apparently does not support this version constraint syntax. Taking back to draft—apologies for requesting review prematurely!

src/commcare_cloud/ansible/roles/git/tasks/main.yml
```diff
@@ -14,3 +14,23 @@
...
+  register: git_version_check
+  changed_when: false
+  check_mode: false
+  tags:
+    - install_git
+
+- name: Assert minimum git version
+  assert:
+    that:
+      - (git_version_check.stdout | regex_search('\d+\.\d+\.\d+')) is version('2.49.0', '>=')
```
millerdev, line 29: The "Install git packages" `apt` task above does not do an upgrade if git is already installed. Given a machine with a lower version of git, will this always fail until/unless manual intervention is taken to do something like `apt upgrade git`?

changelog/0097-min-git-version.yml
```diff
@@ -0,0 +1,34 @@
...
+  releases without having to first check what's installed on each machine.
+  The `ppa:git-core/ppa` repository (already configured by this role) ships
+  versions well past 2.49 for all supported Ubuntu releases, so the upgrade
+  is just a matter of running the install_git tag.
+
+update_steps: |
+  Update git on all machines in your environment:
+
+  ```bash
+  cchq <env> deploy-stack --tags=install_git
```
millerdev, line 25: Since 62f944a56260eb87994904792f7c6ba4006c1eab, I think this will not upgrade git anymore. Now it only checks the version and fails if it's not at least 2.49.

millerdev (changes requested):
> Ansible's `apt` apparently does not support this version constraint syntax.

Is that because we are using an old version of Ansible, or it just doesn't work at all despite what the documentation says?

If it's an Ansible version thing, let's hold off on this PR until I have upgraded Ansible (coming up soon).

src/commcare_cloud/ansible/roles/git/tasks/main.yml
dannyroberts: My understanding is that [`state: latest`](https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/apt_module.html#parameter-state), which the "Install git packages" task is using, will install the latest available version, even if a prior version is already installed. Paired with updating the apt cache ("Update package list" right above it) I think this should reliably upgrade git to the latest version. (`state: latest` was already what it was set to, so this PR just adds a sanity check; it doesn't actually change the install version behavior.)

dannyroberts:
It might be because it's an older version of ansible. I do see that the docs since 2.13 (the earliest I could find for this page) through the latest version all advertise support for `>=` syntax: https://docs.ansible.com/projects/ansible-core/2.13/collections/ansible/builtin/apt_module.html .. https://docs.ansible.com/projects/ansible-core/2.20/collections/ansible/builtin/apt_module.html

src/commcare_cloud/ansible/roles/git/tasks/main.yml
millerdev: Oh, thanks, I missed that `state: latest` was already there, and assumed that it didn't work that way given how old the version of git is on some machines. Guess we just have not run `deploy-stack` on all of our machines very recently.

Forgive me if you have already said so, but have you tested to verify that it works that way on a machine with an older version of git installed?

dannyroberts: My impression when I was rolling this out on staging is that it did install newer versions of git—they said "changed"—but it was not a specific check I did beforehand to see what versions they used to be on like it probably should have been. Thankfully, we log installs, and I can see in the CloudWatch logs from the last 24 hours lines like

```
Upgrade: git:amd64 (1:2.49.0-2~ppa1~ubuntu22.04.1, 1:2.54.0-0ppa1~ubuntu22.04.1), git-man:amd64 (1:2.49.0-2~ppa1~ubuntu22.04.1, 1:2.54.0-0ppa1~ubuntu22.04.1)
```

which looks to me like upgrading git from `1:2.49.0` to `1:2.54.0`, so I think that's the confirmation we were looking for.

millerdev (approved):

PR #6871, opened 2026-05-07
Support non-interactive cchq secrets edit

dannyroberts:
https://dimagi.atlassian.net/browse/SAAS-19725 (subtask of https://dimagi.atlassian.net/browse/SAAS-19724)

##### Environments Affected
None — CLI ergonomics only, no environment files change.

## Summary
- Add a `--from-stdin` flag to `cchq <env> secrets edit` so the new value can be piped in instead of typed at a `getpass` prompt:
  ```
  echo "$MY_SECRET" | cchq <env> secrets edit MY_SECRET --from-stdin
  ```
- A single trailing newline is stripped so the common `echo … | …` pattern works without the newline ending up in the stored value. JSON parsing of the value (existing behavior) is preserved.
- `--from-stdin` is rejected with an argparse usage error (exit 2) if paired with any subcommand other than `edit` (`view`, `list-append`, `list-remove`), rather than being silently ignored.

## Test plan
- [x] `echo "hello" | cchq staging secrets edit MY_TEST_SECRET --from-stdin` followed by `cchq staging secrets view MY_TEST_SECRET` returns `hello` (no trailing newline).
- [x] `cchq staging secrets edit MY_TEST_SECRET` without the flag still prompts interactively and works as before.
- [x] `cchq staging secrets edit --help` lists the new option.
- [x] `cchq <env> secrets view FOO --from-stdin` exits non-zero with an argparse usage error mentioning that `--from-stdin` only works with `edit`.

### DIFF
--- a/src/commcare_cloud/commands/secrets.py
+++ b/src/commcare_cloud/commands/secrets.py
@@ -49,9 +49,7 @@ class Secrets(CommandBase):
 
     def _secrets_edit(self, environment, secret_name, from_stdin=False):
         if from_stdin:
-            secret_value = sys.stdin.read()
-            if secret_value.endswith('\n'):
-                secret_value = secret_value[:-1]
+            secret_value = sys.stdin.read().strip()
         else:
             environment.secrets_backend.prompt_user_input()
             secret_value = getpass.getpass(f"New value for '{environment.name}' secret '{secret_name}': ")
--- a/src/commcare_cloud/commands/secrets.py
+++ b/src/commcare_cloud/commands/secrets.py
@@ -1,5 +1,6 @@
 import getpass
 import json
+import sys
 
 import yaml
 from clint.textui import puts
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/commands/secrets.py
```diff
@@ -40,9 +47,14 @@ def _secrets_view(self, environment, secret_name):
...
             print(yaml.safe_dump(secret))
 
-    def _secrets_edit(self, environment, secret_name):
-        environment.secrets_backend.prompt_user_input()
-        secret_value = getpass.getpass("New value for '{}' secret '{}': ".format(environment.name, secret_name))
+    def _secrets_edit(self, environment, secret_name, from_stdin=False):
+        if from_stdin:
+            secret_value = sys.stdin.read()
+            if secret_value.endswith('\n'):
+                secret_value = secret_value[:-1]
```
gherceg: I suppose this is sufficient and therefore no change is needed, but it also seems like whitespace shouldn't be part of any secrets, and using `strip()` would suffice? Though maybe keeping this flexible is best?

src/commcare_cloud/commands/secrets.py
```diff
@@ -54,7 +54,9 @@ def _secrets_edit(self, environment, secret_name, from_stdin=False):
                 secret_value = secret_value[:-1]
         else:
             environment.secrets_backend.prompt_user_input()
-            secret_value = getpass.getpass("New value for '{}' secret '{}': ".format(environment.name, secret_name))
+            secret_value = getpass.getpass(
+                "New value for '{}' secret '{}': ".format(environment.name, secret_name)
+            )
```
gherceg: nit: while you're here, use f-string which also keeps the line length to under 115?
```suggestion
            secret_value = getpass.getpass(f"New value for '{environment.name}' secret '{secret_name}':")
```

gherceg (approved):
Nothing blocking. Looks good.

src/commcare_cloud/commands/secrets.py
dannyroberts: Good call! Since I was just moving the existing line, I hadn't actually looked that carefully at it, but this is better. Thanks!

src/commcare_cloud/commands/secrets.py
dannyroberts: That's fair. I initially was thinking, do the minimal thing, but it's actually probably more surprising for a command that strips a newline from the end of the input _not_ to then also strip _all_ whitespace from the _start and end_. So I agree, I'll update.

gherceg: Yeah I should have been clear about "leading and trailing whitespace shouldn't be part of any secrets". Thanks!

dannyroberts:
@gherceg Addressed both comments, ready for re-review.

gherceg (approved):
Thanks!

PR #6872, opened 2026-05-07
Add tooling to manage AWS IAM access keys

dannyroberts:
[SAAS-19726](https://dimagi.atlassian.net/browse/SAAS-19726)

- Adds `scripts/aws/manage-iam-keys.py` for inspecting and rotating IAM access keys, building blocks for SES SMTP credential rotation
- Subcommands: `list` (JSON), `create` (CSV in the AWS console download format consumed by `derive_ses_smtp_password.py`), `deactivate`, `reactivate`, `remove`
- `remove` refuses without `--force` unless the key is `Inactive` and either never used or last used 24h+ ago

**Note** for this to work you need an aws profile (e.g. in your `~/.aws/config`) that has access to the IAM keys you're interested in—it'll need to have the correct account number and role.

##### Environments Affected
None — this adds a new standalone script under `scripts/aws/` and does not modify any environment files.

## Test plan
- [x] `./scripts/aws/manage-iam-keys.py ses-iam ses-staging list` outputs a JSON array with `AccessKeyId`, `Status`, `CreateDate`, `LastUsedDate`, `LastUsedService`, `LastUsedRegion`
- [x] `./scripts/aws/manage-iam-keys.py ses-iam ses-swiss create` returns CSV that `derive_ses_smtp_password.py` can consume (extract field 2 as the secret)
- [x] `./scripts/aws/manage-iam-keys.py ses-iam ses-swiss list` shows a key
- [x] `./scripts/aws/manage-iam-keys.py ses-iam ses-swiss remove <key-id> --force` succeeds, even while it's active. (Recreate the key, this time with `> /dev/null` since we don't need to see the output, for the next tests.)
- [x] `./scripts/aws/manage-iam-keys.py ses-iam ses-swiss remove <key-id>` refuses while the key is `Active`
- [x] `./scripts/aws/manage-iam-keys.py ses-iam ses-swiss deactivate <key-id>` flips status to `Inactive`; `reactivate <key-id>` flips it back to `Active` (the deactivate again for the next test)
- [x] `./scripts/aws/manage-iam-keys.py ses-iam ses-swiss remove <key-id>` succeeds for an `Inactive` key that is never-used or last used 24h+ ago

[SAAS-19726]: https://dimagi.atlassian.net/browse/SAAS-19726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/aws/manage-iam-keys.py
+++ b/scripts/aws/manage-iam-keys.py
@@ -6,7 +6,9 @@ Usage:
     ./scripts/aws/manage-iam-keys.py <aws-profile> <iam-user> <subcommand>
 
 Subcommands:
-    list    Output a JSON list of the user's access keys, including last-used info.
+    list                    Output a JSON list of the user's access keys, including last-used info.
+    deactivate <key-id>     Set the key status to Inactive and print the updated key info.
+    reactivate <key-id>     Set the key status to Active and print the updated key info.
 """
 
 import argparse
@@ -25,6 +27,16 @@ def main():
 
     subparsers.add_parser('list', help="List the user's access keys as JSON.")
 
+    deactivate_parser = subparsers.add_parser(
+        'deactivate', help='Set the given access key to Inactive.'
+    )
+    deactivate_parser.add_argument('key_id', help='Access key ID to deactivate.')
+
+    reactivate_parser = subparsers.add_parser(
+        'reactivate', help='Set the given access key to Active.'
+    )
+    reactivate_parser.add_argument('key_id', help='Access key ID to reactivate.')
+
     args = parser.parse_args()
 
     session = boto3.session.Session(profile_name=args.aws_profile)
@@ -32,6 +44,10 @@ def main():
 
     if args.subcommand == 'list':
         print_json(list_keys(iam, args.iam_user))
+    elif args.subcommand == 'deactivate':
+        print_json(set_key_active(iam, args.iam_user, args.key_id, active=False))
+    elif args.subcommand == 'reactivate':
+        print_json(set_key_active(iam, args.iam_user, args.key_id, active=True))
 
 
 def list_keys(iam, iam_user):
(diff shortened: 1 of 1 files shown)


scripts/aws/manage-iam-keys.py
```diff
@@ -0,0 +1,72 @@
...
+        print_json(list_keys(iam, args.iam_user))
+
+
+def list_keys(iam, iam_user):
+    keys = iam.list_access_keys(UserName=iam_user)['AccessKeyMetadata']
+    result = []
+    for key in keys:
+        last_used = iam.get_access_key_last_used(
+            AccessKeyId=key['AccessKeyId']
+        )['AccessKeyLastUsed']
```
millerdev, line 43: Could this hit rate limits? Is there a way to do a bulk query for last used times for a set of keys?

scripts/aws/manage-iam-keys.py
```diff
@@ -0,0 +1,157 @@
...
+    if args.subcommand == 'list':
+        print_json(list_keys(iam, args.iam_user))
+    elif args.subcommand == 'create':
+        print_keyfile(create_key(iam, args.iam_user))
+    elif args.subcommand == 'deactivate':
+        print_json(set_key_active(iam, args.iam_user, args.key_id, active=False))
+    elif args.subcommand == 'reactivate':
+        print_json(set_key_active(iam, args.iam_user, args.key_id, active=True))
+    elif args.subcommand == 'remove':
+        print_json(remove_key(iam, args.iam_user, args.key_id, force=args.force))
```
millerdev, line 72: Is it possible to run without a subcommand? Would it make sense to add an `else` clause to handle the case of no matching subcommand?

millerdev (approved):
Couple suggestions, but I think nothing blocking.

scripts/aws/manage-iam-keys.py
dannyroberts: I think there's already a built-in limit of two keys per IAM user.

Yeah okay here's where it says that: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html.

> You can have a maximum of two access keys per user.

So no I don't see that being an issue—probably not worth trying to optimize further

scripts/aws/manage-iam-keys.py
dannyroberts: Here's the behavior without a subcommand, tested locally:

```
(ccc) ➜  commcare-cloud git:(dmr/manage-iam-keys) ./scripts/aws/manage-iam-keys.py commcare-staging:ses ses-staging
usage: manage-iam-keys.py [-h] aws_profile iam_user {list,create,deactivate,reactivate,remove} ...
manage-iam-keys.py: error: the following arguments are required: subcommand
(ccc) ➜  commcare-cloud git:(dmr/manage-iam-keys) echo $?
2
```
So the parser will notice if the subcommand is missing, print an error, and exit with a non-zero status code. I think due to the `required=True` in the `'subcommand'` arg (assigned to the `subparser` variable) above.

dannyroberts: I suppose I could add something like

```
else:
    assert False, "Since `subcommand` is required, this code should not be reachable"
```

to make that extra-explicit, but on the flip side it's kind of redundant with the `required=True` on `subcommand` above, so it's a bit of a wash—in general I do like to see if/elif have an else like you're saying when it's clearly being used as like a `switch` like this, but I'm not sure it adds much in this case.

PR #6877, opened 2026-05-08
Make derive_ses_smtp_password.py more naturally scriptable

dannyroberts:
https://dimagi.atlassian.net/browse/SAAS-19728

Replaces the `--secret <secret>` argument (an anti-pattern: secrets on the command line leak into shell history and `ps`) with a positional file argument that accepts a path or `-` for stdin. Adds a `--csv` flag that interprets the input as the access-keys CSV that AWS provides for download, so you can pipe / pass that file directly without preprocessing.

### Before
```
SES_SECRET=$(cat ${CREDENTIAL_FILE} | head -n2 | tail -n1 | cut -d',' -f2)
./scripts/aws/derive_ses_smtp_password.py --region us-east-1 --secret ${SES_SECRET}
```

### After
```
./scripts/aws/derive_ses_smtp_password.py --region us-east-1 --csv /path/to/iam_user_name_accessKeys.csv
# or
cat /path/to/iam_user_name_accessKeys.csv | ./scripts/aws/derive_ses_smtp_password.py --region us-east-1 --csv 
# or (with just the bare secret in the clipboard)
pbpaste | ./scripts/aws/derive_ses_smtp_password.py --region us-east-1 -
```

### Test plan
Tested the following manually against a real (now-expired) access key before and after the change to make sure the output was in all cases the same as it was previously:

- [x] Bare secret with a file
- [x] Bare secret via stdin (`-`)
- [x] `--csv` with a file
- [x] `--csv` via stdin

I also checked that our internal documentation pointing to this command references the comment in this file and does not provide its own instructions, so it does not need to be updated.

### DIFF
--- a/scripts/aws/derive_ses_smtp_password.py
+++ b/scripts/aws/derive_ses_smtp_password.py
@@ -2,14 +2,14 @@
 
 # Copy-paste-edited from https://docs.aws.amazon.com/ses/latest/DeveloperGuide/smtp-credentials.html
 # Example Usage:
-#   CREDENTIAL_FILE=/path/to/downloaded/iam_user_name_accessKeys.csv
-#   SES_SECRET=$(cat ${CREDENTIAL_FILE} | head -n2 | tail -n1 | cut -d',' -f2)
-#   ./scripts/aws/derive_ses_smtp_password.py --region us-east-1 --secret ${SES_SECRET}
+#   ./scripts/aws/derive_ses_smtp_password.py --region us-east-1 --csv /path/to/iam_user_name_accessKeys.csv
+#   pbpaste | ./scripts/aws/derive_ses_smtp_password.py --region us-east-1 -
 
 import hmac
 import hashlib
 import base64
 import argparse
+import csv
 
 # Values that are required to calculate the signature. These values should
 # never change.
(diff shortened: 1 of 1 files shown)


dannyroberts:
(All the linting "errors" are on lines that aren't actually touched by this PR, so I'm going to decline to respond to them.)

scripts/aws/derive_ses_smtp_password.py
```diff
@@ -63,7 +83,15 @@ def main():
             action="store")
     args = parser.parse_args()
 
-    calculateKey(args.secret.strip(), args.region)
+    with args.file as f:
+        content = f.read()
```
gherceg, line 87: Not sure if this is relevant, but somewhat similar to https://github.com/dimagi/commcare-hq/pull/37706, if you pass in `-` for the `file` arg, would this context manager end up closing stdin? And if so, would that be an issue or no?

gherceg (commented):
One comment I want to clarify before approving.

scripts/aws/derive_ses_smtp_password.py
dannyroberts: Reasonable question. I think it does close stdin, but that basically just acts as an assertion that we don't attempt to read from this same stdin handle later—and that makes sense, we're intending to consume all of stdin right here.

gherceg: Yeah that makes sense to me.

gherceg (approved):

PR #6878, opened 2026-05-08
Support creating AWS profiles for SES via aws-sign-in

dannyroberts:
[SAAS-19727](https://dimagi.atlassian.net/browse/SAAS-19727)

##### Summary

The `manage-iam-keys.py` script in #6872 needs an AWS profile set up the way commcare-cloud already sets up other ones — but pointing at the account where SES lives and with permissions to manage specific IAM user credentials. This is in service of scripting the SES SMTP credential rotation described in [this runbook](https://dimagi.atlassian.net/wiki/spaces/cc/pages/2146599918/How+to+rotate+AWS+SES+SMTP+email+credentials).

This PR adds a `--ses` flag to `cchq <env> aws-sign-in` that reads new config from each env's `aws.yml` and writes out such a profile. The new `<aws_profile>:ses` profiles live alongside the existing `<aws_profile>:session` used by terraform, and a developer doesn't need to know how either is set up.

A couple of things worth flagging for review, since they don't show up in the diff:

- `iam_username` is unused so far — added preemptively for one of the next tasks in https://dimagi.atlassian.net/browse/SAAS-19724, which will use it manage the SES IAM user during rotation.
- `aws_sign_in_for_ses` deliberately skips the v1-credentials sync that the regular SSO flow does. That sync is a terraform compatibility shim; SES tooling uses modern AWS CLI / boto3, which read SSO natively.

##### Environments Affected
production, staging, india, eu — all gain a new `ses_config` block in `aws.yml`. No services or hosts are touched; the field is read only by the new `cchq <env> aws-sign-in --ses` flow.

##### Testing

- [x] `cchq staging aws-sign-in --help` shows the new flag
- [x]  existing terraform tests still pass
- [x] The new `cchq staging aws-sign-in --ses` command creates a profile that works with `manage-iam-keys.py` (clear with `rm -rf ~/.aws/sso ~/.aws/cli` first)
- [x] Normal `cchq staging aws-sign-in` also still works (clear with `rm -rf ~/.aws/sso ~/.aws/cli` first)

[SAAS-19727]: https://dimagi.atlassian.net/browse/SAAS-19727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commcare_cloud/commands/terraform/aws.py
+++ b/src/commcare_cloud/commands/terraform/aws.py
@@ -443,8 +443,7 @@ def aws_sign_in_for_ses(environment):
 def _ensure_sso_profile_signed_in(
         profile_name, sso_start_url, sso_region, sso_account_id, sso_role_name, region):
     """Make sure ~/.aws/config has a [profile <profile_name>] section, and that we have a valid SSO session."""
-    _ensure_all_dirs(AWS_DOT_DIR)
-    for path in (AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
+    for path in (AWS_DOT_DIR, AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
         _ensure_all_dirs(path)
 
     # todo: add `... or if _date_modified(AWS_CONFIG_PATH) > _date_modified(AWS_CREDENTIALS_PATH)`
--- a/src/commcare_cloud/commands/terraform/aws.py
+++ b/src/commcare_cloud/commands/terraform/aws.py
@@ -338,8 +338,6 @@ def aws_sign_in(environment, duration_minutes=DEFAULT_SIGN_IN_DURATION_MINUTES, 
 
     _ensure_all_dirs(AWS_DOT_DIR)
     if environment.aws_config.credential_style == 'sso':
-        for path in (AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
-            _ensure_all_dirs(path)
         return _aws_sign_in_with_sso(environment)
     else:
         return _aws_sign_in_with_iam(environment.terraform_config.aws_profile, duration_minutes=duration_minutes,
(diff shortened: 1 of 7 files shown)


src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -394,27 +392,45 @@ def _aws_sign_in_with_sso(environment):
...
+
+    return aws_session_profile
+
+
+def _ensure_sso_profile_signed_in(
+        profile_name, sso_start_url, sso_region, sso_account_id, sso_role_name, region):
+    """Make sure ~/.aws/config has a [profile <profile_name>] section, and that we have a valid SSO session."""
+    _ensure_all_dirs(AWS_DOT_DIR)
+    for path in (AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
+        _ensure_all_dirs(path)
```
gherceg, line 447: Messed up my autosuggestion, but 
```
    _ensure_all_dirs(AWS_DOT_DIR)
    for path in (AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
        _ensure_all_dirs(path)
```
can be simplified to
```
    for path in (AWS_DOT_DIR, AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
        _ensure_all_dirs(path)
```

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -394,27 +392,45 @@ def _aws_sign_in_with_sso(environment):
...
+    )
+
+    if not _has_valid_v1_session_credentials(aws_session_profile):
+        _sync_sso_to_v1_credentials(aws_session_profile)
+
+    return aws_session_profile
+
+
+def _ensure_sso_profile_signed_in(
+        profile_name, sso_start_url, sso_region, sso_account_id, sso_role_name, region):
```
gherceg, line 444: I can't decide if the sso_ prefix on args feels redundant or not given this function is `_ensure sso_profile_signed_in`.

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -311,13 +311,20 @@ class AwsSignIn(CommandBase):
...
     ]
 
     def run(self, args, unknown_args):
         environment = get_environment(args.env_name)
-        duration_minutes = args.duration_minutes
-        aws_sign_in(environment, duration_minutes, force_new=True)
+        if args.ses:
+            aws_sign_in_for_ses(environment)
+        else:
+            aws_sign_in(environment, args.duration_minutes, force_new=True)
```
gherceg, line 327: We already have a similar if/else block, granted for checking credential style (sso vs iam), in `aws_sign_in`. This way is nice since it avoids `aws_sign_in` needing to know about this `ses` param, but I also wonder if the generally more intuitive way is for `aws_sign_in` to receive a "profile" arg and act call the appropriate function there?

gherceg (commented):
First pass. Looks good. Just a couple of comments

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: Oh, sure. Yeah I missed that

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: Yeah, I think I'll keep them, because some of them are specifically sso parameters (things that are needed to initiate the SSO) and others aren't. E.g. `sso_region` is the region where SSO is handled, and `region` is the default region for the profile.

src/commcare_cloud/commands/terraform/aws.py
dannyroberts: I think I'll keep it with the branching up here, because this is branching on an command arg, whereas the one inside is branching on a config param (incidentally one that is only ever set to 'sso' anymore, so we can probably safely remove it). Sometimes it makes sense to pass down a command arg, like when it's a minor modification of functionality like --quiet or something, but in this case it's kind of a top level thing saying what thing to create, and those two things have slightly different, if similar/parallel, ways they get created.

src/commcare_cloud/commands/terraform/aws.py
```diff
@@ -393,28 +398,71 @@ def _aws_sign_in_with_sso(environment):
...
+    )
+    return aws_ses_profile
+
+
+def _ensure_sso_profile_signed_in(
+        profile_name, sso_start_url, sso_region, sso_account_id, sso_role_name, region):
+    """Make sure ~/.aws/config has a [profile <profile_name>] section, and that we have a valid SSO session."""
+    _ensure_all_dirs(AWS_DOT_DIR)
+    for path in (AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
+        _ensure_all_dirs(path)
```
dannyroberts:
```suggestion
    for path in (AWS_DOT_DIR, AWS_SSO_CACHE_DIR, AWS_CLI_CACHE_DIR):
        _ensure_all_dirs(path)
```

gherceg (approved):

PR #6886, opened 2026-05-19
Clean up control/init.sh and various other old cruft

millerdev:
Removes a collection of long-obsolete code paths so commcare-cloud installs cleanly on modern Python and `control/init.sh` only does work that's still relevant.

:blowfish: Review by commit.

##### Environments Affected

None. Developer/packaging cleanup only, no environment file edits.


docs/source/reference/1-commcare-cloud/commands.md
```diff
@@ -1171,60 +1171,6 @@ authenticate using the pem file (or prompt for root password if there is no pem
...
-
-The name of the obsolete fab command.
-
-##### Options
-
-###### `-l`
-
-Use `-l` instead of a command to see the full list of commands.
-
-##### Obsolete fab commands
```
gherceg, line 1194: Certainly not advocating for keeping this, but worth mentioning that I did find myself using this specific section from time to time to remember what command to run. I don't _think_ there is an equivalent doc in commcare-cloud that translates more intuitive english names for some action into a commcare-cloud command, but I imagine that could be nice? Out of the scope of this PR for sure, just curious what you think.

src/commcare_cloud/environment/paths.py
```diff
@@ -22,9 +22,10 @@ def get_virtualenv_bin_path():
     return os.path.dirname(sys.executable)
 
 
+_SITE_PACKAGES = get_paths()['purelib']
```
gherceg, line 25: It's not obvious to me why `'purelib'` is the right key here, but I also don't think anything needs to exist in code to explain that since it is pretty clear that it gets you the site packages from good naming. Just wish the commit message elaborated a bit on that, but maybe just leaving a note on this thread could achieve the same thing.

gherceg (approved):
No qualms. Everything being removed falls well outside of the 6 week boundary, so it is fair to have expected someone to have run `init.sh` since.

docs/source/reference/1-commcare-cloud/commands.md
gherceg: In case unclear, I guess I'm saying the fab commands had slightly more intuitive names than the current ansible commands.

src/commcare_cloud/environment/paths.py
millerdev: Yes, agreed, I also had to [look it up](https://docs.python.org/3/library/sysconfig.html#installation-paths) after Claude suggested `'purelib'`.

> gets you the site packages from good naming

That was how I hoped this would work.

docs/source/reference/1-commcare-cloud/commands.md
millerdev: Interesting. I don't think I've ever looked at this, but I can understand the use case. Seems reasonable to add a page to the docs or rename commands if the remaining names are not intuitive.

PR #6894, opened 2026-05-28
Migrate install workflow from pip+pip-tools to uv

millerdev:
Switches commcare-cloud's installation and dependency-management pipeline from `pip` + `pip-tools` to [uv](https://docs.astral.sh/uv/). All bootstrap entry points (`init.sh`, Vagrant provisioning, monolith install, control-user provisioning, CI) now run `uv sync`, giving fast, reproducible installs. The user-facing shell flow no longer activates a venv — `cchq`, `commcare-cloud`, and `ansible-playbook` are exposed by prepending the project venv's `bin/` to `PATH`.

I have started work on optimizing `ansible-galaxy install` to speed that up as well, but decided not to include it in this PR because it seems like a separate concern. That change will be done in a follow-up PR.

:blowfish: Review by commit.

## Environments Affected

All control machines. No special action is required, the switch to uv should be automatic when code is updated.


millerdev:
Except for `jsonobject` and `wrapt`, requirements versions did not change in this PR, so the security alerts are interesting but were not introduced here. Upgrades to fix them are considered out of scope for this PR.

src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
```diff
@@ -97,10 +97,21 @@
...
     name:
-      - uv>=0.7
       - sh
     umask: "0022"
+
+- name: Remove pip-installed uv
   become: yes
+  pip:
+    name: [uv]
+    absent: true
```
gherceg, line 117: Do you plan to come back in ~6 weeks and remove this? Just curious

gherceg (approved):
🎉

src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
millerdev: No, because I don't think we expect people to run `deploy_stack` every 6 weeks. I see it being here ~indefinitely, which I don't love, but also not sure how to do it better. Guess we could publish a change log.

PR #6903, opened 2026-06-02
Migrate test suite from nose to pytest

millerdev:
The Python test suite now runs on **pytest** instead of **nosetests**. `nose` is unmaintained and was pinned to a Dimagi fork installed from GitHub. No tests were dropped; pytest collects and passes all tests previously run by nose.

:blowfish: Review by commit.

##### Environments Affected

None


tests/test_file_migration.py
```diff
@@ -95,5 +94,4 @@ def test_parse_plan():
 def _check_file_contents(generated_path, expected_path):
     expected_script = get_file_contents(expected_path)
     script_source = get_file_contents(generated_path)
-    assert_multi_line_equal.__self__.maxDiff = None
```
MartinRiese, line 98: This was only for the output, correct?

src/commcare_cloud/commands/terraform/tests/test_parameter_groups.py
```diff
@@ -17,12 +17,12 @@ def test_validate_references_to_parameter_groups():
     )
 
     # assert invalid reference
-    with assert_raises(ValueError) as context:
+    with pytest.raises(ValueError) as context:
```
MartinRiese: I think you could do `with pytest.raises(ValueError, match=r".*pg0.*nonexistent.*")` and skip the two asserts at the end too. But for the sake of being a clean conversion this is probably better.

src/commcare_cloud/commands/terraform/tests/test_postgresql_units.py
```diff
@@ -18,16 +18,16 @@ def test_convert_to_unit__bytes():
...
+    assert_contains(context.value.args[0], '12.5 GB')
 
 
 def test_convert_to_unit__mixed_units():
-    with assert_raises(ValueError) as context:
+    with pytest.raises(ValueError) as context:
         convert_to_unit('1kB', MS)
-    assert_contains(context.exception.args[0], '1kB')
-    assert_contains(context.exception.args[0], 'TimeInMilliseconds')
+    assert_contains(context.value.args[0], '1kB')
```
MartinRiese, line 29: Shouldn't these lines be indented one more level? Or how are they accessing `context` at the same indentation as the `with`?

MartinRiese (commented):
LGTM other than those indentations

src/commcare_cloud/commands/terraform/tests/test_postgresql_units.py
millerdev: `context` does not go out of scope at the end of the with block.

tests/test_file_migration.py
millerdev: Correct. pytest does a better job of highlighting differences while minimizing irrelevant matching content, so it should not be necessary anymore.

gherceg (approved):
🎉

src/commcare_cloud/commands/terraform/tests/test_parameter_groups.py
millerdev: b00bc905aa35ff36050803be6aa3e19fece129ac

gherceg (approved):

PR #6904, opened 2026-06-02
Support zsh in control/init.sh

gherceg:
Just setting up a new machine and ran into this with zsh. Was flagged because running `source control/init.sh` led to the commcare-cloud repo being cloned again incorrectly.

I opted for telling the operator to update their .zshrc file with the autocomplete directives as that is better practice than commcare-cloud attempting to update that in this script. 

##### Environments Affected

None

### DIFF
--- a/control/init.sh
+++ b/control/init.sh
@@ -80,11 +80,12 @@ NC='\033[0m' # No Color
 
 if [ -z ${CI_TEST} ]; then
   if ! grep -q init-ansible ~/.profile 2>/dev/null; then
-    if [ $NO_INPUT == 1 ]; then
+    if [[ $NO_INPUT == 1 ]]; then
       yn='y'
     else
       printf "${YELLOW}Do you want to have the CommCare Cloud environment setup on login?${NC}\n"
-      read -t 30 -p "(y/n): " yn
+      printf "(y/n): "
+      read -t 30 yn
     fi
     case $yn in
         [Yy]* )
--- a/control/init.sh
+++ b/control/init.sh
@@ -72,11 +72,20 @@ fi
 alias update-code='${COMMCARE_CLOUD_REPO}/control/update_code.sh && (cd ${COMMCARE_CLOUD_REPO} && uv sync --quiet)'
 alias update_code=update-code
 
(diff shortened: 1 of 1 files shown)


millerdev (approved):

control/init.sh
```diff
@@ -1,7 +1,17 @@
 #! /bin/bash
 NO_INPUT=${NO_INPUT:-0}
 
-if [[ $_ == $0 ]]; then
+# Resolve this script's path in both bash and zsh
+is_sourced=0
+if [ -n "$ZSH_VERSION" ]; then
+    [[ $ZSH_EVAL_CONTEXT == *:file* ]] && is_sourced=1
+    script_path="$0"
+elif [ -n "$BASH_VERSION" ]; then
```
millerdev, line 9: Should this use `else` rather than `elif`? Not sure if other shells provide `BASH_SOURCE`? Is there a more portable fall-back?

gherceg: That does seem like it would be more consistent behavior with how it has been, but I wonder if it would be better to just have an additional `else` case where we raise an error saying only bash and zsh are supported, or does that seem unfair?

millerdev: I think change this condition to `-n "${BASH_SOURCE[0]}"` and add an else clause that prints a _script path cannot be found_ warning.

~There is a question of whether we should check likely locations like `./commcare-cloud/control/init.sh` and `./control/init.sh`, and set `script_path` to the first found if either of those exist?~ Refuse the temptation to guess.

Option 2 is to fall back to not setting `script_path` and falling through to `COMMCARE_CLOUD_REPO=${COMMCARE_CLOUD_REPO:-${HOME}/commcare-cloud}`, which may re-clone the repo at `~/commcare-cloud`.

Option 3 is to set `script_path=./control/init.sh` which will fail with an error

```
It looks like you're running this script from an unexpected location...
```
if `./control/update_code.sh` does not exist. (Is this too much like guessing?)

Thoughts?

gherceg: > I think change this condition to -n "${BASH_SOURCE[0]}" and add an else clause that prints a script path cannot be found warning.

I think this a good option.

gherceg: https://github.com/dimagi/commcare-cloud/pull/6913

PR #6905, opened 2026-06-02
Ansible module to restart ec2 instances 

AmitPhulera:
https://dimagi.atlassian.net/browse/SAAS-19382 

A redo of https://github.com/dimagi/commcare-cloud/pull/6858, the changes were so much that it made no sense continuing on the existing branch.

This is part of the groundwork required to automate the machine restarts. 

I have tested the command locally on staging `web13` machine and the results were as expected, following are the output for each command -
 
- describe

```
$ cat > /tmp/args.json <<'EOF'
{"ANSIBLE_MODULE_ARGS": {"instance_ids": ["i-0fb86252d8c3490e9"], "command": "describe"}}
EOF
python src/commcare_cloud/ansible/library/ec2_instance_state.py /tmp/args.json

{"changed": false, "command": "describe", "instances": [{"instance_id": "i-0fb86252d8c3490e9", "previous_state": "running", "current_state": "running", "name": "web13-staging", "instance_type": "t3a.xlarge", "availability_zone": "us-east-1a", "private_ip": "10.201.10.244", "public_ip": null, "tags": {"Name": "web13-staging", "Environment": "staging", "Group": "webworkers"}, "launch_time": "2025-08-30T17:34:24+00:00"}], "unchanged_instance_ids": [], "diff": {"before": {"states": {"i-0fb86252d8c3490e9": "running"}}, "after": {"states": {"i-0fb86252d8c3490e9": "running"}}}, "invocation": {"module_args": {"instance_ids": ["i-0fb86252d8c3490e9"], "command": "describe", "wait": true, "region": null}}}
```
- start

```
$ cat > /tmp/args.json <<'EOF'
{"ANSIBLE_MODULE_ARGS": {"instance_ids": ["i-0fb86252d8c3490e9"], "command": "start"}}
EOF
python src/commcare_cloud/ansible/library/ec2_instance_state.py /tmp/args.json

{"changed": true, "command": "start", "instances": [{"instance_id": "i-0fb86252d8c3490e9", "previous_state": "stopped", "current_state": "running", "name": "web13-staging", "instance_type": "t3a.xlarge", "availability_zone": "us-east-1a", "private_ip": "10.201.10.244", "public_ip": null, "tags": {"Name": "web13-staging", "Environment": "staging", "Group": "webworkers"}, "launch_time": "2026-06-05T10:14:44+00:00"}], "unchanged_instance_ids": [], "diff": {"before": {"states": {"i-0fb86252d8c3490e9": "stopped"}}, "after": {"states": {"i-0fb86252d8c3490e9": "running"}}}, "invocation": {"module_args": {"instance_ids": ["i-0fb86252d8c3490e9"], "command": "start", "wait": true, "region": null}}}
```

- stop

```
$ cat > /tmp/args.json <<'EOF'
{"ANSIBLE_MODULE_ARGS": {"instance_ids": ["i-0fb86252d8c3490e9"], "command": "stop"}}          
EOF
python src/commcare_cloud/ansible/library/ec2_instance_state.py /tmp/args.json

{"changed": true, "command": "stop", "instances": [{"instance_id": "i-0fb86252d8c3490e9", "previous_state": "running", "current_state": "stopped", "name": "web13-staging", "instance_type": "t3a.xlarge", "availability_zone": "us-east-1a", "private_ip": "10.201.10.244", "public_ip": null, "tags": {"Name": "web13-staging", "Environment": "staging", "Group": "webworkers"}, "launch_time": "2026-06-05T10:12:22+00:00"}], "unchanged_instance_ids": [], "diff": {"before": {"states": {"i-0fb86252d8c3490e9": "running"}}, "after": {"states": {"i-0fb86252d8c3490e9": "stopped"}}}, "invocation": {"module_args": {"instance_ids": ["i-0fb86252d8c3490e9"], "command": "stop", "wait": true, "region": null}}}
```

- stop_start

```
$ cat > /tmp/args.json <<'EOF'
{"ANSIBLE_MODULE_ARGS": {"instance_ids": ["i-0fb86252d8c3490e9"], "command": "stop_and_start"}}
EOF
python src/commcare_cloud/ansible/library/ec2_instance_state.py /tmp/args.json

{"changed": true, "command": "stop_and_start", "instances": [{"instance_id": "i-0fb86252d8c3490e9", "previous_state": "running", "current_state": "running", "name": "web13-staging", "instance_type": "t3a.xlarge", "availability_zone": "us-east-1a", "private_ip": "10.201.10.244", "public_ip": null, "tags": {"Name": "web13-staging", "Environment": "staging", "Group": "webworkers"}, "launch_time": "2026-06-05T10:12:22+00:00"}], "unchanged_instance_ids": [], "diff": {"before": {"states": {"i-0fb86252d8c3490e9": "running"}}, "after": {"states": {"i-0fb86252d8c3490e9": "running"}}}, "invocation": {"module_args": {"instance_ids": ["i-0fb86252d8c3490e9"], "command": "stop_and_start", "wait": true, "region": null}}}
```
##### Environments Affected

All


src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -0,0 +1,141 @@
...
+---
+module: ec2_instance_state
+
+short_description: Start, stop, stop_and_start, or describe EC2 instances by ID.
+
+description:
+    - Manages the running state of EC2 instances given an explicit list of
+      instance IDs. Supports four commands - describe, start, stop, stop_and_start,
+      and is idempotent (no API call is made if the instance is already in the requested state).
+    - Designed to run with delegate_to localhost. AWS credentials and the target region are picked up from the standard boto3 credential chain; in the commcare-cloud workflow the AWS_PROFILE and AWS_REGION environment variables are exported automatically before ansible runs.
```
gherceg: I'm just not too familiar with this. Do you mind elaborating on the "delete_to localhost" workflow?

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -0,0 +1,141 @@
...
+    }
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=True)
+    params = module.params
+
+    instance_ids = params['instance_ids']
+    if not instance_ids:
+        module.fail_json(msg="'instance_ids' must be a non-empty list.")
+
+    bad = [i for i in instance_ids if not INSTANCE_ID_RE.match(i)]
+    if bad:
```
gherceg: nit
```suggestion
    bad_ids = [i for i in instance_ids if not INSTANCE_ID_RE.match(i)]
    if bad_ids:
```

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -117,6 +118,29 @@ class InstanceState(str, Enum):
...
+
+
+def _get_ec2_client(region):
+    """Return a boto3 EC2 client. Defined as a module-level function so tests can patch it."""
+    try:
+        import boto3
+    except ImportError:
+        raise RuntimeError(
+            "boto3 is required by ec2_instance_state but is not installed."
+        )
```
gherceg: Why not call `module.fail_json` directly here, similar to what `_get_region` does, and then that way `main` doesn't need to wrap the call to _get_ec2_client in a try/except.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -140,6 +140,17 @@ def _get_ec2_client(region):
         )
     return boto3.client('ec2', region_name=region)
 
+class _Ctx:
+    """Per-run context shared by the flow helpers.
+
+    Bundles the EC2 client, the AnsibleModule, so these don't have to be
+    passed as arguments to every helper.
+    """
```
gherceg:
```suggestion
    """
    Bundles the EC2 client and Ansible module for convenience
    """
```

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -140,6 +140,17 @@ def _get_ec2_client(region):
         )
     return boto3.client('ec2', region_name=region)
 
+class _Ctx:
```
gherceg: Not my favorite name, but need to continue reviewing to offer suggestions.

tests/ansible.py
```diff
@@ -25,16 +25,22 @@ def run(module, args):
     provided to load from some other location); otherwise a module
     object is expected.
     """
+    warnings = []
     def exit_json(*args, **kw):
+        kw.setdefault("warnings", list(warnings))
```
gherceg: Isn't `warnings` already a list here? Why wrap in `list(...)`?

tests/test_ec2_instance_state.py
```diff
@@ -0,0 +1,99 @@
+# Add the library directory to sys.path so we can import the module by name.
+# We must also ensure that the tests/ directory does NOT appear before the
+# library dir on sys.path, because tests/ansible.py would otherwise shadow
+# the real `ansible` package when ec2_instance_state does
+# `from ansible.module_utils.basic import AnsibleModule`.
```
gherceg: nit: I'd put these right next to the TESTS_DIR and LIBRARY_DIR definitions.

tests/test_ec2_instance_state.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def test_invalid_command_fails(self):
+        os.environ['AWS_REGION'] = 'us-east-1'
+        result = run_module({
+            'instance_ids': ['i-0123456789abcdef0'],
+            'command': 'bogus',
+        })
+        assert result['failed']
+        expected_msg = "value of command must be one of: describe, start, stop, stop_and_start, got: bogus"
+        assert result['result']['msg'] == expected_msg, "Expected {expected_msg} but got {result_msg}".format(expected_msg=expected_msg, result_msg=result['result']['msg'])
```
gherceg: nit: f-string used above, why change to `.format(...)` for the rest of these tests?

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -140,6 +140,72 @@ def _get_ec2_client(region):
...
+        self.previous_state = self.state
+        self.current_state = self.state
+
+    @property
+    def state(self):
+        return self.raw['State']['Name']
+
+    @property
+    def is_terminal(self):
+        return self.state in TERMINATED_STATES
```
gherceg, line 162: I know the naming here is tricky, but if we want to call this `is_terminal`, then I think we should call the variable representing the list of states `TERMINAL_STATES`. Other options that come to mind are `DESTROYED_STATES`/`is_destroyed` or `PURGED_STATES`/`is_purged`, but I don't think those are meaningfully better than `TERMINAL_STATES`/`is_terminal`, which does seem like the most accurate.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -140,6 +140,72 @@ def _get_ec2_client(region):
...
+
+    @property
+    def name(self):
+        """The 'Name' tag value, if the instance has one. Otherwise return the private IP address."""
+        for tag in self.raw.get('Tags', []):
+            if tag['Key'] == 'Name':
```
gherceg, line 168: Just noticing the difference in accessing `self.raw.get('Tags', [])` vs `tag['Key']`. I assume you don't expect a KeyError to be raised in the second case, but just confirming.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -140,6 +140,72 @@ def _get_ec2_client(region):
...
+                return tag['Value']
+        return self.raw.get('PrivateIpAddress')
+
+    @property
+    def label(self):
+        """Human-friendly identifier for error messages."""
+        return f"{self.instance_id} ({self.name})"
```
gherceg: nit: if this is meant to be human friendly, what do you think about flipping this
```suggestion
        return f"{self.name} ({self.instance_id})"
```

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -140,6 +140,72 @@ def _get_ec2_client(region):
...
+
+    @property
+    def can_start(self):
+        """True if a StartInstances call is required to reach 'running'.
+
+        A 'stopping' instance is included: it must first be awaited to 'stopped',
+        then started.
+        """
```
gherceg: I'm between getting rid of the docstring entirely versus keeping just the bit about needing to wait for 'stopping' to reach 'stopped' before issuing a start command. Given this method doesn't handle any of the wait logic, it seems fine to not mention anything related to waiting for states here.

If you did want to keep that bit...
```suggestion
        """
        A 'stopping' instance is included but cannot be started until it reaches 'stopped'
        """
```

Applies to the `can_stop` method below as well.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -253,7 +253,71 @@ def _build_payload(instances, command, changed, unchanged_instance_ids):
...
+def _wait_for(ctx, waiter_name, wait_instances):
+    if not wait_instances or ctx.module.check_mode:
+        return
+    waiter = ctx.client.get_waiter(waiter_name)
+    try:
+        waiter.wait(InstanceIds=[i.instance_id for i in wait_instances])
+    except Exception as e:  # noqa: BLE001 - surface any waiter failure as module failure
+        ctx.module.fail_json(
+            msg=f"Waiter {waiter_name!r} failed for {_labels(wait_instances)}: {e}")
+        return
```
gherceg: nit: `wait_instances` -> `instances`

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -253,7 +253,71 @@ def _build_payload(instances, command, changed, unchanged_instance_ids):
...
+            instances[iid].current_state = InstanceState.PENDING
+
+    return _build_payload(instances, InstanceCommand.START, changed, unchanged)
+
+def _labels(instances):
+    """Render a comma-separated list of human-friendly labels for the Instances."""
+    return ', '.join(i.label for i in instances)
+
+
+def _check_not_terminated(ctx, instances, action):
```
gherceg: nit: my preference would be to not pass `action` into this method since when I first saw `_check_not_terminated(ctx, instances, InstanceCommand.START)`, my first thought was "why would the action impact a check for terminated instances?". I guess the alternatives are generalize the error message or leave it up to the caller to call module.fail_json.

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -321,7 +321,52 @@ def _wait_for(ctx, waiter_name, wait_instances):
...
+
+    if targets:
+        try:
+            ctx.client.stop_instances(InstanceIds=targets)
+        except Exception as e:  # noqa: BLE001
+            labels = _labels(instances[iid] for iid in targets)
+            ctx.module.fail_json(msg=f"StopInstances failed for {labels}: {e}")
+            return
+
+    wait_for_stopped = list(targets) + list(already_stopping)
```
gherceg: `targets` and `already_stopping` are added together on line 336 as well. Could put this variable up with where `targets` and `already_stopping` are already defined. I also don't think these need to be wrapped in `list(...)` right?

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -370,7 +370,47 @@ def _do_stop(ctx, instance_ids, wait):
 
 
 def _do_stop_and_start(ctx, instance_ids, wait):
-    return {}
+    stop_payload = _do_stop(ctx, instance_ids, wait=True)
+
+    # Honor the user's `wait` choice for the final running wait.
+    start_payload = _do_start(ctx, instance_ids, wait=wait)
+
+    # Combine: previous_state = state before the stop; current_state = after start.
```
gherceg: nit: remove

src/commcare_cloud/ansible/library/ec2_instance_state.py
```diff
@@ -370,7 +370,47 @@ def _do_stop(ctx, instance_ids, wait):
...
+    # check there.
+    changed = True
+    if not ctx.module.check_mode and not start_payload['changed']:
+        ctx.module.fail_json(
+            msg="stop_and_start invariant violated: start phase reported no "
+                "change after a completed stop phase.",
+            start_payload=start_payload)
+
+    # unchanged = ids that were no-ops in BOTH phases.
+    # Highly unlikely to happen in practice, but we sort to make the result deterministic.
```
gherceg: How would this be possible?

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: It is `delegate_to localhost`. This is the command that we want to run locally on the host from where the command was run instead of running it on the target machine.
For this case the AWS credentials will be on our local system so we want to run the module from our local systems.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Thats a good point. Will update it.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Yes. If there are objects in tags, they will definitely have `Key` and `Value`. This would be guaranteed by AWS.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: This would only happen when the instance state would be terminated but given the fact we are checking for it already so it should never happen. But I kept it for to keep the output same.

src/commcare_cloud/ansible/library/ec2_instance_state.py
millerdev: All of the functions that accept this type (like `_do_start(ctx, ...)`) look like instance methods. What do you think of renaming this to something like `StopStarter` and moving those functions to methods?
```py
class StopStarter:

    def __init__(self, client, module):
        self.client = client
        self.module = module

    def describe(self, instance_ids):
        ...

    def start(self, instance_ids, wait):
        ...

    def stop(self, instance_ids, wait):
        ...

    ...
```

Feel free to pick a different name, that was just the first thing that come to mind.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/changes/46189a29d5c906a7c26a8338c468b1dacfe4bee6

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/changes/69749cbd47d6a77767c88f3f35a0878fab56dc1a

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/changes/46189a29d5c906a7c26a8338c468b1dacfe4bee6#diff-c6913cbc417347f1f21fdf62868aba7c3a544d1561404dc8c4ff1c5581760b6bL409-L414

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/changes/46189a29d5c906a7c26a8338c468b1dacfe4bee6#diff-c6913cbc417347f1f21fdf62868aba7c3a544d1561404dc8c4ff1c5581760b6bL310-L316

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Good call. https://github.com/dimagi/commcare-cloud/pull/6905/changes/02e4d104d8dc63f6e665ce3bb926d3afcdc19afe

tests/test_ec2_instance_state.py
AmitPhulera: Realised that they are not needed at all because now I am using `ansible.run` already had do it.
https://github.com/dimagi/commcare-cloud/pull/6905/commits/813a7d54c4179eda59be4727af93896955e561b7

tests/test_ec2_instance_state.py
AmitPhulera: I added that test long back and have used format then - have updated to f-strings in https://github.com/dimagi/commcare-cloud/pull/6905/commits/9e12be6dbfc01e0ecb05ab74d55aaf394a0e669e

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: I ended up updating the property name to `is_terminated` which match with the variable. https://github.com/dimagi/commcare-cloud/pull/6905/commits/88b69ede46700d1620f8c30b0acc3a62f16cdd58

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/commits/46189a29d5c906a7c26a8338c468b1dacfe4bee6

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Removed docstrings altogether - https://github.com/dimagi/commcare-cloud/pull/6905/commits/10e870304e05b50c4feebd75e7c36a496205d258

tests/ansible.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/commits/6d3d240476b233353cca73b38326db891b61337f

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6905/commits/10e870304e05b50c4feebd75e7c36a496205d258

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: Thanks @millerdev. Used `EC2InstanceManager` class name https://github.com/dimagi/commcare-cloud/pull/6905/commits/616ad06c2b17e679dbdd92453671b5b93375ae21
This looks much better.

src/commcare_cloud/ansible/library/ec2_instance_state.py
AmitPhulera: While moving stuff to `EC2InstanceManager`, this was addressed. 
https://github.com/dimagi/commcare-cloud/pull/6905/commits/616ad06c2b17e679dbdd92453671b5b93375ae21

gherceg (approved):
Thanks for addressing the feedback. Looks like a good start point to build on. One last nit around commit messages, like the most recent one cf25fe25e3ec5fe82161963fed4cc4900a8f6558. I don't think we have any explicit guideline around this, but we probably should! I think the 50/72 rule is something we follow, so keeping the first line/subject line to 50 characters and then the multi-line description lengths to 72.

PR #6907, opened 2026-06-03
Ssh server is required

millerdev:
Previously ssh without a server caused a traceback with an unhelpful error:
```
TypeError: expected string or bytes-like object
```

`Lookup`'s server has `nargs="?"`, which makes the argument optional. This is valid for `Lookup`, but not for `_Ssh` and its subclasses.

##### Environments Affected

None


MartinRiese (approved):

gherceg (approved):

millerdev:
I need to rework this because of https://github.com/dimagi/commcare-cloud/commit/1477d757a236109d39dd11a74379b4c6a100af78. Should have read the whole help message that was moved _before_ creating this PR :facepalm:

millerdev:
Fixed. It's not as bad as I thought.

gherceg (approved):

PR #6910, opened 2026-06-08
Swap staging buckets

AmitPhulera:
https://dimagi.atlassian.net/browse/SAAS-19843

In the process of consolidating S3 buckets in the same AWS account as entire infra, we have already copied data over from the existing bucket to the new bucket that we created on staging account.

Now we would be switching staging to use the new account. 

##### Environments Affected

Staging

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -201,8 +201,8 @@ S3_BLOB_DB_SETTINGS = {
 {% if localsettings.get('BLOB_DB_MIGRATING_FROM_S3_TO_S3', False) %}
 OLD_S3_BLOB_DB_SETTINGS = {
     'url': '{{ old_s3_blob_db_url }}',
-    'access_key': {% if old_s3_blob_db_access_key %}'{{ old_s3_blob_db_access_key }}'{% else %}None{% endif %},
-    'secret_key': {% if old_s3_blob_db_secret_key %}'{{ old_s3_blob_db_secret_key }}'{% else %}None{% endif %},
+    'access_key': None,
+    'secret_key': None,
     's3_bucket': '{{ old_s3_blob_db_s3_bucket }}',
     'bulk_delete_chunksize': {{ s3_bulk_delete_chunksize }},
     'config': {'signature_version': 's3v4'},
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
@@ -204,7 +204,7 @@ OLD_S3_BLOB_DB_SETTINGS = {
     'access_key': {% if old_s3_blob_db_access_key %}'{{ old_s3_blob_db_access_key }}'{% else %}None{% endif %},
     'secret_key': {% if old_s3_blob_db_secret_key %}'{{ old_s3_blob_db_secret_key }}'{% else %}None{% endif %},
     's3_bucket': '{{ old_s3_blob_db_s3_bucket }}',
-    'bulk_delete_chunksize': {{ old_s3_bulk_delete_chunksize }},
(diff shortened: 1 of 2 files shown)


src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -204,7 +204,7 @@ OLD_S3_BLOB_DB_SETTINGS = {
     'access_key': {% if old_s3_blob_db_access_key %}'{{ old_s3_blob_db_access_key }}'{% else %}None{% endif %},
     'secret_key': {% if old_s3_blob_db_secret_key %}'{{ old_s3_blob_db_secret_key }}'{% else %}None{% endif %},
     's3_bucket': '{{ old_s3_blob_db_s3_bucket }}',
-    'bulk_delete_chunksize': {{ old_s3_bulk_delete_chunksize }},
+    'bulk_delete_chunksize': {{ s3_bulk_delete_chunksize }},
```
AmitPhulera, line 207: This parameter is used in - https://github.com/dimagi/commcare-hq/blob/9e83873106475b4c8ada323953fbeba8a25971f1/corehq/blobs/s3db.py#L137-L146

and it should be fine to have just one value for both the source and destination buckets rather than having two different values.

millerdev (approved):

src/commcare_cloud/ansible/roles/commcarehq/templates/localsettings.py.j2
```diff
@@ -201,10 +201,10 @@ S3_BLOB_DB_SETTINGS = {
 {% if localsettings.get('BLOB_DB_MIGRATING_FROM_S3_TO_S3', False) %}
 OLD_S3_BLOB_DB_SETTINGS = {
     'url': '{{ old_s3_blob_db_url }}',
-    'access_key': {% if old_s3_blob_db_access_key %}'{{ old_s3_blob_db_access_key }}'{% else %}None{% endif %},
-    'secret_key': {% if old_s3_blob_db_secret_key %}'{{ old_s3_blob_db_secret_key }}'{% else %}None{% endif %},
+    'access_key': None,
+    'secret_key': None,
```
millerdev, line 205: I'm curious about this change. Are `old_s3_blob_db_access_key` and `old_s3_blob_db_secret_key` set? Do we not want to allow those variables to be used in the future?

AmitPhulera: > Are old_s3_blob_db_access_key and old_s3_blob_db_secret_key set? 

Yeah. They were already set.  Probably from last time when we might have attempted it? I wanted to delete them but wanted to run them by Danny first. 

>Do we not want to allow those variables to be used in the future?

Part of me want to say no that we won't be using them given the fact that we rely on AWS IAM roles rather than access keys and we won't be doing this migration again ever(?)

AmitPhulera: @millerdev I am merging this PR now to ensure that no one runs update-config without this branch, but would post here on what I plan to do with the existing secrets.

millerdev: I think changing this will prevent anyone from ever migrating from one S3 blob db to another (on any environment), which doesn't seem right. I think the right thing to do here is to ensure that `old_s3_blob_db_access_key` and `old_s3_blob_db_secret_key` are not set in the environment(s) where you want this to apply.

millerdev: ...that is, migrating from S3 to S3 where the old and new access keys are different.

AmitPhulera: Fair point. I am remembering the MinIO issue where third party hosters might have to migrate to different service. I will roll it back and will delete the older keys today.

AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6912

PR #6915, opened 2026-06-15
Expose a cli command to restart ec2 instances

AmitPhulera:
Building on top of https://github.com/dimagi/commcare-cloud/pull/6905, we are exposing a cli which will help us to do restarts of ec2 instances. 

The PR creates a role, a playbook and a cli for `ec2_instance_state` command. 

The role would be used later in the playbooks where we would be issuing start and stop commands to the instances. 

The playbook was required to expose the role for the CLI.
##### Environments Affected

All

### DIFF
--- a/docs/source/reference/1-commcare-cloud/commands.md
+++ b/docs/source/reference/1-commcare-cloud/commands.md
@@ -1735,13 +1735,12 @@ have been made to our actual resources in AWS.
 
 ---
 
-#### ``ec2-instance-state`` Command
+#### ``ec2`` Command
 
 Manage the EC2 instance state (start/stop/describe) of hosts in an AWS environment.
 
 ```
-commcare-cloud <env> ec2-instance-state [--no-wait]
-                                        {describe,start,stop,stop_and_start} inventory_group [inventory_group ...]
+commcare-cloud <env> ec2 [--no-wait] {describe,start,stop,stop_and_start} inventory_group [inventory_group ...]
 ```
 
 `start` and `stop` show a check-mode preview of the state transitions
@@ -1755,9 +1754,9 @@ To force setup, use --control-setup=yes instead.
 ##### Example
 
 ```
-commcare-cloud <env> ec2-instance-state describe webworkers
-commcare-cloud <env> ec2-instance-state stop celery:pillowtop
-commcare-cloud <env> ec2-instance-state stop_and_start 10.201.11.133 10.201.11.134
+commcare-cloud <env> ec2 describe webworkers
+commcare-cloud <env> ec2 stop celery:pillowtop
+commcare-cloud <env> ec2 stop_and_start 10.201.11.133 10.201.11.134
 ```
 
 ##### Positional Arguments
--- a/docs/source/reference/1-commcare-cloud/commands.md
+++ b/docs/source/reference/1-commcare-cloud/commands.md
@@ -1744,6 +1744,13 @@ commcare-cloud <env> ec2-instance-state [--no-wait]
                                         {describe,start,stop,stop_and_start} inventory_group [inventory_group ...]
 ```
 
+`start` and `stop` show a check-mode preview of the state transitions
+and ask for confirmation before applying; `stop_and_start` asks for
+confirmation directly (a check-mode preview of a full cycle would show
+no net state change); `describe` just runs.
+
+When used with --control, this command skips the slow setup.
+To force setup, use --control-setup=yes instead.
 
 ##### Example
 
@@ -1766,7 +1773,7 @@ What to do to the matched instances.
 
 One or more inventory items to act on. Each can be a group
 (e.g. `webworkers`), an individual host name or private IP
-(e.g. `10.201.11.133`), or any valid ansible host pattern
+(e.g. `10.201.11.133`), or any ansible host pattern
 (e.g. `celery:pillowtop`).
 
 ##### Options
--- a/docs/source/reference/1-commcare-cloud/commands.md
+++ b/docs/source/reference/1-commcare-cloud/commands.md
@@ -1740,13 +1740,19 @@ have been made to our actual resources in AWS.
 Manage the EC2 instance state (start/stop/describe) of hosts in an AWS environment.
 
 ```
-commcare-cloud <env> ec2 [--no-wait] {describe,start,stop,stop_and_start} inventory_group [inventory_group ...]
+commcare-cloud <env> ec2 [--no-wait] [--unsafe]
+                         {describe,start,stop,stop_and_start} inventory_group [inventory_group ...]
 ```
 
 `start` and `stop` show a check-mode preview of the state transitions
 and ask for confirmation before applying; `stop_and_start` asks for
 confirmation directly (a check-mode preview of a full cycle would show
 no net state change); `describe` just runs.
+
+`stop` and `stop_and_start` will eventually stop the services running
+on the hosts gracefully before stopping the EC2 instances. Until that
+is implemented they stop the instances outright, without draining
+services first, so they must be confirmed by passing `--unsafe`.
 
 When used with --control, this command skips the slow setup.
 To force setup, use --control-setup=yes instead.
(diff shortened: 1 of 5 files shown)


src/commcare_cloud/commands/ec2_instance_state.py
```diff
@@ -0,0 +1,120 @@
...
+    `start` and `stop` show a check-mode preview of the state transitions
+    and ask for confirmation before applying; `stop_and_start` asks for
+    confirmation directly (a check-mode preview of a full cycle would show
+    no net state change); `describe` just runs.
+
+    Example:
+
+    ```
+    commcare-cloud <env> ec2-instance-state describe webworkers
+    commcare-cloud <env> ec2-instance-state stop celery:pillowtop
```
gherceg: Is `celery:pillowtop` a valid example of an ansible host pattern? This applies to line 52 as well.

src/commcare_cloud/commands/ec2_instance_state.py
```diff
@@ -0,0 +1,120 @@
...
+                "(no terraform config found for {!r})".format(environment.name))
+
+        instance_ids_by_host = get_instance_ids_by_host(environment, args.inventory_group)
+
+        all_hosts = {host.name for host in environment.inventory_manager.get_hosts('all')
+                     if not host.implicit}
+        if set(instance_ids_by_host) >= all_hosts:
+            raise CommandError(
+                "Refusing to run ec2-instance-state against all hosts in {!r}. "
+                "Target a specific group, host, or pattern instead.".format(environment.name))
```
gherceg: nit: would prefer a f-string here.

src/commcare_cloud/commands/ec2_instance_state.py
```diff
@@ -0,0 +1,120 @@
...
+        shared_args.QUIET_ARG,
+    )
+
+    def run(self, args, unknown_args):
+        ansible_context = AnsibleContext(args)
+        environment = ansible_context.environment
+        if not is_aws_env(environment):
+            raise CommandError(
+                "ec2-instance-state can only be used in AWS environments "
+                "(no terraform config found for {!r})".format(environment.name))
```
gherceg: nit: would prefer an f-string here

src/commcare_cloud/commands/ec2_instance_state.py
```diff
@@ -0,0 +1,120 @@
...
+            'ec2_instance_state_wait': not args.no_wait,
+        }
+
+        describe = args.action == 'describe'
+        return run_ansible_playbook(
+            'ec2_instance_state.yml',
+            ansible_context,
+            # describe is read-only; apply it directly without prompting
+            skip_check=args.skip_check or describe,
+            quiet=args.quiet or describe,
```
gherceg, line 110: Is there a world where you would want verbose output on a `describe`? Cause if so this would make that impossible right?

src/commcare_cloud/commands/ec2_instance_state.py
```diff
@@ -0,0 +1,120 @@
...
+    CommandBase,
+    CommandError,
+)
+from commcare_cloud.commands.terraform.aws import is_aws_env
+
+ACTIONS = ['describe', 'start', 'stop', 'stop_and_start']
+
+
+class Ec2InstanceState(CommandBase):
+    command = 'ec2-instance-state'
```
gherceg: What if we just called this command `ec2`? In the context of `cchq staging ec2 describe <host>` it seems fitting, but will defer to you if you are concerned that isn't specific enough.

docs/source/reference/1-commcare-cloud/commands.md
```diff
@@ -1735,6 +1735,49 @@ have been made to our actual resources in AWS.
 
 ---
 
+#### ``ec2-instance-state`` Command
+
+Manage the EC2 instance state (start/stop/describe) of hosts in an AWS environment.
+
+```
+commcare-cloud <env> ec2-instance-state [--no-wait]
+                                        {describe,start,stop,stop_and_start} inventory_group [inventory_group ...]
```
gherceg: this makes it look like `inventory_group` is a subcommand/required keyword parameter? Or am I misreading this?

src/commcare_cloud/commands/ec2_instance_state.py
AmitPhulera: Yes. Its same as using "," It is a [union operator](https://docs.ansible.com/projects/ansible/latest/inventory_guide/intro_patterns.html#common-patterns) in ansible.

src/commcare_cloud/commands/ec2_instance_state.py
AmitPhulera: I have thought about it, a two other options were `ec2_manager`, `ec2_operator` . 
I was unable to come to a conclusion and went with the existing module name.

`cchq staging ec2 describe <host>` do have a ring to it. But this means all the future ec2 api calls might got routed via this command which is fine(and probably a problem for future)

src/commcare_cloud/commands/ec2_instance_state.py
AmitPhulera: If I understood your question correctly - You are worried that quiet won't render the output of describe? If that is the concern - 


quiet here doesn't control output verbosity, it's only passed through to run_action_with_check_mode, where it's used to auto-answer the confirmation prompt (ask(...)). 

So forcing it on describe just means we skip the pointless "apply without running check first?" prompt for a read-only action.

docs/source/reference/1-commcare-cloud/commands.md
AmitPhulera: The documentation was autogenerated by the command - `manage-commcare-cloud make-docs > docs/source/reference/1-commcare-cloud/commands.md`. Which uses the command to extract out the documentation.

I stripped out 

> `start` and `stop` show a check-mode preview of the state transitions
    and ask for confirmation before applying; `stop_and_start` asks for
    confirmation directly (a check-mode preview of a full cycle would show
    no net state change); `describe` just runs.

from this doc because I felt it is not adding much value here. But because of that build [here](https://github.com/dimagi/commcare-cloud/actions/runs/27542855631/job/81408733623?pr=6915#step:6:2771) is failing.

Now coming to your question - I asked Claude and it said that - It is argparse's standard rendering of a positional with nargs='+'

src/commcare_cloud/commands/ec2_instance_state.py
AmitPhulera: Let me know how you think about it. I don't have any preference on it, I would be fine going by your recommendation.

src/commcare_cloud/commands/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6915/commits/cd5a167e92b32591570cf8e1902f1b268a069bb7

src/commcare_cloud/commands/ec2_instance_state.py
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6915/commits/cd5a167e92b32591570cf8e1902f1b268a069bb7

src/commcare_cloud/commands/ec2_instance_state.py
AmitPhulera: Your question raised one valid point regarding displaying the information on describe - So I added it in the role itself that after every execution it should print its state - https://github.com/dimagi/commcare-cloud/pull/6915/commits/6ddcbc52ee70659a261a31014299ad7a679e792c

src/commcare_cloud/commands/ec2_instance_state.py
gherceg: We offlined about this and it still felt a bit up in the air, though I think there is a preference for a simpler `ec2` command. Eventually the idea is to have commands like `safe_restart` or `safe_stop` which include additional steps like shutting down services prior to the instance stop command. However there are also times where we may want to do an `unsafe_restart` on a machine that OOMed during a fire for example. I talked to claude about naming for a bit, running the idea of passing an argument instead by it, like `cchq staging ec2 stop --unsafe` or something, and I actually like what it spit out.

It mentioned that `ec2` can serve as the commands for directly interacting with AWS, while something else, like `server` or `host` could act as the commcare-cloud layer that handles safely bringing services down first. Though server/host don't feel quite distinct enough from ec2, so I'm not sure exactly what it should be, but just throwing it out there.

AmitPhulera: Had a conversation with @millerdev last week and decided to go with `ec2` as command name  [10a5a0c](https://github.com/dimagi/commcare-cloud/pull/6915/commits/10a5a0cceda55e5e634623621e9ed61b6c5ce086) and now we are calling `stop` and `stop_and_start` with --unsafe option only. The safer versions to be implemented soon.

AmitPhulera:
@gherceg The comments have been addressed and this is ready for review again.

gherceg (approved):

PR #6922, opened 2026-06-19
Remove CitusDB automation

millerdev:
CitusDB (the sharded-PostgreSQL extension formerly used by ICDS) is no longer in use. This branch removes all of its install/management automation and untangles its coupling with the regular PostgreSQL codepath, so that DB hosts, pgbouncer routing, pg_hba ACLs, monitoring, and the pg_upgrade flow are handled uniformly with no citus special-casing. The now-unused `citusdb`, `citusdb_master`, and `citusdb_worker` inventory groups are dropped, along with the unused pgbouncer client-TLS ("certificate auth") automation that existed only for citus worker connections.

https://dimagi.atlassian.net/browse/SAAS-19889

##### Environments Affected
None

##### Announce New Release

No changelog entry is included: CitusDB was not used outside of Dimagi-managed environments. The only public documentation we had was for upgrading it. Othrewise it was undocumented, and required special app-specific configuration to be of any benefit.


src/commcare_cloud/manage_commcare_cloud/monitors/agent_check.yml
```diff
@@ -73,6 +73,6 @@ options:
   thresholds: {critical: 1, ok: 1, warning: 1}
   timeout_h: 0
   include_tags: true
-query: '"datadog.agent.up".over("*").exclude("group:pgsynclog","environment:enikshay","host:citus0.india.commcarehq.org","host:citus1.india.commcarehq.org","host:citus2.india.commcarehq.org").by("host").last(2).count_by_status()'
+query: '"datadog.agent.up".over("*").exclude("group:pgsynclog","environment:enikshay").by("host").last(2).count_by_status()'
```
gherceg, line 76: Outside the scope of this PR, but while editing this could also remove the enikshay reference https://github.com/dimagi/commcare-cloud/pull/6362

gherceg (approved):
One non-blocking comment that could be handled separately.

PR #6924, opened 2026-06-24
Remove RabbitMQ support from commcare-cloud

millerdev:
Removes all RabbitMQ support from commcare-cloud. RabbitMQ was previously used as the Celery message broker; all environments have since migrated to Redis. This cleans up the dead code and configuration left behind.

https://dimagi.atlassian.net/browse/SAAS-19890

:blowfish: Review by commit.

##### Environments Affected

None.


docs/source/services/rabbitmq/migrate-to-redis.rst
gherceg, line 1: We should keep this around just in case self hosters haven't yet moved off of rabbit since the changelog references this file specifically. Merging the rest of this PR should be a forcing function to migrate, so it seems reasonable to remove this by the end of the year at the latest.

gherceg (commented):
Thank you! One comment about the migration doc which I think we should keep around for a bit longer.

docs/source/services/rabbitmq/migrate-to-redis.rst
millerdev: https://github.com/dimagi/commcare-cloud/pull/6924/commits/2a0809efab0a8e6f6f9361850bdd6c3ac0662f12

gherceg (approved):

PR #6926, opened 2026-06-26
Auto-analyze dependency-update PRs with Claude

AmitPhulera:
##### Environments Affected
None

## Summary
Adds a GitHub Actions workflow that runs a Claude analysis on dependency-update PRs and posts a single review comment summarizing changelogs, breaking changes, migration notes, and likely codebase impact under `src/commcare_cloud/`.

This mirrors the equivalent workflow in commcare-hq, adapted to commcare-cloud's stack (uv for Python, plus the github-actions ecosystem; no JS).

## How it triggers
- Automatically on any PR opened by `dependabot[bot]` (including the security-update PRs we already receive).
- Manually from the **Actions** tab via `workflow_dispatch`, passing a PR number — so we can run it on hand-written dependency-upgrade PRs too.

No `dependabot.yml` is added: this only consumes existing Dependabot PRs and manual triggers, and does not change our dependency-update policy.

## Requirements
The `ANTHROPIC_API_KEY` repo secret must be configured for the workflow to run.

### DIFF
--- a/.github/workflows/claude-dependabot.yml
+++ b/.github/workflows/claude-dependabot.yml
@@ -46,7 +46,7 @@ jobs:
           echo "ref=$REF" >> $GITHUB_OUTPUT
 
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
         with:
           ref: ${{ steps.pr.outputs.ref }}
           fetch-depth: 1
--- a/.github/workflows/claude-dependabot.yml
+++ b/.github/workflows/claude-dependabot.yml
@@ -1,0 +1,141 @@
+# Claude Dependabot PR Review Workflow
+#
+# Automatically runs Claude analysis on Dependabot PRs to:
+# - Identify dependency changes and their versions
+# - Look up changelogs for updated packages
+# - Assess breaking changes and security impacts
+# - Provide actionable recommendations for the development team
+#
+# Covers the dependabot ecosystems used by commcare-cloud: uv (Python) and github-actions
+#
+# Requirements: ANTHROPIC_API_KEY secret must be configured
+
+name: Claude Dependabot PR Review
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number to analyze'
+        required: true
+        type: string
+
+jobs:
+  dependabot-review:
+    if: github.actor == 'dependabot[bot]' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-dependabot.yml
```diff
@@ -0,0 +1,141 @@
...
+#
+# Covers the dependabot ecosystems used by commcare-cloud: uv (Python) and github-actions
+#
+# Requirements: ANTHROPIC_API_KEY secret must be configured
+
+name: Claude Dependabot PR Review
+
+on:
+  pull_request:
+    types: [opened, synchronize]
```
gherceg: What is the rationale for having `synchronize` here? I know we have this set in hq as well, but same question probably applies for me there. Dependabot opens a new PR if a new version is released, and I don't think we want Claude to do a full analysis if we push a commit to an existing PR to resolve a breaking change right? Or am I missing something here?

.github/workflows/claude-dependabot.yml
```diff
@@ -0,0 +1,141 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          PR_NUMBER="${{ inputs.pr_number || github.event.pull_request.number }}"
+          echo "pr_number=$PR_NUMBER" >> $GITHUB_OUTPUT
+          REF=$(gh pr view "$PR_NUMBER" --repo "${{ github.repository }}" --json headRefName --jq '.headRefName')
+          echo "ref=$REF" >> $GITHUB_OUTPUT
+
+      - name: Checkout code
+        uses: actions/checkout@v4
```
gherceg: I think the latest version of checkout is v7.

.github/workflows/claude-dependabot.yml
AmitPhulera: Good point. I thought earlier that it might be useful when we push fixup commits, but your point makes sense. Nothing major would probably change in the commits that we will make so a re-review would not make sense even if it does we can manually trigger it. Updating it.

.github/workflows/claude-dependabot.yml
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6926/commits/24b0972a708f4ceb7b1844a66a5cd2ab6f54d320

.github/workflows/claude-dependabot.yml
AmitPhulera: https://github.com/dimagi/commcare-cloud/pull/6926/commits/d2884c9f0b469fe867e45dce3a324d0c6d701810

gherceg (approved):

PR #6927, opened 2026-06-29
Add ansible-lint to CI

millerdev:
Introduces `ansible-lint` to commcare-cloud so playbook and role regressions are caught automatically in CI, without forcing a cleanup of the large existing backlog of lint findings. Follow-up PR(s) will fix pre-existing lint errors to pare down `.ansible-lint-ignore`.

Also folds the `test` dependency group into `dev` so a single sync gives developers everything.

I'm considering making this check a required pass on Github Actions before a PR can be merged. Any objections?

:blowfish: Review by commit.

https://dimagi.atlassian.net/browse/SAAS-19958

##### Environments Affected
None


pyproject.toml
```diff
@@ -41,6 +41,7 @@ manage-commcare-cloud = "commcare_cloud.manage_commcare_cloud.manage_commcare_cl
 
 [dependency-groups]
 test = [
+    "ansible-lint==5.4.0",
```
AmitPhulera, line 44: Is pinning required because we are on a outdated ansible version?

.github/workflows/ansible-lint.yml
```diff
@@ -0,0 +1,35 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  ansible-lint:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v5
```
AmitPhulera, line 18: nit: v7 is the latest version.

```suggestion
    - uses: actions/checkout@v7
```

AmitPhulera (approved):
> I'm considering making this check a required pass on Github Actions before a PR can be merged. Any objections?

No objections from my end.

src/commcare_cloud/ansible_lint.py
```diff
@@ -0,0 +1,133 @@
+"""Run ansible-lint on commcare-cloud playbooks and roles.
+
+Runs ansible-lint over every playbook and role and fails only on
+``<file> <rule>`` violations that are not already recorded in the committed
+``.ansible-lint-ignore`` file::
+
+    ansible-lint-cloud                    # check against .ansible-lint-ignore
+    ansible-lint-cloud --generate-ignore  # regenerate .ansible-lint-ignore
+
+This backports the `.ansible-lint-ignore` workflow from ansible-lint 6.2+ onto
```
gherceg, line 10: "This" is referring to this entire file right? And this entire file can be removed once `ansible-lint` is upgraded to >= 6.2?

src/commcare_cloud/ansible_lint.py
```diff
@@ -0,0 +1,133 @@
...
+# ANSIBLE_DIR is <repo>/src/commcare_cloud/ansible; the repo root is three
+# levels up. Lint runs from there so reported paths (and the ignore file) are
+# repo-relative.
+REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(ANSIBLE_DIR)))
+CONTENT_DIR = os.path.relpath(ANSIBLE_DIR, REPO_ROOT)
+IGNORE_FILE = os.path.join(REPO_ROOT, ".ansible-lint-ignore")
+IGNORE_REL = os.path.relpath(IGNORE_FILE, REPO_ROOT)
+
+
+def lint_env():
```
gherceg, line 36: nit: I get that this is the env for linting, but the name `lint_env` sounds like this function would be the one actually linting. Not a big deal though.

gherceg (approved):
Also no objections to blocking merges. I was going to leave comments about f-strings but given it seems src/commcare_cloud/ansible_lint.py will be short-lived anyway, it's fine. I also don't know how strongly people feel about f-strings. Other comments are also non-blocking.

pyproject.toml
millerdev: Yes.

.github/workflows/ansible-lint.yml
millerdev: Thanks. I'll fix that as well as [the one in tests.yml](https://github.com/dimagi/commcare-cloud/blob/0444f9c2c4c030e3d784ef201a7d5e6e0652fed2/.github/workflows/tests.yml#L29) in a follow-up PR.

src/commcare_cloud/ansible_lint.py
millerdev: Correct.

src/commcare_cloud/ansible_lint.py
millerdev: Oh, true, I didn't think about the possibility that this name could be a verb. I'm going to leave it since it will hopefully be short-lived.

PR #6930, opened 2026-07-02
Fix ansible-lint risky-file-permissions with explicit modes

millerdev:
Ansible-lint's `risky-file-permissions` rule flags any `file`, `copy`, or `template` task that doesn't set an explicit `mode`, since the resulting permissions otherwise depend on the umask of the machine running Ansible. This PR works through the roles that were suppressing that rule in `.ansible-lint-ignore` and gives each task an explicit mode, removing the corresponding ignore entries.

Most changes simply pin the mode Ansible was already applying by default (e.g. `0644` for config files, `0755` for directories), so there is no behavior change. A handful of tasks also tighten permissions and/or add explicit `owner`/`group` where the existing default was looser than it should be for sensitive files – these were split into separate commits with explicit messaging on how the risks handled.

https://dimagi.atlassian.net/browse/SAAS-19980

:blowfish: Review by commit.

##### Environments Affected
All.

`deploy-stack` has been run on staging as part of testing for this PR.

##### Announce New Release

- [x] (Dimagi only) After merging, I will follow [these instructions](https://confluence.dimagi.com/display/saas/Announcing+changes+affecting+third+parties#Announcingchangesaffectingthirdparties-announcing) to announce a new commcare-cloud release.


docs/source/changelog/0100-explicit-file-permissions.md
```diff
@@ -63,6 +63,11 @@ This covers the common/ssh/datadog/lpar2rrd roles, postgresql/pgbouncer,
 elasticsearch, kafka/zookeeper, commcarehq, couchdb2 (including its
 keepalived/haproxy setup), nginx, http_proxy, cloudwatch_logs, and pg_repack.
 
+Elasticsearch note: `--skip-tags=es_restart` prevents a full cluster restart
+which could cause downtime and may require cluster maintenance to reallocate
+replica shards. Use a safe (one node at a time) restart procedure if your ES
+nodes need to be restarted after `deploy-stack` has completed.
```
gherceg, line 69: Is it still safe to restart each ES node, even if one at a time, if cluster routing is disabled? I'm imagining accumulating unassigned shards. I suppose our docs on ES mention cluster routing, but should that also be mentioned here?

gherceg (commented):
Is it true that some of the services impacted by this PR are not actively used by us, but could be essential for self hosters? Primarily thinking of keepalived, and am curious what verification was done to ensure that still behaves as expected with the file permission changes there?

docs/source/changelog/0100-explicit-file-permissions.md
millerdev: The suggested "one node at a time" is a simplification of the [rolling restart](https://www.elastic.co/docs/deploy-manage/maintenance/start-stop-services/full-cluster-restart-rolling-restart-procedures#restart-cluster-rolling) procedure outlined in ES docs; the other operative word in that sentence is _safe_, which implies doing other steps necessary to allow nodes to transition from yellow to green after they are restarted. Reiterating the whole restart process is out of scope for this document, especially since it is recommending to skip the restart entirely.

millerdev:
@gherceg 

> Is it true that some of the services impacted by this PR are not actively used by us, but could be essential for self hosters?

Yes, and I'm not sure what to do about that. I don't like when we have configuration for services that we do not use. I wonder if `keepalived` in particular has not been used since ICDS, and maybe we should remove it? Do you have a suggestion of a better approach?

> Primarily thinking of keepalived, and am curious what verification was done to ensure that still behaves as expected with the file permission changes there?

No explicit verification was done other than to inspect way it is configured and run. As noted in the commit message:

_The keepalived daemon runs as root; 0640 removes world-read of the VRRP auth_pass._

This [seems correct](https://github.com/dimagi/commcare-cloud/blob/fd795a3c8c5d4e40742e66db8b2e5bf38e363b4b/src/commcare_cloud/ansible/roles/keepalived/defaults/main.yml#L12), and looks safe to me.

gherceg (approved):
Offlined with Daniel and feel that the changes made here to services we don't use in our own stack appear safe to move forward with.

PR #6931, opened 2026-07-06
Clean up ansible-lint violations

millerdev:
This branch works through the ansible-lint findings across the `commcare-cloud` playbooks and fixes or annotates each one, rather than leaving them all suppressed in the ignore file. Most violations are fixed outright. The remaining violations that are intentional get an inline `# noqa: <rule>` with a comment explaining why.

`.ansible-lint-ignore` itself is cut down to only the handful of exceptions that cannot be expressed as an inline `# noqa`, and now carries a generated header explaining that constraint so it doesn't grow back into a dumping ground.

https://dimagi.atlassian.net/browse/SAAS-19980

:blowfish: Review by commit.

##### Environments Affected
None — this is a lint/cleanup pass over role tasks and handlers, not environment-specific configuration.


src/commcare_cloud/ansible/roles/nginx/tasks/reload.yml
```diff
@@ -23,10 +23,11 @@
...
 - name: Assert nginx is running
   become: yes
-  command: 'service nginx status'
-  args:
-    warn: no
+  # `systemctl status` is exempt from ansible-lint's command-instead-of-module
+  # rule but `is-active` is not, hence the noqa. is-active is used because it
+  # gives a precise active/inactive result instead of substring-matching on
+  # human-readable status output.
+  command: systemctl is-active nginx  # noqa: command-instead-of-module
```
gherceg, line 30: Just curious if this is something ansible would fix in the future or if this is intended behavior of the linter? I don't have a great grasp of why systemctl status would be exempt but systemctl is-active would not.

src/commcare_cloud/ansible/roles/aws-efs/tasks/main.yml
```diff
@@ -7,6 +7,8 @@
     fail_msg: "efs_mount_uid={{ efs_mount_uid }} efs_mount_gid={{ efs_mount_gid }} efs_mount_dir={{ efs_mount_dir }}"
     success_msg: "efs_mount_uid={{ efs_mount_uid }} efs_mount_gid={{ efs_mount_gid }} efs_mount_dir={{ efs_mount_dir }}"
 
+# load-failure (see .ansible-lint-ignore): This cross-role import resolves
+# relative to the playbook dir; ansible-lint resolves relative to the role root.
```
gherceg, line 11: I assume there is not a way to tell ansible-lint to resolve to the playbook dir?

gherceg (approved):

src/commcare_cloud/ansible/roles/nginx/tasks/reload.yml
millerdev: I agree and hope a future version of Ansible will fix this.

> why systemctl status would be exempt but systemctl is-active would not

My guess is it's just an oversight, and there is no technical reason.

src/commcare_cloud/ansible/roles/aws-efs/tasks/main.yml
millerdev: I don't know, but I hope the new version of ansible-lint that we will get later in the Ansible upgrade will fix the issue.

gherceg (approved):

PR #6935, opened 2026-07-13
[Part - 1] : Use prebuilt static files on deploys - Add util to download static files

AmitPhulera:
https://dimagi.atlassian.net/browse/SAAS-20027

We already build static files on Github actions everytime something is pushed to `autostaging` branch on CommCare HQ repo.

This PR moves a step in the direction of using prebuilt static files in deploys. Earlier we were considering it to be useful for third party hosters and were trying to figure out the artifacts public because currently you need a token with read access to commcare hq repo to download the artifacts.

@gherceg and I recently started talking about it again and we decided why not just test it internally with the actions workflow and using the same token that we pass to compile deploy summary to get the artifact. This is the first PR that tries to setup basic structure. In the followup PRs we would be using the utils introduced in this PR to be integrated into deploy playbooks. 

##### Environments Affected

Would be testing on staging for now, but will be used for all Dimagi environments.

Review by commit 🐡 

36df7be5f8299a87c05783a509e5b3487c163e03 is the entire spec on how this entire feature is planned. 
4c432fcff83306d161ccfe0c3cc19fd334184838 is what is in this PR.

I am intentionally keeping them around until the feature is integrated, would remove them later.

### DIFF
--- a/src/commcare_cloud/commands/deploy/command.py
+++ b/src/commcare_cloud/commands/deploy/command.py
@@ -59,6 +59,13 @@ class Deploy(CommandBase):
             Generate new localsettings.py rather than copying from the previous
             release.
         """),
+        Argument('--prebuilt-static', action='store_true', help="""
+            Fetch the prebuilt static files artifact for the deployed commit
+            from GitHub Actions instead of relying solely on the on-host
+            build. Requires a GitHub token (GITHUB_TOKEN env var or
+            config.py) with Actions read access on dimagi/commcare-hq.
+            Falls back to the on-host build if the artifact is unavailable.
+        """),
         shared_args.QUIET_ARG,
         shared_args.BRANCH_ARG,
     )
@@ -128,6 +135,7 @@ def _warn_about_non_formplayer_args(args):
         '--commcare-rev',
         '--ignore-kafka-checkpoint-warning',
         '--update-config',
+        '--prebuilt-static',
     ]:
         dest = arg.lstrip("-").replace("-", "_")
         if getattr(args, dest) not in [None, False]:
--- a/src/commcare_cloud/commands/deploy/commcare.py
+++ b/src/commcare_cloud/commands/deploy/commcare.py
@@ -5,6 +5,9 @@ from datetime import datetime, timedelta
 from commcare_cloud.alias import commcare_cloud
 from commcare_cloud.cli_utils import ask
 from commcare_cloud.colors import color_error, color_notice, color_summary
+from commcare_cloud.commands.ansible.ansible_playbook import (
+    run_ansible_playbook,
+)
 from commcare_cloud.commands.ansible.run_module import (
     AnsibleContext,
     BadAnsibleResult,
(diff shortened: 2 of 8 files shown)


src/commcare_cloud/commands/deploy/static_artifact.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def _headers(token):
+    return {
+        "Authorization": f"Bearer {token}",
+        "Accept": "application/vnd.github+json",
+    }
+
+
+def main(argv=None):
```
dannyroberts, line 145: Nit: the newspaper principal / vertical ordering would have us put this at the top of the file, rather than the bottom.

src/commcare_cloud/commands/deploy/static_artifact.py
```diff
@@ -0,0 +1,174 @@
+"""Fetch the prebuilt REQUIRED_STATIC_FILES.zip GitHub Actions artifact.
```
dannyroberts, line 1: The first thing that pops into my head reading through this is: could this module be written much more concisely with `PyGithub` instead of raw `requests`?

src/commcare_cloud/commands/deploy/commcare.py
```diff
@@ -97,6 +97,25 @@ def deploy_commcare(environment, args, unknown_args):
     return 0
 
 
+def get_prebuilt_static_args(args):
+    """Ansible -e args enabling the prebuilt static artifact fetch, or [].
+
+    Verifies that Github token is present and sets the GITHUB_TOKEN env var. This just passes the arg to tell playbook to use the prebuilt static artifacts.The artifact itself is fetched later, by a playbook task on the control machine.
```
dannyroberts, line 103: Probably worth having Claude do a pass on tightening up comments, to catch things like this formatting error.

dannyroberts (commented):
Had a couple thoughts, so will wait to see what you think before reviewing further. Overall, I like the direction!

src/commcare_cloud/commands/deploy/static_artifact.py
```diff
@@ -0,0 +1,174 @@
...
+
+ARTIFACTS_URL = "https://api.github.com/repos/dimagi/commcare-hq/actions/artifacts"
+RUNS_URL = ("https://api.github.com/repos/dimagi/commcare-hq"
+            "/actions/workflows/build-static.yml/runs")
+INNER_ZIP_NAME = "REQUIRED_STATIC_FILES.zip"
+POLL_TIMEOUT = 1800
+POLL_INTERVAL = 30
+REQUEST_TIMEOUT = 60
+PROGRESS_INTERVAL = 10  # seconds between download progress lines
+RUNNING_STATUSES = ("queued", "in_progress", "pending", "waiting", "requested")
```
gherceg, line 25: Are there too many statuses? What is the difference between in_progress, pending, and waiting?

src/commcare_cloud/commands/deploy/command.py
gherceg, line 1: We talked about it offline, but to me this belongs in the next PR that actually makes use of it. I know you saw it as this PR setting up all of the boilerplate. Just wanted to make a note of this discussion.

PR #6941, opened 2026-07-13
Flush ansible cache before update-supervisor-confs

gherceg:
Without this, unexpected behavior with update-supervisor-confs can occur where an expected change is not applied to the configuration because ansible is operating on a stale cache.

Tested on staging just to make sure `update-supervisor-confs` works as expected and it does. It seems tough to test this against the actual bug itself since it isn't easy to reproduce (not sure how to get a stale cache).

##### Environments Affected

None

### DIFF
--- a/src/commcare_cloud/commands/ansible/ansible_playbook.py
+++ b/src/commcare_cloud/commands/ansible/ansible_playbook.py
@@ -387,7 +387,7 @@ class UpdateSupervisorConfs(_AnsiblePlaybookAlias):
 
     def run(self, args, unknown_args):
         args.playbook = 'deploy_stack.yml'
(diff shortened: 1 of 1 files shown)


dannyroberts (approved):
I think the only concern would be if whatever was being cached takes a long time to retrieve. My guess is not long enough that it's noticeable for a command that already takes a minute or two to run

AmitPhulera (approved):

gherceg:
> I think the only concern would be if whatever was being cached takes a long time to retrieve. My guess is not long enough that it's noticeable for a command that already takes a minute or two to run

Yeah something to keep in mind.

PR #6949, opened 2026-07-17
Allow tuning gunicorn worker recycling without an HQ deploy

dannyroberts:
Follow-up to https://github.com/dimagi/commcare-hq/pull/37927, which raised gunicorn's `max_requests` from 240 to 2000 and `graceful_timeout` from 300s to 900s. The values live in commcare-hq's `gunicorn_conf.py`, so adjusting them requires a full HQ deploy — and they do need adjusting: an hq_webworker OOMed within a day of the increase, so worker memory grows with age more than expected. I'm [reverting](https://github.com/dimagi/commcare-hq/pull/37935) that (commcare-hq) PR and replacing with this (commcare-cloud) PR.

This moves both settings into `app-processes.yml` by passing them as gunicorn command-line arguments in the supervisor config. Gunicorn gives CLI arguments precedence over the config file, so these cleanly override the values baked into the HQ release. Operators can now tune them with:

```
cchq <env> update-supervisor-confs
```

and the rolling webworker restart it offers — no HQ deploy needed.

Two commits:
1. The plumbing (schema, defaults, supervisor template). Defaults are the long-standing pre-#37927 values (240 / 300s), so environments opt in to less frequent recycling explicitly. `--max-requests-jitter` must be passed explicitly since the jitter computed inside `gunicorn_conf.py` derives from that file's own `max_requests`, not the CLI value.
2. Production override: `gunicorn_max_requests: 700` and `gunicorn_graceful_timeout: 900` — recycling ~3x less often than the old 240 (rather than #37927's 8x, which OOMed), with the drain window matched to the 900s request budget nginx allows so recycling can't kill requests that would otherwise complete.

##### Environments Affected

All environments get explicit defaults equal to the long-standing pre-#37927 behavior — which means environments running an HQ release containing #37927 revert to 240/300 when their supervisor confs are next updated; only production keeps the less aggressive recycling (700/900). If other environments should share production's values, that's a per-environment follow-up.

### DIFF
--- a/src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_django.conf.j2
+++ b/src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_django.conf.j2
@@ -11,6 +11,9 @@ command={{ app_processes_config.django_command_prefix }}
     -c deployment/gunicorn/gunicorn_conf.py
     -k gevent
     --workers {{ item.gunicorn_workers }}
+    --max-requests {{ app_processes_config.gunicorn_max_requests }}
+    --max-requests-jitter {{ app_processes_config.gunicorn_max_requests // 2 }}
+    --graceful-timeout {{ app_processes_config.gunicorn_graceful_timeout }}
     --bind 0.0.0.0:{{ django_port }}
     --log-file {{ log_home }}/{{ project }}.gunicorn.log
     --log-level debug
--- a/environments/production/app-processes.yml
+++ b/environments/production/app-processes.yml
@@ -1,5 +1,9 @@
 gunicorn_workers_static_factor: 0
 gunicorn_workers_factor: 2
+# graceful_timeout matches the 900s request budget nginx allows, so that
+# worker recycling can't kill an in-flight request that would have completed
+gunicorn_max_requests: 700
+gunicorn_graceful_timeout: 900
 formplayer_memory: "31g"
 formplayer_g1heapregionsize: "16m"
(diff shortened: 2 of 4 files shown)


gherceg (approved):

src/commcare_cloud/ansible/roles/commcarehq/templates/supervisor_django.conf.j2
```diff
@@ -11,6 +11,9 @@ command={{ app_processes_config.django_command_prefix }}
     -c deployment/gunicorn/gunicorn_conf.py
     -k gevent
     --workers {{ item.gunicorn_workers }}
+    --max-requests {{ app_processes_config.gunicorn_max_requests }}
+    --max-requests-jitter {{ app_processes_config.gunicorn_max_requests // 2 }}
+    --graceful-timeout {{ app_processes_config.gunicorn_graceful_timeout }}
```
gherceg, line 16: Sorry silly question. If these are not defined, will this command fail? Should the inclusion of each flag be conditional on whether it is defined in app processes?

dannyroberts: Not a silly question. But I think it's correct as it is. The reason is that these should _always_ be defined, because they're defined in [environmental-defaults/app-processes.yml‎](https://github.com/dimagi/commcare-cloud/pull/6949/changes#diff-538bd5127fa33bb837c1b09b86915595bc5dc5deb126427f171bd77afccd369a).

gherceg: Oh yeah duh. I panicked after approving and did not internalize the defaults. Thanks

PR #6951, opened 2026-07-20
Provision standalone DB for ProjectDB

esoergel:
From discussion on [ProjectDB Security and Infrastructure Approach Review
](https://docs.google.com/document/d/1ayuYQx7DD6dfc2-zzeLat4R8kXZkTp6WlPcD4rcw0YU/edit?tab=t.8pn5qhzhr6br)

This adds some scaffolding such that environments can provision a separate PostgreSQL database for ProjectDB, and turns it on for staging.  In conjunction with https://github.com/dimagi/commcare-hq/pull/37938, this should make it such that development and test environments have ProjectDB tables created in the `default` postgres database, but environments managed by `commcare-cloud` will error if it isn't explicitly provisioned.

This also handles creating the necessary extensions.  The features that build on these extensions are coming in a different PR: https://github.com/dimagi/commcare-hq/pull/37932 

##### Environments Affected
Staging only, for now.  Will test there and do prod separately later

[Staging rollout plan](https://docs.google.com/document/d/1LICyhHI5ogHhItz-ouGrb4t0T0lb0LSqOljvjguJmwk/edit?tab=t.0#heading=h.fkbc9acm8x0)

### DIFF
--- a/docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
+++ b/docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
@@ -68,6 +68,25 @@ Configuration for the db that formplayer uses
 
 Configuration for the db that UCRs are copied into.
 
+``project_db``
+^^^^^^^^^^^^^^^^^^
+
+
+* Type: `db config <#db-config-type>`_
+* Default: not created (opt-in)
+* Default values (when enabled):
+  .. code-block:: yaml
+
+     django_alias: project_db
+     name: commcarehq_project_db
+     django_migrate: False
+
+Optional db that stores ProjectDB tables. It is only created in environments
+that add a ``project_db`` block to `\ ``dbs`` <#dbs>`_ and map it in `\
+``REPORTING_DATABASES`` <#reporting-databases>`_. Its ``host`` and pgbouncer
+settings default to those of the ``ucr`` db when not set explicitly. Installs
+required extensions automatically as part of database setup.
+
 ``synclogs``
 ^^^^^^^^^^^^^^^^
 
@@ -335,7 +354,13 @@ without a different host explicitly set in `\ ``dbs`` <#dbs>`_.
 * Default: ``{"ucr": "ucr"}``
 * Status: Custom
 
-Specify a mapping of UCR "engines".
+Specify a mapping of Reporting DB "engines" (UCR, ProjectDB).
+
+The default maps only the ``ucr`` engine. Environments that enable the
+optional `\ ``project_db`` <#project-db>`_ database override this to also map
+the ``project_db`` engine, e.g. ``{"ucr": "ucr", "project_db": "project_db"}``.
+Note that setting this value replaces the whole mapping rather than merging,
+so the ``ucr`` entry must be re-listed.
 
 The keys define engine aliases, and can be anything. The values are either
 
--- a/docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
+++ b/docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
@@ -356,11 +356,9 @@ without a different host explicitly set in `\ ``dbs`` <#dbs>`_.
 
 Specify a mapping of Reporting DB "engines" (UCR, ProjectDB).
 
-The default maps only the ``ucr`` engine. Environments that enable the
-optional `\ ``project_db`` <#project-db>`_ database override this to also map
-the ``project_db`` engine, e.g. ``{"ucr": "ucr", "project_db": "project_db"}``.
-Note that setting this value replaces the whole mapping rather than merging,
-so the ``ucr`` entry must be re-listed.
+Setting this replaces the default (``{"ucr": "ucr"}``) entirely rather than
+merging into it, so when you add an engine (e.g. `\ ``project_db``
+<#project-db>`_) you must re-list ``ucr`` or it will be dropped.
 
 The keys define engine aliases, and can be anything. The values are either
 
--- a/environments/staging/postgresql.yml
+++ b/environments/staging/postgresql.yml
@@ -1,5 +1,9 @@
 DEFAULT_POSTGRESQL_HOST: rds_pg0
 
+REPORTING_DATABASES:
+  ucr: ucr
+  project_db: project_db
+
 postgres_override:
   postgresql_max_connections: 300
   postgresql_version: '14'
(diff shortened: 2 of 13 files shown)


src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
```diff
@@ -146,3 +146,20 @@
...
+    db: '{{ postgresql_dbs.project_db.name }}'
+    port: "{{ postgresql_port }}"
+    login_host: "{{ db_is_remote and postgresql_host or omit }}"
+    login_user: "{{ db_is_remote and postgres_users.root.username or omit }}"
+    login_password: "{{ db_is_remote and postgres_users.root.password or omit }}"
+    query: |
+      CREATE EXTENSION IF NOT EXISTS cube;
+      CREATE EXTENSION IF NOT EXISTS earthdistance;
+      CREATE EXTENSION IF NOT EXISTS pg_trgm;
+      CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;
```
esoergel, line 163: These are created here for managed environments, and will be created by commcare-hq code for DEBUG and test contexts

src/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -337,6 +350,12 @@ class UcrDBOptions(DBOptions):
     django_migrate = False
 
 
+class ProjectDbDBOptions(DBOptions):
+    name = constants.project_db_name
+    django_alias = 'project_db'
+    django_migrate = False
+
```
dannyroberts, line 357: In live code review, @esoergel suggested adding `create = False` here so that this db wouldn't be created on non-target environments.

environments/staging/postgresql.yml
```diff
@@ -35,6 +39,8 @@ dbs:
     pgbouncer_hosts:
       - pgproxy6
     query_stats: True
+  project_db:
+    create: True
```
dannyroberts: ... or remove this and replace with just

```suggestion
  project_db: {}
```

dannyroberts (approved):
Looks good, left a note from our live review, but not blocking

docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
```diff
@@ -335,7 +354,13 @@ without a different host explicitly set in `\ ``dbs`` <#dbs>`_.
...
 * Status: Custom
 
-Specify a mapping of UCR "engines".
+Specify a mapping of Reporting DB "engines" (UCR, ProjectDB).
+
+The default maps only the ``ucr`` engine. Environments that enable the
+optional `\ ``project_db`` <#project-db>`_ database override this to also map
+the ``project_db`` engine, e.g. ``{"ucr": "ucr", "project_db": "project_db"}``.
+Note that setting this value replaces the whole mapping rather than merging,
+so the ``ucr`` entry must be re-listed.
```
millerdev: Where is the original listing that must be "re-listed" here? I'm not understanding this last note sentence.

environments/staging/postgresql.yml
```diff
@@ -35,6 +39,7 @@ dbs:
     pgbouncer_hosts:
       - pgproxy6
     query_stats: True
+  project_db: {}
```
millerdev, line 42: Does this have to be set? It's a little weird to set it to an empty dict, which means "use the defaults." Put another way, is it too magical to use the defaults if this line were removed?

millerdev (approved):

environments/staging/postgresql.yml
esoergel: As written, this _does_ have to be set.  The predicate in `set_up_dbs.yml` looks like this:

```yaml
  when: postgresql_dbs.project_db and postgresql_dbs.project_db.create
        and postgresql_host == postgresql_dbs.project_db.host
```

If `project_db` is not defined, that first `postgresql_dbs.project_db` is falsy.  This is what makes project_db on for staging and not for other environments.  You can verify this like so:

```bash
$ cchq staging list-postgresql-dbs | grep project_db
        - commcarehq_project_db
$ cchq production list-postgresql-dbs | grep project_db
```

I removed the create in this last commit because it felt misleading.  The actual gate was the presence of a `project_db` block in `postgresql.yml` - the base `DBOptions` defaults `create = True` anyways.  I suppose another option here would be to set `create=False` on `ProjectDbDBOptions`.  Then environments would need to define project_db _and_ enable `create`.  That does feel a bit more explicit - what do you think?

docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
esoergel: I can reword this.  If you don't specify `REPORTING_DATABASES`, you get the default value of `{"ucr": "ucr"}`, but if you _do_ specify it, you need to manually include UCR.  
```yml
REPORTING_DATABASES:
  project_db: project_db
```
^ This would result in not getting a separate DB for UCR, effectively overriding the default.

environments/staging/postgresql.yml
millerdev: Is `create` needed? Can it be _true_ if `project_db` is defined, and _false_ if not?

This would change the predicate to something like
```yml
when: postgresql_dbs.project_db is defined
      and postgresql_host == postgresql_dbs.project_db.host
```

`postgresql_dbs.project_db.host` must exist if `project_db` is defined.

esoergel: `create` is defined by the parent class, `DBOptions`.  Looks like it's only set to `False` on the `StandbyDBOptions` class, since standby dbs are created independently of this (I think).  I do think that maybe I should just ignore the field entirely here as you suggest.  Though FWIW, `is defined` wouldn't work here, as configs that exclude the field have it defined but set to `null`, as seen by the test updates in this PR.  Pushing a commit that removes `create` from the predicate.

docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
millerdev: Are you planning to reword it in this PR or a follow-up?

esoergel: oh yeah forgot about that, thanks for the nudge.  Just pushed an update

millerdev (approved):

PR #6955, opened 2026-07-28
Modernize the Vagrant dev environment and add libvirt support

millerdev:
This restores the Vagrant-based dev environment to a working, easier-to-maintain state. It replaces the outdated Ubuntu box, fixes the permission, service-limit, and dependency-install errors that broke provisioning, and adds support for contributors who use `libvirt` instead of VirtualBox.

Default service versions (Elasticsearch, CouchDB, HAProxy) are brought up to date. Control-node provisioning has been reworked to bind-mount the repo rather than symlink it, which resolves several errors seen when initializing `commcare-cloud` inside the VM. Fixes a handful of other smaller, including minor cleanup of a `filebeat_inputs` setting that was duplicated identically across environment files.

`deploy-stack` on Vagrant VMs is now possible end-to-end without errors.

:blowfish: Review by commit.

## Environments Affected

- **development**: primary target of this branch; box, provisioning, and env-var changes are Vagrant-specific.
- **All environments**: `couchdb_version` default now matches what `production`, `staging`, `eu`, `india` already set explicitly (3.5.0), so no behavior change there. `haproxy_version` default changes from 1.8 to 2.4 (upgrade long ago via changelog 0071).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,38 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+# VirtualBox is the default provider, and does not require any special set up.
+#
+# libvirt host set up
+#
+# 1. Create "cchq" storage pool
+#
+#   COMMCARE_CLOUD_LIBVIRT_POOL=cchq
+#   COMMCARE_CLOUD_LIBVIRT_DIR=/path/to/libvirt-cchq  # >35G disk space required full deploy-stack
+#   mkdir $COMMCARE_CLOUD_LIBVIRT_DIR
+#   virsh --connect qemu:///session pool-define-as $COMMCARE_CLOUD_LIBVIRT_POOL \
+#     dir --target $COMMCARE_CLOUD_LIBVIRT_DIR
+#   virsh --connect qemu:///session pool-build $COMMCARE_CLOUD_LIBVIRT_POOL
+#   virsh --connect qemu:///session pool-start $COMMCARE_CLOUD_LIBVIRT_POOL
+#   virsh --connect qemu:///session pool-autostart $COMMCARE_CLOUD_LIBVIRT_POOL
+#
+#   # assign SELinux tags if your system requires that
+#   COMMCARE_CLOUD_DIR=~/commcare-cloud
+#   sudo semanage fcontext -a -t svirt_home_t "$COMMCARE_CLOUD_LIBVIRT_DIR(/.*)?"
+#   sudo semanage fcontext -a -t svirt_home_t "$COMMCARE_CLOUD_DIR(/.*)?"
+#   sudo restorecon -Rv $COMMCARE_CLOUD_LIBVIRT_DIR $COMMCARE_CLOUD_DIR
+#
+# 2. Configure "libvirt-cchq" network
+#
+#   sudo nmcli connection add type bridge ifname libvirt-cchq con-name libvirt-cchq \
+#     ipv4.method manual ipv4.addresses 192.168.33.1/24 ipv6.method disabled
+#   sudo nmcli connection up libvirt-cchq
+#   echo "allow libvirt-cchq" | sudo tee -a /etc/qemu/bridge.conf
+#
+# 3. Configure environment and start VMs
+#
+#   vagrant up --provider=libvirt
+
 # Vagrantfile API/syntax version. Don't touch unless you know what you're doing!
 VAGRANTFILE_API_VERSION = "2"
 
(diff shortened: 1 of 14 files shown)


millerdev:
Note: auto-merge enabled so it doesn't get stale while I'm offline for a bit.

Vagrantfile
```diff
@@ -11,34 +43,41 @@ require_relative './provisioning/key_authorization'
...
-    end
-  end
 
-  config.vm.define "control" do |host|
-    host.vm.hostname = "control"
-    host.vm.network "private_network", ip: "192.168.33.14"
-    host.vm.hostname = "control.hq.dev"
-    host.vm.provider "virtualbox" do |v|
-      v.memory = 768
-      v.cpus = 1
```
gherceg, line 40: nit: would have been nice to collapse control in a separate commit from adding support for libvirt here.

src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
```diff
@@ -91,6 +91,7 @@
   become: yes
   community.general.snap:
     name: [astral-uv]
+    classic: true
```
gherceg, line 94: nit: a more descriptive commit message would be nice here to explain how this fixes the install task.

gherceg: Realized that we install `uv` with `--classic` so this is making the equivalent change here, but still would be nice to point it out for reviewers.

environments/development/public.yml
```diff
@@ -2,10 +2,11 @@ internal_domain_name: null
...
 
-http_proxy_address: '192.168.33.14'
-http_proxy_port: '3128'
-http_proxy_parent_address: '192.168.33.1'
-http_proxy_parent_port: '3128'
+# HTTP proxy breaks Vagrant networking
+#http_proxy_address: '192.168.33.14'
+#http_proxy_port: '3128'
+#http_proxy_parent_address: '192.168.33.1'
+#http_proxy_parent_port: '3128'
```
gherceg, line 9: Just curious if this is used when virtualbox is the provider, or if this is just not useful on a dev environment? Asking because we noticed the network configs are virtualbox and libvirt differ:

virtualbox
```
        override.vm.network "private_network", ip: ip
```
vs

libvirt
```
        override.vm.network "public_network", ip: ip,
          dev: "libvirt-cchq", type: "bridge", auto_config: true
```

gherceg (commented):
Did a first pass with @AmitPhulera and it looks good. We both feel we need to understand the Vagrantfile changes more before approving.

Another general question is what do we gain by adding support for libvirt? It seems the cost of supporting multiple tools is high, and can be felt by some of the workarounds needed to support libvirt in this PR. Open to adding it if there is a compelling reason though.

millerdev:
> what do we gain by adding support for libvirt

It's easier (than VirtualBox) to install on Linux, and in my experience it performs better too.

> the cost of supporting multiple tools is high, and can be felt by some of the workarounds needed to support libvirt in this PR

It's true that I did implement some workarounds to support libvirt. I expect that I would have had to implement similar (possibly different) workarounds for VirtualBox. As far as I can tell Vagrant has not been used by anyone on our team recently, so the existing configuration was pretty stale and frankly incomplete. It required one-off manual tweaks and workarounds that I have automated in this PR. In most cases I think those tweaks and workarounds would have been needed regardless of the virtualization backend.

If you think it is not worth maintaining support for both VirtualBox and libvirt then I would prefer to drop support for VirtualBox at least until someone wants it again. With this PR it is possible to run a full deploy-stack without errors on libvirt. Apparently libvirt can work on Mac OS, but I have never used it there. Either way, more testing would be required for other developer platforms, but I don't plan to do that because I don't need it for the Ansible upgrade.

Vagrantfile
millerdev: Good point. Noted for future work, but I'm not planning to rebase to split the commit at this time.

src/commcare_cloud/ansible/roles/common_installs/tasks/main.yml
millerdev: Yes, I should have done that. I'm not planning to rebase to fix it, so I will add it here:

The omission of `classic: true` was a bug (it's required to install `uv`), but it did not surface before because we had manually installed uv on all relevant machines and/or we never actually tested this task.

environments/development/public.yml
millerdev: It's a good question. I did not test on VirtualBox, so I'm not sure if it works there but I suspect it might not.

I expect the current configuration (with this commented out) will work on VirtualBox.

gherceg:
Fair points, and given the work to support libvirt is done here, it would be silly to not move forward with that!

gherceg (approved):

PR #6960, opened 2026-08-05
Configure a new queue `24_hour_queue`

AmitPhulera:
https://dimagi.atlassian.net/browse/SAAS-20143

This queue is supposed to host the long running tasks which have SLAs in hours. The motivation behind it is that recently there was an issue where saved reports were delayed because a domain deleted a lot of users which resulted in their forms getting deleted which rebuilt the cases, all of this happened in background_queue which was blocked for hours. 

Going forward if we see such tasks which are delaying the SLAs in any queue and they themselves have higher threshold for dealy we are going to add these tasks to the queue. 

This PR hosts that changes just to staging.

##### Environments Affected

Staging

##### Announce New Release

I am not actually sure if we should announce a changelog for it. We are rolling it in a way that it would not affect any third party hosters.

- [ ] (Dimagi only) After merging, I will follow [these instructions](https://confluence.dimagi.com/display/saas/Announcing+changes+affecting+third+parties#Announcingchangesaffectingthirdparties-announcing)


src/commcare_cloud/environment/schemas/app_processes.py
```diff
@@ -107,6 +107,7 @@ def __new__(cls, name, required=True, is_queue=True, blockage_threshold=None,
 
 
 CELERY_PROCESSES = [
+    CeleryProcess("24_hour_queue", required=False, blockage_threshold=24 * 60 * 60),
```
gherceg, line 110: I continue to waffle on `24_hour_queue` vs something like `eventual_queue`, but I ultimately think I like the clear definition of the `24_hour_queue`. Plus `eventual_queue` is no different from `background_queue`.

gherceg (commented):
To provide some additional context, we talked about this in a meeting recently and felt that there wasn't an obvious benefit to having entirely feature based queues (`repeat_record_queue`, etc) or entirely time based queued (`1_hour_queue`, etc). They both have tradeoffs, but having a clear "slow queue" to put tasks that just need to succeed eventually feels right.

> I am not actually sure if we should announce a changelog for it. We are rolling it in a way that it would not affect any third party hosters.

I think a changelog that notes this optional change is good.

Also just need to fix the conflict on master now that we removed the redundant celery machine on staging.

src/commcare_cloud/environment/schemas/app_processes.py
millerdev: My thought is that "24 hours" implies that backlogs greater than 24h would seem to be breaking an SLA where one does not exist, while "eventual" is a softer boundary. That's a soft +1 for "eventual" over "24 hour". Maybe we could use a name that more clearly differentiates from `background_queue` such as `slower_background_queue`?

It would be nice to have a comment somewhere explaining the purpose of this queue and its relationship to `background_queue`.

dannyroberts: I don't feel strongly about it, but my rough conceptual reason for proposing 24h as a sort of soft-SLA or proxy-SLA is because 24 hours is our natural load cycle, so if a queue cannot clear itself in 24h it likely cannot clear itself in 48 hours, and, if things keep up like that, ever. (We do have weaker but distinct cycles at the weekly and annual level, but that seems like too long a timespan to try and factor in here.) So I was sort of proposing we set 24h as the de facto SLA for this queue, not because we care so much specifically about 24h, but because that's a rough proxy of it being at risk of _never_ clearing.

millerdev: > if a queue cannot clear itself in 24h it likely cannot clear itself in 48 hours

I agree for normal operations, but when someone does something unusual (example from PR description: _a domain deleted a lot of users_) I can imagine it could take a long time to get through that backlog, and that's fine as long as it's an infrequent operation. Maybe these things are still likely to clear in less than 24 hours, and we should be looking into it if not? If that's true, then the 24h name is fine.

AmitPhulera: My understanding is similar to Danny. Another thing that I would like to add - In the specific incident which prompted creation of this queue, the bulk of the jobs when ran together taken updates the queue resources and `background_queue` was not able to process other tasks. 
The tasks that ran were not super urgent but the tasks it blocked were time sensitive. 
So I perceive this queue as something to be used to protect the time sensitive tasks from other queues. We will start with the ones that we caused issues this time and whenever we will face any such situation in future because of some other tasks they would be moved in the queue. 
I am also not attached with the 24h_* name, I am fine with it but would be open to changing it if we agree on a better name.

gherceg: Another possible takeaway is that whatever was more urgent on the background_queue maybe shouldn't be on the background_queue? It seems like that was created with a similar intention to be the queue where it is safe for things to happen "in the background".

AmitPhulera:
I dropped ball on this one 😅 
@millerdev @dannyroberts @gherceg Do you think there are any blockers in rolling it out?

gherceg (approved):
Happy to approve. Curious to see what we end up putting on this queue. Can always rename it, or remove it if we don't find value in it.

PR #6963, opened 2026-08-10
Look up per-domain Project DB roles with pgbouncer auth_query

esoergel:
These three stacked PRs introduce a database-level safeguard that keeps a Project DB query from reading another domain's data, even if the query layer above it has a bug or is handed hostile SQL.

## Part I: HQ Query Users (https://github.com/dimagi/commcare-cloud/pull/6962)

Each domain gets its own postgres role, which has read-only access to only that domain's tables.  Queries on behalf of that domain are authenticated as that user, so postgres enforces separation between tenants.

HQ provisions these query users using a constrained `SECURITY DEFINER` function called `projectdb_provision_role` which narrowly defines the action.  This function is created during database provisioning.  Passwords are derived from the domain name via `django.utils.crypto.salted_hmac`, never stored.

## Part II: PgBouncer

PgBouncer connection pooling complicates things in two ways

### Part IIa: PGBouncer Authentication (This PR)

PgBouncer authenticates queries itself, so it has to know user passwords.  Currently, it validates against a file called `userlist.txt` which is managed by commcare-cloud.  Adding a new user requires amending that file, which just isn't doable in this context.  Instead, we give pgbouncer an `auth_query` it can run to get the username and password to authenticate a user that's not in `userlist.txt`.

To do this, we need a few things in place:
 * A `user_lookup` function as described in the pgbouncer docs (https://www.pgbouncer.org/config.html), supporting the `auth_query`
 * An `auth_user` pgbouncer can authenticate as when running the `auth_query`, and a password for that user
 * A role for the pgbouncer auth user with only LOGIN permissions by default
   * This role must also be granted access to the `user_lookup` function

### Part IIb: (https://github.com/dimagi/commcare-cloud/pull/6965)

PgBouncer keys its pools by `(user, database)` pair, meaning that each domain will have it's own pool in pgbouncer.  Right now, we allow up to 490 open connections per pool (`default_pool_size`). Since ProjectDB uses many pools, we can instead set a `max_db_connections` as a limit across all ProjectDB pools, and set a much lower `pool_size` to prevent one domain from claiming all that for itself.

##### Environments Affected

Staging, since it has `project_db` configured already.  Should only affect other environments when that's enabled.

##### Announce New Release

No.  Shouldn't require action by anyone but me.

### DIFF
--- a/docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
+++ b/docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
@@ -87,6 +87,13 @@ that add a ``project_db`` block to `\ ``dbs`` <#dbs>`_ and map it in `\
 settings default to those of the ``ucr`` db when not set explicitly. Installs
 required extensions automatically as part of database setup.
 
+Enabling it in an environment requires one manual step: set
+``PGBOUNCER_AUTH_PASSWORD`` in the vault.  Each domain gets its own postgres
+login role, provisioned from HQ using ``projectdb_provision_role`` and
+``projectdb_drop_role``.  To validate these logins, pgbouncer runs an
+``auth_query`` to get the password hash from the DB, connecting to postgres
+using its own ``auth_user`` and ``pgbouncer_auth`` role.
+
 ``synclogs``
 ^^^^^^^^^^^^^^^^
 
--- a/src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
@@ -18,4 +18,4 @@ postgresql_port: 5432
 
 pgbouncer_processes: 1
 pgbouncer_current_processes: [0]
-pgbouncer_min_version: "1.13"
+pgbouncer_min_version: "1.14"
--- a/src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
@@ -13,7 +13,7 @@
 {{ db.name }} = host={{ "localhost" if db.host == inventory_hostname else db.host }} port={{ postgresql_port }}
 {%- if db.pgbouncer_pool_size %} pool_size={{ db.pgbouncer_pool_size // pgbouncer_processes }}{% endif %}
 {%- if db.pgbouncer_reserve_pool_size is not none %} reserve_pool={{ db.pgbouncer_reserve_pool_size // pgbouncer_processes }}{% endif %}
-{%- if postgresql_dbs.project_db and db.name == postgresql_dbs.project_db.name %} auth_user={{ pgbouncer_auth_username }} auth_query='SELECT usename, passwd FROM pgbouncer.user_lookup($1)'{% endif %}
+{%- if postgresql_dbs.project_db and db.name == postgresql_dbs.project_db.name %} auth_user={{ pgbouncer_auth_username }} auth_query='SELECT usename, passwd FROM pgbouncer.project_db_user_lookup($1)'{% endif %}
 
 {% endfor %}
 
--- a/src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
+++ b/src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
@@ -10,7 +10,10 @@
 {% set current_host = inventory_hostname %}
 {% if is_monolith %}{% set current_host = DEFAULT_POSTGRESQL_HOST %}{% endif %}
 {% for db in postgresql_dbs.by_pgbouncer_host.get(current_host, []) %}
-{{ db.name }} = host={{ "localhost" if db.host == inventory_hostname else db.host }} port={{ postgresql_port }}{% if db.pgbouncer_pool_size %} pool_size={{ db.pgbouncer_pool_size // pgbouncer_processes }}{% endif %}{% if db.pgbouncer_reserve_pool_size is not none %} reserve_pool={{ db.pgbouncer_reserve_pool_size // pgbouncer_processes }}{% endif %}
+{{ db.name }} = host={{ "localhost" if db.host == inventory_hostname else db.host }} port={{ postgresql_port }}
+{%- if db.pgbouncer_pool_size %} pool_size={{ db.pgbouncer_pool_size // pgbouncer_processes }}{% endif %}
+{%- if db.pgbouncer_reserve_pool_size is not none %} reserve_pool={{ db.pgbouncer_reserve_pool_size // pgbouncer_processes }}{% endif %}
+{%- if postgresql_dbs.project_db and db.name == postgresql_dbs.project_db.name %} auth_user={{ pgbouncer_auth_username }} auth_query='SELECT usename, passwd FROM pgbouncer.user_lookup($1)'{% endif %}
 
 {% endfor %}
 
--- a/src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
+++ b/src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
@@ -169,20 +169,20 @@
         state: present
         db: '{{ postgresql_dbs.project_db.name }}'
         port: "{{ postgresql_port }}"
-        login_host: "{{ db_is_remote and postgresql_host or omit }}"
-        login_user: "{{ db_is_remote and postgres_users.root.username or omit }}"
-        login_password: "{{ db_is_remote and postgres_users.root.password or omit }}"
+        login_host: "{{ postgresql_host if db_is_remote else omit }}"
+        login_user: "{{ postgres_users.root.username if db_is_remote else omit }}"
+        login_password: "{{ postgres_users.root.password if db_is_remote else omit }}"
         encrypted: True
       no_log: true
 
 - name: Set up the project_db database
   become: yes
-  become_user: "{{ db_is_remote and 'ansible' or 'postgres' }}"
+  become_user: "{{ 'ansible' if db_is_remote else 'postgres' }}"
   postgresql_query:
     db: '{{ postgresql_dbs.project_db.name }}'
     port: "{{ postgresql_port }}"
-    login_host: "{{ db_is_remote and postgresql_host or omit }}"
-    login_user: "{{ db_is_remote and postgres_users.root.username or omit }}"
-    login_password: "{{ db_is_remote and postgres_users.root.password or omit }}"
+    login_host: "{{ postgresql_host if db_is_remote else omit }}"
+    login_user: "{{ postgres_users.root.username if db_is_remote else omit }}"
+    login_password: "{{ postgres_users.root.password if db_is_remote else omit }}"
     query: "{{ lookup('template', 'project_db_setup.sql.j2') }}"
   when: postgresql_dbs.project_db and postgresql_host == postgresql_dbs.project_db.host
--- a/src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
+++ b/src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
@@ -151,7 +151,7 @@
 # to validate dynamically provisioned roles (for ProjectDB)
 - name: Create pgbouncer auth user
   become: yes
-  become_user: "{{ db_is_remote and 'ansible' or 'postgres' }}"
+  become_user: "{{ 'ansible' if db_is_remote else 'postgres' }}"
   vars:
     ansible_ssh_pipelining: true
   when: postgresql_dbs.project_db and postgresql_host == postgresql_dbs.project_db.host
--- a/src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
+++ b/src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
@@ -147,6 +147,34 @@
       END $$;
   when: postgresql_dbs.repeaters and postgresql_host == postgresql_dbs.repeaters.host
 
+# Create pgbouncer user for auth_query calls
+# to validate dynamically provisioned roles (for ProjectDB)
+- name: Create pgbouncer auth user
+  become: yes
+  become_user: "{{ db_is_remote and 'ansible' or 'postgres' }}"
+  vars:
+    ansible_ssh_pipelining: true
+  when: postgresql_dbs.project_db and postgresql_host == postgresql_dbs.project_db.host
+  block:
+    - name: Check that the pgbouncer auth password is set
+      assert:
+        that: PGBOUNCER_AUTH_PASSWORD
+        fail_msg: 'PGBOUNCER_AUTH_PASSWORD must be set for envs with a project_db'
+
+    - name: Create the role
+      postgresql_user:
+        name: "{{ pgbouncer_auth_username }}"
+        password: "{{ PGBOUNCER_AUTH_PASSWORD }}"
+        role_attr_flags: 'LOGIN,NOSUPERUSER,NOCREATEROLE,NOCREATEDB'
+        state: present
+        db: '{{ postgresql_dbs.project_db.name }}'
+        port: "{{ postgresql_port }}"
+        login_host: "{{ db_is_remote and postgresql_host or omit }}"
+        login_user: "{{ db_is_remote and postgres_users.root.username or omit }}"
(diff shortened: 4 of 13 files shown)


src/commcare_cloud/ansible/roles/pgbouncer/templates/pgbouncer.ini.j2
```diff
@@ -47,6 +47,11 @@ so_reuseport = 1
 ;;;
 auth_type = {{ pgbouncer_auth_type|default('md5') }}
 auth_file = {{ pgbouncer_users }}
+{% if postgresql_dbs.project_db %}
+; Support for authenticating as dynamically created users for ProjectDB
+auth_user = {{ pgbouncer_auth_username }}
+auth_query = SELECT usename, passwd FROM pgbouncer.user_lookup($1)
```
esoergel, line 53: Setting this does mean that any attempts to login with a bad user will error in dbs that don't have `user_lookup` defined

MartinRiese (commented):
Makes sense to me. But I don't feel comfortable approving.

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
```diff
@@ -51,3 +51,23 @@ GRANT EXECUTE ON FUNCTION projectdb_drop_role(text) TO {{ postgres_users.commcar
...
+CREATE SCHEMA IF NOT EXISTS pgbouncer;
+
+CREATE OR REPLACE FUNCTION pgbouncer.user_lookup(username text)
+RETURNS TABLE(usename name, passwd text)
+LANGUAGE sql
+SECURITY DEFINER
+SET search_path = pg_catalog, pg_temp
+AS $$
+  SELECT rolname, CASE WHEN rolvaliduntil < now() THEN NULL ELSE rolpassword END
+  FROM pg_authid WHERE rolname = username AND rolcanlogin;
```
millerdev: I see that this is a general function, but for our purposes should it have a `LIKE 'projectdb\_%'` namespace check? As is it also works for any role.

src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
```diff
@@ -147,6 +147,34 @@
       END $$;
   when: postgresql_dbs.repeaters and postgresql_host == postgresql_dbs.repeaters.host
 
+# Create pgbouncer user for auth_query calls
+# to validate dynamically provisioned roles (for ProjectDB)
+- name: Create pgbouncer auth user
+  become: yes
+  become_user: "{{ db_is_remote and 'ansible' or 'postgres' }}"
```
millerdev:
```suggestion
  become_user: "{{ 'ansible' if db_is_remote else 'postgres' }}"
```

There are other instances of the `<condition> and <true-value> or <false-value>` pattern below. The `if/else` syntax is easier to read and not prone to a wrong outcome when `<true-value>` evaluates to false.

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
```diff
@@ -70,4 +70,3 @@ $$;
 REVOKE EXECUTE ON FUNCTION pgbouncer.user_lookup(text) FROM PUBLIC;
 GRANT USAGE ON SCHEMA pgbouncer TO {{ pgbouncer_auth_username }};
 GRANT EXECUTE ON FUNCTION pgbouncer.user_lookup(text) TO {{ pgbouncer_auth_username }};
-GRANT CONNECT ON DATABASE {{ postgresql_dbs.project_db.name }} TO {{ pgbouncer_auth_username }};
```
millerdev: d969db4be8560b9a9d8f68ac917def8997d440ed commit message is "pgbouncer" - would have been nice to elaborate on what that means? No need to rebase/amend the commit message now, but maybe leave a comment here with a bit more detail about why this was removed?

millerdev (approved):

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
esoergel: Oh, sorry about that - I'd meant to rebase and fixup that commit into the commit that introduced that line, so it never showed up in the PR, but must've forgotten.  That came up in my own code review of this with Claude, that it shouldn't be necessary to `GRANT CONNECT`, as that's already the default.

src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
esoergel: Totally agree

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
esoergel: It could - that's something I considered when implementing this, though I think I was hedging towards keeping this generic in case we needed to adopt this pattern more broadly in the future. However it does seem better to be stricter about the scope if there's no big disadvantage to doing so, as this could conceivably be part of a privilege escalation avenue

src/commcare_cloud/ansible/roles/postgresql_base/tasks/set_up_dbs.yml
```diff
@@ -147,6 +147,34 @@
...
+      postgresql_user:
+        name: "{{ pgbouncer_auth_username }}"
+        password: "{{ PGBOUNCER_AUTH_PASSWORD }}"
+        role_attr_flags: 'LOGIN,NOSUPERUSER,NOCREATEROLE,NOCREATEDB'
+        state: present
+        db: '{{ postgresql_dbs.project_db.name }}'
+        port: "{{ postgresql_port }}"
+        login_host: "{{ db_is_remote and postgresql_host or omit }}"
+        login_user: "{{ db_is_remote and postgres_users.root.username or omit }}"
+        login_password: "{{ db_is_remote and postgres_users.root.password or omit }}"
```
millerdev: The second part of my [earlier comment](https://github.com/dimagi/commcare-cloud/pull/6963#discussion_r3785465328) was easy to miss. Do you agree that these should also be changed to `<true-value> if <condition> else <false-value>` syntax?

esoergel: Oh thanks yeah I did miss that.  I'm guessing Claude did it that way because that's apparently the pattern in this file.  Not sure whether the existing references intend to fallback to `omit` when the value is Falsy, but I can at least deal with that here

millerdev (approved):

MartinRiese:
@millerdev @gherceg @esoergel Could I get another review?

The force pushes that involved Ethans commits were all rebasing using github's stacks. So this is only for the four I added. With the [last one](https://github.com/dimagi/commcare-cloud/pull/6963/changes/4b62d8df6834fe4df8963610c78452038ceae94a) being the only controversial one.

AWS RDS does not expose pg_authid to anybody. So added a separate table that we manage would be one option. That is what (A)I implemented. The password is already salted and hashed at that point so storing them in this way should be fine.

An alternative that Ethan pointed out would be to use a common password between the users that would be provided at deploy time like for the other db users. Claude complained about that because the whole point is to separate access to different domains. Since we menage the passwords anyways and there would still be a separate user I don't think that would be a problem. I would like to hear your take on that.

p.s. the [run book](https://docs.google.com/document/d/18sYTyrEjsMUvchrOnARr-SNkQDNSg1LvsBVkh-GbxyQ/edit?tab=t.0) now passes.

src/commcare_cloud/environment/schemas/postgresql.py
```diff
@@ -164,6 +164,7 @@ def replace_hosts(self, environment):
                 self.dbs.project_db.pgbouncer_hosts = list(ucr.pgbouncer_hosts)
             if self.dbs.project_db.pgbouncer_endpoint is None:
                 self.dbs.project_db.pgbouncer_endpoint = ucr.pgbouncer_endpoint
+            self.pgbouncer_override.pgbouncer_auth_type = 'scram-sha-256'
```
gherceg, line 167: nit: would put this line before the if blocks just for easier readability, but I know this is very nitpicky.

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
```diff
@@ -5,6 +5,17 @@ CREATE EXTENSION IF NOT EXISTS earthdistance;
 CREATE EXTENSION IF NOT EXISTS pg_trgm;
 CREATE EXTENSION IF NOT EXISTS fuzzystrmatch;
 
+-- RDS doesn't expose pg_authid, create our own stand-in
+CREATE SCHEMA IF NOT EXISTS pgbouncer;
```
gherceg, line 9: Perhaps I'm misunderstanding, but if we are adding our own schema, should we name it something more custom/specific since it seems possible for this schema name to conflict if pgbouncer decided it wanted a schema in dbs it connects to?

gherceg (commented):
I did a pass over those recent commits but feel like I would need to do more of my own research to give feedback on your approach. If @millerdev hasn't looked yet I'd be interested in pairing on this as well.

MartinRiese:
@gherceg both fair points. I'll wait for more feedback before renaming anything.

src/commcare_cloud/environment/schemas/postgresql.py
millerdev: Is a migration from the old `md5` hashes to `scram-sha-256` planned? How will it be rolled out?

docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
```diff
@@ -87,6 +87,13 @@ that add a ``project_db`` block to `\ ``dbs`` <#dbs>`_ and map it in `\
 settings default to those of the ``ucr`` db when not set explicitly. Installs
 required extensions automatically as part of database setup.
 
+Enabling it in an environment requires one manual step: set
+``PGBOUNCER_AUTH_PASSWORD`` in the vault.  Each domain gets its own postgres
+login role, provisioned from HQ using ``projectdb_provision_role`` and
+``projectdb_drop_role``.  To validate these logins, pgbouncer runs an
+``auth_query`` to get the password hash from the DB, connecting to postgres
+using its own ``auth_user`` and ``pgbouncer_auth`` role.
```
gherceg: Do we need to put a unique password on each of these users?  The complexity around maintaining separate passwords seems like overkill, since it should not be possible to run a query as another domain's user, regardless of whether you know the password or not.

src/commcare_cloud/environment/schemas/postgresql.py
gherceg: Just to confirm, is it the case that we are not using this `pgbouncer_auth_type` at all currently, and therefore we don't need to worry about migrating old md5 hashes to scram-sha-256?

src/commcare_cloud/ansible/roles/pgbouncer/defaults/main.yml
```diff
@@ -12,6 +12,7 @@ pgbouncer_default_pool: 15
 pgbouncer_reserve_pool: 4
 pgbouncer_pool_timeout: 2
 pgbouncer_pool_mode: session
+pgbouncer_auth_type: md5
```
gherceg: Can this just default to scram if this is the first time we are using it?

gherceg (commented):
Paired with @millerdev

docs/source/reference/1-commcare-cloud/2.1-postgresql-config.rst
MartinRiese: @gherceg @millerdev That matches what Ethan suggested as alternative too.

- reverted the commit creating the table and the ones setting the hashing algorithm
- [19e76dd](https://github.com/dimagi/commcare-cloud/pull/6963/commits/19e76dd333df742057fbb1b04305b9b682f8c27f) uses a secret now that is shared between domains. I am wondering if we need code supporting updating that password?
- One extra commit for `DROP TABLE IF EXISTS pgbouncer.project_db_passwords;` which we can revert after deploying to staging.

I reverted instead of dropping to avoid force pushing. If you are ok with this approach I would try getting this on staging again.

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
```diff
@@ -62,8 +65,8 @@ LANGUAGE sql
 SECURITY DEFINER
 SET search_path = pg_catalog, pg_temp
 AS $$
-  SELECT rolname, CASE WHEN rolvaliduntil < now() THEN NULL ELSE rolpassword END
-  FROM pg_authid
+  SELECT rolname, '{{ PROJECTDB_USERS_PASSWORD }}'
```
millerdev: Why does the password need to be included in this SELECT statement? Would it make sense to put it in localsettings so HQ can access it like it does other secrets?

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
```diff
@@ -59,6 +59,10 @@ REVOKE CREATE ON SCHEMA public FROM PUBLIC;
 -- scoped by which role a domain is given, not by that role having a secret password.
 -- Scope limited to ProjectDB's per-domain roles, which is all pgbouncer needs it for
 CREATE SCHEMA IF NOT EXISTS pgbouncer;
+
+-- Leftover from when each role's password was stored here instead of shared.
+DROP TABLE IF EXISTS pgbouncer.project_db_passwords;
```
millerdev: We could run this directly on the staging projectdb database and not include it here since that table only exists on staging.

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
```diff
@@ -18,6 +18,7 @@ BEGIN
     RAISE EXCEPTION 'refusing to manage role %', role_name;
   END IF;
   EXECUTE format('CREATE ROLE %I WITH LOGIN PASSWORD %L', role_name, role_password);
+  EXECUTE format('GRANT %I TO %I', role_name, current_user);
```
millerdev: Can we move this line into `projectdb_drop_role` so the definer is only a member of the projectdb role temporarily while dropping it?

millerdev:
^ The previous was a pair review with @gherceg

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
MartinRiese: I already ran this on staging. So this commit can be dropped in the clean up PR. I just found it easier to do it this way than to figure out how to get to the postgres shell.

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
MartinRiese: 9821e40457fd599257640b5554cd134a9015e5cb

Tested on staging with a new domain to avoid existing grants (` cchq staging django-manage manage_project_db martins-smart-links --drop --sync --populate`)

src/commcare_cloud/ansible/roles/postgresql_base/templates/project_db_setup.sql.j2
MartinRiese: This is not called by HQ but by pgbouncer (see `{%- if postgresql_dbs.project_db and db.name == postgresql_dbs.project_db.name %} auth_user={{ pgbouncer_auth_username }} auth_query='SELECT usename, passwd FROM pgbouncer.project_db_user_lookup($1)'{% endif %}` above).

1. HQ sends request with username and password (form localsettings) to pq bouncer
1. pg bouncer call auth_query with username
1. function checks that role exists and returns the expected password (that's what the code above does)
1. pg bounces compares the passwords
1. if they match forwards request.

gherceg: Right because despite removing the unique password per user requirement, the following still holds true:
> PgBouncer authenticates queries itself, so it has to know user passwords. Currently, it validates against a file called userlist.txt which is managed by commcare-cloud. Adding a new user requires amending that file, which just isn't doable in this context. Instead, we give pgbouncer an auth_query it can run to get the username and password to authenticate a user that's not in userlist.txt.

millerdev: Ah, I forgot that part. In that case it sounds like this can't be avoided.

PR #6967, opened 2026-08-19
Curb OOMs on pgbouncer-a6 by recycling server connections

dannyroberts:
pgbouncer-a6-production (the plproxy host) keeps OOMing. I believe this is at least in part due to the way postgres manages memory on reused connections: as queries run, they push up the high water mark of memory used by that connection, and that memory is not released until the connection is closed. Currently, these connections are set to live and be reused for an hour (pgbouncer's default `server_lifetime` of 3600) or until they're idle for 10m (default `server_idle_timeout` of 600). Capping the lifetime at 5 minutes on this host bounds how long any bloated backend survives. `server_idle_timeout` is deliberately left at the default for production, because with a 5-minute lifetime it would almost never be the operative limit.

Nothing changes for other hosts or environments: config changes on those environments just explicitly set these values to their existing default values.

**Rollout:** an ini-only change triggers the role's per-process **reload** handler, so deploying doesn't drop connections. (The full-restart handler only fires on systemd unit file changes, which this PR doesn't touch.)

```
cchq production ap deploy_postgres.yml --tags=pgbouncer --limit=pgbouncer_a6
```

This bounds how long bloat survives, not how big one query can get; a proxy-DB `statement_timeout` and paginating the unbounded callers are possible follow-ups.

##### Environments Affected

production (pgbouncer-a6 only)

### DIFF
--- a/environments/production/inventory.ini
+++ b/environments/production/inventory.ini
@@ -86,7 +86,7 @@ mobile_webworkers
 [pgbouncer_a5]
 10.202.40.44 hostname=pgbouncer-a5-production ufw_private_interface=ens5 ansible_python_interpreter=/usr/bin/python3 ec2_instance_id=i-0421eb020cc421ad6 root_encryption_mode=aws pgbouncer_processes=2
 [pgbouncer_a6]
-10.202.40.98 hostname=pgbouncer-a6-production ufw_private_interface=ens5 ansible_python_interpreter=/usr/bin/python3 ec2_instance_id=i-0ccf6dcc1a4cb2b9b root_encryption_mode=aws pgbouncer_processes=2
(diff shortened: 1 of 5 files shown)


millerdev (approved):
Open questions:
- How/when will we know if this made a positive impact?
- What is the down-side? How will we know if it makes things worse?

dannyroberts:
One potential sign it's helping would be if the steady state memory is noticeably lower—that would align with connections holding onto less memory on average due to more frequent recycling.

But yeah, it could help in the pathological case without showing much immediate sign, and we wouldn't really get a positive signal for that.

Agree we should also keep an eye out for things getting worse — higher memory, higher cpu, slower average query times, etc

dannyroberts:
Tests are hung on what seems to be an unrelated github actions issue, and I don't think the remaining tests would plausibly catch a bug in pgbouncer settings, so I'm going to go ahead and merge as is.

millerdev:
Follow-up: this definitely improved memory usage on the machine where it was applied, so it seems like a very good change.

PR #6972, opened 2026-09-08
Add Molecule testing infrastructure and swap role tests

millerdev:
Adds [Molecule](https://ansible.readthedocs.io/projects/molecule/) as the testing framework for commcare-cloud's Ansible roles, and the first tests using it, for the `swap` role. This gives maintainers a repeatable, containerized way to verify a role's behavior (and its idempotence) without needing a full VM or a real target host.

https://dimagi.atlassian.net/browse/SAAS-19960

:blowfish: Review by commit.

##### Environments Affected

None. Only adds test tooling and CI configuration.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,8 @@ ansible-lint-cloud = "commcare_cloud.ansible_lint:main"
 test = [
     "ansible-lint==5.4.0",
     "docopt",
+    "molecule",
+    "molecule-plugins[docker,podman]",
     "pytest",
     "pytest-unmagic>=1.1.0",
     "requests-mock",
--- a/tests/molecule/README.md
+++ b/tests/molecule/README.md
@@ -1,0 +1,53 @@
+# Molecule tests
+
+Molecule tests for roles under `src/commcare_cloud/ansible/roles/` live
+here, one directory per role/scenario pair:
+
+```
+tests/molecule/<role>-<scenario>/
+    molecule.yml
+    converge.yml
+    prepare.yml   (optional)
+    verify.yml
+```
+
+## Running scenarios
+
+```sh
+./tests/molecule/run.sh [ROLE [SCENARIO]]
+```
+
+## Adding a new scenario for a role
+
+```sh
+./tests/molecule/init.sh ROLE [SCENARIO]
+```
+
+## Container choice
+
+The `containers` molecule driver is used, which picks whichever of
+`podman`/`docker` is available. If your container engine is rootless — check
+with `podman info`/`docker info` — that matters for two things role authors
+should know about:
+
+- **Anything needing real root against the host kernel** (enabling swap,
+  non-namespaced sysctls, raw device access, etc.) **fails in a rootless
+  container**, even with `privileged: true` — its "root" is a mapped
+  unprivileged host UID. Work around it in the scenario (e.g. stub the
+  binary in `prepare.yml`), not the role.
+- **Non-namespaced sysctls** can be written to their config file, but the live
+  `/proc/sys` value never actually changes, so `ansible.posix.sysctl` reports
+  `changed` every run, breaking idempotence. Fix with `module_defaults`
+  (`sysctl_set`/`reload: false`) on the scenario's `include_role` task.
+
+The platform image is `docker.io/geerlingguy/docker-ubuntu2204-ansible`, rather
+than a generic systemd/Ubuntu image. It's built specifically for
+Ansible/Molecule testing — systemd, `sudo`, and `python3` pre-configured.
+
+## About the directory structure
+
+Molecule assumes one role per repo, and by default puts each role's scenarios
+beside its source files. commcare-cloud, however, keeps all tests under the
+`tests` directory at the repo root, so scenarios live in `tests/molecule/`.
+Molecule also requires scenario names to be unique with no notion of "role" so
+the role name is folded into the scenario directory name itself.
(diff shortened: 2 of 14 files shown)


pyproject.toml
```diff
@@ -44,6 +44,8 @@ ansible-lint-cloud = "commcare_cloud.ansible_lint:main"
 test = [
     "ansible-lint==5.4.0",
     "docopt",
+    "molecule",
```
AmitPhulera, line 47: Is molecule package not dependent on ansible version? If yes, do we want to pin it?

tests/molecule/README.md
```diff
@@ -0,0 +1,53 @@
...
+  non-namespaced sysctls, raw device access, etc.) **fails in a rootless
+  container**, even with `privileged: true` — its "root" is a mapped
+  unprivileged host UID. Work around it in the scenario (e.g. stub the
+  binary in `prepare.yml`), not the role.
+- **Non-namespaced sysctls** can be written to their config file, but the live
+  `/proc/sys` value never actually changes, so `ansible.posix.sysctl` reports
+  `changed` every run, breaking idempotence. Fix with `module_defaults`
+  (`sysctl_set`/`reload: false`) on the scenario's `include_role` task.
+
+The platform image is `docker.io/geerlingguy/docker-ubuntu2204-ansible`, rather
```
AmitPhulera, line 43: Is this official image that is recommended?

tests/molecule/run.sh
AmitPhulera, line 1: The benefit of having this script is that we won't have to pass env variables everytime we run the tests, right?

tests/molecule/swap-default/molecule.yml
```diff
@@ -1,13 +1,31 @@
...
+  inventory:
+    group_vars:
+      all:
+        swap_size: 512M
+scenario:
+  # Drop dependency/cleanup/side_effect -- they're unused here,
+  # and leaving them in emits warnings on every run
+  test_sequence:
+    - lint
+    - destroy
```
AmitPhulera, line 22: Why is destroy needed in test sequence?

tests/molecule/swap-removal/converge.yml
AmitPhulera, line 1: Is it expected to have this level of duplication across all the scenarios in role? Is there a way in molecule to pull in common playbook(?) to a place and refer them from there.

tests/molecule/swap-default/molecule.yml
```diff
@@ -1,13 +1,31 @@
 ---
 dependency:
   name: galaxy
+  enabled: false
 driver:
   name: containers
 platforms:
   - name: instance
-    image: quay.io/centos/centos:stream8
+    image: docker.io/geerlingguy/docker-ubuntu2204-ansible
```
AmitPhulera, line 9: Do we always want to pull the latest image for it?

AmitPhulera (commented):
This is looking good. Super excited to have it integrated into our testing framework. Thankyou for making it a reality. 🤩 

I have some open questions which I have flagged in the comments.

pyproject.toml
millerdev: uv will pick a compatible version of related/unpinned dependencies if Ansible is pinned.

tests/molecule/swap-default/molecule.yml
millerdev: There is only one tag on this image (`latest`, which is the default), so I think it's fine this way.

tests/molecule/swap-default/molecule.yml
millerdev: This is Molecule's default `test_sequence` minus a few that were emitting warnings. The only reason I added it is to avoid the warnings.

AmitPhulera (approved):
Got on a call with Daniel and we discussed the questions that were put up in the PR.  

We agreed that this is the right way to go and Daniel is going to add details about them as response to the questions here.

tests/molecule/run.sh
millerdev: Yes, and it also adds a convenient way to run all scenarios for a single role which is not so easy by default.

tests/molecule/README.md
millerdev: This is the best image I could find for what we need. Images like `ubuntu:22.04` (the base image of this one) do not provide systemd, sudo, etc., which many of our roles depend on.

tests/molecule/swap-removal/converge.yml
millerdev: I think yes. I will be looking for ways to avoid duplication as I build out more scenarios.

tests/molecule/swap-default/molecule.yml
millerdev: Another option is to remove this and add empty stub files where needed to silence the warnings. I'm annoyed that Molecule emits warnings when optional files are not present. Seems like they should be ignored or maybe emit an INFO level message rather than a warning? I want to learn more about how it works before I make judgements about Molecule's architecture.

tests/molecule/swap-removal/converge.yml
millerdev: Fixed in https://github.com/dimagi/commcare-cloud/pull/6974/commits/d271181db7e4e5646ceb50c7f26bef15582bd74f

tests/molecule/swap-default/molecule.yml
millerdev: Switched to a different technique that does not modify/duplicate the test sequence. See https://github.com/dimagi/commcare-cloud/pull/6974/commits/ee16fa88468543073bee41ecd210c7c6627b7c2a and https://github.com/dimagi/commcare-cloud/pull/6974/commits/4053dbb0aed976d48cfc9e5a994962c0888efc53

PR #6974, opened 2026-09-10
Eliminate duplication in Molecule scenarios

millerdev:
Follow-up to https://github.com/dimagi/commcare-cloud/pull/6972

Also added https://github.com/dimagi/commcare-cloud/pull/6974/commits/7515893de3b7a3b463761313ac565393b11ca072 to fix an issue with the scenario naming scheme.

:blowfish: Review by commit.

##### Environments Affected
None. Tests only.


tests/molecule/swap-default/molecule.yml
```diff
@@ -10,22 +10,12 @@ platforms:
...
+    side_effect: ../lib/stub.yml
   inventory:
     group_vars:
       all:
         swap_size: 512M
-scenario:
-  # Drop dependency/cleanup/side_effect -- they're unused here,
-  # and leaving them in emits warnings on every run
-  test_sequence:
-    - lint
```
AmitPhulera, line 21: When I was looking at molecules [default test_sequence](https://docs.ansible.com/projects/molecule/workflow/#example-flow) it was  - 

```
dependency → cleanup → destroy → syntax → create → prepare → converge → idempotence → side_effect → verify → cleanup → destroy
```

I don't see a `lint` step in there. I was not sure if `syntax` and `lint` are the same. 

Wanted to flag if this was something you looked into.

AmitPhulera (approved):
Looking good. Thanks for the cleanup. 

I have an open question which is non blocking.

tests/molecule/swap-default/molecule.yml
millerdev: It's a good question, and I didn't look into it. I didn't notice a functionality difference in how molecule ran tests when I switched this. We have a separate linter job that runs on GHA, so a lint step here seems unnecessary/redundant anyway.

PR #6976, opened 2026-09-11
Add Molecule tests for the datadog role, fix idempotence bugs

millerdev:
Adds a Molecule test scenario for the `datadog` Ansible role, which previously had no automated coverage. Vagrant's `deploy-stack` does not exercise it since `DATADOG_ENABLED is False` in the development environment.

Building the test surfaced several real, pre-existing bugs in the role that caused it to report spurious "changed" state on every deploy after the first (not just in tests) — those are fixed here too:

- a403e2fe8ca8d16b3390129a5c1bfe5d3765f14c An `apt-key del` cleanup step always reported changed regardless of whether it removed anything; it now checks presence first and only reports/acts on keys that actually exist.
- 40c89a649908834de08dfca3286096e83c996573 The "remove old datadog apt keys" cleanup list included the *current* key file, so every deploy deleted and immediately recreated it.
- 0fae3bc7d43c45c193131739f71f1f1ea5ed5401 The new signed-by apt repo and the legacy repo's cleanup file collided on the same auto-generated filename, so every deploy deleted and recreated the current repo config.
- 1ecd8776731e7a20630f688357978e242e51e9a0 The `datadog_integration_*` enable flags only existed in `group_vars/all.yml`, so the role silently depended on that file being loaded rather than providing sane defaults itself.

The first two commits on this branch are housekeeping/tooling improvements. Notably, cdb5a57df93121c4a99e2a6e451c195e69be565b adds a `-c/--command` option to `tests/molecule/run.sh` for running a single molecule subcommand (e.g. `converge`, `idempotence`) against one scenario instead of the full `test` sequence. Useful for fast local iteration when debugging a scenario.

https://dimagi.atlassian.net/browse/SAAS-20255

##### Environments Affected

The `datadog_integration_*` default changes and the role bug fixes apply to every environment with `DATADOG_ENABLED: true` (production, staging, eu, india, confluence, ...). All are behavior-neutral.

### DIFF
--- a/tests/molecule/datadog.default/prepare.yml
+++ b/tests/molecule/datadog.default/prepare.yml
@@ -1,0 +1,79 @@
+---
+- name: Prepare
+  hosts: all
+  become: true
+  tasks:
+    - name: Ensure dpkg-dev is present (for dpkg-scanpackages)
+      ansible.builtin.apt:
+        name: dpkg-dev
+        state: present
+        update_cache: true
+
+    # The role installs the datadog-agent package from apt.datadoghq.com. Serve
+    # a fake one from a local repo so the role apt task installs it rather than
+    # hitting Datadog's repo on every CI run.
+    - name: Create fake datadog-agent package tree
+      ansible.builtin.file:
+        path: /tmp/fake-datadog-agent/DEBIAN
+        state: directory
+        mode: "0755"
+
+    - name: Write fake package control file
+      ansible.builtin.copy:
+        dest: /tmp/fake-datadog-agent/DEBIAN/control
+        content: |
+          Package: datadog-agent
+          Version: 999:1.0.0
+          Architecture: all
+          Maintainer: Molecule Test <test@example.com>
+          Description: Fake datadog-agent stub for molecule testing
+
+    - name: Write fake package postinst
+      ansible.builtin.copy:
+        dest: /tmp/fake-datadog-agent/DEBIAN/postinst
+        mode: "0755"
+        content: |
+          #!/bin/sh
+          set -e
+          getent group dd-agent >/dev/null || groupadd --system dd-agent
+          getent passwd dd-agent >/dev/null || useradd --system --gid dd-agent --no-create-home --shell /usr/sbin/nologin dd-agent
+          mkdir -p /etc/datadog-agent/conf.d /etc/datadog-agent/checks.d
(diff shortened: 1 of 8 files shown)


tests/molecule/datadog.default/prepare.yml
```diff
@@ -0,0 +1,79 @@
...
+  tasks:
+    - name: Ensure dpkg-dev is present (for dpkg-scanpackages)
+      ansible.builtin.apt:
+        name: dpkg-dev
+        state: present
+        update_cache: true
+
+    # The role installs the datadog-agent package from apt.datadoghq.com. Serve
+    # a fake one from a local repo so the role apt task installs it rather than
+    # hitting Datadog's repo on every CI run.
```
AmitPhulera, line 14: Should that be an issue? I think it would make tests more comprehensive if we pulled the actual package rather than faking it.

AmitPhulera: I see support for it added in dd786179db10d5f663c4a3bcff54e52fb9042312

AmitPhulera: Q - Would it make sense to make default to be installing datadog agent instead of stubbed version?

tests/molecule/datadog.default/verify.yml
```diff
@@ -10,6 +10,7 @@
     - name: Assert stub package version was kept (real repo not used)
       ansible.builtin.shell: dpkg-query -W -f='${Version}' datadog-agent | grep -Fq "999:1.0.0"
       changed_when: false
+      when: lookup('env', 'MOLECULE_DATADOG_REAL_INSTALL') != '1'
```
AmitPhulera: Should this be documented somewhere?

AmitPhulera (approved):
Looking good. I have one question - non blocking that I have added above.

tests/molecule/datadog.default/prepare.yml
millerdev: I think it's better to use the stubbed version by default. When I asked Claude about it, it also agreed that it's better not to hammer apt.datadoghq.com in CI jobs. It noted that this can be seen as abusive behavior and some repos have been known to block IP ranges to prevent it.

`MOLECULE_DATADOG_REAL_INSTALL=1` was specifically added to allow occasional local tests to verify a full package install as a sanity check.

